PR #26219, opened 2026-01-15
feat(eslint-config-fluid)!: make package flat-config-only

tylerbutler:
## Description

Makes `@fluidframework/eslint-config-fluid` flat-config-only. The package root now exports the named flat configs (`base`, `recommended`, `strict`, `strictBiome`) and server configs directly for `eslint.config.mts` consumers, while retaining the explicit `flat.mts` and `server.mts` subpaths.

This removes the legacy `.eslintrc` CommonJS entry points (`index.js`, `base.js`, `recommended.js`, `strict.js`, and `strict-biome.js`) and drops `@eslint/eslintrc`, since `FlatCompat` is no longer needed. The README and changelog now document the flat-config usage and migration impact.

## Breaking Changes

Legacy `.eslintrc` consumers must migrate to ESLint flat config using `eslint.config.mts`; the removed CommonJS config entry points are no longer supported.

## Reviewer Guidance

The review process is outlined on [this wiki page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

PR #26017, opened 2025-12-10
feat(build-tools): add additionalConfigFiles and ${repoRoot} token for task file tracking

tylerbutler:
## Summary

Adds two related features to the fluid-build incremental build system:

- **`additionalConfigFiles` property** on `TaskFileDependencies` — lets task definitions declaratively track extra configuration files (e.g., shared root-level eslint or biome configs) for incremental build invalidation, with `"..."` spread syntax for package-level extension of global definitions
- **`${repoRoot}` token** for file paths and globs in task definitions — allows referencing repository-root files without fragile depth-dependent relative paths; supported in `additionalConfigFiles`, `inputGlobs`, and `outputGlobs`

These replace the hardcoded shared-config scanning in `EsLintTask` with a general-purpose, declarative mechanism available to all task types.

### Implementation details

- `LeafWithDoneFileTask` base class gains an `additionalConfigFiles` getter that resolves paths from the task definition, integrating into both timestamp-based and hash-based done file comparisons
- `TscDependentTask` now composes task-specific configs (via abstract `taskSpecificConfigFiles`) with additional configs from the task definition in a cached `configFileFullPaths` getter
- `EsLintTask` simplified: removed hardcoded `getSharedConfigFiles()` in favor of the new declarative approach
- `DeclarativeLeafTask` resolves `${repoRoot}` tokens in `inputGlobs`/`outputGlobs` before passing them downstream

## Note to reviewers

This PR is "soft-blocking" https://github.com/microsoft/FluidFramework/pull/26219. However, this is a huge change _just_ to unblock that issue, but I believe this is a generally useful feature and will enable us to keep even more custom task definitions out of the fluid-build codebase while retaining the incremental builds. Moreover, should we move to another system, most I have used have similar concepts to those introduced here, for the same reasons. My preference is to get this merged (it's been open for 2 months) rather than a throwaway smaller change, but if this is hard to review, I'll go that route.

PR #26384, opened 2026-02-09
fix(eslint-config-fluid): add tsconfigRootDir and node types for IDE support

tylerbutler:
## Summary

- Adds `tsconfigRootDir: process.cwd()` to the shared `useProjectService` parser options, fixing a VS Code ESLint error where typescript-eslint's auto-detection finds multiple candidate tsconfig root directories as files from different monorepo packages are visited in the same session.
- Adds a `tsconfig.json` to the eslint-config-fluid package for IDE support (TypeScript type-checking of `.mts` config files and scripts).
- Adds `@types/node` as a dev dependency so `process.cwd()` and other Node.js APIs resolve correctly in the config source.
- Strips `tsconfigRootDir` from printed configs since the value is environment-specific.
- Lockfile deduplication: removes stale typescript-eslint 8.53.0 entries that were superseded by 8.54.0.


common/build/eslint-config-fluid/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+	// This tsconfig is for IDE support only - the package doesn't compile TypeScript.
+	// Scripts run via tsx which handles ESM/CJS interop automatically.
+	"extends": "../build-common/tsconfig.base.json",
+	"compilerOptions": {
+		// Override target for modern Node.js features (import.meta, etc.)
+		// "target": "ES2022",
+		// "lib": ["ES2022"],
```
tylerbutler: Verify that this is needed - doesn't seem to be?

common/build/eslint-config-fluid/scripts/print-configs.ts
```diff
@@ -101,6 +101,11 @@ async function generateConfig(filePath: string, config: FlatConfigArray): Promis
 		delete cleanConfig.languageOptions.globals;
 	}
 
+	// Remove tsconfigRootDir since it varies by environment (it's set to process.cwd())
+	if (cleanConfig.languageOptions?.parserOptions?.tsconfigRootDir) {
```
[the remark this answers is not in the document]
jason-ha: I think `typeof cleanConfig.languageOptions?.parserOptions === "object"` would be a sufficient check.

common/build/eslint-config-fluid/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+
+		// Enable Node.js types - base config sets "types": [] which excludes all @types packages
+		"types": ["node"],
+
+		// Use modern module settings for ESM (.mts files) and tsx runtime
+		// Differs from standard Node16 to allow flexible import syntax that tsx handles
+		"module": "ESNext",
```
jason-ha, line 11: Can we fix this package? I don't see why it couldn't be ESM and declare exports in package.json such that it is very clear. Is nice to have good examples in most places. (Would definitely be separate from this PR.)

common/build/eslint-config-fluid/tsconfig.json
jason-ha: Remove commented out lines?

jason-ha (approved):
Some notes for clean up but solution appears to work at least for me.

FWIW there are TS complaints in print-configs.ts for configsToPrint that can be address with shallow copy. For example:
`		config: [...recommended],`

common/build/eslint-config-fluid/scripts/print-configs.ts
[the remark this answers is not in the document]
tylerbutler:
```suggestion
	if (typeof cleanConfig.languageOptions?.parserOptions === "object") {
```

common/build/eslint-config-fluid/tsconfig.json
tylerbutler: I don't think it can be ESM-only and use exports until we remove the old configs. Or, rather, going ESM-only is a breaking change, in progress in #26219. That includes exports and whatnot.

But that is in turn mildly blocked on https://github.com/microsoft/FluidFramework/pull/26017, because without it (or a different smaller change), every lint task outputs this in the terminal:

```
Warning: Expected shared ESLint config file not found: /Volumes/Code/Fluid-workspace/scratch/common/build/eslint-config-fluid/index.js. ESLint cache invalidation may not work correctly.
Warning: Expected shared ESLint config file not found: /Volumes/Code/Fluid-workspace/scratch/common/build/eslint-config-fluid/base.js. ESLint cache invalidation may not work correctly.
Warning: Expected shared ESLint config file not found: /Volumes/Code/Fluid-workspace/scratch/common/build/eslint-config-fluid/strict.js. ESLint cache invalidation may not work correctly.
Warning: Expected shared ESLint config file not found: /Volumes/Code/Fluid-workspace/scratch/common/build/eslint-config-fluid/recommended.js. ESLint cache invalidation may not work correctly.
Warning: Expected shared ESLint config file not found: /Volumes/Code/Fluid-workspace/scratch/common/build/eslint-config-fluid/minimal-deprecated.js. ESLint cache invalidation may not work correctly.
Warning: Expected shared ESLint config file not found: /Volumes/Code/Fluid-workspace/scratch/common/build/eslint-config-fluid/strict-biome.js. ESLint cache invalidation may not work correctly.
```

Josmithr (approved):

common/build/eslint-config-fluid/tsconfig.json
jason-ha: I don't know why you'd need to change the package surface to go ESM. #26219 removes all of the individual configs, but you don't have to do that. You would just need to list all of the exports indicating which imports are allowed.

PR #26385, opened 2026-02-09
fix(eslint-config-fluid): add tsconfigRootDir to all package eslint configs

tylerbutler:
## Summary

- Adds `tsconfigRootDir: import.meta.dirname` to the `parserOptions` in every package-level `eslint.config.mts` across the monorepo (160 files).
- Fixes the "No tsconfigRootDir was set" error from typescript-eslint's parser, which occurs in the monorepo when the parser encounters multiple candidate `tsconfig.json` roots and cannot resolve the correct one automatically.
- For configs that used a single-line array spread (e.g. `[...strict]`), the array is expanded to multi-line with the new config object appended. For configs that already had a multi-line array, the new config object is appended as the last entry.

### DIFF
--- a/azure/packages/azure-local-service/eslint.config.mts
+++ b/azure/packages/azure-local-service/eslint.config.mts
@@ -6,6 +6,15 @@
 import type { Linter } from "eslint";
 import { strict } from "../../../common/build/eslint-config-fluid/flat.mts";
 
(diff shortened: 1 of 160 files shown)


jason-ha:
I thought you were trying to avoid this big change?
Are there cases that aren't dealt with by the other changes?

PR #26392, opened 2026-02-09
chore(eslint-config-fluid): configure no-empty-object-type allowInterfaces and remove overrides

tylerbutler:
## Summary

- Configures `@typescript-eslint/no-empty-object-type` in the shared ESLint config with `allowInterfaces: "with-single-extends"`, permitting the common pattern of empty interfaces extending a single interface.
- Removes per-package `no-empty-object-type: "off"` overrides from `build-tools`, `property-changeset`, and `tree` that are no longer needed with the relaxed shared config.
- In the `tree` package, replaces the blanket `"off"` with a stricter local override that additionally allows `allowObjectTypes: "always"` for its `{}` type usages.
- Removes `eslint-disable` comments for `no-empty-object-type` across tree source and test files where the rule no longer flags (empty interfaces with single extends, and `{}` object types now permitted by the local override).
- Adds one targeted `eslint-disable` comment on `MakeNominal` (an intentionally empty nominal-brand interface) in `typeCheck.ts`.
- Updates all printed config snapshots to reflect the new rule options.


taylorsw04:
@tylerbutler I don't think we want to enable the rule as-is, in general or in tree, but definitely not in tree. All of the occurrences I saw in tree take the form of Foo extends Bar<T1, ...Tn> {}, which is AFAICT not the intended target of this rule given that this actually specializes the type. At the very least, we should configure it to allow interfaces with a single extends (`allowInterfaces?:
      | 'never'
      | 'with-single-extends'
      /** Whether to allow empty interfaces. */
      | 'always';`
but perhaps we even want to always allow this in the case of interfaces.

CraigMacomber:
> @tylerbutler I don't think we want to enable the rule as-is, in general or in tree, but definitely not in tree. All of the occurrences I saw in tree take the form of Foo extends Bar<T1, ...Tn> {}, which is AFAICT not the intended target of this rule given that this actually specializes the type. At the very least, we should configure it to allow interfaces with a single extends (`allowInterfaces?: | 'never' | 'with-single-extends' /** Whether to allow empty interfaces. */ | 'always';` but perhaps we even want to always allow this in the case of interfaces.

We have cases in tree which we do want this check for, but they are already suppressed.

I agree with Taylor though, these new cases look like false positives to me.

There are two kinds of empty interfaces:
- Interfaces who's type is the "empty object" type. (We might want to lint these)
- Interfaces which extend something that is not the "empty object" type, but have no additional members. (I think we should allow these)

I don't get why the no-empty-object rule is flagging interfaces which are that second case, as they typing has nothing to do with empty object types. I think the setting Taylor suggested might fix that (and maybe miss linting interfaces with no members which extend an  "empty object" type, but I'm fine with that. As ErasedTypes which are triggering this are a utility mostly used in tree, but supported outside of it, it might be good for that tweak to go in the default setting of that rule. Creating interfaces like that is a useful thing to do from a devX perspective as it gives better compile errors and IntelliSense in some cases so I think we should allow them with our lint config.

noencke (approved):
🌲

PR #26399, opened 2026-02-10
feat(client-fluid-framework): expose getPresence from fluid-framework

jason-ha:
- Move support to `@fluidframework/fluid-static`, re-export from `fluid-framework`, and deprecate from `@fluidframework/presence`
- Update fluid-static exports to support alpha and beta
- `fluid-framework` only exports `getPresence*`. Callers can access remaining surface from `@fluidframework/presence`.
- Update example and test uses


packages/framework/presence/src/getPresence.ts
```diff
@@ -161,6 +161,8 @@ const ContainerPresenceFactory = {
  * @returns the {@link Presence}
  *
  * @beta
+ *
+ * @deprecated Import from `fluid-framework/beta` instead. The export to be removed in 2.100.0 release.
```
alexvy86:
```suggestion
 * @deprecated Import from `fluid-framework/beta` instead. This export will be removed in the 2.100.0 release.
```

packages/framework/presence/src/getPresence.ts
```diff
@@ -170,6 +172,8 @@ export const getPresence: (fluidContainer: IFluidContainer) => Presence = getPre
  * @returns the {@link PresenceWithNotifications}
  *
  * @alpha
+ *
+ * @deprecated Import from `fluid-framework/alpha` instead. The export to be removed in 2.100.0 release.
```
alexvy86:
```suggestion
 * @deprecated Import from `fluid-framework/alpha` instead. This export will be removed in the 2.100.0 release.
```

packages/framework/presence/src/getPresence.ts
```diff
@@ -15,23 +15,23 @@ import { assert, fail } from "@fluidframework/core-utils/internal";
 import type { IFluidContainer } from "@fluidframework/fluid-static";
 import { isInternalFluidContainer } from "@fluidframework/fluid-static/internal";
 import type {
-	ExtensionCompatibilityDetails,
+	ExtensionCompatibilityDetails as getPresenceAlphaDeprecated,
 	FluidDataStoreContextInternal,
 	IFluidDataStoreContext,
 } from "@fluidframework/runtime-definitions/internal";
```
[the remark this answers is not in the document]
jason-ha: These are weird merges with origin unknown. Reverting.

packages/framework/presence/src/getPresence.ts
```diff
@@ -161,6 +161,8 @@ const ContainerPresenceFactory = {
  * @returns the {@link Presence}
  *
  * @beta
+ *
+ * @deprecated Import from `fluid-framework/beta` instead. The export is to be removed in 2.100.0 release.
```
alexvy86:
```suggestion
 * @deprecated Import from `fluid-framework/beta` instead. This export will be removed in the 2.100.0 release.
```

packages/framework/presence/src/getPresence.ts
```diff
@@ -170,6 +172,8 @@ export const getPresence: (fluidContainer: IFluidContainer) => Presence = getPre
  * @returns the {@link PresenceWithNotifications}
  *
  * @alpha
+ *
+ * @deprecated Import from `fluid-framework/alpha` instead. The export is to be removed in 2.100.0 release.
```
alexvy86:
```suggestion
 * @deprecated Import from `fluid-framework/beta` instead. This export will be removed in the 2.100.0 release.
```

packages/service-clients/end-to-end-tests/azure-client/src/test/multiprocess/childClient.tool.ts
```diff
@@ -19,6 +19,7 @@ import { LogLevel } from "@fluidframework/core-interfaces";
 import type { ScopeType } from "@fluidframework/driver-definitions/legacy";
 import type { ContainerSchema, IFluidContainer } from "@fluidframework/fluid-static";
 import {
+	// eslint-disable-next-line import-x/no-deprecated -- TODO#59157: relocating to fluid-static
```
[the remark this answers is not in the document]
alexvy86: Do we need the export in fluid-static? Users of `azure-client` are expected to also depend on `fluid-framework`, no? So they could get the API from there.

packages/framework/fluid-framework/eslint.config.mts
```diff
@@ -6,6 +6,14 @@
...
 
-const config: Linter.Config[] = [...strict];
+const config: Linter.Config[] = [
+	...strict,
+	{
+		files: ["src/index.ts"],
+		rules: {
+			"import-x/order": "off",
+		},
+	},
```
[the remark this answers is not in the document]
alexvy86: I also don't see why disable the rule here. If it's to keep the new lines in `src/index.ts` together when auto-ordering would interleave them with other things, seems better to just wrap them in `/* eslint-disable import-x/order */` and the corresponding `...enable...`.

.changeset/beige-candles-sleep.md
```diff
@@ -0,0 +1,14 @@
+---
+"fluid-framework": minor
+"@fluidframework/presence": minor
+"__section": deprecation
+---
+getPresence is being relocated to fluid-framework package away from @fluidframework/presence
```
alexvy86: Nit
```suggestion
getPresence is being relocated from @fluidframework/presence to the fluid-framework package
```

alexvy86 (commented):
Some doc-related comments

shlevari (approved):

packages/framework/fluid-framework/src/index.ts
```diff
@@ -83,6 +83,35 @@ export * from "@fluidframework/tree/alpha";
...
+export const getPresence = getPresenceDeprecated;
+
+/**
+ * {@inheritdoc @fluidframework/presence#getPresenceAlpha}
+ *
+ * @alpha
+ */
+export const getPresenceAlpha: (
+	// eslint-disable-next-line @typescript-eslint/consistent-type-imports -- complicated redeclaration required to satisfy api-extractor
+	fluidContainer: import("@fluidframework/fluid-static").IFluidContainer,
```
Josmithr: I really think we need to avoid doing these inline imports.

packages/framework/fluid-framework/eslint.config.mts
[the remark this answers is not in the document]
jason-ha: I don't think the rule makes sense for the file. I can disable for the custom re-exports region, which is the only part of the file with imports and should only ever be the region with imports. So that is effectively the whole file. Making the change...

.changeset/beige-candles-sleep.md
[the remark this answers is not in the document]
jason-ha: I had those tick but removed them. I thought we didn't want ticks in the title (for reason I didn't question).?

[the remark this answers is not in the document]
jason-ha: Oh wait - this isn't a human comment. @Josmithr or @alexvy86, can you confirm if ticks should be used or not?
Also, should a period be added?
(The comma doesn't seem controversial.)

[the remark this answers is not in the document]
Josmithr: Yeah, our current guidance is to not use inline code syntax in changeset titles to prevent noisy looking headers in the resulting changeset. I think this is a policy we could discuss as a team if folks feel strongly that we _should_ allow them (I have mixed feelings personally). But for now, the guidance is no.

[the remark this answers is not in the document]
alexvy86: What Josh said, and I'll add that there shouldn't be a trailing period.

packages/framework/presence/src/getPresence.ts
jason-ha: I was trying to avoid "will" because I've seen things slip and "is to be" felt sufficiently wishy washy. :)

.changeset/beige-candles-sleep.md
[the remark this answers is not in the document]
jason-ha: The one reason I know to allow them is to avoid the @ mentions that happen without them. We've mentioned `@alpha` and `@beta` in the past. Or maybe we've fixed that since.?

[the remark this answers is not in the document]
Josmithr: Oh, interesting. I wasn't aware of that issue. You can always escape the `@` if needbe though. At least I would assume that works.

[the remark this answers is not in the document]
alexvy86: I don't think the issue is fixed. That's a minor reason why I personally don't mind the backticks, but not a battle I care about fighting 😄 . Somehow trying to escape the `@` had never occurred to me... it does work when used directly in comment boxes here (the picker doesn't come up). Hopefully it would work in the rendered release notes.

packages/framework/fluid-framework/src/index.ts
jason-ha: I have gone through and made all of the changes that avoiding this requires.
See https://github.com/jason-ha/FluidFramework/pull/9.
It does lead to some unfortunate `{@link}` qualifications, but we can live with those.
cc @alexvy86 
LMK what you think.

packages/service-clients/end-to-end-tests/azure-client/src/test/multiprocess/childClient.tool.ts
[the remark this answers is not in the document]
jason-ha: I need to relocate it outside of the `presence` package. `fluid-static` is the best home I can think of. Otherwise, I will create a new package. `fluid-framework` can continue to have the export but it will import from `fluid-static` instead.
I can't move `getPresence` straight to `fluid-static` because `presence` currently depends on `fluid-static`. There will be presence package refactor to do the other shuffling. But in order to deprecate it, I need a replacement. Hence this staging.

packages/framework/presence/src/getPresence.ts
alexvy86: Personally I would go with "will be removed no earlier than ..." in that case. "Is to be" left me unsure if it was addressing Fluid consumers of Fluid devs, with my brain leaning towards the latter.

packages/framework/fluid-framework/src/index.ts
alexvy86: As much as I don't like how many types end up in the API report in that PR, aren't they supposed to be there? If we do the inline imports here without exporting all those extra types, doesn't that leave the API surface incomplete in a way where its consumers might not have all the required type definitions for parameters, return types, etc, that they could encounter?

And I don't quite understand how things are complete/correct today in the presence package itself, where API reports have references to (for example) `DeepReadonly` but don't export it, like the API report for `fluid-framework` ends up doing in the PR. Am I missing something there?

packages/framework/presence/src/getPresence.ts
jason-ha: That is definitely a consumer note (that devs should of course be aware of). `@deprecation` notes are published on our docs site. I'd put dev notes in privateRemarks.

packages/framework/fluid-framework/src/index.ts
jason-ha: 1. They'd still have everything they'd need. Types aren't required to be exported for them to be "visible" to consumers.
If we had this:
```ts
interface Foo { prop: string };
interface Bar { prop: number };
export declare function getBarFromFoo(arg: Foo): Bar;
```
whomever imports `getBarFromFoo` will still see the definition for `arg` and return type but they won't be able to use `Foo` or `Bar` directly.
```ts
import { getBarFromFoo } from "api";
type ExtractedBar = ReturnType<typeof getBarFromFoo>; // = Bar = { prop: number; }
```
In general, I think it is a good idea to reveal things but there are plenty of times it is just a maintenance burden. There are a set of things that should never be needed but still are forced to be exported. I *think* a reason for `api-extractor` aggressiveness is to support `api-extractor`'s roll-up pattern that redeclares API. Those hidden APIs could be included in the rollup but I don't know that it does that and instead relies on forcing reveal. We aren't using roll-ups for other reasons; so, I haven't looked at things in a long time.
I don't think any of the things in the various `Internal*` namespaces need to be exported. There are a handful of others I'd keep out.
For `fluid-framework` specifically, it is also possible for consumers to use other packages directly.
Personally, I'd prefer to keep all of the `/alpha` things out of `fluid-framework` and have consumers interested in that use subpackages. (I will likely remove `getPresenceAlpha` from `fluid-framework` in a little while. I included it to be able to successfully deprecate from `presence` package and still provide non-deprecated access. In the end it will only appear from `fluid-static` package.)

Josmithr: > They'd still have everything they'd need. Types aren't required to be exported for them to be "visible" to consumers.
> If we had this:
> interface Foo { prop: string };
> interface Bar { prop: number };
> export declare function getBarFromFoo(arg: Foo): Bar;
> whomever imports getBarFromFoo will still see the definition for arg and return type but they won't be able to use Foo or Bar directly.

I disagree with this being a reasonable status quo. We don't know exactly how a consumer is using our code. They may want to be able to reference Foo or Bar directly for any number of reasons that we can't predict. Forcing them to use implicit typing or manually derive types using patterns like `ReturnType<getBarFromFoo>` is not a good developer model IMO.

jason-ha: I think it definitely depends. There are some things that the only reasonable (maintainable) approach is not use the explicit type. I have plenty of experiences where code does nothing with a type except to carry it from one place to another, then the ends change but the middle is broken. In some but not all of those cases, some form of type erasure would have been best. There are certainly many places where we should be clear.
I am not opposed to exposing things if there is customer use. I do prefer to hold things back if there isn't need. Presence is being used successfully for Boards and Designer now without any need for the 70 or so exports from core-interfaces that are being exposed now via fluid-framework.

jason-ha: 2. `DeepReadonly` is itself fine. But I was surprised by the `DeepReadonly` collateral that was required. Something in the linting was missing those and I have plugged the hole with [improvement(client-core-interfaces): ensure complete internal/exposedUtilityTypes exports](https://github.com/jason-ha/FluidFramework/pull/9/commits/085eeffabe0f16400a8ca8e261affc2e5eb3f1dd) in the referenced PR.

I also updated that PR with renames of the core-interfaces and presence Internal* namespace which cleaned up most of the `{@link}` ugliness.

I think presence got away with not exposing things because the referenced APIs come from core-interface and those are considered okay external references. But fluid-framework is a full aggregator and isn't allowed to rely on other packages.

Josmithr: > I think it definitely depends. There are some things that the only reasonable (maintainable) approach is not use the explicit type. I have plenty of experiences where code does nothing with a type except to carry it from one place to another, then the ends change but the middle is broken.

> I am not opposed to exposing things if there is customer use. I do prefer to hold things back if there isn't need. Presence is being used successfully for Boards and Designer now without any need for the 70 or so exports from core-interfaces that are being exposed now via fluid-framework.

I don't think we should be in the business of deciding when this is / is not appropriate. Whether or not to consume our types seems best left to the consumer. We take away that choice if we don't export them.

> In some but not all of those cases, some form of type erasure would have been best. There are certainly many places where we should be clear.

100%

jason-ha: We should definitely be deciding. There is a clear minimal surface best practice for platforms. We had also better be well-informed of how our APIs are used in order to provide the best service. As long as we are diligent about where the lines are and how things might shift, we can provide a better API surface. Note that we were diligent with those ~70 APIs and while I'd rather no customer need concern themselves with them, if they get out, their quality is good. Even so I don't expect any customer to reference and use them correctly.
None of those can be type-erased as their whole value is in knowing their whole definition.

jason-ha: I've decided to go ahead and walk back the `getPresenceAlpha` deprecation and not expose any alpha APIs from fluid-framework at this time. That can be seen in [jason-h#9](https://github.com/jason-ha/FluidFramework/pull/9). If we keep the inline, then I'll move that here.

jason-ha: I think my last thought on this is that the flat exposure of everything from fluid-framework seems crazy. I see lots of `tree` stuff that looks generic but is so `tree` specific (example: `asAlpha`). The entire `presence` API surface is not here currently. I know there are about 27 APIs that are not exported. No one would be able to use `getPresence` in a meaningful way without also using `@fluidframework/presence` based on what is exposed in the PR right now. Long-term `fluid-framework` should either `export * from "@fluidframework/presence"` like `tree` package or it should have slim exports, IMO.

Josmithr: > I think my last thought on this is that the flat exposure of everything from fluid-framework seems crazy. I see lots of tree stuff that looks generic but is so tree specific (example: asAlpha).

1000% percent agree. I pushed for us to do something different when we were prepping for 2.0, but I was outvoted. The global namespace of this package is crazy.

This is something we should revisit before 3.0.

jason-ha: I don't think I will pursue exposure from `fluid-framework` so as not to exacerbate that surface funkiness.
In either case I have extracted some of the API cleanup and verification to #26435

packages/service-clients/end-to-end-tests/azure-client/src/test/multiprocess/childClient.tool.ts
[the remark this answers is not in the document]
jason-ha: `presence` has been refactored and now `getPresence` is implemented in `fluid-static`.

packages/framework/fluid-framework/src/index.ts
jason-ha: The funkiness has all been resolved with the noted PR and recent the presence refactor.

.changeset/beige-candles-sleep.md
```diff
@@ -0,0 +1,14 @@
+---
+"fluid-framework": minor
+"@fluidframework/presence": minor
+"__section": deprecation
+---
+getPresence is being relocated from @fluidframework/presence to the fluid-framework package
+
+To prepare, make changes following this pattern:
```
Josmithr, line 10: Nit: I think we should expand this a bit. Something like "The export from `@fluidframework/presence` is now deprecated and will be removed in a future release. Please update any existing imports as follows:"

Josmithr (commented):
I think I'm missing some context on the changes here. The user-facing aspects of this look good to me, and the docs look good (aside from 1 nitpick).

But the move to `fluid-static` is a bit surprising. The scope / intention of that package isn't well documented, but I always viewed it as the APIs needed to implement a service client. Since the service clients don't use presence (right?), that placement seems a bit weird to me.

I'm assuming there were some external conversations that led to this outcome?

jason-ha:
> I think I'm missing some context on the changes here. The user-facing aspects of this look good to me, and the docs look good (aside from 1 nitpick).
> 
> But the move to `fluid-static` is a bit surprising. The scope / intention of that package isn't well documented, but I always viewed it as the APIs needed to implement a service client. Since the service clients don't use presence (right?), that placement seems a bit weird to me.
> 
> I'm assuming there were some external conversations that led to this outcome?

I would look at this thread: https://github.com/microsoft/FluidFramework/pull/26399#discussion_r2790425991.

I will post extra notes there.

packages/service-clients/end-to-end-tests/azure-client/src/test/multiprocess/childClient.tool.ts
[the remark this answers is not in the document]
jason-ha: @Josmithr, fluid-static is the package that pulls together the "encapsulated" layers and enables the "declarative" use pattern. In particular, it implements `FluidContainer` that is the host for `Presence` connection as it is for DDSes.

I am guessing that `IFluidContainer` may get extended to just have a `presence` member. But for now, we let customers demand `Presence` via `getPresence` but are placing that in the vicinity.

If `presence` is the general `Presence` package, it must not pull in "encapsulated" or "declarative" specific things. The `getPresence*` access functions violated that. I am not sure where the "encapsulated" accessor will land, but that is probably a new (small) package.

.changeset/beige-candles-sleep.md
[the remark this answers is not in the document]
jason-ha: I have added ticks since the mention issue isn't addressed.

packages/service-clients/end-to-end-tests/azure-client/src/test/multiprocess/childClient.tool.ts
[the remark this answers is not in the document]
jason-ha: I made draft of what the alternative is using a new package. See #27004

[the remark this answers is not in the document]
Josmithr: > I am guessing that IFluidContainer may get extended to just have a presence member. But for now, we let customers demand Presence via getPresence but are placing that in the vicinity.

This seems reasonable to me. And I personally don't think the overhead of an extra package is worth it here.

Josmithr (approved):

Issue #26383, opened 2026-02-07
Performance mark and measure entries are never cleaned up, accumulating indefinitely in long-running host applications


cosmincatrinescu:
## Description

`PerformanceEvent.timedExecAsync` in `@fluidframework/container-loader` creates `performance.mark()` and `performance.measure()` entries but never cleans them up after the operation completes. This causes performance entries to accumulate indefinitely in long-running applications.

## Impact

In applications that load/unload Fluid containers multiple times, these entries accumulate:
- After loading 100 containers: ~600+ orphaned marks/measures
- In long-running sessions: thousands of entries

This contributes to memory pressure and degrades DevTools performance when inspecting the Performance timeline.

## Affected Events

| Event Name | Location |
|------------|----------|
| `Resolve` | `loader.js:92` |
| `Load` | `container.js:122` |
| `CodeLoad` | `container.js:1527` |
| `InstantiateRuntime` | `container.js:1571` |
| `WaitOps` | `container.js:1065` |
| `WaitOpProcessing` | `container.js:1066` |

## Reproduction

```javascript
// Load a Fluid container 3 times, then check:
performance.getEntries().filter(e => ['mark', 'measure'].includes(e.entryType))
// Result: 66 entries (6 events × 3 loads × ~3-4 marks/measures each)

tylerbutler:
@cosmincatrinescu Do you have a repro of this? Just doing console testing, ~100K markers in a single trace doesn't seem to have any impact. At ~1M there is some minor slowdown but nothing like what you're reporting. Do you have a test harness that shows this behavior? Not arguing we don't have a bug here, but it's hard to square my obsrvations with what you're reporting.

PR #23668, opened 2025-01-27
Replace IsomorphicPerformance with a performanceNow() method

TommyBrosman:
IsomorphicPerformance is an internal API in client-utils. Currently it includes a number of methods supported by the [Performance](https://developer.mozilla.org/en-US/docs/Web/API/Performance) API, including `mark()` and `measure()`. The only member of the API that is currently used is `now()`. This change removes the other members and reduces the amount of work needed to restructure client-utils. Note that there is another copy of IsomorphicPerformance in common-utils. The common-utils version is deprecated, so this PR does not change it.

PR #26401, opened 2026-02-10
fix(telemetry-utils): prevent performance mark/measure entry accumulation

tylerbutler:
## Summary

- **Fixes** https://github.com/microsoft/FluidFramework/issues/26383
- `PerformanceEvent.cancel()` now clears orphaned start marks from the Performance API — previously it set `this.event = undefined` without any cleanup, leaking the start mark forever
- `performanceEndMark()` now calls `clearMarks()`/`clearMeasures()` after creating the measure, so entries don't accumulate on the success path either
- `timedExec()` and `timedExecAsync()` use a `finally` block for scope-bound cleanup, ensuring marks are always cleared regardless of code path


packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -780,7 +788,21 @@ export class PerformanceEvent {
 		if (this.startMark !== undefined && this.event) {
 			const endMark = `${this.event.eventName}-end`;
 			window.performance.mark(endMark);
-			window.performance.measure(`${this.event.eventName}`, this.startMark, endMark);
+			window.performance.measure(this.event.eventName, this.startMark, endMark);
```
ChumpChief: Not strictly necessary for this PR, but probably would be good to make this isomorphic.  I see that #23668 stripped away relevant methods from `IsomorphicPerformance`, but you might be able to get away with just replacing `window` with `globalThis` and call it a day?

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -780,7 +788,21 @@ export class PerformanceEvent {
...
+			window.performance.clearMarks(this.startMark);
+			window.performance.clearMarks(endMark);
+			window.performance.clearMeasures(this.event.eventName);
```
ChumpChief: I'm not super familiar with the usage scenario here, but doesn't this defeat the purpose of marking/measuring?  Like doesn't this prevent the customer from actually observing the measurement?

Maybe this instead wants some other trigger for clearing marks/measures (sliding time window, count, explicit invocation, memory-pressure-based, etc.)?

tylerbutler:
Marking as draft while I work through the feedback.

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -729,6 +737,37 @@ export class PerformanceEvent {
...
+				for (const name of PerformanceEvent.pendingMarkCleanup) {
+					globalThis.performance.clearMarks(name);
+				}
+				for (const name of PerformanceEvent.pendingMeasureCleanup) {
+					globalThis.performance.clearMeasures(name);
+				}
+				PerformanceEvent.pendingMarkCleanup.clear();
+				PerformanceEvent.pendingMeasureCleanup.clear();
+				PerformanceEvent.cleanupScheduled = false;
+			}, 10_000);
```
ChumpChief: Since this is not a sliding window but rather kind of a "chunk" clear, it seems like it might have the potential to be frustrating to use (e.g. hard to control whether some measure under investigation spans a clear).  Marks that were made even a few milliseconds prior to the scheduled clear time would be cleared.

Looking back at the issue filed, are we sure it was diagnosed correctly? 600 marks should not have a major impact on devtools performance.  Might be worth checking the repro first, and if nothing else it'll be good for validating the fix against.

If it was diagnosed correctly though I'd probably suggest a sliding window, which might just look like a double-buffer of items for cleanup (thereby guaranteeing that every mark will live for at least N seconds).

tylerbutler: I'll look into a repro.

PR #26405, opened 2026-02-10
Add composable labels for transactions

daesunp:
## Description

This PR adds a composable `labels` type that keeps track of all labels in nested transactions.


packages/dds/tree/src/core/rebase/types.ts
```diff
@@ -234,6 +235,17 @@ export interface LocalChangeMetadata extends CommitMetadata {
...
 	 */
 	readonly label?: unknown;
+
+	/**
+	 * A label tree auto-composed from nested transaction labels.
+	 * Each nesting level that provides a {@link RunTransactionParams.label | label} contributes
+	 * a node, with inner transaction labels becoming children of outer ones.
+	 *
+	 * @remarks
+	 * This is only defined when the outermost transaction has a label.
```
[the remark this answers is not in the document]
noencke: Agreed

packages/dds/tree/src/util/labelTree.ts
```diff
@@ -0,0 +1,78 @@
...
+		this.label = label;
+		this.children = children;
+	}
+
+	/**
+	 * Returns `true` if this tree contains the given label at this node or any descendant node.
+	 *
+	 * @remarks Comparison uses reference equality (`===`).
+	 */
+	public has(label: T): boolean {
```
noencke:
```suggestion
	public has(label: T, areEqual: (a: T, b: T) => boolean = Object.is): boolean {
```
What do you think of this? Seems like a nice to have and it's low cost for us to implement

packages/dds/tree/src/util/labelTree.ts
```diff
@@ -0,0 +1,78 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * A tree of labels that supports recursive search and Set-like read operations.
```
noencke: Interestingly, there is nothing in this class that is actually specific to labels. It's really just a tree of values. So you have a choice:

1. Rename this to `ValueTree`, and adjust its documentation accordingly. It's now a generic tree of values (no mention of labels here) and can be used for other features besides labels, if ever desired.
2. Keep it as `LabelTree`, but double-down on the documentation. Talk about labels here, and how they are used in transactions, and cross-link to the other label APIs, etc. Basically, if you're going to make this a piece of code that can only be used by one feature (transaction labels), then benefit from that by strongly connecting it to that feature via docs and links.

packages/dds/tree/src/util/labelTree.ts
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @remarks
+ * Each node in the tree has a label of type `T` and zero or more children.
+ * The tree is immutable after construction.
```
noencke:
```suggestion
```
IMO just say "An immutable tree of..." in the summary doc, that's just as good.

packages/dds/tree/src/util/labelTree.ts
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @sealed @alpha
+ */
+export class LabelTree<T = unknown> {
```
noencke: Prefer to export an _interface_ called `LabelTree` and then you can have it implemented here by a class (`LabelTreeNode` or some other contrived name, whatever) - but the class isn't exported outside of the package. It probably won't matter, but it's usually best practice since it's easier for us to make compatible API changes if we're dealing with an interface vs. a class.

packages/dds/tree/src/test/util/labelTree.spec.ts
```diff
@@ -0,0 +1,108 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { strict as assert } from "node:assert";
+
+// Allow importing from this specific file which is being tested:
+/* eslint-disable-next-line import-x/no-internal-modules */
+import { LabelTree } from "../../util/labelTree.js";
```
noencke:
```suggestion
import { LabelTree } from "../../util/index.js";
```
I think? Then you don't need the eslint ignore above?

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -396,6 +398,24 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+	private transactionLabels?: LabelTree;
+
+	/**
+	 * Stack of label frames for composing a {@link LabelTree} from nested transactions.
+	 *
+	 * @remarks
+	 * Each frame tracks the label for that nesting level and the {@link LabelTree} children
+	 * accumulated from completed inner transactions.
+	 */
+	private readonly labelFrameStack: { label: unknown; children: LabelTree[] }[] = [];
```
noencke: It's unfortunate that we have both `transactionLabels` _and_ `labelFrameStack`. Is it necessary?

If you follow [the suggestion here](https://github.com/microsoft/FluidFramework/pull/26405/changes#r2819221923), you should be able to accomplish this with an additional tweak. Have the customer-facing _interface_ be immutable, but have the implementation class (used in this file) be mutable. Then, you can just have a single field with a label tree and update it as the transaction open and close.

packages/dds/tree/src/core/rebase/types.ts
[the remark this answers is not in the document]
daesunp: updated to suggested change

packages/dds/tree/src/util/labelTree.ts
daesunp: updated to suggested change

packages/dds/tree/src/util/labelTree.ts
daesunp: Updated to ValueTree

packages/dds/tree/src/util/labelTree.ts
daesunp: Updated to `ValueTree` and docs

packages/dds/tree/src/util/labelTree.ts
daesunp: Updated to `ValueTree` and to an interface implemented by a class

packages/dds/tree/src/test/util/labelTree.spec.ts
daesunp: updated to suggested change

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Removed labelFrameStack

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -399,22 +398,16 @@ export class TreeCheckout implements ITreeCheckoutFork {
 	private transactionLabel?: unknown;
```
noencke: This is redundant now, no? Anywhere using `transactionLabel` should be able to use `currentLabelTreeNode?.value` instead.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -399,22 +398,16 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
-
-	/**
-	 * Stack of label frames for composing a {@link LabelTree} from nested transactions.
-	 *
-	 * @remarks
-	 * Each frame tracks the label for that nesting level and the {@link LabelTree} children
-	 * accumulated from completed inner transactions.
-	 */
-	private readonly labelFrameStack: { label: unknown; children: LabelTree[] }[] = [];
+	private currentLabelTreeNode: ValueTreeNode | undefined;
```
noencke: This says that it points to the root of the label tree, but it actually points to the most recent leaf? I think?

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -468,58 +461,79 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
-	 * When a transaction commits, the label and accumulated children from inner transactions
-	 * are composed into a {@link LabelTree} node. That node is added as a child of the parent
-	 * frame (if nested) or set as {@link TreeCheckout.transactionLabels} (if outermost).
+	 * When a transaction commits, the cursor moves up to the parent node.
+	 * If the current node has no label, its children are promoted to the parent.
+	 * When the outermost transaction commits, {@link TreeCheckout.currentLabelTreeNode}
+	 * is left pointing to the root of the composed tree (or cleared if no labels exist).
 	 *
-	 * When a transaction aborts, the frame is simply discarded.
+	 * When a transaction aborts, the node is removed from its parent and the cursor moves up.
```
noencke: Cursor? I think I know what this is trying to say, but it's not clear what a cursor is at a glance.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -468,58 +461,79 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+	 * Pops a label node and composes the label tree.
 	 *
 	 * @param committed - Whether the transaction was committed (true) or aborted (false).
 	 *
 	 * @remarks
-	 * When a transaction commits, the label and accumulated children from inner transactions
-	 * are composed into a {@link LabelTree} node. That node is added as a child of the parent
-	 * frame (if nested) or set as {@link TreeCheckout.transactionLabels} (if outermost).
+	 * When a transaction commits, the cursor moves up to the parent node.
+	 * If the current node has no label, its children are promoted to the parent.
```
noencke: What? How can the current node have children? Wouldn't the current node always be a leaf? I'm so confused lol

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -468,58 +461,79 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+	 * Called at the start of each transaction (including nested ones).
+	 * Creates a new {@link ValueTreeNode} as a child of the current node (if any)
+	 * and moves the cursor down to it. When the transaction completes,
+	 * {@link TreeCheckout.popLabelFrame} moves the cursor back up.
 	 */
 	public pushLabelFrame(label: unknown): void {
-		this.labelFrameStack.push({ label, children: [] });
+		const node = new ValueTreeNode(label, this.currentLabelTreeNode);
+		if (this.currentLabelTreeNode !== undefined) {
+			this.currentLabelTreeNode.children.push(node);
```
noencke: Having to update the parent of one node and the children of another - or else an invariant is broken - is a drawback of this "doubly-linked" scheme. Why doesn't the ValueTreeNode do this automatically? If you pass in an array of children, it will link both ways, but if you pass in a single parent, it will only link upwards.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -468,58 +461,79 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
 
-		const parent = getLast(this.labelFrameStack);
-		if (frame.label !== undefined) {
-			const node = new LabelTree(frame.label, frame.children);
-			if (parent === undefined) {
-				// Outermost level — set the composed tree for the commit metadata.
-				this.transactionLabels = node;
+		if (current.value === undefined) {
+			if (parent !== undefined) {
+				// No label at this level — promote children to parent.
```
noencke: This is an optimization to make the tree minimally sized but I don't think it's worth the code complexity. It only _happens_ rarely - when users have outer transactions with labels and also inner transactions without labels - and it only _matters_... never, probably. We would need hundreds or thousands of transactions inside of each other before any of this even begins to show up on a profiler.

packages/dds/tree/src/util/valueTree.ts
```diff
@@ -0,0 +1,110 @@
...
+ * while the class still satisfies the readonly {@link ValueTree} interface for consumers.
+ */
+export class ValueTreeNode<T = unknown> implements ValueTree<T> {
+	public readonly value: T;
+	public readonly children: ValueTreeNode<T>[] = [];
+	public parent: ValueTreeNode<T> | undefined;
+
+	public constructor(
+		value: T,
+		parentOrChildren?: ValueTreeNode<T> | readonly ValueTreeNode<T>[],
```
noencke: Let's just do away with the parent pointers - see rationale in other comments. Data structures are so much simpler if they have clear, 1-way ownership. Parent nodes own the child nodes, no uppointers.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -447,6 +460,82 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+	 */
+	public popLabelFrame(committed: boolean): void {
+		const current = this.currentLabelTreeNode;
+		if (current === undefined) {
+			return;
+		}
+
+		const parent = current.parent;
+
+		if (!committed) {
```
noencke: Why do we need to special case aborts? If an abort happens, we should pop the frame. If a commit happens, we should apply the commit and then pop the frame. But I don't see why the frarme-popping itself should care, if that makes sense.

noencke (changes requested):
This whole "double linking" scheme is making the code significantly more complicated and hard to read than it would otherwise be.

1. Why do we need parent pointers? Because the "cursor" is pointing to the _bottom_ of the tree, so we need to be able to climb back up it to find the other nodes.
2. Why do we need to point to the bottom of the tree? Because we want to access the "current" label node in O(1). 
3. Why do we want to access the current label node in O(1)? Because... why not?
4. Why not? Because it makes the code way more complicated. Let's do something else.

`currentLabelTreeNode` - which we should rename to just `labelTreeNode` - can simply always point to the root node of the tree. Whenever you want to find the "current" node, walk down the right side of the tree. I.e. just do `node.children.at(-1).children.at(-1).children.at(-1)...` until you reach the bottom. That is the current node. A negligible performance impact (except in pathological cases), and in exchange we don't need parent pointers, which will dramatically simplify ValueNode and this push/pop code.

Same logic for removing the "make the tree as flat as possible" optimization. Everything is just way simpler if you have the LabelTree always be isomorphic to the actual transaction tree, whether there are labels or not. Even if there are no labels at all, but lots of transactions, still make the label tree have a node for every transaction. In the very unlikely event that this ever proves to be practically inefficient, we can optimize it at that time.

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated to point to root

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated confusing doc comment

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated confusing doc comment

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: removed parent pointer

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: removed unnecessary optimization

packages/dds/tree/src/util/valueTree.ts
daesunp: removed parent pointer

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: removed special case for aborts

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated to use currentLabelTreeNode.value

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -389,12 +390,23 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+	 */
+	private labelTreeNode: ValueTreeNode | undefined;
+
+	/**
+	 * Tracks which label tree nodes belong to transactions that are still in progress.
+	 * Used by {@link TreeCheckout.currentLabelNode} to walk past committed (closed) nodes
+	 * and find the deepest open node.
 	 */
-	private transactionLabel?: unknown;
+	private readonly openLabelNodes = new WeakSet<ValueTreeNode>();
```
noencke: This should be `static`.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -447,6 +459,90 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+		}
+	}
+
+	/**
+	 * Returns the deepest open node on the right side of the label tree, which corresponds
+	 * to the current (most recently pushed, not yet committed or aborted) transaction.
+	 * Only valid to call when {@link TreeCheckout.labelTreeNode} is defined.
+	 */
+	private currentLabelNode(): ValueTreeNode {
+		assert(this.labelTreeNode !== undefined, "Expected label tree to exist");
```
noencke: nit: I think best practice would be to have this return `ValueTreeNode | undefined`. The callers should determine whether to do the assert instead. For example, `pushLabelFrame` would assert before getting the `currentLabelNode`, but `closeLabelFrame` and `popLabelFrame` would tolerate the `undefined` and return early (as they currently do).

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -447,6 +459,90 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+			}
+			node = lastChild;
+		}
+		return node;
+	}
+
+	/**
+	 * Returns whether any node in the label tree has a defined value.
+	 */
+	private hasDefinedLabels(): boolean {
```
noencke: This is a curious optimization. Is there a strong rationale for it? I don't see any real harm in having the labelTree be present, even if there are no labels anywhere in it. `has` will still work (returning false always).

In fact... do you think we should _always_ provide a tree? It seems inconvenient from the user's perspective that they have to check whether `labels` is undefined. What if it were always defined, no matter what? Maybe `labels` should actually point to a "shell" object, which has the operations on it (like `has`, `values`, etc.). The shell would also have a `tree` property, which would contain what is currently `labelTreeNode`. The nodes themselves would then not necessarily need to have `has`, `values`, etc. - those could just be on the top-level shell.

Or, when firing the event, if `labelTreeNode` is undefined, you would instead just return an empty `ValueTree`. That's a little simpler, but also slightly inconsistent, because the shape of the value tree that the user gets no longer always exactly corresponds to the transaction structure. They can't tell the difference between a transaction that ran with no label and no transaction at all. Maybe we don't care.

What do you think of these different approaches?

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -447,6 +459,90 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+			}
+		}
+		return false;
+	}
+
+	/**
+	 * Marks the current label node as closed when a transaction commits.
+	 * The node remains in the tree so it appears in the final {@link LocalChangeMetadata.labels}.
+	 */
+	public closeLabelFrame(): void {
```
noencke: All of this method's behavior is already present in `popLabelFrame`. I would delete this and just add an `aborted` flag to `popLabelFrame`.

noencke (commented):
This is looking much cleaner!

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated `openLabelNodes` to static

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated return type

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Decided to go with your first suggestion :)

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Updated `popLabelFrame` and removed `closeLabelFrame`

packages/dds/tree/src/shared-tree/treeCheckout.ts
noencke: Okay. What was your reasoning? What pros and cons did you consider when making your decision?

If the value can be potentially undefined, `has` is still easy: `labels?.has(...)`.
values is slightly worse, but not by much, since you can do: `labels?.values() ?? []`.

packages/dds/tree/src/core/rebase/types.ts
```diff
@@ -234,6 +235,19 @@ export interface LocalChangeMetadata extends CommitMetadata {
 	 * This can be used by undo/redo to group or classify edits.
 	 */
 	readonly label?: unknown;
+
+	/**
+	 * A value tree auto-composed from nested transaction labels.
```
noencke: 1. The reader probably doesn't know what a "value tree" means
2. "auto-composed" isn't a particularly useful phrase. Auto, as opposed to manual? A reader won't know what composed means either, probably.
3. "nested transaction labels" - but also un-nested transaction labels. Nesting isn't required - this could be misleading.

noencke: The summary should be descriptive to a user, not necessarily descriptive of the mechanics of the thing. Like, what really is this thing? Why would I want it?

"A collection of labels for all transactions (nested or otherwise) that made up this change". Something like that.

Then in the remarks you can give the details and examples. Describe how the labels contribute to the tree structure, show an example of using `has`, etc.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -447,6 +459,75 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+	private currentLabelNode(): ValueTreeNode | undefined {
+		if (this.labelTreeNode === undefined) {
+			return undefined;
+		}
+		let node: ValueTreeNode = this.labelTreeNode;
+		while (node.children.length > 0) {
+			const lastChild = node.children[node.children.length - 1];
+			assert(lastChild !== undefined, "Expected label tree node to have children");
+			if (!TreeCheckout.openLabelNodes.has(lastChild)) {
+				break;
```
noencke, line 530: According to this logic you will break the first time that you find a last child that is closed, and will use its parent as the current node. When you find this closed-last-child, is it not the case that **it is always the most recently closed transaction**?

If it is indeed always the most recently closed transaction, then you do not actually need a WeakSet. You only need a single pointer to the "most recently closed transaction", which you update on pop.

Put another way, suppose you've marked 5 transactions in the tree as "closed" (by removing them from the set). But you could only possibly traverse to 1 of those 5, I think. That's the one to point to.

packages/dds/tree/src/shared-tree/treeCheckout.ts
noencke: That said, it still seems maybe an unnecessary implementation detail being exposed. Suppose that `labels` was simply a `ReadonlySet<unknown> & { tree?: ValueTree }`. We could pretty easily offer that, and it'd be more familiar to the user. You _always_ get a Set, but it might be empty. This is more normalized than "you _might_ get a Set, but you have to check in case you didn't, but if you _do_, then it still only _might_ have something in it". It's better (normalized) to say either "you always get a set, but it might be empty" or "you might get set, but it will always have something in it" (and the former is generally preferable).

Or, here is yet another option:

* ValueTree doesn't have `has` or `forEach` - all it has is `values()`. This is _honest_, because `has()` is not O(1) - if a user is bummed that they have to iterate over the values rather than having instant lookup, well, we were iterating under the hood anyway.
* `labels` is an object of type `Iterable<unknown> & { tree?: ValueTree }`.

This leaves it up to the user what to do with the values - if they want to put them in a set, they can with `new Set(labels)`. If they're just looking for a single label, they can do `[...labels].includes("foo")` or if they have 30,000 labels and don't want to spread they can write a for loop lol.

These are all different design decisions with different tradeoffs. I'm not yet sure which I prefer, but I'd love to hear your thoughts!

packages/dds/tree/src/core/rebase/types.ts
daesunp: Updated docs and provided examples

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Updated with suggestions

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated to `ReadonlySet<unknown> & { tree?: ValueTree }` after discussion off call :)

packages/dds/tree/src/util/valueTree.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Mutable implementation of {@link ValueTree}.
+ *
+ * @remarks
+ * Used internally to incrementally build a value tree as nested scopes open and close.
+ * The mutable `children` array allows in-place tree construction,
+ * while the class still satisfies the readonly {@link ValueTree} interface for consumers.
+ */
+export class ValueTreeNode<T = unknown> implements ValueTree<T> {
```
noencke: Do we need this class anymore? :) It doesn't do anything except copy the child array when constructing - probably not useful.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -108,6 +111,52 @@ import { SharedTreeChangeFamily, hasSchemaChange } from "./sharedTreeChangeFamil
...
 
+/**
+ * Shared empty set used as the `labels` value for remote changes and non-transactional local changes.
+ */
+const emptyLabelsSet: ReadonlySet<unknown> = new Set<unknown>();
+
+/**
+ * Recursively collects all defined (non-`undefined`) values from a {@link ValueTree} into a target set.
+ */
+function collectTreeValues(node: ValueTree, target: Set<unknown>): void {
```
noencke: I think this would be better as a generator that yields values. That is a more general and powerful function with a simpler implementation. And the usage for new sets is just `new Set(collectTreeValues(node))`

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -108,6 +111,52 @@ import { SharedTreeChangeFamily, hasSchemaChange } from "./sharedTreeChangeFamil
...
+	return false;
+}
+
+/**
+ * Builds the labels set for a change event from the label tree.
+ * If the tree exists and contains at least one defined label, returns a set of all
+ * values with the tree attached. Otherwise returns an empty set.
+ */
+function buildLabelsSet(labelTreeNode: ValueTreeNode | undefined): TransactionLabels {
+	if (labelTreeNode === undefined || !hasDefinedLabel(labelTreeNode)) {
```
noencke: If the goal of doing this "pre-walk" is to avoid allocating the net Set below, then I think we could make it simpler.

1. Start with the set undefined `let set: Set<unknown> | undefined;`.
2. [ Use an iterator](https://github.com/microsoft/FluidFramework/pull/26405/changes#r2861404392) to walk the nodes `for const (node of collectTreeValues(labelTreeNode)) {`
3. If/when you find a node, do `(set ??= new Set()).add(node)`.
4. Return `set ?? emptyLabelSet`.

That way you only loop once, no need for `hasDefinedLabel`, but you still only allocate the new set if necessary.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -389,12 +438,27 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+	 */
+	private labelTreeNode: ValueTreeNode | undefined;
+
+	/**
+	 * Points to the most recently closed (committed) label node on the right spine of the tree.
+	 * Used by {@link TreeCheckout.currentLabelNode} to stop descending past committed nodes.
 	 *
 	 * @remarks
-	 * This label is used to implement {@link TreeCheckout.runWithTransactionLabel}.
+	 * Only one closed node is ever reachable during the right-spine traversal, so a single
```
noencke: nit: I'd prefer to see this kind of comment next to the actual algorithm, and the docs for this are more just like "used by {@link currentLabelNode}"

packages/dds/tree/src/util/valueTree.ts
daesunp: Removed class, but had to add 2 casts (`as ValueTree[]`) since `ValueTree.children` are readonly, so `push/pop` operations aren't available. But if you prefer to prevent casting, I could create a simple interface which is a mutable version of `ValueTree` and use that instead :)

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated `collectTreevalues` to return a generator and removed `hasDefinedLabels`

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated with suggestions

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: moved comment next to the algorithm

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -117,44 +116,34 @@ import type { ISharedTreeEditor, SharedTreeEditBuilder } from "./sharedTreeEditB
...
 /**
  * Builds the labels set for a change event from the label tree.
  * If the tree exists and contains at least one defined label, returns a set of all
  * values with the tree attached. Otherwise returns an empty set.
  */
-function buildLabelsSet(labelTreeNode: ValueTreeNode | undefined): TransactionLabels {
-	if (labelTreeNode === undefined || !hasDefinedLabel(labelTreeNode)) {
+function buildLabelsSet(labelTreeNode: ValueTree | undefined): TransactionLabels {
+	if (labelTreeNode === undefined) {
 		return emptyLabelsSet;
```
noencke: nit: this function has three returns, two of which are duplicated and are each empty cases. You could structure the function as 

if labelTreeNode !== undefined {
  if set !== undefined {
    return Object.assign(...)
  }
}
return emptyLabelSet

to collapse to just two returns.

packages/dds/tree/src/util/valueTree.ts
noencke: Good question. A type conundrum (compile time) is usually not a good reason to make a runtime change. A mutable version would be a good solution.

But it might be even simpler than that. The `changed` event is the only way that a user is passed one of these things, right? We pass them our `checkout.labelTreeNode`. But then, we clear `checkout.labelTreeNode` (set it to undefined) before we ever edit it further, yes? So, we don't mutate a `ValueTree` after we pass it to the user. In which case, who cares if they mutate it. In fact, it might even be best practice to let them, because it's an output type - not an input type - that they take exclusive ownership of. Us telling them they can't modify it, when we don't even have a reference to it anymore, is perhaps silly and annoying.

So if you want, you can just make ValueTree mutable *shrug* both inside and out *shrug*. If we decide we need an immutable version later for some other theoretical feature, we can make the second type at that time.

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Updated with suggested change

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -133,17 +133,16 @@ function* collectTreeValues(node: ValueTree): IterableIterator<unknown> {
...
-	}
-	if (set === undefined) {
-		return emptyLabelsSet;
+	if (labelTreeNode !== undefined) {
+		let set: Set<unknown> | undefined;
+		for (const value of collectTreeValues(labelTreeNode)) {
+			(set ??= new Set()).add(value);
+		}
+		if (set !== undefined) {
+			return Object.assign(set, { tree: labelTreeNode });
```
noencke: nit: Can you just declare the type of `set` as `TransactionLabels | undefined`, and then just do `set.tree = labelTreNode` here rather than `Object.assign`? Not sure if that works but it'd be slightly nicer.

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1468,6 +1470,11 @@ export type TransactionConstraint = NodeInDocumentConstraint;
 // @alpha @sealed
 export type TransactionConstraintAlpha = TransactionConstraint | NoChangeConstraint;
 
+// @alpha @sealed
+export type TransactionLabels = ReadonlySet<unknown> & {
```
noencke: If `ValueTree` is no mutable, do you want `TransactionLabels` to be mutable too? Should the `tree` property be mutable? Should the `ReadonlySet` be a `Set`? If so, you would need to create a new Set every time (you couldn't use your empty set singleton) but you could make it lazy, e.g.:

```ts
let set: Set<unknown> | undefined;
this.#events.emit("changed", {
  ...
  get labels(): {
    set ??= new Set();
    Object.assign(set, { tree: labelsTree });
    return set
  }
});
```

But I give this example to be academic - I think you should just always create a new set, even if its empty. From copilot:

**Prompt [How performant is constructing an empty set in JS? | Try in Copilot Chat](https://m365.cloud.microsoft/chat/entity1-d870f6cd-4aa5-4d42-9626-ab690c041429/eyJpZCI6IlZYTmxjbFl4ZkdoMGRIQnpPaTh2YzNWaWMzUnlZWFJsTFhOa1ppMXBiblF1YjJabWFXTmxMbU52YlM5OFQwbEVPbVkwWmpBMk5HRTBMVEU1TW1NdE5HWmhNUzFoWTJObExUQmtZMkU0TTJWaU5UY3dZbnd5WkRSbVlXVXdOUzFoWVdNMUxUUmhZVFV0T1Raa1pTMWxOV015WkRBek1UQmpNekI4TWpBeU5pMHdNaTB5TjFReE56b3dPRG93T0M0M09UYzBNRGRhIiwic2NlbmFyaW8iOiJzaGFyZUxpbmtWaWFSaWNoQ29weSIsInByb3BlcnRpZXMiOnsicHJvbXB0U291cmNlIjoidXNlciIsImNsaWNrVGltZXN0YW1wIjoiMjAyNi0wMi0yN1QxNzowODowOC4wNzBaIn0sImNoYXRUeXBlIjoid29yayIsInZlcnNpb24iOjEuMX0)**
How performant is constructing an empty set in JS?
**Response**
Short answer: **constructing an empty `Set` in JavaScript is extremely cheap**—so cheap that in almost all real code it’s effectively noise.

Here’s the more precise, performance‑oriented breakdown you’d expect as a JS/engine nerd.

***

## What actually happens when you do `new Set()`

For modern engines (V8, SpiderMonkey, JavaScriptCore):

*   `new Set()` allocates:
    *   A small JS object header
    *   An *empty* internal hash table (often lazily initialized or with a tiny initial capacity)
*   **No hashing, no buckets populated, no iteration**
*   Time complexity: **O(1)**
*   Allocation size: **very small** (tens of bytes)

V8 exposes this explicitly at the API level: `v8::Set::New(isolate)` is documented as “Creates a new empty Set” with no work beyond allocation. [\[v8.github.io\]](https://v8.github.io/api/head/classv8_1_1Set.html)

***

## How fast is it in practice?

While exact numbers depend on CPU, engine version, and GC state, the qualitative facts are consistent:

*   You can construct **millions of empty Sets per second**
*   It is comparable in cost to:
    *   `{}`
    *   `[]`
    *   `new Map()`

In other words, if `new Set()` shows up in a profile, **something else is wrong** (e.g. you’re doing it in a hot loop unnecessarily).

Benchmarks and engine discussions consistently focus on **operations on populated Sets** (add / has / delete), not construction cost, because construction is negligible relative to any meaningful work. [\[stackoverflow.com\]](https://stackoverflow.com/questions/39007637/javascript-set-vs-array-performance), [\[dev.to\]](https://dev.to/abdulghaffar349/optimizing-javascript-collections-comparing-sets-and-arrays-e07)

***

## Relative cost vs alternatives

Rough ordering (fastest → slowest), all O(1):

```js
{}            // object literal
[]            // array literal
new Set()
new Map()
```

But the gap between them is **tiny**. The difference is measured in **nanoseconds**, not microseconds.

Choosing between these should be based on:

*   semantics (uniqueness, iteration order, lookup complexity)
*   downstream operations

—not construction cost.

***

## When `new Set()` *can* matter

The only time it becomes relevant:

### ❌ Bad pattern (hot path)

```js
for (...) {
  const seen = new Set(); // repeated millions of times
  ...
}
```

If the loop is extremely hot and the Set escapes or grows, allocation + GC pressure can show up.

### ✅ Better pattern

```js
const seen = new Set();
for (...) {
  seen.clear(); // often cheaper than reallocating
}
```

Even then, **`clear()` vs reallocate** is workload-dependent and usually dominated by how many elements were in the Set.

***

## TL;DR

*   ✅ `new Set()` is **O(1)** and **very fast**
*   ✅ Cost is comparable to `{}` or `[]`
*   ✅ Safe to use freely in normal code
*   ❌ Only worth thinking about in extreme hot loops or allocation-heavy paths

If you’re worried about this, you’re already operating at a level where **algorithmic behavior and allocation patterns matter far more than constructor cost**.

If you want, I can:

*   show a microbenchmark pattern that *actually* reveals Set construction cost
*   compare `new Set()` vs reuse + `clear()` under different workloads

daesunp: Updated `TransactionLabels` to be mutable and removed `emptySet`

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Updated to just make the set every time after updating with suggestions from other comment

packages/dds/tree/src/core/rebase/types.ts
```diff
@@ -234,8 +235,66 @@ export interface LocalChangeMetadata extends CommitMetadata {
...
+	 * like {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has | has}
+	 * to check for specific labels.
+	 *
+	 * The optional {@link ValueTree | tree} property provides the structural nesting of the transactions.
+	 * Each transaction contributes a node whose {@link ValueTree.value} is its label
+	 * (or `undefined` if no label was provided).
+	 * When transactions are nested, inner transaction nodes become children of outer ones.
+	 *
+	 * The `tree` property is present whenever the change was produced by a transaction that
+	 * includes at least one defined (non-`undefined`) label. If all transactions are unlabeled,
```
noencke:
```suggestion
	 * includes at least one label. If the change was unlabeled,
```

packages/dds/tree/src/core/rebase/types.ts
```diff
@@ -234,8 +235,66 @@ export interface LocalChangeMetadata extends CommitMetadata {
...
+	 */
+	readonly labels: TransactionLabels;
 }
 
+/**
+ * A set of transaction labels with an optional structural tree.
+ *
+ * @remarks
+ * The set contains all label values from the transactions that produced the change.
+ * Use standard {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set | Set}
```
noencke, line 304: Nah, remove this - customers should know how to use a JS Set :)

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -447,6 +489,83 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
+				break;
+			}
+			node = lastChild;
+		}
+		return node;
+	}
+
+	/**
+	 * Pops the current label frame from the label tree.
+	 *
```
noencke:
```suggestion
```

packages/dds/tree/src/util/valueTree.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * A tree where each node holds a value of type `T` and zero or more children.
+ *
+ * @sealed @alpha
+ */
+export interface ValueTree<T = unknown> {
```
noencke: Now that this interface is so trivial, and no longer has a class associated with it, or any functionality, I think we should reconsider whether we want it to be "generic". We originally made it a "value" tree instead of a "label" tree because then other features could possibly use it in the future. This made sense because we had implemented various methods and logic and it's nice to re-use those and make them into a "mini library". 

However, this is now nothing more than a type with two properties. In which case, I think the benefits of having nice names/docs that are specific to the transaction feature outweighs it being generic. If you agree, I'd suggest something like:

```ts
interface LabelTree {
  label: unknown;
  sublabels: LabelTree[];
}
```

This may also let you move some of the documentation. I expect some things you had to write out on the long `labels` doc comment can be moved to this type and its properties, and you can link to it.

packages/dds/tree/src/core/rebase/types.ts
daesunp: Updated with suggested change

packages/dds/tree/src/core/rebase/types.ts
daesunp: removed :)

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated with suggested change

packages/dds/tree/src/util/valueTree.ts
daesunp: Removed `ValueTree` interface and replaced with suggested interface

noencke (approved):

PR #26408, opened 2026-02-11
feat(common-utils,protocol-definitions): upgrade to ESLint 9 flat config

tylerbutler:
## Summary

Upgrades `common-utils` and `protocol-definitions` from ESLint 8 with `@fluidframework/eslint-config-fluid` to standalone ESLint 9 flat configs.

- **protocol-definitions**: Replaces `.eslintrc.cjs` and `.eslintignore` with `eslint.config.mjs` using `@eslint/js`, `typescript-eslint`, `@rushstack/eslint-plugin`, and `eslint-plugin-unicorn`. Enables `@rushstack/no-new-null` and `unicorn/text-encoding-identifier-case` so existing inline disable comments remain valid. Auto-fixes index signatures to `Record<>` style.
- **common-utils**: Replaces `.eslintrc.cjs` with `eslint.config.mts` using the same core stack plus `eslint-plugin-import`. Restores all comments and rules from the old config (`import/no-nodejs-modules`, `import/no-deprecated`, unicorn overrides, test override). Enables rules from the shared config that are referenced by inline `eslint-disable` comments (`@rushstack/no-new-null`, `no-bitwise`, `default-case`, `no-restricted-syntax`, etc.).
- Both configs include a TODO: [AB#59243](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/59243) to migrate back to the shared config once it supports ESLint 9.


common/lib/protocol-definitions/src/storage.ts
```diff
@@ -132,8 +132,8 @@ export interface ITree {
  */
 export interface ISnapshotTree {
 	id?: string;
-	blobs: { [path: string]: string };
-	trees: { [path: string]: ISnapshotTree };
+	blobs: Record<string, string>;
```
tylerbutler, line 135: Autofix got applied unintentionally. Can revert if needed. I prefer Record myself.

ChumpChief (approved):

common/lib/protocol-definitions/src/storage.ts
jason-ha: I think there are plenty of places where explicit index notation is better.
The improvement is that you get to say a bit about the intent of the index. Here it was clearly a path string and not just a string. With `Record` that is lost. So, I would say this is one of those places. I am going to propose a revert of these changes - #26505
There are lots of examples in presence where we want to keep the named index.

PR #26413, opened 2026-02-11
feat!: remove public API surface for external custom DDS implementations

tylerbutler:
> [!NOTE]
> **A minimal alternative is available:** [#26439](https://github.com/microsoft/FluidFramework/pull/26439) achieves the same "no custom DDSes" goal by only internalizing `SharedObject` and `SharedObjectCore` (2 files, 7 lines changed). This PR is the maximal version that also internalizes the runtime/factory layer. See the comparison table below.

## Summary

Removes the public API surface that enabled external custom DDS implementations, making these types `@internal`:

**`@fluidframework/shared-object-base`**: `SharedObject`, `SharedObjectCore`, `ISharedObjectKind`, `createSharedObjectKind` — base classes and factory interfaces for DDS implementations

**`@fluidframework/datastore-definitions`**: `IChannelFactory` — the factory interface required to register custom DDSes

**`@fluidframework/datastore`**: `FluidDataStoreRuntime`, `ISharedObjectRegistry`, `mixinRequestHandler`, `mixinSummaryHandler` — runtime infrastructure for DDS hosting

**`@fluidframework/aqueduct`**: `DataObjectFactory` — the high-level factory that wires up DDSes with data objects

**`@fluidframework/test-runtime-utils`**: `MockContainerRuntime`, `MockContainerRuntimeFactory`, `MockFluidDataStoreRuntime` — test mocks that depend on the now-internal types

All existing DDSes (SharedTree, SharedMap, SharedCell, etc.) continue to work via `SharedObjectKind`, which remains public and sealed. The `@fluid-experimental/tree` package is updated to `@internal` release tags throughout since it depends on the now-internal `SharedObject` base class.

~45 example and test files are updated to import the affected types from `/internal` entry points instead of `/legacy`.

## Comparison: maximal (this PR) vs minimal ([#26439](https://github.com/microsoft/FluidFramework/pull/26439))

| | Maximal (this PR) | Minimal (#26439) |
|---|---|---|
| **Files changed** | 113 | 2 |
| **Lines changed** | +750 / -2026 | +7 / -64 |
| **Breaking change packages** | 5+ (shared-object-base, datastore-definitions, datastore, aqueduct, experimental tree) | 1 (shared-object-base) |
| **Example files changed** | ~41 | 0 |
| **Symbols internalized** | ~15+ (SharedObject, SharedObjectCore, IChannelFactory, ISharedObjectKind, ISharedObjectRegistry, FluidDataStoreRuntime, DataObjectFactory, PureDataObjectFactory, TreeDataObjectFactory, mixinRequestHandler, mixinSummaryHandler, experimental tree types) | 2 (SharedObject, SharedObjectCore) |
| **Can create custom DDS?** | No | No |
| **Can iterate on IChannelFactory?** | Yes (now internal) | No (still legacy beta) |
| **Can iterate on FluidDataStoreRuntime?** | Yes (now internal) | No (still legacy beta) |

### When to use which

- **#26439 (minimal)** if the only goal is preventing custom DDSes with minimal disruption
- **This PR (maximal)** if the team also wants freedom to iterate on the runtime/factory layer without legacy compatibility constraints


examples/apps/diceroller/src/container/main.tsx
```diff
@@ -4,7 +4,9 @@
  */
 
 import type { EventEmitter } from "@fluid-example/example-utils";
-import { DataObject, DataObjectFactory } from "@fluidframework/aqueduct/legacy";
+// eslint-disable-next-line import-x/no-internal-modules
+import { DataObjectFactory } from "@fluidframework/aqueduct/internal";
```
CraigMacomber, line 8: We are adding a lot of cases of examples doing things we don't want customers to do.

If our own examples can't manage to show how customers are supposed to use our APIs, we might need to adjust our examples and/or API.

Also DataObjects are a totally different thing from SharedObjects, and we allow subclassing and extending them as a supported and heavily used legacy API: I don't think we should be impacting the data object API surface as part of this change.

Maybe we have to tweak the data object APIs to avoid leaking/depending on shared object or something like that, but I think that should be its own separate change, as its rather high risk and likely to break existing applications and thus need a revert and rework.

experimental/dds/tree/src/TransactionInternal.ts
```diff
@@ -1026,7 +1026,7 @@ export namespace TransactionInternal {
 
 	/**
 	 * Error thrown when a detach operation is given an invalid or malformed Range
-	 * @alpha
+	 * @internal
```
CraigMacomber, line 1029: This makes no sense as part of this change. Hiding the error types experimental tree probably should not be done, but certainly seems out of scope for this PR.

packages/dds/ordered-collection/src/consensusOrderedCollectionFactory.ts
```diff
@@ -62,13 +62,13 @@ export class ConsensusQueueFactory implements IConsensusOrderedCollectionFactory
 
 /**
  * {@inheritDoc ConsensusQueueClass}
- * @legacy @beta
+ * @internal
```
CraigMacomber, line 65: Removing the ability to use this DDS seems unrelated to the claimed action of this PR which is to remove the ability of external users to create new DDS kinds. This seems like the opposite of what you want.

experimental/dds/tree/src/TransactionInternal.ts
tylerbutler: SOrry, this is already fixed locally - pushing the latest.

packages/dds/ordered-collection/src/consensusQueue.ts
```diff
@@ -32,7 +32,7 @@ class SnapshotableQueue<T> extends SnapshotableArray<T> implements IOrderedColle
  * Implementation of a consensus stack
  *
  * An derived type of ConsensusOrderedCollection with a queue as the backing data and order.
- * @legacy @beta
+ * @internal
```
CraigMacomber, line 35: If we remove the need to leak this class to the API, we shouldn't need to package export it.

packages/dds/tree/api-report/tree.legacy.beta.api.md
```diff
@@ -116,7 +116,7 @@ export type ConciseTree<THandle = IFluidHandle> = Exclude<TreeLeafValue, IFluidH
 export function configuredSharedTreeBeta(options: SharedTreeOptionsBeta): SharedObjectKind<ITree>;
 
 // @beta @legacy
-export function configuredSharedTreeBetaLegacy(options: SharedTreeOptionsBeta): ISharedObjectKind<ITree> & SharedObjectKind<ITree>;
+export function configuredSharedTreeBetaLegacy(options: SharedTreeOptionsBeta): SharedObjectKind<ITree>;
```
CraigMacomber, line 119: This makes shared tree unusable for legacy API users. This API only exists to expose the APis needed by legacy users: if you remove those you should just remove configuredSharedTreeBetaLegacy entirely instead (of course we can't do either of these changes since there are users).

This API was carefully designed to not leak the SharedObject classes: it should be fine as is.

packages/dds/tree/src/test/simple-tree/api/tree.spec.ts
```diff
@@ -34,7 +38,7 @@ class NodeMap extends schema.map("NoteMap", schema.string) {}
 class NodeList extends schema.array("NoteList", schema.string) {}
 class Canvas extends schema.object("Canvas", { stuff: [NodeMap, NodeList] }) {}
 
-const factory = SharedTree.getFactory();
+const factory = (SharedTree as unknown as ISharedObjectKind<ITree>).getFactory();
```
CraigMacomber, line 41: Don't make tests less type safe as part of removing support for third parties to implement a particular interface.

plans/custom-dds-removal.md
```diff
@@ -0,0 +1,311 @@
+# Custom DDS Removal Plan
+
+## Goals
+
+1. **Reduce public API surface** - Remove DDS base classes and interfaces (`SharedObject`, `SharedObjectCore`, `IChannel`, `IChannelFactory`, etc.) from the public API to prevent external/third-party custom DDS implementations.
```
CraigMacomber, line 5: You don't want to remove the interfaces. You want to make the classes internal and seal the interfaces.

plans/custom-dds-removal.md
```diff
@@ -0,0 +1,311 @@
...
+
+2. **Enable faster iteration** - When these types are public or `@legacy @beta`, changes must move slowly due to compatibility requirements. Making them internal allows the team to iterate more rapidly on DDS infrastructure.
+
+3. **Simplify the API** - Consolidate around SharedTree as the recommended DDS for most use cases, while maintaining existing DDSes for specific scenarios.
```
CraigMacomber, line 9: Pivoting more to shared tree should be out of scope for this work.

plans/custom-dds-removal.md
```diff
@@ -0,0 +1,311 @@
...
+
+---
+
+## Current State Analysis
+
+### DDS Audit Results
+Found **20 total DDS implementations** in the codebase - all are Fluid-owned:
```
CraigMacomber, line 16: Lol, of course the fluid codebase only has fluid owned DDSes.

tylerbutler: Poorly worded. This is intended to represent that all DDSes known "in the wild" are "Fluid-owned."

PR #26415, opened 2026-02-11
ci: Upload all tinylicious.log files correctly

alexvy86:
## Description

Currently when we run the stress tests against tinylicious (in the build pipeline, not in the separate pipeline for stress tests), the step to upload the tinylicious log to the pipeline artifacts [logs a warning](https://dev.azure.com/fluidframework/public/public%20Team/_build/results?buildId=377020&view=logs&j=6bb09b85-9174-5850-be23-a20013a777c6&t=aad6019f-4142-5ba1-d2ba-152116590ec5&l=12) because it's looking for the log file in the wrong place (the path where the file ends up in when the real-svc tests run, but not when the stress tests run).

Also, we're not uploading the tinylicious.log files from other tests that also produce them.

This adds logic to look for all the relevant log files produced in the repo and upload them no matter where they are.

## Reviewer Guidance

### DIFF
--- a/tools/pipelines/templates/build-npm-client-package.yml
+++ b/tools/pipelines/templates/build-npm-client-package.yml
@@ -817,11 +817,20 @@ extends:
                     displayName: Upload tinylicious log
                     condition: always()
                     continueOnError: true # Keep running subsequent tasks even if this one fails (e.g. the tinylicious log wasn't there)
+                    env:
+                      TEST_NAME: '${{ test.name }}'
                     inputs:
                       targetType: inline
                       script: |
                         set -eu -o pipefail
-                        PATH_TO_TINYLICIOUS_LOG=$(Build.SourcesDirectory)/packages/test/test-end-to-end-tests/tinylicious.log;
+                        # Determine package path based on test type
+                        if [[ "$TEST_NAME" == *"stress"* ]]; then
+                          PACKAGE_PATH="packages/test/test-service-load"
+                        else
+                          PACKAGE_PATH="packages/test/test-end-to-end-tests"
+                        fi
+
+                        PATH_TO_TINYLICIOUS_LOG=$(Build.SourcesDirectory)/$PACKAGE_PATH/tinylicious.log;
                         if [ -f $PATH_TO_TINYLICIOUS_LOG ] ; then
                           echo "Found file at '$PATH_TO_TINYLICIOUS_LOG'. Uploading.";
                           echo "##vso[task.uploadfile]$PATH_TO_TINYLICIOUS_LOG";
--- a/tools/pipelines/templates/build-npm-client-package.yml
+++ b/tools/pipelines/templates/build-npm-client-package.yml
@@ -830,8 +830,8 @@ extends:
                           PACKAGE_PATH="packages/test/test-end-to-end-tests"
                         fi
 
-                        PATH_TO_TINYLICIOUS_LOG=$(Build.SourcesDirectory)/$PACKAGE_PATH/tinylicious.log;
-                        if [ -f $PATH_TO_TINYLICIOUS_LOG ] ; then
+                        PATH_TO_TINYLICIOUS_LOG="$(Build.SourcesDirectory)/$PACKAGE_PATH/tinylicious.log";
+                        if [ -f "$PATH_TO_TINYLICIOUS_LOG" ] ; then
                           echo "Found file at '$PATH_TO_TINYLICIOUS_LOG'. Uploading.";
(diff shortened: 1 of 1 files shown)


alexvy86:
@jason-ha this is ready for review now that https://github.com/microsoft/FluidFramework/pull/26450 is closed.

tools/pipelines/templates/build-npm-client-package.yml
```diff
@@ -814,19 +814,43 @@ extends:
 
                 - ${{ if contains(test.name, 'tinylicious') }}:
                   - task: Bash@3
-                    displayName: Upload tinylicious log
+                    displayName: Upload tinylicious log(s)
                     condition: always()
```
jason-ha, line 824: Do we care about the storage size of the new logs that will be accumulated?
I don't know how big these are. I could guess that the stress ones are a lot.?
If we only care about them when there is a failure, then we could condition this on `failure()` instead of `always()`. If we want stress and realsvc jobs captured differently, then the condition could be split for those cases.

tools/pipelines/templates/build-npm-client-package.yml
```diff
@@ -814,19 +814,43 @@ extends:
...
+
+                          # Extract the directory path and filename separately
+                          DIR_PATH=$(dirname "$RELATIVE_PATH")
+                          FILENAME=$(basename "$RELATIVE_PATH")
+
+                          # Note: Azure DevOps only preserves the filename when uploading, not the directory structure,
+                          # so multiple files named "tinylicious.log" would overwrite each other.
+                          # Thus we create a unique name for each log file by using a safe version of its directory path.
+                          # Example: packages/test/foo/tinylicious.log -> packages_test_foo-tinylicious.log
+                          SAFE_NAME="${DIR_PATH//\//_}-${FILENAME}"
```
jason-ha: I think this also need to make unique based on the job. Some packages have both stress and realsvc scripts and those will now collide.

jason-ha:
Please also add some comments explaining why we are uploading these files.
My hope is that with more intention captured when the bots come in, they can do a better job.

tools/pipelines/templates/build-npm-client-package.yml
alexvy86: > Do we care about the storage size of the new logs that will be accumulated?

Not as far as I can tell; storage for Azure Pipelines is all managed by the product itself with no way for us to even see how much is being used, as far as I know. And no documentation or guidelines about how big/small logs should be overall.

> If we only care about them when there is a failure, then we could condition this on failure() instead of always()

In stress tests nothing fails in the pipeline if something went wrong while things were running, we only see error telemetry in Kusto. I know that's true for the stress tests pipeline and I believe it's also the case when we run them in the build pipeline. So a `condition` wouldn't work in that case. For e2e tests... I think those would fail here (the build pipeline), yeah, but not sure if the possibility (and impact) is worth the complication of a separate codepath? Each tinylicious.log file ends up being ~150Mb. Not small, but I'd think our pipelines don't run so frequently that it'd be a problem. Even a few GBs/day worth of these files to me feel like a drop in an ocean compared to what other services/products/clients must be consuming in terms of log storage for their pipeline runs.

I lean towards a simpler approach where we always upload them, but if you still have concerns we can discuss more.

tools/pipelines/templates/build-npm-client-package.yml
alexvy86: In our current setup I don't think we'd see collissions; the two kinds of tests run in separate jobs, thus separate cloud VMs, and wouldn't see each other. But we can future-proof this against changes in that setup. I'll add the sanitized test task to the filename.

alexvy86:
@jason-ha let me know if the new comments match what you had in mind.

jason-ha (commented):
Looks good!
Is there a test run confirming the uploads? (Maybe I don't know where to look.)
I see that [Run Task Test StressTinyliciousTest - Upload tinylicious log(s)](https://dev.azure.com/fluidframework/public/_build/results?buildId=378229&view=logs&j=6bb09b85-9174-5850-be23-a20013a777c6&t=aad6019f-4142-5ba1-d2ba-152116590ec5) says uploading, but when I look at [Artifacts](https://dev.azure.com/fluidframework/public/_build/results?buildId=378229&view=artifacts&pathAsName=false&type=publishedArtifacts) I don't see such a file. Otherwise, I approve.

jason-ha (approved):
We found that log file names are truncated. Probably from zip file limits. These are also logs and not build artifacts (Jason now knows to download Logs)

PR #26417, opened 2026-02-11
tree(test): make staged schema upgrade test more conservative

yann-achard-MS:
## Description

Updates our staged schema upgrade tests so they don't assume that it is safe to read from a tree that has thrown an exception.
It's unclear whether the current production code should put the tree in an explicitly broken state, but future changes (like support for two-step build & attach) may end up doing so. It seems best to keep our tests on the safe side no matter what.

## Breaking Changes

None


CraigMacomber (commented):
Some of copilot's suggestions seem good. Otherwise looks good.

CraigMacomber (approved):

PR #26424, opened 2026-02-12
fix(eslint-config-fluid): resolve TypeScript errors in flat config types

tylerbutler:
## Summary

- Add type assertions in `base.mts` for third-party ESLint plugin types (typescript-eslint, import-x, rushstack, eslint-config-prettier) whose `.d.ts` declarations don't match ESLint 9's updated type interfaces
- Fix `print-configs.ts` to accept readonly config arrays from `as const` exports, removing the need for shallow copies at each call site
- Fix `outputPath` possibly-undefined error
- Remove invalid `indentSize` option from `sortJson` call — `indentSize` only applies to sort-json's file-writing overload, not the in-memory sorting overload used here. Formatting is already handled by `JSON.stringify(sortedConfig, null, 4)` on the next line, so the option was silently ignored at runtime.


common/build/eslint-config-fluid/library/configs/base.mts
```diff
@@ -110,5 +115,5 @@ export const baseConfig: FlatConfigArray = [
 		},
 	},
 	// Prettier disables conflicting rules - must come after custom rules
-	prettierConfig,
+	prettierConfig as Linter.Config,
```
alexvy86: Add a comment like the ones above?

alexvy86 (approved):
One small thing below. I think it would be nice to have references to existing open issues in the new comments but I could not find any. Oh well.

common/build/eslint-config-fluid/library/configs/base.mts
```diff
@@ -59,17 +59,22 @@ export const baseConfig: FlatConfigArray = [
 	// eslint:recommended
 	eslintJs.configs.recommended,
 	// @typescript-eslint/recommended-type-checked and stylistic-type-checked
-	...tseslint.configs.recommendedTypeChecked,
-	...tseslint.configs.stylisticTypeChecked,
+	// Type assertions needed: typescript-eslint's PluginFlatConfig.languageOptions lacks the
+	// string index signature that ESLint core's LanguageOptions requires.
+	...(tseslint.configs.recommendedTypeChecked as Linter.Config[]),
+	...(tseslint.configs.stylisticTypeChecked as Linter.Config[]),
```
jason-ha: Are you certain erasing the specific types from these are correct? I have a strong feeling that issue can be resolved without these casts. We should limit `as` as much as possible.

jason-ha (dismissed):
Putting a light block here - see prior questions

common/build/eslint-config-fluid/library/configs/base.mts
tylerbutler: Do you have something specific in mind?

jason-ha: I looked at the error and I am confused. When I try to inspect things VS Code says that the type of these are
```ts
{
    name?: string;
    rules?: object;
}[]
```
The visible error shown is a complaint about LanguageOptions. That doesn't add up.
If you remove the other entries following those, then there isn't a complaint. I narrowed it down to lines 65, 66, and 72 as contributing to the problem.
If you comment those out, then there is no error.
You can see that importXPlugin is more of the issue. Temporarily adding:
```ts
const X = tseslint.configs.recommendedTypeChecked satisfies Linter.Config[];
const Y = importXPlugin.flatConfigs.recommended satisfies Linter.Config;
```
will show that importXPlugin does not satisfy Linter.Config, which you already know as those lines have been changed.
I think TypeScript error is bad (bug) and these casts can be removed. You can use `satisfies` in place of `as` on these lines. Then the comment for casting moves down to importXPlugin that is a real problem.

common/build/eslint-config-fluid/library/configs/base.mts
jason-ha: Should be able to just remove this cast similar to removing `tseslint.configs.*` casts.

common/build/eslint-config-fluid/scripts/print-configs.ts
```diff
@@ -26,11 +26,9 @@ import { recommended, strict, minimalDeprecated } from "../flat.mjs";
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
-type FlatConfigArray = Linter.Config[];
-
 interface ConfigToPrint {
 	name: string;
-	config: FlatConfigArray;
+	config: readonly Linter.Config[];
```
jason-ha: Alternatively import FlatConfigArray from base.mts definition? That definition also makes Linter.Config readonly (= `readonly Readonly<Linter.Config>[]`)

jason-ha:
While here print-configs.ts's `configsToPrint` is better set as
```ts
const configsToPrint = [
 ...
] as const satisfies readonly ConfigToPrint[];
```
since it is a constant.

common/build/eslint-config-fluid/library/configs/base.mts
tylerbutler: Thanks for the reminder of `satisfies`. I keep forgetting about it.

jason-ha (approved):

PR #26426, opened 2026-02-12
Refactor ClientVersionDispatchingCodecBuilder to support unstable and unused formats.

CraigMacomber:
## Description

This generalizes ClientVersionDispatchingCodecBuilder  so it can be used in more cases, including with under development and discontinued formats.

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/codec/versioned/codec.ts
+++ b/packages/dds/tree/src/codec/versioned/codec.ts
@@ -3,11 +3,11 @@
  * Licensed under the MIT License.
  */
 
-import { assert } from "@fluidframework/core-utils/internal";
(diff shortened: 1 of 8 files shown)


packages/dds/tree/src/codec/versioned/codec.ts
```diff
@@ -179,6 +179,14 @@ export interface CodecVersionBase<
 	T = unknown,
 	TFormatVersion extends FormatVersion = FormatVersion,
 > {
+	/**
+	 * When `undefined` the codec never be selected as a write version unless via override.
```
yann-achard-MS:
```suggestion
	 * When `undefined` the codec will never be selected as a write version except via override.
```

yann-achard-MS (approved):

PR #25806, opened 2025-11-03
build(client): Move eslint-config-fluid to client release group

tylerbutler:
## Description

This change moves the `eslint-config-fluid` package into the client release group and makes it private.

### Changes:
- Moved `@fluidframework/eslint-config-fluid` to the client release group
- Updated the package version and mark it private
- Updated all references across the repository to use the workspace dependency

PR #26427, opened 2026-02-12
build(eslint-config-fluid): restore as independent workspace

tylerbutler:
## Summary

Reverses #25806 which temporarily moved `@fluidframework/eslint-config-fluid` into the client release group. The package is restored as an independently versioned workspace at **v9.0.0**, preparing it for its next published release with ESLint 9 flat config support.

### Key changes

- **eslint-config-fluid**: Version set to `9.0.0`, removed `private: true`, removed `typeValidation` block in package.json. All ESLint 9 work (flat config, typescript-eslint 8.54, etc.) is preserved.
- **pnpm-workspace.yaml**: Package removed from workspace members. Added an `eslint` catalog group for `eslint`, `eslint-config-prettier`, and `@fluid-internal/eslint-config-fluid`.
- **Root package.json**: Uses `link:` protocol with a pnpm override to resolve references across the workspace, plus a `postinstall` script to install transitive deps. ESLint version moved to `catalog:eslint`.
- **Client packages**: Dependencies changed from `workspace:~` to `catalog:eslint` for `eslint-config-fluid`, `eslint`, and `eslint-config-prettier`.
- **fluidBuild.config.cjs**: Removed temporary `npm-private-packages` policy exclusion; added `npm-package-types-field` exclusion.
- **syncpack.config.cjs**: Added ignore rules for `eslint-config-fluid` (link: protocol is unresolvable by syncpack).
- **build-eslint-config-fluid.yml**: Re-enabled `taskPack` for publishing.


examples/apps/blobs/package.json
```diff
@@ -65,8 +65,8 @@
 		"@fluid-tools/build-cli": "^0.63.0",
 		"@fluidframework/build-common": "^2.0.3",
 		"@fluidframework/build-tools": "^0.63.0",
-		"@fluidframework/eslint-config-fluid": "workspace:~",
-		"@fluidframework/test-tools": "^1.0.195075",
+		"@fluidframework/eslint-config-fluid": "~9.0.0",
+		"@fluidframework/test-tools": "^2.0.0",
```
alexvy86, line 70: I'd say let's split the update of test-tools to its own PR.

common/build/eslint-config-fluid/package.json
```diff
@@ -15,7 +14,6 @@
 	"scripts": {
 		"build": "npm run print-configs && npm run prettier",
 		"build:readme:disabled": "markdown-magic --files \"**/*.md\"",
-		"clean": "rimraf --glob dist \"**/*.build.log\"",
```
alexvy86, line 18: This shouldn't be removed, should it?

fluidBuild.config.cjs
```diff
@@ -416,6 +417,7 @@ module.exports = {
 				"^build-tools/",
 				"^common/lib/common-utils/package.json",
 			],
+			"npm-package-types-field": ["common/build/eslint-config-fluid/package.json"],
```
alexvy86, line 434:
```suggestion
			// Packages that don't need type tests
			"npm-package-types-field": ["common/build/eslint-config-fluid/package.json"],
```

package.json
```diff
@@ -365,6 +367,7 @@
 		],
 		"overrides": {
 			"@biomejs/biome": "~2.3.11",
+			"@fluidframework/eslint-config-fluid": "link:common/build/eslint-config-fluid",
```
alexvy86, line 374: Does this work correctly when other packages are at different depths? I.e. this relative path works fine for the root package.json, but is it interpreted correctly for things inside `packages/` and at different levels?

Also, is this temporary until we publish 9.0.0 of our eslint config and then we'll go back to pulling it from npm, or is the intention to continue referencing the local version?

examples/apps/blobs/package.json
tylerbutler: Already done: https://github.com/microsoft/FluidFramework/pull/26555

common/build/eslint-config-fluid/package.json
tylerbutler: We don't need a clean task AFAIK in this package - it doesn't build anything to dist. I'll double-check.

fluidBuild.config.cjs
tylerbutler: It's more that the package doesn't export types.

package.json
tylerbutler: > Does this work correctly when other packages are at different depths?

It seems to.

> Also, is this temporary until we publish 9.0.0 of our eslint config and then we'll go back to pulling it from npm, or is the intention to continue referencing the local version?

Planning to go back to npm. That said, if we like the ability to quickly make changes to the config and take effect immediately, then this is my recommended approach. But the plan is going back to publishing.

pnpm-lock.yaml
alexvy86, line 92: This is still not reflecting the move back to an external package.

tylerbutler: It won't yet - the dep is overridden to use a link dep in the root overrides. Once the independent package is published, then we can use an external dep. This PR just gets the release groups split up so I can release the config.

common/build/eslint-config-fluid/package.json
tylerbutler: Not needed.

pnpm-lock.yaml
alexvy86: Oh... and have we tried putting the link dep in the catalog instead of in the overrides?

package.json
```diff
@@ -80,6 +80,7 @@
 		"generate:packageList:public": "flub list client --no-private --feed public --outFile feeds/public.txt",
 		"generate:packagesMd": "flub check layers --info layerInfo.json --md .",
 		"preinstall": "node scripts/only-pnpm.cjs",
+		"postinstall": "pnpm -C common/build/eslint-config-fluid install --frozen-lockfile",
 		"install:agency": "curl -sSfL https://aka.ms/InstallTool.sh | sh -s agency",
```
jason-ha, line 83: Could we make this a preinstall?
It feels like having everything set for it before "integrating" within client is safer.

package.json
```diff
@@ -353,6 +354,8 @@
 	},
 	"pnpm": {
 		"comments": [
+			"eslint-config-fluid is an independent package in a separate workspace; the override resolves it to the local copy for all client packages.",
+			"biome is overridden to make review of the upgrade easier. This can be removed once merged.",
```
jason-ha: nit: swap order to match override order

package.json
```diff
@@ -80,6 +80,7 @@
 		"generate:packageList:public": "flub list client --no-private --feed public --outFile feeds/public.txt",
 		"generate:packagesMd": "flub check layers --info layerInfo.json --md .",
 		"preinstall": "node scripts/only-pnpm.cjs",
+		"postinstall": "pnpm -C common/build/eslint-config-fluid install --frozen-lockfile",
```
tylerbutler:
```suggestion
		"preinstall": "node scripts/only-pnpm.cjs && pnpm -C common/build/eslint-config-fluid install --frozen-lockfile",
```

package.json
tylerbutler: Done.

common/build/eslint-config-fluid/README.md
```diff
@@ -103,7 +103,6 @@ a diff to review as part of a PR -- just like we do with API reports for code ch
 |--------|-------------|
 | `build` | `npm run print-config` |
```
jason-ha, line 104: pre-existing: build does `npm run print-configs && npm run prettier`

packages/framework/tree-agent-langchain/package.json
```diff
@@ -135,8 +135,8 @@
 		"concurrently": "^9.2.1",
 		"copyfiles": "^2.4.1",
 		"cross-env": "^10.1.0",
-		"eslint": "~9.39.1",
-		"eslint-config-prettier": "~10.1.8",
+		"eslint": "catalog:eslint",
+		"eslint-config-prettier": "catalog:eslint",
```
jason-ha, line 139: ~aside: I've found that tree-agent- packages seemed to be created with rough settings that aren't correct. Does this need `eslint-config-prettier`?
Also, only some packages have `eslint-config-prettier` - what is the need? sometimes the dep in eslint-config-fluid doesn't suffice? (I don't see reference in client workspace outside of package.json deps.)

tools/pipelines/build-client.yml
jason-ha, line 1: These removals don't seem correct. When those files change they impact client build. Only once we have a proper feed dependency can these be safely removed.

tools/pipelines/build-eslint-config-fluid.yml
```diff
@@ -102,5 +102,3 @@ extends:
     taskLint: false
     taskTest:
       - test
-    # Disabled while eslint-config-fluid is part of the client release group
-    taskPack: false
```
jason-ha, line 106: I am guessing this is okay - but just adding an "are we sure?" comment per the triggers change that I don't think is okay.

package.json
```diff
@@ -367,6 +369,8 @@
 			"diff: overridden to patched versions to resolve a known ReDoS vulnerability. diff@3.x and diff@7.x have no fix in their major range so they are bumped to the nearest patched major."
 		],
 		"overrides": {
+			"@biomejs/biome": "~2.4.5",
+			"@fluidframework/eslint-config-fluid": "link:common/build/eslint-config-fluid",
```
jason-ha, line 374: Does this need to be an override? Why can't we just have this in the catalog settings?
Bonus: you'll be able to make better comments there.

syncpack.config.cjs
```diff
@@ -225,8 +243,6 @@ module.exports = {
 				"@fluidframework/build-common",
 				"@fluidframework/build-tools",
 				"@fluidframework/common-utils",
-				// Temporarily disabled while eslint-config-fluid is part of the client release group
-				// "@fluidframework/eslint-config-fluid",
```
jason-ha, line 226: Not sure what is going on here.
Before: comment and commented out setting
Now: comment is gone and the commented out setting, too. The setting should come back when we are using the feed, right? So, leave comments for now?

pnpm-lock.yaml
jason-ha, line 1: I feel like this might mess with Component Governance - there will be installs that are not in the lockfile.
Assuming this is just a transition step, then okay!

jason-ha (commented):
I like it but I think some changes are not yet ready to be made.

pnpm-workspace.yaml
```diff
@@ -37,3 +36,8 @@ catalogs:
     "@fluid-tools/version-tools": ^0.64.0
     "@fluidframework/build-tools": ^0.64.0
     "@fluidframework/bundle-size-tools": ^0.64.0
+  # eslint and related packages
+  eslint:
```
ChumpChief, line 41: I haven't looked much at catalogs yet, but what's the benefit of dividing the packages amongst multiple catalogs like buildTools vs. eslint?  I would have expected a single monolithic catalog, but I guess that's not the design?  Anyway just wondering how you decided on "eslint" as a catalog scoping.

common/build/eslint-config-fluid/package.json
ChumpChief: It does still write build.log files for prettier and print-configs -- might still be nice to retain that portion

packages/framework/tree-agent-langchain/package.json
tylerbutler: As I recall when there was some staged rollout of prettier and these were supposed to be cleaned up when prettier. They could be cleaned up now, but out of scope for this PR, which I am trying to keep as close to a pure revert as possible.

package.json
tylerbutler: Because catalogs weren't in when I made this PR over a month ago. I'm just trying to get it merged ASAP.

syncpack.config.cjs
tylerbutler: This is no longer needed because @fluidframework/eslint-config-fluid is not part of the release group any more, so the synpack settings here no longer apply to it. This is an example of "putting it back the way it was."

pnpm-workspace.yaml
tylerbutler: There is no "design" here because there is no owner nor oracle to consult when it comes to making these decisions. I guess we could have one large catalog, but I don't want to litigate it in this PR.

package.json
tylerbutler: ERR_PNPM_CATALOG_ENTRY_INVALID_SPEC  The entry for '@fluidframework/eslint-config-fluid' in catalog 'eslint' declares a dependency using the 'link' protocol. This is not yet supported, but may be in a future version of pnpm.

package.json
tylerbutler: ERR_PNPM_CATALOG_ENTRY_INVALID_SPEC  The entry for '@fluidframework/eslint-config-fluid' in catalog 'eslint' declares a dependency using the 'link' protocol. This is not yet supported, but may be in a future version of pnpm.

common/build/eslint-config-fluid/package.json
tylerbutler: Reverted in latest.

tools/pipelines/build-client.yml
tylerbutler: reverted in latest with an updated comment.

tools/pipelines/build-eslint-config-fluid.yml
tylerbutler: This is _probably_ safe to revert now. YOLO!

tylerbutler: Ugh, vscode's review UI is confusing. Yes, this is intended to be removed.

package.json
```diff
@@ -171,7 +171,7 @@
 		"@fluid-tools/markdown-magic": "workspace:~",
 		"@fluidframework/build-common": "^2.0.3",
 		"@fluidframework/build-tools": "catalog:buildTools",
-		"@fluidframework/eslint-config-fluid": "workspace:^",
+		"@fluidframework/eslint-config-fluid": "link:common/build/eslint-config-fluid",
```
jason-ha: not catalog:eslint?

pnpm-workspace.yaml
jason-ha: I appreciate the grouping. It suggests things that are related and should be considered together.

syncpack.config.cjs
jason-ha: The way it was prior to https://github.com/microsoft/FluidFramework/commit/4854dc1a823dd0fe77a49513a31ffa04ee4b77b4#diff-e7c97b4272fdec5c94fdcd2a6c9e0072f985d80d2ee75996029f81eab9e3977c was to have `"@fluidframework/eslint-config-fluid",` uncommented.
Convince me no entry is that OR offer a different reason it shouldn't ever be here in the future.

package.json
jason-ha: Comments would be nice - not a blocker

jason-ha (commented):
two notes holding up me approving...

jason-ha:
Something in build wants to change the README. Not sure why client build found that versus eslint-config-fluid build.

pnpm-lock.yaml
tylerbutler: I don't follow. What is not covered by the root and eslint-config-fluid lockfiles?

tylerbutler: Catalogs don't support link: protocol.

pnpm-workspace.yaml
alexvy86: My gut feeling aligns with Jason's. When I introduced the buildTools catalog I was also unsure of how we'd want to split deps into separate catalogs but I did expect at least a few vs one monolithic one. That said, if anyone feels strongly against or finds good evidence to sway us one way or the other, my opinions here are held pretty loosely.

pnpm-lock.yaml
jason-ha: It is probably fine. The root lockfile no longer represents the entirety of dependencies that the client workspace has. I would hope that all installs for a job are checked - but it feels like there could be missing association.

pnpm-workspace.yaml
ChumpChief: This was purely me asking for education purposes, not a pointed question.

package.json
jason-ha: I'm still wondering about this one. Can't be catalog reference?

alexvy86: Per https://github.com/microsoft/FluidFramework/pull/26427#discussion_r2968386458, apparently not (as long as we want the `link:` reference).

jason-ha: I get that for link can't be in the catalog. But lots of other package.json specs are `catalog:eslint`. Why is this one special?

alexvy86: For extra context, I chatted with Jason because I didn't quite understand his question. The point is: other packages have their `"@fluidframework/eslint-config-fluid"` deps pointing to the catalog, because the catalog points to `^9.0.0` (not `link:...`) and it's the pnpm override that then replaces that with the `link`. Which seems like it should also work here.

pnpm-lock.yaml
alexvy86: If I understood this correctly, the concern about the deps in eslint-config-fluid not being represented in the root lockfile is valid but the infra/tooling should still protect us. The install of eslint-config-fluid's deps should (right?) follow the lockfile in that folder, and all lockfiles in the repo are scanned during the Component Governance check. (For better and for worse; I'm not a fan of how we need to run _all_ pipelines after updating the root lockfile, or one of the server lockfiles, so all pipelines declare a given alert resolved).

jason-ha (approved):

PR #26432, opened 2026-02-12
Add tree context API

noencke:
This adds a new tree static function `Tree.context(node)` which yields a new type: a `TreeContext`. A `TreeContext` is a further generalization of branch and view. So the hierarchy (from most general to most specific) is now:

`TreeContext` - A place for functions that don't require the caller to have hydrated/attached content - i.e., the ST DDS / Fluid are not necessarily reachable.
`TreeBranch` - A place for functions that _do_ require the caller to have hydrated content, and can reach the ST/Fluid.
`TreeView` - A _schematized_ tree branch

Accordingly, you use `isBranch()` to downcast a `TreeContext` to a `TreeBranch`, and then `hasRootSchema` to downcast a `TreeBranch` to a `TreeView`.

`TreeContext` is now the home of `runTransaction`, since `runTransaction` can operate on both hydrated and unhydrated nodes - in the latter case, it simply runs the transaction delegate. This lets you do `Tree.context(node).runTransaction(...)` without having to worry about whether `node` is hydrated or not. Note that this "upcasted" version of `runTransaction` does not have any options (like labels, or constraints) because those don't make sense until you downcast to a branch.

`Tree.branch` is now deprecated in favor of `Tree.context()` + `context.isBranch()`. This is slightly more verbose for some situations but it is trivial for a consumer to write a helper to replicate the previous behavior of `Tree.branch` if they want the one-liner. By removing `Tree.branch`, we avoid an awkward naming dilemma: most other `Tree.*` functions are single nouns which retrieve that noun, and the same is true for `branch` - but "branch" is also a _verb_, which makes it sound like the API is _forking_.


packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,86 @@ export interface TreeBranch extends IDisposable {
...
+	 * @param transaction - The function to run as the body of the transaction.
+	 * It may return a {@link WithValue | value }, which will be returned by the `runTransaction` call.
+	 * @returns A result object of {@link TransactionResultExt | TransactionResultExt} type. It includes the following:
+	 *
+	 * - A "success" flag indicating whether the transaction was successful or not.
+	 * - The success or failure value as returned by the transaction function.
+	 *
+	 * @remarks
+	 * If `runTransaction` is invoked on the context of a {@link TreeStatus.InDocument | node in the document }, the transaction will be applied to the {@link TreeBranchAlpha | branch associated with that node}.
+	 * Use {@link TreeContextAlpha.isBranch | isBranch() } to check whether the returned context is associated with a branch and gain {@link TreeBranchAlpha | access to more transaction capabilities} if so.
```
yann-achard-MS: >returned [](http://example.com/codeflow?start=77&length=8)

This bit about checking the returned context seems a bit confusing. You're presumably referring to `TreeAlpha.context`, but there's no mention of it here. Maybe make "returned" a link to `TreeAlpha.context`. #Resolved

yann-achard-MS: This bit about checking the returned context seems a bit confusing. You're presumably referring to `TreeAlpha.context`, but there's no mention of it here. Maybe make "returned" a link to `TreeAlpha.context`. #Resolved

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,86 @@ export interface TreeBranch extends IDisposable {
...
+/**
+ * Context for a tree node which provides additional context-aware APIs.
+ * @alpha
+ */
+export interface TreeContextAlpha {
+	/**
+	 * Run a synchronous transaction which applies one or more edits to the tree as a single atomic unit.
+	 * @param transaction - The function to run as the body of the transaction.
+	 * It may return a {@link WithValue | value }, which will be returned by the `runTransaction` call.
+	 * @returns A result object of {@link TransactionResultExt | TransactionResultExt} type. It includes the following:
```
yann-achard-MS: Can this just be simplfied to
```suggestion
	 * @returns A result object of {@link TransactionResultExt} type. It includes the following:
``` #Resolved

File: packages/dds/tree/src/simple-tree/api/tree.ts [hunk 3]
```diff
@@ -175,6 +179,86 @@ export interface TreeBranch extends IDisposable {
...
+	 *
+	 * - A "success" flag indicating whether the transaction was successful or not.
+	 * - The success or failure value as returned by the transaction function.
+	 *
+	 * @remarks
+	 * If `runTransaction` is invoked on the context of a {@link TreeStatus.InDocument | node in the document }, the transaction will be applied to the {@link TreeBranchAlpha | branch associated with that node}.
```
yann-achard-MS, line 201: I wonder if it would make sense to link to `TreeBranchAlpha.runTransaction` since it has more details on the transaction semantics. This would also avoid scenarios where a user doesn't realize that we offer a more full-featured version of `runTransaction`. #Resolved

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: I wanted to but I can't figure out how to properly link to `TreeAlpha` with TSDoc. You have to do the special `@link (TreeAlpha:interface).context` syntax (because we have a `TreeAlpha` type as well as a `TreeAlpha` object), but then VSCode linking simply doesn't work.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: I don't see why not. This is just copied from our existing `runTransaction` docs (on TreeBranchAlpha) so that they are consistent. It seems all the type links do this funny duplication.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: I poked around and it seems like our website docs respect the extended tsdoc syntax, even if VSCode does not. So I guess that's a good enough reason to still do it. At least our customers will have nice docs, even if we have extra ugly and hard to read docs with broken links.

packages/dds/tree/src/shared-tree/treeAlpha.ts
```diff
@@ -1079,3 +1095,41 @@ function borrowFieldCursorFromTreeNodeOrValue(
...
+	public async runTransactionAsync(
+		t: () => Promise<WithValue<unknown> | void>,
+	): Promise<TransactionResultExt<unknown, unknown> | TransactionResult> {
+		return UnhydratedTreeContext.wrapTransactionResult(await t());
+	}
+
+	private static wrapTransactionResult<TValue>(
+		value: WithValue<TValue> | void,
+	): TransactionResultExt<TValue, TValue> | TransactionResult {
+		if (value?.value !== undefined) {
```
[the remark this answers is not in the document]
noencke: Huh? Am I crazy, or are you?

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Regardless, I think the best thing to do here is just say "this context" since this is a method on the context. You are right, it was weird/incorrect as it was previously phrased.

noencke: Duplicate comment, [see other conversation](https://github.com/microsoft/FluidFramework/pull/26432#discussion_r2800991456).

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: I cleaned up these as well as some of the existing ones.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: I linked it, using the sketchy tsdoc:1 selector. Hopefully it works on our website.

packages/dds/tree/src/simple-tree/api/tree.ts
yann-achard-MS: I didn't realize this was just a copy. Thanks for the cleanup.

packages/dds/tree/src/simple-tree/api/tree.ts
yann-achard-MS: Looks like the CodeFlow extension lied to me about failing to add this comment. :)

yann-achard-MS (approved):

yann-achard-MS (approved):

.changeset/green-sides-watch.md
```diff
@@ -0,0 +1,49 @@
+---
+"@fluidframework/tree": minor
```
CraigMacomber, line 2: if it impacts the tree API, it impacts the fluidframework project which reexports it as well.

.changeset/green-sides-watch.md
```diff
@@ -0,0 +1,49 @@
...
+"__section": tree
+---
+Added `TreeAlpha.context(node)` to provide context-aware APIs for any SharedTree node, plus a new `TreeContextAlpha` surface for transactions and branch checks.
```
CraigMacomber: We don't code format or put a period at the end of changeset titles. Also this is a bit long for a title.

https://github.com/microsoft/FluidFramework/wiki/Changesets#formatting


```suggestion
Added TreeAlpha.context(node) to provide context-aware APIs for SharedTree nodes, and TreeContextAlpha for transactions and branch checks
```

.changeset/green-sides-watch.md
```diff
@@ -0,0 +1,49 @@
...
+
+This release introduces a node-scoped context that works for nodes inserted into the tree as well as new nodes that are not yet inserted.
```
CraigMacomber:
```suggestion
This release introduces a node-scoped context that works for hydrated and [unhydrated](https://fluidframework.com/docs/api/fluid-framework/unhydrated-typealias) [TreeNodes](https://fluidframework.com/docs/api/fluid-framework/treenode-class).
```

.changeset/green-sides-watch.md
```diff
@@ -0,0 +1,49 @@
...
+The new `TreeContextAlpha` interface exposes `runTransaction` / `runTransactionAsync` and an `isBranch()` type guard.
```
CraigMacomber:
```suggestion
The new `TreeContextAlpha` interface exposes `runTransaction` / `runTransactionAsync` methods and an `isBranch()` type guard.
```

.changeset/green-sides-watch.md
```diff
@@ -0,0 +1,49 @@
...
+`TreeBranchAlpha` now extends `TreeContextAlpha`, so you can keep using branch APIs when available.
+
+### Migration
```
CraigMacomber:
```suggestion
#### Migration
```

.changeset/green-sides-watch.md
noencke: "unhydrated" causes the PR docs watchdog to yell at me - should I add it to the dictionary? Where is that again?

.changeset/green-sides-watch.md
```diff
@@ -0,0 +1,49 @@
...
+if (context.isBranch()) {
+	// Same branch APIs as before
+	context.fork();
+}
+```
+
+`TreeAlpha.branch(node)` is now deprecated.
+Prefer the context API above.
+
+### New transaction entry point
```
CraigMacomber:
```suggestion
#### New transaction entry point
```

.changeset/green-sides-watch.md
```diff
@@ -0,0 +1,49 @@
...
+You can now run transactions from a node context, regardless of whether the node is hydrated:
+
+```ts
+const context = TreeAlpha.context(node);
+
+// No return value
```
CraigMacomber: I find this comment confusing. We are assigning to "result", which is presumably the return value of runTransaction. Are you instead talking about the callback passed to runTransaction?

Also as there has been no mention of return values for this entire changeset, calling out that something doesn't have one here seems a little lacking in context.

CraigMacomber: Now I see the code below this. Maybe it would make sense to remove the "result =", or to give a type to result (which is undefined?). Typing the result in both these cases would be helpful: I can't tell if they are wrappers or not.

noencke: You're right, since this isn't anything new WRT to the existing runTransaction APIs, I'll just show a couple of examples and keep it simple

CraigMacomber: Also adjusting this comment to something like "A transaction without a return value" (if result is undefined/void), or "A transaction without a custom return value" otherwise might help.

.changeset/green-sides-watch.md
```diff
@@ -0,0 +1,49 @@
...
+const result = context.runTransaction(() => {
+	node.count += 1;
+});
+
+// Return a value by wrapping it in `{ value }`
+const resultWithValue = context.runTransaction(() => ({ value: node.count }));
+```
+
+For asynchronous work:
```
CraigMacomber:
```suggestion
For asynchronous transactions:
```

.changeset/green-sides-watch.md
```diff
@@ -0,0 +1,49 @@
...
+
+```ts
+const result = await context.runTransactionAsync(async () => {
+	await doWork();
```
CraigMacomber: I think it might make more sense if you passed node into doWork:

```suggestion
	await doWork(node);
```

noencke: Good idea.

packages/dds/tree/src/simple-tree/api/transactionTypes.ts
```diff
@@ -90,22 +95,18 @@ export type VoidTransactionCallbackStatus = Omit<
  * The result of the {@link RunTransaction | RunTransaction} API when it was successful.
```
CraigMacomber, line 95: What do "successful" and "failed" mean for these?

Does failed mean the user was unable to make the change to the tree that they desired? Does it mean the result was non-transactional? Does it mean an exception left the tree in a broken state? There are a lot of things it could mean.

Generally I think we should not use the word "failed" for well defined in memory operations unless the outcome fails to align with the name of the API (Like getFoo can fail if you don't get a foo).

Here I don't think failed means we didn't run the transaction, so I think we should use a more specific term.

For similar reasons I don't think a success Boolean is a good API.

If someone tries to roll back their transaction, and success is "false" does that mean the rollback failed? We could have that APi shape for unhydrated transactions. Or do we return failed when an async transaction fails to merge and thus is reverted when it was not requested to be? This naming is confusing.

If this is suppose to indicate "applied" verses "reverted" then I think having an enum or string union indicated such would make more sense.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -11,6 +11,9 @@ import type {
 	RevertibleAlphaFactory,
 	RevertibleFactory,
 } from "../../core/index.js";
+// This is referenced by doc comments.
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { TreeStatus } from "../../feature-libraries/index.js";
```
CraigMacomber, line 16: As far as I know, we don't have a policy on if we should or should not includes imports needed by doc comments. They are not required for our build and publishing process, they just impact intelisense and not anything else. I'd love if someone actually landed on a decision about this and documented it, and ensure it aligned with out tooling. Usually we do not include such imports, but I don't have a strong opinion other than we should decide if we want to have these as a project level and not case by case.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
 	dispose(error?: Error): void;
 }
 
+/**
+ * Context for a tree node which provides additional context-aware APIs.
```
CraigMacomber: Would be good to note what a "context" is.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+export interface TreeContextAlpha {
+	/**
+	 * Run a synchronous transaction which applies one or more edits to the tree as a single atomic unit.
+	 * @param transaction - The function to run as the body of the transaction.
+	 * It may return a {@link WithValue | value }, which will be returned by the `runTransaction` call.
+	 * @returns A {@link TransactionResultExt | result object}.
+	 * It includes the following:
+	 *
+	 * - A "success" flag indicating whether the transaction was successful or not.
+	 * - The success or failure value as returned by the transaction function.
```
CraigMacomber: Does `the transaction function` refer to `runTransaction` or `transaction`. If its suppose to be the later, then 
```suggestion
	 * - The success or failure value as returned by the `transaction` function.
```

or 


```suggestion
	 * - The success or failure value as returned by `transaction`.
```

CraigMacomber: Why are success and failure defined here and not on the type which defines the success field?

Also, this does not say what successful means since it defines successful to mean if it was "successful". THis provides slightly more detail than you did on the interfaces, but its still not very helpful to define a word interms of itself especially when a user might be trying to revert the change, or trying to apply it, or have many ways they can fail.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 *
+	 * @remarks
+	 * If `runTransaction` is invoked on the context of a {@link TreeStatus.InDocument | node in the document }, the transaction will be applied to the {@link TreeBranchAlpha | branch associated with that node}.
```
CraigMacomber, line 201: This also applies to removed nodes.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 * Use {@link TreeContextAlpha.isBranch | isBranch() } to check whether this context is associated with a branch and gain {@link TreeBranchAlpha.(runTransaction:1) | access to more transaction capabilities} if so.
+	 *
+	 * If `runTransaction` is invoked on the context of an {@link TreeStatus.New | unhydrated } node, it is equivalent to running the `transaction` delegate directly (i.e. `runTransaction` does nothing additional).
```
CraigMacomber: we have a doc link for unhydrated you can link:  https://fluidframework.com/docs/api/tree/unhydrated-typealias

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 * The transaction will always succeed.
```
CraigMacomber: This, as two sentences, doesn't seem right. Did you mean:


```suggestion
	 * If `runTransaction` is invoked on the context of an {@link TreeStatus.New | unhydrated } node, it is equivalent to running the `transaction` delegate directly (i.e. `runTransaction` does nothing additional):
	 * the transaction will always succeed.
```

Also, isn't that not quite true, as it does something: it recreates the return value object adding a new member.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 */
+	runTransaction<TValue>(
+		transaction: () => WithValue<TValue>,
+	): TransactionResultExt<TValue, TValue>;
+	/**
```
CraigMacomber, line 211: Generally I like having a blank line between members, especially if they have lots of lines within them.


```suggestion

	/**
```

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 * Run a synchronous transaction which applies one or more edits to the tree as a single atomic unit.
+	 * @param transaction - The function to run as the body of the transaction.
+	 * @remarks
+	 * If `runTransaction` is invoked on the context of a {@link TreeStatus.InDocument | node in the document }, the transaction will be applied to the {@link TreeBranchAlpha | branch associated with that node}.
+	 * Use {@link TreeContextAlpha.isBranch | isBranch() } to check whether this context is associated with a branch and gain {@link TreeBranchAlpha.(runTransaction:2) | access to more transaction capabilities} if so.
+	 *
+	 * If `runTransaction` is invoked on the context of an {@link TreeStatus.New | unhydrated } node, it is equivalent to running the `transaction` delegate directly (i.e. `runTransaction` does nothing additional).
+	 * The transaction will always succeed.
+	 */
+	runTransaction(transaction: () => void): TransactionResult;
```
CraigMacomber, line 221: Doesn't typescript pick the first overload which fits, which means this overload should probably be first as its the more specific one, or not exist because its not necessary and just a subset?

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	/**
```
CraigMacomber, line 223: This has been going on way too long without any line breaks. Makes it hard to see the boundary between these.


```suggestion

	/**
```

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 * Run an asynchronous transaction which applies one or more edits to the tree as a single atomic unit.
```
CraigMacomber, line 224: Isn't it legal for this to do zero edits?

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 * @param transaction - The function to run as the body of the transaction.
```
CraigMacomber, line 225: Defining transaction using the world transaction isn't great, and I don't think "body" is a very informative word here. Sure I know what this is doing, so I know what it implies, but I didn't already know what was intended, I think this might be confusing. For example the `<body>` means all of the content, but can be styled by css.

Something like:


```suggestion
	 * @param transaction - A callback run during the transaction to perform the user controlled operations.
```

Might be more informative.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 * It may return a {@link WithValue | value }, which will be returned by the `runTransactionAsync` call.
+	 * @returns A promise that resolves to a {@link TransactionResultExt | result object}.
+	 * It includes the following:
+	 *
+	 * - A "success" flag indicating whether the transaction was successful or not.
+	 * - The success or failure value as returned by the transaction function.
```
CraigMacomber: Having to duplicate this is why it should be part of the docs on the type you are returning where this field is defined, not repeated in every single function here.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 *
+	 * @remarks
+	 * If `runTransactionAsync` is invoked on the context of a {@link TreeStatus.InDocument | node in the document }, the transaction will be applied to the {@link TreeBranchAlpha | branch associated with that node}.
```
CraigMacomber, line 234: Also applies to removed nodes.

Might be better to just link the sync version, and call out the differences.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 * Use {@link TreeContextAlpha.isBranch | isBranch() } to check whether this context is associated with a branch and gain {@link TreeBranchAlpha.(runTransactionAsync:1) | access to more transaction capabilities} if so.
+	 *
+	 * If `runTransactionAsync` is invoked on the context of an {@link TreeStatus.New | unhydrated } node, it is equivalent to running the `transaction` delegate directly (i.e. `runTransactionAsync` does nothing additional).
+	 * The transaction will always succeed.
+	 */
+	runTransactionAsync<TValue>(
```
CraigMacomber, line 240: Nowhere in the docs for this did it cover the implications of this being async. That seems like the only real important thing to document here.

How does this impact the main branch if used on it?

What are the risks? We should note what can happen thats bad if you fail to finish a transaction. We should not running a transaction does not get you a separate view of the content, so it impacts code not inside the transaction if the transaction is still open, and recommend that async transactions crate a branch specific for the transaction to avoid unrelated changes getting grouped in with it. We should not that transactions must be nested and can't start within one async transaction and end after it.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+	 * Run an asynchronous transaction which applies one or more edits to the tree as a single atomic unit.
+	 * @param transaction - The function to run as the body of the transaction.
+	 * @remarks
+	 * If `runTransactionAsync` is invoked on the context of a {@link TreeStatus.InDocument | node in the document }, the transaction will be applied to the {@link TreeBranchAlpha | branch associated with that node}.
+	 * Use {@link TreeContextAlpha.isBranch | isBranch() } to check whether this context is associated with a branch and gain {@link TreeBranchAlpha.(runTransactionAsync:2) | access to more transaction capabilities} if so.
+	 *
+	 * If `runTransactionAsync` is invoked on the context of an {@link TreeStatus.New | unhydrated } node, it is equivalent to running the `transaction` delegate directly (i.e. `runTransactionAsync` does nothing additional).
+	 * The transaction will always succeed.
+	 */
+	runTransactionAsync(transaction: () => Promise<void>): Promise<TransactionResult>;
```
CraigMacomber, line 254: I got to the end of these docs, and I don't think any of it mentioned the ability to roll back transactions, what happens if incompatible changes merge (like something that violates a constraint in the transaction or a schema change) before an async transaction finishes, or before any transaction is sequenced.

Also don't we have some new tech for putting labels on transactions? I don't see any mention of that here either.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -175,6 +179,88 @@ export interface TreeBranch extends IDisposable {
...
+
+	/**
+	 * True if this context is associated with a {@link TreeBranchAlpha | branch} and false if it is associated with an {@link TreeStatus.New | unhydrated } node.
+	 * @remarks If this returns true, the context can be safely inferred or cast to {@link TreeBranchAlpha} to access additional branch-specific APIs.
+	 * @example
+	 * ```typescript
+	 * const context = tree.context(someNode);
+	 * if (context.isBranch()) {
+	 *   context.fork(); // `fork` is a method on TreeBranchAlpha, so this is only accessible if `context` is a branch context.
```
CraigMacomber: I think this is a bad example. It leaks a branch.

Maybe asserting the context is a branch, then forking and doing an async transaction on it would be a good example. Or doing a transaction with rollback.

packages/dds/tree/src/simple-tree/api/transactionTypes.ts
noencke: These are good questions and I think we should answer them in a different PR. This PR is attempting to keep the "new" transaction APIs as similar to the existing ones as makes sense (they are, in fact, the same API just with different typing). 

But as a next step, and definitely before they become beta, we should address these ambiguities and terminology questions across all of them at once.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: This is a copy paste - I'm just making it consistent with our other transaction docs for now, but I agree it could be better. We'll update it all at once - including in the other transaction APIs - in a future PR before going to beta.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: It's necessary because the return value is different, even though the input parameter is a subset. We want it second because it's the "easier" one to match, so it should be matched last. If you _do_ pass a function that returns a value, we want to match with the other function, not this one, even though it _could_ match as you pointed out.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Yes, good callout. Another thing we should fix when we refresh all the transaction docs together in a followup PR.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: I agree. I'll add this to a "general transaction API improvement fixes" list for the next PR.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Well, you can't rollback with these APIs, or put constraints, or labels, and that's on purpose.

.changeset/green-sides-watch.md
noencke: "accept.txt" - added.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Yea, I'm not sure either. We already have precedent in this file for this, so I'm inclined to include them for now. Easy to remove them all in a batch later if we decide.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: I reworded it since it was kind of circular to call a context a context.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Good point, updated.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Nice, thanks.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Good call.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Added to TODO list

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Added to TODO list

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: I improved the example but still kept it short.

CraigMacomber (approved):
Docs seem good, especially given planned follow-ups to dedupe

PR #26434, opened 2026-02-12
Add ExtensibleUnionNode.isValid and more tests

CraigMacomber:
## Description

Add ExtensibleUnionNode.isValid and more tests.

Throw when accessing ExtensibleUnionNode.union if in an invalid state.

## Reviewer Guidance


packages/dds/tree/src/extensibleUnionNode.ts
```diff
@@ -64,8 +67,30 @@ export namespace ExtensibleUnionNode {
...
+		 * Returns true, unless this node is in a invalid state.
+		 * @remarks
+		 * Invalid in this context means that the internal implementation details of this node have had their invariants violated.
+		 * This can happen when:
+		 * - Using weakly typed construction APIs like {@link TreeBeta.importConcise} or {@link TreeBeta.importVerbose} to construct an invalid state directly.
+		 * Using such APIs, even when not creating invalid nodes, is not supported for this schema,
+		 * since doing so requires knowing the implementation details of this node which are subject to change.
+		 * - By editing a document using a different client using a different schema for this node.
+		 * - Violating the TypeScript types to directly manipulate the node internals.
+		 * - A bug in this node's implementation.
```
noencke: This line probably isn't necessary - you can say this for any of our APIs, and we don't deem it necessary to do that.

packages/dds/tree/src/extensibleUnionNode.ts
```diff
@@ -64,8 +67,30 @@ export namespace ExtensibleUnionNode {
...
 		 * The child wrapped by this node has one of the types allowed by the union,
 		 * or `undefined` if the type is one which was added to the union by a future version of this schema.
+		 *
+		 * @throws if {@link isValid} is false.
 		 */
 		readonly union: T | undefined;
+
+		/**
+		 * Returns true, unless this node is in a invalid state.
+		 * @remarks
```
noencke, line 77: This gives a detailed outline of how this API functions, but what is a user supposed to use it for? Should they be checking this every time they are about to access `.union`? Can they ignore it most of the time? If so, when can they not ignore it? Would an example help?

noencke (approved):

packages/dds/tree/src/extensibleUnionNode.ts
```diff
@@ -64,8 +67,30 @@ export namespace ExtensibleUnionNode {
...
+		 * Invalid in this context means that the internal implementation details of this node have had their invariants violated.
+		 * This can happen when:
+		 * - Using weakly typed construction APIs like {@link TreeBeta.importConcise} or {@link TreeBeta.importVerbose} to construct an invalid state directly.
+		 * Using such APIs, even when not creating invalid nodes, is not supported for this schema,
+		 * since doing so requires knowing the implementation details of this node which are subject to change.
```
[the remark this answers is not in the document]
CraigMacomber: I think our linting bans ding this.

packages/dds/tree/src/extensibleUnionNode.ts
CraigMacomber: Most of our APIs cannot leave the content of a persisted tree in well defined in schema state which will pass schema validation but could cause exceptions in future clients when those future clients are bug free.

Currently its just table and this code which can have such bugs. If the bug manifested as something other than a usage error, and/or only impacted the clients with the bug, I'd remove this, but since its persisted data which can give a usage error, it seems like we should note that.
This code isn't very likely to have such a bug, but more complex cases like table could pretty reasonably have such a but, and we would want someone to report it if it occurs. As we are establishing a bit of a pattern here other schema might follow, being complete seems like a good idea.

packages/dds/tree/src/extensibleUnionNode.ts
CraigMacomber: Added.

noencke (approved):

noencke (approved):

jatgarg (approved):

PR #24785, opened 2025-06-08
improvement(client-core-interfaces): revise JsonDeserialized recursion

jason-ha:
Use `OpaqueJsonDeserialized` at recursion points for inexactly deserializable types instead of limited unrolling. Callers will need to use cast helper to resolve, which will produce a repeatable pattern.

Supporting changes:
 1. Allow `OpaqueJson*` branded types (classes) when checking for recursion and hidden (non-public) properties.
 1. Separate `RecursionMarker` `AllowExactly` logic re-use from the `AllowExactly` use in `JsonDeserializedOpaqueConversion`
 1. Internally, ancestor types are now explicitly limited to `object` types.
 1. Remove unused `Control` option in `JsonSerializableImpl`'s use of `IfNonPublicProperties`.
 1. Indexed properties with unsupported types no longer gain `| undefined` by being explicitly optional. 
 1. Better handle branded index keys - TypeScript has an issue comparing indices that have been manipulated. Branded indices like `type FooBarString = string & { foo: "bar" }`, when filtered, may lose `FooBarString` and `string & { foo: "bar" }` will not match via `IfSameType` implementation.
    - Make sure to use original key type (when possible, looks like all cases), when generating keys for mapped types.
 1. Branded property keys are assumed to be indexed properties. (Complete test is not currently possible and it is impractical to use a branded literal property key.)
    - Introduce `IfIndexOrBrandedKey` and replace all `IfIndexKey` use.
    - Relocate `IfIndexKey` to common section and rename to `IfVariableStringOrNumber`.
 1. Tests:
    - Update case for class with private data in a recursive situation.
    - Add explicit verification that no result contains `any`.
    - Add branded index keys cases.
 1. Correct RawInboundExtensionMessage  to use OpaqueJsonDeserialized
 1. Correct Presence test case showing missing OpaqueJson wrapping

PR #24733, opened 2025-05-29
feat(client-core-interfaces): opaque JSON placeholder types

jason-ha:
Introduce `OpaqueJsonDeserialized`, and `OpaqueJsonSerializable` as "handles" for generic values.

Opaque Json types can be used when there is no direct need to work with a complex or unknown type, but still need to respect serializability. Common use is when a value that has passed through `JsonSerializable` or `JsonDeserialized` filter utilities needs to be stored, especially in a generic context. The Opaque types are recognized by `JsonSerializable` and `JsonDeserialized` allowing those stored captures to themselves be part of flow that checks for serializability or deserialization.

Test cases:
  JsonDeserialized
    positive compilation tests
      fully supported object types are preserved
        ✔ OpaqueJsonDeserialized<{number:number}>
      partially supported object types are modified
        OpaqueJsonSerialized becomes OpaqueJsonDeserialized counterpart
          ✔ OpaqueJsonSerializable<{number:number}> becomes OpaqueJsonDeserialized<{number:number}>
          ✔ OpaqueJsonSerializable<{number:number}>&OpaqueJsonDeserialized<{number:number}> becomes OpaqueJsonDeserialized<{number:number}>
        opaque Json types requiring extra allowed types have extras removed
          ✔ opaque serializable object with `bigint`
          ✔ opaque deserialized object with `bigint`
          ✔ opaque serializable and deserialized object with `bigint`
          ✔ opaque serializable object with number array expecting `bigint` support
          ✔ opaque deserialized object with number array expecting `bigint` support
          ✔ opaque serializable and deserialized object with number array expecting `bigint` support
    special cases
      using alternately allowed types
        are preserved
          ✔ opaque serializable object with `bigint`
          ✔ opaque deserialized object with `bigint`
          ✔ opaque serializable and deserialized object with `bigint`

  JsonSerializable
    positive compilation tests
      supported object types
        opaque Json types
          ✔ opaque serializable object
          ✔ opaque deserialized object
          ✔ opaque serializable and deserialized object
    negative compilation tests
      unsupported types cause compiler error
        opaque Json types requiring extra allowed types
          ✔ opaque serializable object with `bigint`
          ✔ opaque deserialized object with `bigint`
          ✔ opaque serializable and deserialized object with `bigint`
          ✔ opaque serializable object with number array expecting `bigint` support
          ✔ opaque deserialized object with number array expecting `bigint` support
          ✔ opaque serializable and deserialized object with number array expecting `bigint` support

[...]

PR #26435, opened 2026-02-12
improvement(client): API surface cleanup

jason-ha:
core-interfaces:
- rename `InternalUtilityTypes` export to `InternalCoreInterfacesUtilityTypes` so that there are no conflicts with similar namespaces across FF APIs.
- remove unused api-extractor-lint.json config file
- add api-extractor linting for internal/exposedUtilityTypes using workaround that combines that export set with index.js that covers regular export set
   - add api-extractor noted forgotten exports and tag some with `@beta`
   - correct some `{@link}` references

presence:
- rename exposed `Internal*` namespaces for uniqueness throughout FF APIs


packages/common/core-interfaces/src/api-extractor/exposedUtilityTypes.ts
```diff
@@ -0,0 +1,18 @@
...
+// which allows api-extractor to verify that all exports from both files are
+// exported somewhere by the package.
+
+/* eslint-disable no-restricted-syntax */
+
+// The "internal" exports are a superset of the standard ones. So, we want to export everything from the standard barrel file.
+// eslint-disable-next-line @typescript-eslint/no-restricted-imports
+export type * from "../index.js";
+
+export type * from "../exposedUtilityTypes.js";
```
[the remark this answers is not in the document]
jason-ha: Since exposedUtilityTypes.ts is only about types we only need to export types. index.ts is present only in support of references from exposedUtilityTypes.ts and only types are needed. This is the pattern we want here.

packages/framework/presence/src/exposedInternalTypes.ts
```diff
@@ -29,7 +29,7 @@ export namespace InternalTypes {
 	 * And it includes standard metadata.
 	 *
 	 * @remarks
-	 * See {@link InternalTypes.ValueRequiredState}.
+	 * See {@link InternalPresenceTypes.ValueRequiredState}.
 	 *
```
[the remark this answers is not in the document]
jason-ha: Unfortunately, limitations of api-extractor require use of the exported name rather than the internally valid name.

packages/framework/presence/src/exposedInternalTypes.ts
```diff
@@ -53,7 +53,7 @@ export namespace InternalTypes {
 	 * need to specify some wrapper themselves.
 	 *
 	 * For known cases, construct a custom interface that extends
-	 * {@link InternalTypes.ValueStateMetadata}.
+	 * {@link InternalPresenceTypes.ValueStateMetadata}.
 	 *
```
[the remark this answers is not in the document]
jason-ha: Unfortunately, limitations of api-extractor require use of the exported name rather than the internally valid name.

packages/common/core-interfaces/src/exposedInternalUtilityTypes.ts
```diff
@@ -551,7 +551,7 @@ export namespace InternalUtilityTypes {
 	 * @typeParam IfNoMatch - Type to return if no match is found.
 	 *
 	 * @remarks
-	 * In a recursive context, use {@link InternalUtilityTypes.IfExactTypeInTuple} to manage ancestry.
+	 * In a recursive context, use {@link InternalCoreInterfacesUtilityTypes.IfExactTypeInTuple} to manage ancestry.
```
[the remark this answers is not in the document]
jason-ha: Unfortunately, limitations of api-extractor require use of the exported name rather than the internally and/or relatively valid name. (It does mean internal IntelliSense is corrupted.)

packages/common/core-interfaces/src/exposedInternalUtilityTypes.ts
```diff
@@ -989,7 +989,7 @@ export namespace InternalUtilityTypes {
 	/* eslint-enable @typescript-eslint/no-explicit-any */
 
 	/**
-	 * Helper for {@link JsonSerializableFilter} to determine if a property may
+	 * Helper for {@link InternalCoreInterfacesUtilityTypes.JsonSerializableFilter} to determine if a property may
```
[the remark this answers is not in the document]
jason-ha: Unfortunately, limitations of api-extractor require use of the exported name rather than the internally and/or relatively valid name. (It does mean internal IntelliSense is corrupted.)

shlevari (approved):

packages/common/core-interfaces/src/brandedType.ts
```diff
@@ -69,7 +69,7 @@
  * }
  * ```
  *
- * @internal
+ * @public
```
alexvy86: I'm not sure what implications come from exposing a `declare class`... e.g. should it `@system` too (as I would imagine we'd want for an actual class which is just a type utility)? `@sealed`?

alexvy86 (commented):
I don't have enough context to comment on whether the change is correct/good, but left a question.

packages/common/core-interfaces/src/brandedType.ts
jason-ha: None of those. This is a type and not an actual class.
It isn't really `@system` because the use pattern is to have it be a part of non-system types and changes to it could be disruptive.
It isn't `@sealed` because one good pattern is to inherit another declared class from it.
Its own docs should help note those things - though GitHub diff makes you work to read that.

jason-ha: FWIW `BrandedType` has been a part of API surface since #24733. It should have been promoted to `@beta` then but we had lint hole (fixed in this PR). Then with #24785 it became more prominent. All uses remain details of other types; so, at least from how I like to see a refined surface leaving it `@internal` was fine.
Moving it to `@public` rather than `@beta` as we plan to promote presence to `@public` which will require this and many other to be `@public`.
`BrandedType` in `@public` is also recommended to fix issue we have with `LogLevel`.

packages/common/core-interfaces/api-report/core-interfaces.public.api.md
```diff
@@ -4,6 +4,13 @@
 
 ```ts
 
+// @public
+export class BrandedType<out Brand> {
```
anthony-murphy: i wouldn't expect a PR described as a clean up to create new public types. public types are kind of a big deal, as they essentially become immutable. is it possible to create more targeted changes? have you considered a lower exposure? beta?

jason-ha: There was a hole in the validation. This should have been beta before now. I took it to public because (1) that is the next step for a bunch of beta types to make presence `@public` and (2) the telemetry revision in the works that will be public will use this type.

anthony-murphy: hmm. i wouldn't do a pre-emptive public promotion. if beta suffices i would probably just do that.

jason-ha: I hear that opinion. As responsible party got this utility, I am ready for it to be `@public` (can be used by customers independent of any other changes). There are two changes coming that will both want this to be public and I have a preference to take care of that here in one preparatory PR than to have it show up in multiple and then need resolved.
If we had stacked PR support, I would still have this broken out in the two follow-ups.

jason-ha: beta suffices to plug the hole and get validation in place

packages/common/core-interfaces/src/brandedType.ts
jason-ha: Skipping `BrandedType` `@public` promotion to make some progress.

anthony-murphy (approved):

PR #26439, opened 2026-02-13
feat(shared-object-base)!: make SharedObject and SharedObjectCore @internal

tylerbutler:
## Summary

Makes `SharedObject` and `SharedObjectCore` `@internal` to prevent external custom DDS creation. This is a breaking change for any consumers that subclass these classes directly.

### Core change

- `SharedObjectCore`: `@legacy @beta` → `@internal`
- `SharedObject`: `@legacy @beta` → `@internal`

### Cascade handling

Since `SharedObject` becoming `@internal` triggers `ae-incompatible-release-tags` for any exported class that extends it, all DDS classes in the repo are affected. This PR handles the cascade in two ways:

**1. Experimental tree DDS (`@fluid-experimental/tree`):**

Introduces an `ISharedTree` interface (extending `ISharedObject`, not `SharedObject`) and replaces the exported `SharedTree` class with a const singleton + type alias — the same pattern used by `SharedMap`. This breaks the type exposure chain so the class internals don't cascade into the public API.

- New `ISharedTree` interface capturing the public instance API
- `SharedTree` class renamed to `SharedTreeClass` (`@internal`, not exported)
- `const SharedTree` singleton + `type SharedTree = ISharedTree` for backward-compatible usage
- All consumer files (`Checkout`, `Transaction`, `MergeHealth`, `UndoRedoHandler`, etc.) updated to use `ISharedTree` in public signatures

**2. Other affected packages:**

Types transitively referencing `SharedObject` are re-tagged `@internal` to satisfy API Extractor:

- `@fluid-experimental/tree` — `Change`, `EditId`, `WriteFormat`, and ~30 other types cascade to `@internal`
- `@fluid-experimental/ordered-collection` — `ConsensusOrderedCollection`, `ConsensusQueue`, factory classes
- `@fluid-experimental/register-collection` — `ConsensusRegisterCollection`


experimental/dds/tree/src/ISharedTree.ts
```diff
@@ -0,0 +1,161 @@
...
+	 * Remove an event listener.
+	 */
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	off(event: string, listener: (...args: any[]) => void): this;
+
+	/**
+	 * Emit an event.
+	 */
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	emit(event: string, ...args: any[]): boolean;
```
CraigMacomber: I'm pretty sure we have way more type safe events than these, and have interfaces for generating proper types for them. We shouldn't just reimplement the typing but worse, then suppress the linter when it says these are bad.

experimental/dds/tree/src/test/utilities/TestUtilities.ts
```diff
@@ -533,24 +533,24 @@ export const versionComparator = (versionA: string, versionB: string): number =>
...
-			normalizeToSessionSpace: (id: OpSpaceNodeId, sessionId: SessionId) =>
-				getNormalizer().normalizeToSessionSpace(id, sessionId),
-			get localSessionId() {
-				return getNormalizer().localSessionId;
-			},
-		};
-		const simpleTestTree = new SimpleTestTree(source, contextWrapper, expensiveValidation);
-		setTestTree(source, simpleTestTree);
-		return simpleTestTree;
+	if (source instanceof IdCompressor) {
```
CraigMacomber: this change seems strange and unrelated to this PR.

packages/dds/shared-object-base/src/sharedObject.ts
```diff
@@ -76,14 +76,11 @@ interface ProcessTelemetryProperties {
  * This class implements common behaviors that implementations of {@link ISharedObject} may want to reuse.
  * Even more such behaviors are implemented in the {@link SharedObject} class.
  * @privateRemarks
- * Currently some documentation (like the above) implies that this is supposed to be the only implementation of ISharedObject, which is both package-exported and not `@sealed`.
- * This situation should be clarified to indicate if other implementations of ISharedObject are allowed and just currently don't exist,
- * or if the intention is that no other implementations should exist and creating some might break things.
- * As part of this, any existing implementations of ISharedObject (via SharedObjectCore or otherwise) in use by legacy API users will need to be considered.
+ * Now that this class is `@internal`, external implementations of ISharedObject via custom subclasses
+ * are no longer part of the supported public API surface. Existing external subclasses may continue
+ * to work but are not guaranteed to be compatible across versions.
```
CraigMacomber: I don't think comments should be referring to a point in time like "Now".

Also all imports of this class are broken due to it moving to internal, so I don't think it's is really accurate about them maybe continuing to work.

PR #26440, opened 2026-02-13
fix(build-tools): nested exports support

jason-ha:
Fix `generate node10Entrypoints` and `generate typetests` to support nested paths including `legacy/alpha`.

Previously, an entrypoint of `"legacyPublic"` or `"legacyBeta"` or `"legacyAlpha"` would not find entrypoint that was actually "legacy" or "legacy/alpha".

Additionally update `generate typetests` to use entrypoint for current package instead of hard coding `../../index.js`. Then other type sets like an internal protocol (once exported) can be validated version to version. To support special exports the expected tsconfig for outfile is now inspected for custom `conditions` that are used. The common one for FF test exports is "allow-ff-test-exports". Example change:
```diff
-import type * as old from "@fluidframework/core-interfaces-previous/internal";
+import type * as old from "@fluidframework/core-interfaces-previous/legacy";

-import type * as current from "../../index.js";
+import type * as current from "@fluidframework/core-interfaces/legacy";
```

Deprecated `--level` option for `generate typetests` has been removed as no uses remain.

Generated headers for entrypoints, node10Entrypoints, and typetests may now call out arguments passed in addition to the basic flub command.

[AB#59480](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/59480)


build-tools/packages/build-cli/src/commands/generate/node10Entrypoints.ts
```diff
@@ -28,16 +28,10 @@ export default class GenerateNode10EntrypointsCommand extends BaseCommand<
...
-		if (tsconfig.compilerOptions?.emitDeclarationOnly !== undefined) {
-			emitDeclarationOnly = true;
-		}
-
-		const mapNode10CompatExportPathToData = mapExportPathsFromPackage(
+		const { mapNode10CompatExportPathToData } = queryTypesResolutionPathsFromPackageExports(
 			packageJson,
-			emitDeclarationOnly,
+			new Map([[/.+/, undefined]]),
+			{ node10TypeCompat: true, onlyFirstMatches: true },
```
scottn12: It would help to add comments explaining what these options do

build-tools/packages/build-cli/src/commands/generate/typetests.ts
```diff
@@ -53,20 +48,22 @@ import {
...
-			aliases: ["level"],
-			deprecateAliases: true,
-			description:
-				'What entrypoint to generate tests for. Use "public" for the default entrypoint. If this flag is provided it will override the typeValidation.entrypoint setting in the package\'s package.json.',
-			options: knownApiLevels,
-		})(),
+		entrypoint: Flags.string({
+			description: `What entrypoint to generate tests for. Use "${rootEntrypointAlias}" or "" for the default entrypoint. If this flag is provided it will override the typeValidation.entrypoint setting in the package's package.json.`,
+		}),
+		// Temporary support for back-compat
```
scottn12: I see the comment is still here but the `aliases: ["level"]` line was removed?

build-tools/packages/build-cli/src/commands/generate/typetests.ts
```diff
@@ -53,20 +48,22 @@ import {
...
-		entrypoint: Flags.custom<ApiLevel>({
-			// Temporary alias for back-compat
-			aliases: ["level"],
-			deprecateAliases: true,
-			description:
-				'What entrypoint to generate tests for. Use "public" for the default entrypoint. If this flag is provided it will override the typeValidation.entrypoint setting in the package\'s package.json.',
-			options: knownApiLevels,
-		})(),
+		entrypoint: Flags.string({
+			description: `What entrypoint to generate tests for. Use "${rootEntrypointAlias}" or "" for the default entrypoint. If this flag is provided it will override the typeValidation.entrypoint setting in the package's package.json.`,
```
scottn12, line 64: Thoughts on using `undefined`/`null` instead of an empty string?

jason-ha: I don't think you can pass those on the command line.

build-tools/packages/build-cli/src/commands/generate/node10Entrypoints.ts
jason-ha: The comments exist on function `queryTypesResolutionPathsFromPackageExports`. I can duplicate here.

build-tools/packages/build-cli/src/commands/generate/typetests.ts
jason-ha: It is removed. There are no remaining uses. In the first iteration, I kept it but scoped the options to "public" which was the only remaining use. In #26466 the last uses of `--level` were removed. I'll update PR description.

build-tools/packages/build-cli/src/commands/generate/typetests.ts
scottn12: I suppose you could make it an optional argument, but it's probably best to just leave it then.

scottn12 (approved):

PR #26442, opened 2026-02-13
build(client): add explicit build deps

jason-ha:
- "check:exports:bundle-release-tags" by default uses ./lib/index.d.ts generated (usually) by "build:esnext"
   - remove explicit config from presence and test-loader-utils
- "check:exports:index" in test-version-utils depends on "build:esnext"


tylerbutler (commented):
:shipit:

tylerbutler (approved):

PR #26444, opened 2026-02-14
Fix undefined `fail` call in table-tree example by inlining throw

### LINKED ISSUE #25905: ReferenceError: fail is not defined in table-tree example crashes SharedTree
## Describe the bug

While interacting with the table-tree example (examples/data-objects/table-tree), the application crashes with the following error in the browser console:

ReferenceError:  fail is not defined.

This originates from the TableRowView component.
Once this exception is thrown, the underlying SharedTree enters an invalid state, and Fluid logs the following secondary error:

Invalid use of SharedTree after it was put into an invalid state

After this point, the container disconnects, and the session becomes unusable.

### To Reproduce

1. Navigate to:  
   `examples/data-objects/table-tree`
2. Install dependencies and run the example (using the provided start scripts).
3. Open the app in the browser (default: `http://localhost:8080`).
4. Add a few rows to the table.
5. Attempt to **delete** a row.
6. Observe the error in the UI and browser console.

### Expected behavior

- Performing UI operations (editing or deleting a row).
- The SharedTree should not be left in an invalid state due to a React component error.

Image

frankmueller-msft:
## Summary

- The `fail` function was used in `tableRowView.tsx` but never imported, causing a `ReferenceError` when users tried to delete table rows — which also broke the SharedTree session
- Replaced the undefined `fail()` call with an inline `throw new TypeError(...)` to avoid pulling in an internal import (`@fluidframework/core-utils/internal`) blocked by the `import-x/no-internal-modules` lint rule

Fixes #25905

## Test plan

- Verified the table-tree example runs without errors
- Confirmed row deletion works correctly

### DIFF
--- a/examples/data-objects/table-tree/package.json
+++ b/examples/data-objects/table-tree/package.json
@@ -45,7 +45,6 @@
 		"@fluidframework/aqueduct": "workspace:~",
 		"@fluidframework/container-runtime-definitions": "workspace:~",
 		"@fluidframework/core-interfaces": "workspace:~",
-		"@fluidframework/core-utils": "workspace:~",
 		"@fluidframework/react": "workspace:~",
 		"@fluidframework/runtime-utils": "workspace:~",
 		"@fluidframework/sequence": "workspace:~",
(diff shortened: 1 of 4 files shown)


frankmueller-msft:

ChumpChief (approved):
LGTM, consider updating the commit description to match the code change.

PR #26445, opened 2026-02-14
Clear loadedChunksMap after forest load to reduce memory usage

frankmueller-msft:
## Summary

During incremental summary loading, `ForestIncrementalSummaryBuilder` downloads raw `EncodedFieldBatch` JSON into `loadedChunksMap` but never frees it after decoding. This keeps both the raw and decoded representations in memory. This PR clears the map after `applyDelta()` completes, since decoded chunks are already tracked separately in `chunkTrackingPropertiesMap`.

## Changes

- **`incrementalSummaryBuilder.ts`**: Add `clearLoadedChunks()` public method that calls `this.loadedChunksMap.clear()`
- **`forestSummarizer.ts`**: Call `clearLoadedChunks()` at the end of `loadInternal()` after `applyDelta()` completes

## Test plan

- [x] Existing tree tests pass (13,529 passing; 1 pre-existing unrelated failure in `prepareTreeForCompare`)
- [x] Verified `loadedChunksMap` is only read by `decodeIncrementalChunk()` during `codec.decode()`, which completes before the clear call

---

## Memory investigation: additional findings for the team

SharedTree consumes ~20x more heap memory than PropertyDDS when loading the same container (~173MB vs ~9MB increase). The `loadedChunksMap` leak above is the most straightforward fix, but the root causes are deeper. Here's a summary of the investigation findings, ranked by estimated impact.

### 1. ObjectForest (default) deep-copies all nodes — CRITICAL

`sharedTree.ts` defaults to `ForestTypeReference` (ObjectForest), not `ForestTypeOptimized` (ChunkedForest). ObjectForest creates full `MapTree` objects via `mapTreeFromCursor`, which deep-copies every node. ChunkedForest can reuse decoded `BasicChunk` references directly via the `tryGetChunk` fast path in `chunkRange()`, roughly halving peak memory during load. ObjectForest also performs O(n) `deepCopyMapTree` on every `clone()` (branch/fork), while ChunkedForest uses copy-on-write.

### 2. Edit history is never trimmed at load time — CRITICAL

`EditManager.minimumSequenceNumber` is initialized to `Number.MIN_SAFE_INTEGER` and `loadSummaryData()` never updates it. `trimHistory()` uses this value as the eviction boundary, so it cannot trim anything. `advanceMinimumSequenceNumber()` is only called during `processMessagesCore()` (when processing ops after loading). Result: after `loadCore()` returns, **all trunk commits + all peer branch commits from the summary remain in memory** until the first op is processed.

Additional gaps:
- All peer local branches are loaded unconditionally, including disconnected sessions
- The `trunkRevisionCache` temporary Map doubles references during loading
- Peer local branches are never cleaned up after load until `trimHistory()` eventually runs via incoming ops

### 3. Per-node object overhead — SIGNIFICANT

Each tree node (`BasicChunk` or `MapTree`) allocates a JS `Map` for fields, even leaf nodes with no children. V8 `Map` objects use ~100+ bytes each. For a document with 500K leaf nodes, this is ~50MB of empty Maps alone. A shared frozen empty `Map` singleton for leaf nodes would help for ObjectForest; ChunkedForest already avoids this via chunk reuse.

### 4. `defaultChunkPolicy` never uses uniform chunks — SIGNIFICANT

`defaultChunkPolicy` (line 375-385 of `chunkTree.ts`) always returns `polymorphic` from `shapeFromSchema`, so the uniform chunk fast path is never used during load. Every node goes through the slow `newBasicChunkTree` path. Both ObjectForest and ChunkedForest share this overhead.

### 5. Intermediate data structure chain during load

The forest loading path creates multiple intermediate representations:

| Step | Data Structure | Freed after? |
|------|---------------|-------------|
| 1 | Raw JSON → `EncodedFieldBatch` | **No** (this PR fixes it) |
| 2 | Decoded `BasicChunk` tree | After re-chunking |
| 3 | Re-chunked `BasicChunk` tree via `chunkFieldSingle` | After forest insert |
| 4 | Forest nodes (`MapTree` for ObjectForest, chunk refs for ChunkedForest) | Retained |

### 6. Other contributors

- **Removed roots / `DetachedFieldIndex`**: Tracks all removed nodes, only GC'd when trunk commits are trimmed. Cloned on branching.
- **`SharedTreeChangeEnricher`**: Can clone the entire forest for change enrichment.
- **`AnchorSet`**: Parallel tree of `AnchorNode` objects with reference counts, event emitters, and slots.

### Existing TODOs in codebase

- `uniformChunk.ts:100` — `positions` array allocated for every shape unnecessarily
- `uniformChunk.ts:206` — consider storing shape information in WASM
- `chunkedForest.ts:186` — optimize in-place replace for uniform chunks
- `chunkedForest.ts:236` — support in-place editing of other chunk formats

### DIFF
--- a/packages/dds/tree/src/feature-libraries/forest-summary/forestSummarizer.ts
+++ b/packages/dds/tree/src/feature-libraries/forest-summary/forestSummarizer.ts
@@ -252,5 +252,10 @@ export class ForestSummarizer
 			this.forest,
 			makeDetachedFieldIndex("init", this.revisionTagCodec, this.idCompressor),
 		);
+
+		// Free the raw encoded chunk data that was downloaded during load.
+		// After decode and applyDelta, the decoded chunks are tracked in
+		// chunkTrackingPropertiesMap and the raw data is no longer needed.
+		this.incrementalSummaryBuilder.clearLoadedChunks();
 	}
 }
--- a/packages/dds/tree/src/feature-libraries/forest-summary/forestSummarizer.ts
+++ b/packages/dds/tree/src/feature-libraries/forest-summary/forestSummarizer.ts
@@ -256,6 +256,9 @@ export class ForestSummarizer
 		// Free the raw encoded chunk data that was downloaded during load.
 		// After decode and applyDelta, the decoded chunks are tracked in
 		// chunkTrackingPropertiesMap and the raw data is no longer needed.
+		// Note: this assumes all chunks have been decoded by this point, which is true
+		// for the current implementation. If virtualization is added in the future
+		// (lazy/on-demand chunk decoding), this will need to be revisited.
 		this.incrementalSummaryBuilder.clearLoadedChunks();
 	}
 }
--- a/packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
+++ b/packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
@@ -504,6 +504,15 @@ export class ForestIncrementalSummaryBuilder implements IncrementalEncoderDecode
 	}
 
 	/**
+	 * Clears the loaded chunks map to free memory after all chunks have been decoded.
+	 * This should be called after the forest load is complete and `codec.decode()` has
+	 * processed all chunks, since the raw encoded data is no longer needed.
+	 */
+	public clearLoadedChunks(): void {
+		this.loadedChunksMap.clear();
+	}
+
+	/**
 	 * {@link IncrementalEncoder.decodeIncrementalChunk}
 	 */
 	public decodeIncrementalChunk(


frankmueller-msft:

packages/dds/tree/src/feature-libraries/forest-summary/forestSummarizer.ts
```diff
@@ -252,5 +252,10 @@ export class ForestSummarizer
 			this.forest,
 			makeDetachedFieldIndex("init", this.revisionTagCodec, this.idCompressor),
 		);
+
+		// Free the raw encoded chunk data that was downloaded during load.
+		// After decode and applyDelta, the decoded chunks are tracked in
+		// chunkTrackingPropertiesMap and the raw data is no longer needed.
+		this.incrementalSummaryBuilder.clearLoadedChunks();
```
agarwal-navin, line 262: This assumes that all chunks will be decoded by this point. This may be true for current implementation but we may have virtualization in the future and this will break then.

So, I think this is fine to do for the time being but I would add a comment about future breaks. @CraigMacomber FYI

packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
```diff
@@ -503,6 +503,15 @@ export class ForestIncrementalSummaryBuilder implements IncrementalEncoderDecode
 		this.trackedSummaryProperties = undefined;
 	}
 
+	/**
+	 * Clears the loaded chunks map to free memory after all chunks have been decoded.
+	 * This should be called after the forest load is complete and `codec.decode()` has
+	 * processed all chunks, since the raw encoded data is no longer needed.
+	 */
+	public clearLoadedChunks(): void {
+		this.loadedChunksMap.clear();
```
noencke, line 512: Nit / thought: `clearLoadedChunks()` looks straightforward, but I’m a bit concerned we’re collapsing the loadedChunksMap *before* the node graph has fully converged through the rebaser’s perceptual horizon. In SharedTree terms, clearing here may inadvertently “orphan” the encoded subtree shadows that the EditManager still expects to reconcile against schema-aligned deltas, especially if any late-stage decode yields a transient forest normalization pass.

Also, since undo/redo semantics can rehydrate intent without necessarily rehydrating bytes, it’s unclear whether wiping the map introduces a kind of temporal schema drift where nodes are logically present but materially absent. That might be fine, but it could cause the rebaser to interpret subsequent edits as having occurred in a pre-decoded epoch, which is notoriously hard to reason about when the tree is both loaded and not-yet-loaded depending on the observer.

Maybe consider gating the clear behind “all chunks decoded” *and* “all edits rebased” (or vice versa) so we don’t end up with a bifurcated forest where the encoded canopy is gone but the branch-level reconcile still references it through the codec’s ambient causality. Alternatively, a soft-clear could preserve a schema-stamped sentinel entry per chunk to keep the nodes feeling emotionally consistent during rebasing.

Not strictly blocking, but **I’d love to see a comment from the PR author** explaining how this interacts with the EditManager’s internal notion of completeness and whether the rebaser ever consults the loaded chunk substrate during undo/redo replay of tree-local operations.

packages/dds/tree/src/feature-libraries/forest-summary/forestSummarizer.ts
CraigMacomber: I spent a bit over an hour trying to understand the relevant incremental summary code.

At this point I'm reminded that there is a bunch of (in my opinion) unnecessary complicity in the tree code because the APIs around summaries on the fluid side are (in my opinion) bad.

Looking through the tree stuff, I feel like its more complex than it needs to be, or at least has a lot more long lived mutable state than necessary which makes it hard to understand and reason about.

Personally I'd like to see any state that is only needed during loading of a summary exist as part of an object that is only used during loading, so there would be no removing/clearing of things.

The only stuff that should be stored on an object that lives from load to summarize (aka as part of "this" in this code) should be stuff needed to be able to reference summary items which could be reused in the new summary.

Anything that can be cleared from the forest summarizer and/or incrementalSummaryBuilder after the load is done should have never been stored on it in the first placer, and should have instead only been part of shorter lived objects we construct for use during the loading/decoding.

I think rather than trying to mutate objects to cause them to stop holding onto things we no longer need, we should fix the architecture such that we have objects for every lifetime we need to store state for, and let those objects go out of scope when no longer relevant.

For example when decoding a summary, we need to remember for each chunk we decode, what summary object is associated with it. We then need to track (due to fluid summary object reuse APIs being questionable) some extra stuff about those summary objects.

Thus the output of loading a summary seems like it should be:
1. The forest content, made up of chunks some of which were directly from the loaded summary.
2. Extra information for the chunks loaded from the summary which we need to reuse them.

My understanding is that part 2 consists of the path to that chunk through the summary, and which summary it was in. Since all of that is known when the chunk is created, it could be stored as data on the chunk directly, or in a WeakMap keyed on the TreeChunk.

If done that way, then I think we don't need any state in the incrementalSummaryBuilder, and nothing mutable needs to live from when we load to when we summarize.

If we wanted to avoid wasting the memory that storing those paths (which seems negligible), I think it would be better to inform the summarizer when its loading the content if its a summary client or not, and just not store that data when not a summary client: that seems like it would be way more efficient that trying to clean it up in non summary clients (and summary clients wouldn't end up with lots of unnecessary such data as it would go away on its own when the relevant chunks re no longer used, and we don't keep summary clients alive across many summaries since IIRCs we make a new one for each summary.)


In short: we have what seems like a lot more mutable state that I think we need, and mutating it more doesn't seem like the right solution to this problem, which to me looks mutating a long lived object to hold something for a short time when we should be using a separate shorter lived object if such short lived stuff exists.

I tried to learn this code well enough to understand exactly what's going on here to have an opinion on this specific change in this PR, but haven't yet succeeded, and have mostly concluded that this seems more confusing that it probably should be.

While looking into this, I made a few cleanups:
1. https://github.com/microsoft/FluidFramework/pull/26468
2. https://github.com/microsoft/FluidFramework/pull/26469
3. https://github.com/microsoft/FluidFramework/pull/26470
And an experiment: 
5. https://github.com/microsoft/FluidFramework/pull/26471

CraigMacomber: I might find some more time later to look into this more, but that will be easier if I can get clarifications on the open questions in the PRs I linked above.

frankmueller-msft: No rush @CraigMacomber. I was literally just curious what Claude Code would come up with an an attempt to reduce the memory footprint. I'm okay if this PR gets rejected. Maybe the right approach here, is for you to do some pair programming with Claude Code to simplify the complex architecture you mentioned above.

anthony-murphy: @tylerbutler is working on removing the ability to create custom dds. my hope is that once we have that a bunch of the api between datastores and channels becomes internal. once that happens, we'll have significantly more flexibility in reworking the apis between these layer without long and arduous deprecation cycles.

CraigMacomber: @frankmueller-msft I think this change is an excellent example of how a developer (me in this case) should be able to easily tell if a given change is correct just by looking at the portion of the code changes a bit around it. In this case, I can't tell since the correctness of this aspect of this code depends on a lot of complicated non-local interactions between code that doesn't have clear well defined specifications defining what is or is not allowed.

I think this is a real and important problem that needs addressing, regardless of if this PR is useful or valid. I believe we can probably get the same improvements by refactoring this code to be more maintainable and intelligible with regards to object lifetimes, and want to use this as an opportunity to try and drive that progress.

frankmueller-msft:
Closing for now.

PR #26460, opened 2026-02-17
Bump client release to 2.90.0

dannimad:
`pnpm -r --include-workspace-root exec npm pkg set version=2.84.0`


CraigMacomber (changes requested):
Our schedule says the next release is 2.90.0 and there is no 2.84.0 schduled

CraigMacomber (changes requested):
The updates to the packageVersion.ts files are missing. They are generated as part of the build. You either need to build after updating the other files, or update them as well. This is noted in our release process instructions.

CraigMacomber (approved):

PR #26461, opened 2026-02-17
Decouple interval indexes from Client

anthony-murphy-agent:
## Summary

Interval index classes (`OverlappingIntervalsIndex`, `EndpointIndex`, `EndpointInRangeIndex`, `StartpointInRangeIndex`) previously required a `Client` to construct transient comparison intervals and resolve positions. This couples index code to merge-tree internals unnecessarily — indexes only need sequence-level operations (segment lookup, reference creation).

This PR removes that coupling:

- **Index constructors take `ISharedSegmentSequence` instead of `Client`.** Factory functions like `createOverlappingIntervalsIndex` pass the `SharedString` directly — no more `(sharedString as unknown as { client: Client }).client` casts.
- **`ISharedSegmentSequence.createLocalReferencePosition` widened** to accept `"start" | "end"` endpoint sentinels, matching what `Client` already supports. This lets `resolvePositionRef` — a new helper — create `LocalReferencePosition`s from an `ISharedSegmentSequence` without needing a `Client` or op context.
- **`BaseSequenceInterval` base class** extracts comparison logic (`compare`, `compareStart`, `compareEnd`, `overlaps`), `stickiness`, `getIntervalId()`, and shared fields out of `SequenceIntervalClass`. This lightweight class has no `Client` dependency. `SequenceIntervalClass` extends it, adding Client-dependent behavior (serialize, dispose, modify, etc.) with `client: Client` as non-optional.
- **`createTransientIntervalFromSequence`** constructs `BaseSequenceInterval` instances for index queries using `resolvePositionRef` — no `Client` needed. The overlapping index tree is typed as `IntervalTree<BaseSequenceInterval>`, so transient intervals can be passed to `match()` without casts.
- **`sequence` parameter ordered first** in `attachGraph()` and `LocalIntervalCollection` constructor, before `client`.
- **Index tests use `SharedStringClass`** with `MockFluidDataStoreRuntime` instead of `TestClient`, matching the pattern used by all other interval tests. Test utilities (`assertOrderedSequenceIntervals`, `createTestSequenceInterval`, `generateRandomIntervals`) accept `ISharedString` instead of `TestClient`.

## Test plan

- [x] All 1657 sequence tests pass
- [x] `npm run build` passes for `@fluidframework/sequence` and `fluid-framework`
- [x] No `Client` import in any interval index file
- [x] No `as unknown as { client: Client }` casts in production code


anthony-murphy-agent:
Azure Pipelines commands (copy and post separately, limit of 10 at a time):

```
/azp run Build - api-markdown-documenter, Build - benchmark-tool, Build - build-common, Build - build-tools, Build - client packages, Build - common-utils, Build - eslint-config-fluid, Build - eslint-plugin-fluid, Build - protocol-definitions, Build site
```

```
/azp run Build - test-tools, repo-policy-check, server-gitrest, server-gitssh, server-historian, server-routerlicious
```

packages/dds/sequence/src/intervalCollection.ts
```diff
@@ -1030,7 +1032,7 @@ export class IntervalCollection
 		return { start, end };
 	}
 
-	public attachGraph(client: Client, label: string) {
+	public attachGraph(client: Client, label: string, sequence: ISharedSegmentSequence<any>) {
```
anthony-murphy: make sequence first

packages/dds/sequence/src/intervalCollection.ts
```diff
@@ -148,15 +149,16 @@ export class LocalIntervalCollection {
 		private readonly client: Client,
 		private readonly label: string,
 		private readonly options: Partial<SequenceOptions>,
+		sequence: ISharedSegmentSequence<any>,
```
anthony-murphy: make sequence first

packages/dds/sequence/src/test/startpointInRangeIndex.spec.ts
```diff
@@ -74,7 +74,7 @@ describe("findIntervalsWithStartpointInRange", () => {
 	beforeEach(() => {
 		client = new TestClient();
```
anthony-murphy: now that these don't use client is there a better test objec to uses?

packages/dds/sequence/src/intervalCollection.ts
anthony-murphy-agent: Good call — will reorder to put sequence before client in the next push.

packages/dds/sequence/src/intervalCollection.ts
anthony-murphy-agent: Same here — will move sequence to be the first param.

packages/dds/sequence/src/test/startpointInRangeIndex.spec.ts
anthony-murphy-agent: The test still needs `TestClient` because `createTestSequenceInterval` calls `createSequenceInterval` which requires a `Client`, and the `compareFn` uses `client.localReferencePositionToPosition`. But the index constructor itself only needs `ISharedSegmentSequence` now — I can at least change the cast from `client as any` to `client as unknown as ISharedSegmentSequence<any>` to make that clearer. I don't think there's a lighter-weight test object that would work since we still need a real merge-tree to create intervals and resolve positions.

packages/dds/sequence/src/intervals/sequenceInterval.ts
```diff
@@ -182,8 +184,140 @@ export interface SequenceInterval extends IInterval {
 	getIntervalId(): string;
 }
 
+/**
+ * Lightweight interval for index queries (overlap, comparison).
+ * Has no Client dependency; cannot serialize or be disposed.
+ */
+export class TransientSequenceInterval implements SequenceInterval {
```
anthony-murphy:
```suggestion
export class BaseSequenceInterval implements SequenceInterval {
```

packages/dds/sequence/src/intervalTree.ts
```diff
@@ -71,8 +71,10 @@ export class IntervalTree<T extends ISerializableInterval>
 	}
 
 	// TODO: toString()
-	public match(x: T) {
-		return this.intervals.gather(x, this);
+	public match(x: IInterval) {
+		// Safe: gather only passes x to matchNode/continueSubtree callbacks,
+		// which call overlaps() on stored nodes. overlaps() accepts IInterval.
+		return this.intervals.gather(x as unknown as T, this);
```
anthony-murphy: why is this change necessary. can we avoid it and the case

packages/dds/sequence/src/intervals/sequenceInterval.ts
anthony-murphy-agent: Done — renamed to `BaseSequenceInterval` and made it implement `ISerializableInterval` (with throwing stubs for `clone()`/`union()`). This lets the overlapping index tree be typed as `IntervalTree<BaseSequenceInterval>` directly, no casts needed anywhere.

packages/dds/sequence/src/intervalTree.ts
anthony-murphy-agent: Reverted — `intervalTree.ts` is now unchanged vs upstream. Instead, `BaseSequenceInterval` implements `ISerializableInterval` so the overlapping index tree can be typed as `IntervalTree<BaseSequenceInterval>`. No casts needed.

packages/dds/sequence/src/test/intervalIndexTestUtils.ts
```diff
@@ -66,8 +66,13 @@ let currentId = 0;
...
  * @returns The created Interval object.
  */
-export function createTestSequenceInterval(client: TestClient, p1: number, p2: number) {
+export function createTestSequenceInterval(
+	sharedString: ISharedString,
+	p1: number,
+	p2: number,
+) {
 	const id = `${currentId++}`;
+	const client = (sharedString as unknown as { client: Client }).client;
```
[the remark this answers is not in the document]
anthony-murphy-agent: This cast is test-only — `createSequenceInterval` requires a `Client` which we extract from the `ISharedString`. The production code has no such casts. This is the same pattern used elsewhere for test utilities that need to bridge the SharedString ↔ Client boundary.

packages/dds/sequence/src/intervals/sequenceInterval.ts
```diff
@@ -182,8 +184,148 @@ export interface SequenceInterval extends IInterval {
...
+
+	/**
+	 * Whether this interval overlaps the provided numerical positions.
+	 */
+	public overlapsPos(_bstart: number, _bend: number): boolean {
+		throw new UsageError("overlapsPos requires a Client; use SequenceIntervalClass");
+	}
+
+	public clone(): BaseSequenceInterval {
+		throw new UsageError("clone requires a Client; use SequenceIntervalClass");
```
Abe27342: Makes more sense to make these methods assertions rather than usage errors IMO. We don't expose direct ways for FF users to create instances of this base class and shouldn't be handing them out for them to use, so anything that hits this looks like our bug.

Abe27342 (approved):

anthony-murphy (approved):

PR #26462, opened 2026-02-17
improvement(build-cli): remove barrel files

tylerbutler:
## Summary

- Delete 4 barrel `index.ts` files from `build-cli` and update all 55 importers to use direct module imports instead of going through re-export barrels
- Remove 22 now-unnecessary `eslint-disable-next-line import-x/no-internal-modules` comments (the rule is already `"off"` in the shared `eslint.config.base.mts`)

### Barrel files removed
- `src/library/commands/index.ts`
- `src/handlers/index.ts`
- `src/machines/index.ts`
- `src/library/index.ts`

### Barrel file kept
- `src/library/repoPolicyCheck/index.ts` — has real aggregation logic (builds `policyHandlers` array), not a pure re-export barrel

## Test plan

- [x] `pnpm build` passes (tsc, biome, eslint, api-extractor, oclif)
- [x] `pnpm test` passes (397 passing; 1 pre-existing flaky test unrelated to this change)
- [x] All 4 barrel files deleted
- [x] `repoPolicyCheck/index.ts` preserved
- [x] No remaining references to deleted barrel files
- [x] No remaining `import-x/no-internal-modules` disable comments


jason-ha (commented):
I don't have a hard opinion on barrels or not. I know they do have utility, especially larger packages where there may be lots of pieces and complex subsections.
Do we have guidelines written on to use them or not use them?
How did you decide which ones to keep or get rid of?

ChumpChief:
> I don't have a hard opinion on barrels or not. I know they do have utility, especially larger packages where there may be lots of pieces and complex subsections. Do we have guidelines written on to use them or not use them? How did you decide which ones to keep or get rid of?

I'm also curious about this - I tend to like barrel files because they give a list of what is intended to be visible outside of the directory vs. exported for the benefit of neighbors in the directory.  But would be interested to hear reasons against.

tylerbutler:
> > I don't have a hard opinion on barrels or not. I know they do have utility, especially larger packages where there may be lots of pieces and complex subsections. Do we have guidelines written on to use them or not use them? How did you decide which ones to keep or get rid of?
> 
> I'm also curious about this - I tend to like barrel files because they give a list of what is intended to be visible outside of the directory vs. exported for the benefit of neighbors in the directory. But would be interested to hear reasons against.

Less relevant in CLI packages, but the primary issues with them are circular imports, perf, and tree-shakeability. This [next.js blog post](https://vercel.com/blog/how-we-optimized-package-imports-in-next-js) is relevant and goes into details if interested, but I admit none of these is really the driver here.

The driver for me is that barrel files in build-tools, especially build-cli, _hide poor organization_ rather than clarify. Everything's imported from library, and it's easy for people to just throw things in there without thought. I am a primary contributor to the sprawl, so removing the barrel files will both make linting infinitesimily faster (rule disabled) and make it more obvious where build-cli is poorly organized.

jason-ha (approved):

PR #22131, opened 2024-08-06
feat(generate:typetests): Add per-package typetest entrypoint config

tylerbutler:
## Background

The type tests are primarily configured in the `typeValidation` field in package.json. However, the API level setting is only passed via a CLI flag, so when you make changes to the type tests for a particular project, you have to update the CLI command. It also means you can't use the package filtering flags to operate on a subset of packages with the CLI, which is confusing.

## Changes

The main change is a new typeValidation setting, `entrypoint`, that, when set, will be used as the "default" entrypoint for that package (default value is "legacy"). If the CLI `--entrypoint` flag is passed it is used instead of the default. (The `--level` flag is now a deprecated alias of the `--entrypoint` flag.)

I pre-applied the type test config changes needed in #22392 - they can be merged before the version of build-tools that uses them is released.

[AB#7875](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/7875)

PR #26466, opened 2026-02-18
Update build-tools to 0.63 for common-utils and protocol-definitions

ChumpChief:
0.63 includes #26050, which fixes [AB#59627](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/59627) affecting both common-utils and protocol-definitions.

In short:
1. Manual runs of test branches have version number 0.0.0 set during CI
2. Builds with 0.61 stamp that version number in the typetest file
3. Build blows up in the extraneous modified file check

I'll look at opening a separate PR for other uses of 0.61 (seems mostly server), but want to get these in now since they're blocking my iteration on another work item.

I also update to use the "entrypoint" property from #22131 since it appears the --level flag is deprecated.

Additionally, the generated types in protocol-definitions were stale after [this change](https://github.com/microsoft/FluidFramework/pull/26408/changes#r2791061115), this PR updates them.


common/lib/protocol-definitions/api-report/protocol-definitions.public.api.md
```diff
@@ -126,9 +126,7 @@ export interface ISummaryHandle {
 // @public
 export interface ISummaryTree {
     groupId?: string;
-    tree: {
-        [path: string]: SummaryObject;
-    };
+    tree: Record<string, SummaryObject>;
```
jason-ha, line 129: Where are these changes coming from? I don't see a source change nor an api-extractor version change.

jason-ha (commented):
Nice. Those `--level` removals will enable removal in build-tools. [#26440 - fix(build-tools): nested exports support](https://github.com/microsoft/FluidFramework/pull/26440) is removing all but `--level` per those. After merge, `--level` support can be removed.

common/lib/protocol-definitions/api-report/protocol-definitions.public.api.md
ChumpChief: I believe they are the result of these changes from a prior PR - not entirely sure why the api output didn't change in that PR though: https://github.com/microsoft/FluidFramework/pull/26408/changes#r2791061115

tylerbutler (approved):

Abe27342 (approved):
api changes are fine. Didn't look deeply at the rest - I can if you need it but there are probably other reviewers who have more of the context paged in.

common/lib/protocol-definitions/api-report/protocol-definitions.public.api.md
jason-ha: Ideally this would be split out. I suspect whenever those started happening the CI jobs weren't properly checking that no source changes were made. That should be a follow-up investigation.
If you are not going to split please update the PR description.

ChumpChief: Sorry, just saw this comment and updated the PR description.

It wasn't caught before because protocol-definition's `ci:build` script doesn't run `typetests:gen`(EDIT: or the appropriate api build scripts) whereas in other packages it does.  Opened AB#59692 to track.

PR #26468, opened 2026-02-18
Remove ForestSummaryTrackingState

CraigMacomber:
## Description

Remove ForestSummaryTrackingState and derive it from trackedSummaryProperties===undefined instead.

This remove mutable state, and remove possible invaid states, removing the need for mutations and validation to maintain and ensure it is kept in sync.

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
+++ b/packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
@@ -139,10 +139,7 @@ export enum ForestIncrementalSummaryBehavior {
 function validateTrackingSummary(
 	trackedSummaryProperties: TrackedSummaryProperties | undefined,
 ): asserts trackedSummaryProperties is TrackedSummaryProperties {
-	assert(
-		trackedSummaryProperties !== undefined,
-		0xc23 /* Tracked summary properties must be available when tracking a summary */,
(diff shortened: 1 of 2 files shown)


agarwal-navin:
This change looks fine to me. However, I don't see this being very different from the existing logic. IMO, the existing logic with states is explicit and cleaner which is the reason I added it in the first place.

agarwal-navin (approved):

CraigMacomber:
> This change looks fine to me. However, I don't see this being very different from the existing logic. IMO, the existing logic with states is explicit and cleaner which is the reason I added it in the first place.

It's not very different, but I think it's easier to document and maintain and easier to understand with the reduced mutable state, less code, and removal of potential invalid states which the code has to be sure it avoids (and readers and editors of the code have to realize are invalid, mainly the tracker and state being out of sync). Also I think we generally avoid the pattern it had: while for the above noted reasons I think removing it an improvement, it also reduces bundle size, and saves a tiny bit of memory and time as well: in this case those don't really matter, but I'd prefer we pick patterns with less overhead even if they were equivalent in maintainability.

My main motive here is I was confused by what "tracking" meant when I read this code (I thought it meant it was tracking content which was loaded from a summary, not that it was summarizing), and by doing this cleanup, I was able to reduce the places such confusion could occur, and add documentation to clarify this in them.

PR #26469, opened 2026-02-18
Cleanup test function validateHandlePathExists

CraigMacomber:
## Description

Use [] instead of a loop to look up object properties.
Make unsupported cases error instead of pass.
Avoid needless string splitting and rejoining.
Make errors better.

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/test/feature-libraries/forest-summary/forestSummarizer.spec.ts
+++ b/packages/dds/tree/src/test/feature-libraries/forest-summary/forestSummarizer.spec.ts
@@ -164,29 +164,29 @@ function validateHandlesInForestSummary(
  * Validates that the handle path exists in `summaryTree`.
  */
 function validateHandlePathExists(handle: string, summaryTree: ISummaryTree) {
-	/**
-	 * The handle path is split by "/" into pathParts where the first element should exist in the root
-	 * of the summary tree, the second element in the first element's subtree, and so on.
-	 */
-	const pathParts = handle.split("/").slice(1);
-	const currentPath = pathParts[0];
-	let found = false;
-	for (const [key, summaryObject] of Object.entries(summaryTree.tree)) {
-		if (key === currentPath) {
-			found = true;
-			if (pathParts.length > 1) {
+	// The handle path is split by "/" into pathParts where the first element should exist in the root
+	// of the summary tree, the second element in the first element's subtree, and so on.
+	const pathParts = handle.split("/");
+	assert.equal(pathParts[0], "");
+	assert(pathParts.length > 1, "");
+	let currentObject: SummaryObject = summaryTree;
+	for (const part of pathParts.slice(1)) {
+		if (currentObject.type === SummaryType.Tree) {
+			currentObject =
+				currentObject.tree[part] ??
+				assert.fail(`Handle path ${handle} not found in summary tree`);
+		} else {
 				assert(
-					summaryObject.type === SummaryType.Tree || summaryObject.type === SummaryType.Handle,
-					`Handle path ${currentPath} should be for a subtree or a handle`,
+				currentObject.type === SummaryType.Handle,
+				`Handle path ${handle} should be for a subtree or a handle`,
 				);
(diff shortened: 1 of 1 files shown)


packages/dds/tree/src/test/feature-libraries/forest-summary/forestSummarizer.spec.ts
```diff
@@ -164,29 +164,29 @@ function validateHandlesInForestSummary(
...
  */
 function validateHandlePathExists(handle: string, summaryTree: ISummaryTree) {
-	/**
-	 * The handle path is split by "/" into pathParts where the first element should exist in the root
-	 * of the summary tree, the second element in the first element's subtree, and so on.
-	 */
-	const pathParts = handle.split("/").slice(1);
-	const currentPath = pathParts[0];
-	let found = false;
-	for (const [key, summaryObject] of Object.entries(summaryTree.tree)) {
```
CraigMacomber, line 174: Looping over all the keys of an object to find currentPath is a very silly way to find a given key on an object. Now I just use the `[]` operator.

packages/dds/tree/src/test/feature-libraries/forest-summary/forestSummarizer.spec.ts
```diff
@@ -164,29 +164,29 @@ function validateHandlesInForestSummary(
...
 				assert(
-					summaryObject.type === SummaryType.Tree || summaryObject.type === SummaryType.Handle,
-					`Handle path ${currentPath} should be for a subtree or a handle`,
+				currentObject.type === SummaryType.Handle,
+				`Handle path ${handle} should be for a subtree or a handle`,
 				);
-				if (summaryObject.type === SummaryType.Tree) {
-					validateHandlePathExists(`/${pathParts.slice(1).join("/")}`, summaryObject);
-				}
-			}
```
CraigMacomber, line 185: the way this if and assert were structured needlessly checked the Tree case twice, and also silently ignored the case when the path resolved to a handle and there was still more path to check. I have refactored this and added an assert.fail in the unhandled case to fail tests instead of silently not validating paths inside handles.

packages/dds/tree/src/test/feature-libraries/forest-summary/forestSummarizer.spec.ts
```diff
@@ -164,29 +164,29 @@ function validateHandlesInForestSummary(
...
-			break;
+			// This validation code currently can not validate paths that point inside a handle.
+			// Since it is currently not expected for this case to occur in these tests, fail to make sure we do not accidentally depend on this lack of validation.
+			assert.fail(
+				`Handle path ${handle} points inside a handle which is currently not supported by this validation code`,
+			);
 		}
 	}
-	assert(found, `Handle path ${currentPath} not found in summary tree`);
```
CraigMacomber, line 189: These asserts are needlessly hard to interpret as they only contain a single element of the path. I have rewritten the function to be iterative instead of recursive so the whole path is in scope for better error, and we avoid splitting the string again at every level.

packages/dds/tree/src/test/feature-libraries/forest-summary/forestSummarizer.spec.ts
```diff
@@ -164,29 +164,29 @@ function validateHandlesInForestSummary(
  * Validates that the handle path exists in `summaryTree`.
  */
 function validateHandlePathExists(handle: string, summaryTree: ISummaryTree) {
-	/**
-	 * The handle path is split by "/" into pathParts where the first element should exist in the root
-	 * of the summary tree, the second element in the first element's subtree, and so on.
-	 */
-	const pathParts = handle.split("/").slice(1);
```
CraigMacomber, line 171: if there was part of the path before the first slash, this silently ignored it, which seems wrong, so I assert instead.

agarwal-navin:
The title and description of this PR are not meaningful. I would suggest moving the content from some of the comments to the description / title.

agarwal-navin (approved):
LGTM. As per my other comment, you should update the title and description.

PR #26471, opened 2026-02-18
Don't reuse referenceId

CraigMacomber:
## Description

Attempt to simplify and document how referenceIds are allocated and scoped.

## Reviewer Guidance

I know this fails a test, but I'd like to know if that's the test being overly specific about testing implementation details, or a caught real bug.

Also I'd like to know if there is some reason the referenceId's work the way they do instead of the way I suggest which seems simpler.

### DIFF
--- a/packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
+++ b/packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
@@ -77,8 +77,11 @@ interface ChunkSummaryProperties {
 	/**
 	 * The reference ID of the chunk which uniquely identifies it under its parent's summary tree.
 	 * The summary for this chunk will be stored against this reference ID as key in the summary tree.
+	 * @privateRemarks
+	 * TODO: there seems to be no reason to store this as it appears to only be used to store new references to this chunk under this id,
+	 * and there is no reason a new reference to a chunk should use the same ChunkReferenceId as a previous one.
 	 */
-	readonly referenceId: ChunkReferenceId;
+	// readonly referenceId: ChunkReferenceId;
 	/**
 	 * The path for this chunk's summary in the summary tree relative to the forest's summary tree.
 	 * This path is used to generate a summary handle for the chunk if it doesn't change between summaries.
@@ -310,6 +313,7 @@ export class ForestIncrementalSummaryBuilder implements IncrementalEncoderDecode
 				});
 
 				const chunkReferenceIdNumber = Number(chunkReferenceId);
+				// TODO: this should be unneeded
 				this.nextReferenceId = brand(
 					Math.max(this.nextReferenceId, chunkReferenceIdNumber + 1),
 				);
@@ -383,17 +387,20 @@ export class ForestIncrementalSummaryBuilder implements IncrementalEncoderDecode
 				this.latestSummarySequenceNumber,
 				chunk,
 			);
+
+			// Generate a new reference ID for the chunk.
+			// TODO: this should be able to just use the index in the `chunks` array.
+			// If done we can remove this.nextReferenceId.
+			const newReferenceId: ChunkReferenceId = brand(this.nextReferenceId++);
+
 			if (previousChunkProperties !== undefined && !this.trackedSummaryProperties.fullTree) {
 				chunkProperties = previousChunkProperties;
 				this.trackedSummaryProperties.parentSummaryBuilder.addHandle(
-					`${chunkProperties.referenceId}`,
+					`${newReferenceId}`,
 					SummaryType.Tree,
 					`${this.trackedSummaryProperties.latestSummaryBasePath}/${chunkProperties.summaryPath}`,
 				);
 			} else {
-				// Generate a new reference ID for the chunk.
-				const newReferenceId: ChunkReferenceId = brand(this.nextReferenceId++);
-
 				// Add the reference ID of this chunk to the chunk summary path and use the path as the summary path
 				// for the chunk in its summary properties.
 				// This is done before encoding the chunk so that the summary path is updated correctly when encoding
@@ -401,7 +408,6 @@ export class ForestIncrementalSummaryBuilder implements IncrementalEncoderDecode
 				this.trackedSummaryProperties.chunkSummaryPath.push(newReferenceId);
 
 				chunkProperties = {
-					referenceId: newReferenceId,
 					summaryPath: this.trackedSummaryProperties.chunkSummaryPath.join("/"),
 				};
 
@@ -435,7 +441,7 @@ export class ForestIncrementalSummaryBuilder implements IncrementalEncoderDecode
 				chunk,
 				chunkProperties,
 			);
-			chunkReferenceIds.push(chunkProperties.referenceId);
+			chunkReferenceIds.push(newReferenceId);
 		}
 		return chunkReferenceIds;
 	}
--- a/packages/dds/tree/src/test/feature-libraries/forest-summary/incrementalSummaryBuilder.spec.ts
+++ b/packages/dds/tree/src/test/feature-libraries/forest-summary/incrementalSummaryBuilder.spec.ts
@@ -483,12 +483,10 @@ describe("ForestIncrementalSummaryBuilder", () => {
 				builder: summaryTreeBuilder2,
 			});
 			const referenceIds2 = builder.encodeIncrementalField(testCursor, () => mockEncodedChunk);
-			// Should reuse the same reference ID
-			assert.deepEqual(referenceIds1, referenceIds2, "Reference IDs should match");
 
 			// Verify that a handle was added to the summary builder
-			assert.equal(referenceIds1.length, 1);
-			const referenceId1 = referenceIds1[0];
+			assert.equal(referenceIds2.length, 1);
+			const referenceId1 = referenceIds2[0];
 			builder.completeSummary({
 				incrementalSummaryContext: incrementalSummaryContext2,
 				forestSummaryRootContent: mockForestSummaryRootContent,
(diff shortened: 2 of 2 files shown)


packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
```diff
@@ -383,25 +387,27 @@ export class ForestIncrementalSummaryBuilder implements IncrementalEncoderDecode
 				this.latestSummarySequenceNumber,
 				chunk,
 			);
+
+			// Generate a new reference ID for the chunk.
+			// TODO: this should be able to just use the index in the `chunks` array.
```
CraigMacomber, line 392: I'd like to make this change, but I'd like to understand if the other (I think lower impact) change I made is valid first.

packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
```diff
@@ -310,6 +313,7 @@ export class ForestIncrementalSummaryBuilder implements IncrementalEncoderDecode
 				});
 
 				const chunkReferenceIdNumber = Number(chunkReferenceId);
+				// TODO: this should be unneeded
 				this.nextReferenceId = brand(
```
CraigMacomber, line 317: I'd like to remove this as well, though awaiting clarification on if the change I did make is ok or not first.

agarwal-navin:
@CraigMacomber Can you add some reasoning in the description as to why this change is being made? The implications of this change (if any) aside, what gain are you expecting by removing referenceId?

packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
```diff
@@ -383,25 +387,27 @@ export class ForestIncrementalSummaryBuilder implements IncrementalEncoderDecode
...
+			// Generate a new reference ID for the chunk.
+			// TODO: this should be able to just use the index in the `chunks` array.
+			// If done we can remove this.nextReferenceId.
+			const newReferenceId: ChunkReferenceId = brand(this.nextReferenceId++);
+
 			if (previousChunkProperties !== undefined && !this.trackedSummaryProperties.fullTree) {
 				chunkProperties = previousChunkProperties;
 				this.trackedSummaryProperties.parentSummaryBuilder.addHandle(
-					`${chunkProperties.referenceId}`,
+					`${newReferenceId}`,
```
agarwal-navin, line 399: The summary handle points to the location of this chunk's summary tree in the previous summary. That location is identified by the chunk's reference id. If a new reference id is used in the handle, that won't exist in the previous summary, right?

agarwal-navin: Do any tests fail with this change? I would expect tests to fail where a chunk doesn't change.

CraigMacomber:
> @CraigMacomber Can you add some reasoning in the description as to why this change is being made? The implications of this change (if any) aside, what gain are you expecting by removing referenceId?

As noted in the review guidance section. may min interest here is:

> I'd like to know if there is some reason the referenceId's work the way they do instead of the way I suggest which seems simpler.

I'm trying to understand why we have complexity which seems unnecessary. I'd like to either correct my understanding of the requirements (and improve the documentation to make that more discoverable) OR simplify the code so that the referenceIds care more locally scoped, avoids the centralized mutable next id counter, avoids adjusting ids based on if they are reusing content or not, and avoids having to trach the max id encountered when reading a summary. If possible, that simplification would reduce coupling (maintaining and understanding easier), reduce the code needed (lowering bundle size and reducing maintenance costs) and (less importantly) very slightly reduce summary size by using smaller referenceIds.

This is a draft because I don't understand this code, and looking to learn why it does what seems like needless things. Removing some of these apparently needless things breaks tests, which is concerning so I wanted to ask (from the PR description reviewer guidance):

> I know this fails a test, but I'd like to know if that's the test being overly specific about testing implementation details, or a caught real bug.

packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
CraigMacomber: As noted in the reviewer guidance for this PR, yes a test is broken. But I want to know: 
> I know this fails a test, but I'd like to know if that's the test being overly specific about testing implementation details, or a caught real bug.

To me, it seems like there should be no reason that when creating a new SummaryObject, that the key a given handle is stored under needs to be the same as it was stored under in the summary which create the SummaryObject that handle refers to. The fact that this code goes out of its way to reuse the same key is surprising and I'm trying to determine if this is required behavior or not.

I don't get why each SummaryObject we create does not simple use referenceIds "1", "2", "3", etc in order. That's what I originally envisioned, and IIRC you implied that was more difficult for some reason, but looking at the code it seems like it would be simpler to implement it that way, but maybe it breaks something?

agarwal-navin: The reference Id doesn't need to be reused. You can create a new reference Id for every summary object in every summary. I don't think that should break anything in theory. I would recommend adding end-to-end tests where we use real service to ensure that summary handles work as expected.

To reply to your other comment (https://github.com/microsoft/FluidFramework/pull/26471#issuecomment-3946818007), your suggestions would slightly reduce complexity but IMO, the current logic is not unncessarily complex. It reuses ids which is a nice optimization in cases where chunks don't change across summaries as it keeps the reference ids small. Summaries happen frequently and if there are a lot of incremental chunks, the reference id can become quite big. Again, like you mentioned, this is not super important.

agarwal-navin: Just ensure that the `chunkProperties.summaryPath` is correctly updated in `this.chunkTrackingPropertiesMap` when creating a new reference id.

CraigMacomber: >Just ensure that the chunkProperties.summaryPath is correctly updated in this.chunkTrackingPropertiesMap when creating a new reference id.

That's done at the bottom of this function: https://github.com/microsoft/FluidFramework/pull/26471/files/fab7690a3424852d27d10fe8de2e5e02c12f7c96#diff-e7186f997f7e1f831543a53f4e1d69c95b2b558cbeee85f7b7f91a3c71b888b6R444 so I think it should be correct.

> IMO, the current logic is not unncessarily complex. It reuses ids which is a nice optimization in cases where chunks don't change across summaries as it keeps the reference ids small. Summaries happen frequently and if there are a lot of incremental chunks, the reference id can become quite big. Again, like you mentioned, this is not super important.

If simpler code can result in even smaller ids, I think its unnecessary complex, and a counter productive optimization. If a given summary tree could has 6 children to reference, using ids 0-5 seems like the simple choice and the smaller option instead of using what ever number it had last. This particular PR only removes some of the logic, and thus does not get to that point, but it adds the TODOs to make the follow-up changes which could get there.

CraigMacomber: Ah, I forgot that encodeIncrementalField can get called multiple times within a single field batch. That makes a few more changes needed to do my intended simplification, but I still think its cleaner. I'll give it a try.

agarwal-navin: > If simpler code can result in even smaller ids, I think its unnecessary complex, and a counter productive optimization. If a given summary tree could has 6 children to reference, using ids 0-5 seems like the simple choice and the smaller option instead of using what ever number it had last. This particular PR only removes some of the logic, and thus does not get to that point, but it adds the TODOs to make the follow-up changes which could get there.

This PR will not result in smaller ids (as you mentioned already) always. In the scenario you mentioned, the ids will keep on increasing with every summary irrespective of if something changed or not. If your plan is to eventually get to smaller ids, then it looks fine to me.
I disagree that it's a counter productive optimization. There are more things to consider than simply smaller ids or simpler code. For instance, the property that a chunk that doesn't change always has the same reference id makes it easier for me personally to reason about the system and helps in debugging. When looking at multiple summaries (which I have had to do when debugging livesite issues at times), it makes it simpler to debug and correlate sub-trees. If the ids changes and / or are re-used, it will make it harder to debug.

It's essentially a trade-off and I am fine with the approach you suggested as long we document what the summary will look like.

Also, if I understand correctly, in the final state (after TODOs are done), in every summary, we will essentially start from 0, 1, 2 and so on for reference Ids, correct?

agarwal-navin: > That's done at the bottom of this function: https://github.com/microsoft/FluidFramework/pull/26471/files/fab7690a3424852d27d10fe8de2e5e02c12f7c96#diff-e7186f997f7e1f831543a53f4e1d69c95b2b558cbeee85f7b7f91a3c71b888b6R444 so I think it should be correct.

I am not sure that is correct. In the `if (previousChunkProperties !== undefined && !this.trackedSummaryProperties.fullTree)` block, the `chunkProperties`'s summaryPath needs to be updated. It still contains the old reference id but that is now updated.
Consider this scenarios:
- Summary 1 - A chunk has referenced Id `0`. The summary tree looks like this:
  ```
   Forest
  ├── contents
  ├── 0
      ├── /.../Forest/0
  ```
- Summary 2 - The chunk doesn't change. It's reference id is now `1`. The summary handle value is `/.../Forest/0`. The summary tree looks like this:
  ```
  Forest
  ├── contents
  ├── 1
      ├── /.../Forest/0
  ```
- Summary 3 - The chunk doesn't change. It's reference id is now `2`. The summary handle value should be `/.../Forest/1` not `/../Forest/0`. The summary tree should look like this:
  ```
  Forest
  ├── contents
  ├── 2
      ├── /.../Forest/1
  ```

PR #26480, opened 2026-02-18
build: cleanup ci:build definitions

jason-ha:
Use fluidBuild.config definition of `ci:build` for common packages:
```json
		"ci:build": {
			dependsOn: [
				"compile",
				"lint",
				"ci:build:api-reports",
				"ci:build:docs",
				"build:manifest",
				"build:readme",
			],
```
to ensure that `ci:build:api-reports` and `ci:build:docs` are run.

Remove `ci:build` from packages that had been relocate under client workspace and no longer need the entry (that was otherwise similarly light).

### DIFF
--- a/common/lib/common-utils/package.json
+++ b/common/lib/common-utils/package.json
@@ -31,7 +31,7 @@
 		"build:test:types": "tsc --project ./src/test/types/tsconfig.json",
 		"bump-version": "npm version minor --no-push --no-git-tag-version",
 		"check:release-tags": "api-extractor run --local --config ./api-extractor-lint.json",
-		"ci:build": "npm run build:compile",
+		"ci:build": "fluid-build . --task ci:build",
 		"ci:build:docs": "api-extractor run --typescript-compiler-folder ./node_modules/typescript && copyfiles -u 1 \"./_api-extractor-temp/doc-models/*\" ../../../_api-extractor-temp/",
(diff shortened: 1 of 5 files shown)


tylerbutler (approved):

alexvy86 (approved):
Seems fine as far as I can tell. The removals in particular; the changes in the `common/*` packages I'm not as confident about, but they seem reasonable.

ChumpChief (approved):
Also makes sense!  Should one of the policy checks be enforcing that these scripts are what we expect (maybe as a followup)?

jason-ha:
> Seems fine as far as I can tell. The removals in particular; the changes in the `common/*` packages I'm not as confident about, but they seem reasonable.

common is the impetus for the changes. Recently some code changes went in that changed API but the reports were not updated - not checked by CI. I confirmed that the tasks expected are now run under CI (and would have caught the change).

jason-ha:
> Also makes sense! Should one of the policy checks be enforcing that these scripts are what we expect (maybe as a followup)?

Maybe. There are very few of these; so, ROI may not be there. I do think more of the workspaces should be using ci:build. I also found that api-markdown-documenter has an API lint task that is completely invalid. (The command line references a config file that doesn't exist.) It doesn't have a `ci:build` script. It will when I'm done.

PR #26485, opened 2026-02-19
refactor(register-collection): hide implementation class from legacy API

tylerbutler:
## Summary

- Mark the `ConsensusRegisterCollection` implementation class as `@internal`, removing it from the legacy (`/legacy`) API surface
- The `IConsensusRegisterCollection` interface, `ConsensusRegisterCollection` singleton const, and `ConsensusRegisterCollectionFactory` remain at `@legacy @beta` -- no changes for consumers using the public API
- The legacy API report is updated to reflect the removal of `ConsensusRegisterCollectionClass` from the legacy surface

## Why

This is part of the custom DDS removal effort. Implementation classes for DDSes should not be on the legacy API surface since they expose internal details that are not needed by consumers. Consumers should use the `IConsensusRegisterCollection` interface and the `ConsensusRegisterCollection` singleton/factory to create instances.

## What changed

- `packages/dds/register-collection/src/consensusRegisterCollection.ts`: Changed the TSDoc release tag on the `ConsensusRegisterCollection` class from `@legacy @beta` to `@internal`
- `packages/dds/register-collection/api-report/register-collection.legacy.beta.api.md`: Regenerated -- the `ConsensusRegisterCollectionClass` type alias is removed from the legacy API report

## Test plan

- [x] Package builds successfully
- [x] API reports regenerated correctly
- [x] No type test changes needed (class was already not directly referenced in type tests)


ChumpChief (approved):

tylerbutler:
I won't merge this until 2.100 since it's a breaking change. The deprecation is going in 2.90.

PR #26487, opened 2026-02-19
Optimize shared tree text prototype, and fix formatRange API

CraigMacomber:
## Description

Profiling of the text editor turned up a few major and easy performance improvements.

Now production builds of the text editor avoid making full destructive replaces of all content when editing plain text, and use much faster APIs to get the current content and length.

With this, plain text editing goes from lagging horribly to faster than I type for the largish strings I tried.

Formatted text is now optimized to save the ~45% of the time that was spent in fullString by making full string way faster (so much faster by profiling the app, its hard to measure anymore, but likely over 20x faster).

This also fixes a bug in the removeRange API where the second parameter was called length but was uses as the end. Its name has been corrected. The formatRange APi has also been updated to align with the correct removeRange.

## Reviewer Guidance


examples/data-objects/text-editor/src/plain/quillView.tsx
```diff
@@ -60,20 +61,8 @@ const TextEditorView = withMemoizedTreeObservations(({ root }: { root: TextAsTre
...
-
-					// Clear existing content and insert new text
-					const length = [...root.characters()].length;
-					if (length > 0) {
-						root.removeRange(0, length);
-					}
-					if (text.length > 0) {
-						root.insertAt(0, text);
-					}
+					const newText = quill.getText();
```
yann-achard-MS, line 64: Doesn't Quill generate deltas when edits are performed? Converting a Quill Delta to a set of SharedTree changes seems like it ought to be much more efficient.

Using `syncTextToTree` seems like a fine steppingstone though.

yann-achard-MS (approved):

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -43,16 +44,51 @@ class TextNode
...
+		return result;
+	}
+
+	public characterCount(): number {
+		return this.content.length;
+	}
+
 	public fullString(): string {
+		const result = this.content.fullString();
+		debugAssert(
```
CraigMacomber, line 73: This design has two problems related to how reading from the tree had observation tracking side effects:
1. The optimized version does not track observations. We should have a util which borrows a cursor and tracks observation of the whole subtree to address that.
2. This debugAssert has observation tracking side-effects. Thus we should either suppress those, or not use this pattern with tree reads. We want tests which do full validation to be able to catch bugs like 1 above, and not have that masked by the validation side-effects.

examples/data-objects/text-editor/src/plain/quillView.tsx
CraigMacomber: I think so, and I think the formatted version uses it. TextArea probably needs this logic, but this version should likely share logic with the formatted version instead of text area.

packages/dds/tree/src/text/textDomainFormatted.ts
CraigMacomber: Add deep observation tracking for cursor helper, and lots of test coverage. Should be good enough for now.

brrichards (approved):

PR #26490, opened 2026-02-19
build-tools: consolidate all packages into single @fluidframework/build-tools package

alexvy86:
## Summary

Consolidates all 5 packages under `build-tools/packages/` into a single package at `build-tools/packages/build-tools/`.

### Packages merged
- `@fluidframework/build-tools` (core, CJS)
- `@fluid-tools/build-cli` (flub CLI, ESM)
- `@fluid-tools/build-infrastructure` (buildProject CLI, ESM)
- `@fluid-tools/version-tools` (fluv CLI, CJS)
- `@fluidframework/bundle-size-tools` (library, CJS)

### Source organization
Source is organized as sub-projects under `src/`:
- `src/core/` — original build-tools
- `src/build-cli/` — flub CLI
- `src/build-infrastructure/` — buildProject CLI
- `src/version-tools/` — fluv CLI
- `src/bundle-size-tools/` — library

### Key changes
- **TypeScript Project References** with solution-level tsconfig for correct build ordering
- **Mixed CJS/ESM** preserved via marker `package.json` files in ESM sub-projects
- **Subpath exports**: `.`, `./build-cli`, `./build-cli/policy`, `./build-infrastructure`, `./version-tools`, `./bundle-size-tools`
- **4 stub packages** for backward compatibility (re-export from consolidated package)
- **Merged eslint config** with scoped rules per sub-project
- **Updated biome exclusions** for relocated test data

### Verification
- All **824 tests passing** (137 core + 237 version-tools + 398 build-cli + 52 build-infrastructure)
- Full workspace build succeeds (`pnpm build` from `build-tools/`)
- All subpath exports verified working
- Stub package imports verified working


tylerbutler:
FWIW, I really dont like this approach. The code quality varies massively between the packages, and this sort of re-org makes blame impossible to follow in my experience. It's also not clear to me what problem this solves.

alexvy86:
> FWIW, I really dont like this approach. The code quality varies massively between the packages, and this sort of re-org makes blame impossible to follow in my experience. It's also not clear to me what problem this solves.

That's fair, for me this is just an experiment in pushing Claude based on Craig at some point suggesting this avenue as a potential way to streamline build-tools. I don't recall what specific problems it would address based on that conversation (which I'd have to find), but my personal gut feeling is that a single package is probably a more desirable end goal that keeping them separate, so wanted to get an idea of what that change would look like. But seeing it at this point, I don't intend to merge it.

PR #26491, opened 2026-02-19
Update README.md for codespace instructions

brrichards:
# Summary

Updated the `README.md` in `examples/data-objects/text-editor` to include needing Tinylicious in another terminal in the "Getting Started" instructions.  Changed `EXAMPLE_APP_README_HEADER:usesTinylicious=FALSE` to `EXAMPLE_APP_README_HEADER:usesTinylicious=TRUE`

Also updated auto-generated `README.md` tinylicious section to include instructions on how to use tinylicious in codespaces

### DIFF
--- a/tools/markdown-magic/src/transforms/exampleGettingStartedTransform.cjs
+++ b/tools/markdown-magic/src/transforms/exampleGettingStartedTransform.cjs
@@ -49,6 +49,10 @@ const generateExampleGettingStartedSection = (
 	);
 
 	sectionBody.push(
+		`1. If using codespaces, set tinylicious (port 7070) visibility to "public". "Private to Organization" will not work. Check your IDE's terminal or ports panel to find available ports.`,
+	);
+
+	sectionBody.push(
 		`1. If you want to run the app against SharePoint, follow the instructions in [webpack-fluid-loader](https://github.com/microsoft/FluidFramework/blob/main/examples/utils/webpack-fluid-loader/README.md#sharepoint) to get auth credentials. Then run \`pnpm start:spo\` or \`pnpm start:spo-df\` and open <http://localhost:8080> like above.`,
 	);
 
--- a/tools/markdown-magic/src/transforms/exampleGettingStartedTransform.cjs
+++ b/tools/markdown-magic/src/transforms/exampleGettingStartedTransform.cjs
@@ -40,16 +40,16 @@ const generateExampleGettingStartedSection = (
 
 	if (includeTinyliciousStep) {
 		sectionBody.push(
-			`1. In a separate terminal, start a Tinylicious server by following the instructions in [Tinylicious](https://github.com/microsoft/FluidFramework/tree/main/server/routerlicious/packages/tinylicious).`,
+			`1. In a separate terminal, start a Tinylicious server by running pnpm Tinylicious in this directory.`,
+		);
+
+		sectionBody.push(
+			`1. If using codespaces in a browser, set tinylicious (port 7070) visibility to "public". "Private to Organization" will not work. Check your IDE's terminal or ports panel to find available ports.`,
 		);
 	}
 
 	sectionBody.push(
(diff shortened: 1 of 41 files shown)


tools/markdown-magic/src/transforms/exampleGettingStartedTransform.cjs
```diff
@@ -48,6 +48,10 @@ const generateExampleGettingStartedSection = (
 		`1. Run \`pnpm start\` from this directory and open <http://localhost:8080> in a web browser to see the app running.`,
 	);
 
+	sectionBody.push(
```
CraigMacomber: Shouldn't this be in the `if (includeTinyliciousStep) {` section above?

Also maybe its out of scope for this PR, but in my opinion, the instructions in that section are not great. It directs you to effectively build a local copy of the server which is annoying as it has native deps. Since our examples that use it take a dev dep, I think this would be better

> 1. In a separate terminal, start a [Tinylicious](https://github.com/microsoft/FluidFramework/tree/main/server/routerlicious/packages/tinylicious) server by running `pnpm tinylicious` in this directory.

tools/markdown-magic/src/transforms/exampleGettingStartedTransform.cjs
```diff
@@ -48,6 +48,10 @@ const generateExampleGettingStartedSection = (
...
+		`1. If using codespaces, set tinylicious (port 7070) visibility to "public". "Private to Organization" will not work. Check your IDE's terminal or ports panel to find available ports.`,
```
CraigMacomber: This step is not needed for me. Private works fine. This requirement might be specific to using codepsaces via the browser (I'm connecting to them from a local vs-code install, which has some benefits like better plugin support)

tools/markdown-magic/src/transforms/exampleGettingStartedTransform.cjs
brrichards: You're right, I missed that closing bracket for `if (includeTinyliciousStep) {`. I'll also update the tinylicious server instructions

tools/markdown-magic/src/transforms/exampleGettingStartedTransform.cjs
brrichards: I'll specify browser only

tools/markdown-magic/src/transforms/exampleGettingStartedTransform.cjs
```diff
@@ -40,18 +40,18 @@ const generateExampleGettingStartedSection = (
 
 	if (includeTinyliciousStep) {
 		sectionBody.push(
-			`1. In a separate terminal, start a Tinylicious server by following the instructions in [Tinylicious](https://github.com/microsoft/FluidFramework/tree/main/server/routerlicious/packages/tinylicious).`,
+			`1. In a separate terminal, start a Tinylicious server by running pnpm Tinylicious in this directory.`,
```
CraigMacomber:
```suggestion
			`1. In a separate terminal, start a Tinylicious server by running \`pnpm Tinylicious\` in this directory.`,
```

tools/markdown-magic/src/transforms/exampleGettingStartedTransform.cjs
```diff
@@ -40,18 +40,18 @@ const generateExampleGettingStartedSection = (
...
+		);
+
+		sectionBody.push(
+			`1. If using codespaces in a browser, set tinylicious (port 7070) visibility to "public". "Private to Organization" will not work. Check your IDE's terminal or ports panel to find available ports.`,
```
CraigMacomber: If using codespaces in a browser, I think we can assume they are using VS-Code.

```suggestion
			`1. If using codespaces in a browser, set tinylicious (port 7070) visibility to "public". "Private to Organization" will not work. See [sharing a port](https://docs.github.com/en/codespaces/developing-in-a-codespace/forwarding-ports-in-your-codespace#sharing-a-port) for how to do this.`,
```

tylerbutler (approved):

PR #26503, opened 2026-02-19
Pin "@rushstack/node-core-library": "5.19.1"

markfields:
## Description

The 5.20.0 release of Client RG: Pin `@rushstack/node-core-library` changed how their build outputs are structured which broke our internal build pipelines.

## Reviewer Guidance

This branch so far only updates the Client ReleaseGroup since that's the most impactful.  Others are coming.


markfields:
THis won't work, closing

PR #26504, opened 2026-02-19
ci: Use pnpm with an override to install build-tools in the publish stage

alexvy86:
## Description

A release of a transitive dependency of build-tools (`@rushstack/node-core-library`) is broken and that is causing the installation of build-tools in the publishing stage of our pipeline to fail.

This changes the way we do that install in the pipeline stage so we can leverage pnpm overrides to pin to a working version of the dependency.

## Reviewer Guidance

[AB#59774](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/59774)

### DIFF
--- a/tools/pipelines/templates/include-publish-npm-package-deployment.yml
+++ b/tools/pipelines/templates/include-publish-npm-package-deployment.yml
@@ -58,35 +58,27 @@ jobs:
           steps:
           - template: /tools/pipelines/templates/include-use-node-version.yml@self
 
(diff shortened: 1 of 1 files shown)


alexvy86:
`@rushstack/node-core-library` released a patch fixing the issue so we don't technically need this anymore. It might be useful to have an escape hatch along these lines but for now we won't merge this.

PR #26507, opened 2026-02-20
deprecate(experimental-tree): deprecate direct SharedTree class usage

tylerbutler:
## Summary

- Mark the `SharedTree` class in `@fluid-experimental/tree` as `@deprecated`, directing users to the new `ISharedTree` interface
- Add `ISharedTree` interface extracted from the class as the migration target
- Fix `ISharedTreeEvents` event name mismatch (`'appliedSequencedEdit'` → `'sequencedEditApplied'` to match the actual emitted event)

This is a **non-breaking** change — no APIs are removed, only deprecated. It prepares for a future removal of the class from the public API surface as part of the custom DDS removal effort.

## Test plan

- [x] Build passes (`pnpm build @fluid-experimental/tree`)
- [x] CI passes


examples/apps/tree-comparison/src/model/legacyTreeInventoryList.ts
```diff
@@ -185,12 +193,11 @@ export class LegacyTreeInventoryList extends DataObject implements IInventoryLis
...
 		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-		this._tree = await this.root
-			.get<IFluidHandle<LegacySharedTree>>(legacySharedTreeKey)!
-			.get();
+		this._tree = await this.root.get<IFluidHandle<ISharedTree>>(legacySharedTreeKey)!.get();
 
 		// We must use a checkout in order to get "viewChange" events - it doesn't change any of the rest of our usage though.
-		const checkout = new EagerCheckout(this._tree);
+		// eslint-disable-next-line import-x/no-deprecated
+		const checkout = new EagerCheckout(this._tree as LegacySharedTree);
```
tylerbutler, line 200: Can we avoid the cast? If not, can we use `satisfies` instead?

examples/apps/tree-comparison/src/model/legacyTreeInventoryList.ts
```diff
@@ -3,10 +3,17 @@
...
+	BuildNode,
+	ISharedTree,
+	TraitLabel,
+	TreeView,
+	TreeViewNode,
+} from "@fluid-experimental/tree";
 import {
 	Change,
 	EagerCheckout,
+	// eslint-disable-next-line import-x/no-deprecated
```
tylerbutler, line 16: Why do we still need this import?

examples/apps/tree-comparison/src/model/legacyTreeInventoryList.ts
```diff
@@ -301,6 +308,7 @@ export class LegacyTreeInventoryList extends DataObject implements IInventoryLis
 export const LegacyTreeInventoryListFactory = new DataObjectFactory<LegacyTreeInventoryList>(
 	"legacy-tree-inventory-list",
 	LegacyTreeInventoryList,
+	// eslint-disable-next-line import-x/no-deprecated
```
tylerbutler, line 311: Can't this use the new patterns?

examples/benchmarks/bubblebench/experimental-tree/src/main.tsx
```diff
@@ -83,5 +86,6 @@ export class Bubblebench extends DataObject {
 export const BubblebenchInstantiationFactory = new DataObjectFactory({
 	type: Bubblebench.Name,
 	ctor: Bubblebench,
+	// eslint-disable-next-line import-x/no-deprecated
 	sharedObjects: [SharedTree.getFactory(WriteFormat.v0_1_1)],
```
tylerbutler, line 90: Isn't there a different pattern for this?

examples/utils/bundle-size-tests/src/experimentalSharedTree.ts
```diff
@@ -3,6 +3,7 @@
  * Licensed under the MIT License.
  */
 
+/* eslint-disable import-x/no-deprecated -- This bundle-size test intentionally uses the deprecated SharedTree value. */
```
tylerbutler, line 6: Shouldn't we use the new patterns here?

examples/version-migration/tree-shim/src/model/legacyTreeInventoryListController.ts
```diff
@@ -126,10 +128,11 @@ export class LegacyTreeInventoryListController extends EventEmitter implements I
 
 	private readonly _inventoryItems = new Map<string, LegacyTreeInventoryItem>();
 
-	public constructor(private readonly _tree: LegacySharedTree) {
+	public constructor(private readonly _tree: ISharedTree) {
 		super();
 		// We must use a checkout in order to get "viewChange" events - it doesn't change any of the rest of our usage though.
-		const checkout = new EagerCheckout(this._tree);
+		// eslint-disable-next-line import-x/no-deprecated
+		const checkout = new EagerCheckout(this._tree as LegacySharedTree);
```
tylerbutler, line 135: Seems to be the wrong type, no? Shouldn't it be ISharedTree? Is there some reason that EagerCheckout can't take an ISharedTree?

examples/benchmarks/bubblebench/experimental-tree/src/main.tsx
tylerbutler: I looked into this -- DataObjectFactory's sharedObjects parameter only accepts IChannelFactory[], so .getFactory() is required. This example uses SharedTree.getFactory(WriteFormat.v0_1_1) from @fluid-experimental/tree with a specific write format. Is there a different pattern you have in mind, or should this example be reworked more fundamentally (e.g., migrated away from @fluid-experimental/tree)?

examples/utils/bundle-size-tests/src/experimentalSharedTree.ts
tylerbutler: This bundle-size test intentionally measures the deprecated SharedTree value from @fluid-experimental/tree. Other DDS bundle-size tests use the .getFactory() pattern, but this test's purpose is to track the deprecated export's size specifically. Should it be removed, or updated to use .getFactory() instead?

examples/version-migration/tree-shim/src/model/legacyTreeInventoryListController.ts
tylerbutler: Checked -- EagerCheckout's constructor currently takes the concrete SharedTree class, but the implementation only uses public interface members. I'll update EagerCheckout to accept ISharedTree, which will also eliminate the cast and the LegacySharedTree import in the other two comments on this file.

examples/apps/tree-comparison/src/model/legacyTreeInventoryList.ts
tylerbutler: Changing EagerCheckout to accept ISharedTree would cascade into changing the Checkout base class (which stores `public readonly tree: SharedTree`), and since ISharedObject doesn't expose `emit`, it would require worse-looking casts elsewhere (e.g. `this.tree as unknown as EventEmitterWithErrorHandling<ISharedTreeEvents>` for error forwarding, plus test code casting back to SharedTree for `listenerCount`). Net result is more casts, not fewer.

examples/apps/tree-comparison/src/model/legacyTreeInventoryList.ts
tylerbutler: Same as above — changing EagerCheckout to accept ISharedTree would eliminate this cast, but it cascades into the Checkout base class and introduces worse casts elsewhere (error forwarding via `emit`, test assertions via `listenerCount`). Keeping the cast here is the lesser evil. `satisfies` can't work here since ISharedTree doesn't structurally satisfy the concrete SharedTree class type — it's a narrowing assertion, not a widening check.

examples/apps/tree-comparison/src/model/legacyTreeInventoryList.ts
tylerbutler: DataObjectFactory's `sharedObjects` parameter only accepts `IChannelFactory[]`, so `.getFactory()` is required. The only alternative would be switching to `TreeDataObjectFactory` (which auto-registers SharedTree), but that's a larger refactor of this legacy example.

examples/version-migration/tree-shim/src/model/legacyTreeInventoryListController.ts
tylerbutler: Tried this — EagerCheckout's constructor takes the concrete SharedTree class, and changing it cascades into the Checkout base class which uses `this.tree.emit('error', error)` for error forwarding. Since ISharedObject (via IEventProvider) doesn't expose `emit`, this requires `this.tree as unknown as EventEmitterWithErrorHandling<ISharedTreeEvents>` plus test code casting back to SharedTree for `listenerCount`. Net result is more casts pushed into worse places, so keeping the cast here in the example is the lesser evil.

tylerbutler:
Replaced by two PRs that split non-breaking and breaking changes:

- **Non-breaking**: #26721 — adds ISharedTree interface, deprecates SharedTree class, migrates examples
- **Breaking**: #26722 — decouples ISharedTree from ISharedObject, fixes event name, removes DDS types from alpha API

PR #26508, opened 2026-02-20
generate server release reports

kekachmar:
## Description

- Adds a manifest upload for server packages (historian, gitrest, routerlicious).
- Generates prerelease manifests for the server, and publishes exact version mappings for the selected release group.
- Adds a release report flag to build tools to use the current checked-out version when tag history does not yet contain that version.

## Reviewer Guidance
- Creates a shared file for the upload steps. `include-upload-release-reports.yml`.
- Server files are uploaded to blob storage with the path `manifest-files/server/prerelease/<filename>.json`

### DIFF
--- a/tools/pipelines/templates/include-publish-npm-package.yml
+++ b/tools/pipelines/templates/include-publish-npm-package.yml
@@ -18,6 +18,22 @@ parameters:
 - name: buildDirectory
   type: string
 
+- name: buildToolsVersionToInstall
+  type: string
+  default: repo
+
+- name: serverManifestPathSegment
+  type: string
+  default: prerelease
+  values:
+  - prerelease
+  - release
+
+# The path to the pnpm store.
+- name: pnpmStorePath
+  type: string
+  default: $(Pipeline.Workspace)/.pnpm-store
+
 stages:
 
 # This stage only runs for test-branch builds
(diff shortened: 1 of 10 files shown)


tools/pipelines/templates/include-publish-npm-package.yml
```diff
@@ -115,6 +131,22 @@ stages:
...
-      parameters:
-        buildDirectory: '${{ parameters.buildDirectory }}'
+    - ${{ if or(eq(parameters.tagName, 'server'), eq(parameters.tagName, 'historian'), eq(parameters.tagName, 'gitrest')) }}:
+      - template: /tools/pipelines/templates/upload-server-manifest.yml@self
+        parameters:
+          buildDirectory: '${{ parameters.buildDirectory }}'
+          blobPathPrefix: ${{ format('server/{0}', parameters.serverManifestPathSegment) }}
+          serverManifestPathSegment: ${{ parameters.serverManifestPathSegment }}
+          ${{ if eq(parameters.tagName, 'server') }}:
+            releaseGroup: 'server'
```
Abe27342, line 141: should be able to replace the if/else here with

```
releaseGroup: '${{ parameters.tagName }};
releaseReportCategory: '${{ parameters.tagNam }}-'
```

tools/pipelines/templates/upload-server-manifest.yml
```diff
@@ -0,0 +1,144 @@
...
+        displayName: Generate latest prerelease reports
+        inputs:
+          targetType: 'inline'
+          workingDirectory: ${{ parameters.buildDirectory }}
+          script: |
+            set -eu -o pipefail
+            mkdir generate_release_reports
+            mkdir upload_release_reports
+            flub release report -g ${{ parameters.releaseGroup }} --useCurrentVersion -o generate_release_reports --baseFileName manifest
+            REPORT_VERSION="$(version)" RELEASE_GROUP="${{ parameters.releaseGroup }}" BUILD_SOURCEBRANCH='$(Build.SourceBranch)' node -e '
```
Abe27342, line 47: the extra node script here and below seems unnecessary. What's the intent? Pipeline build number is available in build variables. If we need more information in the context of the pipeline, we should have the release report generation flub command be writing to ADO pipeline environment variables (similar to what the version command bump does).

PR #26511, opened 2026-02-20
Install fluid-build to a local pnpm project in publish jobs

Abe27342:
## Description

A recent bad publish of the \@rushstack libraries (see https://github.com/microsoft/rushstack/issues/5644) made us realize that problems caused by new releases of the (transitive) dependencies of build-tools can block us from releasing as well. This is because our publish step currently installs Fluid's build-tools (to run the publish command) from a fresh context without a lock file.

This PR implements a relatively self-contained mitigation mechanism: it moves the global installation of fluid-build via npm to instead be a locally spun up project installation via `pnpm` that is then added to the PATH. This allows leveraging pnpm's overrides mechanism, which is how we would address similar sorts of dependency issues.

Test run can be found [here](https://dev.azure.com/fluidframework/internal/_build/results?buildId=378885&view=results)

### DIFF
--- a/tools/pipelines/templates/include-publish-npm-package-deployment.yml
+++ b/tools/pipelines/templates/include-publish-npm-package-deployment.yml
@@ -121,32 +121,17 @@ jobs:
                 ls -la
                 pnpm --version
 
-                # Set up pnpm global bin directory explicitly
-                export PNPM_HOME="$HOME/.local/share/pnpm"
-                mkdir -p "$PNPM_HOME"
-                export PATH="$PNPM_HOME:$PATH"
-                pnpm config set global-bin-dir "$PNPM_HOME"
-
                 echo "Installing build tools from tarballs (uses overrides from package.json)"
                 pnpm install ./*.tgz
 
-                echo "Linking build tools globally"
-                for tgz in *.tgz; do
-                  pkg_name=$(tar -xzf "$tgz" --to-stdout package/package.json | jq -r '.name')
-                  # Check if the package has a bin field (i.e., provides CLI commands)
-                  has_bin=$(tar -xzf "$tgz" --to-stdout package/package.json | jq -e '.bin' > /dev/null 2>&1 && echo "yes" || echo "no")
-                  if [ "$has_bin" = "yes" ]; then
-                    echo "Linking $pkg_name globally"
-                    pnpm link --global "node_modules/$pkg_name"
-                  fi
-                done
+                # Add the local node_modules/.bin to PATH for subsequent pipeline steps
+                BIN_DIR="$(pwd)/node_modules/.bin"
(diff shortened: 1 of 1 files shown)


tools/pipelines/templates/include-publish-npm-package-deployment.yml
```diff
@@ -58,6 +58,55 @@ jobs:
...
+
+          # We can use this step to add overrides that might be necessary to allow for a correct
+          # installation of build-tools from the tarball artifacts produced by the build stage.
+          # Since this is a lockfile-less install (ideally we should figure out how to change that),
+          # overrides is one recourse when a dependency releases a new version that causes build-tools installation to break.
+          # Current overrides:
+          # - Build-tools packages: map all versions to local tarballs so interdependencies resolve correctly.
+          - task: Bash@3
+            name: CreatePackageJsonForOverrides
+            displayName: Create package.json with npm overrides
```
alexvy86: Nit, aligning with other copilot comments.
```suggestion
            displayName: Create package.json with pnpm overrides
```

tools/pipelines/templates/include-publish-npm-package-deployment.yml
```diff
@@ -58,6 +58,55 @@ jobs:
...
+                  tarball_overrides=$(echo "$tarball_overrides" | jq --arg name "$pkg_name" --arg file "file:$tgz" '. + {($name): $file}')
+                done
+
+                # Merge manual overrides with tarball overrides and create package.json
+                jq -n \
+                  --argjson manual "$manual_overrides" \
+                  --argjson tarballs "$tarball_overrides" \
+                  '{
+                    "name": "build-tools-install",
+                    "packageManager": "pnpm@10.18.3+sha512.bbd16e6d7286fd7e01f6b3c0b3c932cda2965c06a908328f74663f10a9aea51f1129eea615134bf992831b009eabe167ecb7008b597f40ff9bc75946aadfb08d",
```
markfields, line 99: Is this something we'll need to update periodically?

alexvy86 (approved):
Looks good to me overall. I think I'd apply all copilot suggestions to make this as defensive as possible.

tools/pipelines/templates/include-publish-npm-package-deployment.yml
alexvy86: Yeah, ideally at the same time we do it for the package.jsons that are committed. Even if it goes out of sync for a bit, I don't think it's a huge deal.

alexvy86 (approved):

Abe27342:
updated test run after taking pr feedback [here](https://dev.azure.com/fluidframework/internal/_build/results?buildId=379527&view=results). Merging now :)

PR #26512, opened 2026-02-20
Improve alpha transaction API

noencke:
In preparation for promoting SharedTree transaction APIs to beta, several improvements have been made:

* `TreeContextAlpha.runTransaction` now accepts `params`, just like `TreeBranchAlpha.runTransaction`.
The params have been documented such that the properties are sensible for both interfaces (even though they don't do very much for unhydrated contexts). This leaves `rollback` as the only API discrepancy between the unhydrated and hydrated transaction APIs. We are unable to implement rollback the traditional way for unhydrated transactions because there is no rebaser. We could make and restore a copy of the tree instead, but that might be expensive. See Open Question below.
* The documentation on all alpha transaction functions has been deduplicated and leverages doc cross-linking instead.
The documentation that remains has been improved overall as well.
* Asynchronous transactions now error if begun when any other transaction is already in progress.
This policy is more conservative than before to allow us more design options going forward.
* UnhydratedTreeContext has been moved to its own file and the Transactor API has been cleaned up now that `isAsync` is not necessary to report errors.

Open Question:

Should we include the `rollback` return parameter in `TreeContextAlpha.runTransaction` so that it can be fully consistent with `TreeBranchAlpha.runTransaction`? This would make the API types simpler, but would mean we'd have to throw an error if you attempt to rollback an unhydrated transaction.


packages/dds/tree/src/shared-tree/schematizingTreeView.ts
```diff
@@ -348,7 +352,8 @@ export class SchematizingSimpleTreeView<
 			| TransactionCallbackStatus<TSuccessValue, TFailureValue>
 			| VoidTransactionCallbackStatus
 			| void,
-		params?: RunTransactionParams,
+		params: RunTransactionParams | undefined,
+		isAsync: boolean,
```
yann-achard-MS: >isAsync [](http://example.com/codeflow?start=2&length=7)

It looks like the `isAsync` parameter is not need (and never was??) #Closed

packages/dds/tree/src/shared-tree-core/transaction.ts
```diff
@@ -46,13 +46,11 @@ export interface Transactor {
...
-	 * @param isAsync - Whether the transaction is asynchronous.
-	 * An error will be thrown if an asynchronous transaction is started while a synchronous transaction is in progress.
 	 *
-	 * @remarks Asynchronous transactions are not supported on the root checkout,
-	 * since it is always kept up-to-date with the latest remote edits and the results of this rebasing (which might invalidate
-	 * the transaction) is not visible to the application author.
-	 * Instead,
+	 * @remarks
+	 * Asynchronous transactions are not supported on the root checkout,
+	 * since it is always kept up-to-date with the latest remote edits and the results of this rebasing (which might invalidate the transaction) is not visible to the application author.
```
yann-achard-MS: I failed to remove this remark in my PR. We do support async transactions on the root checkout. #Closed

packages/dds/tree/src/simple-tree/api/transactionTypes.ts
```diff
@@ -134,17 +133,18 @@ export type TransactionResult =
...
 	 * An optional list of {@link TransactionConstraintAlpha | constraints} that are checked just before the transaction begins.
+	 * @remarks
 	 * If any of the constraints are not met when `runTransaction` is called, an error will be thrown.
-	 * If any of the constraints are not met after the transaction has been ordered by the service, it will be rolled back on
-	 * this client and ignored by all other clients.
+	 * If any of the constraints are not met after the transaction has been ordered by the service, it will be rolled back on this client and ignored by all other clients.
+	 *
+	 * Constraints only apply to the "outermost" transaction.
+	 * If a transaction is started while another transaction is already in progress, then its constraints will be ignored.
 	 */
```
yann-achard-MS, line 139: As things stand, we do support constraints in nested transactions, but we don't do a great job of it because we consider some constraints violated in scenarios where they really should not be (this will be improved by [https://dev.azure.com/fluidframework/internal/_workitems/edit/55763](https://dev.azure.com/fluidframework/internal/_workitems/edit/55763)). That said, we do correctly consider constraints violated when they ought to be, so our handling of constraints on nested transactions is technically correct.

So I think that means we don't need a comment, and we very much do not want a comment that tells people their constraints will be ignored.
 #Closed

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -188,70 +188,55 @@ export interface TreeBranch extends IDisposable {
...
+	 * Note that this is guaranteed by Fluid for any sequence of changes that are submitted synchronously, whether in a transaction or not.
+	 * However, using a transaction has the following additional consequences:
+	 *
+	 * - If {@link Revertible | reverted } (e.g. via an "undo" operation), all the changes in the transaction are reverted together.
+	 * Only the "outermost" transaction commits a change to the synchronized tree state and therefore only the outermost transaction can be reverted.
+	 * If a transaction is started and completed while another transaction is already in progress, then the inner transaction will be reverted together with the outer transaction.
+	 * - The internal data representation of a transaction with many changes is generally smaller and more efficient than that of the changes when separate.
+	 *
+	 * {@link (TreeBeta:interface).on | Change events } will be emitted for each mutation as the transaction is being applied.
 	 */
```
yann-achard-MS, line 213: After reading this whole description, I think someone could be confused because we say "All of the changes in the transaction are applied synchronously and therefore no other changes (either from this client or from a remote client) can be interleaved with those changes." as well as "Change events will be emitted for each mutation as the transaction is being applied". They could appear to be in tension to someone who hasn't internalize how different entities and layers experience the same transaction, which we don't explicitly cover. It seems like we should maybe start there.

I think this PR is a vast improvement though, so I'd be happy to not gate it on such concerns, #Closed

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -188,70 +188,55 @@ export interface TreeBranch extends IDisposable {
...
 	 * @remarks
-	 * If `runTransactionAsync` is invoked on the context of a {@link TreeStatus.InDocument | node in the document }, the transaction will be applied to the {@link TreeBranchAlpha | branch associated with that node}.
-	 * Use {@link TreeContextAlpha.isBranch | isBranch() } to check whether this context is associated with a branch and gain {@link TreeBranchAlpha.(runTransactionAsync:1) | access to more transaction capabilities} if so.
+	 * As with synchronous transactions, all of the changes in an asynchronous transaction are treated as a unit.
+	 * Therefore, no other changes (either from this client or from a remote client) can be interleaved with the transaction changes.
 	 *
-	 * If `runTransactionAsync` is invoked on the context of an {@link Unhydrated | unhydrated } node or removed node,
-	 * it is equivalent to running the `transaction` delegate directly and the transaction will always succeed.
+	 * Unlike with synchronous transactions, it is possible that other changes (either from this client by merging a branch or from a remote client) may be applied to the branch while this transaction is in progress.
```
yann-achard-MS: >from this client by merging a branch [](http://example.com/codeflow?start=84&length=36)

We don't allow merging into a branch with an open transaction. Nor do we allow rebasing a branch with an open transaction onto another branch.

We likely won't ever support the former but we could support the latter if people wanted it. #Closed

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Hmm, yea... the reality is that we simply don't allow customers to do edits in those event callbacks, right? So yes, there is a cut point where their code can come into play during the transaction, but it's not allowed to mutate. Maybe we just emphasize that it is _forbidden_ to do that, rather than what this says - that "therefore, it cannot happen" - which isn't quite true or sensical.

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
noencke: Oops, that was leftover from another approach I was taking. Good eye.

packages/dds/tree/src/shared-tree-core/transaction.ts
noencke: Removed.

packages/dds/tree/src/simple-tree/api/transactionTypes.ts
noencke: Okay, I got rid of it.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: I tweaked it a bit. Probably not perfect, but maybe a bit better for now.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: Removed

packages/dds/tree/src/simple-tree/api/tree.ts
yann-achard-MS: I do think that's an improvement. My concern (which I didn't communicate clearly) was more focused on who gets to see the transaction changes as individual steps as they happen, and who gets to see them as as single atomic change.

yann-achard-MS (approved):
:shipit:

yann-achard-MS (approved):

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -188,70 +188,59 @@ export interface TreeBranch extends IDisposable {
...
+	 * @remarks
+	 * All of the changes in the transaction are applied synchronously and therefore no other changes from a remote client can be interleaved with those changes.
+	 * Note that this is guaranteed by Fluid for any sequence of changes that are submitted synchronously, whether in a transaction or not.
 	 *
-	 * - A "success" flag indicating whether the transaction was successful or not.
-	 * - The success or failure value as returned by `transaction`.
+	 * {@link (TreeBeta:interface).on | Change events } will be emitted on this client for each edit as it happens, just as they would be if the changes were made outside of a transaction.
+	 * However, after the transaction completes the edits are effectively "squashed".
+	 * Any other/future clients that receive the transaction will apply its changes all at once, emitting only a single event per node (even if that node was edited multiple times in the transaction).
+	 * Edits to the tree are not permitted within these event callbacks, therefore no other local changes from this client will be interleaved with the changes in this transaction.
```
yann-achard-MS: Last nit: other branches on the same client won't get events for each edit. Is "context" the right descriptor for a `TreeCheckout`? Or`TreeView`?
```suggestion
	 * {@link (TreeBeta:interface).on | Change events } will be emitted on this context for each edit as it happens, just as they would be if the changes were made outside of a transaction.
	 * However, after the transaction completes the edits are effectively "squashed".
	 * Any other/future contexts that receive or merge the transaction will apply its changes all at once, emitting only a single event per node (even if that node was edited multiple times in the transaction).
	 * Edits to the tree are not permitted within these event callbacks, therefore no other local changes from this client will be interleaved with the changes in this transaction.
```

noencke: Good suggestion. The events are actually being emitted on the _nodes_, so I changed it to that and I think it dodges the problem. I also reworded so that all of this is still appropriate documentation for the unhydrated case, but still provides clarity about the hydrated case, which is what people will care about the most.

yann-achard-MS (approved):

PR #26513, opened 2026-02-20
text copy/paste formatting

brrichards:
# Summary

Currently copy/pasting into the text-editor example will retain bold, italicized, underline formatting but will drop font size/family. Using Quill's built in clipboard module provides information about the pasted text. Using this information to cross-check the pasted text with the supported size/family in `formattedTextAsTree` and applying the proper attribute changes will allow the example to retain font size/family. 

Files changed: 
- `quillFormattedView.tsx` - add clipboard module and two helper functions to parse font family and size


examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
```diff
@@ -72,6 +72,53 @@ const sizeReverse = { 10: "small", 18: "large", 24: "huge" } as const;
...
+function parseCssFontFamily(node: HTMLElement): string | undefined {
+	const style = node.style.fontFamily;
+	if (style === "") return undefined;
+
+	// Get the last font in the stack (generic family)
+	const fonts = style.split(",");
+	const last = fonts[fonts.length - 1]
+		?.trim()
+		.replace(/^["']/, "")
+		.replace(/["']$/, "")
```
CraigMacomber: Looking at some docs  ( https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-family ) I think this isn't quite the right approach.

Seems like we are supposed to try fonts from first to last.

I think this will fail to parse cases with commas inside quotes string correctly, but I guess thats fine as non of the strings we try and match have comas in them. Might be good to note that (and lack of support for handeling escapes) in a comment.

examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
```diff
@@ -72,6 +72,53 @@ const sizeReverse = { 10: "small", 18: "large", 24: "huge" } as const;
...
+		.toLowerCase();
+
+	// Map generic families to Quill values
+	if (last === "monospace") return "monospace";
+	if (last === "serif") return "serif";
+	if (last === "sans-serif") return "sans-serif";
```
CraigMacomber: Having a set with our allowed font strings in it, then testing against that seems like a cleaner way to do these checks.

examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
```diff
@@ -261,6 +308,26 @@ const FormattedTextEditorView = React.forwardRef<
...
+
+						const size = parseCssFontSize(node);
+						const font = parseCssFontFamily(node);
+
+						// If no formatting to apply, return unchanged
+						if (size === undefined && font === undefined) return delta;
+
+						// Build attributes object
+						const attrs: Record<string, unknown> = {};
+						if (size !== undefined) attrs.size = size;
```
CraigMacomber: I there any downside to setting this unconditionally? I know it would create a propery with value undefined, but is that a problem in this context? If now, we can remove the ifs.

examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
```diff
@@ -261,6 +308,26 @@ const FormattedTextEditorView = React.forwardRef<
...
+				clipboard: [
+					Node.ELEMENT_NODE,
+					(node: Node, delta: Delta): Delta => {
+						if (!(node instanceof HTMLElement)) return delta;
+
+						const size = parseCssFontSize(node);
+						const font = parseCssFontFamily(node);
+
+						// If no formatting to apply, return unchanged
+						if (size === undefined && font === undefined) return delta;
```
CraigMacomber: This early return looks like its just does as a performance optimization, and not a correctness thing? It would be good to add a couple words to the comment to claify.

Also, I think it might be better to refactor this to have a single return at the bottom, and just skip some code with an if here. That way if someone adds another kind of attribute than font or size here, it will help avoid a bug where they put it at the bottom and the early return skips it.

The pattern of update the delta for each kind of property we need, then return it at the end is more maintainble as its easier to change which properties exist without breaking the code.

examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
```diff
@@ -261,6 +308,26 @@ const FormattedTextEditorView = React.forwardRef<
...
+
+						// Build attributes object
+						const attrs: Record<string, unknown> = {};
+						if (size !== undefined) attrs.size = size;
+						if (font !== undefined) attrs.font = font;
+
+						// Apply formatting using compose/retain pattern per Quill docs
```
CraigMacomber: would be good to link the docs

examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
```diff
@@ -261,6 +308,26 @@ const FormattedTextEditorView = React.forwardRef<
 					[{ font: [] }],
 					["clean"],
 				],
+				clipboard: [
+					Node.ELEMENT_NODE,
+					(node: Node, delta: Delta): Delta => {
```
CraigMacomber: This function looks like something that could be tested if it were declared as a top level function. Might be nice to split it out as its own stand alone thing.

examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
```diff
@@ -72,6 +72,53 @@ const sizeReverse = { 10: "small", 18: "large", 24: "huge" } as const;
 const defaultSize = 12;
 /** Default font when no explicit font is specified. */
 const defaultFont = "Arial";
+/**
+ * Parse CSS font-size from a pasted HTML element's inline style.
+ * Returns a Quill size name if the pixel value matches a supported size, undefined otherwise.
+ * 12px is the default size and returns undefined (no Quill attribute needed).
+ */
+function parseCssFontSize(node: HTMLElement): string | undefined {
```
CraigMacomber: I like these two functions: they are pure functional, making them very easy to unit test. Might be nice to include a minimal unit test for each.

examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
brrichards: Since generics are at the back and thats all we support now besides ariel I thought it would be faster to check just the last element. You're right looking at the docks I don't know if we would ever read arial or future supported fonts if we don't going first to last. I'll change and add the comment as well

examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
brrichards: Good point I'll go ahead and refactor using this approach

examples/data-objects/text-editor/src/test/textEditor.test.tsx
```diff
@@ -785,6 +789,79 @@ describe("textEditor", () => {
...
+					assert.equal(parseCssFontSize(styledElement({ fontSize: "42px" })), undefined);
+				});
+			});
+
+			describe("parseCssFontFamily", () => {
+				it("returns undefined when no fontFamily is set", () => {
+					assert.equal(parseCssFontFamily(styledElement({})), undefined);
+				});
+
+				it("returns first recognized font in a comma-separated stack", () => {
```
CraigMacomber, line 822: I feel like this should include a case there you code would fail if it returned the last recognized item, or if it didn't handle quotes, or maybe a case other than monospace.

examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
```diff
@@ -68,10 +68,83 @@ FormattedMainView.displayName = "FormattedMainView";
...
+ * and escape sequences within font names are not supported. This is fine
+ * since none of the font names we match against contain commas or escapes.
+ */
+export function parseCssFontFamily(node: HTMLElement): string | undefined {
+	const style = node.style.fontFamily;
+	if (style === "") return undefined;
+
+	const fonts = style.split(",");
+	for (const raw of fonts) {
+		const font = raw.trim().replace(/^["']/, "").replace(/["']$/, "");
```
CraigMacomber, line 116:
```suggestion
		// Trim whitespace and leading and trailing quotes
		const font = raw.trim().replace(/^["']/, "").replace(/["']$/, "");
```

examples/data-objects/text-editor/src/formatted/quillFormattedView.tsx
```diff
@@ -68,10 +68,83 @@ FormattedMainView.displayName = "FormattedMainView";
...
+}
+
+/**
+ * Parse CSS font-family from a pasted HTML element's inline style.
+ * Tries fonts in priority order (first to last per CSS spec) and returns
+ * the first recognized Quill font value.
+ *
+ * Note: splitting on "," does not handle commas inside quoted font names,
+ * and escape sequences within font names are not supported. This is fine
+ * since none of the font names we match against contain commas or escapes.
```
CraigMacomber: This note is about the why this specific implementation is valid, and thus should be a comment inside this function on the relevant line, rather than a doc comment on the function. If this design decision impacted users of this function (like if the allowed font set was passed in, and wouldn't handle escapes or commas), then this would make sense where it is (though probably in a remarks block)

CraigMacomber (approved):

PR #26515, opened 2026-02-20
Replace hydrateUnsafe with a simpler and safer hydrateNode, and cleanup simpleTreeBench

CraigMacomber:
## Description

Replace hydrateUnsafe with a simpler and safer hydrateNode.

Refactor simpleTreeBench to reduce duplicated code, and fix incorrect references to hydrated trees as flex trees (flex trees are a lower level of abstraction which we might want to compare to, but these tests do not compare to).

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/test/simple-tree/simpleTree.bench.ts
+++ b/packages/dds/tree/src/test/simple-tree/simpleTree.bench.ts
@@ -7,7 +7,7 @@ import { strict as assert, fail } from "node:assert";
 
 import { BenchmarkType, benchmark, isInPerformanceTestingMode } from "@fluid-tools/benchmark";
 
-import { SchemaFactory, SchemaFactoryAlpha } from "../../simple-tree/index.js";
+import { SchemaFactory, SchemaFactoryAlpha, type TreeNode } from "../../simple-tree/index.js";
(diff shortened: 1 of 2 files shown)


packages/dds/tree/src/test/simple-tree/simpleTree.bench.ts
```diff
@@ -89,52 +89,40 @@ describe("SimpleTree benchmarks", () => {
...
+						title: `${title} (${doHydration ? "hydrated" : "unhydrated"} node)`,
+						before: () => {
+							tree = unhydratedNodeInitFunction();
+							if (doHydration) {
+								hydrateNode(tree);
+							}
+						},
+						benchmarkFn: () => {
+							readNumber = treeReadingFunction(
+								tree ?? fail("Expected unhydratedTree to be set"),
```
brrichards: Should this be `("Expected tree to be set")` with the variable rename?

CraigMacomber:
```suggestion
								tree ?? fail("Expected tree to be set"),
```

taylorsw04 (approved):

brrichards (approved):

PR #26520, opened 2026-02-23
Fix parallel forked containers causing IdCompressor "Ranges finalized out of order" 

dannimad:
### Bug scenario:
When a container forks (two instances load from the same stashed state), the ID Compressor in each fork has the same session ID. Both independently generate IDs and submit IdAllocation ops:
Fork 1 sends: "Session ABC claims IDs 1-10"
Fork 2 sends: "Session ABC claims IDs 1-10" (same range — they started from the same state)
When finalizeCreationRange processes both, the compressor sees overlapping ranges for the same session, hitting "Ranges finalized out of order" inside the compressor itself, which is a confusing error with no indication it's a fork.

### Fix
Detect duplicate (overlapping) IdAllocation ranges from forked containers that share the same session ID but submit from different clients. IdAllocation batches use ignoreBatchId, so the DuplicateBatchDetector cannot catch these — this adds range-level overlap detection before finalizeCreationRange.

- Add a new e2e test variant that explicitly covers parallel forks with ID allocation ops.
- Add unit tests covering: overlapping ranges (same session, different clients), sequential ranges (same session, allowed), different sessions (allowed), and partial overlap detection.

### DIFF
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -1326,6 +1326,13 @@ export class ContainerRuntime
 	// Once it loads, it will process all such ops and we will stop accumulating further ops - ops will be processes as they come in.
 	private pendingIdCompressorOps: IdCreationRange[] = [];
 
+	/**
+	 * Tracks the next expected firstGenCount for each session's IdAllocation ranges.
+	 * Used to detect duplicate (overlapping) IdAllocation ranges from forked containers
+	 * that share the same session ID but submit from different clients.
+	 */
+	private readonly finalizedIdAllocationRanges = new Map<string, number>();
+
 	// Id Compressor serializes final state (see getPendingLocalState()). As result, it needs to skip all ops that preceeded that state
 	// (such ops will be marked by Loader layer as savedOp === true)
 	// That said, in "delayed" mode it's possible that Id Compressor was never initialized before getPendingLocalState() is called.
@@ -3431,6 +3438,36 @@ export class ContainerRuntime
 						this.pendingIdCompressorOps.length === 0,
 						0x979 /* there should be no pending ops! */,
 					);
+
+					// Detect duplicate (overlapping) IdAllocation ranges from forked containers.
+					// Forked containers share the same session ID but submit from different clients,
+					// producing IdAllocation ranges with the same sessionId and overlapping genCounts.
+					// IdAllocation batches use ignoreBatchId, so the DuplicateBatchDetector cannot
+					// catch these duplicates — we must detect them here before finalizeCreationRange.
+					if (range.ids !== undefined) {
+						const nextExpectedGenCount = this.finalizedIdAllocationRanges.get(
+							range.sessionId,
+						);
+						if (
+							nextExpectedGenCount !== undefined &&
+							range.ids.firstGenCount < nextExpectedGenCount
+						) {
+							throw new DataCorruptionError(
+								"Duplicate IdAllocation range detected - overlapping ID ranges for the same session (likely a container fork)",
+								{
+									sessionId: range.sessionId,
+									firstGenCount: range.ids.firstGenCount,
+									count: range.ids.count,
+									nextExpectedGenCount,
+								},
+							);
+						}
+						this.finalizedIdAllocationRanges.set(
+							range.sessionId,
+							range.ids.firstGenCount + range.ids.count,
+						);
+					}
+
 					this._idCompressor.finalizeCreationRange(range);
 				}
 			}
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -3445,9 +3445,7 @@ export class ContainerRuntime
 					// IdAllocation batches use ignoreBatchId, so the DuplicateBatchDetector cannot
 					// catch these duplicates — we must detect them here before finalizeCreationRange.
 					if (range.ids !== undefined) {
-						const nextExpectedGenCount = this.finalizedIdAllocationRanges.get(
-							range.sessionId,
-						);
+						const nextExpectedGenCount = this.finalizedIdAllocationRanges.get(range.sessionId);
 						if (
 							nextExpectedGenCount !== undefined &&
 							range.ids.firstGenCount < nextExpectedGenCount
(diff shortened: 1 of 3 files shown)


packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3431,6 +3438,34 @@ export class ContainerRuntime
 						this.pendingIdCompressorOps.length === 0,
 						0x979 /* there should be no pending ops! */,
 					);
+
+					// Detect duplicate (overlapping) IdAllocation ranges from forked containers.
+					// Forked containers share the same session ID but submit from different clients,
+					// producing IdAllocation ranges with the same sessionId and overlapping genCounts.
+					// IdAllocation batches use ignoreBatchId, so the DuplicateBatchDetector cannot
```
markfields: We can get more to the point here by saying "IdAllocation batches don't have a stable batch identity over resubmit, so we don't assign them batchIds and the DuplicateBatchDetector cannot..."

Or you can phrase it more clearly.  The point is that on resubmit, ID allocations are recalculated based on what's actuall resubmitted, rather than the original ID allocations being resubmitted.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3431,6 +3438,34 @@ export class ContainerRuntime
...
+								"Duplicate IdAllocation range detected - overlapping ID ranges for the same session (likely a container fork)",
+								{
+									sessionId: range.sessionId,
+									firstGenCount: range.ids.firstGenCount,
+									count: range.ids.count,
+									nextExpectedGenCount,
+								},
+							);
+						}
+						this.finalizedIdAllocationRanges.set(
```
markfields, line 3463: What would it look like to move this duplicate tracking into ID Compressor itself? (Inside finalizeCreationRange)

packages/test/test-end-to-end-tests/src/test/offline/stashedOps.spec.ts
```diff
@@ -2251,7 +2251,137 @@ describeCompat(
...
+					counter1.value,
+					incrementValue,
+					"container1 should have incremented to 3 before closing",
+				);
+			},
+		);
+
+		itExpects(
+			`Parallel Forks: Closes (ForkedContainerError and DuplicateBatchError) when hydrating twice and submitting in parallel (via Counter DDS)
+			with ID Allocation op`,
```
markfields, line 2374: Can we share code between these 2 tests?  Reading through the code are they identical except for that one line that enables/disables ID Compressor?

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -370,6 +370,145 @@ describe("Runtime", () => {
...
+					// Second range: genCounts 6-10 (sequential, no overlap)
+					containerRuntime.process(
+						makeIdAllocationOp(localSessionId, 6, 5, remoteClientId, 2),
+						false /* local */,
+					);
+					// No error — sequential ranges for the same session are fine (normal reconnection)
+				});
+
+				it("Allows IdAllocation ranges from different sessions", () => {
+					const otherSessionId = "aaaaaaaa-bbbb-4ccc-dddd-eeeeeeeeeeee";
```
markfields, line 474:
```suggestion
					const otherSessionId = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee";
```

Haha I'm kidding, it doesn't matter what it is but I thought it was funny the "4" in the middle

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -370,6 +370,145 @@ describe("Runtime", () => {
...
+				});
+
+				it("Detects partial overlap where new range starts before the end of the previous one", () => {
+					// First range: genCounts 1-5
+					containerRuntime.process(
+						makeIdAllocationOp(localSessionId, 1, 5, remoteClientId, 1),
+						false /* local */,
+					);
+
+					// Overlapping range starting at genCount 3 (before nextExpected=6)
```
markfields, line 496: I wonder if we should make this case an assert - no known way to trigger this case in reality is there?

packages/runtime/container-runtime/src/containerRuntime.ts
dannimad: I've been playing with it and by doing so it breaks some other id compressor tests. I think this is a better place to target the forked scenario better, given that finalizeCreationRange is used in different places and id compressor already has other mechanisms to detect similar scenarios.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
dannimad: I don't think so but given I'm no expert on id compressor I wouldn't assert that. Seems only relevant to internal id compressor logic so I wouldn't add that assert in this PR

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
dannimad: I wonder what on earth Claude was thinking writing that 4 mmm

packages/runtime/container-runtime/src/containerRuntime.ts
markfields: Ok. Yeah I can see how trying to modify the core `finalizeCreationRange` fn could be troublesome.

I'm always leery of adding a too much niche logic to this file/class.  Can we add a new fn to `IIdCompressorCore` like `tryFinalizeCreationRange` that returns status about whether it succeeded (rather than throwing), and includes this check?

That type is legacy-beta so you'd have to add it as optional but I know Taylor wants to move it to internal anyway so we can consider it pretty flexible.

markfields: @taylorsw04 can you please review this new logic?  Any thoughts on moving it into the ID Compressor impl rather than adding it here?

packages/test/test-end-to-end-tests/src/test/offline/stashedOps.spec.ts
```diff
@@ -2230,141 +2230,160 @@ describeCompat(
...
+				category: "error",
+			},
+			{
+				eventName: "fluid:telemetry:Container:ContainerClose",
+				category: "error",
+			},
+		] as const;
+
+		itExpects(
+			`Parallel Forks: Closes (ForkedContainerError and DuplicateBatchError) when hydrating twice and submitting in parallel (via Counter DDS)`,
```
markfields, line 2363: How do you know this won't fail for t9s? Isn't it equivalent to the former single test?

markfields: Wait, previously it was only run for local.  Shouldn't you keep it that way for both new tests?

PR #26522, opened 2026-02-23
fix(client-tree): fix cross-package schema consumption via entrypoint restructuring

WillieHabi:
## Description

Consumer packages compiling against `@fluidframework/tree` could generate `.d.ts` files with incorrect import paths, as tree's entrypoint files all pointed to `index.js`. This change fixes the issue by giving each API tier (`.`, `./alpha`, `./beta`, etc.) its own dedicated JS entrypoint under `src/entrypoints/`.

 - Entrypoint source files are now committed under `packages/dds/tree/src/entrypoints/` instead of being generated at build time. Changes for exposed API surface require updating files under `src/entrypoints`, which may use the `generate:entrypoint-sources` script.
 - Each export path in `package.json` now maps to its own `.js` and `.d.ts` file rather than sharing `index.js`.
 - Until 0.64 build-tools are used, `legacy.d.ts` Node10 file won't be generated and is temporarily checked in.


examples/utils/cross-package-schema/schema-provider/package.json
```diff
@@ -0,0 +1,61 @@
...
+	"exports": {
+		".": {
+			"import": {
+				"types": "./lib/index.d.ts",
+				"default": "./lib/index.js"
+			}
+		},
+		"./schema": {
+			"types": "./src/index.ts",
+			"default": "./lib/index.js"
```
jason-ha: Should only support "import" condition - like is used in `"."`.

examples/utils/cross-package-schema/schema-provider/tsconfig.bundler.json
```diff
@@ -0,0 +1,12 @@
+{
+	"extends": "../../../../common/build/build-common/tsconfig.node16.json",
+	"include": ["src/**/*"],
+	"compilerOptions": {
+		"rootDir": "./src",
+		"outDir": "./lib",
+		"exactOptionalPropertyTypes": false,
```
jason-ha: Why do we have two tsconfigs targetting `./lib`? There should only be one tsconfig per outDir except for special overlaid cases with unique sets of included source files.

examples/utils/cross-package-schema/schema-provider/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+	"extends": "../../../../common/build/build-common/tsconfig.node16.json",
+	"include": ["src/**/*"],
+	"compilerOptions": {
+		"rootDir": "./src",
+		"outDir": "./lib",
+		"exactOptionalPropertyTypes": false,
```
jason-ha: Do you need this? We prefer to not have this setting when possible.
I think you do - should comment on the need.

examples/utils/cross-package-schema/README.md
```diff
@@ -0,0 +1,85 @@
...
+  }
+}
+```
+
+### Consumer side (`schema-consumer`)
+
+Import from the `/schema` subpath
+
+```typescript
+import { AppState } from "@my-provider/schema";
```
jason-ha: This is not importing from a schema subpath as the first element is a scope per the `@`. This imports from a package named `@my-provider/schema`.

examples/utils/cross-package-schema/README.md
```diff
@@ -0,0 +1,85 @@
+# Cross-Package SharedTree Schema Example
+
+This example demonstrates a pattern for consuming SharedTree schemas across package boundaries
+when those schemas are defined using `SchemaFactoryAlpha.objectAlpha()`.
+
+## The Problem
+
+When SharedTree schema classes are defined using `objectAlpha()` and compiled with
+TypeScript 5.9+ using `moduleResolution: "bundler"`, the `.d.ts` emitter normalizes import
```
jason-ha: Is using bundler resolution a part of the problem? I didn't think that was part of the issue. The only issue as I understood it was working with generated .d.ts files.

examples/utils/cross-package-schema/README.md
```diff
@@ -0,0 +1,85 @@
...
+paths in a way that breaks cross-package consumption.
+
+## The Solution
+
+Add a `/schema` subpath export in the provider's `package.json` whose `"types"` field
```
jason-ha: Let's not say `"types"` field. We can say "types" import resolution.

examples/utils/cross-package-schema/schema-provider/package.json
```diff
@@ -0,0 +1,61 @@
...
+		}
+	},
+	"scripts": {
```
jason-ha: For completeness, there should be a "files" block in here. "files" says what files are actually part of a published package. See `presence` package for an example. In this case, we want to call out `src` explicitly as some or all of it is needed. (Even if there aren't test files, I would exclude those paths to set a good example.)

The above is an old comment that was never apparently published. It matters more and less now that we don't need to publish src. More because a proper package wouldn't have to publish them. Less because this is super simple now.

examples/utils/cross-package-schema/README.md
jason-ha: Whoa. I'm reading step 1 below. That suggests bundler is part of the problem. So, the solution is maybe is don't use bundler resolution for your schema package? (office-bohemia doesn't require bundler resolution. It is largely used because it has relaxed rules. But should be easy to build schema package with node16 resolution.)

examples/utils/cross-package-schema/schema-provider/tsconfig.bundler.json
jason-ha: After reading some other things, you might want to emit bundler to a `bundler` directory and export that (additional export).
Build config will need to be updated to make sure all is built.

biome.jsonc
```diff
@@ -34,6 +34,7 @@
 			"!**/src/**/test/types/**/*.generated.ts",
 			"!**/src/packageVersion.ts",
 			"!**/src/layerGenerationState.ts",
+			"!**/src/entrypoints/*.ts",
```
jason-ha: We do not want a broad disable of src/entrypoints from biome. The disable in tree should be temporary. If the package level biome config is a problem, then make this "absolute" just for tree with a comment.

pnpm-lock.yaml
```diff
@@ -30425,7 +30496,7 @@ snapshots:
       mdast: 3.0.0
       mdast-util-heading-range: 4.0.0
       mdast-util-to-string: 4.0.0
-      minimatch: 10.1.1
+      minimatch: 10.2.1
```
jason-ha: okay, I think, but rogue unrelated change

examples/utils/cross-package-schema/schema-consumer/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+	"extends": "../../../../common/build/build-common/tsconfig.node16.json",
+	"include": ["src/**/*"],
+	"compilerOptions": {
+		"rootDir": "./src",
+		"outDir": "./lib",
+		// Need to skip lib check since some .d.ts from provider are invalid.
+		"skipLibCheck": true,
```
jason-ha: No longer the case. All builds should be good.

jason-ha (commented):
Looks pretty good - you'll obviously need another review since I have decent contributions in here.
Do update the PR title and description.
Apologies if this posts with seemingly irrelevant comments. I apparently never submitted notes I had from last week. I tried to cleanup things that I could see to be accurate.

jason-ha:
> Do update the PR title and description.

Do borrow liberally from https://github.com/jason-ha/FluidFramework/commit/4c04950418c2b8fe0fa3540c9f54d0f75764e794 description. I am not planning on submitting those separately as we want to have tests in place. You get the credit. :)

CraigMacomber:
If possible, do testing of stuff like this in the `import-testing` package, which exists for the purpose of testing the exporting of content using fluid types then importing of them.

Realistically I believe what you want to test is d.ts generation and consumption, which can be done with two ts projects in a single package, which is how import-testing is setup. This approach should be able to test what we need, as well as serve as examples of how to do it, while not requiring extra packages.

Note that this package already has tests/example for a lot of schema related APIs.

packages/dds/tree/src/entrypoints/alpha.ts
```diff
@@ -0,0 +1,301 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/*
+ * THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
CraigMacomber, line 7: I'd prefer we don't commit auto generated files, and instead have them produces as part of the build, unless it is intended for human auditing like API reports. As we already have an API report for these, I don't think we want this as well for the same purpose.

jason-ha: This is exactly the plan. The API report files give us one thing, but they don't tell us what is exactly happening and we can miss things.
`InternalTypes` is an example of that. Our current "generate entrypoints" actually emits and error for `InternalTypes` but you don't see it when run as part of the task infrastructure - run it on its own to see. You can also look at the current lib/public.d.ts to see the funkiness.
In the future (see AB#14504), we'll be able to review API surface changes more easily (easier to understand) than with API reports. (Though we'll likely keep one API report to keep track of documented/undocumented and tags.)
In some cases, we may want to temporarily or permanently handcraft parts of the entrypoint files.

This result is not exactly what we want, but it is sufficient.

jason-ha:
> If possible, do testing of stuff like this in the `import-testing` package, which exists for the purpose of testing the exporting of content using fluid types then importing of them.
> 
> Realistically I believe what you want to test is d.ts generation and consumption, which can be done with two ts projects in a single package, which is how import-testing is setup. This approach should be able to test what we need, as well as serve as examples of how to do it, while not requiring extra packages.
> 
> Note that this package already has tests/example for a lot of schema related APIs.

I think testing can be relocated. I don't know that need (see below) to bother with two ts projects, though I know you can test that way as one form of setup in my personal iterations had observed that.

I do know that these cases need to be isolated. I observed too many things that would avoid recreating the issue. Maybe separate project is the best to help ensure that.

I don't think it is bad to have a clear example of practical use. Multiple ts projects does complicate the build a bit and is a lot less obvious.

I'm okay with this or rolled into `import-testing`.

jason-ha:
@WillieHabi, to fix the build break probably should apply this diff to fluidBuild.config.cjs:
```diff
-		"build:api-reports:current": ["api-extractor:esnext"],
-		"build:api-reports:legacy": ["api-extractor:esnext"],
-		"ci:build:api-reports:current": ["api-extractor:esnext"],
-		"ci:build:api-reports:legacy": ["api-extractor:esnext"],
+		// Depend on "build:esnext" in case there is no "api-extractor:esnext".
+		"build:api-reports:current": ["api-extractor:esnext", "build:esnext"],
+		"build:api-reports:legacy": ["api-extractor:esnext", "build:esnext"],
+		"ci:build:api-reports:current": ["api-extractor:esnext", "build:esnext"],
+		"ci:build:api-reports:legacy": ["api-extractor:esnext", "build:esnext"],
```

examples/utils/cross-package-schema/schema-consumer/tsconfig.json
WillieHabi: Think this maybe still be needed as removing failed CI here: https://dev.azure.com/fluidframework/public/public%20Team/_build/results?buildId=381732&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=bfe27d5f-9219-57c0-0d67-1b2946a32921&l=198

WillieHabi: Also some CI failures at 'npm run ci:check:are-the-types-wrong' stage: https://dev.azure.com/fluidframework/public/public%20Team/_build/results?buildId=381745&view=logs&j=5122cb2b-0fab-5462-27ab-7d04b7dfe6ec&t=90a828af-f36d-57be-f9c1-197780797175&l=2273

examples/utils/cross-package-schema/schema-consumer/tsconfig.json
```diff
@@ -4,7 +4,9 @@
 	"compilerOptions": {
 		"rootDir": "./src",
 		"outDir": "./lib",
-		// Need to skip lib check since some .d.ts from provider are invalid.
+		// Tree's .d.ts files have internal type incompatibilities (e.g. UnhydratedFlexTreeField.parent
+		// is UnhydratedFlexTreeNode | undefined but FlexTreeField interface requires FlexTreeNode).
+		// These surface as errors in any downstream consumer without skipLibCheck.
 		"skipLibCheck": true,
```
jason-ha: We could do that, but I am not sure it is in the spirit of these examples.
I think the problem is that tree has a problem with `exactOptionalPropertyTypes: true`. Setting that to `false` here like is done in provider should solve the problem. Then both tsconfigs should have a comment.

Alternative solution is to avoid it by testing via import-testing.

examples/utils/cross-package-schema/schema-consumer/tsconfig.json
jason-ha: That looks like a subtle fluid build config flaw. `tree`'s "api" task should be executed. The trouble is the script: false setting for "api" task when I added an actual command to that package entry.
That step was usually getting skipped - we only check at the very end of the build.
Anyway, I fixed that in https://github.com/jason-ha/FluidFramework/tree/api/tree-schema-expressions

jason-ha (commented):
Current state for me, I see three issues remaining:
1. PR description needs updated as ts 5.9 and bundler are not relevant
2. producer exactOptionalPropertyTypes needs commented
3. consumer skipLibCheck should be [commented] exactOptionalPropertyTypes: false

Alternatively for 2+3, replace with import-testing as preferred by Craig.

WillieHabi:
> Current state for me, I see three issues remaining:
> 
> 1. PR description needs updated as ts 5.9 and bundler are not relevant
> 2. producer exactOptionalPropertyTypes needs commented
> 3. consumer skipLibCheck should be [commented] exactOptionalPropertyTypes: false
> 
> Alternatively for 2+3, replace with import-testing as preferred by Craig.

Ok I've deleted the cross-package-schema packages. I was looking to move over to import-testing, but was wondering what does import-testing not already test for (compared to what was being tested with the cross-schema-package tests)? Might be mistaken but it looks redundant to move the minimal example we had over there -- have these import tests caught any of the issues we had previously?

jason-ha:
> Might be mistaken but it looks redundant to move the minimal example we had over there -- have these import tests caught any of the issues we are seeing previously?

Part of the issue is that other code in the vicinity of the minimal example code we crafted can alter the result.
1) As with all testing, you should verify the new test fails without the implementation correction.
2) Because code in vicinity can alter the result, it is perhaps too easy to break the test and not notice unless the test code is very isolated. It can definitely be done with added tsconfig in import-testing, but it is not as simple as copy the files over.

examples/utils/import-testing/package.json
```diff
@@ -17,12 +17,14 @@
...
-		"build:test": "npm run build:test:esm && npm run build:test:cjs",
+		"build:esnext:crosspackage": "tsc --project ./tsconfig.crossPackage.json",
+		"build:test": "npm run build:test:esm && npm run build:test:cjs && npm run build:test:crosspackage",
 		"build:test:cjs": "fluid-tsc commonjs --project ./src/test/tsconfig.cjs.json",
+		"build:test:crosspackage": "tsc --project ./src/test/tsconfig.crossPackage.json",
 		"build:test:esm": "tsc --project ./src/test/tsconfig.json",
 		"check:biome": "biome check .",
 		"check:format": "npm run check:biome",
-		"clean": "rimraf --glob dist lib \"**/*.tsbuildinfo\" \"**/*.build.log\" nyc",
+		"clean": "rimraf --glob dist lib lib-crosspackage \"**/*.tsbuildinfo\" \"**/*.build.log\" nyc",
```
jason-ha: No need to introduce another output directory.
Change ./tsconfig.json to exclude crossPackageSchemaExport (should be done anyway)
Let ./tsconfig.crossPackage.json out to lib.

examples/utils/import-testing/package.json
```diff
@@ -17,12 +17,14 @@
 		"build:commonjs": "fluid-build . --task commonjs",
 		"build:compile": "fluid-build . --task compile",
 		"build:esnext": "tsc --project ./tsconfig.json",
```
jason-ha: `build:esnext` should cover both builds. This should use concurrently or compound command line.

examples/utils/import-testing/src/crossPackageSchemaExports.ts
```diff
@@ -3,9 +3,20 @@
...
+By compiling only these schemas, we ensure the test accurately detects regressions in
+objectAlpha/mapAlpha/optional cross-package consumption.
+*/
+
 /* eslint-disable jsdoc/require-jsdoc */
 
-import { sf } from "./schemaUtils.js";
+import { SchemaFactoryAlpha } from "@fluidframework/tree/alpha";
+
+const sf = new SchemaFactoryAlpha("cross-package-example");
```
jason-ha: So, the test fails if tree isn't restructured?

examples/utils/import-testing/src/test/tsconfig.crossPackage.json
jason-ha, line 1: I don't think there is need for this. The import isn't where the issue happens. The normal test project should be sufficient.

examples/utils/import-testing/src/test/crossPackageImporter.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+/*
+Compile-time only test — if this file compiles, the test passes. Not a .spec.ts, not run by mocha.
+The schemas in crossPackageSchemaExports.ts are compiled by a separate tsconfig (tsconfig.crossPackage.json)
+*/
+
+import { TreeViewConfiguration } from "@fluidframework/tree";
+
+import { AppState, Container, Dimensions, Position } from "../crossPackageSchemaExports.js";
```
jason-ha: Recommend adding exports to package.json and importing via package reference here. Helps with cross package clarity.

examples/utils/import-testing/tsconfig.crossPackage.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "../../../common/build/build-common/tsconfig.node16.json",
```
jason-ha: Recommend simplifying with reference to ./tsconfig.json

jason-ha (changes requested):

examples/utils/import-testing/src/crossPackageSchemaExports.ts
WillieHabi: Yes verified that the test fails without the entrypoint restructuring and passes with it.

examples/utils/import-testing/tsconfig.crossPackage.json
```diff
@@ -1,11 +1,5 @@
...
+	"extends": "./tsconfig.json",
 	"include": ["src/crossPackageSchemaExports.ts"],
-	"compilerOptions": {
-		"rootDir": "./src",
-		"outDir": "./lib-crosspackage",
-		"noUnusedLocals": false,
-		"skipLibCheck": false,
-		"exactOptionalPropertyTypes": false,
-	},
+	"exclude": [],
```
jason-ha: fine to have it, but I don't think you need it. I don't think include/exclude is inherited

examples/utils/import-testing/package.json
```diff
@@ -12,19 +12,29 @@
...
+		}
+	},
 	"scripts": {
 		"build": "fluid-build . --task build",
 		"build:commonjs": "fluid-build . --task commonjs",
 		"build:compile": "fluid-build . --task compile",
-		"build:esnext": "tsc --project ./tsconfig.json",
-		"build:esnext:crosspackage": "tsc --project ./tsconfig.crossPackage.json",
-		"build:test": "npm run build:test:esm && npm run build:test:cjs && npm run build:test:crosspackage",
+		"build:esnext": "tsc --project ./tsconfig.json && tsc --project ./tsconfig.crossPackage.json",
```
jason-ha: Is our incremental build system happy with this? I don't think it will be. I think these need to be on their own lines.

examples/utils/import-testing/src/crossPackageSchemaExports.ts
jason-ha: I checked out your branch locally and I don't get failure when I make things (tree) in a bad way. But I do if I then also restore this to split as I had.
Can you publish a branch with just the import-testing changes? Then we can double check results.

examples/utils/import-testing/eslint.config.mts
```diff
@@ -7,6 +7,16 @@ import type { Linter } from "eslint";
...
 
-const config: Linter.Config[] = [...strict, ...sharedConfig];
+const config: Linter.Config[] = [
+	...strict,
+	...sharedConfig,
+	{
+		files: ["src/test/crossPackageImporter.ts"],
+		rules: {
+			// Allow importing the cross-package schema via package self-reference subpath.
+			"import-x/no-internal-modules": "off",
```
jason-ha: I think there is a preference for in file suppression when there are a general set of things that need suprressed across several files.

jason-ha (commented):
I remain suspicious :)

examples/utils/import-testing/package.json
WillieHabi: so there's lint vs. policy checker issue. splitting into separate scripts causes a fluid-build-tasks-tsc policy failure since both tsconfigs output crossPackageSchemaExports.d.ts to lib/. the file is in both tsconfigs because eslint's project service needs it in tsconfig.json to lint it. so this compound command in a single build:esnext avoids this conflict. I'm not sure about incremental build concerns

examples/utils/import-testing/tsconfig.crossPackage.json
WillieHabi: Left a comment, exclude is inherited and overwrites the exclude from base config https://www.typescriptlang.org/tsconfig/#extends

examples/utils/import-testing/src/crossPackageSchemaExports.ts
WillieHabi: Ok I spent time experimenting and you're right that the test as it was couldn't reproduce the failure with **Node16 resolution**:

The bug requires moduleResolution: "bundler" to manifest. When tree's export tiers all point to the same JS module (lib/index.js), the bundler-mode declaration emitter normalizes import("@fluidframework/tree/alpha") to import("@fluidframework/tree") in the provider's .d.ts output. Since ObjectNodeSchema is alpha-only and not in the public tier, a consumer can't resolve it, causing `TS2322: Type 'typeof AppState' is not assignable to type 'ImplicitFieldSchema'.`

With Node16 resolution, TypeScript preserves the original import path regardless of the export structure — so the test always passed.

So now, to reproduce in import-testing without a two-package setup, we added:
  - tsconfig.crossPackage.bundler.json — builds the same schema source with moduleResolution: "bundler", outputting to lib/bundler/
  - crossPackageBundlerImporter.ts — consumes those .d.ts files via a ./crossPackageSchemaBundler export

With your tree's fixed exports, the bundler can't normalize the paths since they resolve to different modules and the test passes. With old exports, the bundler normalizes and the consumer fails.

examples/utils/import-testing/package.json
jason-ha: When you build, fluid build will append `(non-incremental)` to any tasks it doesn't understand dependencies for. We have other packages with multiple "production" build steps. See presence for an example package.json.
For linting, typescript-eslint's projectService won't discover secondary projects that aren't referenced by ones that it recognizes. See various eslint.config.mts 's with "tsconfig" in them.
(This is all of the goo BTW that we'd avoid by using the example packages you originally established.)

examples/utils/import-testing/tsconfig.crossPackage.json
jason-ha: I guess I was thinking of references.

examples/utils/import-testing/tsconfig.crossPackage.bundler.json
```diff
@@ -0,0 +1,15 @@
...
+	"extends": "./tsconfig.json",
+	"include": ["src/crossPackageSchemaExports.ts"],
+	// Override the exclude from tsconfig.json which excludes this file.
+	"exclude": [],
+	"compilerOptions": {
+		"outDir": "./lib/bundler",
+		// Use bundler resolution to reproduce TS 5.9's import path normalization bug:
+		// when all tree export tiers resolve to the same JS module (lib/index.js),
+		// TS 5.9's bundler emitter normalizes import("@fluidframework/tree/alpha")
+		// to import("@fluidframework/tree"), losing alpha-only types like ObjectNodeSchema.
```
jason-ha: Two problems with comment:
1. 5.9 is not invoked here. 5.4 is used.
2. There isn't specifically a bundler resolution bug. There is a general one that appears to be exacerbated by bundler.
The general problem is that type reference that needs resolved is declared in some deep file. The chain of files looks like alpha.d.ts -> index.js -> ... -> origin.js. Resolution will look for origin.js but failing that will work through the chain to find the deepest exported. The bug in compiler is that it will match index.js from "." export.
I *think* the Node16 has a hint or preference for existing imports, which means that it first checks alpha.d.ts (from existing tree/alpha import).
Assuming I am right, we need to check the known Node16 test case as that ensures isolation. (In other words, if bundler support were updated to peek at alpha.d.ts, then the test would be broken.)
It is best to compile for both Node16 and Bundler.

examples/utils/import-testing/tsconfig.crossPackage.bundler.json
```diff
@@ -0,0 +1,15 @@
+{
+	"extends": "./tsconfig.json",
+	"include": ["src/crossPackageSchemaExports.ts"],
+	// Override the exclude from tsconfig.json which excludes this file.
+	"exclude": [],
```
jason-ha: A better organization of the tsconfig's would be to place common settings in tsconfig.base.json. Then tsconfig.json and these would extend it. No include or exclude in the base.

examples/utils/import-testing/package.json
```diff
@@ -12,11 +12,29 @@
...
+	"exports": {
+		"./crossPackageSchema": {
+			"import": {
+				"types": "./lib/crossPackageSchemaExports.d.ts",
+				"default": "./lib/crossPackageSchemaExports.js"
+			},
+			"require": {
+				"types": "./dist/crossPackageSchemaExports.d.ts",
+				"default": "./dist/crossPackageSchemaExports.js"
+			}
```
jason-ha: Where does this get built?

examples/utils/import-testing/package.json
```diff
@@ -12,11 +12,29 @@
 	"license": "MIT",
 	"author": "Microsoft and contributors",
 	"type": "module",
+	"exports": {
+		"./crossPackageSchema": {
+			"import": {
+				"types": "./lib/crossPackageSchemaExports.d.ts",
+				"default": "./lib/crossPackageSchemaExports.js"
+			},
```
jason-ha: FYI, since we aren't doing type remapping. This can be simplified to `"import": "./lib/crossPackageSchemaExports.js"`. TypeScript resolution will let it look at the matching .d.ts.

examples/utils/import-testing/src/crossPackageSchemaExports.ts
jason-ha: Okay. Let's test both and use the harder pattern for us to get right - split to two files; so that Node16 problematic case is also covered.

examples/utils/import-testing/package.json
```diff
@@ -12,11 +12,29 @@
 	"license": "MIT",
 	"author": "Microsoft and contributors",
 	"type": "module",
+	"exports": {
+		"./crossPackageSchema": {
```
jason-ha: nit: qualify this by Node16 instead of leaving it implicit

examples/utils/import-testing/src/test/crossPackageBundlerImporter.ts
```diff
@@ -0,0 +1,37 @@
...
+	Position,
+} from "@fluid-example/import-testing/crossPackageSchemaBundler";
+/* eslint-enable import-x/no-internal-modules */
+
+const _config = new TreeViewConfiguration({ schema: AppState });
+const _container = new Container({
+	id: "test",
+	position: new Position({ x: 0, y: 0 }),
+	dimensions: new Dimensions({ width: 100, height: 100 }),
+});
```
jason-ha: both cases could be tested in one consumer file (matches the common multiple variants in one file pattern) - separate is okay too

examples/utils/import-testing/tsconfig.crossPackage.bundler.json
WillieHabi: Just responding to number one, I don't think this is the case. in my experiments with Claude I see that TS 5.9 is what gets invoked, which was weird. The explanation is that this package has multiple typescript versions as devDeps (typescript, typescript-5.4 through typescript-5.9) and they all register a tsc binary.  was explained saying "Only one can win in node_modules/.bin/tsc — in this case pnpm picked typescript-5.9". 

I was skeptical but you can verify by running `npx tsc --version` from examples/utils/import-testing/ and it prints 5.9.3.       

using ts version 5.4 actually results in the correct .d.ts alpha import path. So somewhere from 5.4 -> 5.9 something went wrong.       

EDIT: this is all for bundler module resolution btw

also sent Claude to run a test (haven't verfied 5.5-5.8 manually yet): 
| TS Version | Bundler .d.ts import paths              |
  |------------|-----------------------------------------|
  | 5.4        | import("@fluidframework/tree/alpha") ✅ |
  | 5.5        | import("@fluidframework/tree/alpha") ✅ |
  | 5.6        | import("@fluidframework/tree") ❌       |
  | 5.7        | import("@fluidframework/tree") ❌       |
  | 5.8        | import("@fluidframework/tree") ❌       |
  | 5.9        | import("@fluidframework/tree") ❌       |

examples/utils/import-testing/src/crossPackageSchemaExports.ts
WillieHabi: Hmm trying to repro with two-file split but still passes with Node16 , I have a experiement branch you can check out here I made with Claude: https://github.com/microsoft/FluidFramework/compare/main...WillieHabi:FluidFramework:cross-package-schema-split-experiment

examples/utils/import-testing/tsconfig.crossPackage.bundler.json
jason-ha: We need to fix that silent 5.9 use. That is not what is intended.
I didn't try Bundler, but I know for Node16 + 5.4 that it will fail. That is after all what we had in the split packages version - no 5.9 references.

examples/utils/import-testing/src/crossPackageSchemaExports.ts
WillieHabi: Fixed the experiment branch to mirror our PR, just without the tree fix

examples/utils/import-testing/package.json
WillieHabi: Now have `"build:esnext:bundler": "tsc --project ./tsconfig.crossPackage.bundler.json",` script explicitly defined

examples/utils/import-testing/src/test/crossPackage/bundlerImporter.ts
```diff
@@ -7,11 +7,8 @@
 Compile-time test for cross-package schema consumption where schemaDefinitions.d.ts
 was generated under Bundler module resolution.
 
-Currently a disabled test: @ts-expect-error documents known failures caused by a
-TypeScript type resolution bug that ignores tree's exports type override and generates
-invalid import paths. Bundler module resolution in the exporter (TS 5.6+) exacerbates the issue. Fixing tree's
-exports (giving each tier its own JS entrypoint) will resolve these — at which point the
-@ts-expect-error lines should be removed.
+Tree's exports now give each tier its own JS entrypoint, so TypeScript correctly
```
jason-ha: nit: don't say "now". The comment exists as-is without the context of what came before.
"now" is okay in a description.

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.json
jason-ha, line 1: Just delete the consumer project. Test build only needs one build now.

examples/utils/import-testing/src/test/tsconfig.json
```diff
@@ -19,9 +19,6 @@
 		"incremental": false,
 	},
 	"include": ["./**/*"],
-	// Temporary: cross-package importers excluded because skipLibCheck: false
-	// checks the cross-package .d.ts output which has the known broken import paths.
-	// Remove this exclusion when tree's exports are fixed.
 	"exclude": ["./crossPackage/**/*"],
```
jason-ha: It's now longer temporary?

packages/dds/tree/package.json
```diff
@@ -232,6 +232,18 @@
...
+			],
+			"build:api-reports:legacy": [
+				"build:esnext"
+			],
+			"ci:build:api-reports:current": [
+				"build:esnext"
+			],
+			"ci:build:api-reports:legacy": [
+				"build:esnext"
+			],
```
jason-ha: These should be in fluidBuild.config.cjs with the similar changes.

jason-ha (approved):

packages/dds/tree/package.json
jason-ha: (https://github.com/microsoft/FluidFramework/pull/26522#issuecomment-3988238994)

packages/dds/tree/api-report/tree.alpha.api.md
jason-ha, line 1: Shouldn't be any diffs here - rebuild and check.

WillieHabi: Thanks restored

jason-ha (approved):

jason-ha (approved):

packages/dds/tree/scripts/generate-entrypoint-sources.sh
```diff
@@ -0,0 +1,29 @@
...
+# therefore a build should always be run after this script to restore validity
+# of .d.ts files.
+
+# 0. Make sure lib/entrypoints exists (a limitation of "flub generate entrypoints")
+
+mkdir -p ./lib/entrypoints
+
+# 1. Generate entrypoint .d.ts files in lib/entrypoints.
+
+flub generate entrypoints --outFileLegacyBeta legacy --outDir ./lib/entrypoints
```
CraigMacomber: this relies on a global install of flub. Should we use `pnpm flub` here to run the version from our dev deps?

packages/dds/tree/scripts/generate-entrypoint-sources.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+# Generate entrypoint .d.ts files in lib/entrypoints, then transform them into
+# source .ts files under src/entrypoints.
```
CraigMacomber, line 6: Can we have some centralized documentation about these entry point files somewhere, and link it from here?

I think this is a workaround for a tsc bug: I'd love the centralized docs to link to an upstream issue in tsc, explain how it applies to us, and why our workaround is specifically part of the tree package.

packages/dds/tree/scripts/generate-entrypoint-sources.sh
WillieHabi: Changed to `pnpm flub`. I did get a ' flub: command not found' before so this seems more reliable

packages/dds/tree/src/entrypoints/internal.ts
```diff
@@ -0,0 +1,13 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+// This file provides one module of indirection from index.ts to control
+// TypeScript import spec generation for expression declarations. Importantly,
```
CraigMacomber: It would be good to clarify if this is talking about how typescript processes this package when compiling it, or how it impacts other packages which reference types from this package.

packages/dds/tree/src/index.ts
```diff
@@ -3,6 +3,10 @@
  * Licensed under the MIT License.
  */
 
+// For the time being, if exports are changed, additional files under entrypoints
+// directory probably need updated. They can be manually adjusted or run
+// generate:entrypoint-sources script.
```
CraigMacomber: Those files say:
> THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.

So I don't think we should say they can be "manually adjusted".

```suggestion
// For the time being, if exports are changed, additional files under entrypoints need updated.
// Run `pnpm generate:entrypoint-sources` to update them.
```

packages/dds/tree/scripts/generate-entrypoint-sources.sh
jason-ha: It would be better to prefix with `pnpm`, but it does not rely on a global... The script is run under package.json and will use package bin. So current package flub will be used.
But if you did run it manually and not from package.json script, then indeed it would look for a global version.

packages/dds/tree/src/index.ts
CraigMacomber: If someone forgets to do this step, will the build fail on CI or will get let them build and merge with out of date files?

If this can get through CI I'd like to see a TODO and a tracking work item reference here tracking integrating this into the build properly.

If this will fail on CI (presumably due to being generated as part of the build and failing the no unchanged files check), I'd like the comment above clarifies to not that running a build will also be sufficient.

packages/dds/tree/src/index.ts
```diff
@@ -369,6 +373,12 @@ import * as InternalTypes from "./internalTypes.js";
 /**
  * Contains types used by the API, but which serve mechanical purposes and do not represent semantic concepts.
  * They are used internally to implement API aspects, but are not intended for use by external consumers.
+ *
+ * @public
+ * @system
+ *
+ * @privateRemarks These TS-Docs are not recognized by API-Extractor, but the
```
CraigMacomber, line 379: What is "These TS-Docs" referring to? The public tag here? The description above it? Everything inside of InternalTypes? A more specific noun phrase would be helpful here.

packages/dds/tree/src/entrypoints/internal.ts
WillieHabi: specified '...in consumer packages' .d.ts output'

packages/dds/tree/src/index.ts
WillieHabi: It's the system tag right? I can change that here @jason-ha

jason-ha: Exactly "these TS-Docs" - those that existing in this block. Can be clarified.

packages/dds/tree/src/index.ts
jason-ha: It depends on what isn't done. If the API surface changes are tested, then CI will definitely fail. If someone adds a new API and doesn't test it via API surface, then probably not. Of course, the *.api.md files won't change either. So, I feel like it should be pretty obvious.
@WillieHabi, you had to refresh these since time had passed. What was your experience?

WillieHabi: Yeah so when I rebased onto latest main, the entrypoints were stale so src/entrypoints/alpha.ts was missing types like ObjectNodeSchemaWorkaround that had been added to the alpha API surface since the original commit.

The cross-package schema provider build failed because TypeScript couldn't find those types through the exports map. Running `pnpm generate:entrypoint-sources` and rebuilding fixed it immediately. So seems like CI would catch this, the build itself fails, not just the api-extractor lint.

packages/dds/tree/src/index.ts
CraigMacomber: By this block, do you mean the privateRemarks block? I assume not?

Seems like this is intending to explain that https://fluidframework.com/docs/api/fluid-framework/internaltypes-namespace/ doesn't get the description or tags included above

```suggestion
 * @privateRemarks
 * Doc comments on module exports are not handled by API-Extractor and thus do not properly appear in the generated API documentation.
 * The release tags are handled flub entrypoint generation.
```

jason-ha: @WillieHabi, it is the whole thing - entire block. None of the comments or tags in here are fully used. These comments do not apply to generated documentation as api-extractor doesn't recognize the comment. See https://fluidframework.com/docs/api/fluid-framework/internaltypes-namespace/. The only thing captured is that this namespace exists.

jason-ha: It is more than generated documentation. IntelliSense also doesn't pick it up.
The only thing that pays attention to it is our custom entrypoint generation. Adding `@public` makes it clear to our tooling that this isn't forgotten (suppresses a warning) and to us that it is effectively `@public`. For api-extractor it is undeclared and default public - no warning.

jason-ha (approved):

packages/dds/tree/scripts/generate-entrypoint-sources.sh
jason-ha: Will follow up with more post merge.

CraigMacomber (approved):
Assuming the people driving this will continue to refine the build process and ensure the process for keeping these new files up to date is ergonomic, this seems good from the tree perspective.

PR #26525, opened 2026-02-24
Tweak docs on processMessagesCore

CraigMacomber:
## Description

Adjust the docs on processMessagesCore to make more sense to callers (not just be for implementers),
and to make more sense for implementers since the method is abstract and must be implemented.
Docs now provide a bit of higher-level context to how these fits into fluid (that is how ops are sequenced) and note some TODOs for future work.

## Reviewer Guidance


packages/dds/shared-object-base/src/sharedObject.ts
```diff
@@ -398,26 +398,31 @@ export abstract class SharedObjectCore<
...
 	 *   - They are all based on the same reference sequence number.
 	 *   - They are not interleaved with messages from other clients.
 	 * - They are not interleaved with messages from other DDS in the container.
-	 * Derived classes should override this if they need to do custom processing on a 'bunch' of remote messages.
-	 * @param messageCollection - The collection of messages to process.
-	 *
+	 * @param messageCollection - The 'bunch' of sequenced ops to apply to this shared object.
+	 * @privateRemarks
+	 * TODO: We should have cental linkable documentation (and ideally types) for the special cases of `IRuntimeMessageCollection`
+	 * that are known to be a valid "grouped batch" or a "bunch".
```
CraigMacomber:
```suggestion
	 * TODO: We should have central linkable documentation (and ideally more specific types) for the
	 * "grouped batch" and "bunch" cases of `IRuntimeMessageCollection`.
```

packages/dds/shared-object-base/src/sharedObject.ts
```diff
@@ -398,26 +398,31 @@ export abstract class SharedObjectCore<
 
 	/* eslint-disable jsdoc/check-indentation */
 	/**
-	 * Derived classes must override this to do custom processing on a 'bunch' of remote messages.
+	 * Apply a 'bunch' of sequenced ops to this shared object.
 	 * @remarks
+	 * These ops have been sequenced by the service and are now finalized with respect to their ordering.
```
CraigMacomber:
```suggestion
	 * These ops have been sequenced by the service and now have a finalized ordering.
```

packages/dds/shared-object-base/src/sharedObject.ts
```diff
@@ -398,26 +398,31 @@ export abstract class SharedObjectCore<
...
 	 *   - They are not interleaved with messages from other clients.
 	 * - They are not interleaved with messages from other DDS in the container.
-	 * Derived classes should override this if they need to do custom processing on a 'bunch' of remote messages.
-	 * @param messageCollection - The collection of messages to process.
-	 *
+	 * @param messageCollection - The 'bunch' of sequenced ops to apply to this shared object.
+	 * @privateRemarks
+	 * TODO: We should have central linkable documentation (and ideally more specific types) for the
+	 * "grouped batch" and "bunch" cases of `IRuntimeMessageCollection`.
+	 * TODO:Performance: AB#59783: Allowing this to process more messages at once (more than the current definition of 'bunch') could improve performance of clients which fall behind,
```
agarwal-navin, line 409: This TODO should not be added here. I can't say for sure if the description in AB#59783 can actually be integrated into bunching or is a completely separate feature. The options mentioned there looks fine in theory but in practice, they may not work or may not be worth doing because it likely covers a very small set of scenarios.
So, I would not add this TODO here until at least the general direction of that work item is more defined. Adding it before that gives the impression that there is more coming which may not be the case.

packages/dds/shared-object-base/src/sharedObject.ts
```diff
@@ -398,26 +398,31 @@ export abstract class SharedObjectCore<
...
 	 *   - They are all from the same client.
 	 *   - They are all based on the same reference sequence number.
 	 *   - They are not interleaved with messages from other clients.
 	 * - They are not interleaved with messages from other DDS in the container.
-	 * Derived classes should override this if they need to do custom processing on a 'bunch' of remote messages.
-	 * @param messageCollection - The collection of messages to process.
-	 *
+	 * @param messageCollection - The 'bunch' of sequenced ops to apply to this shared object.
+	 * @privateRemarks
+	 * TODO: We should have central linkable documentation (and ideally more specific types) for the
```
agarwal-navin: Consider moving this comment block to `IRuntimeMessageCollection` and adding a link to that here (and eventually in other places). That way this TODO can be taken care of in this PR itself.

packages/dds/shared-object-base/src/sharedObject.ts
CraigMacomber: All possible ways to implement that work item involve changing what is passed in here (regardless of if its larger bunches or multiple bunches). It will deliver a large and important performance improvement generally (make slow clients much more able to keep up with lots of tree edits) which is being funded as part of a feature I own (text) where I have measured it would be a large win already. I have an active work item tracking implementing this, and plan to attempt it soon. If that attempt fails I will clean this up along with the work item itself.

CraigMacomber: For some context, I am making this PR because I was reading this code as the first steps of starting that implementation work.

CraigMacomber: Also, personally I think linking a work item for future work in impacting a part of the code is a good thing to do, even if that work item isn't scheduled and likely to happen soon like this one is. If someone wants to know if/when it will happen, they can check the work item to see when/if its scheduled. Linking such items helps prevent redundant work items or implementation and ensures if someone does implement this, they will know about the item and know to coordinate with its owner (if any) and adjust or close if when they are done.

packages/dds/shared-object-base/src/sharedObject.ts
CraigMacomber: Done.

packages/dds/shared-object-base/src/sharedObject.ts
CraigMacomber: Anyway, above is my explanation for why I think including this TODO makes sense. If you still think it should be removed, just say so and I will do it: it's better to get this in without it (and discuss that issue independently as another change) than not merge this.

agarwal-navin (approved):

PR #26526, opened 2026-02-24
Add `override` to processMessagesCore implementations, and remove inheritDoc

CraigMacomber:
## Description

Add `override` to processMessagesCore implementations, and remove inheritDoc which breaks intelisense and seems to offer little to no value.

## Reviewer Guidance

I think https://github.com/microsoft/FluidFramework/wiki/TSDoc-Guidelines#override might need to be corrected/updated. I made this PR based on what I think our standards should be, but its draft for now as we don't have a clear policy.

### DIFF
--- a/packages/dds/task-manager/src/taskManager.ts
+++ b/packages/dds/task-manager/src/taskManager.ts
@@ -697,10 +697,7 @@ export class TaskManagerClass
 		}
 	}
 
-	/**
-	 * {@inheritDoc @fluidframework/shared-object-base#SharedObject.processMessagesCore}
-	 */
-	protected processMessagesCore(messagesCollection: IRuntimeMessageCollection): void {
+	protected override processMessagesCore(messagesCollection: IRuntimeMessageCollection): void {
 		const { envelope, local, messagesContent } = messagesCollection;
 		for (const messageContent of messagesContent) {
 			this.processMessage(envelope, messageContent, local);
--- a/experimental/PropertyDDS/packages/property-dds/src/propertyTree.ts
+++ b/experimental/PropertyDDS/packages/property-dds/src/propertyTree.ts
@@ -364,10 +364,7 @@ export class SharedPropertyTree extends SharedObject {
 		this._root._reportDirtinessToView();
 	}
 
-	/**
-	 * {@inheritDoc @fluidframework/shared-object-base#SharedObject.processMessagesCore}
-	 */
-	protected processMessagesCore(messagesCollection: IRuntimeMessageCollection): void {
+	protected override processMessagesCore(messagesCollection: IRuntimeMessageCollection): void {
 		const { envelope, messagesContent } = messagesCollection;
 		for (const messageContent of messagesContent) {
 			this.processMessage(envelope, messageContent.contents);
--- a/experimental/dds/ot/ot/src/ot.ts
+++ b/experimental/dds/ot/ot/src/ot.ts
@@ -107,10 +107,7 @@ export abstract class SharedOT<TState, TOp> extends SharedObject {
 
 	protected onDisconnect(): void {}
 
-	/**
-	 * {@inheritDoc @fluidframework/shared-object-base#SharedObject.processMessagesCore}
-	 */
-	protected processMessagesCore(messagesCollection: IRuntimeMessageCollection): void {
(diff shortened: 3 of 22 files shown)


Josmithr (approved):

packages/dds/task-manager/src/taskManager.ts
```diff
@@ -697,10 +697,7 @@ export class TaskManagerClass
 		}
 	}
 
-	/**
-	 * {@inheritDoc @fluidframework/shared-object-base#SharedObject.processMessagesCore}
-	 */
-	protected processMessagesCore(messagesCollection: IRuntimeMessageCollection): void {
+	protected override processMessagesCore(messagesCollection: IRuntimeMessageCollection): void {
```
jason-ha, line 700: This is for `TaskManagerClass` which is exported as `@legacy+beta` and therefore until tooling is fixed inheritDoc is needed to get doc generation, at per last discussion of this 3/18/2025.
It would be nice if the api report noted that this method is now undocumented OR if we checked in the generated doc json file that shows what will be relayed for documentation generation so that we could see this delta.
All other cases for exported classes need similar keeping. (Also, better if we just stop exporting classes - especially sealed ones.)
cc: @Josmithr

Josmithr: I don't understand. Release tag levels have no impact on whether or not something gets API documentation. API-Extractor does not automatically inherit documentation in _any_ circumstances. So, this change does have the effect of removing explicit method documentation from the resulting API docs, but this is a trade-off we've made elsewhere (prioritizing intellisense over API docs). And it doesn't have anything to do with the release tagging.

Also note that type-level inheritance _is_ captured by the API docs, so a reader can navigate to the base declaration (in this case, `SharedObjectBase`) to view the documentation. It's certainly a worse experience overall, but we are unfortunately required to make these trade-offs until one of the following:
1. Intellisense actually supports TSDoc (rather than JSDoc)
2. API-Extractor is updated to persist inheritance information more granularly 
3. We do hacky things in `api-markdown-documenter` to try to re-derive this information dynamically.

CraigMacomber: Can we get https://github.com/microsoft/FluidFramework/wiki/TSDoc-Guidelines#inheritDoc updated indicate when inheritDoc should be omitted to let the inheritance happen implicitly? Sounds like maybe we need it for package exported non-internal APIs? Or maybe just those APIs when crossing between packages?

jason-ha: @Josmithr, release tags are relevant because that indicates a non-internal use. We don't generate docs for `@internal` and will just rely on IntelliSense (that is happier without `@inheritDoc`).

Josmithr: @jason-ha What `@internal` types are involved here?

Josmithr: > Can we get https://github.com/microsoft/FluidFramework/wiki/TSDoc-Guidelines#inheritDoc updated indicate when inheritDoc should be omitted to let the inheritance happen implicitly? Sounds like maybe we need it for package exported non-internal APIs? Or maybe just those APIs when crossing between packages?

I'll file an item for myself to update the guidance for that and `@override`.

jason-ha: I don't know. I really only looked at this class. I do know that we have inheritDoc in places where it is detrimental.

CraigMacomber: > if the api report noted that this method is now undocumented

It does. I just failed to include the API report updates. That has been corrected.

That said, Tyler is working to make all these classes internal currently, so they will soon be out of the API reports :). We do not actually intend to support any non-internal implanters or callers of this API, but a few soon to be gone exports are keeping it in the legacy API at the moment.

jason-ha: Maybe we don't have API reports for task-manager. I don't see a now undocumented note for this in an api.md file.

jason-ha: Also, I don't have an issue with not having a comment for this. Just want to make sure we know what we are doing.

Josmithr (approved):

jzaffiro (approved):

PR #26529, opened 2026-02-24
Specify sources directory explicitly in ADO pipelines

ChumpChief:
[AB#60000](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/60000)

ADO Pipelines [changes the location of checked-out repos](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/multi-repo-checkout?view=azure-devops#checkout-path) depending on whether a single vs. multiple repos are checked out.  As a result, checking out a second repo would break most of our pipelines.  To fix the broken bundle size telemetry, we want to !conditionally! check out a second repo in "internal" runs, which also exacerbates this quirk.

Fix here is to be explicit about checkout locations for the repo (see include-vars.yml) such that it will be consistent no matter how many repos are checked out, and then fix up a bunch of places that make assumptions about install location.  Broadly, these are:

1. Stop using the Build.SourcesDirectory variable, which exhibits (a variant of) the quirk - it can follow the new explicit checkout location for a single repo but will break again when a second repo is checked out.
2. "checkout" steps get a path relative to Pipeline.Workspace (a rooted/absolute path is not an option, this causes an error).  This value is provided as a variable in include-vars.yml
3. Bash task's workingDirectory - when given a relative path, this is implicitly evaluated as relative to the SourcesDirectory.  Fix is to pass a rooted path instead.
    * Similar fixes for tasks CopyFiles, Npm, and AzureCLI
4. buildDirectory parameter - this remains a relative path, but relative to the Pipeline.Workspace variable (rather than relative to the SourcesDirectory).
    * Also fix a few bugs where the path was being double-applied.  These weren't getting caught when the buildDirectory was "." (e.g. Archive Build Output Content's WORKING_DIRECTORY + workingDirectory)
5. Clarify paths that aren't repo-agnostic as being relative to the FluidFramework directory specifically - several places were using paths like `$(Build.SourcesDirectory)/scripts/...`, these are made to point more explicitly to the FF repo checkout rather than just the SourcesDirectory or buildDirectory.
6. Add explicit checkout steps to stages that didn't include a checkout step - when omitted, a self-checkout is implicitly added to the stage using the quirky default location, and several of our pipelines were depending on this implicit checkout.  Adding an explicit checkout step lets us specify the location.

It is difficult to confirm that I've updated all permutations and run options that are touched by this change, but I've manually run a wide variety in various configurations - I'll be sure to remain available to monitor following merge in case I've missed anything.

### DIFF
--- a/tools/pipelines/templates/build-npm-client-package.yml
+++ b/tools/pipelines/templates/build-npm-client-package.yml
@@ -371,11 +371,11 @@ extends:
               - task: Bash@3
                 displayName: Archive Build Output Content
                 env:
-                  WORKING_DIRECTORY: '${{ parameters.buildDirectory }}'
+                  WORKING_DIRECTORY: '.'
                 inputs:
                   targetType: filePath
                   workingDirectory: '$(Pipeline.Workspace)/${{ parameters.buildDirectory }}'
-                  filePath: $(Build.SourcesDirectory)/scripts/pack-build-output.sh
+                  filePath: $(Pipeline.Workspace)/${{ parameters.buildDirectory }}/scripts/pack-build-output.sh
 
               - task: CopyFiles@2
                 displayName: Copy build_output_archive to artifact staging directory
--- a/tools/pipelines/templates/build-npm-client-package.yml
+++ b/tools/pipelines/templates/build-npm-client-package.yml
@@ -812,7 +812,7 @@ extends:
                   displayName: 'npm run test:copyresults'
                   inputs:
                     command: custom
-                    workingDir: '${{ parameters.buildDirectory }}'
+                    workingDir: '$(Pipeline.Workspace)/${{ parameters.buildDirectory }}'
                     customCommand: 'run test:copyresults'
 
                 - ${{ if contains(test.name, 'tinylicious') }}:
--- a/tools/pipelines/templates/build-npm-client-package.yml
+++ b/tools/pipelines/templates/build-npm-client-package.yml
@@ -692,13 +692,13 @@ extends:
                 displayName: Publish Code Coverage
                 condition: and(succeededOrFailed(), eq(variables['ReportDirExists'], 'true'))
                 inputs:
-                  summaryFileLocation: ${{ parameters.buildDirectory }}/nyc/report/cobertura-coverage-patched.xml
+                  summaryFileLocation: $(Pipeline.Workspace)/${{ parameters.buildDirectory }}/nyc/report/cobertura-coverage-patched.xml
                   failIfCoverageEmpty: true
               - task: CopyFiles@2
                 displayName: Copy code coverage report to artifact staging directory
                 condition: and(succeededOrFailed(), eq(variables['ReportDirExists'], 'true'))
                 inputs:
-                  sourceFolder: '${{ parameters.buildDirectory }}/nyc/report'
+                  sourceFolder: $(Pipeline.Workspace)/${{ parameters.buildDirectory }}/nyc/report
                   targetFolder: $(Build.ArtifactStagingDirectory)/codeCoverageAnalysis
               - task: Bash@3
                 displayName: Report Code Coverage Comparison
--- a/tools/pipelines/templates/build-npm-client-package.yml
+++ b/tools/pipelines/templates/build-npm-client-package.yml
@@ -245,7 +245,7 @@ extends:
                 displayName: Parameters
                 inputs:
                   targetType: inline
-                  workingDirectory: '${{ parameters.buildDirectory }}'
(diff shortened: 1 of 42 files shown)


tools/pipelines/templates/include-test-real-service.yml
```diff
@@ -236,6 +236,7 @@ stages:
 
         - template: /tools/pipelines/templates/include-install-pnpm.yml@self
           parameters:
+            # TODO: Make sure this is a path relative to $(Pipeline.Workspace)
```
Abe27342: dangling TODO which should be checked before merge :)

tools/pipelines/templates/include-vars.yml
```diff
@@ -29,6 +29,19 @@ variables:
...
 - name: skipComponentGovernanceDetection
   value: true
+
+# ADO changes the value of Build.SourcesDirectory depending on whether 1 vs. many repositories are checked out.
+# Using 's' is consistent with the behavior when multiple repositories are checked out and allows for more
+# consistent build scripts.
+# A relative path is required in some places (like "checkout" steps). To get the absolute path, concatenate it
+# with the predefined variable "Pipeline.Workspace", e.g.: '$(Pipeline.Workspace)/${{ variables.FluidFrameworkDirectory }}'
+- name: consistentSourcesDirectory
+  value: 's'
```
Abe27342, line 40: maybe worth pointing out this constant is documented on https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml under Build.SourcesDirectory (it looks like an implementation detail we should not rely on otherwise, but AFAICT seems relatively safe, even if not ideal)

Abe27342: or linking the docs you put in the PR description. That might be a better explanation of the rationale.

tools/pipelines/templates/build-npm-client-package.yml
```diff
@@ -364,22 +367,22 @@ extends:
...
                   command: custom
-                  workingDir: '${{ parameters.buildDirectory }}'
+                  workingDir: '$(Pipeline.Workspace)/${{ parameters.buildDirectory }}'
                   customCommand: 'run webpack'
 
               - task: Bash@3
                 displayName: Archive Build Output Content
                 env:
-                  WORKING_DIRECTORY: '${{ parameters.buildDirectory }}'
+                  WORKING_DIRECTORY: '.'
```
alexvy86: I'd suggest giving this an absolute path, since the `sh` script logs it and it'd be more useful to see the actual path than just a period.

alexvy86 (commented):
Overall I agree with the direction. Just one thing:

> buildDirectory parameter - this remains a relative path, but relative to the Pipeline.Workspace variable (rather than relative to the SourcesDirectory).

We're not using it for the `checkout` path which you mention cannot be an absolute one, so would it be better to make `buildDirectory` absolute everywhere? We could even have small validation steps as the first thing in any template that has such a parameter to ensure that whatever is passed is an absolute path. Seems easier to enforce that invariant "at the top" instead of having to prepend `$(Pipeline.Workspace)` everywhere we use it.

ChumpChief:
> We're not using it for the `checkout` path which you mention cannot be an absolute one, so would it be better to make `buildDirectory` absolute everywhere? We could even have small validation steps as the first thing in any template that has such a parameter to ensure that whatever is passed is an absolute path. Seems easier to enforce that invariant "at the top" instead of having to prepend `$(Pipeline.Workspace)` everywhere we use it.

That's a good idea, I'll take a look.  I think I started off in this direction as an attempt to minimize the delta to current behavior, but agree that seems preferable.

ChumpChief:
> That's a good idea, I'll take a look. I think I started off in this direction as an attempt to minimize the delta to current behavior, but agree that seems preferable.

@alexvy86 Relative path is still used in build-docker-service to locate the test coverage report relative to the ArtifactStagingDirectory - should be solvable but I'd probably prefer to stick with the relative paths for initial checkin at least.

tools/pipelines/templates/include-test-real-service.yml
ChumpChief: Fixed in latest

tools/pipelines/templates/include-vars.yml
ChumpChief: Fixed in latest

tools/pipelines/templates/build-npm-client-package.yml
ChumpChief: Fixed in latest

alexvy86 (approved):
LGTM as far as one can tell with pipeline changes just by looking at them.

PR #26530, opened 2026-02-24
Recommend snapshotSchemaCompatibility

CraigMacomber:
## Description

Now that snapshotSchemaCompatibility is beta, it shouldn't be renamed and links to it should be stable.
Since it is unchanged since last released (as alpha) the links to it are already valid, so we can add them to docs.

Use these links to recommend snapshotSchemaCompatibility in more places.

Also included are a few related docs tweaks noted while working on this.

## Reviewer Guidance


Josmithr (approved):
Looks good!

Nit: there is a small typo in the PR description ("Usse" instead of "Use"), FYI

PR #26533, opened 2026-02-24
Simplify TestClientLogger addNewLogLine

CraigMacomber:
## Description

I profiled the merge-tree tests and found we spend about 28% of the time in getSegString so I updated the comment about that.

I also noticed a lot of time in addNewLogLine. I did simplify it in a way that might help, but in practice did not make a significant difference to the runtime, but seems work keeping.

## Reviewer Guidance

### DIFF
--- a/packages/dds/merge-tree/src/test/testClientLogger.ts
+++ b/packages/dds/merge-tree/src/test/testClientLogger.ts
@@ -208,12 +208,12 @@ export class TestClientLogger {
 
 	private addNewLogLine(): void {
 		if (this.incrementalLog) {
-			while (this.roundLogLines.length > 0) {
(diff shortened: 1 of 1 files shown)


packages/dds/merge-tree/src/test/testClientLogger.ts
```diff
@@ -208,12 +208,12 @@ export class TestClientLogger {
 
 	private addNewLogLine(): void {
 		if (this.incrementalLog) {
```
anthony-murphy, line 210: most the time we don't run in incremental mode, which makes sense why this wouldn't have impact.

anthony-murphy (approved):

PR #26534, opened 2026-02-25
Port FieldBatchCodec and ForestCodec to ClientVersionDispatchingCodecBuilder

CraigMacomber:
## Description

Port FieldBatchCodec and ForestCodec to ClientVersionDispatchingCodecBuilder.

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/codec/versioned/codec.ts
+++ b/packages/dds/tree/src/codec/versioned/codec.ts
@@ -391,9 +391,14 @@ export class ClientVersionDispatchingCodecBuilder<
 	/**
 	 * Produce a single codec which can read any supported format, and writes a version selected based on the provided options.
 	 */
-	public build(
-		options: TBuildOptions,
-	): IJsonCodec<TDecoded, JsonCompatibleReadOnly, JsonCompatibleReadOnly, TContext> {
+	public build(options: TBuildOptions): IJsonCodec<
+		TDecoded,
+		JsonCompatibleReadOnly,
+		JsonCompatibleReadOnly,
+		TContext
+	> & {
+		readonly writeVersion: TFormatVersion;
+	} {
 		const applied = this.applyOptions(options);
 		const writeVersion = getWriteVersion(this.name, options, applied);
 		const fromFormatVersion = new Map<
@@ -415,6 +420,7 @@ The client which encoded this data likely specified an "minVersionForCollab" val
 				}
 				return codec.codec.decode(data, context);
 			},
+			writeVersion: writeVersion.formatVersion,
 		};
 	}
 
--- a/packages/dds/tree/src/codec/versioned/codec.ts
+++ b/packages/dds/tree/src/codec/versioned/codec.ts
@@ -397,6 +397,11 @@ export class ClientVersionDispatchingCodecBuilder<
 		JsonCompatibleReadOnly,
 		TContext
 	> & {
+		/**
+		 * The format version which this codec writes.
+		 * @remarks
+		 * This is selected based on the provided {@link CodecWriteOptions}.
+		 */
 		readonly writeVersion: TFormatVersion;
 	} {
 		const applied = this.applyOptions(options);
--- a/packages/dds/tree/src/feature-libraries/chunked-forest/codec/codecs.ts
+++ b/packages/dds/tree/src/feature-libraries/chunked-forest/codec/codecs.ts
@@ -116,7 +116,7 @@ export type FieldBatchCodec = ReturnType<typeof fieldBatchCodecBuilder.build>;
  * Creates the encode/decode functions for a specific FieldBatch format version.
  */
 function makeFieldBatchCodecForVersion(
-	writeVersion: FieldBatchFormatVersion,
+	version: FieldBatchFormatVersion,
 	uncompressedEncodeFn: (batch: FieldBatch) => EncodedFieldBatch,
 	schemaCompressedEncodeFn: (
 		schema: StoredSchemaCollection,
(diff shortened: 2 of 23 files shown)


packages/dds/tree/src/feature-libraries/forest-summary/codec.ts
```diff
@@ -53,40 +57,75 @@ export function clientVersionToForestFormatVersion(
...
-			const keys: FieldKey[] = [];
-			const fields: ITreeCursorSynchronous[] = [];
-			for (const [key, value] of data) {
-				keys.push(key);
-				fields.push(value);
-			}
-			return { keys, fields: inner.encode(fields, context), version: writeVersion };
+/**
+ * Codec builder for forest summarizer codecs.
+ * Uses ClientVersionDispatchingCodecBuilder to dispatch to the appropriate version based on minVersionForCollab.
```
Josmithr: Nit
```suggestion
 * @remarks Uses {@link ClientVersionDispatchingCodecBuilder} to dispatch to the appropriate version based on `minVersionForCollab`.
```

packages/dds/tree/src/feature-libraries/forest-summary/codec.ts
```diff
@@ -9,33 +9,37 @@ import {
...
-	FieldSet,
-	Format,
-	JsonCompatibleReadOnly,
-	FieldBatchEncodingContext
->;
+export type ForestCodec = ReturnType<typeof forestCodecBuilder.build>;
+
+/**
+ * Options for building the forest summarizer codec.
+ * Extends CodecWriteOptions with the required fieldBatchCodec dependency.
```
Josmithr: Nit
```suggestion
 * @remarks Extends {@link CodecWriteOptions} with the required `fieldBatchCodec` dependency.
```

packages/dds/tree/src/feature-libraries/forest-summary/codec.ts
```diff
@@ -53,40 +57,75 @@ export function clientVersionToForestFormatVersion(
...
+					const out: Map<FieldKey, ITreeCursorSynchronous> = new Map();
+					const fields = options.fieldBatchCodec.decode(data.fields, context);
+					assert(data.keys.length === fields.length, 0x891 /* mismatched lengths */);
+					for (const [index, field] of fields.entries()) {
+						out.set(data.keys[index] ?? oob(), field);
+					}
+					return out;
+				},
+				schema: formatSchema,
+			};
```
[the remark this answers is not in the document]
Josmithr: @CraigMacomber I had more or less the same thought here. Or was the duplication here intentional to avoid future changes incorrectly affecting other (older) codec versions? If so, might be worth a quick comment.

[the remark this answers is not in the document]
CraigMacomber: No, this is bad. Looks like back when I had Claud try and make this change (and I had to split a ton of PRs out of it to make it reasonable), I only fixed the duplication in the other codec and not this one.

This isn't even valid as is since it duplicated a tagged assert: we still need to make CI catch that however.

packages/dds/tree/src/codec/versioned/codec.ts
```diff
@@ -391,9 +391,14 @@ export class ClientVersionDispatchingCodecBuilder<
...
-	public build(
-		options: TBuildOptions,
-	): IJsonCodec<TDecoded, JsonCompatibleReadOnly, JsonCompatibleReadOnly, TContext> {
+	public build(options: TBuildOptions): IJsonCodec<
+		TDecoded,
+		JsonCompatibleReadOnly,
+		JsonCompatibleReadOnly,
+		TContext
+	> & {
+		readonly writeVersion: TFormatVersion;
```
Josmithr, line 405: Nit: docs?

packages/dds/tree/src/feature-libraries/chunked-forest/codec/codecs.ts
```diff
@@ -114,63 +110,24 @@ export interface FieldBatchEncodingContext {
...
-		}
-	}
-
-	// Both the encode and decode logic here support both v1 and v2, as does `validVersions` and `EncodedFieldBatch`.
-	// This makes this use of makeVersionedValidatedCodec atypical as it is a single call being used to make a codec that supports all versions,
-	// instead of one call per version, then using another utility to select between them based on version.
-	return makeVersionedValidatedCodec(options, validVersions, EncodedFieldBatch, {
+function makeFieldBatchCodecForVersion(
+	writeVersion: FieldBatchFormatVersion,
+	uncompressedEncodeFn: (batch: FieldBatch) => EncodedFieldBatch,
```
Josmithr, line 120: Nit: `makeForestSummarizerCodec` takes an `options` object and the `writeVersion`. Would it makes sense to do the same here and bucket the non-`writeVersion` params under a `FieldBatchCodecOptions` param? Would likely make the call sites a bit more legible since the different functions would be labeled.

Josmithr: Might also make sense to bundle the `writeVersion` in the property bag, but then we'd probably want to update `makeForestSummarizerCodec` as well.

Josmithr (approved):

packages/dds/tree/src/feature-libraries/chunked-forest/codec/codecs.ts
CraigMacomber: The options object required in makeForestSummarizerCodec is provided to the build API by the user getting a codec instance (and likely will be removed soon, and can't contain the version as it has the options needed before selecting the version), where this one takes in nothing which is needed at that point and these are instead specific to the codec version and provided by the registration code below. Thus they are rather different cases and I don't think aligning them to the same pattern really makes sense.

A parameters object might be cleaner here, but with this not being file exported and only used by the two calls below I don't think its worth it.

Josmithr (approved):

PR #26543, opened 2026-02-25
Fix explicit sources directory breaks

ChumpChief:
Fixes:
* deploy-website
    * Didn't include the variables from any other location, this change adds them directly
    * Two instances of specifying buildDirectory as an absolute path when it should be relative
    * Missed an explicit path on one of the checkout steps
* publish package step (multiple pipelines)
    * Needs a relative path for pnpm install (exact location shouldn't matter too much since caching is disabled)
* include-generate-notice-steps.yml
    * I had just missed updating this template.

### DIFF
--- a/tools/pipelines/deploy-website.yml
+++ b/tools/pipelines/deploy-website.yml
@@ -196,6 +196,7 @@ stages:
       displayName: Guardian tasks
       steps:
         - checkout: self
+          path: ${{ variables.FluidFrameworkDirectory }}
           submodules: false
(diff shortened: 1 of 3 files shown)


Abe27342 (approved):

ChumpChief:
I see NOTICE file generation seems to fail sporadically even before my changes (looks like maybe overloaded service) so I don't plan to block on that for merge.

alexvy86 (approved):

PR #26548, opened 2026-02-25
Add logLevel property to telemetry events

MarioJGMsoft:
## Description

Added a new optional logLevel property to telemetry events to indicate their importance for diagnostics. This enables consumers to make sampling or filtering decisions based on the event's importance level.

The property can have three values:

- **verbose** — Chatty logs useful for debugging but likely not needed in production.
- **info** — Session information that could be omitted in some sessions if needed.
- **essential**— Critical diagnostic information that should always be transmitted.

Events without a `logLevel` value should be treated as **essential**.

## Reviewer Guidance

The goal is to eventually replace the existing LogLevel type with this new property. I'd appreciate feedback on the naming of the property and its values and whether the default behavior (treating missing logLevel as essential) makes sense.

[AB#58991](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/58991)


packages/runtime/container-runtime/src/summary/orderedClientElection.ts
```diff
@@ -688,6 +690,7 @@ export class OrderedClientElection
 		sequenceNumber: number,
 		forceSend: boolean = false,
 		reason?: string,
+		logLevel?: string,
```
jason-ha: not required?
If not specified, I think we'll emit undefined. Does undefined just get dropped?

jason-ha (commented):
We're replacing `LogLevel` with `string`? Any string is okay?
How are we documenting to our customers what they should be checking?

MarioJGMsoft:
> We're replacing `LogLevel` with `string`? Any string is okay? How are we documenting to our customers what they should be checking?

Originally I wanted to avoid the issue we had with the logLevel enum, so I decided to make it a simple string. But thinking more about it now, I could create an internal enum that can work as both documentation for what costumers should be checking and a way to always have a list of expected values.

I was thinking of adding a section in one of the .md files in the docs folder, but I wasn's sure where.

packages/runtime/container-runtime/src/summary/orderedClientElection.ts
MarioJGMsoft: My idea was to have undefined be treated as the `essential` log level, this means that if there is no `logLevel` property the event **shouldn't** be dropped.

This is because there are a lot of events in our codebase and I think it's better to have it so that we have more information diagnosing rather than missing some that wasn't properly set to `essential`.

.changeset/tricky-zoos-see.md
```diff
@@ -0,0 +1,18 @@
+---
+"@fluidframework/container-runtime": minor
+"@fluidframework/shared-object-base": minor
+"__section": feature
+---
+Add logLevel property to events
```
markfields:
```suggestion
Add logLevel property to logging events
```

.changeset/tricky-zoos-see.md
```diff
@@ -0,0 +1,18 @@
...
+
+Events now include an optional `logLevel` property that indicates their importance for diagnostics and enables consumers to make sampling or filtering decisions.
+
+There are currently three values that the property can have:
+1. 'verbose'
+Chatty logs useful for debugging but likely not to be sent over the wire in production.
+2. 'info'
+Information about the session. These logs could be ommitted in some sessions if needed.
+3. 'essential'
+Essential information about the operation of Fluid that should always be transmitted for diagnostic purposes.
```
markfields:
```suggestion
There are currently three supported values:
1. 'verbose'
Chatty logs useful for local debugging. They need not be collected in production.
2. 'info'
Information about the session. These logs could be omitted in some sessions if needed (e.g. to reduce overall telemetry volume).  If any are collected from a particular session, all should be.
3. 'essential'
Essential information about the operation of Fluid. It's recommended that they should always be collected even in production, for diagnostic purposes.
```

packages/dds/shared-object-base/src/sharedObject.ts
markfields, line 204: Let's make this class apply a default `logLevel: "info"` itself. (and don't bother setting below)

.changeset/tricky-zoos-see.md
```diff
@@ -0,0 +1,18 @@
+---
+"@fluidframework/container-runtime": minor
+"@fluidframework/shared-object-base": minor
+"__section": feature
+---
+Add logLevel property to events
+
+Events now include an optional `logLevel` property that indicates their importance for diagnostics and enables consumers to make sampling or filtering decisions.
+
+There are currently three values that the property can have:
```
markfields: Let's add some kind of helpers for this.

I wish would could do this, but it involves touching legacy-beta stuff (because `ITelemetryLoggerExt` is overexposed at the moment):

> Update `ITelemetryGenericEventExt` and related types to include `logLevel?: "verbose" | "info" | "essential"`
>
> Then whenever you use our internal loggers, you'd have type safety on what you pass for `logLevel`.
> And other helpers like SampledTelemetryHelper would benefit as well

But let's not for now.  Instead we can just add this to telemetry-utils logger.ts:

`export const LogLevel = { verbose: "verbose", info: "info", essential: "essential" }`

And then use that instead of the string literals everywhere.

packages/dds/shared-object-base/src/sharedObject.ts
MarioJGMsoft: Applied change

.changeset/tricky-zoos-see.md
MarioJGMsoft: Thanks for the feedback, I applied the changes

packages/runtime/container-runtime/src/summary/orderedClientElection.ts
Josmithr: Nit: for APIs not exposed to end users, I generally recommend making all arguments required (and allow undefined), rather than making them optional. This will help prevent usages from accidentally forgetting to provide a parameter it really wants to provide.

Josmithr: (Feel free to disregard, especially since it would require updating existing parameters here).

.changeset/tricky-zoos-see.md
MarioJGMsoft: Applied change

packages/runtime/container-runtime/src/summary/orderedClientElection.ts
MarioJGMsoft: After thinking about it for a bit, from what I could see the sendPerformanceEvent() funciton is used only in this file, and the events related to the client elected performance events might all be ommitable? (@kian-thompson  @markfields can confirm?)
So if that's the case then the parameter won't be needed at all and they can all default to being `info`

jason-ha: I think we should always have the property. It is best that our code set it - preferably it is explicitly chosen, but you can have a helper if we need to delay explicit decision on the level for some events. Then clean those up later. By the end we wanted to get all the events audited; so, they all will be visited.
Defaults are hard to manage for external consumers. We should just tell them the value.

jason-ha:
> > We're replacing `LogLevel` with `string`? Any string is okay? How are we documenting to our customers what they should be checking?
> 
> Originally I wanted to avoid the issue we had with the logLevel enum, so I decided to make it a simple string. But thinking more about it now, I could create an internal enum that can work as both documentation for what costumers should be checking and a way to always have a list of expected values.
> 
> I was thinking of adding a section in one of the .md files in the docs folder, but I wasn's sure where.

You can avoid the constrained enum issue by using the branded type pattern I previously shared.

With a string, how can a consumer know what to do with a string value they weren't previously told about?

packages/utils/telemetry-utils/src/test/sampledTelemetryHelper.spec.ts
```diff
@@ -91,7 +91,7 @@ describe("SampledTelemetryHelper", () => {
 		assert.strictEqual(event.count, 1);
 	});
 
-	it("includes aggregate properties when it should", () => {
+	it.only("includes aggregate properties when it should", () => {
```
jason-ha: Accidental .only commited

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -11,6 +11,8 @@ import {
 	type Tagged,
 	type TelemetryBaseEventPropertyType,
 } from "@fluidframework/core-interfaces";
+// eslint-disable-next-line import-x/no-internal-modules
+import type { BrandedType } from "@fluidframework/core-interfaces/legacy";
```
jason-ha: importing from `/legacy` but using as `@beta` below.
You can import from `/beta` (it doesn't exist currently) or `/internal`. (Or `BrandedType` can be promoted to `@public`.)

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -999,3 +1001,41 @@ export const tagCodeArtifacts = <
 					})
 		| (T[P] extends undefined ? undefined : never);
 } => tagData<TelemetryDataTag.CodeArtifact, T>(TelemetryDataTag.CodeArtifact, values);
+
+/**
+ * Numerical values that indicate the importance of a telemetry event for diagnostics,
+ * enabling consumers to make sampling or filtering decisions.
+ *
+ * If an event does not contain a `logLevelValue` value, it should be treated as `essential`.
```
jason-ha: The second paragraph should be in `@remarks` block.

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -999,3 +1001,41 @@ export const tagCodeArtifacts = <
...
+	 * if needed (e.g. to reduce overall telemetry volume). If any are collected
+	 * from a particular session, all should be.
+	 */
+	info: 15 as 15 & BrandedType<"LogLevelValue">,
+
+	/**
+	 * Essential information about the operation of Fluid. It is recommended that
+	 * these should always be collected, even in production, for diagnostic purposes.
+	 */
+	essential: 25 as 25 & BrandedType<"LogLevelValue">,
```
jason-ha: Looks like we are redoing the system altogether.
If so:
- can we spread these out more to provide room for injection?
- there is no utility distinguishing essential and what was previously called critical?

packages/utils/telemetry-utils/src/index.ts
```diff
@@ -69,6 +69,7 @@ export {
 	tagCodeArtifacts,
 	TelemetryDataTag,
 	type TelemetryEventPropertyTypes,
+	LogLevelValue,
```
jason-ha: maintain sorting

packages/utils/telemetry-utils/src/logger.ts
jason-ha: I think we want to avoid events not containing `logLevelValue`. When does this pertain?

packages/runtime/container-runtime/src/blobManager/blobManagerSnapSum.ts
```diff
@@ -7,7 +7,10 @@ import type { IContainerContext } from "@fluidframework/container-definitions/in
 import { readAndParse } from "@fluidframework/driver-utils/internal";
 import type { ISummaryTreeWithStats } from "@fluidframework/runtime-definitions/internal";
 import { SummaryTreeBuilder } from "@fluidframework/runtime-utils/internal";
-import type { ITelemetryLoggerExt } from "@fluidframework/telemetry-utils/internal";
+import {
+	LogLevelValue,
+	type ITelemetryLoggerExt,
+} from "@fluidframework/telemetry-utils/internal";
```
jason-ha: Keep type imports split from runtime

.changeset/tricky-zoos-see.md
```diff
@@ -0,0 +1,18 @@
+---
+"@fluidframework/container-runtime": minor
+"@fluidframework/shared-object-base": minor
+"__section": feature
+---
+Add logLevel property to logging events
+
+Events now include an optional `logLevel` property that indicates their importance for diagnostics and enables consumers to make sampling or filtering decisions.
```
jason-ha, line 8: If they do, shouldn't there be a type that says so?

packages/utils/telemetry-utils/src/logger.ts
jason-ha: Changeset says `logLevel`, but here it says `logLevelValue`.

.changeset/tricky-zoos-see.md
```diff
@@ -0,0 +1,18 @@
...
+
+There are currently three supported values:
```
jason-ha: I wouldn't say "supported". I think you can say "emitted". I think the intent is to support any number. Constrained values is what lead to trouble in the first place.

jason-ha (commented):
I'm still looking for more comprehensive docs on how to consume the changes.
We can't be saying look for one of three values since that is what is causing trouble now.

Tests or examples are needed to demonstrate proper use (consumption).

packages/runtime/container-runtime/src/blobManager/blobManagerSnapSum.ts
markfields: Is this just a style thing?  We should get a lint rule for it if it matters.

packages/utils/telemetry-utils/api-report/telemetry-utils.legacy.beta.api.md
```diff
@@ -55,6 +55,16 @@ export interface ITelemetryPerformanceEventExt extends ITelemetryGenericEventExt
 // @beta @legacy
 export type ITelemetryPropertiesExt = Record<string, TelemetryEventPropertyTypeExt | Tagged<TelemetryEventPropertyTypeExt>>;
 
+// @beta
+export const LogLevelValue: {
+    readonly verbose: 10 & BrandedType<"LogLevelValue">;
+    readonly info: 15 & BrandedType<"LogLevelValue">;
+    readonly essential: 25 & BrandedType<"LogLevelValue">;
```
markfields: Why not 10 20 30?

packages/utils/telemetry-utils/src/index.ts
markfields: Another place for a lint rule, you shouldn't have to think about this as a reviewer IMO

packages/utils/telemetry-utils/src/logger.ts
markfields: Could we have lint rules for this kind of thing...?

markfields (commented):
It seems like we're still trying to figure out the API and implementation of defaults.  Here's what I'm thinking:

Consumer API:

- Export `type LogLevelValue` as `@beta` (consumers need to reason over logs' levels)
- Update `ITelemetryBaseEvent` (which is `@public`) to explicitly include optional `logLevel` alongside `category` and `eventName` (optional to avoid breaking change, but in 3.0 could be made required)

Instrumentation API:

- Export `const  LogLevelValue` as `@internal` (Instrumentation is only internal, external code will not be setting log level)
- Update `ITelemetryGenericEventExt` and similar event interfaces to include optional `logLevel?: LogLevelValue`.
  - You can put a comment on it suggesting a default value (similar to `category`), but that's not enforceable
  - You can update `TelemetryLogger` base class to fill in `Essential` if it's missing. All our actual logger impl's extend this, so this is a good choke point for implementing the default.
  - For emphasis, you could even strengthen the type of `event` in `abstract send` on `TelemetryLogger` to require `logLevel`.  Note that this `send` function is the same as the one on `ITelemetryBaseLogger`, the `@public` API that can have `logLevel` required in 3.0 if we want.

Maybe paste this idea to Claude and ask it what it thinks to see if I missed something :P

packages/utils/telemetry-utils/src/test/sampledTelemetryHelper.spec.ts
MarioJGMsoft: Fixed

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: I wanted to add it to the changeset so that we could have time to go over all of the events, because in my opinion if this line isn't explicitly said, then that means that this PR should include the change to all of the events that we emit. From my understanding that's out of scope for this PR.

packages/utils/telemetry-utils/api-report/telemetry-utils.legacy.beta.api.md
MarioJGMsoft: I wanted to avoidclashing against the current declaration of `LogLevel` just in case it could break anything but thinking about it now, I think we should be okay with making them those values.

packages/utils/telemetry-utils/src/logger.ts
jason-ha: The lint rule is to only use public (no sub-path) or /internal.
The rule we don't have is that each suppression has to have a comment.

jason-ha:
> Consumer API:
> 
> * Export `type LogLevelValue` as `@beta` (consumers need to reason over logs' levels)
> * Update `ITelemetryBaseEvent` (which is `@public`) to explicitly include optional `logLevel` alongside `category` and `eventName` (optional to avoid breaking change, but in 3.0 could be made required)

`type LogLevelValue` would need to be `@public` to be used in `@public` `ITelemetryBaseEvent` even if optional. I do think that listing it is the right thing to do.
Internally I think we should use version of `ITelemetryBaseEvent` that has `logLevel` required.
 
> Instrumentation API:
> 
> * Export `const  LogLevelValue` as `@internal` (Instrumentation is only internal, external code will not be setting log level)

I think should be at least `@beta` (and probably `@public`) since we are going to provide guidance on how to use `logLevel` property.
Note that if we want levels at different release levels we can have `LogLevelValueAlpha` etc.

packages/runtime/container-runtime/src/blobManager/blobManagerSnapSum.ts
jason-ha: I would say functional style. There are several maintenance things that are better with keeping them separate. An example is making it more clear when there is a runtime dependency or just a compile time one. (And it matters.)
There are old discussions but never enough priority to make a change.
Basically all new code should prefer the split. Especially if the file already follows that pattern as this one does.

.changeset/tricky-zoos-see.md
MarioJGMsoft: Applied change

markfields:
@jason-ha - Sending an event through a logger is not or should not be a supported scenario outside the repo.

markfields:
> Internally I think we should use version of ITelemetryBaseEvent that has logLevel required.

That's the `Ext` types.  We should not introduce a new type layer between the `Base` types and `Ext` types.

jason-ha:
> @jason-ha - Sending an event through a logger is not or should not be a supported scenario outside the repo.

@markfields, maybe it shouldn't be but there aren't docs that say that and it happens. Pages does it. I suspect the way encapsulated model is built that it lends customers to call it.

markfields:
> there aren't docs that say that and it happens

@jason-ha - You're right\, I see how Bohemia's logger types all extend `ITelemetryBaseLogger` - Or they used to, now it's their copy of it!

Anyway, if people are going to interpret (read) `logLevel` they need the known values to be exported publicly anyway.

Let's just be sure that additions to that list are non-breaking. I'm not confident if that's the case with the current const object literal.

PS - Long-term, let's discuss marking `ITelemetryBaseLogger` as `@sealed`, to discourage folks from using our logging types as an SDK for building their own loggers.

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: Yes, I can spread out the numbers more, I can change it to be through the hundreds now.

And there's no utility in distinguishing between critical and essential since they both refer to events that should always be collected.

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: Added change

packages/runtime/container-runtime/src/blobManager/blobManagerSnapSum.ts
MarioJGMsoft: Applied change

.changeset/tricky-zoos-see.md
```diff
@@ -0,0 +1,18 @@
...
+---
+Add logLevel property to logging events
+
+Events now include an optional `logLevel` property that indicates their importance for diagnostics and enables consumers to make sampling or filtering decisions.
+
+There are currently three emitted values:
+1. 'verbose'
+Chatty logs useful for local debugging. They need not be collected in production.
+2. 'info'
+Information about the session. These logs could be omitted in some sessions if needed (e.g. to reduce overall telemetry volume).  If any are collected from a particular session, all should be.
```
ChumpChief, line 14: "Session" seems a little vague.  Can we define this more concretely?  E.g. is it per-logger-instance, per-unique user, per-document, etc.  Maybe even recommend the customer make the decision up-front as they create their logger to drive the point home?

Perhaps include the code sample in this changeset as well (the one from the level definitions)

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -999,3 +1011,62 @@ export const tagCodeArtifacts = <
...
+ * @example Sample usage
+ *
+ * ```typescript
+ * // Each telemetry event carries a logLevel property whose value indicates
+ * // how important that event is for diagnosing Fluid Framework behaviour.
+ * // Use it in your logger's send() to decide which events to sample or drop:
+ * public send(event: ITelemetryBaseEvent): void {
+ *   const eventLogLevel = event.logLevel as number;
+ *   if (eventLogLevel < LogLevelValue.essential) {
+ *     // Non-essential events can be sampled to reduce telemetry volume
```
ChumpChief: "Sampling" implies taking a random subset in most contexts.

Consider a different word (i.e. probably "filter") to describe taking all-or-none in both the comments and event property name.

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -999,3 +1011,62 @@ export const tagCodeArtifacts = <
...
+ *     event.sampleEvent = true;
```
ChumpChief: Can this just be early-exit?  I think this would be clearer that we're dropping the telemetry on the floor.

It's also probably not a good practice for the logger to modify the event on the way out the door.  Not sure if adding properties would also pollute telemetry?

```suggestion
 *     return;
```

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -999,3 +1011,62 @@ export const tagCodeArtifacts = <
 					})
 		| (T[P] extends undefined ? undefined : never);
 } => tagData<TelemetryDataTag.CodeArtifact, T>(TelemetryDataTag.CodeArtifact, values);
+
+/**
+ * Numerical values that indicate the importance of a telemetry event for diagnostics,
+ * enabling consumers to make sampling or filtering decisions.
```
ChumpChief: If the rule is all-or-none then we shouldn't say "sampling"

```suggestion
 * enabling consumers to make filtering decisions.
```

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -999,3 +1011,62 @@ export const tagCodeArtifacts = <
...
+ *
+ * @remarks
+ * If an event does not contain a `logLevel` property, it should be treated as an `essential`.
+ *
+ * @example Sample usage
+ *
+ * ```typescript
+ * // Each telemetry event carries a logLevel property whose value indicates
+ * // how important that event is for diagnosing Fluid Framework behaviour.
+ * // Use it in your logger's send() to decide which events to sample or drop:
```
ChumpChief:
```suggestion
 * // Use it in your logger's send() to decide which events to drop:
```

packages/utils/telemetry-utils/src/test/telemetryLogger.spec.ts
```diff
@@ -430,3 +431,89 @@ describe("convertToBasePropertyType", () => {
...
+		});
+
+		assert.strictEqual(logger.events.length, 1, "One event should be logged");
+		const event = logger.events[0];
+
+		// Simulate what a consumer's send() would do:
+		// if the event's logLevel value is below essential, mark as sampleEvent
+		const eventLogLevel = event.logLevel as number;
+		if (eventLogLevel < LogLevelValue.essential) {
+			event.sampleEvent = true;
```
ChumpChief: Similar recommendation here as in interface definition file, avoid putting the property on the event.

packages/utils/telemetry-utils/src/telemetryTypes.ts
```diff
@@ -78,6 +80,12 @@ export interface ITelemetryGenericEventExt extends ITelemetryPropertiesExt {
 	 * @defaultValue "generic"
 	 */
 	category?: TelemetryEventCategory;
+
+	/**
+	 * Optional log level of the event.
+	 * @defaultValue {@link LogLevelValue.essential}
+	 */
+	logLevel?: LogLevelValueType;
```
ChumpChief: This naming is a little confusing, as my understanding is that there's already a "LogLevel" (verbose/default/error) that is distinct from this one.  And also the mismatch of `logLevel` vs. `LogLevelValue` or `LogLevelValueType` probably invites some confusion as well.  E.g. imagine a customer is on an FF documentation page for something called "log level" - how would they know if they're looking at the right thing?

I'd probably recommend deviating completely from the "log level" naming to avoid conflation.  Maybe something like "importance" would convey?

I also forget what we landed on for its interaction/relationship to the pre-existing LogLevel, but that interaction/relationship should also be documented either here or in LogLevel's documentation

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -999,3 +1011,62 @@ export const tagCodeArtifacts = <
...
+	 * They need not be collected in production.
+	 */
+	verbose: 10 as 10 & BrandedType<"LogLevelValueType">,
+
+	/**
+	 * Information about the session. These logs could be omitted in some sessions
+	 * if needed (e.g. to reduce overall telemetry volume). If any are collected
+	 * from a particular session, all should be.
+	 */
+	info: 20 as 20 & BrandedType<"LogLevelValueType">,
```
ChumpChief: Essential/info/verbose might be fine, but maybe consider a tri-state that more clearly indicates a linear scale of "ability to drop safely".  Maybe

* essential | recommended | optional
* mandatory | recommended | optional

Point being that even though yes the "verbose" items might be "verbose" and the "info" items might be informational, this feature is primarily used to make a "can I drop it" decision and so that's more important to convey than the verbosity/intent of the telemetry.

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -1038,4 +1049,4 @@ export const LogLevelValue = {
  * If an event does not contain a `logLevelValue` value, it should be treated as `essential`.
  * @beta
  */
-export type LogLevelValue = number & BrandedType<"LogLevelValue">;
+export type LogLevelValueType = number & BrandedType<"LogLevelValueType">;
```
jason-ha: Why rename? I think we prefer to keep the type and the value names the same.

packages/utils/telemetry-utils/src/telemetryTypes.ts
jason-ha: Isn't this the case that we don't want to leave optional? This is internal and how we can make sure we are emitting it for all log events, yes?

MarioJGMsoft: The decided upon interaction with the existing LogLevel was to replace it. I'll update the docs to point that out

MarioJGMsoft: But this is a `@legacy+@beta` interface, wouldn't adding a new required parameter break people that are using it outside of Fluid?

.changeset/tricky-zoos-see.md
markfields: Logger instance seems clear cut and tractable

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -842,8 +850,12 @@ function convertToBaseEvent({
 	category,
 	eventName,
 	...props
-}: ITelemetryEventExt): ITelemetryBaseEvent {
-	const newEvent: ITelemetryBaseEvent = { category, eventName };
+}: ITelemetryEventExt): ITelemetryBaseEvent & { logLevel: LogLevelValueType } {
+	const newEvent: ITelemetryBaseEvent & { logLevel: LogLevelValueType } = {
+		category,
+		eventName,
+		logLevel: LogLevelValue.essential,
```
markfields, line 859: Nit
```suggestion
		logLevel: LogLevelValue.essential, // Will be overwritten if props specifies a different LogLevelValue
```

Or put a more general comment above `const newEvent` since the same is true for `category` and `eventName`

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: applied change

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -190,7 +194,11 @@ export abstract class TelemetryLogger implements ITelemetryLoggerExt {
 		logLevel: typeof LogLevel.verbose | typeof LogLevel.default = LogLevel.default,
```
markfields, line 194: We need to unify this param with the event props.  Could be a follow-up if you can do it right away but don't want to block this PR.

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -999,3 +1011,62 @@ export const tagCodeArtifacts = <
...
+ * enabling consumers to make sampling or filtering decisions.
+ *
+ * @remarks
+ * If an event does not contain a `logLevelValue` value, it should be treated as `essential`.
+ *
+ * @see {@link LogLevelValue}
+ *
+ * @beta
+ */
+export type LogLevelValueType = number & BrandedType<"LogLevelValueType">;
```
markfields: I think naming a type `FooType` is generally discouraged.  Too bad we can't just call the const and type `LogLevel` (I mean, we could, since it's exported from another package, but shouldn't).

How about `LogLevelNumber` or `AnyLogLevel` or `OutputLogLevel`?  I won't object to merging as-is but if we can find consensus around another type name that would be good.

cc @jason-ha, @ChumpChief

packages/utils/telemetry-utils/src/sampledTelemetryHelper.ts
```diff
@@ -291,6 +292,7 @@ export class SampledTelemetryHelper<
 				...bucketProperties, // If the bucket doesn't exist and this is undefined, things work as expected
 				...measurements,
 				...processedCustomData,
+				logLevel: LogLevelValue.info,
```
markfields, line 297: Let's add a comment above, maybe in "remarks" section for this class, that states that the event samples are emitted with LogLevelValue.Info

packages/utils/telemetry-utils/src/telemetryTypes.ts
markfields: Maybe `EventLogLevel`? or `EventLevel`?  "Log Level" is a pretty common concept, so I'd lean towards sticking with that rather than coining a new term.

markfields: (and if we do get a new term, then let's name the const and the type the same thing - will resolve my other comment about having "Type" in the type name)

markfields: This is the input type to the `TelemetryLogger` base class, which is where default values are filled in, so I don't think this needs it.

If in the future we want to force all devs instrumenting code to explicitly set it (which is debatable IMO) then yeah make it required here.

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/test/telemetryLogger.spec.ts
```diff
@@ -430,3 +431,89 @@ describe("convertToBasePropertyType", () => {
...
+		}
```
markfields: It looks like these tests are trying to be that "example usage", right?  If so then I'd take it a bit further and create a class implementing `ITelemetryBaseLogger` that has this sampling logic (call it `ExampleSamplingLogger` or something).

As-is, this test is just asserting that `LogLevelValue.info < LogLevelValue.essential` (because `TestTelemetryLogger` doesn't do anything interesting with LogLevel anyway).

So instead of using `TestTelemetryLogger`, pass your `ExampleSamplingLogger` to `createChildLogger`, and then log to the Child Logger, and then have `ExampleSamplingLogger` track whatever you want.  Probably combine all these tests into one, that sends a variety of logs at different levels (including no level) and then asserts how many of different level were logged, and if you have custom sampling threshold then how many were above/below that or whatever.

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: I agree, it can be a follow up PR when we deprecate the LogLevel

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: It was something I needed to do since they were going to have two different tags (@public and @beta) but now I've made them both beta because `BrandedType` would also need to be updated and that seems out of scope for this PR to me

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: I updated the name to be the same.

packages/utils/telemetry-utils/src/sampledTelemetryHelper.ts
MarioJGMsoft: Added comment

packages/utils/telemetry-utils/src/test/telemetryLogger.spec.ts
MarioJGMsoft: Created test

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -190,7 +194,11 @@ export abstract class TelemetryLogger implements ITelemetryLoggerExt {
 		logLevel: typeof LogLevel.verbose | typeof LogLevel.default = LogLevel.default,
 	): void {
 		this.sendTelemetryEventCore(
-			{ ...event, category: event.category ?? "generic" },
+			{
+				...event,
+				category: event.category ?? "generic",
+				logLevel: event.logLevel ?? LogLevelValue.essential,
```
lindsnguyen, line 200: If im reading this correctly I think setting `event.logLevel` here to `LogLevelValue.essential` if it's empty/null is redudent because in `convertToBaseEvent` we set the newEvent to have `LogLevelValue.essential` and will only change it back to the originally set prop in the for loop.

Not sure if we want to remove this. Alternatively, if we find that this line is neccessary here / we want to set the value of logLevel here and not do it later on in `convertToBaseEvent`, do we need/want to add it into our function `sendPerformanceEvent` when calling into `sendTelemetryEventCore`? Wondering if I'm missing a reason we're not explictly handing that case or if we missed that scenario.

From what I can tell the 3 public functons are `sendTelemetryEvent`, `sendErrorEvent`, and `sendPerformanceEvent` so is there a reason we're not explictly handing performance?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1757,6 +1758,7 @@ export class ContainerRuntime
 				// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 				gcOptions_gcGeneration: runtimeOptions.gcOptions[gcGenerationOptionName],
 			}),
+			logLevel: LogLevelValue.info,
```
markfields: @MarioJGMsoft  This log event was removed in #26748, you'll have a merge conflict, just keep it deleted.

packages/utils/telemetry-utils/src/test/telemetryLogger.spec.ts
markfields: The new tests look great to me!

packages/utils/telemetry-utils/src/logger.ts
markfields: Just please do it ASAP after this.  cc @shlevari (I wouldn't call this workstream "done" without that deprecation)

packages/utils/telemetry-utils/src/logger.ts
markfields: (I prefer the current semantics which is descriptive of the nature of the data - what the consumer does it with that is up to them . I feel Matt's suggestion makes it more prescriptive - telling consumers what they can/should do)

I'm open-minded though, this stuff is definitely subjective and we'll learn from how our decisions play out!

packages/utils/telemetry-utils/src/test/telemetryLogger.spec.ts
```diff
@@ -430,3 +436,83 @@ describe("convertToBasePropertyType", () => {
 		});
 	});
 });
+
+describe.only("LogLevelValue-based sampling", () => {
```
markfields, line 440: .only!

packages/utils/telemetry-utils/src/test/telemetryLogger.spec.ts
```diff
@@ -430,3 +436,83 @@ describe("convertToBasePropertyType", () => {
...
+	/**
+	 * Example implementation of ITelemetryBaseLogger that uses LogLevelValue
+	 * to filter events. Events with a logLevel below the configured sampling
+	 * threshold are dropped; the rest are kept.
+	 */
+	class ExampleSamplingLogger implements ITelemetryBaseLogger {
```
markfields, line 446: nit: I'd call it filtering.  The sampling part would be the fact that the filter only applies to 5% of sessions or whatever, which is outside the scope of what's being tested here.

markfields: i.e. this is more testing the evolution of "minLogLevel" concept (which is being removed since it's naturally the log consumer's responsibility as illustrated here)

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -190,7 +194,11 @@ export abstract class TelemetryLogger implements ITelemetryLoggerExt {
 		logLevel: typeof LogLevel.verbose | typeof LogLevel.default = LogLevel.default,
 	): void {
 		this.sendTelemetryEventCore(
```
markfields, line 196: I think we can/should make logLevel required on the event passed to `sendTelemetryEventCore`

markfields: (related to Lindsey's comment below about the redundant fallback logic - getting the types clear will help sort this out)

packages/utils/telemetry-utils/src/logger.ts
markfields: > From what I can tell the 3 public functons are sendTelemetryEvent, sendErrorEvent, and sendPerformanceEvent so is there a reason we're not explictly handing performance?

I think this proves your point about how `convertToBaseEvent`'s fallback is sufficient and this is redundant?

Would be great to have a test case per public function that confirms that the default value is added.

packages/utils/telemetry-utils/src/telemetryTypes.ts
markfields, line 1: (I thought I wrote this comment already but can't find it)

I think we should probably get rid of `ITelemetryEventExt` in this PR.  It's hardly used, and its not clear what its purpose is v. `ITelemetryGenericEventExt`.

If you don't want to get rid of it, then at least add `logLevel` as required on there (or not, maybe it depends on what you do with `convertToBaseEvent`)

markfields (approved):
Approving with a few more comments, just to say the direction is good and you're really close!  Will approve again after you merge main and address the final comments from me and others.

jason-ha:
> Approving with a few more comments, just to say the direction is good and you're really close! Will approve again after you merge main and address the final comments from me and others.

@markfields, I recommend against Approving Draft PRs.

packages/utils/telemetry-utils/src/test/telemetryLogger.spec.ts
```diff
@@ -430,3 +436,83 @@ describe("convertToBasePropertyType", () => {
...
+		public events: ITelemetryBaseEvent[] = [];
+
+		public constructor(
+			private readonly samplingThreshold: LogLevelValue = LogLevelValue.essential,
+		) {}
+
+		public send(event: ITelemetryBaseEvent): void {
+			const eventLogLevel = (event.logLevel as number) ?? LogLevelValue.essential;
```
jason-ha, line 454: Should not use `as` (and I don't understand what you are trying to do with it).

jason-ha, line 454: This looks like we are telling customers that they need to select a default. I think we said we don't want customers to have to assign a default level.

.changeset/tricky-zoos-see.md
jason-ha: Logger instance is probably not sufficient. Especially with a component model there can be multiple logger instances.
What is the lifetime boundary that FF wants? If it is container then we can say that.
Is there a volume level that we think is needed to be statistically significant? Can we recommend a minimum number of sessions be complete?

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -174,7 +175,10 @@ export abstract class TelemetryLogger implements ITelemetryLoggerExt {
 	 *
 	 * @param event - the event to send
 	 */
-	public abstract send(event: ITelemetryBaseEvent, logLevel?: LogLevel): void;
+	public abstract send(
+		event: ITelemetryBaseEvent & { logLevel: LogLevelValue },
```
jason-ha, line 179: Can we name this type instead?

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -174,7 +175,10 @@ export abstract class TelemetryLogger implements ITelemetryLoggerExt {
...
+		logLevel?: LogLevel,
```
jason-ha, line 180: Please add guidance on how to use logLevel parameter.

packages/utils/telemetry-utils/src/logger.ts
jason-ha, line 274: Looks like `convertToBaseEvent` should take LogLevel into account and won't be able to internally deprecate.
Right now more things become essential including `LogLevel.verbose` events.

packages/utils/telemetry-utils/src/logger.ts
jason-ha: It is probably useful to understand what that deprecation actually looks like before taking a partial step.

packages/utils/telemetry-utils/src/logger.ts
jason-ha: I am concerned about being able to achieve the goal without trying to enumerate the work that needs to be done.
Here you should explain how/why we get to this point and that it is intended to be temporary.
Having found at least one place where verbose events become essential, I find it hard to suggest anyone try to use the new property.

jason-ha (changes requested):
It feels like this effort still could benefit from having an update to date design document. Maybe that is just a much more verbose PR description.
It is not clear to me how we could ask anyone to start using the new property since there appears to be verbose events that are now essential. Perhaps it is a non-issue if those are all filtered already or those are all already being regularly captured today.
A precondition for approval of this should be to connect it on Pages side, generate telemetry, and audit the results.

packages/utils/telemetry-utils/src/test/telemetryLogger.spec.ts
MarioJGMsoft: I missed it, thanks

packages/utils/telemetry-utils/src/telemetryTypes.ts
jason-ha: While the goal is to replace it they are going to be sitting next to each other for potentially long time. I expect it will be confusing even for FF devs. So, I would recommend something else as well.
Some other possibilities adjacent to `LogImportance`: `LogRelevance`, `LogRelevancy`, `LogSignificance`, `LogPertinence`, `LogBearing`

packages/utils/telemetry-utils/src/logger.ts
jason-ha: Thinking about it further I can see other complications in the deprecation. As with the regular deprecation process it is best to prepare a branch where the deprecated element doesn't exist to make sure bases are covered.

packages/utils/telemetry-utils/src/telemetryTypes.ts
MarioJGMsoft: I think we should go with `LogImportance`

markfields:
> Having found at least one place where verbose events become essential

@jason-ha what was this case?

jason-ha:
> > Having found at least one place where verbose events become essential
> 
> @jason-ha what was this case?

@markfields, https://github.com/microsoft/FluidFramework/pull/26548/changes/BASE..f3948e8099f459e383893fd3f04c474f842ae259#r2949445239
`sendPerformanceEvent` with `LogLevel.verbose` calls `sendTelemetryEventCore` (ignores logLevel parameter) calls `convertToBaseEvent` (always sets "essential" as default for `event.logLevel`)

PR #26549, opened 2026-02-25
Fix telemetry upload pnpm install

ChumpChief:
One more fix, telemetry upload also needs to use a relative path for the pnpm install.


tylerbutler (approved):

MarioJGMsoft (approved):
LGTM!

PR #26550, opened 2026-02-25
Move pnpm overrides from eslint-config-fluid to workspace root

frankmueller-msft:
## Summary
- Moves `pnpm.overrides` (`mocha>serialize-javascript` CVE fix) and its comment from `common/build/eslint-config-fluid/package.json` to the workspace root `package.json`
- Removes the entire `pnpm` section from `eslint-config-fluid/package.json` since the `onlyBuiltDependencies` entries (`esbuild`, `unrs-resolver`) are already in the root
- **Eliminates pnpm config warnings** that appear on every install and build: pnpm 10 only reads `pnpm.overrides` and `pnpm.onlyBuiltDependencies` from the workspace root

## Test plan
- [ ] `pnpm install` completes without pnpm config field warnings
- [ ] CI build passes (the `mocha>serialize-javascript` override still applies workspace-wide from the root)

### DIFF
--- a/common/build/eslint-config-fluid/package.json
+++ b/common/build/eslint-config-fluid/package.json
@@ -76,18 +76,6 @@
 			}
 		}
 	},
-	"pnpm": {
-		"commentsOverrides": [
-			"serialize-javascript - CVE-2024-11831 impacts version 6.0.0 which is pinned by mocha 10.4.0, which in turn comes from mocha-multi-reporters 1.5.1 (which has no updated version at this time)"
-		],
-		"overrides": {
-			"mocha>serialize-javascript@6.0.0": "^6.0.2"
-		},
-		"onlyBuiltDependencies": [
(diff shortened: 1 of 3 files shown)


frankmueller-msft:
Closing this PR — the approach is wrong.

`eslint-config-fluid` has its own `pnpm-workspace.yaml` making it a standalone workspace for independent builds. Removing `pnpm.overrides` from it reverts `serialize-javascript` to the vulnerable 6.0.0 in the independent build (`ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` confirmed in CI). The root workspace overrides don't apply to independent builds since they use their own lockfile.

The pnpm warnings in client workspace builds are an unavoidable cost of the dual-workspace structure (`eslint-config-fluid` is both a standalone workspace root and a client workspace member).

PR #26551, opened 2026-02-25
Suppress repeated port mapping warnings in getTestPort

frankmueller-msft:
## Summary
- Checks `fs.existsSync()` before reading `testportmap.json` and silently falls back to port 8081 when the file doesn't exist
- **Eliminates 10+ identical warnings** per CI build from `getTestPort()` when the port map hasn't been generated
- The warning now only fires when the file exists but can't be parsed, which indicates an actual problem

The file is absent whenever jest configs are loaded without a prior `assign-test-ports` run — the normal case for local single-package testing and for CI steps that evaluate jest configs outside of the `test:jest` flow.

## Bonus fix
Also fixes a latent bug: when the port map file exists but doesn't contain the requested package name, the original code returned `undefined` despite the function declaring a `string` return type. The `?? "8081"` fallback now correctly returns the default port in this case.

## Test plan
- [x] `Build - test-tools` pipeline passes
- [x] repo-policy-check passes

### DIFF
--- a/tools/test-tools/src/getTestPort.ts
+++ b/tools/test-tools/src/getTestPort.ts
@@ -7,27 +7,29 @@ import fs from "fs";
 import os from "os";
 import path from "path";
 
+const portMapPath = path.join(os.tmpdir(), "testportmap.json");
+
 /**
  * Get the port for the pkg from the mapping.  Use a default if the file or the entry doesn't exist
  * (e.g. an individual test is being run and the file was never generated), which should presumably
  * not lead to collisions.
  */
 export function getTestPort(pkgName: string): string {
-	let mappedPort: string;
+	if (!fs.existsSync(portMapPath)) {
+		return "8081";
+	}
 
 	try {
-		const portMapPath: string = fs
-			.readFileSync(path.join(os.tmpdir(), "testportmap.json"))
-			.toString();
+		const portMapContent: string = fs.readFileSync(portMapPath).toString();
 		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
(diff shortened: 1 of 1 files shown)


alexvy86 (changes requested):
I don't think we should make this change.

First, the warnings serve a good purpose, letting someone know that they need to fix something or they might run into issues when running jest tests in parallel because the same port will be reused across packages, probably resulting in unexpected results. Defaulting to a single port with no other information would make it harder to diagnose/troubleshoot issues in the scenario this infrastructure is designed to support (running jest tests in parallel across packages).

I'm assuming the logs the AI agent looked at are the ones from `npm run checks` in the Policy checks stage, like [these](https://dev.azure.com/fluidframework/public/public%20Team/_build/results?buildId=380141&view=logs&j=eefa678c-f484-5837-78d3-627f1635913b&t=a53dad46-1a42-5abb-a3cd-1368a15b6673&l=120). I don't recall the exact details but those happen because somehow the fact of _reading_ jest config files (like checks have to do) causes the port-assignment-check code to run. The logs don't really mean that anything bad is happening in this case. I tried to fix that at some point, unsuccessfully.

Finally, the `test-tools` package itself doesn't have jest tests, so while running `build - test-tools` because the changed file is in the `test-tools` package makes sense, it is not good validation that this change actually has the intended effect.

Marking as request-changes to block merging, but really I think we should just close this PR.

frankmueller-msft:
Closing per Alex's feedback — the warnings are intentional and serve as a diagnostic signal for parallel jest port assignment. The CI noise is cosmetic and harmless.

PR #26271, opened 2026-01-23
Add timeouts to summarizer stopping

kian-thompson:
There are some peculiar cases of documents hitting 10k ops where a summarizer doesn't shut down properly. The parent container continually recognizes it is the elected client, but it still has a summarizer hanging somewhere. Some timeouts were added in various places to ensure the summarizer closes properly for the parent to spawn a new one.

Looking at telemetry, the P99.9 of a summarizer's first event to proper connection was 35 seconds. Thus, a timeout of 2 minutes should be more than enough.

One interesting potential case is the following:
1. Parent container spawns summarizer
2. Summarizer starts load flow
3. Parent container disconnects (`summarizer?.stop(...)` is not called since load flow has not finished and returned the `summarizer` instance)
4. Summarizer finishes load flow, and we attempt to call `summarizer.run(...)`
5. At the start of `summarizer.run(...)`, it waits to see a `"connected"` event from the ContainerRuntime
i. This may not always happen if the summarizer gets into some weird state and isn't able to get a connection from service, causing the whole summarizer flow to hang

[AB#54693](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/54693)

PR #26557, opened 2026-02-26
fix(container-runtime): Fix stop timeout race in SummaryManager (0x263, 0x264)

frankmueller-msft:
## Summary

Fixes assert failures 0x263 ("Expected: starting") and 0x264 ("Expected: Not Off") in `SummaryManager` that occur during Real Service Stress Tests.

- **Root cause**: PR #26271 added a 2-minute stop timeout that calls `cleanupAfterSummarizerStop()` when the summarizer hangs. But the `.finally()` block in `startSummarization()` also calls `cleanupAfterSummarizerStop()` unconditionally when the promise chain completes. When the timeout fires first, the subsequent `.finally()` performs a redundant cleanup that either hits assert 0x264 (state is already `Off`) or cascades into assert 0x263 (double-cleanup corrupts a new summarizer cycle's state).
- **Fix**: Adds a `summarizerGeneration` counter that is incremented each time `cleanupAfterSummarizerStop()` runs. The promise chain captures the generation at the start of each `startSummarization()` call; `.finally()` skips cleanup if the generation has changed, meaning another path already performed it. Logs a `SummarizerCleanupAlreadyDone` telemetry event when this occurs.
- **Tests**: Two new tests reproduce both race scenarios (timeout fires while disconnected; timeout fires, reconnects, starts new cycle, then old chain completes).

## Race scenario for 0x264

1. `startSummarization()`: state → Starting → Running
2. `summarizer.run()` starts
3. Disconnect → `stop()` → state = Stopping, 2-minute timeout set
4. Summarizer hangs under stress test load
5. **Timeout fires** → `cleanupAfterSummarizerStop()` → state = Off
6. `summarizer.run()` promise finally resolves
7. `.finally()` → `assert(state !== Off)` → **0x264**

## Race scenario for 0x263

5b. Timeout → cleanup → state = Off → `shouldSummarize` true → `startSummarization()` #2 → state = Starting
6. Old `.finally()` fires → passes 0x264 (Starting ≠ Off) → calls `cleanupAfterSummarizerStop()` **again** → state = Off
7. #2's delay resolves → `assert(state === Starting)` → **0x263** (state is Off)

## Change details

**`summaryManager.ts`**:
- New `summarizerGeneration` field (monotonically increasing counter)
- `startSummarization()` captures generation at start
- `.finally()` checks generation before cleanup; skips with telemetry if stale
- `cleanupAfterSummarizerStop()` increments generation on entry

**`summaryManager.spec.ts`**:
- "Should not assert when stop timeout fires before promise chain completes" — reproduces 0x264
- "Should not corrupt new cycle when stop timeout fires and shouldSummarize is true" — reproduces 0x263 cascade

## Test plan

- [x] All 14 Summary Manager unit tests pass (ESM + CJS), including 2 new tests
- [x] CI passes
- [x] Stress tests no longer hit 0x263/0x264

[AB#59694](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/59694)

### DIFF
--- a/packages/runtime/container-runtime/src/summary/summaryManager.ts
+++ b/packages/runtime/container-runtime/src/summary/summaryManager.ts
@@ -104,6 +104,14 @@ export class SummaryManager
 	private summarizer?: ISummarizer;
 	private _disposed = false;
 	private summarizerStopTimeout?: ReturnType<typeof setTimeout>;
+	/**
+	 * Monotonically increasing counter that tracks summarizer lifecycle generations.
+	 * Incremented each time {@link cleanupAfterSummarizerStop} runs. Used by the
+	 * promise chain in {@link startSummarization} to detect that cleanup has already
+	 * been performed by another path (e.g. the stop timeout), so it can skip
+	 * redundant cleanup that would corrupt the state machine.
+	 */
+	private summarizerGeneration = 0;
 
 	public get disposed(): boolean {
 		return this._disposed;
@@ -245,6 +253,8 @@ export class SummaryManager
 
 		assert(this.summarizer === undefined, 0x262 /* "Old summarizer is still working!" */);
 
+		const generation = this.summarizerGeneration;
+
 		this.delayBeforeCreatingSummarizer()
 			.then(async (startWithInitialDelay: boolean) => {
 				if (this.disposed) {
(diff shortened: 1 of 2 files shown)


shlevari (commented):
@frankmueller-msft is there urgency to completing this PR? I'd like to wait for @kian-thompson to get to take a look at this, and he's out until Thursday. While this does in fact seem like it would prevent the assert being hit, I'm worried that we're just covering up a deeper issue that that assert is there to make sure we see. I want to make sure his original PR is actually fixing what it intended and just missing this case, as opposed to this issue revealing something deeper about the original fix.

frankmueller-msft:
Sure, we wait for @kian-thompson to sign off on this PR.

kian-thompson (approved):
LGTM. This handling doesn't cover up any deeper issue. There was just an overlap in the cleanup and that code assumed no one else ran it beforehand prior to my change.

PR #26558, opened 2026-02-26
Add policy check for ci:build script (Bug 59692, Part 2)

frankmueller-msft:
[AB#59692](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/59692)

## Summary

- Adds a `packageNames` scope to `ScriptRequirement` so the `ci:build` policy check applies only to `@fluidframework/protocol-definitions` and `@fluidframework/common-utils`
- Replaces the previous `optional` flag approach, which was semantically contradictory and couldn't catch missing scripts
- Catches both wrong script bodies and entirely missing `ci:build` scripts for scoped packages, while skipping workspace packages that don't need it

## Context

Bug 59692 reported that `protocol-definitions`' `ci:build` script was `"npm run build:compile"` instead of `"fluid-build . --task ci:build"`, silently skipping typetests:gen and API reports in CI. Part 1 (the script fix) was PR #26480. This is Part 2: a policy check to prevent future regressions.

## Test plan

- [x] Verify `fluidBuild.config.cjs` loads without errors: `node -e "require('./fluidBuild.config.cjs')"`
- [ ] Trace handler for `protocol-definitions` (in `packageNames`, has correct body → PASS)
- [ ] Trace handler for a client workspace package like `@fluidframework/container-runtime` (not in `packageNames` → SKIP)
- [ ] Confirm API report matches interface change (`packageNames?: string[]` in alphabetical position)

### DIFF
--- a/build-tools/packages/build-cli/api-report/build-cli.api.md
+++ b/build-tools/packages/build-cli/api-report/build-cli.api.md
@@ -124,6 +124,7 @@ export interface ScriptRequirement {
     body: string;
     bodyMustMatch?: boolean;
     name: string;
+    packageNames?: string[];
 }
 
 // (No @packageDocumentation comment for this package)
--- a/build-tools/packages/build-cli/src/config.ts
+++ b/build-tools/packages/build-cli/src/config.ts
@@ -325,6 +325,12 @@ export interface ScriptRequirement {
 	 * @defaultValue `false`
 	 */
 	bodyMustMatch?: boolean;
+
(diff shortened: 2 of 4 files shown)


frankmueller-msft:
Closing — the `packageNames` scoping feature needs to be released in build-tools first before the config change can land. Will split into two PRs: (1) build-tools feature, (2) config update after release.

PR #26561, opened 2026-02-26
ci: Add missing trigger dependency to pipelines

alexvy86:
## Description

Adds a missing trigger dependency to pipelines, on `include-build-lint.yml` which is used by `build-npm-package.yml` and `build-npm-client-package.yml`, so basically all our pipelines should have been depending on this file as well since they depend on one of the other two.

Caught by looking at https://github.com/microsoft/FluidFramework/pull/26544 which made changes to ìnclude-build-lint.yml` but caused no pipelines to trigger other than `repo-policy-check`.

## Reviewer Guidance


tylerbutler (approved):
One of my least favorite ADO behaviors.

PR #26564, opened 2026-02-26
Updating SharedTree's public transaction docs

jumyhre:
This PR updates the docs for SharedTree's transaction API. It adds:

- A brief explanation of the new TreeAlpha runTransaction API, and what benefits it provides over the old API
- Snippets for new API use, including async forms and transaction nesting
- A section on Constraints and their usage


docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+Requires that a specific node exists in the document.
+
+```typescript
+const importantNode = view.root;
+
+view.runTransaction(
+    () => {
+        view.root.value = 42;
+    },
+    {
```
jumyhre, line 216: Question for reviewers: In theory, the old transaction API supports constraints, but in practice the old transaction API only supports `nodeInDocument`. I didn't document it because by my understanding we're not planning to invest in the old API due to its correctness issues and want to replace it.  Should I add documentation on the old API specifically for nodeInDocument?

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -3,8 +3,6 @@ title: Transactions
 sidebar_position: 6
 ---
 
-TODO:#27374: update this page to describe the new transaction API
```
Josmithr, line 6: Oof. We should avoid comments like this in our website docs. Users would see this. Glad to see it removed.

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
 
 There are example transactions here: [Shared Tree Demo](https://github.com/microsoft/FluidExamples/tree/main/brainstorm).
+
+## New Transaction API
+
+:::caution
+
+The APIs described in this section are currently `@alpha` and may change in future releases.
+
+Unlike `Tree.runTransaction`, the Alpha `runTransaction` does **not** automatically roll back the transaction if the callback throws an error. If your callback may throw, you should handle errors explicitly and return `{ rollback: true }` when required
```
Josmithr: Nit: Our Markdown best practices generally recommend line breaking along sentence boundaries. This helps make future changes a bit easier to review / reason about.

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+
+:::
+
+The TreeBranchAlpha API introduces `TreeBranchAlpha.runTransaction` as a replacement for `Tree.runTransaction`. This new API exists to solve correctness issues where errors being thrown in runTransaction could result in an undefined state. This new API enables developers to customize how their application handles this error case.
```
Josmithr: Nit: I would recommend linking to the API docs for the referenced APIs here (assuming they've been published).

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+```typescript
+view.runTransaction(
+    () => {
+        view.root.name = "Dave";
+    },
+    { label: "rename-user" },
+);
+```
+
+If transactions are nested, only the outermost transaction's label is used.
```
Josmithr, line 140: Nit: this makes sense, but it took me a minute to reason out why. I wonder if it would be worth noting the _why_.

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+
+### Nested Transactions
+
+`runTransaction` and `runTransactionAsync` can be called from within the callback of another transaction, with some limitations. Nested transactions have the following behavior:
+
+-   If the inner transaction rolls back, only the inner transaction's changes are discarded. The outer transaction continues.
+-   Constraints are applied to the outermost transaction. A single commit is generated for the outermost transaction, encompassing all inner transactions.
```
Josmithr, line 148: What does this mean for constraints applied to inner transactions? Are they ignored? Or does this just mean that any constraints within the transaction (including nested transactions) are applied to the entire outer-most transaction?

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+-   Undo reverts the outermost transaction and all of its inner transactions together.
+-   Only the outermost transaction's label is used; inner labels are ignored.
+
+:::warning
+
+An asynchronous transaction cannot be started inside a synchronous transaction. The other three nesting combinations (sync-in-sync, sync-in-async, async-in-async) are all supported.
```
Josmithr: What is the error mode that a user will see if they try to do async in sync?

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+Constraints are specified via the `preconditions` field in the `params` argument to `runTransaction`, or returned from the transaction callback via the `preconditionsOnRevert` field:
+
+```typescript
+view.runTransaction(
+    () => {
+        view.root.content = "new value";
+    },
+    {
+        preconditions: [
+            { type: "nodeInDocument", node: someNode },
```
Josmithr, line 182: This could probably use an inline comment explaining what the precondition does.

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+}
+```
+
+**With return values:**
+
+```typescript
+const result = view.runTransaction(() => {
+    const oldName = view.root.name;
+    view.root.name = "Bob";
+    return { rollback: false, value: oldName };
```
Josmithr: Does `rollback` have a default? Or are users required to return something?

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+### Transaction Labels
+
+You can attach a label to a transaction via the `params` argument. The label is surfaced through `ChangeMetadata`, and can be useful for grouping related actions.
+
+```typescript
+view.runTransaction(
+    () => {
+        view.root.name = "Dave";
+    },
+    { label: "rename-user" },
```
Josmithr, line 130: An example of how this will affect the change metadata (either as an inline comment, or as notes following the codeblock) would probably be useful.

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+);
+```
+
+If transactions are nested, only the outermost transaction's label is used.
+
+### Nested Transactions
+
+`runTransaction` and `runTransactionAsync` can be called from within the callback of another transaction, with some limitations. Nested transactions have the following behavior:
```
Josmithr: These are obviously existing APIs, so this isn't feedback for this PR or anything, but I'm curious: would it be possible to make the async variation of this be an overload of `runTransaction`? Do we actually need 2 different function names?

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+
+Constraints are specified via the `preconditions` field in the `params` argument to `runTransaction`, or returned from the transaction callback via the `preconditionsOnRevert` field:
+
+```typescript
+view.runTransaction(
+    () => {
+        view.root.content = "new value";
+    },
+    {
+        preconditions: [
```
Josmithr, line 181: Nit: A quick comment above this indicating that `preconditions` applies to the application of the transaction itself might be useful (to help identify how this example is different from the next one).

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+    },
+);
+```
+
+```typescript
+view.runTransaction(() => {
+    view.root.content = "updated";
+    return {
+        rollback: false,
+        preconditionsOnRevert: [
```
Josmithr, line 194: Nit: A quick comment above this line indicating that the `preconditionsOnRevert` apply to if/when the associated transaction is *reverted* might be useful (to help identify how this example is different from the previous one).

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +34,181 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+            { type: "nodeInDocument", node: view.root },
+        ],
+    };
+});
+```
+
+### Supported Constraint Types
+
+// TODO:#59981: Document new constraint types as they are implemented. Consider breaking this section into its own page when we have enough constraint types to warrant it.
```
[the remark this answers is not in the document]
Josmithr: Yeah, we'll want to use mdx comment syntax here to avoid this appearing in the user-facing docs.

```suggestion
{/* TODO:#59981: Document new constraint types as they are implemented. Consider breaking this section into its own page when we have enough constraint types to warrant it. */}
```

docs/docs/data-structures/tree/transactions.mdx
Josmithr: IMO, it would probably still be worth at least a quick note. I'm not sure how quickly we will stabilize these new APIs (we have not really been bumping many things to public recently, so I fear it might be a while). But I'm curious how others feel.

docs/docs/data-structures/tree/transactions.mdx
jumyhre: It's optional and assumed false by default. I've removed the rollback: false lines so people dont think it's required, and I've added a comment showing that you can return { rollback: true } if you need to

docs/docs/data-structures/tree/transactions.mdx
jumyhre: I added a note to it that says it can also be applied via the old API. If anyone else has opinions please let us know

docs/docs/data-structures/tree/transactions.mdx
jumyhre: UsageError. I added a note

docs/docs/data-structures/tree/transactions.mdx
jumyhre: i'm leaning weakly towards no, just because I'm in favor of keeping the doc succinct, but i'm open to it if we think this is important info for API consumers

docs/docs/data-structures/tree/transactions.mdx
jumyhre: I _think_ it's technically possible to put it all into one API, but I'm not 100% on that

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +37,211 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+
+The `nodeInDocument` constraint is also available for use in the old transaction API via `Tree.runTransaction`.
+This is the only constraint which can be applied via `Tree.runTransaction`.
+
+:::
+
+#### `noChange` (Alpha)
+
+Requires that the document is in the exact same state when the transaction is applied as it was when the transaction was created.
+If _any_ concurrent edit is sequenced before the transaction, the constraint is violated and the transaction is rolled back.
```
Josmithr, line 236: This probably needs to be updated when your other PR (#26763) goes in to describe the new, more tolerant requirements. If this PR gets merged first, we probably want to make the changes in the PR changing the constraint.

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +37,211 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+Constraints are specified via the `preconditions` field in the `params` argument to `runTransaction`, or returned from the transaction callback via the `preconditionsOnRevert` field:
+
+```typescript
+view.runTransaction(
+    () => {
+        view.root.content = "new value";
+    },
+    {
+        preconditions: [
+            // preconditions applies to the transaction itself.
```
Josmithr: Nit: here, we put the comment _inside_ the `preconditions` declaration, but below we put the comperable comment _outside_ the declaration. We should probably be consistent.

Josmithr: Nit (to differentiate the API term from simple English):

```suggestion
            // `preconditions` applies to the transaction itself.
```

docs/docs/data-structures/tree/transactions.mdx
```diff
@@ -36,3 +37,211 @@ Tree.runTransaction(myTreeView, (treeView) => {
...
+        ],
+    },
+);
+```
+
+```typescript
+view.runTransaction(() => {
+    view.root.content = "updated";
+    return {
+        // preconditionsOnRevert only applies if the transaction is reverted (e.g. via an undo operation).
```
Josmithr: Nit (to differentiate the API term from simple English):
```suggestion
        // `preconditionsOnRevert` only applies if the transaction is reverted (e.g. via an undo operation).
```

Josmithr (approved):
Left a couple more nitpicks, but overall looks good to me! Thanks for adding this great documentation!

docs/docs/data-structures/tree/transactions.mdx
jumyhre: Good call. Will update this file in that PR once this is in

Josmithr (approved):

PR #26569, opened 2026-02-26
ci: add missing trigger paths for biome-formatted files

frankmueller-msft:
## Summary

PR #26547 changed `.devcontainer/devcontainer.json` and `.vscode/extensions.json` but no CI pipeline checked formatting, allowing a trailing comma error to merge (fixed by PR #26565).

**Root cause:** `biome check .` (run by `check:format:repo`) covers the entire repo, but no pipeline was running it on every PR. The `build-client` pipeline only triggers on specific paths, and `.devcontainer/` and `.vscode/` weren't included.

**Fix (two parts):**

1. **`repo-policy-check.yml`** — Add `check:format:repo` step (`biome check .`). This pipeline runs on every PR with no path filter, so formatting is always validated regardless of which files change. Note: `npm run checks` (the broader check) can't be used here because it depends on `fluid-build`/`flub` and full workspace deps, but this pipeline only installs root deps. `policy-check` and `layer-check` already run as separate steps.

2. **`build-client.yml`** — Add missing root-level config files that biome formats to the trigger paths:
   - `.pnpmfile.cjs`
   - `_buildProject.config.cjs`
   - `assertTagging.config.mjs`

## Test plan

- [ ] Verify the pipeline YAML is valid (no syntax errors)
- [ ] Confirm `repo-policy-check` pipeline runs `check:format:repo` on PRs touching any files

### DIFF
--- a/tools/pipelines/repo-policy-check.yml
+++ b/tools/pipelines/repo-policy-check.yml
@@ -117,10 +117,10 @@ extends:
             customCommand: 'run layer-check'
 
         - task: Npm@1
-          displayName: Format Check (repo)
+          displayName: Repo-wide checks
           inputs:
             command: 'custom'
-            customCommand: 'run check:format:repo'
+            customCommand: 'run checks'
 
         - task: Bash@3
           displayName: Prune pnpm store
--- a/tools/pipelines/repo-policy-check.yml
+++ b/tools/pipelines/repo-policy-check.yml
@@ -116,6 +116,12 @@ extends:
             command: 'custom'
             customCommand: 'run layer-check'
 
+        - task: Npm@1
+          displayName: Format Check (repo)
+          inputs:
+            command: 'custom'
+            customCommand: 'run check:format:repo'
+
         - task: Bash@3
           displayName: Prune pnpm store
           inputs:
--- a/tools/pipelines/repo-policy-check.yml
+++ b/tools/pipelines/repo-policy-check.yml
@@ -117,10 +117,10 @@ extends:
             customCommand: 'run layer-check'
 
         - task: Npm@1
-          displayName: Repo-wide checks
+          displayName: Format Check (repo)
           inputs:
             command: 'custom'
-            customCommand: 'run checks'
+            customCommand: 'run check:format:repo'
 
         - task: Bash@3
           displayName: Prune pnpm store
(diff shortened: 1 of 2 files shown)


alexvy86 (commented):
I don't like `.vscode/` as a trigger for the whole `build - client` pipeline; changes there shouldn't affect the CI build. Would something like the `repo-policy-check` pipeline be enough to check for formatting of those files? Same for `.devcontainer/`. The cjs/mjs files do feel like good additions.

tylerbutler:
> I don't like `.vscode/` as a trigger for the whole `build - client` pipeline; changes there shouldn't affect the CI build. Would something like the `repo-policy-check` pipeline be enough to check for formatting of those files? Same for `.devcontainer/`. The cjs/mjs files do feel like good additions.

I agree - we can add a format step to the repo-policy-check pipeline instead, and that pipeline is already triggered on almost any change - maybe every change.

frankmueller-msft:
Good call. Removed `.devcontainer/*` and `.vscode/*` from the `build-client` triggers and added `check:format:repo` (`biome check .`) to `repo-policy-check.yml` instead. Since that pipeline runs on every PR with no path filter, it covers formatting for all files without triggering the full build. Kept the cjs/mjs additions in `build-client`.

tools/pipelines/repo-policy-check.yml
```diff
@@ -116,6 +116,12 @@ extends:
             command: 'custom'
             customCommand: 'run layer-check'
 
+        - task: Npm@1
+          displayName: Format Check (repo)
+          inputs:
+            command: 'custom'
+            customCommand: 'run check:format:repo'
```
tylerbutler: nit

```suggestion
          displayName: Repo-wide checks
          inputs:
            command: 'custom'
            customCommand: 'run checks'
```

This would run some additional checks. Some are client-only, but it's also cheap to run.

tylerbutler (approved):
Looks good once CI passes.

tools/pipelines/repo-policy-check.yml
frankmueller-msft: Done — switched to `npm run checks` in d3c24e5.

frankmueller-msft: Reverted in 06215ca — `npm run checks` runs `fluid-build --task checks` which depends on flub/build-tools and full workspace deps, but this pipeline only installs root deps (`pnpm install --workspace-root`). Went back to `check:format:repo` since that's the only missing check here — `policy-check` and `layer-check` already run as separate steps above.

alexvy86 (approved):

PR #26573, opened 2026-02-26
ci: suppress baseline-browser-mapping warnings in webpack steps

frankmueller-msft:
## Summary

- PR #26544 added `BASELINE_BROWSER_MAPPING_IGNORE_OLD_DATA=1` to the build and lint steps in `include-build-lint.yml` to suppress the 100+ "data is over two months old" warnings from the `baseline-browser-mapping` transitive dependency of `browserslist`.
- That PR missed the two webpack `Npm@1` steps in `build-npm-client-package.yml` — the main `npm run webpack` step and the devtools `Build devtools` step — which also invoke browserslist and produce the same warnings.
- This PR adds the same `env` block to both webpack tasks, matching the pattern already established in `include-build-lint.yml`.

## Test plan

- [ ] Verify the YAML is syntactically valid (CI pipeline validation on PR)
- [ ] Confirm webpack steps no longer emit baseline-browser-mapping staleness warnings in build logs

### DIFF
--- a/tools/pipelines/templates/build-npm-client-package.yml
+++ b/tools/pipelines/templates/build-npm-client-package.yml
@@ -365,6 +365,9 @@ extends:
 
               - task: Npm@1
                 displayName: 'npm run webpack'
+                env:
+                  # Suppress baseline-browser-mapping staleness warnings during webpack.
+                  BASELINE_BROWSER_MAPPING_IGNORE_OLD_DATA: 1
                 inputs:
                   command: custom
                   workingDir: '$(Pipeline.Workspace)/${{ parameters.buildDirectory }}'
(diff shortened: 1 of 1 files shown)


alexvy86:
I don't see the warnings for either of those steps, neither in the `internal` CI runs nor the `public` PR ones (like I did for #26544 ). Can we point to where they show up that prompted these additions? I'd just don't want to sprinkle this env var in places where it is not necessary.

frankmueller-msft:
@alexvy86 You're right — I checked multiple CI builds (both before and after #26544 merged) and the `baseline-browser-mapping` warnings do **not** appear in the `npm run webpack` or `Build devtools` steps. They only show up in the build/lint steps, which #26544 already covers.

Builds checked:
- **380476** (PR #26568, post-#26544) — webpack log (log 68): no matches
- **380476** devtools log (log 81): no matches
- **380041** (PR `refactor/structured-return-types`, pre-#26544) — webpack log (log 68): no matches
- **380041** devtools log (log 81): no matches

This PR was a preemptive addition without confirming the warnings actually appear here. I'll close it.

PR #26575, opened 2026-02-26
Drop CommonJS support from React package

CraigMacomber:
## Description

Drop CommonJS support from the @fluidframework/reactpackage.
This package is rather new, and only has alpha APIs,
so this does not drop support for anything stable.

This enables this package to add future dependencies which are ESM only, and to have a simpler and faster build.

## Reviewer Guidance


Josmithr (approved):

CraigMacomber:
This is failing due to https://dev.azure.com/fluidframework/public/public%20Team/_build/results?buildId=380514&view=logs&j=5122cb2b-0fab-5462-27ab-7d04b7dfe6ec&t=90a828af-f36d-57be-f9c1-197780797175&l=2095 .

Looks like our CI is correctly detecting that we dropped CJS support, and is reporting that as an error.

Josmithr (approved):

PR #26577, opened 2026-02-27
Add stagingModeAutoFlushThreshold for staging mode batch control

anthony-murphy-agent:
## Summary

- Add `stagingModeAutoFlushThreshold` option to `ContainerRuntimeOptions` that controls automatic batch flushing during staging mode
- When in staging mode, suppress turn-based/async flush scheduling until the accumulated batch reaches the threshold op count
- Default threshold: **1000 ops** (tied to `largeBatchThreshold` constant) — incoming ops always break the batch regardless
- Wrap `exitStagingMode` in a `PerformanceEvent` reporting duration, exit method, batch count, and batches at or over threshold
- Remove the `DisableFlushBeforeProcess` kill-bit flag (split out to #26770)

## Default Justification (from production telemetry)

- Copy-paste operations routinely produce batches of 1000+ ops (435K instances over 30 days via `GroupLargeBatch` telemetry)
- All observed large batches are non-reentrant single-turn batches from normal user actions (not reconnection replay — reconnect preserves batch boundaries)
- Receivers on modern Fluid versions (2.74+) handle 1000-op batches without jank (p99 processing duration ~5ms)
- 1000 matches the existing "large batch" telemetry threshold in `OpGroupingManager`
- The threshold only affects cross-turn accumulation; single-turn operations (like paste) are unaffected

## Key Design Points

- Only affects `scheduleFlush()` — direct `flush()` calls (incoming ops, connection changes, stashing, exit staging mode) bypass the threshold entirely
- No effect outside staging mode
- Exposed on the public `ContainerRuntimeOptions` interface (`@legacy @beta`) with `forwardCompat: false` type validation break acknowledged — consumers using `Partial<ContainerRuntimeOptions>` (the typical pattern via `IContainerRuntimeOptions`) are unaffected
- Config override (`Fluid.ContainerRuntime.StagingModeAutoFlushThreshold`) > runtime option > default (1000)

## Telemetry

- `ExitStagingMode` perf event: duration, `exitMethod` (commit/discard), `autoFlushThreshold`, `batches`, `batchesAtOrOverThreshold`
- `GroupLargeBatch` threshold changed from `>=` to `>` so staging-mode auto-flush batches (exactly at threshold) don't trigger the event

## Test plan

- [x] Ops accumulate under threshold during staging mode
- [x] Ops flush when threshold is reached (with telemetry assertion)
- [x] Incoming ops break batch regardless of threshold
- [x] Incoming non-runtime ops break batch during staging mode
- [x] Reconnect breaks batch during staging mode
- [x] Exit staging mode (commit) flushes remaining ops
- [x] Exit staging mode (discard) flushes outbox before rollback
- [x] enterStagingMode flushes pending outbox as non-staged
- [x] No effect outside staging mode
- [x] Default threshold suppresses turn-based flushing during staging mode
- [x] Config override > runtime option > default precedence
- [x] Runtime option > default precedence
- [x] IdAllocation + reconnect while in staging mode (highest risk — b4e1fd1dd25 interaction)
- [x] Reconnect resubmits pre-staged batches with threshold active
- [x] All 944 existing tests pass
- [ ] Manual testing with Word integration


packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1849,6 +1874,7 @@ export class ContainerRuntime
 		} else {
 			this._flushMode = runtimeOptions.flushMode;
 		}
+		this.stagingModeMaxBatchOps = runtimeOptions.stagingModeMaxBatchOps;
```
anthony-murphy: default here, so stagingModeMaxBatchOps  always has a value, and remove undefined

anthony-murphy: add a config override as well

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4809,6 +4838,20 @@ export class ContainerRuntime
 	}
 
 	private scheduleFlush(): void {
+		// During staging mode with a batch threshold, suppress automatic flush scheduling.
+		// Only flush when the main batch exceeds the threshold.
+		// Incoming ops still break the batch via direct this.flush() calls elsewhere
+		// (deltaManager "op" handler, process(), connection changes, getPendingLocalState,
+		// exitStagingMode). Those all bypass scheduleFlush(), so they're unaffected by this check.
+		// Additionally, outbox.maybeFlushPartialBatch() (called on every submit) detects
+		// sequence number changes and forces a flush as a safety net.
```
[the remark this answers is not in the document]
anthony-murphy-agent: Fixed in a977adc244 — the comment now reads:

> By default it throws if unexpected changes are detected; it only forces a flush as a safety net when partial-batch flushing is enabled via Fluid.ContainerRuntime.DisableFlushBeforeProcess.

This matches the suggested wording.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4441,6 +4441,187 @@ describe("Runtime", () => {
...
+
+					submitDataStoreOp(runtimeWithThreshold, "3", genTestDataStoreMessage("op3"));
+					// The 3rd op should trigger scheduleFlush to fall through to normal scheduling
+					await Promise.resolve();
+
+					assert.equal(
+						submittedOps.length,
+						0,
+						"Ops should not be submitted while in staging mode (flushed into PSM only)",
+					);
```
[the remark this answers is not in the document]
anthony-murphy-agent: Fixed in a977adc244 — the test now asserts that `outbox.mainBatchMessageCount` drops to 0 after the threshold is reached, confirming the outbox was actually emptied (ops moved into PendingStateManager as a staged batch).

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4441,6 +4441,187 @@ describe("Runtime", () => {
...
+						type: MessageType.ClientJoin,
+						contents: "test content",
+					});
+
+					// Ops are not submitted to the wire during staging mode — they're flushed to PSM
+					assert.equal(
+						submittedOps.length,
+						0,
+						"Ops should not be submitted to wire while in staging mode",
+					);
```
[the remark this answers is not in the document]
anthony-murphy-agent: Fixed in a977adc244 — the test now asserts that `outbox.mainBatchMessageCount` drops to 0 after the incoming op, confirming the outbox was flushed into PendingStateManager as a staged batch (not just that nothing was submitted to the wire).

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4809,6 +4838,20 @@ export class ContainerRuntime
 	}
 
 	private scheduleFlush(): void {
+		// During staging mode with a batch threshold, suppress automatic flush scheduling.
+		// Only flush when the main batch exceeds the threshold.
```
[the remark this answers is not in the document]
anthony-murphy-agent: Fixed in a977adc244 — the comment now reads "reaches or exceeds the threshold" to accurately reflect the `>=` condition.

anthony-murphy-agent:
The following pipeline checks are missing (expected for fork PRs). To trigger them, copy and paste these `/azp run` commands:

```
/azp run Build - api-markdown-documenter
/azp run Build - benchmark-tool
/azp run Build - build-common
/azp run Build - build-tools
/azp run Build - client packages
/azp run Build - common-utils
/azp run Build - eslint-config-fluid
/azp run Build - eslint-plugin-fluid
/azp run Build - protocol-definitions
/azp run Build - test-tools
```

```
/azp run repo-policy-check
/azp run server-gitrest
/azp run server-gitssh
/azp run server-historian
/azp run server-routerlicious
```

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -608,6 +620,16 @@ const defaultMaxBatchSizeInBytes = 700 * 1024;
 
 const defaultChunkSizeInBytes = 204800;
 
+/**
+ * Default maximum ops per staging-mode batch before automatic flush scheduling resumes.
+ *
+ * Chosen based on production telemetry: copy-paste operations routinely produce batches
+ * of 1000+ ops (435K instances over 30 days), and receivers on modern Fluid versions
+ * handle them without issues. 1000 also matches the existing "large batch" telemetry
```
markfields: Can we update the "large batch" telemetry to differentiate these large batches from "natural" ones?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4809,6 +4838,22 @@ export class ContainerRuntime
...
 	private scheduleFlush(): void {
+		// During staging mode, suppress automatic flush scheduling until the main batch
+		// reaches or exceeds the threshold.
+		// Incoming ops still break the batch via direct this.flush() calls elsewhere
+		// (deltaManager "op" handler, process(), connection changes, getPendingLocalState,
+		// exitStagingMode). Those all bypass scheduleFlush(), so they're unaffected by this check.
+		// Additionally, outbox.maybeFlushPartialBatch() (called on every submit) detects
+		// sequence number changes. By default it throws if unexpected changes are detected; it only
+		// forces a flush as a safety net when partial-batch flushing is enabled via
+		// Fluid.ContainerRuntime.DisableFlushBeforeProcess.
```
markfields: We can/should get rid of this flag.  I'll happily review a PR from a bot if you decide to kick it off.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -608,6 +620,16 @@ const defaultMaxBatchSizeInBytes = 700 * 1024;
...
 
+/**
+ * Default maximum ops per staging-mode batch before automatic flush scheduling resumes.
+ *
+ * Chosen based on production telemetry: copy-paste operations routinely produce batches
+ * of 1000+ ops (435K instances over 30 days), and receivers on modern Fluid versions
+ * handle them without issues. 1000 also matches the existing "large batch" telemetry
+ * threshold ({@link OpGroupingManager}).
+ */
+const defaultStagingModeAutoFlushThreshold = 1000;
```
markfields: Tie this in code to the "Large Batch" telemetry?

markfields:
I left some individual comments about logging in there, but thinking about it holistically, I think we want:
  - Log when this threshold is hit
  - Log when a batch is large in bytes (regardless of the main change here)
  - Be harmonious with existing "GroupLargeBatch" event (PS - Think about hosts that doesn't enable Grouped Batching)
  - Do we also want telemetry about large blobAttach batches? (regardless of the main change here)

markfields:
@anthony-murphy  Claude and I chatted about test cases (and I mentioned the ones you included in the PR description).  Here's the test plan that came out of it, worth browsing, there are some important corner cases here:

# stagingModeAutoFlushThreshold — Test Plan

## Already covered

| Scenario                                                 | Status |
| -------------------------------------------------------- | ------ |
| Ops accumulate under threshold                           | ✅      |
| Ops flush when threshold reached                         | ✅      |
| Incoming ops break batch regardless of threshold         | ✅      |
| Exit via `commitChanges` flushes remaining ops           | ✅      |
| No effect outside staging mode                           | ✅      |
| Default threshold (1000) suppresses turn-based flushing  | ✅      |

## To add

### 1. Direct-flush codepaths break the batch during staging mode

These tests demonstrate that every codepath calling `flush()` directly still breaks
the accumulated batch, regardless of threshold. Each should verify: outbox is drained,
ops move to PSM as a staged batch, and no ops are sent to the wire.

#### 1a. Incoming runtime op breaks batch

Already covered by existing test ("incoming ops break the batch regardless of
threshold"). Listed here for completeness.

#### 1b. Incoming non-runtime op breaks batch

Same as 1a but with a non-runtime (signal/system) op arriving. The `process()` path
calls `flush()` unconditionally (line 3066) unless `skipSafetyFlushDuringProcessStack`
is set. Verify a non-runtime inbound message also drains the outbox mid-accumulation.

#### 1c. Connection state change (reconnect) breaks batch

1. Enter staging mode, submit ops (under threshold, sitting in outbox)
2. Simulate disconnect then reconnect (`canSendOps` transitions true → false → true)
3. The reconnect flush (line 2962) should drain the outbox before `replayPendingStates`
4. Verify the accumulated ops became a staged batch in PSM
5. Verify pre-staged ops are resubmitted correctly after reconnect

#### 1d. `enterStagingMode` flushes any pending outbox contents

This covers the edge case where ops were submitted in the same JS turn before
`enterStagingMode()` is called (so a flush was scheduled but hasn't fired yet).

1. Submit ops (not yet flushed — still in outbox)
2. Call `enterStagingMode()` in the same turn
3. Verify those ops were flushed as a **non-staged** batch (they predate staging mode)
4. Submit more ops while in staging mode
5. `commitChanges()` — verify only the post-entry ops are staged

### 2. Exit via `discardChanges` flushes outbox before rollback

Same shape as the existing `commitChanges` exit test but using `discardChanges`.
Verify outbox is drained and rolled-back ops match what was submitted.

### 3. IdAllocation + reconnect while in staging mode (b4e1fd1dd25 interaction)

**This is the highest-risk gap.**

The fix in b4e1fd1dd25 added `scheduleFlush()` after `submitIdAllocationOpIfNeeded`
during `replayPendingStates` to ensure the IdAllocation op is flushed before new ops
with different refSeqs arrive. With threshold suppression, that `scheduleFlush()` will
now return early if in staging mode and under threshold — potentially re-introducing
the original bug.

Test scenario:

1. Enter staging mode
2. Disconnect, generate a compressed ID (queued in idAllocationBatch)
3. Reconnect — `replayPendingStates` submits IdAllocation op + calls `scheduleFlush()`
4. Simulate remote op arriving (bumps refSeq)
5. Generate 2nd compressed ID + submit a data store op
6. Verify no `outboxSequenceNumberCoherencyCheck` error

**If this test fails**, the fix is to exempt the `scheduleFlush()` call in
`replayPendingStates` from threshold suppression (e.g., pass a `force` flag, or
call `flush()` directly instead of `scheduleFlush()`).

### 4. Reconnect resubmits pre-staged batches while threshold is active

Verify that ops submitted *before* entering staging mode are correctly resubmitted
on reconnect, and that the threshold does not interfere with their resubmission
(since resubmitted pre-staging batches go through `replayPendingStates`, not
`scheduleFlush()`).

### 5. Config override > runtime option > default

Single test: create runtime with both a config override and a runtime option set to
different values. Verify the config override wins. Then verify runtime option wins
over default when no config override is present.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Done: #26770

packages/runtime/container-runtime/src/opLifecycle/opGroupingManager.ts
```diff
@@ -123,7 +130,7 @@ export class OpGroupingManager {
 			return batch as OutboundSingletonBatch;
 		}
 
-		if (batch.messages.length >= 1000) {
+		if (batch.messages.length >= largeBatchThreshold) {
```
anthony-murphy: we don't want staging mode batches to trigger this. the easiest fix might be to make this just be >, rather than >=, the alternative to only allow 999 ops in staging mode is just a tiny bit more complicated. we should also add a note here

packages/runtime/container-runtime/src/pendingStateManager.ts
```diff
@@ -775,6 +776,7 @@ export class PendingStateManager implements IDisposable {
 
 		const initialPendingMessagesCount = this.pendingMessages.length;
 		let remainingPendingMessagesCount = this.pendingMessages.length;
+		const replayedBatches: Record<number, IPendingMessage["batchInfo"]> = {};
```
anthony-murphy: Both `replayPendingStates` and `popStagedBatches` track batches using `Record<number, batchInfo>` keyed by `batchStartCsn`. For staged batches that never reach the wire, `clientSequenceNumber` is `undefined` and the fallback assigns `batchStartCsn = -1`. When multiple auto-flushes occur (e.g., threshold=3, 7 ops → 3 batches), all entries collide at key `-1`. The `??=` operator keeps only the first, so `ExitStagingMode` telemetry reports `batches: 1` instead of the actual count.

**Suggestion:** Replace `Record<number, batchInfo>` with an array (`batchInfo[]`). Push each batch's info where the current `??=` assignments are. Alternatively, key by the unique `clientId` (assigned via `uuid()` per unsent batch) instead of `batchStartCsn`. This correctly counts all staged batches regardless of their CSN value. (~10-line change across both methods.)

packages/runtime/container-runtime/src/opLifecycle/opGroupingManager.ts
anthony-murphy: +1 — this change is still pending. Changing `>=` to `>` as suggested here would prevent staging-mode auto-flush batches (exactly 1000 ops) from triggering `GroupLargeBatch` noise, while still capturing genuinely oversized batches (1001+ ops). Worth confirming with @markfields that this resolves [his telemetry concern](https://github.com/microsoft/FluidFramework/pull/26577#issuecomment-4078134877) and is compatible with existing production dashboards that may rely on the `>= 1000` definition.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -474,6 +476,18 @@ export interface ContainerRuntimeOptions {
...
+	 * Controls automatic batch flushing during staging mode.
+	 * Normal turn-based/async flush scheduling is suppressed while in staging mode
+	 * until the accumulated batch reaches this many ops, at which point the batch
+	 * is flushed. Incoming ops always break the current batch regardless of this setting.
+	 *
+	 * Set to Infinity to only break batches on system events (incoming ops).
+	 *
+	 * @defaultValue 1000
+	 */
+	readonly stagingModeAutoFlushThreshold: number;
```
anthony-murphy, line 490: The PR description states this option is "Placed in `ContainerRuntimeOptionsInternal` (not public)." However, `stagingModeAutoFlushThreshold` is added here as a required field on the exported `ContainerRuntimeOptions` beta interface, and the API report confirms it on the public surface. With `forwardCompat: false` in `package.json`, consumers implementing `ContainerRuntimeOptions` directly will fail to compile on upgrade (though most use `Partial<ContainerRuntimeOptions>` and are unaffected).

**Suggestion:** Either (a) move this property to `ContainerRuntimeOptionsInternal` only — aligning with the stated intent and removing the forward-compat break, or (b) update the PR description to acknowledge the intentional public placement and confirm the break is acceptable for this release train.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4785,6 +4845,20 @@ export class ContainerRuntime
...
+		// Incoming ops still break the batch via direct this.flush() calls elsewhere
+		// (deltaManager "op" handler, process(), connection changes, getPendingLocalState,
+		// exitStagingMode). Those all bypass scheduleFlush(), so they're unaffected by this check.
+		// Additionally, outbox.maybeFlushPartialBatch() (called on every submit) detects
+		// sequence number changes and throws if unexpected changes are detected.
+		if (this.inStagingMode) {
+			if (this.outbox.mainBatchMessageCount < this.stagingModeAutoFlushThreshold) {
+				return;
+			}
+			this.stagingModeAutoFlushCount++;
```
anthony-murphy: can we remove this variable. its computable by looking at the batches in the exit staging mode perf event, every batch over the threshold was auto flushed. these saves use from adding a member to an already gigantic class

packages/runtime/container-runtime/src/pendingStateManager.ts
anthony-murphy: Revisiting after latest changes: this `batchStartCsn` collision issue has been resolved. `popStagedBatches` now uses `Set<batchInfo>` with reference-identity dedup and `replayPendingStates` uses array-based accumulation, correctly counting batches without key collisions. This thread can be resolved. 👍

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4785,6 +4845,20 @@ export class ContainerRuntime
...
+		// reaches or exceeds the threshold.
+		// Incoming ops still break the batch via direct this.flush() calls elsewhere
+		// (deltaManager "op" handler, process(), connection changes, getPendingLocalState,
+		// exitStagingMode). Those all bypass scheduleFlush(), so they're unaffected by this check.
+		// Additionally, outbox.maybeFlushPartialBatch() (called on every submit) detects
+		// sequence number changes and throws if unexpected changes are detected.
+		if (this.inStagingMode) {
+			if (this.outbox.mainBatchMessageCount < this.stagingModeAutoFlushThreshold) {
+				return;
+			}
```
anthony-murphy: `this.stagingModeAutoFlushCount++` executes before the `if (this.flushScheduled) { return; }` guard. When ops exceed the threshold in a single synchronous turn (e.g., threshold=1000, 1200-op paste), ops 1001–1200 each increment the counter but hit the `flushScheduled` early return — producing `autoFlushCount: 201` alongside only 1 actual flush.

**Suggestion:** Move the increment below the `flushScheduled` guard:
```typescript
if (this.flushScheduled) {
    return;
}
if (this.inStagingMode) {
    this.stagingModeAutoFlushCount++;
}
```
This ensures the counter only increments when a flush is actually being scheduled. Consider adding a test with ops exceeding the threshold in a single synchronous turn to verify count accuracy.

packages/runtime/container-runtime/src/opLifecycle/opGroupingManager.ts
```diff
@@ -17,6 +17,13 @@ import type {
 	OutboundSingletonBatch,
 } from "./definitions.js";
 
+/**
+ * The number of ops in a batch at or above which the batch is considered "large"
+ * for telemetry purposes. Used by both {@link OpGroupingManager} (GroupLargeBatch event)
```
anthony-murphy, line 22: The JSDoc says "at or above which the batch is considered 'large'" but the code now uses `batch.messages.length > largeBatchThreshold` (strictly greater). A batch of exactly 1000 ops won't be considered "large" — contradicting the documentation.

**Suggestion:** Update to "above which the batch is considered 'large'" to match the strict-`>` behavior.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4785,6 +4845,20 @@ export class ContainerRuntime
 	}
 
 	private scheduleFlush(): void {
```
markfields, line 4839: Note that one potentially problematic `scheduleFlush` call is removed by PR #26784

packages/runtime/container-runtime/src/pendingStateManager.ts
```diff
@@ -775,6 +776,7 @@ export class PendingStateManager implements IDisposable {
 
 		const initialPendingMessagesCount = this.pendingMessages.length;
 		let remainingPendingMessagesCount = this.pendingMessages.length;
+		const replayedBatchInfos: IPendingMessage["batchInfo"][] = [];
```
anthony-murphy: this is a clear error, and won't work correctly. i'm surprised this is wrong while popStagedBatches is correct

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: The `stagingModeAutoFlushCount` counter has been removed entirely in the latest revision, so the overcounting bug is no longer applicable. However, the PR description's Telemetry section still references `autoFlushCount` as an `ExitStagingMode` field — see the updated review comment for the remaining documentation mismatch.

packages/runtime/container-runtime/src/opLifecycle/opGroupingManager.ts
anthony-murphy: This has been fixed — the JSDoc now correctly says "above which", matching the strict `>` condition. ✅

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1834,6 +1862,10 @@ export class ContainerRuntime
 			this.closeFn(error);
 			throw error;
 		}
+		this.stagingModeAutoFlushThreshold =
```
anthony-murphy, line 1865: Defensive hardening suggestion: if `getNumber()` returns `NaN` (rather than `undefined`), it passes through the `??` chain since `NaN` is not nullish. The downstream check `count < NaN` is always `false`, which disables threshold suppression entirely.

This **fails open safely** — you get more flushes (pre-feature default), not fewer — so it's not a correctness bug. But a guard would make the intent explicit:

```typescript
const configVal = this.mc.config.getNumber("Fluid.ContainerRuntime.StagingModeAutoFlushThreshold");
const threshold = (Number.isFinite(configVal) && configVal >= 0) ? configVal : undefined;
```

Non-blocking — just a nice-to-have for robustness.

anthony-murphy:
## Deep Review

The core `scheduleFlush()` threshold-suppression mechanism is architecturally sound, well-tested (574 new lines covering Mark Fields' 13-scenario test plan), and all prior review findings have been addressed across revisions — including the `batchStartCsn` collision, JSDoc mismatch, API placement clarification, and the `autoFlushCount` stale reference. Only minor polish items remain; no correctness bugs or blocking issues.

### Findings

1. **`GroupLargeBatch` telemetry boundary change (`>=` → `>`) affects non-staging batches of exactly 1000 ops** — `opGroupingManager.ts:130` — The threshold was historically `>= 1000`. This PR changes it to `> largeBatchThreshold` with an inline code comment explaining the rationale: staging-mode auto-flush batches at exactly the threshold should not trigger telemetry noise. However, this also silences non-staging batches of exactly 1000 ops across all batch paths. The change is deliberate and documented, but confirm with telemetry owners (Mark Fields) that downstream Kusto/ADX queries filtering on `>= 1000` are compatible with the boundary shift.

2. **`getNumber()` returning `NaN` bypasses nullish coalescing with no diagnostic signal** — `containerRuntime.ts:1862-1865` — If `getNumber()` returns `NaN` (not nullish), it passes through the `??` chain. The downstream check `count < NaN` is always `false`, disabling threshold suppression entirely. This **fails open safely** (more flushes, not fewer), so it's not a correctness bug. A `Number.isFinite()` guard would make the intent explicit. Non-blocking. (See [inline comment](https://github.com/microsoft/FluidFramework/pull/26577#discussion_r2967472402).)

3. **Required property on `ContainerRuntimeOptions` is a type-level breaking change** — `containerRuntime.ts:477-489` — `readonly stagingModeAutoFlushThreshold: number` is required on the exported `@legacy @beta` interface. Consumers implementing it directly will fail to compile, but most use `Partial<ContainerRuntimeOptions>` and are unaffected. With `forwardCompat: false` the blast radius is narrow. Informational only — ensure release notes mention the new required property.

### Path to Ready

To reach 9-10/10:
- [ ] **Get Mark Fields' sign-off** on the `>=` → `>` boundary change in `GroupLargeBatch` — confirm it aligns with his telemetry harmony request and is compatible with existing production dashboards
- [ ] **Get Scott Norton's validation** that `stagingModeAutoFlushThreshold` placement in `defaultsNotAffectingDocSchema` follows the established runtime options pattern
- [ ] **Complete manual testing with Word integration** — the only unchecked item in the test plan
- [ ] *(Optional)* Add `Number.isFinite()` guard on config value — trivial defensive hardening for the NaN edge case (Finding #2)

### Context for Reviewers

- **Staging mode is young and iteratively hardening** — introduced April 2025 (PR #23783, Mark Fields + Tony Murphy) with 5 follow-up PRs. The iterative pattern is normal feature maturation, not instability.
- **Two-tier flush safety architecture** — threshold suppression only affects the deferred `scheduleFlush()` path. All safety-critical flush paths (incoming ops, reconnection, stashing, enter/exit staging mode) call `flush()` directly and bypass the threshold.
- **The 1000-op default is data-driven** — matches the existing `largeBatchThreshold` in `OpGroupingManager`, backed by production telemetry (435K instances of 1000+ op batches over 30 days). Mark Fields explicitly requested this alignment.
- **All prior review findings fully addressed** — `batchStartCsn` collision fixed (Set-based dedup), JSDoc corrected ("above which"), API placement acknowledged, `autoFlushCount` reference removed from PR description, overcounting counter removed entirely.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: this is handled internally by config. not necessary for every caller

anthony-murphy: Deep Review: Fair point — if the config layer already handles non-finite values internally, an additional guard at each call site is unnecessary. Dropping this suggestion.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -613,6 +627,16 @@ const defaultMaxBatchSizeInBytes = 700 * 1024;
 
 const defaultChunkSizeInBytes = 204800;
 
+/**
+ * Default maximum ops per staging-mode batch before automatic flush scheduling resumes.
+ *
+ * Chosen based on production telemetry: copy-paste operations routinely produce batches
+ * of 1000+ ops (435K instances over 30 days), and receivers on modern Fluid versions
```
markfields, line 634: Nit: Do we want these numbers here? Feels a little out of context in the OSS repo

packages/runtime/container-runtime/src/pendingStateManager.ts
```diff
@@ -748,7 +747,9 @@ export class PendingStateManager implements IDisposable {
 	 * states in its queue. This includes triggering resubmission of unacked ops.
 	 * ! Note: successfully resubmitting an op that has been successfully sequenced is not possible due to checks in the ConnectionStateHandler (Loader layer)
 	 */
```
markfields, line 755: Add doc comment explaining the return value

packages/runtime/container-runtime/src/pendingStateManager.ts
```diff
@@ -748,7 +747,9 @@ export class PendingStateManager implements IDisposable {
...
-	public replayPendingStates(options?: ReplayPendingStateOptions): void {
+	public replayPendingStates(
+		options?: ReplayPendingStateOptions,
+	): IPendingMessage["batchInfo"][] {
```
markfields, line 758: @anthony-murphy, @dannimad -- Side note, sometimes I think about changing `replayPendingStates` to something like `dequeuePendingBatchesForReplay`. This would simplify the logic (no replay callback, no queue cursor management), and would give the ContainerRuntime access to all the batches both before and after replay.  Something to keep in mind if that refactor would ever simplify future work in this area.

packages/runtime/container-runtime/src/pendingStateManager.ts
```diff
@@ -894,6 +899,8 @@ export class PendingStateManager implements IDisposable {
 				clientId: this.stateHandler.clientId(),
 			});
 		}
+
+		return [...replayedBatchSet];
```
markfields, line 909: This shallow copy seems unnecessary since the set is instantiated inside this function's closure.

packages/runtime/container-runtime/src/pendingStateManager.ts
```diff
@@ -921,6 +930,7 @@ export class PendingStateManager implements IDisposable {
 			this.pendingMessages.toArray().every((m) => m.batchInfo.staged !== true),
 			0xb89 /* Shouldn't be any more staged messages */,
 		);
+		return [...batchSet];
```
markfields, line 939: This shallow copy seems unnecessary since the set is instantiated inside this function's closure.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -475,6 +477,18 @@ export interface ContainerRuntimeOptions {
...
 
+	/**
+	 * Controls automatic batch flushing during staging mode.
+	 * Normal turn-based/async flush scheduling is suppressed while in staging mode
+	 * until the accumulated batch reaches this many ops, at which point the batch
+	 * is flushed. Incoming ops always break the current batch regardless of this setting.
+	 *
+	 * Set to Infinity to only break batches on system events (incoming ops).
+	 *
+	 * @defaultValue 1000
```
markfields: nit: this stated default value could drift relative to largeBatchThreshold which is the true default value. Can you link there instead of saying 1000 explicitly?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3615,20 +3647,38 @@ export class ContainerRuntime
...
+						// NOTE: We can't use this.flush() here, because orderSequentially uses StagingMode and in the rollback case we'll hit assert 0x24c
+						this.outbox.flush();
+
+						this.stageControls = undefined;
+
+						// During Staging Mode, we avoid submitting any ID Allocation ops (apart from resubmitting pre-staging ops).
+						// Now that we've exited, we need to submit an ID Allocation op for any IDs that were generated while in Staging Mode.
+						this.submitIdAllocationOpIfNeeded({ staged: false });
+						const batchInfos = discardOrCommit();
+						event.reportProgress({
```
markfields, line 3669: I'd recommend putting these 2 props inside a details object, to avoid indicating two new first-class columns in a telemetry DB.  No need to JSON.stringify, just wrap it in `{ details: { ... } }`.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3615,20 +3647,38 @@ export class ContainerRuntime
...
-				// Now that we've exited, we need to submit an ID Allocation op for any IDs that were generated while in Staging Mode.
-				this.submitIdAllocationOpIfNeeded({ staged: false });
-				discardOrCommit();
-
-				this.channelCollection.notifyStagingMode(false);
+				PerformanceEvent.timedExec(
+					this.mc.logger,
+					{
+						eventName: "ExitStagingMode",
+						exitMethod,
```
markfields: @anthony-murphy - Might be easier to glance at telemetry if you just put this in the eventName.  `ExitStagingMode_${exitMethod}`.  As-is, for our internal telemetry `exitMethod` will be hiding in `Payload`.  Minor annoyance but you probably always want to immediately see which path was taken anyway so that's why I suggest putting it in the eventName

packages/runtime/container-runtime/src/containerRuntime.ts
markfields: Move the property `autoFlushThreshold` out of the initial property bag and into `details` here along with the other detailed info.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4785,6 +4838,20 @@ export class ContainerRuntime
...
 	private scheduleFlush(): void {
+		// During staging mode, suppress automatic flush scheduling until the main batch
+		// reaches or exceeds the threshold.
+		// Incoming ops still break the batch via direct this.flush() calls elsewhere
+		// (deltaManager "op" handler, process(), connection changes, getPendingLocalState,
+		// exitStagingMode). Those all bypass scheduleFlush(), so they're unaffected by this check.
+		// Additionally, outbox.maybeFlushPartialBatch() (called on every submit) detects
+		// sequence number changes and throws if unexpected changes are detected.
+		if (
+			this.inStagingMode &&
```
markfields, line 4848: I'm worried that this could be the wrong flag to check. In the case where we're (re)submitting pre-staged changes, even though the ContainerRuntime is "in staging mode", the resubmitted batches (including the possible ID Allocation op from `replayPendingStates` itself) are unstaged and need to be flushed.

There are two callsites for scheduleFlush:
1. `submit` - this will actually be fine for the resubmitted pre-staged batches, because `resubmitBatch` calls flush at the end
2. `replayPendingStates` - This one is about a corner case where an ID Allocation was pre-staged but there were no other ops that followed. The reason the scheduledFlush call matters is to ensure that op doesn't end up in a batch with ops with a later refSeq.  Likewise, in StagingMode this will sit here until we exit, and then be flushed, and we may get staged and unstaged changes intermingled, which is not supported and crashes the container.

So I believe we do need to plumb through a bit to let the scheduleFlush happen from that second codepath.  Please think carefully about this stuff yourself in case you disagree.  You can look at the commit/PR that addeed that 2nd `scheduleFlush` to learn more about the scenario

(Side note: I have another PR that removes this call, so if that gets approved, it's moot)

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4353,6 +4362,570 @@ describe("Runtime", () => {
...
+						provideEntryPoint: mockProvideEntryPoint,
+					})) as unknown as ContainerRuntime_WithPrivates;
+				}
+
+				afterEach(() => {
+					runtimeWithThreshold?.dispose();
+					runtimeWithThreshold = undefined as unknown as ContainerRuntime_WithPrivates;
+				});
+
+				it("ops accumulate under threshold during staging mode", async () => {
```
markfields, line 4407: I believe this test passes even without the new threshold logic.  This is just testing staging mode.

I think you should add an assert after the 5 submits to say that Outbox has 5 ops and/or PSM has none.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4353,6 +4362,570 @@ describe("Runtime", () => {
...
+					await Promise.resolve();
+
+					assert.equal(
+						submittedOps.length,
+						1,
+						"Op should flush normally when not in staging mode",
+					);
+				});
+
+				it("default threshold suppresses turn-based flushing during staging mode", async () => {
```
markfields, line 4584: Same as above - this is only assrting that submittedOps is empty, not that Outbox has those ops unflushed.

The key difference is "flush" in staging mode _does not submit_.  It merely collects the batch(es) from the Outbox and puts them in the PSM to be replayed upon exiting staging mode.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4353,6 +4362,570 @@ describe("Runtime", () => {
...
+					);
+					// Nothing sent to wire
+					assert.equal(
+						submittedOps.length,
+						0,
+						"No ops should be submitted to wire after discard",
+					);
+				});
+
+				it("enterStagingMode flushes any pending outbox contents as non-staged", async () => {
```
markfields: Does this depend on the threshold?  If not, move out of this describe block

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4353,6 +4362,570 @@ describe("Runtime", () => {
...
+					// Submit more ops while in staging mode
+					submitDataStoreOp(runtimeWithThreshold, "3", genTestDataStoreMessage("staged1"));
+					runtimeWithThreshold.flush();
+					assert.equal(submittedOps.length, 2, "Staged ops should NOT be submitted to wire");
+
+					controls.commitChanges();
+					assert.equal(submittedOps.length, 3, "All ops should be submitted after commit");
+				});
+
+				it("config override takes precedence over runtime option", async () => {
```
markfields, line 4669: IMO this could be simplified to just read the final value of `ContainerRuntime.stagingModeAutoFlushThreshold`, rather than also testing that the threshold is respected (that's the concern of earlier tests)

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4353,6 +4362,570 @@ describe("Runtime", () => {
...
+
+					assert.equal(
+						// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+						(runtimeWithThreshold as any).outbox.mainBatchMessageCount,
+						0,
+						"Outbox should be empty — config override threshold (5) should win over runtime option (50)",
+					);
+				});
+
+				it("runtime option takes precedence over default", async () => {
```
markfields, line 4696: Same comment as previous test

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4353,6 +4362,570 @@ describe("Runtime", () => {
...
+					assert.equal(submittedOps.length, 0, "No ops sent to wire during staging mode");
+					assert.equal(
+						// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+						(runtimeWithThreshold as any).outbox.mainBatchMessageCount,
+						0,
+						"Outbox should be empty — non-runtime op should break the batch",
+					);
+				});
+
+				it("reconnect breaks batch during staging mode", async () => {
```
markfields: Move out of threshold describe block?

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4353,6 +4362,570 @@ describe("Runtime", () => {
...
+					);
+				});
+
+				it("IdAllocation + reconnect while in staging mode does not hit coherency check", async () => {
+					// This is the highest-risk scenario from Mark's test plan.
+					// The fix in b4e1fd1dd25 added scheduleFlush() after submitIdAllocationOpIfNeeded
+					// during replayPendingStates. With threshold suppression, that scheduleFlush()
+					// returns early if in staging mode and under threshold. But the "op" handler
+					// calls flush() directly, so the IdAllocation op still gets flushed before
+					// new ops with different refSeqs are submitted.
```
markfields, line 4751: Are you saying that we don't need that scheduleFlush at all...?  I'll have to come back to this if my other PR that removes that bonus ID Allocation op doesn't get approved, I'm a bit confused trying to think through these strange corner cases.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4353,6 +4362,570 @@ describe("Runtime", () => {
...
+							submitDataStoreOp(
+								runtimeWithThreshold,
+								"someDS",
+								genTestDataStoreMessage({ id: id2 }),
+							),
+						"Should not throw coherency check — IdAllocation op should have been flushed by the 'op' handler",
+					);
+				});
+
+				it("reconnect resubmits pre-staged batches with threshold active", async () => {
```
markfields: Also seems like it doesn't belong in the threshold describe block

markfields (commented):
Looks good overall.  Left some comments for both the agent and human

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: i don't see the harm, and knowing this is data driven is useful to the platform. i don't think it leaks anything, as it is just aggregate statistics we have access too.

packages/runtime/container-runtime/src/pendingStateManager.ts
anthony-murphy: Done - added doc comment explaining the return value.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Done - updated to reference largeBatchThreshold instead of hardcoding 1000.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Done - wrapped the telemetry properties in a details object.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Done - moved exitMethod into the eventName as ExitStagingMode_${exitMethod}.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
anthony-murphy: Done - added assert after the 5 submits to verify the outbox op count.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
anthony-murphy: Done - added outbox count assertion in the default threshold test.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
anthony-murphy: Done - moved the enterStagingMode test out of the threshold describe block.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
anthony-murphy: Done - simplified to just verify the final value of stagingModeAutoFlushThreshold.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
anthony-murphy: Done - simplified the runtime option test similarly.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
anthony-murphy: Done - moved the reconnect test out of the threshold describe block.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
anthony-murphy: Done - moved the reconnect resubmit test out of the threshold describe block.

packages/runtime/container-runtime/src/pendingStateManager.ts
anthony-murphy: The spread here is a Set→Array type conversion (the return type is `IPendingMessage["batchInfo"][]`), not a defensive shallow copy. The Set is used to deduplicate batch infos (multiple messages can share the same `batchInfo` reference), and the spread is the idiomatic way to convert it to an array for the return.

packages/runtime/container-runtime/src/pendingStateManager.ts
anthony-murphy: Same as above — the spread is converting a Set to an Array for the return type, not a defensive copy. The Set provides deduplication of batch infos.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: fixed in #26784

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
anthony-murphy: i think this is resolved by #26784

packages/runtime/container-runtime/src/pendingStateManager.ts
markfields: oh, of course, cool.

packages/runtime/container-runtime/src/pendingStateManager.ts
anthony-murphy: @copilot create an issue for this

markfields (approved):
LGTM!  Looking forward to hearing how this goes in practice.  Makes StagingMode feel simpler.

markfields (approved):

anthony-murphy (approved):

PR #26581, opened 2026-02-27
feat(text-editor): Add devtools initialization to app setup

Josmithr:
Also refactors container creation/load logic to deduplicate some boilerplate.

image


examples/data-objects/text-editor/src/app.tsx
```diff
@@ -106,55 +155,66 @@ async function initFluid(): Promise<DualUserViews> {
...
+		client2,
+		containerId,
+	);
+	const user2Container = container2;
+	const user2View = treeView2;
+
+	console.log(`User 2 connected to document: ${containerId}`);
+
+	// Initialize Devtools
+	initializeDevtools({
```
CraigMacomber, line 196: does having this result in any non trivial performance overhead when processing ops? If so, we probably want to disable it in production builds which we use for perf testing.

Josmithr: It shouldn't by default. You have to explicitly opt into op-related telemetry. And data visualization requests are on demand by the extension. But it also wouldn't hurt to make devtools be predicated on presence/absence of an environment variable or something. What flag does the performance testing look for?

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -79,18 +81,65 @@ interface DualUserViews {
 	containerId: string;
 }
 
+async function createAndAttachNewContainer(client: AzureClient): Promise<{
+	container: IFluidContainer;
+	containerId: string;
+	treeView: TreeView<typeof TextEditorRoot>;
+}> {
+	const { container } = await client.createContainer(containerSchema, "2");
```
CraigMacomber, line 93: I know it was this way before your change, but an undocumented "2" is a pretty odd choice of magic string. seems like using a `const containerName = "text demo";` or something like that might make more sense.

Josmithr: This isn't the container name, it's the compatibility mode.

Josmithr: https://fluidframework.com/docs/api/fluid-static/compatibilitymode-typealias

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -79,18 +81,65 @@ interface DualUserViews {
...
+
+	const treeView = container.initialObjects.tree.viewWith(treeConfig);
```
CraigMacomber, line 95: Did you add a way to make a container with a root tree instead of the full containerSchema and initialObjects thing? Maybe not this PR, but would be nice to not have the extra cruft wrapping our tree.

Josmithr: The API for that is currently internal, and the general consensus was that it wasn't what we wanted long-term. We could use it, but it would make the code less useful as an example.

examples/data-objects/text-editor/src/app.tsx
CraigMacomber: And that's what I get for reading code without inteliSense.

Josmithr: Also a downside of magic strings in general. IMO, the `createContainer` API would be better if it took its arguments as a property bag so these values could be labeled.

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -106,55 +155,66 @@ async function initFluid(): Promise<DualUserViews> {
 		containerId = rawContainerId;
 		console.log(`Loading document for both users: ${containerId}`);
 
-		const { container: container1 } = await client1.getContainer(
+		// User 1 connects to existing document
+		const { container: container1, treeView: treeView1 } = await loadExistingContainer(
```
CraigMacomber: I believe you can use de-structuring directly if you use () to make it an expression.

```
({ container: user1Container, treeView: user1View } = await loadExistingContainer(client1, containerId ));
```

CraigMacomber: Same below

examples/data-objects/text-editor/src/app.tsx
CraigMacomber: I don't mean performance tests/benchmarks. I mean developers running the app with a production webpack bundle and measuring the performance in the browser.

If you did want to disable this in production builds, there are two approaches: doing it dynamically (for example by checking nonProductionConditionalsIncluded) or if you wanted to save bundle size and not include the code in production builds, we would need a new API, expose skipInProduction, or kludge it into a debugAssert callback.

CraigMacomber: This sounds fins as is though. If there is overhead, it will show up and we can deal with it then.

examples/data-objects/text-editor/src/app.tsx
CraigMacomber: Someday we need to have an API we like it when people use it, but today is not that day.

examples/data-objects/text-editor/src/app.tsx
Josmithr: I forgot that syntax was an option! Good call. Updated.

CraigMacomber (approved):

CraigMacomber (approved):

PR #26583, opened 2026-02-27
test: increase ReconnectFarm timeout from 30s to 60s

frankmueller-msft:
## Summary

Increases the timeout for the `ReconnectFarm` stress tests from 30s to 60s. These tests run randomized merge-tree operations with reconnection and can intermittently exceed the 30s timeout on CI machines under load, causing flaky `Timeout of 30000ms exceeded` failures.

Observed in build 379358: `MergeTree.Client > ReconnectFarm_2_0` timed out with 1554 other tests passing.

## Coverage audit

No change to what the test validates — same operations, same assertions, same seed-based determinism. Only the timeout threshold changed.

## Test plan

- [x] Change is a single constant: `30 * 1000` → `60 * 1000`

### DIFF
--- a/packages/dds/merge-tree/src/test/client.reconnectFarm.spec.ts
+++ b/packages/dds/merge-tree/src/test/client.reconnectFarm.spec.ts
@@ -120,7 +120,7 @@ function runReconnectFarmTests(opts: IReconnectFarmConfig, extraSeed?: number): 
 				testOpts,
 				applyMessagesWithReconnect,
 			);
-		}).timeout(30 * 1000);
+		}).timeout(60 * 1000);
 	});
 }
 


anthony-murphy:
what is the impetus for this change? I'm not aware of these tests being flaky in the PR build or CI? is there any associated issue? In general, i don't like just increasing timeouts, as it can lead to performance regression over time.

i managed to find [379358](https://dev.azure.com/fluidframework/public/_build/results?buildId=379358&view=results), but that is a topic branch run, not a run on main. i think we need more evidence that this is all a net good, before we make changes. I worry that our build machines are under resourced for the parallelism, and we are trading time for stability, which is not a good trade off.

frankmueller-msft:
Fair points. To give context: the timeout failure in build 379358 was on the `ci/combined-pipeline-parallelization` topic branch where I was developing #26586 (splitting DDS coverage shards). That PR runs the farm/fuzz tests with `mocha --parallel`, which adds CPU contention and could push these long-running fuzz tests closer to the 30s boundary.

I haven't seen this fail on main, so the evidence for a standalone timeout increase is weak. The right path is probably to close this and monitor after #26586 merges — if `--parallel` causes ReconnectFarm timeouts, the timeout increase should be part of that PR, not standalone.

Closing this.

PR #26585, opened 2026-02-27
ci: skip bundle-size comparison until token is rotated (AB#56981)

frankmueller-msft:
## Summary

Disables the bundle-size comparison step (`bundle-analysis:run`) with `condition: false` until the `githubPublicRepoSecret` token is rotated in `prague-key-vault` ([AB#56981](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/56981)).

The step was failing with 401/forbidden on every PR build, and `continueOnError: true` was masking it as `partiallySucceeded` — hiding real failures behind the same yellow status.

### What changed

- Set `condition: false` so the step is visibly skipped, not silently swallowed
- Added `(DISABLED)` to the display name
- Added `TODO([AB#56981](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/56981))` comment for re-enabling
- Preserved the original condition as a comment for easy restoration
- Kept `continueOnError: true` as a safety net for when the step is re-enabled

### Re-enabling

When the token is rotated, restore the original condition and remove `(DISABLED)` from the display name:
```yaml
# condition: false
condition: and(succeeded(), eq(variables['Build.Reason'], 'PullRequest'))
```

## Test plan

- [x] YAML is valid
- [x] PR builds should show the step as skipped (not partially succeeded)
- [x] No functional changes to any other pipeline steps

### DIFF
--- a/tools/pipelines/templates/build-npm-client-package.yml
+++ b/tools/pipelines/templates/build-npm-client-package.yml
@@ -437,11 +437,11 @@ extends:
                           echo "Bundle size comparison completed successfully."
                         else
                           exit_code=$?
-                          if grep -qi "Bad credentials\|401\|authentication\|DANGER_GITHUB_API_TOKEN" /tmp/bundle-analysis.log; then
-                            echo "##[warning]Bundle size comparison failed due to GitHub authentication (exit code $exit_code). The GitHub token (githubPublicRepoSecret) may need to be rotated."
-                            echo "##[warning]Skipping bundle size comparison — this does not affect build correctness."
+                          if grep -q "Bad credentials" /tmp/bundle-analysis.log; then
+                            echo "##[warning]Bundle size comparison skipped — GitHub token (githubPublicRepoSecret) returned 'Bad credentials'. Token may need rotation."
                           else
-                            echo "##[warning]Bundle size comparison failed (exit code $exit_code). See log above for details."
+                            echo "##[error]Bundle size comparison failed (exit code $exit_code). See log above."
+                            exit 1
                           fi
                         fi
 
--- a/tools/pipelines/templates/build-npm-client-package.yml
+++ b/tools/pipelines/templates/build-npm-client-package.yml
@@ -437,8 +437,8 @@ extends:
                           echo "Bundle size comparison completed successfully."
                         else
                           exit_code=$?
-                          if grep -q "Bad credentials" /tmp/bundle-analysis.log; then
-                            echo "##[warning]Bundle size comparison skipped — GitHub token (githubPublicRepoSecret) returned 'Bad credentials'. Token may need rotation."
+                          if grep -qE "Bad credentials|forbids access via a personal access token" /tmp/bundle-analysis.log; then
+                            echo "##[warning]Bundle size comparison skipped — GitHub token (githubPublicRepoSecret) was rejected. Token may need rotation or its lifetime may exceed enterprise policy limits."
                           else
                             echo "##[error]Bundle size comparison failed (exit code $exit_code). See log above."
                             exit 1
--- a/tools/pipelines/templates/build-npm-client-package.yml
+++ b/tools/pipelines/templates/build-npm-client-package.yml
@@ -421,18 +421,29 @@ extends:
                   # At this point we want to publish the artifact with the bundle size analysis,
                   # but as part of 1ES migration that's now part of templateContext.outputs below.
 
-                  - task: Npm@1
+                  - task: Bash@3
                     displayName: (PR only) Compare bundle sizes against baseline
                     condition: and(succeeded(), eq(variables['Build.Reason'], 'PullRequest'))
-                    continueOnError: true
                     env:
                       ADO_API_TOKEN: $(System.AccessToken)
                       DANGER_GITHUB_API_TOKEN: $(githubPublicRepoSecret)
                       TARGET_BRANCH_NAME: '$(targetBranchName)'
                     inputs:
-                      command: custom
-                      workingDir: '$(Pipeline.Workspace)/${{ parameters.buildDirectory }}'
-                      customCommand: 'run bundle-analysis:run'
+                      targetType: inline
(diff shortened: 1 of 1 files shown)


tools/pipelines/templates/build-npm-client-package.yml
```diff
@@ -421,18 +421,29 @@ extends:
...
                       DANGER_GITHUB_API_TOKEN: $(githubPublicRepoSecret)
                       TARGET_BRANCH_NAME: '$(targetBranchName)'
                     inputs:
-                      command: custom
-                      workingDir: '$(Pipeline.Workspace)/${{ parameters.buildDirectory }}'
-                      customCommand: 'run bundle-analysis:run'
+                      targetType: inline
+                      workingDirectory: '$(Pipeline.Workspace)/${{ parameters.buildDirectory }}'
+                      script: |
+                        set -uo pipefail
```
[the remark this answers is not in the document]
frankmueller-msft: Good catch — both concerns here were addressed in the follow-up commit ("Only swallow auth failures; let other failures propagate"):

1. **`set -e` not needed**: The script uses explicit `if/else/exit` flow control. The `if npm run ... | tee ...; then` block captures the exit code directly, and the `else` branch either warns (auth failure) or `exit 1` (all other failures). `set -uo pipefail` ensures the pipeline and undefined variables are handled correctly.

2. **Exit code capture**: The `if` statement captures success/failure directly. On failure, `$?` is captured into `exit_code` for the error message, and non-auth failures now correctly `exit 1` instead of being silently swallowed.

tools/pipelines/templates/build-npm-client-package.yml
```diff
@@ -421,18 +421,29 @@ extends:
...
+                        if npm run bundle-analysis:run 2>&1 | tee /tmp/bundle-analysis.log; then
+                          echo "Bundle size comparison completed successfully."
+                        else
+                          exit_code=$?
+                          if grep -qi "Bad credentials\|401\|authentication\|DANGER_GITHUB_API_TOKEN" /tmp/bundle-analysis.log; then
```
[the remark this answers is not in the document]
frankmueller-msft: Agreed — this was addressed in the second commit. The grep pattern was narrowed from the broad `"Bad credentials\|401\|authentication\|DANGER_GITHUB_API_TOKEN"` to just `"Bad credentials"`, which is the specific error string GitHub returns for expired/invalid PATs. Non-auth failures now correctly propagate via `exit 1`.

ChumpChief:
I generally don't like hiding a failed step behind a green check, as it gives a false sense of security that things are working correctly.  If the goal is to get to a green check in the rollup, maybe consider just skipping the step for now until we can investigate AB#56981 - this should then roll up to a green check for the job but make it more obvious at a glance that this step is not passing.

frankmueller-msft:
Good call — updated to just skip the step entirely with `condition: false`. The step now shows as visibly skipped in the pipeline instead of masking behind a green check. Restoring it is a one-line change once the token is rotated.

alexvy86:
> Good call — updated to just skip the step entirely with `condition: false`. The step now shows as visibly skipped in the pipeline instead of masking behind a green check. Restoring it is a one-line change once the token is rotated.

Seems like the change wasn't pushed.

frankmueller-msft:
Sorry about that — the commit was pushed to the wrong remote. It's there now. Ready for your review.

tools/pipelines/templates/build-npm-client-package.yml
```diff
@@ -421,10 +421,11 @@ extends:
...
                   # but as part of 1ES migration that's now part of templateContext.outputs below.
 
+                  # TODO(AB#56981): Re-enable once githubPublicRepoSecret is rotated in prague-key-vault.
+                  # Disabled because the token is expired/rejected.
                   - task: Npm@1
-                    displayName: (PR only) Compare bundle sizes against baseline
-                    condition: and(succeeded(), eq(variables['Build.Reason'], 'PullRequest'))
-                    continueOnError: true
+                    displayName: (PR only) Compare bundle sizes against baseline (DISABLED)
+                    condition: false
```
alexvy86, line 429: Can we leave the previous condition here as a comment so the change is easier to undo once things are fixed? Instead of having to find the PR. Also, I think we should keep the `continueOnError: true`.

alexvy86, line 429: Can we leave the previous condition here as a comment so the change is easier to undo once things are fixed? Instead of having to find the PR. Also, I think we should keep the `continueOnError: true`.

frankmueller-msft: Done — added the original condition as a comment and restored `continueOnError: true`.

frankmueller-msft: (Duplicate — addressed above.)

tools/pipelines/templates/build-npm-client-package.yml
```diff
@@ -421,9 +421,12 @@ extends:
...
                   # but as part of 1ES migration that's now part of templateContext.outputs below.
 
+                  # TODO(AB#56981): Re-enable once githubPublicRepoSecret is rotated in prague-key-vault.
+                  # Disabled because the token is expired/rejected.
+                  # Original condition: and(succeeded(), eq(variables['Build.Reason'], 'PullRequest'))
                   - task: Npm@1
-                    displayName: (PR only) Compare bundle sizes against baseline
-                    condition: and(succeeded(), eq(variables['Build.Reason'], 'PullRequest'))
+                    displayName: (PR only) Compare bundle sizes against baseline (DISABLED)
+                    condition: false
```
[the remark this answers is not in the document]
frankmueller-msft: PR description has been updated with accurate re-enabling instructions — includes restoring the original condition and removing `(DISABLED)` from the display name.

alexvy86 (approved):

PR #26587, opened 2026-02-27
Move text-editor views to react package

CraigMacomber:
## Description

This is an incremental step toward exporting text editor react components.

Once I moved the views out of the text-editor package, it became clear the remaining code had no tests, so I added a top level test.

When sorting out the dependencies I noticed we were using an old version of delta, and skip lib check to suppress errors in use of delta. I reworked the workarounds and better documented them.

## Reviewer Guidance

### DIFF
--- a/packages/framework/react/src/test/text/textEditor.test.tsx
+++ b/packages/framework/react/src/test/text/textEditor.test.tsx
@@ -10,7 +10,7 @@ import { TreeAlpha } from "@fluidframework/tree/alpha";
 import { independentView, TextAsTree } from "@fluidframework/tree/internal";
 import { render } from "@testing-library/react";
 import globalJsdom from "global-jsdom";
-import Delta from "quill-delta";
+import DeltaPackage from "quill-delta";
 import * as React from "react";
 
 import { toPropTreeNode } from "../../propNode.js";
@@ -32,6 +32,10 @@ import {
 	// eslint-disable-next-line import-x/no-internal-modules
 } from "../../text/plain/index.js";
 import { UndoRedoStacks } from "../../undoRedo.js";
+
+// Workaround for quill-delta's export style not working well with node16 module resolution.
+type Delta = DeltaPackage.default;
+const Delta = DeltaPackage.default;
 
 // Configuration for creating formatted text views
 const formattedTreeConfig = new TreeViewConfiguration({ schema: FormattedTextAsTree.Tree });
--- a/packages/framework/react/src/test/text/textEditor.test.tsx
+++ b/packages/framework/react/src/test/text/textEditor.test.tsx
@@ -5,16 +5,15 @@
 
 import { strict as assert } from "node:assert";
 
-import { toPropTreeNode } from "@fluidframework/react/alpha";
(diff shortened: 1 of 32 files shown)


packages/framework/react/src/text/formatted/quillFormattedView.tsx
```diff
@@ -358,8 +355,12 @@ const FormattedTextEditorView = React.forwardRef<
 				let utf16Pos = 0; // Position in UTF-16 code units (Quill's view)
 				let cpPos = 0; // Position in codepoints (tree's view)
 
-				for (const op of (delta as QuillDelta).ops ?? []) {
+				for (const op of delta.ops) {
 					if (op.retain !== undefined) {
+						// The docs for retain imply this is always a number, but the type definitions allow a record here.
+						// It is unknown why the type definitions allow a record as the have no doc comments.
```
brrichards:
```suggestion
						// It is unknown why the type definitions allow a record as they have no doc comments.
```

packages/framework/react/src/test/text/textEditor.test.tsx
```diff
@@ -64,9 +70,21 @@ const views: { name: string; component: React.FC<MainViewProps> }[] = [
...
-	// since Quill requires document at import time. DOM state may accumulate
-	// across tests. See src/test/mochaHooks.mjs.
+	// Note: JSDOM is initialized once in mochaHooks.ts before Quill is imported,
+	// since Quill requires document at import time. See src/test/mochaHooks.ts.
+	// These tests reset up a clean DOM.
+
+	let cleanup: () => void;
+
+	// TODO: why does making this beforeEach/afterEach instead of before/after cause cleanup to crash?
+	// It seems like each test should be able to have its own clean DOM.
```
brrichards, line 80: Maybe it has something to do with how Quill registers itself on the document at import time and holds references to the DOM nodes. Tearing down JSDOM between each test, destroying the document out from under Quills registered state, would cause a crash when the next test tries to up a new JSDOM and Quill still has stale references.

CraigMacomber: I don't think its an issue with Quill. The error is window not existing when doing the teardown after the first test, even on non- quill tests. This makes no sense to me since we can do the same cleanup after running multiple tests, but not after running one test.

CraigMacomber: I can look into it more later (like removing all but one test so they should be the same), but I decided it was out of scope for this change.

syncpack.config.cjs
```diff
@@ -200,6 +200,10 @@ module.exports = {
 				"react-virtualized-auto-sizer",
 				"@types/react",
 				"@types/react-dom",
+				// "quill-delta" 5 does not support CommonJS so @fluid-experimental/ot is stuck on version 4 for its dev-dep.
+				// Other use of quill delta is in  our react package for test editing, which is ESM only and uses up to date quill packages.
```
brrichards:
```suggestion
				// Other use of quill delta is in our react package for text editing, which is ESM only and uses up to date quill packages.
```

brrichards (approved):

PR #26594, opened 2026-03-01
Parallelize webpack inside ci:build task graph

frankmueller-msft:
## Summary

- Adds `webpack` to `ci:build.dependsOn` in `fluidBuild.config.cjs` so fluid-build can schedule it in parallel with lint, api-reports, and docs
- Removes the separate sequential `npm run webpack` step from `build-npm-client-package.yml` since it's now handled by `ci:build`

Webpack only depends on `^tsc` and `^build:esnext` (defined at line 154 of `fluidBuild.config.cjs`), not on lint or api-reports, so it can start as soon as compilation finishes rather than waiting for all of `ci:build` to complete first.

**Expected savings:** ~1-2 minutes on the build job critical path.

## Test plan

- [ ] CI build passes (webpack output is identical whether run inside or outside ci:build)
- [ ] Verify webpack artifacts are still included in the build output archive
- [ ] Compare build job duration against baseline (~19 min)

### DIFF
--- a/fluidBuild.config.cjs
+++ b/fluidBuild.config.cjs
@@ -39,6 +39,7 @@ module.exports = {
 				"ci:build:docs",
 				"build:manifest",
 				"build:readme",
+				"webpack",
 			],
 			script: false,
 		},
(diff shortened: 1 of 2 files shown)


frankmueller-msft:
Closing — CI results show no measurable build time improvement. The build job ran in 17m 39s, which is within normal variance of recent main builds (17m 31s – 19m 18s). Webpack was likely already being scheduled in parallel by fluid-build, so making the dependency explicit didn't change the critical path.

PR #26606, opened 2026-03-02
[Claude code generated PR] refactor(historian-base): extract createRouteContext helper

dhr-verma:
## Summary
- Extracted the duplicated 4-line router/throttle prologue shared across 8 Historian route modules into a single `createRouteContext()` helper in `routes/utils.ts`
- Added `IHistorianRouteContext` interface for the returned context object
- `summaries.ts` uses a destructuring alias (`tenantThrottleOptions: tenantGeneralThrottleOptions`) to preserve its existing local variable name without touching downstream code
- Removed now-unused `Router`, `IThrottleMiddlewareOptions`, and `Constants` imports from the 8 standard route files

## Test plan
- [x] `pnpm run build:compile` passes (root)
- [x] `npm run test` passes in `historian-base` — 51 tests passing (6 new unit tests for `createRouteContext`)
- [x] New unit tests cover: router shape, `maxTokenLifetimeSec` value, throttle options shape, throttler map wiring, absent-key fallback to `0`

## Related
Closes #26591

## Review notes
- When `maxTokenLifetimeSec` is absent from config, the helper falls back to `0`. In `verifyToken` this causes all tokens to be rejected (safe-fail). A startup warning could be added in a follow-up if desired.
- The `throttleIdPrefix` test uses a defensive ternary; a future cleanup could split it into an explicit type assertion + direct call for clearer failure messages.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


zhangxin511 (approved):
Looks good, lease hace @znewton  for another eye

zhangxin511 (approved):

PR #26613, opened 2026-03-02
Update container-views example to match external-views infrastructure

ChumpChief:
### container-views restructuring
- Restructure `src/` into `src/container/` subdirectory matching external-views layout (`diceRoller/`, `view.tsx`, `container/index.ts`)
- Replace custom `IContainerView` interface with `ContainerViewRuntimeFactory`, `IFluidMountableViewEntryPoint`, and `MountableView` from `@fluid-example/example-utils` (matching the diceroller example pattern)
- Use lower-level `FluidDataStoreRuntime` and `IFluidDataStoreFactory` for the DiceRoller data store (instead of `DataObject`/`DataObjectFactory` from aqueduct)
- Remove unused `src/index.ts` package entry point and `@fluidframework/aqueduct` dependency
- Add multi-service support (local/t9s/odsp) via `@fluid-example/example-driver` and `@fluid-example/example-webpack-integration`
- Switch to default html-webpack-plugin template (remove `src/index.html`)
- Modernize tests to use puppeteer directly with manual browser lifecycle management
- Add cross-container event propagation test ("raises an event in other connected containers")
- Update eslint config from `minimalDeprecated` to `recommended`

### Align all three dice roller examples (container-views, external-views, diceroller)
- Apply external-views dice styling (dynamic color, centered layout, larger fonts) to container-views and diceroller
- Use named React imports (`createElement`, `FC`, `useState`, `useEffect`) instead of default `React.xxx` style
- Use `diceRoller` as the view prop name with destructured parameters
- Add explicit `(): void` return type on useEffect cleanup functions
- Add `"react"` to tsconfig types in external-views
- Fix missing `const` in diceroller `jest.config.cjs`
- Remove unused dependencies: `uuid` from diceroller; `uuid`, `process` from external-views
- Remove dead `main`/`types` fields from external-views `package.json`
- Fix description in external-views: "Data Object" → "data store"
- Fix typo in diceroller: "rolls the device" → "rolls the dice"
- Align test names across examples
- Simplify `DiceRollerFactory.type` to a `readonly` property

## Test plan

- [x] `pnpm run build` passes (biome, tsc, eslint)
- [x] `pnpm run test:jest` passes (puppeteer tests validate button loads and cross-container sync)
- [x] Manual test: `pnpm run start:local` — app loads at localhost:8080, dice rolls work
- [x] Manual test: open a second browser tab with the same container ID — dice rolls sync


ChumpChief:
Human note: This PR updates container-views in the same way that I've updated external-views and diceroller in the past, so most of the changes are there (esp. app.ts).

The changes in external-views and diceroller are more minor or stylistic changes to better align the three examples.

examples/view-integration/container-views/webpack.config.cjs
```diff
@@ -34,17 +38,26 @@ module.exports = (env) => {
...
 		plugins: [
 			new webpack.ProvidePlugin({
 				process: "process/browser.js",
 			}),
 			new HtmlWebpackPlugin({
-				template: "./src/index.html",
+				title: "app-integration-container-views",
 			}),
+			createExampleDriverServiceWebpackPlugin(service),
 		],
```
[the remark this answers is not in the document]
ChumpChief: Will consider options for a subsequent change.  I don't want to default to a service because that feels prone to hiding bugs.  Most likely would be better to update the webpack script to specify the service and adding a throw here in the config if not defined.

examples/apps/diceroller/src/container/main.tsx
```diff
@@ -6,7 +6,7 @@
 import type { EventEmitter } from "@fluid-example/example-utils";
 import { DataObject, DataObjectFactory } from "@fluidframework/aqueduct/legacy";
 import type { IValueChanged } from "@fluidframework/map/legacy";
-import React from "react";
+import React, { type FC, useEffect, useState } from "react";
```
shlevari, line 9: Do you still want the React import as well?

examples/view-integration/container-views/src/container/view.tsx
```diff
@@ -0,0 +1,41 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import React, { type FC, useEffect, useState } from "react";
```
shlevari, line 6: Same question, why the mix of specific and general imports

examples/view-integration/container-views/README.md
```diff
@@ -43,9 +45,9 @@ For in browser testing update `./jest-puppeteer.config.js` to:
 
 ## Data model
 
-Dice Roller uses the following distributed data structures:
+DiceRoller uses the following distributed data structures:
 
--   SharedDirectory - root
+-   SharedMap
```
shlevari, line 50: In general, seems like a good follow up to probably switch our examples to SharedTree since that is what we'd like all new customers to use. But obviously not needed as part of this PR

examples/apps/diceroller/src/container/main.tsx
ChumpChief: Ideally we'd remove it, but without the `React` import TS currently chokes on the JSX.  I believe that with a combination of [tsconfig options](https://www.typescriptlang.org/tsconfig/#jsx) and maybe a disable a [lint rule](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md) this might be ok (since we are on a newer version of React that should be using the [new JSX transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports))?  At least in my personal projects I use `"jsx": "react-jsx"` and omit the `React` import with no issues.

But if we make that change we should probably do it separately as a uniform update across all the examples.

alexvy86 (approved):
Infra changes look good, in line with the ones in the other examples. Shahar's comments are worth addressing IMO. Since it's a relatively mechanical change replicating what you've done before I didn't scrutinize every line, but overall it looked fine.

ChumpChief:
> Infra changes look good, in line with the ones in the other examples. Shahar's comments are worth addressing IMO. Since it's a relatively mechanical change replicating what you've done before I didn't scrutinize every line, but overall it looked fine.

I'm actually attempting an update to use React's new JSX transform at this very moment :)

PR #26614, opened 2026-03-02
feat(build-cli): support pnpm catalog references in flub release report

alexvy86:
## Description

`flub release report` (and the release state machine) would throw `TypeError: Invalid comparator: catalog:X` when any package in the repo used pnpm catalog protocol references in its dependencies.

This PR adds a `catalog.ts` utility that reads `pnpm-workspace.yaml` and resolves `catalog:` / `catalog:default` / `catalog:X` references to their actual version ranges before any semver operations. Both `getFluidDependencies` and `getPreReleaseDependencies` in `package.ts` now resolve catalog references via a per-workspace-root cache, making `getFluidDependencies` async. The `release report` command is updated to await the now-async call.

Includes a `testRepoCatalog` fixture and unit/integration tests that cover the catalog resolution logic end-to-end.

This will enable us to use pnpm catalogs to simplify our version ranges across the repo.

## Reviewer Guidance

### DIFF
--- a/build-tools/packages/build-cli/src/library/catalog.ts
+++ b/build-tools/packages/build-cli/src/library/catalog.ts
@@ -1,0 +1,87 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { readFile } from "node:fs/promises";
+import path from "node:path";
+
+import YAML from "yaml";
+
+/**
+ * Map of catalogName → { packageName → versionRange }.
+ * The default catalog (defined under `catalog:` in pnpm-workspace.yaml) is stored under the key "default".
+ */
+export type PnpmCatalogMap = Map<string, Record<string, string>>;
+
+/**
+ * Reads the pnpm-workspace.yaml at `workspaceRoot` and returns the catalog entries.
+ * Returns an empty map if the file doesn't exist or has no catalogs.
+ */
+export async function readPnpmCatalogs(workspaceRoot: string): Promise<PnpmCatalogMap> {
+	const catalogMap: PnpmCatalogMap = new Map();
+	const yamlPath = path.join(workspaceRoot, "pnpm-workspace.yaml");
+
+	let content: string;
+	try {
+		content = await readFile(yamlPath, "utf-8");
+	} catch {
+		return catalogMap;
+	}
+
+	const workspace = YAML.parse(content) as {
+		catalog?: Record<string, string>;
+		catalogs?: Record<string, Record<string, string>>;
+	};
+
+	if (workspace.catalog !== undefined) {
+		catalogMap.set("default", workspace.catalog);
+	}
+
+	if (workspace.catalogs !== undefined) {
+		for (const [name, entries] of Object.entries(workspace.catalogs)) {
+			catalogMap.set(name, entries);
+		}
+	}
+
+	return catalogMap;
+}
+
+/**
+ * If `version` starts with "catalog:", looks up the real version from `catalogs`.
+ * Otherwise returns `version` unchanged.
+ *
+ * Resolution rules (per pnpm spec):
+ * - `catalog:` or `catalog:default` → look in the "default" entry (the `catalog:` section)
+ * - `catalog:X` → look in the `catalogs.X` entry
+ *
+ * @throws If the referenced catalog or package is not found.
+ */
+export function resolveCatalogVersion(
+	packageName: string,
+	version: string,
+	catalogs: PnpmCatalogMap,
+): string {
+	if (!version.startsWith("catalog:")) {
+		return version;
+	}
+
+	const catalogRef = version.slice("catalog:".length);
+	const catalogName = catalogRef === "" || catalogRef === "default" ? "default" : catalogRef;
+
+	const catalog = catalogs.get(catalogName);
+	if (catalog === undefined) {
+		throw new Error(
+			`pnpm catalog "${catalogName}" not found when resolving "${packageName}". Available catalogs: ${[...catalogs.keys()].join(", ") || "(none)"}`,
+		);
+	}
+
+	const resolved = catalog[packageName];
+	if (resolved === undefined) {
+		throw new Error(
+			`Package "${packageName}" not found in pnpm catalog "${catalogName}"`,
+		);
+	}
+
+	return resolved;
+}
--- a/build-tools/packages/build-cli/src/library/package.ts
+++ b/build-tools/packages/build-cli/src/library/package.ts
@@ -36,9 +36,13 @@ import {
 } from "../filter.js";
 import { isReleaseGroup, type ReleaseGroup, type ReleasePackage } from "../releaseGroups.js";
 import type { DependencyUpdateType } from "./bump.js";
-import { readPnpmCatalogs, resolveCatalogVersion } from "./pnpmCatalog.js";
 import { zip } from "./collections.js";
 import type { Context, VersionDetails } from "./context.js";
+import {
+	type PnpmCatalogMap,
+	readPnpmCatalogs,
+	resolveCatalogVersion,
+} from "./pnpmCatalog.js";
 import { indentString } from "./text.js";
 
 /**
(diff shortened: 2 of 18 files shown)


build-tools/packages/build-cli/src/test/library/package.test.ts
```diff
@@ -128,3 +135,59 @@ describe("typeTestUtils", () => {
 		});
 	});
 });
+
+describe("resolveCatalogVersion", () => {
+	it("returns version unchanged for non-catalog strings", () => {
+		const catalogs: PnpmCatalogMap = new Map([["default", { "pkg-a": "^1.0.0" }]]);
+		assert.equal(resolveCatalogVersion("pkg-a", "^1.0.0", catalogs), "^1.0.0");
```
alexvy86: Since they are the same, this test doesn't make it clear that the resolution is happening to the hardcoded version, not the one from the catalogs.

build-tools/packages/build-cli/src/library/catalog.ts
```diff
@@ -0,0 +1,87 @@
...
+	packageName: string,
+	version: string,
+	catalogs: PnpmCatalogMap,
+): string {
+	if (!version.startsWith("catalog:")) {
+		return version;
+	}
+
+	const catalogRef = version.slice("catalog:".length);
+	const catalogName = catalogRef === "" || catalogRef === "default" ? "default" : catalogRef;
```
alexvy86: The second part of the `or` is redundant, isn't it?

build-tools/packages/build-cli/src/library/pnpmCatalog.ts
alexvy86, line 1: Rename this file to pnpmCatalog.ts to make things a bit clearer.

build-tools/packages/build-cli/src/library/pnpmCatalog.ts
```diff
@@ -0,0 +1,87 @@
...
+export async function readPnpmCatalogs(workspaceRoot: string): Promise<PnpmCatalogMap> {
+	const catalogMap: PnpmCatalogMap = new Map();
+	const yamlPath = path.join(workspaceRoot, "pnpm-workspace.yaml");
+
+	let content: string;
+	try {
+		content = await readFile(yamlPath, "utf-8");
+	} catch {
+		return catalogMap;
+	}
```
alexvy86, line 33: Let's improve this: if the file doesn't exist, it's fine to return an empty catalog map, but if it exists and somethings fail when reading it, we should let that error propagate.

build-tools/packages/build-cli/src/test/library/package.test.ts
```diff
@@ -128,3 +135,59 @@ describe("typeTestUtils", () => {
...
+	it("resolves named catalog:mycat correctly", () => {
+		const catalogs: PnpmCatalogMap = new Map([["mycat", { "pkg-a": "^2.0.0" }]]);
+		assert.equal(resolveCatalogVersion("pkg-a", "catalog:mycat", catalogs), "^2.0.0");
+	});
+
+	it("throws on unknown catalog", () => {
+		const catalogs: PnpmCatalogMap = new Map();
+		assert.throws(
+			() => resolveCatalogVersion("pkg-a", "catalog:nonexistent", catalogs),
+			/nonexistent/,
```
alexvy86: Can we be more specific with the error being checked here?

build-tools/packages/build-cli/src/test/library/package.test.ts
```diff
@@ -128,3 +135,59 @@ describe("typeTestUtils", () => {
...
+			"3.0.0",
+			"Expected pkg-c version resolved from catalog: to be 3.0.0",
+		);
+	});
+
+	it("throws for unknown catalog", () => {
+		const catalogs: PnpmCatalogMap = new Map();
+		assert.throws(
+			() => resolveCatalogVersion("pkg-a", "catalog:buildTools", catalogs),
+			/buildTools/,
```
alexvy86: Same as above, can we be more specific with the error message?

tylerbutler:
Thanks for the PR — I reviewed this with focus on correctness, reliability, and security/perf. I found two actionable items:

1. **Misleading parse error after catalog resolution** (`build-tools/packages/build-cli/src/library/package.ts`)
   - In `getFluidDependencies`, parsing is done on `resolvedVersion`, but the error still reports `dep.version`:
     - current: `throw new Error(\`Failed to parse depVersion: ${dep.version}\`)`
   - This can hide the real bad value when a `catalog:` entry resolves to an invalid semver/range.
   - Suggested fix: include `resolvedVersion` (and optionally original `dep.version` for context).

2. **Test coverage gaps for new catalog behavior**
   - Missing direct unit tests for `readPnpmCatalogs` in `pnpmCatalog.ts` (especially malformed YAML + non-ENOENT file errors).
   - Missing `resolveCatalogVersion` test for "catalog exists but package key is missing" (`Package "..." not found in pnpm catalog ...`).
   - Missing coverage for catalog-resolution path in `getPreReleaseDependencies` (PR added behavior there but no dedicated test).

Security/performance-wise, I didn’t find any major concerns; the caching + async propagation look good.

build-tools/packages/build-cli/src/library/package.ts
```diff
@@ -284,17 +285,30 @@ export async function getPreReleaseDependencies(
 		);
 	}
 
+	// Cache catalogs per workspace root to avoid re-reading the file for each package.
+	const catalogCache = new Map<string, Awaited<ReturnType<typeof readPnpmCatalogs>>>();
```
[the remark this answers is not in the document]
tylerbutler: +1

build-tools/packages/build-cli/src/library/package.ts
```diff
@@ -434,10 +448,10 @@ export function filterVersionsOlderThan(
  * @returns A tuple of {@link PackageVersionMap} objects, one of which contains release groups on which the package
  * depends, and the other contains independent packages on which the package depends.
  */
-export function getFluidDependencies(
+export async function getFluidDependencies(
```
tylerbutler: Does this need to be async really? Everything needed should be in memory, no?

build-tools/packages/build-cli/src/library/pnpmCatalog.ts
```diff
@@ -0,0 +1,90 @@
...
+	try {
+		content = await readFile(yamlPath, "utf-8");
+	} catch (error) {
+		if ((error as NodeJS.ErrnoException).code === "ENOENT") {
+			return catalogMap;
+		}
+		throw error;
+	}
+
+	const workspace = YAML.parse(content) as {
```
tylerbutler: satisfies instead?

tylerbutler: Or type the workspace itself, i.e.

```ts
const workspace: {catalog?: Record<string, string>;
		catalogs?: Record<string, Record<string, string>>;
	} = YAML.parse(content);
```

tylerbutler (commented):
Seems good though I don't like the async change. I'm surprised this is the only place that needs to understand the catalogs, though. Won't other tools continue to be confused, including vs code?

build-tools/packages/build-cli/src/library/package.ts
alexvy86: It needs to read the pnpm-workspace.yml file for the catalog. But since it's build-tools we can do it synchronously, I'll make the change.

alexvy86:
> Won't other tools continue to be confused, including vs code?

Do you have an idea of what exactly would confuse VSCode? The only error that surfaced (and we noticed) when we tried pnpm catalogs last time was this issue with `flub release report` so that's what I focused on for now. I had Claude look for other potential problem codepaths and it called out `getTscCommandDependencies` which should be handled now, and I'm working through `flub bump --updateAllDeps` which seems like ideally it would want to update the entry in the catalog, but I'm wondering if we can just make it skip `catalog:` dependencies and log warnings about them for now. Thoughts? @tylerbutler

alexvy86:
> flub bump --updateAllDeps

Seems like that flag is an escape hatch to restore old behavior which was arguably incorrect, and probably isn't being used? Under that assumption, I'm skipping handling `catalog:` dependencies in that code path.

tylerbutler:
> > flub bump --updateAllDeps
> 
> Seems like that flag is an escape hatch to restore old behavior which was arguably incorrect, and probably isn't being used? Under that assumption, I'm skipping handling `catalog:` dependencies in that code path.

I think this is the right call.

tylerbutler:
> Do you have an idea of what exactly would confuse VSCode?

I was thinking it would get confused about where, say, the biome bin is. But I think I'm comcpletely wrong about that. vscode cares about the actual files on disk in node_modules and that should be the same with or without catalogs.

I'm mostly worried about places where we bump versions or calculate versions. Even though we _should_ skip those, it might be nice if, say, version parsing in version-tools threw a useful error or something when parsing them. But maybe that's not needed in practice. I would trust a Claude review on that front.

tylerbutler (approved):
I'm good with these as-is.

PR #26619, opened 2026-03-03
Migrate all uses of Jest+Puppeteer in the client release group to Playwright

alexvy86:
## Description

- Replaces jest, jest-puppeteer, jest-environment-puppeteer, expect-puppeteer, and puppeteer dependencies with @playwright/test across all packages that used jest + puppeteer it in the client release group.
- Gets rid of the puppeteer dependency entirely and cleans up, including removing the `@types/jest-environment-puppeteer` package.

More in detail about the jest + puppeteer -> Playwright migration:

- Adds playwright.config.ts per package (port via getTestPort from package.json name)
- Rewrites test files using Playwright fixtures and locator API
- Deletes jest.config.cjs and jest-puppeteer.config.cjs from each package
- Updates package.json scripts: test:playwright / test:playwright:verbose
- Adds test:playwright to fluidBuild.config.cjs task graph
- Adds test:playwright to repoPolicyCheck split test script names
- Adds test-results/ to root .gitignore (Playwright last-run artifacts)
- Updates tsconfig.json per package: adds playwright.config.ts to include, replaces skipLibCheck workaround with explicit "node" in types array
- Bumps minimum Node.js requirement to 20.19.6: Node.js 20.19.0–20.19.4 ship `cjs-module-lexer` 1.4.1, which fails to detect `Object.defineProperty`-style named CJS exports when imported from ESM — causing a `TypeError: Class extends value undefined` when Playwright loads test files that transitively import `@fluid-internal/client-utils` ([example](https://dev.azure.com/fluidframework/public/public%20Team/_build/results?buildId=382286&view=logs&j=a82bb044-b235-5622-d70b-42a84e9d6383&t=842d1f7f-d9d5-50bf-c8f7-b2f3a4559929&l=261)). Node.js 20.19.5 bumped `cjs-module-lexer` to 2.1.0, fixing the detection. Requiring 20.19.6 (the latest patch) ensures a stable baseline.

## Reviewer Guidance

Pipeline run for this PR shows [40,800 tests running](https://dev.azure.com/fluidframework/public/_build/results?buildId=382492&view=ms.vss-test-web.build-test-results-tab), vs [40,791 in other PRs](https://dev.azure.com/fluidframework/public/_build/results?buildId=382159&view=ms.vss-test-web.build-test-results-tab), which is due to the fact that the inventory-app mocha tests are currently not running for some reason (see [today](https://dev.azure.com/fluidframework/public/public%20Team/_build/results?buildId=382159&view=logs&j=855df218-f14a-5111-4dba-a3c8eab6872c&t=a5446ac8-4e73-5540-487b-79110ad67165&l=3257), vs [this PR](https://dev.azure.com/fluidframework/public/public%20Team/_build/results?buildId=382492&view=logs&j=855df218-f14a-5111-4dba-a3c8eab6872c&t=a5446ac8-4e73-5540-487b-79110ad67165&l=4292)).


.gitignore
```diff
@@ -4,6 +4,9 @@ dist
 lib
 !common/lib
 nyc
+
+# Playwright test artifacts: failure screenshots/traces and last-run state used by --last-failed
+test-results/
```
alexvy86, line 9:
```suggestion

# Playwright test artifacts: failure screenshots/traces and last-run state used by --last-failed
test-results/

```

Abe27342:
todo: verify CI gates test failure :)

alexvy86:
@Josmithr @Abe27342 I realized this PR only tackled `examples/apps/`, not the other subfolders of examples. There's also `packages/common/client-utils`, and three devtools packages that currently use jest. I'll ask Claude to migrate all of those too, but if we like the direction, we could merge these first to not end up with a single huge(r) PR.

alexvy86:
Claude noticed that a couple of packages are outliers in how they use jest:

> examples/client-logger/app-insights-logger and packages/tools/devtools/devtools-view — These don't fit the "Jest+Puppeteer → Playwright" pattern at all. They use Jest with the jsdom environment plus @testing-library/react (and @testing-library/jest-dom, @testing-library/user-event in the case of devtools-view) to render React components into a synthetic DOM and assert against it — including a11y checks (assertNoAccessibilityViolations) and keyboard-navigation simulation. There is no real browser, no puppeteer/jest-puppeteer, no dev server. Migrating them to Playwright would require either adopting Playwright Component Testing (experimental, fundamentally different API, would need test rewrites) or moving them to a different runner like Vitest or Mocha+jsdom. That's a separate decision from the E2E migration in this branch and should be tracked on its own.
>
> packages/common/client-utils — This one does use jest-puppeteer, but the tests are not E2E. They're unit-style parity tests for IsoBuffer and gitHash (buffer.spec.ts, gitHash.spec.ts) that need to run in a real browser to exercise the browser-targeted code paths, alongside a parallel Mocha suite that runs the same module surface in Node. The package builds Jest tests separately (build:test:jest → CJS) and runs Jest directly against the built CJS via rewire for gitHash patching. Mechanically the migration to Playwright is feasible (load the built bundle into a page and assert), but the test shape is different from the apps migration (no app server, no UI, just module-level assertions in a browser context), so it warrants a deliberate design rather than a copy of the app pattern.

I asked it to skip touching those for now.

examples/data-objects/clicker/tsconfig.json
alexvy86, line 1: Revert the changes to this file (some things just moved, or the comment was added because of an incorrect change that was reverted) and just apply the changes to `types`

build-tools/packages/build-cli/src/library/repoPolicyCheck/npmPackages.ts
```diff
@@ -1372,7 +1372,12 @@ export const handlers: Handler[] = [
 			}
 			const testScript = scripts.test;
 
-			const splitTestScriptNames = ["test:mocha", "test:jest", "test:realsvc"];
+			const splitTestScriptNames = [
+				"test:mocha",
+				"test:jest",
```
Josmithr, line 1377: Nit: maybe leave a TODO to remove this once all of the usages of jest are gone?

examples/apps/blobs/tests/blobs.test.ts
```diff
@@ -3,41 +3,21 @@
...
-		page = await browser.newPage();
-		await page.goto(globals.PATH);
-		await page.waitForFunction(() => typeof globalThis.loadAdditionalContainer === "function");
-		await page.close();
-	}, 20_000);
-
-	beforeEach(async () => {
-		page = await browser.newPage();
-		await page.goto(globals.PATH);
+test.describe("blobs", () => {
```
Josmithr, line 9: This looks SO much nicer than what was here before!

examples/apps/blobs/playwright.config.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const { name } = createRequire(import.meta.url)("./package.json") as { name: string };
+const testPort = getTestPort(name);
+const baseURL = `http://localhost:${testPort}`;
+
+export default defineConfig({
+	testDir: "tests",
+	forbidOnly: !!process.env.CI,
+	retries: 0,
+	timeout: 60_000,
```
Josmithr, line 18: Looks like this got increased from 10,000 to 60,000. Intentional?

examples/apps/blobs/playwright.config.ts
```diff
@@ -0,0 +1,33 @@
...
+import { defineConfig, devices } from "@playwright/test";
+import { getTestPort } from "@fluidframework/test-tools";
+
+const { name } = createRequire(import.meta.url)("./package.json") as { name: string };
+const testPort = getTestPort(name);
+const baseURL = `http://localhost:${testPort}`;
+
+export default defineConfig({
+	testDir: "tests",
+	forbidOnly: !!process.env.CI,
```
Josmithr, line 16: I forgot this was an option with playwright! Love it!

examples/apps/blobs/playwright.config.ts
```diff
@@ -0,0 +1,33 @@
+/*!
```
Josmithr, line 1: How much of this config could we share via extension? Seems like it might be nice to define a base config (maybe at the root of examples?) that all of the package-local configs extend.

examples/apps/presence-tracker/tests/presenceTracker.test.ts
```diff
@@ -265,7 +256,7 @@ describe("presence-tracker", () => {
 		});
 
 		// TODO: AB#28502: presence-tracker example multi-client test should not be skipped
-		it.skip("First client shows two clients connected in UI", async () => {
+		test.skip("First client shows two clients connected in UI", async () => {
```
Josmithr, line 259: Not for this PR, but it might be nice to do an audit of skipped tests and see if they are more reliable using playwright. Probably sufficient to just notify test owners after this PR goes in so they can audit their own skipped tests.

Josmithr (commented):
I didn't go line-by-line, but did a cursory scan through the changes. Overall, this looks like a huge improvement to me. The tests are simpler and much easier to parse. The config files also make more intuitive sense to me.

I left a handful of comments with some questions / suggestions.

alexvy86:
Already done between https://github.com/microsoft/FluidFramework/pull/27964, https://github.com/microsoft/FluidFramework/pull/28114, and https://github.com/microsoft/FluidFramework/pull/28048.

PR #26621, opened 2026-03-03
Add unordered list and header support to formatted text editor

brrichards:
## Summary                                                                                                                              
  - Add support for header (h1-h5) and unordered list formatting in the Quill-based formatted text editor by integrating `StringLineAtom` line-level formatting                
  - Handle line tag parsing from Quill attributes (headers as numbers, lists as strings) and convert them to/from `FormattedTextAsTree` line atoms                                                                                                                              
  - Update `buildDeltaFromTree` to emit correct Quill delta ops for line-level formatting (headers and bullet lists)                      
  - Enable header and bullet list toolbar buttons in the Quill editor                                                                     
  - Add CSS fixes for Quills bullet point vertical alignment in list items
  - Add tests for mimicing Quills trailing newline and the conversion of Quill to formattedText LineTag

  ## Files changed
  - `packages/framework/react/src/text/formatted/quillFormattedView.tsx` — Add `parseLineTag`, `createLineAtom`, `LineTagValue` type; update `buildDeltaFromTree` to handle line atoms; handle line tags in insert/retain delta ops; add header/list toolbar buttons and bullet CSS.
  - `packages/framework/react/src/test/text/textEditor.test.tsx` — Add line formatting test suite covering lineTag

### DIFF
--- a/packages/framework/react/src/text/formatted/quillFormattedView.tsx
+++ b/packages/framework/react/src/text/formatted/quillFormattedView.tsx
@@ -280,7 +280,7 @@ export function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp
 	const ops: QuillDeltaOp[] = [];
 	// Accumulator for current run of identically-formatted text
 	let text = "";
-	let currentAttributes: Record<string, unknown> = {};
+	let previousAttributes: Record<string, unknown> = {};
 	// JSON key for current attributes, used for equality comparison
 	// TODO:Performance: implement faster equality check.
 	let key = "";
@@ -289,7 +289,7 @@ export function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp
 	const pushRun = (): void => {
 		if (!text) return;
 		const op: QuillDeltaOp = { insert: text };
-		if (Object.keys(currentAttributes).length > 0) op.attributes = currentAttributes;
+		if (Object.keys(previousAttributes).length > 0) op.attributes = previousAttributes;
 		ops.push(op);
 	};
 
@@ -297,22 +297,22 @@ export function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp
 	// TODO:Performance: Optimize this loop by adding an API to get runs to FormattedTextAsTree.Tree, and implementing that using cursors.
 	// Something like `getUniformRun(startIndex, maxLength): number` and `substring(startIndex, length): string`.
 	for (const atom of root.charactersWithFormatting()) {
-		const attributes = formatToQuillAttributes(atom.format);
+		const currentAttributes = formatToQuillAttributes(atom.format);
 
 		if (atom.content instanceof FormattedTextAsTree.StringLineAtom) {
 			// Merge line-specific attributes (header/list) into the format
 			const lineTag = atom.content.tag.value;
-			Object.assign(attributes, lineTagToQuillAttributes[lineTag]);
+			Object.assign(currentAttributes, lineTagToQuillAttributes[lineTag]);
 
 			// Line atoms always break the current run and emit a newline
 			pushRun();
 			text = "";
 			key = "";
 			const op: QuillDeltaOp = { insert: "\n" };
-			if (Object.keys(attributes).length > 0) op.attributes = attributes;
+			if (Object.keys(currentAttributes).length > 0) op.attributes = currentAttributes;
 			ops.push(op);
 		} else {
-			const stringifiedAttributes = JSON.stringify(attributes);
+			const stringifiedAttributes = JSON.stringify(currentAttributes);
 			if (stringifiedAttributes === key) {
 				// Same formatting as previous character - extend run
 				text += atom.content.content;
@@ -320,7 +320,7 @@ export function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp
 				// Different formatting - push previous run and start a new one
 				pushRun();
 				text = atom.content.content;
-				currentAttributes = attributes;
+				previousAttributes = currentAttributes;
 				key = stringifiedAttributes;
 			}
 		}
--- a/packages/framework/react/src/text/formatted/quillFormattedView.tsx
+++ b/packages/framework/react/src/text/formatted/quillFormattedView.tsx
@@ -288,31 +288,32 @@ export function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp
 	// TODO:Performance: Optimize this loop by adding an API to get runs to FormattedTextAsTree.Tree, and implementing that using cursors.
 	// Something like `getUniformRun(startIndex, maxLength): number` and `substring(startIndex, length): string`.
 	for (const atom of root.charactersWithFormatting()) {
-		const lineTag =
-			atom.content instanceof FormattedTextAsTree.StringLineAtom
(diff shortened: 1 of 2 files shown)


packages/framework/react/src/text/formatted/quillFormattedView.tsx
```diff
@@ -244,19 +288,34 @@ function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp[] {
 	// TODO:Performance: Optimize this loop by adding an API to get runs to FormattedTextAsTree.Tree, and implementing that using cursors.
 	// Something like `getUniformRun(startIndex, maxLength): number` and `substring(startIndex, length): string`.
 	for (const atom of root.charactersWithFormatting()) {
-		const a = formatToQuillAttrs(atom.format);
-		const k = JSON.stringify(a);
-		if (k === key) {
-			// Same formatting as previous character - extend current run
-			text += atom.content.content;
+		const lineTag =
```
CraigMacomber: Since line tags do have formatting (both in our setup and in quill), lifting the atom.format stuff out of this if and doing the formatting in a way not specific to which atom type we have would be better.

Then we can do a `if (atom.content instanceof FormattedTextAsTree.StringLineAtom)` and add the line specific formatting to the format.

packages/framework/react/src/text/formatted/quillFormattedView.tsx
```diff
@@ -244,19 +288,34 @@ function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp[] {
...
+			atom.content instanceof FormattedTextAsTree.StringLineAtom
+				? atom.content.tag.value
+				: undefined;
+		if (lineTag === undefined) {
+			const a = formatToQuillAttrs(atom.format);
```
CraigMacomber: variable names should be made out of English words in a descriptive manner.

packages/framework/react/src/text/formatted/quillFormattedView.tsx
```diff
@@ -59,6 +59,25 @@ const fontSet = new Set<string>(["monospace", "serif", "sans-serif", "Arial"]);
...
+/** Quill header numbers → LineTag values. */
+const headerToLineTag: Record<number, LineTagValue> = {
+	1: "h1",
+	2: "h2",
+	3: "h3",
+	4: "h4",
+	5: "h5",
+};
+/** LineTag values → Quill attributes. Used by buildDeltaFromTree (tree → Quill). */
+const lineTagToQuillAttrs: Record<LineTagValue, Record<string, unknown>> = {
```
Josmithr: Nit: Our guidelines generally prefer avoiding abbreviations unless there is a particularly compelling reason to use them. I'd recommend updating the usages of "attrs" to spell out "attributes" here and below.
https://github.com/microsoft/FluidFramework/wiki/Coding-Guidelines#-avoid-using-abbreviations

packages/framework/react/src/text/formatted/quillFormattedView.tsx
```diff
@@ -149,6 +168,31 @@ function parseSize(size: unknown): number {
 	return defaultSize;
 }
 
+/** Extract a LineTag from Quill attributes, or undefined if none present. Quill only supports one LineTag at a time. */
```
Josmithr, line 173: Nit: since we never expect to see more than 1 line tag in the input here, should we add an assertion for it?

packages/framework/react/src/text/formatted/quillFormattedView.tsx
```diff
@@ -360,16 +422,36 @@ const FormattedTextEditorView = React.forwardRef<
...
-						);
+						assert(typeof op.retain === "number", "Expected retain count to be a number");
 						// Convert UTF-16 retain count to codepoint count
 						const retainedStr = content.slice(utf16Pos, utf16Pos + op.retain);
 						const cpCount = codepointCount(retainedStr);
 
 						if (op.attributes) {
-							root.formatRange(cpPos, cpPos + cpCount, quillAttrsToPartial(op.attributes));
+							const lineTag = parseLineTag(op.attributes);
+							if (lineTag !== undefined && content[utf16Pos] === "\n") {
```
Josmithr, line 442: Nit: it would probably be nice to add some quick notes to each of the cases here describing the scenario being checked for.

packages/framework/react/src/text/formatted/quillFormattedView.tsx
Josmithr: Nit: since we presumably don't intend for these to be mutated, we could make them `Readonly<Record<...>>`

Josmithr (approved):
Left a handful of nitpicks, but overall looks good to me!

packages/framework/react/src/text/formatted/quillFormattedView.tsx
```diff
@@ -59,6 +59,25 @@ const fontSet = new Set<string>(["monospace", "serif", "sans-serif", "Arial"]);
 const defaultSize = 12;
 /** Default font when no explicit font is specified. */
 const defaultFont = "Arial";
+/** The string literal values accepted by LineTag. */
+type LineTagValue = "h1" | "h2" | "h3" | "h4" | "h5" | "li";
```
CraigMacomber: We should share typing with the formatted text domain in the tree package for this. I think `Paramaters<typeof FormattedTextAsTree.LineTag>[0]` or something like that should do it.

packages/framework/react/src/text/formatted/quillFormattedView.tsx
```diff
@@ -59,6 +59,25 @@ const fontSet = new Set<string>(["monospace", "serif", "sans-serif", "Arial"]);
...
+/** Quill header numbers → LineTag values. */
+const headerToLineTag: Record<number, LineTagValue> = {
+	1: "h1",
+	2: "h2",
+	3: "h3",
+	4: "h4",
+	5: "h5",
```
CraigMacomber, line 72: Might be simpler to store the actual schema here instead of the strings (FormattedTextAsTree.LineTag.h1 for example).

I don't think it really matters though.

packages/framework/react/src/text/formatted/quillFormattedView.tsx
```diff
@@ -149,6 +168,31 @@ function parseSize(size: unknown): number {
...
 }
 
+/** Extract a LineTag from Quill attributes, or undefined if none present. Quill only supports one LineTag at a time. */
+export function parseLineTag(
+	attrs?: Record<string, unknown>,
+): FormattedTextAsTree.LineTag | undefined {
+	if (!attrs) return undefined;
+	if (typeof attrs.header === "number") {
+		const tag = headerToLineTag[attrs.header];
+		if (tag !== undefined) return FormattedTextAsTree.LineTag(tag);
```
CraigMacomber: I'm not sure if dropping headings of an unsupported level is better than defaulting to some level.

`FormattedTextAsTree.LineTag(headerToLineTag[attrs.header] ?? defaultHeading);`

might be better.

packages/framework/react/src/text/formatted/quillFormattedView.tsx
CraigMacomber: Dropping the type entirely from left hand side, and instead using `as const stratifies Record<...>` might be better as that captures the full set of keys, not just the typing used for them for a little more safety (will error if a line is missing when you use it below)

Josmithr (approved):

PR #26623, opened 2026-03-03
Port benchmark to node asserts

CraigMacomber:
## Description

Port benchmark tests to node asserts to better match rest of repo.

## Reviewer Guidance


alexvy86 (approved):
I'd say also apply Copilot's suggestion above.

CraigMacomber:
> I'd say also apply Copilot's suggestion above.

I'll do that, and the followup pnpm i and such in another change.

PR #26624, opened 2026-03-03
ci: parallelize merge-tree mocha tests

frankmueller-msft:
## Summary

- Enable mocha parallel mode (`parallel: true`, `jobs: 4`) for the merge-tree test suite, which is one of the two longest-running mocha suites
- Increase timeouts on `reconnectFarm` and `applyStashedOpFarm` tests from 30s to 300s to match the convention used by other farm tests (`conflictFarm`, `obliterateFarm`, `rollbackFarm`) and prevent timeouts under parallel CPU contention. The timeout is a ceiling, not a target — these tests complete in ~15-20s normally. The increase just prevents false failures when parallel workers compete for CPU.

This is a focused, minimal version of the approach in #26523 — only parallelizing merge-tree (not end-to-end tests, which have a test count inflation issue due to compat config propagation in worker processes). No pipeline changes, no extra CI jobs.

## Results

| | Coverage tests time |
|---|---|
| **Baseline** (4 recent builds) | ~22 min |
| **This PR** | **14 min 16s** |
| **Improvement** | **~8 min faster (36%)** |

The baseline is consistent across 4 recent builds (21m42s – 22m52s). The coverage test job is no longer the bottleneck — RealsvcTinyliciousTest (15m01s) and the Build stage (19m06s) are now the longer jobs.

## Why not parallelize end-to-end tests too?

The end-to-end test suite should not be parallelized until the `configList` computation in `test-version-utils/describeCompat.ts` is made worker-safe. `nconf` parses `process.argv` which differs in mocha worker processes, causing different compat configs to be generated and inflating the test count (3,639 → 4,847 tests). This needs a separate fix.

## Context

Our CI already runs all mocha suites in parallel. The bottleneck is that merge-tree and end-to-end tests are individually slow and not internally parallelized. Splitting packages into separate CI jobs adds setup overhead and pipeline complexity without meaningful gains — this PR fixes the merge-tree bottleneck with 4 lines.

Supersedes #26586. Builds on the approach from #26523.

## Test plan

- [x] Verify merge-tree tests pass with parallel mode enabled
- [x] Verify `ReconnectFarm_2_0` no longer times out
- [x] Confirm CI wall-clock improvement on the coverage test job (~22 min → ~14 min)


packages/dds/merge-tree/.mocharc.cjs
```diff
@@ -8,4 +8,6 @@
 const getFluidTestMochaConfig = require("@fluid-internal/mocha-test-setup/mocharc-common");
 
 const config = getFluidTestMochaConfig(__dirname);
+config.parallel = true;
+config.jobs = 4;
```
CraigMacomber, line 18: We don't want everyone replicating this pattern to other packages as that would make things worse not better. THus I think some docs would be a good idea:

```suggestion
// This test suite is the slowest on CI if not run in parallel, so this parallelization speeds up CI significantly.
config.parallel = true;
// CI assumes that each package will use one thread when computing how many package's tests to run inn parallel.
// To avoid making this computation too inaccurate and slowing things down with excessive memory use
// (and other overhead) from too many threads, we keep the numbers of jobs limited,
// and only enable this parallelization for packages where doing so actually produces an improvement in CI time.
// 4 jobs was measured to give most of the speed up, even when just running this suite, while not slowing down CI.
config.jobs = 4;
```

alexvy86 (approved):
I was a bit concerned about this potentially introducing issues but the PR run seems pretty good. I confirmed the number of passing and skipped tests matches other PRs. I agree with Craig's suggestion above with more explicit docs, but otherwise I think this is a net win.

packages/dds/merge-tree/.mocharc.cjs
frankmueller-msft: Good point — added the documentation. Used your suggested wording (fixed the minor typo: "inn" → "in").

frankmueller-msft: @CraigMacomber, added some comments here. Looking good?

CraigMacomber (approved):
Looks good to me!

anthony-murphy (approved):

packages/dds/merge-tree/src/test/client.applyStashedOpFarm.spec.ts
```diff
@@ -209,7 +209,9 @@ function runApplyStashedOpFarmTests(
 				opts,
 				(s, m, c) => applyMessagesWithReconnect(s, m, c, stashClients),
 			);
-		}).timeout(30 * 1000);
+			// Matches the timeout convention used by other farm tests (conflictFarm, obliterateFarm,
+			// rollbackFarm). Tests complete in ~15-20s normally, but need headroom under parallel execution.
+		}).timeout(30 * 10000);
```
jason-ha, line 214: @frankmueller-msft, is this because in parallel execution there might be stalls introduced? There are parallel executions, but some of those get starved?

frankmueller-msft: The increase prevents false failures when parallel workers compete for CPU.

jikim-msft:
[Incident Ticket](https://portal.microsofticm.com/imp/v5/incidents/details/756974246/summary)

Hey @frankmueller-msft ! I'm this week's OCE and have been investigating the consecutive failures in the [Performance Benchmarks pipeline](https://dev.azure.com/fluidframework/internal/_build?definitionId=96&_a=summary). 

Five consecutive runs are failing in the `@fluidframework/merge-tree stage (perf_unit_tests_runtime)` with:                                                                                           

`TypeError: hook.error is not a function`

Since this PR is the first commit where the failure appears and the root cause is clear, I'm reverting it to unblock the pipeline.

jikim-msft:
I believe the `.mocharc.cjs` change in this PR sets `parallel: true` globally for all mocha runs in merge-tree, including **benchmark tests**. 

Looking at the [MochaReporter.ts](https://github.com/microsoft/FluidFramework/blob/0065839273a38e270e3639bc90297c7c4ab9c7e3/tools/benchmark/src/MochaReporter.ts#L92), `hook.error()` is called for every mocha hook that completes during the test run. 

I think this [Mocha Official Doc](https://mochajs.org/features/parallel-mode) implies that Mocha serializes Hook objects across worker processes somehow (i.e., via IPC) in parallel mode. And I believe during the serialization (`Hook.prototype.serialize()`) some methods (in this case `hook.error()`) are not included and make the benchmark reporter to fail. 

@CraigMacomber

CraigMacomber:
> I believe the `.mocharc.cjs` change in this PR sets `parallel: true` globally for all mocha runs in merge-tree, including **benchmark tests**.
> 
> Looking at the [MochaReporter.ts](https://github.com/microsoft/FluidFramework/blob/0065839273a38e270e3639bc90297c7c4ab9c7e3/tools/benchmark/src/MochaReporter.ts#L92), `hook.error()` is called for every mocha hook that completes during the test run.
> 
> I think this [Mocha Official Doc](https://mochajs.org/features/parallel-mode) implies that Mocha serializes Hook objects across worker processes somehow (i.e., via IPC) in parallel mode. And I believe during the serialization (`Hook.prototype.serialize()`) some methods (in this case `hook.error()`) are not included and make the benchmark reporter to fail.
> 
> @CraigMacomber

I think a good fix for this would be to set parallel to false for perf tests. We don't want them to run in parallel anyway since it would be bad for data quality.

@frankmueller-msft 
Specifically, I think we can make the change in this PR conditional on `if (!process.argv.includes("--perfMode"))`.

First test without that, and confirm you can reproduce the issue where running the perf script in merge tree fails with your change, then add that `if` to disable it for perf tests and confirm that fixes it.

PR #26630, opened 2026-03-03
content derived id allocation batches

dannimad:
### Problem
When two containers rehydrate from the same stashed state ("parallel fork"), both submit identical ID Allocation ops. DuplicateBatchDetector couldn't catch this because ID allocation batches had ignoreBatchId: true, meaning no batchId was tracked for duplicate detection.

### Solution: Content-Derived BatchId
4 source files modified, 3 test files updated with 8 new tests:

### Source Changes
[batchManager.ts](vscode-webview://0setsnuqdgpe6inajv22i20l1mo21fcect4fpqbjjmsc3q1f9bv8/packages/runtime/container-runtime/src/opLifecycle/batchManager.ts) — Renamed ignoreBatchId → skipResubmit in IBatchManagerOptions with updated JSDoc

[containerRuntime.ts](vscode-webview://0setsnuqdgpe6inajv22i20l1mo21fcect4fpqbjjmsc3q1f9bv8/packages/runtime/container-runtime/src/containerRuntime.ts) — Added computeIdAllocBatchId(idRange) which derives a deterministic batchId from JSON.stringify(idRange). Both forked containers produce identical IdCreationRange, so the batchId matches → DuplicateBatchDetector catches it. Modified submitIdAllocationOpIfNeeded to pass this batchId to the outbox.

[outbox.ts](vscode-webview://0setsnuqdgpe6inajv22i20l1mo21fcect4fpqbjjmsc3q1f9bv8/packages/runtime/container-runtime/src/opLifecycle/outbox.ts) — Added pendingIdAllocBatchId field. submitIdAllocation now accepts an optional batchId. flushAll passes it through flushInternal → popBatch which stamps it on the op metadata. Renamed ignoreBatchId → skipResubmit.

[pendingStateManager.ts](vscode-webview://0setsnuqdgpe6inajv22i20l1mo21fcect4fpqbjjmsc3q1f9bv8/packages/runtime/container-runtime/src/pendingStateManager.ts) — Renamed ignoreBatchId → skipResubmit throughout. Added IPendingMessageV1 type for backward compat. Updated patchbatchInfo to migrate old ignoreBatchId field to skipResubmit when loading stashed state.

### Test Changes
duplicateBatchDetector.spec.ts: 3 tests verifying fork detection with content-derived batchIds
outbox.spec.ts: 2 tests verifying batchId stamping and clearing behavior
pendingStateManager.spec.ts: 3 tests verifying skipResubmit replay behavior, remote batch matching, and backward compat migration

### DIFF
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -239,6 +239,7 @@ import {
 } from "./messageTypes.js";
 import type { ISavedOpMetadata } from "./metadata.js";
 import {
+	type BatchId,
 	type LocalBatchMessage,
 	type BatchStartInfo,
 	DuplicateBatchDetector,
@@ -826,6 +827,17 @@ function canStageMessageOfType(
 		// These are typically sent shortly after boot and will not be common in Staging Mode, but it's possible.
 		type === ContainerMessageType.DocumentSchemaChange
 	);
+}
+
+/**
+ * Compute a deterministic batchId for an ID Allocation batch based on the IdCreationRange content.
+ * Both forked containers rehydrating from the same stashed state will produce identical
+ * IdCreationRange objects (same sessionId, firstGenCount, count, localIdRanges), so
+ * JSON.stringify produces the same string, yielding the same batchId.
+ * This enables DuplicateBatchDetector to detect the duplicate and throw a clear fork error.
+ */
+function computeIdAllocBatchId(idRange: IdCreationRange): BatchId {
+	return `idalloc_${JSON.stringify(idRange)}`;
 }
 
 /**
@@ -4702,7 +4714,12 @@ export class ContainerRuntime
 					referenceSequenceNumber: this.deltaManager.lastSequenceNumber,
 					staged,
 				};
-				this.outbox.submitIdAllocation(idAllocationBatchMessage);
+				// Compute a deterministic batchId from the ID range content.
+				// Two forked containers rehydrating from the same stashed state will produce
+				// identical IdCreationRange objects, yielding the same batchId. This enables
+				// DuplicateBatchDetector to detect the duplicate and throw a clear fork error.
+				const batchId = computeIdAllocBatchId(idRange);
+				this.outbox.submitIdAllocation(idAllocationBatchMessage, batchId);
 			}
 		}
 	}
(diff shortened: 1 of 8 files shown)


packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -340,9 +342,10 @@ export class Outbox {
 		this.addMessageToBatchManager(this.blobAttachBatch, message);
 	}
 
-	public submitIdAllocation(message: LocalBatchMessage): void {
+	public submitIdAllocation(message: LocalBatchMessage, batchId?: BatchId): void {
 		this.maybeFlushPartialBatch();
 
+		this.pendingIdAllocBatchId = batchId;
```
markfields, line 348: How about moving this state to the ID Allocation Batch Manager

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -828,6 +829,17 @@ function canStageMessageOfType(
 	);
 }
 
+/**
+ * Compute a deterministic batchId for an ID Allocation batch based on the IdCreationRange content.
+ * Both forked containers rehydrating from the same stashed state will produce identical
+ * IdCreationRange objects (same sessionId, firstGenCount, count, localIdRanges), so
```
markfields, line 835: I wonder if we should have this be a bit relaxed.  Only include sessionId and firstGenCount or something.  If only count is mismatched, it should still fail I think.  I'd have to learn more what "gen count" is, we should run this fn by some ID Compressor SMEs.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -828,6 +829,17 @@ function canStageMessageOfType(
...
+ * JSON.stringify produces the same string, yielding the same batchId.
```
markfields, line 836: Warning, JSON.stringify doesn't sort, it goes off of insertion order.  So if that changes between different code versions this breaks.  Would recommend something like this (considering previous comment too):

```typescript
// DO NOT REORDER the props in the object being stringified!
return `idalloc_${JSON.stringify({ sessionId: idRange.sessionId, genCount: idRange.ids?.firstGenCount })}`
```

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
markfields: ALso this has a problem where multiple calls to `submitIdAllocation` between flushes will include several competing ranges/batchIds... hmmmm...

packages/runtime/container-runtime/src/containerRuntime.ts
markfields: Maybe that's moving too close to just replicating the process logic though.  Keeping it a strict match may be sufficient to fill this gap.

PR #26631, opened 2026-03-03
Switch to React's new JSX transform

ChumpChief:
## Summary

- Switch all packages to React's new JSX transform by setting `"jsx": "react-jsx"` in `tsconfig.base.json`
- Replace all `import React from "react"` and `import * as React from "react"` with direct named imports (e.g. `import { useState, type ReactElement } from "react"`)
- Add inline `type` specifiers to imports that are only used as types
- Remove unused React imports left over from the old JSX transform
- Add `plugin:react/jsx-runtime` to `eslint-config-fluid` (both legacy and flat configs) to disable `react/react-in-jsx-scope` globally

## Test plan

- [x] `pnpm eslint` passes with zero errors
- [ ] Full CI build passes
- [ ] Verify applications render correctly at runtime

### DIFF
--- a/common/build/eslint-config-fluid/minimal-deprecated.js
+++ b/common/build/eslint-config-fluid/minimal-deprecated.js
@@ -522,7 +522,11 @@ module.exports = {
 				// Plugin documentation: https://www.npmjs.com/package/eslint-plugin-react-hooks
 				"react-hooks",
 			],
-			extends: ["plugin:react/recommended", "plugin:react-hooks/recommended"],
+			extends: [
+				"plugin:react/recommended",
+				"plugin:react-hooks/recommended",
+				"plugin:react/jsx-runtime",
+			],
 			rules: {
 				// TODO: These rules should be re-enabled once we are on eslint 9
 				// and the react plugins are upgraded to more recent versions
--- a/examples/apps/blobs/eslint.config.mts
+++ b/examples/apps/blobs/eslint.config.mts
@@ -7,15 +7,6 @@ import type { Linter } from "eslint";
 import { recommended } from "../../../common/build/eslint-config-fluid/flat.mts";
 import sharedConfig from "../../eslint.config.data.mts";
 
-const config: Linter.Config[] = [
-	...recommended,
-	...sharedConfig,
-	{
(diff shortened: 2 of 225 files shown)


common/build/eslint-config-fluid/minimal-deprecated.js
```diff
@@ -522,7 +522,11 @@ module.exports = {
 				// Plugin documentation: https://www.npmjs.com/package/eslint-plugin-react-hooks
 				"react-hooks",
 			],
-			extends: ["plugin:react/recommended", "plugin:react-hooks/recommended"],
+			extends: [
```
Josmithr, line 525: You may need to run the build script in this package - I would expect all of our JSON reports to be updated.

examples/apps/blobs/eslint.config.mts
```diff
@@ -7,6 +7,15 @@ import type { Linter } from "eslint";
 import { recommended } from "../../../common/build/eslint-config-fluid/flat.mts";
 import sharedConfig from "../../eslint.config.data.mts";
 
-const config: Linter.Config[] = [...recommended, ...sharedConfig];
+const config: Linter.Config[] = [
+	...recommended,
+	...sharedConfig,
+	{
+		// TODO: remove this override once eslint-config-fluid has been updated to disable this rule.
+		rules: {
```
Josmithr: The shared config package is now included in the same workspace as the client packages - we should be able to make any necessary changes now, right?

ChumpChief: Good point, maybe I'm doing something wrong (in combination with your other comment).  print-configs doesn't update anything and has weird output.  Will look further.

ChumpChief: Updated the correct location for the flat config, which now makes these overrides unnecessary.  Removed the per-package overrides in latest.

common/build/eslint-config-fluid/minimal-deprecated.js
ChumpChief: Confirmed that we need to update two places - here for the old-style config, and also in overrides.mts for the new flat configs (which then are the ones that print out).  Fixed in latest.

Josmithr (approved):

PR #26633, opened 2026-03-03
tool: ignore Claude personal and temporary files/folder

jason-ha:
This configuration provided by Claude and revised to apply to all subdirectories in case Claude is invoked in a narrow scope.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,7 +57,7 @@ build_output_archive
 
 # Claude Code
 # Local settings contain author-specific information and should not be shared.
-**/.claude/settings.local.json
+**/.claude/**/*.local.*
 # Session-specific task tracking
 **/.claude/todos/
 # Git worktrees created by Claude Code
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,9 @@ UPCOMING.md
 build_output_archive
 
 # Claude Code
-# This is a local file that likely contains author specific information and
-# should not be shared generally.
-.claude/settings.local.json
+# Local settings contain author-specific information and should not be shared.
+**/.claude/settings.local.json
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -56,6 +56,9 @@ UPCOMING.md
 build_output_archive
 
 # Claude Code
-# This is a local file that likely contains author specific information and
-# should not be shared generally.
-.claude/settings.local.json
```
anthony-murphy, line 61: i'd rather we ignore .claude, then opt-in things to share, like !.claude/skills

alexvy86 (commented):
If going the route of excluding things explicitly, I think `**/CLAUDE.local.md` should also be here.

.gitignore
alexvy86: I'm with Tony on this. And to your point @jason-ha , `rules/` does feel like something that should be shared.

.gitignore
```diff
@@ -56,6 +56,9 @@ UPCOMING.md
...
+# Local settings contain author-specific information and should not be shared.
+**/.claude/**/*.local.*
+# Session-specific task tracking
+**/.claude/todos/
+# Git worktrees created by Claude Code
+**/.claude/worktrees/
```
tylerbutler, line 64: Interesting. Claude creates worktrees in .worktrees for me - not in the claude folder.

tylerbutler (approved):

.gitignore
jason-ha: It could depend on the model. https://betterstack.com/community/guides/ai/git-worktrees-claude/ definitely documents use of that folder.
Easy enough to ignore as I can't see any utility in sharing.

PR #26635, opened 2026-03-03
tool: give Claude notes on repo policy checking

jason-ha:
via memory so that work isn't cut short.

----
compare to #26636

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,0 +1,5 @@
+# FluidFramework Claude Code Notes
+
+## Policy Checking
+- Run `pnpm policy-check --path <relative-path>` from the repo root to check a specific package. Do NOT `cd` into the package directory.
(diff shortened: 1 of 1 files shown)


.claude/CLAUDE.md
```diff
@@ -0,0 +1,5 @@
+# FluidFramework Claude Code Notes
+
+## Policy Checking
+- Run `pnpm policy-check --path <relative-path>` from the repo root to check a specific package. Do NOT `cd` into the package directory.
```
anthony-murphy, line 4: i'm very against a change like this. i personally don't think it is necessary, as i never have problems with this check and llm. it also isn't necessary for lots of kinds of work, like exploration, and prototypeing. 

if you find the model can't reliably do this based on its understanding of the repo, and its a task you think must be done frequently, a skill would be a better fit.

anthony-murphy (changes requested):
make it a skill if you really need it

jason-ha:
Confirmed the skill version from #26636 would be leveraged when rerunning the scenario in which it was originally a problem.

PR #26638, opened 2026-03-03
Rework Benchmark Package

CraigMacomber:
## Description

Major rework of benchmark package's implementation and APIs.

Main changes:
1. Mocha specific API surface reduced to "benchmarkIt"
2. Naming of types and functions in API now clear about duration vs memory in all cases.
3. Memory benchmark API has been redesigned, allowing for more robustness and easier to use API.
4. Memory benchmarks now have much more aggressive GC to collect more stable results.
5. Geometric mean has been restored to the results summary (was removed in the past) and correctly handles both larger is better and smaller is better values. This should make evaluating overall impact of changes easier.
6. All benchmarks now have a primary result (included in the geometric mean), and optional additional results.
7. All benchmarks share common utilities, with memory and duration being expressed in terms of the public APIs.
8. Report files have a new format which preserves hierarchy.
9. APIs and types are exported for dealing with the data format used in the report files.

Changes needed to integrate this update can be seen in https://github.com/CraigMacomber/FluidFramework/pull/5

## Breaking Changes

Most of the APIs have changed, especially memory tests.
Report file format has changed.

## Reviewer Guidance


tools/benchmark/src/test/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-	"extends": "@fluidframework/build-common/ts-common-config.json",
```
alexvy86, line 2: If we do this, the build pipeline for benchmark tools should be updated to trigger on changes to `common/build/eslint-config-fluid/` like others do.

tools/benchmark/CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # @fluid-tools/benchmark
 
+## 0.53.0
+
+-   Memory benchmarks now have much more aggressive GC to collect more stable results.
+-   Geometric mean has been restored to the results summary (was removed in the past) and correctly handles both larger is better and smaller is better values. This should make evaluating overall impact of changes easier.
```
alexvy86:
```suggestion
-   Geometric mean has been restored to the results summary (was removed in the past) and correctly handles both larger-is-better and smaller-is-better values. This should make evaluating overall impact of changes easier.
```

tools/benchmark/src/Reporter.ts
```diff
@@ -178,13 +198,18 @@ export class BenchmarkReporter {
...
 
 		for (const [, value] of benchmarksMap) {
 			if (isResultError(value)) {
 				countFailure++;
 			} else {
 				sumRuntime += value.elapsedSeconds;
 				countSuccessful++;
+				const primary = value.data.primary;
+				geometricMeanProductValues.push(
+					primary.type === ValueType.SmallerIsBetter ? primary.value : 1 / primary.value,
```
alexvy86: Division by 0 handling?

tools/benchmark/CHANGELOG.md
alexvy86: For reference, and to validate my understanding: geometric mean is used as an aggregate measure of performance for a whole run, right? Because (as I just learned) it avoids problems trying to compare the overall performance (not the test-wise performance) of different runs, which could be skewed depending on which run's final measurements we take as numerator/denominator when we try to understand the overall perf difference as a ratio between them.

alexvy86 (commented):
Haven't gone through the whole thing, but leaving some comments for now.

tools/benchmark/CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+-   Multiple benchmarks with the same name in the same suite no longer overwrite results, and emit a warning.
+-   Include ArrayBuffer memory usage.
+
+### ⚠ BREAKING CHANGES
+
+-   Mocha specific API surface reduced to `benchmarkIt` which can be used to wrap any kind of benchmark for use in mocha. All tests will need to be edited to accommodate this.
+-   Naming of types and functions are now clear about duration vs memory: most type imports will need to be updated to accommodate this.
+-   Memory benchmarks now require sampling memory before during and after the allocation being measured is retained in memory to allow for the test to know what it's supposed to measure and sanity check that the test is freeing it properly. As this is an entirely different API, all memory tests will need significant changes.
+-   Formatting of the output json results files has changed: code consuming them will have to be updated.
+-   Reporter now accepts `reportFile` instead of `reportDir`, and sames results to a single file.
```
alexvy86:
```suggestion
-   Reporter now accepts `reportFile` instead of `reportDir`, and saves results to a single file.
```

tools/benchmark/src/Configuration.ts
```diff
@@ -233,6 +247,11 @@ export type HookFunction = () => void | Promise<unknown>;
  * // cycles 40
  * // teardown goes here
  * ```
+ *
+ * @privateRemarks
+ * Replace these with a patter pattern.
```
alexvy86:
```suggestion
 * Replace these with a better pattern.
```

tools/benchmark/api-report/benchmark.public.api.md
```diff
@@ -182,16 +186,27 @@ export enum Phase {
     WarmUp = 0
 }
 
-// @public
-export function prettyNumber(num: number, numDecimals?: number): string;
-
 // @public
 export function qualifiedTitle(args: BenchmarkDescription & Titled & {
```
alexvy86, line 188: Nit: since we're making big changes to the API surface, do we need to export this? Feels like an internal utility.

tools/benchmark/src/durationBenchmarking/configuration.ts
```diff
@@ -173,28 +140,16 @@ export interface OnBatch {
...
-		HookArguments,
-		BenchmarkTimingOptions,
-		OnBatch,
-		BenchmarkDescription {}
-
 /**
  * Validates arguments to `benchmark`.
  * @public
  */
 export function validateBenchmarkArguments(
```
alexvy86, line 230: This one also feels like it doesn't need to be exposed from the package.

tools/benchmark/src/durationBenchmarking/getDuration.ts
```diff
@@ -47,26 +51,38 @@ export const defaultTimingOptions: Required<BenchmarkTimingOptions> = {
 	startPhase: Phase.WarmUp,
 };
 
+export const nonPerfTestingArgs: Required<BenchmarkTimingOptions> = {
```
alexvy86:
```suggestion
/**
 * Default values for when performance testing mode is not enabled.
 */
export const nonPerfTestingArgs: Required<BenchmarkTimingOptions> = {
```

tools/benchmark/src/durationBenchmarking/getDuration.ts
```diff
@@ -297,3 +316,17 @@ async function doBatchAsync(
 function tryRunGarbageCollection(): void {
 	global?.gc?.();
 }
+
+/**
+ * Configures a benchmark that uses {@link collectDurationData}
+ * to measure duration and returns the results in a format suitable for reporting via {@link benchmarkIt}.
+ * @public
+ */
+export function benchmarkDuration(
```
alexvy86, line 334: Not sure about `benchmarkDuration` itself, haven't looked at the rest of the PR, but `collectDurationData` feels like it shouldn't be exposed by the package.

tools/benchmark/src/memoryBenchmarking/configuration.ts
```diff
@@ -0,0 +1,85 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Callbacks for allocation and deallocation benchmarks/
```
alexvy86:
```suggestion
 * Callbacks for allocation and deallocation benchmarks
```

tools/benchmark/src/memoryBenchmarking/configuration.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @public
+ */
+export interface MemoryUseBenchmark {
+	/**
+	 * A function that should loop until `callbacks.continue()` returns false, and in each loop call the callbacks in order:
+	 * 1. `callbacks.beforeAllocation()`
+	 * 2. Allocate memory in some way.
+	 * 3. `callbacks.whileAllocated()`
+	 * 4. Deallocate the memory allocated in step 2.
+	 * 5. Optionally call `callbacks.afterDeallocation()` to measure deallocation separately from allocation
```
alexvy86: Making it match the param name.
```suggestion
	 * A function that should loop until `state.continue()` returns false, and in each loop call the callbacks in order:
	 * 1. `state.beforeAllocation()`
	 * 2. Allocate memory in some way.
	 * 3. `state.whileAllocated()`
	 * 4. Deallocate the memory allocated in step 2.
	 * 5. Optionally call `state.afterDeallocation()` to measure deallocation separately from allocation
```

tools/benchmark/src/memoryBenchmarking/configuration.ts
```diff
@@ -0,0 +1,85 @@
...
+	 * (the amounts will both be reported, but the mean of them is the primary result).
+	 * @remarks
+	 * The benchmark will measure memory use inn step 2 and freed in step 4.
```
alexvy86:
```suggestion
	 * The benchmark will measure memory allocated in step 2 and freed in step 4.
```

tools/benchmark/src/memoryBenchmarking/configuration.ts
```diff
@@ -0,0 +1,85 @@
...
+	 * A valid use of this function (to avoid errors anc collect accurate data) these amounts should be the same
```
alexvy86: Double check this one, not sure I fully understand the original intent of the sentence.
```suggestion
	 * In a valid use of this function (to avoid errors and collect accurate data) these amounts should be the same
```

tools/benchmark/src/memoryBenchmarking/configuration.ts
```diff
@@ -0,0 +1,85 @@
...
+	 */
+	benchmarkFn(state: MemoryUseCallbacks): Promise<void>;
+
+	/**
+	 * When set, async garbage collection will be used.
+	 * @remarks
+	 * Defaults to false.
+	 * Enable this to handle cases where a FinalizationRegistry is used and pending finalizers need to be run.
+	 * When enabling this, the `await` will leak some memory,
+	 * which can be mitigated by using {@link MemoryUseCallbacks.afterDeallocation} as the error to that will cancel out the error for the allocation amount.
```
alexvy86: Not sure I get this part, specifically how using `afterDeallocation` cancels out the error.

tools/benchmark/src/test/Benchmark.unit.tests.ts
```diff
@@ -48,7 +54,7 @@ describe("`benchmark` function", () => {
 	});
 
 	it("runBenchmark sync", async () => {
-		await runBenchmark({
+		await collectDurationData({
```
alexvy86: I think test names need to be updated here and for the one in line 78.

tools/benchmark/src/test/Benchmark.unit.tests.ts
```diff
@@ -204,6 +210,112 @@ describe("`benchmark` function", () => {
...
+				minBatchCount: 1,
+				maxBenchmarkDurationSeconds: 0,
+				benchmarkFn: () => {
+					// no-op
+				},
+			}),
+		});
+	});
+
+	describe("duplicate suite name", () => {
```
alexvy86: Some docs on these suites about what they're testing would be useful.

alexvy86 (commented):
Haven't looked at the biggest files yet, but here's another batch of comments.

tools/benchmark/api-report/benchmark.public.api.md
CraigMacomber: If someone is writing a reporter for a runner other than Mocha (there is an internal codebase which did this for jest) they should use this. There are a few APIs which are exported for such purposes (and some missing ones: I need to fix that)

tools/benchmark/src/durationBenchmarking/getDuration.ts
CraigMacomber: Docs and integration tests now cover reasons to use collectDurationData (which are why its exported)

tools/benchmark/src/test/RunnerUtilities.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+			assert.equal(
+				formatMeasurementValue({ name: "items", value: 1000, units: "count" }),
+				"1,000",
+			);
+		});
+
+		it("throws for non-integer count measurements", () => {
+			assert.throws(() =>
+				formatMeasurementValue({ name: "items", value: 1.5, units: "count" }),
+			);
```
alexvy86: Validate error is the expected one.

tools/benchmark/src/test/RunnerUtilities.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+				"1.00 s/op",
+			);
+		});
+
+		it("keeps ns/op units when scaleUnits is false", () => {
+			const result = formatMeasurementValue(
+				{ name: "time", value: 1_000_000, units: "ns/op" },
+				false,
+			);
+			assert.match(result, /ns\/op/);
```
alexvy86: We should be able to check for the exact value, no?

tools/benchmark/src/RunnerUtilities.ts
```diff
@@ -31,22 +24,84 @@ export function prettyNumber(num: number, numDecimals = 3): string {
...
+ * @param values - The values to compute the geometric mean of.
  */
 export function geometricMean(values: number[]): number {
 	// Compute the geometric mean of values, but do it using log and exp to reduce overflow/underflow.
 	let sum = 0;
 	for (const value of values) {
-		assert(value > 0, "invalid value in geometricMean");
+		if (value <= 0) {
+			return 0;
+		}
```
alexvy86: Is this better than the assertion? I remember memory measurements could result in negative values because they're hard to compute accurately, so maybe that's the reason... but I wonder if not throwing here could mask problems.

tools/benchmark/src/Reporter.ts
```diff
@@ -62,135 +88,149 @@ export class BenchmarkReporter {
...
 
 	/**
-	 * Appends a prettified version of the results of a benchmark instance provided to the provided
-	 * BenchmarkResults object.
+	 * Registers the start of a suite and pushes it onto the stack.
+	 * Must be called before any {@link BenchmarkReporter.recordTestResult} calls for tests in the suite,
+	 * and must be paired with a matching {@link BenchmarkReporter.recordSuiteResults} call.
+	 * When using mocha, call this from `EVENT_SUITE_BEGIN`.
+	 *
+	 * @param localName - the local `suite.title` string (tags stripped), used in JSON output.
```
alexvy86: Find it a bit clearer.
```suggestion
	 * @param localName - the name of the suite (in Mocha,`suite.title`), without any tags added by this package. Will be used in JSON output.
```

tools/benchmark/src/reporterUtilities.ts
alexvy86, line 1: Double checking: the suite-stack approach doesn't have issues with parent mode because there's no concurrency, correct? But mocha's `--parallel` mode in the mix would probably have bad consequences.

tools/benchmark/src/test/durationBenchmarking/getDuration.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+	});
+
+	it("collectDurationData output validation", async () => {
+		const before = Configuration.isInPerformanceTestingMode;
+
+		// This test checks that when in performance testing mode,
+		// that the output of duration tests is as expected.
+		// Currently doing this requires forcing `isInPerformanceTestingMode` to true,
+		// which is not ideal but is worth it to have this test.
+		// A better option would to allow setting `isInPerformanceTestingMode` as a parameter to override a default from the environment,
```
alexvy86:
```suggestion
		// A better option would be to allow setting `isInPerformanceTestingMode` as a parameter to override a default from the environment,
```

tools/benchmark/src/test/durationBenchmarking/getDuration.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+		// but that is not currently supported.
+		const mutableConfig = Configuration as Mutable<typeof Configuration>;
+
+		let result: CollectedData;
+
+		// It is critically import to restore this value after the test to avoid messing up unrelated tests,
```
alexvy86:
```suggestion
		// It is critically important to restore this value after the test to avoid messing up unrelated tests,
```

tools/benchmark/src/test/benchmark.integration.ts
```diff
@@ -0,0 +1,293 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { strict as assert } from "node:assert";
+// For these tests, import from the public AIP like a user would.
```
alexvy86:
```suggestion
// For these tests, import from the public API like a user would.
```

tools/benchmark/src/test/benchmark.integration.ts
```diff
@@ -0,0 +1,293 @@
...
+					foo.value++;
+				},
+			});
+		},
+	});
+
+	// Putting these in their own suite helps keep the reporter output from having way too many columns in any given table.
+	describe("memory use", () => {
+		benchmarkIt({
+			title: "collectMemoryUseData test",
```
alexvy86:
```suggestion
			title: "benchmarkMemoryUse test",
```

tools/benchmark/src/test/benchmark.integration.ts
```diff
@@ -0,0 +1,293 @@
...
+						foo.value = new Array(1000000).fill("leak");
+						await state.whileAllocated();
+					}
+					// After test custom cleanup
+					foo.value = -1;
+				},
+			}),
+		});
+
+		// This pattern isn't very useful since does not allow much what beyond benchmarkMemoryUse does.
```
alexvy86:
```suggestion
		// This pattern isn't very useful since it does not allow much beyond what benchmarkMemoryUse does.
```

tools/benchmark/src/test/benchmark.integration.ts
```diff
@@ -0,0 +1,293 @@
...
+					{
+						name: "extra data",
+						value: 1,
+					},
+				];
+			},
+		});
+	});
+
+	describe("duplicate suite name", () => {
```
alexvy86, line 219:
```suggestion
	// This test and its sibling with the same name exist so we can validate (by looking at the output)
	// that duplicate names are handled correctly (reported separately, not merged/mixed in weird ways).
	describe("duplicate suite name", () => {
```

alexvy86 (commented):
Working through the memory files still. A bit concerned about not seeing `ENABLE_MEM_REGRESSION` in the new code. Although I don't recall if it's currently being leveraged. @WayneFerrao do you remember?

tools/benchmark/src/test/memoryBenchmarking/getMemoryUse.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+		title: "no after",
+		...benchmarkMemoryUse({
+			benchmarkFn: async (callbacks) => {
+				while (callbacks.continue()) {
+					const a: number[] = [];
+					await callbacks.beforeAllocation();
+					a.length = 1024;
+					a.fill(0);
+					await callbacks.whileAllocated();
+					// Use "a" to ensure that memory is not freed until whileAllocated.
```
alexvy86: I think I got this right? The other instances below could use the same update if so.
```suggestion
					// Use "a" after the whileAllocated call to ensure that memory is not freed before that call completes
```

tools/benchmark/src/reporterUtilities.ts
CraigMacomber: we don't have a reporter in the child process, so there is no impact on how this works in parent process mode.

I'm not sure exactly how reporters and the events our reporter uses work in --parallel. I think we would have to have one per process, or the suite start and end events we subscribe to are all serialized back to one reporter and don't nest.

I'll look into it.

tools/benchmark/src/Reporter.ts
CraigMacomber: We don't tag suites, and I removed the stripping of tags from suites, so:
```suggestion
	 * @param localName - the name of the suite (in Mocha, `suite.title`).
```

tools/benchmark/src/RunnerUtilities.ts
CraigMacomber: I don't think crashing a possible several minute performance collection would be better than listing a geometric mean  of 0 (infinitely good) if some test which claims lower is better for its state gives 0 or a negative value.

Not only is it clear when looking at a geometric mean of 0 that one of the entries multiplied together must have been 0, but the report that includes the geometric mean also includes all the values which were multiplied to get it, so nothing is lost.

Including negative values would be very misleading as odd numbers of negative values would result in the answer being a complex number, and even amounts of negative values would show up as if they were all positive which is super wrong.

```suggestion
		if (value <= 0) {
			// In this context, smaller numbers are considered better, and 0 is infinitely good
			// (drowns out all other data from the entire geometric mean).
			// A negative value is thus better than infinitely good, which we can approximate as infinitely good,
			// and thus 0.
			// Generally, tests should not produce 0 if they want to use the geometric mean for anything,
			// but we don't have a simply way to know if the user cares about the geometric mean,
			// nor a way to know if 0 or negative values are valid.
			// Thus, for now, we simply cap the geometric mean at 0 for these cases to ensure negative values don't produce a seemingly meaningful but actually misleading result.
			// As this happens at the end of an often very slow data collection,
			// we really don't want to throw here and lose that data, which contains everything the user would need to see why the geometric mean is 0.
			return 0;
		}
```

tools/benchmark/src/reporterUtilities.ts
CraigMacomber: From what I can tell, this should not be an issue with --parallel.

What is an issue with --parallel is that random properties which mocha's types claim exist are missing. `.error` on hooks being one of them which we could mitigate, but `.on` on tests being missing is problematic as we need that to get the actual results in the reporter.

CraigMacomber: After a major refactor, --parallel is now supported, except when using --parentProcess (because --parallel screws with the process arguments causing the child process not be started properly)

tools/benchmark/src/test/tsconfig.json
CraigMacomber: We already trigger on changes to build-common which is what is used here.

tools/benchmark/src/test/Benchmark.unit.tests.ts
CraigMacomber: Done.

tools/benchmark/src/memoryBenchmarking/configuration.ts
CraigMacomber: I wrote a large explanation.

tools/benchmark/CHANGELOG.md
CraigMacomber: Yes. Its much better than the arithmetic mean since it treats making a test twice as fast as equally good, regardless of if the test took 10 seconds or 10 nano seconds (arithmetic mean would consider one of those a billion times more important than the other)

tools/benchmark/src/Configuration.ts
```diff
@@ -270,38 +156,33 @@ export function qualifiedTitle(
...
- * Instead child process will be created to run each test.
- * This has some overhead, but can reduce noise and cross test effects
- * (ex: tests performing very differently based on which tests ran before them due to different jitting).
- * This does not (and can not) remove all causes for effects of earlier tests on later ones.
- * Ex: cpu temperature will still be an issue, and thus running with fixed CPU clock speeds is still recommend
- * for more precise data.
- */
-export const isParentProcess: boolean = process.argv.includes("--parentProcess");
+export const isInPerformanceTestingMode =
+	process.argv.includes("--perfMode") || "FLUID_TEST_PERF_MODE" in process.env;
```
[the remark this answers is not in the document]
CraigMacomber: Do we have any sort of convection or library for how we are supposed to check truthiness of environment variables?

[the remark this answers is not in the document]
alexvy86: Nothing documented that I'm aware of, but seems like mostly we just check against undefined. I wouldn't object to having a better check, even something like `=== "1"` to activate and keep deactivated in all other cases.

tools/benchmark/src/assert.ts
```diff
@@ -3,13 +3,14 @@
  * Licensed under the MIT License.
  */
 
+import chalk from "chalk";
```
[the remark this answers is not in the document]
CraigMacomber: Chalk should autodetect support, and modern browser have support (at least some of them, see https://github.com/chalk/chalk?tab=readme-ov-file#browser-support), so I think this is fine.

tools/benchmark/src/Configuration.ts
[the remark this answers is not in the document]
CraigMacomber: if we compare to "1", then setting it to "true" would set this to false which seems not great. "1" seems like a questionable way to encode "true" in a string (I know we use it and would like to support it, but it seems like a poor only option)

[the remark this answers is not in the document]
CraigMacomber: I went ahead and make the change to support 1, and true (case insensitive)

tools/benchmark/README.md
```diff
@@ -1,186 +1,203 @@
...
-You can change the `@Benchmark` tag to a few other values (like `@Measurement`, `@Perspective`, or `@Diagnostic`) with
-one of the arguments to the functions exposed in this package, and if you do, you can be more specific about which tests
-want to run by passing a different filter, e.g. `--fgrep @Measurement`.
-
-### `--fgrep @ExecutionTime` or `--fgrep @MemoryUsage`
-
-You'll also want to use one of these to only run execution-time/runtime or memory usage tests.
-You can technically run them both at the same time, but the custom mocha reporters (one for runtime tests, one for memory
-usage tests) expect to only see tests of their corresponding type, so in order to use those you'll have to use `--fgrep`
-as described here and do two separate test runs, one for each type of test.
```
alexvy86, line 52: Do we not need this anymore? A package that has both kinds of benchmark tests probably wants to run them separately, no?

tools/benchmark/src/mocha/runnerUtilities.ts
```diff
@@ -5,19 +5,30 @@
...
 	test: Test,
 ): Promise<void> {
 	const { exception, result } = await f();
-	test.emit("benchmark end", result);
+	// Only emit results in perfMode
+	if (isInPerformanceTestingMode) {
+		test.emit("benchmark end", result);
+		// If test is running in parallel mode, the reporter cannot subscribe to events on tests.
+		// It can however access the test body, so cramming the results in there makes them usable on the reporter side.
+		test.body = JSON.stringify(result);
```
alexvy86, line 31: This feels weird... do we know what Mocha would normally use `body` for?

tools/benchmark/src/memoryBenchmarking/getMemoryUse.ts
```diff
@@ -0,0 +1,378 @@
...
+		) {
+			return { used: usedAfter, gcIterations: counter, lastDelta: change };
+		}
+		usedBefore = usedAfter;
+		counter++;
+
+		// Experiments have found two ways to ensure GC runs including the FinalizationRegistry finalizers:
+		// 1. A sync GC then an 8-second wait (unreliable across multiple runs unless a debugger takes a heap snapshot, possibly due to JIT).
+		// 2. Two async GCs in a row.
+		// Option 2 is both more robust and faster, so it is used here.
```
alexvy86, line 69: The code does only one async GC though, and then a sync one (two including the one without the `major` arg). Comment needs some adjustment?

tools/benchmark/src/memoryBenchmarking/getMemoryUse.ts
```diff
@@ -0,0 +1,378 @@
...
+		return { allocatedBytes, freedBytes, meanBytes: (allocatedBytes + freedBytes) / 2 };
+	});
+
+	const processed: ProcessedMeasurement[] = processedAll.slice(trimCount);
+
+	// When debugging the behaviors, inspecting this can be helpful to see if there is noise coming from specific iterations and the trim needs adjusting.
+	if (args.logRawData) {
+		console.log(`Raw data (First ${trimCount} rows discarded):`);
+		console.log(data);
+	}
```
alexvy86, line 216: Should this be printing `trimmed`, or the comment not say any rows are discarded?

tools/benchmark/src/memoryBenchmarking/getMemoryUse.ts
```diff
@@ -0,0 +1,378 @@
...
+
+	// Split up data into "first" and "last" halves and compute stats on those
+	// to see if there is a trend in the data that might indicate leaking memory across iterations.
+	const sizeOfHalf = Math.max(1, Math.floor(processed.length / 2));
+	const firstHalfStats = getArrayStatistics(
+		processed.slice(0, sizeOfHalf).map((x) => x.meanBytes),
+	);
+	const lastHalfStats = getArrayStatistics(processed.slice(-sizeOfHalf).map((x) => x.meanBytes));
+	assert(firstHalfStats.samples.length === sizeOfHalf, `invalid first half size`);
+	assert(lastHalfStats.samples.length === sizeOfHalf, `invalid last half size`);
```
alexvy86, line 237: This won't have issues with an odd number of samples in `processed`?

tools/benchmark/src/memoryBenchmarking/getMemoryUse.ts
```diff
@@ -0,0 +1,378 @@
...
+
+	const averageIndexInStart = (0 + (sizeOfHalf - 1)) / 2;
+	const averageIndexInEnd = (processed.length - sizeOfHalf + (processed.length - 1)) / 2;
+	const iterationsBetweenStartAndEndStats = averageIndexInEnd - averageIndexInStart;
+
+	const startBeforeStats = getArrayStatistics(trimmed.slice(0, sizeOfHalf).map((x) => x.before));
+	const endBeforeStats = getArrayStatistics(trimmed.slice(-sizeOfHalf).map((x) => x.before));
```
alexvy86: Nit: `firstHalfBeforeStats` and `lastHalfBeforeStats` for consistency with the ones above.

alexvy86 (commented):
Done looking at all files, a few more comments. Also still concerned about the memory regression functionality that seems to not have made it over and I pointed out [here](https://github.com/microsoft/FluidFramework/pull/26638#pullrequestreview-3905889020).

tools/benchmark/README.md
CraigMacomber: There are infinite possible kinds of benchmarks you can write, and you can run what ever custom subset of them you want. There is no inherent reason you should do separate runs and reports for memory vs execution time vs other custom things. They all work together properly with the same reporter, and all use the same APIs to report their data. You could even write a single test which measured both runtime and memory use if you wanted.

Also given that mocha silently ignores multiple -fgrep arguments, you really just want to do the narrowing once, by optionally making the top level filter more specific.

tools/benchmark/src/mocha/runnerUtilities.ts
CraigMacomber: Yes, it's a horrible hack. Body normally has the source code of the test in it, which is not something we care about.

Personally I'd prefer if Mocha's docs for writing reporters in parallel mode wasn't that sometimes random properties are undefined and that breaks reporters, and if you are using a reporter that doesn't work, open an issue with mocha about the broken reporter (with no instructions for reporter authors or details on which things don't exist).

Given we are a major version behind on mocha, and that process is for users of reporters not authors, I didn't open such an issue and just found this hack instead.

I'll add some details to the comment.

tools/benchmark/src/memoryBenchmarking/getMemoryUse.ts
CraigMacomber: I have clarified the comment below to better explain this.

tools/benchmark/src/memoryBenchmarking/getMemoryUse.ts
CraigMacomber: The comment is saying which rows of thew data being printed will be discarded.

One of main reasons to use this logging is to tune how much data is discarded, or ro look for odd patterns in the data. Both cases would rather have all of the data rather than just the trimmed data.

I have updated the message (and the one below) to clarify.

tools/benchmark/src/memoryBenchmarking/getMemoryUse.ts
CraigMacomber: Comment updated to clarify how that works.

tools/benchmark/src/memoryBenchmarking/getMemoryUse.ts
CraigMacomber: Done.

tools/benchmark/src/mocha/runnerUtilities.ts
CraigMacomber: Updated.

tools/benchmark/README.md
alexvy86: Ok yeah, I had to think a bit about a new state of things... we'll want to consolidate the handlers that take the reports and write to Kusto, and packages to stop splitting their benchmark-running npm scripts, right?

CraigMacomber: Yes, we can use a single handler. We still might want to keep separate runs since its safer from a data quality perspective to parallelize non execution time tests more, but the format should be the same and the handlers should be the same as well. I have some in progress refactoring the upload code, but now that we export some utils from this package which that code can consume, finishing that work is easiest after this is merged.

CraigMacomber:
> Working through the memory files still. A bit concerned about not seeing `ENABLE_MEM_REGRESSION` in the new code. Although I don't recall if it's currently being leveraged. @WayneFerrao do you remember?

Looking at our codebase, none of our script enable that, and none of our tests set a baselineMemoryUsage, and even if they did, I don't think it would work well as we have a ton of noise and also throwing would make the test with a regression as failed and make us hide the memory use information in the output files, and we would filter out failed tests from the data we reported, so it would hide regressions rather than emphasize them I think.

alexvy86 (approved):
I might have missed something but overall I think it's in very good shape and we should iterate based on trying to integrate it into the client release group.

PR #26639, opened 2026-03-04
Update .gitignore for Claude Code folders

anthony-murphy:
## Summary

Updates the Claude Code section of the root `.gitignore` to better control what gets checked in across the repo.

## What changed

Previously, only `.claude/settings.local.json` was ignored. This PR replaces that with:

| Rule | Purpose |
|---|---|
| `/.claude/` | Exclude the root `.claude` folder (anchored to repo root) |
| `!/.claude/skills/` | Re-include the `skills/` subfolder |
| `**/.claude/*.local.*` | Exclude local files in any `.claude` folder across the repo |

## Why

- **Root `/.claude/` is excluded** — The root `.claude` folder contains project-level settings (`settings.json`, hooks, etc.) that tend to be author-specific and shouldn't be imposed on the whole team.
- **`/.claude/skills/` is re-included** — Skills are reusable, shareable prompt templates. Checking them in lets the team build a shared library of Claude Code skills that anyone can invoke.
- **Nested packages work out of the box** — The rules are anchored with a leading `/`, so they only apply at the repo root. Individual packages can freely check in their own `.claude/` folders (e.g. a `CLAUDE.md` with build instructions) without needing per-package gitignore exceptions. Only `*.local.*` files are excluded everywhere, keeping author-specific config out of source control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,8 +58,8 @@ build_output_archive
 # Claude Code
 # Exclude the root .claude folder to minimize automatically loaded context,
 # but allow reusable skills to be checked in and shared.
-.claude/
-!.claude/skills/
-# Nested packages control their own .claude folders, but exclude local files
-# (e.g. settings.local.json) which contain author-specific configuration.
+/.claude/
+!/.claude/skills/
+# Exclude author-specific local files in any .claude folder across the repo.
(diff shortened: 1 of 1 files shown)


jason-ha:
Personally prefer #26633 instead of this.

PR #26642, opened 2026-03-04
build: disable formatWithErrors in biome

tylerbutler:
Disable the `formatWithErrors` setting in biome, which should address the problem folks have reported where biome further "corrupts" a file with a parser error, especially when using format-on-save.


jason-ha (approved):
Do kind of like "wosrse" though 😛

jason-ha (approved):

alexvy86 (approved):

PR #26645, opened 2026-03-04
Add missing build:test script to inventory-app and text-editor

frankmueller-msft:
## Summary

- Adds `"build:test": "npm run build:test:esm"` to `inventory-app` and `text-editor`
- These packages had `build:test:esm` but no `build:test`, so the `compile → build:test` dependency chain skipped test compilation
- Previously test files were only compiled via the `lint → eslint → build:test:esm` path. With lint moved to a parallel CI job (`ci:lint`), this path no longer runs during `ci:build`, causing coverage tests to fail (`ERR_MODULE_NOT_FOUND: lib/test/mochaHooks.js`)
- Aligns these packages with the convention used by the rest of the repo

## Test plan

- [x] Coverage tests pass (verified in prior CI runs)


fluidBuild.config.cjs
```diff
@@ -34,7 +34,6 @@ module.exports = {
 		"ci:build": {
 			dependsOn: [
 				"compile",
-				"lint",
 				"ci:build:api-reports",
```
CraigMacomber, line 38: I think just moving lint is a good starting point and likely the right choice for this PR, but in the future, we should be able to move "ci:build:api-reports" as well I think since all that does is fail the build if they are out of date.

The same MIGHT be true for some of "ci:build:docs", "build:manifest", "build:readme", but I'm not sure.

examples/data-objects/inventory-app/package.json
```diff
@@ -16,6 +16,7 @@
 		"build": "fluid-build . --task build",
 		"build:compile": "fluid-build . --task compile",
 		"build:esnext": "tsc --project ./tsconfig.json",
+		"build:test": "npm run build:test:esm",
```
CraigMacomber, line 19: This change, and the other one like it seem good to me. Looks like its fixing a bug from when I ported these tests to esm.

tools/pipelines/build-client.yml
```diff
@@ -206,17 +206,16 @@ extends:
...
     publishDocs: true
     # We only care about pipeline run telemetry for the CI runs on the internal project, not for PR runs in the public
     # one. And since we don't batch commits for CI pipeline runs, the only reason we see for automated runs is IndividualCI.
     telemetry: ${{ eq(variables['Build.Reason'], 'IndividualCI') }}
     shouldShip: ${{ or(eq(variables['release'], 'release'), eq(variables['release'], 'prerelease')) }}
     taskTest:
-    # This check must be run after the build, since it relies on built files being present. Eventually it might be moved
-    # to the "pack" stage since it can use the already-packed packages in that case. As it is the pipeline packs some
-    # packages twice.
+    - { name: "lint", jobName: "Lint" }
```
CraigMacomber: This does not seem to work.

Reviewing the output of this task on CI ( https://dev.azure.com/fluidframework/public/_build/results?buildId=383321&view=logs&j=006c0001-782a-5004-c94b-b7174ee579be&t=6c242987-8d02-5616-35d4-17ab9c387bed ) shows it is not linting all the packages.

A good sanity check would be to review the contents of the build task before and after this change by looking at the CI logs, and ensure the items removed show up in the new task.

Without this change, there were about 1510 tasks run in the CI build. After there are 931 in build and 10 in lint. THis implies there are over 500 tasks missing. This includes things like all of the eslint runs, which is the main part of the linting. 

For example, the linting of tree is missing ( a build which includes it shows a line like: https://dev.azure.com/fluidframework/public/public%20Team/_build/results?buildId=383339&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=bfe27d5f-9219-57c0-0d67-1b2946a32921&l=1785 )

frankmueller-msft: Good catch. The root cause is that `fluid-build` is incremental — after extracting the build artifacts, it sees outputs and thinks lint tasks are up-to-date. Only ~10 tasks appeared "stale" enough to run.

Fixed by adding a `ci:lint` script that runs `fluid-build --task lint --force`, and using that in taskTest. The `--force` flag bypasses the incrementality check so all lint tasks run unconditionally in the parallel job.

tools/pipelines/build-client.yml
```diff
@@ -206,17 +206,16 @@ extends:
...
     # We only care about pipeline run telemetry for the CI runs on the internal project, not for PR runs in the public
     # one. And since we don't batch commits for CI pipeline runs, the only reason we see for automated runs is IndividualCI.
     telemetry: ${{ eq(variables['Build.Reason'], 'IndividualCI') }}
     shouldShip: ${{ or(eq(variables['release'], 'release'), eq(variables['release'], 'prerelease')) }}
     taskTest:
-    # This check must be run after the build, since it relies on built files being present. Eventually it might be moved
-    # to the "pack" stage since it can use the already-packed packages in that case. As it is the pipeline packs some
-    # packages twice.
+    # This check must be run after the build, since it relies on built files being present.
+    - { name: "lint", jobName: "Lint" }
```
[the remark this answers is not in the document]
frankmueller-msft: Good catch. Added `BASELINE_BROWSER_MAPPING_IGNORE_OLD_DATA: 1` to both branches (coverage and no-coverage) in `include-test-task.yml`. This is harmless for non-lint tasks and prevents the 100+ warning spam for lint.

fluidBuild.config.cjs
frankmueller-msft: Agreed — `ci:build:api-reports` is a good next candidate. It just validates that reports are up-to-date, so it doesn't need to block the rest of the build. Keeping this PR focused on lint for now.

alexvy86 (commented):
I'm not sure I like the approach, for a couple of reasons:

- If the code has linting errors, do we want to spend the time building it or running other parallel tasks post-build for it? Blocking the success of `ci:build` on the success of linting seems correct to me.
- Other release groups have a `ci:build` script, and if that won't force linting during pipeline runs for them, then we need to update their pipelines too to run linting separately. That feels like a recipe for forgetting to do it in some place.

tools/pipelines/build-client.yml
tylerbutler: But why _isn't_ fluid-build up-to-date? If the build has happened and the outputs are there, there should be no reason to force the tasks to run again.

frankmueller-msft:
@alexvy86 Good points, both of them. I've descoped the PR — removed all the lint parallelization and pipeline changes. What's left is just the `build:test` fix for `inventory-app` and `text-editor`, which were missing the script and failing coverage tests. The lint-as-parallel-job idea needs more thought and isn't worth blocking this bugfix on.

alexvy86 (approved):
The current scope per the comment above seems good to me.

tylerbutler (approved):

PR #26646, opened 2026-03-04
ci: add fluid-release skill and dispatcher workflow

tylerbutler:
## Summary

- Add `fluid-release` Claude skill with reference docs for minor/patch release workflows, release scheduling, and type test updates
- Add GitHub Actions dispatcher workflow (`fluid-release-skill-dispatcher`) that runs the release skill on a weekday schedule and via manual dispatch


frankmueller-msft (commented):
## Review Feedback

Nice architecture here — the safe outputs pattern, threat detection gate, read-only MCP, and human-in-the-loop for the most dangerous steps (merging, release branch creation, ADO builds) are all solid design choices. SHA-pinned actions and domain allowlist firewall are good hygiene.

A few things I'd want addressed:

### Security / Blast Radius

1. **`--allow-all-tools --allow-all-paths` on the agent execution**: The firewall provides the outer boundary, but inside the sandbox the agent is unconstrained. Worth documenting why this is acceptable.

2. **Token fallback chain** (`GH_AW_GITHUB_MCP_SERVER_TOKEN || GH_AW_GITHUB_TOKEN || GITHUB_TOKEN`): If the restricted tokens aren't configured, this silently falls back to `GITHUB_TOKEN` which may have broader permissions. Should this fail loudly instead of silently degrading?

3. **`actions: write` on `safe_outputs` and `conclusion` jobs** is broader than "dispatch these 3 specific workflows" — it covers all Actions in the repo.

### Autonomous Mode Risks (SKILL.md)

4. **`release-blocking` issue creation on failure**: A confused or hallucinating agent filing spurious blocking issues could halt the real release process. Consider: (a) a bot-specific label prefix like `bot/release-blocking`, and (b) dedup logic — the skill has no guard against opening duplicate issues on retries or repeated cron failures.

5. **Pushing `release-prep/` branches directly to upstream**: `release/` branches are protected, but `release-prep/` apparently are not. A misbehaving agent could pollute the upstream repo with junk branches. Is there a cleanup policy or branch protection rule for these?

6. **Version inference from a static schedule file**: If `release-schedule.md` goes stale or date logic hits an edge case, the agent runs `pnpm -r --include-workspace-root exec npm pkg set version=<WRONG>` across every `package.json` in the monorepo. The PR review safety net exists, but the noise would be significant. Who owns keeping the schedule current?

7. **No rate limiting on issue creation**: The workflow runs on a weekday cron. If it fails repeatedly, it could file a new blocking issue every day.

### Minor

8. The 20-minute agent timeout is generous — if the agent loops, that's a lot of compute burn before termination.

9. The workflow markdown source (61 lines) compiles into 1069 lines of YAML via `gh-aw`. The generated output is hard to audit against the markdown intent — is there a way to verify the compilation is faithful?

Overall: the defense-in-depth is real, but the autonomous SKILL.md instructions (PR creation, branch pushing, issue filing against the production repo) are the main risk surface. Dedup/rate-limiting on issue creation, explicit failure on the token chain, and a maintenance plan for the release schedule would tighten this up.

frankmueller-msft (approved):

tylerbutler:
Thanks @frankmueller-msft - I'm flagging the feedback for future iterations of the workflow since this is a first draft.

PR #26648, opened 2026-03-04
fix: update mongodb to 5.x in server-services

frankmueller-msft:
## Summary

- Updates `mongodb` from `4.17.1` to `^5.9.2` in `@fluidframework/server-services`
- MongoDB 5.x moves `@aws-sdk/credential-providers` from a direct optional dependency to an optional peer dependency, reducing the transitive dependency footprint
- All existing MongoDB usage in the codebase is promise-based (no callbacks), which is the main breaking change in the 4.x to 5.x migration

## Test plan

- [ ] Server build passes (routerlicious)
- [ ] Server integration tests pass
- [ ] Verify no callback-based MongoDB usage exists (confirmed — all async/await)

### DIFF
--- a/server/routerlicious/packages/services/package.json
+++ b/server/routerlicious/packages/services/package.json
@@ -69,7 +69,7 @@
 		"ioredis": "^5.6.1",
 		"lodash": "^4.17.21",
 		"lru-cache": "^6.0.0",
-		"mongodb": "4.17.1",
+		"mongodb": "^5.9.2",
 		"nconf": "^0.12.0",
 		"socket.io": "^4.8.0",
 		"telegrafjs": "^0.1.3",
--- a/server/routerlicious/pnpm-lock.yaml
+++ b/server/routerlicious/pnpm-lock.yaml
@@ -1051,8 +1051,8 @@ importers:
         specifier: ^6.0.0
         version: 6.0.0
       mongodb:
-        specifier: 4.17.1
-        version: 4.17.1
+        specifier: ^5.9.2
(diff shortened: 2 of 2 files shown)


frankmueller-msft:
Both Copilot review comments have been addressed in the latest revision:

1. **Open-ended range** — Changed from `>=5.3.5` to `^5.3.5`, constraining to 5.x to prevent unexpected major version bumps.
2. **commentsOverrides entries** — Removed to keep the change minimal and consistent with other overrides in these files.

alexvy86 (commented):
A couple of things:

- 2 changes seem to have been conflated here: updating mongo in routerlicious, and updating fast-xml-parser in historian. They should be in separate PRs.
- The change in historian seems to have been done incorrectly; the lockfile shouldn't have a new override without it being present in package.json as well.
- Can we investigate why mongo 4 doesn't disappear from the lockfile in routerlicious? What brings an old version?

frankmueller-msft:
Thanks @alexvy86 — all three addressed:

1. **Separate PRs** — Removed the historian change. This PR now only touches routerlicious (`packages/services/package.json` + lockfile).

2. **Stale historian lockfile** — That was a leftover from an earlier approach where we used pnpm overrides. Removed from the commit.

3. **Why mongodb 4.x remains in the routerlicious lockfile** — It's pulled in by `@fluidframework/server-services-previous` (aliased to `npm:@fluidframework/server-services@7.0.0`), which is a devDependency used for type compatibility testing. That published version pins `mongodb@4.17.1`. It doesn't ship to customers and won't be affected by this change.

alexvy86 (approved):

PR #26649, opened 2026-03-04
build: initialize hk git hook manager with biome format

tylerbutler:
## Summary

- Adds `hk.pkl` configuration to initialize the [hk](https://hk.jdx.dev/) git hook manager
- Configures a `biome-format` pre-commit hook that auto-formats staged files using `biome format --write` and re-stages the results
- Extends the built-in `Builtins.biome` step with format-only commands (no linting)


alexvy86 (commented):
I'm tempted to approve this blindly just because [HK has a sea-shanty](https://hk.jdx.dev/shanty.html) 😆 .

I feel a bit conflicted about the additional toolset required for this (more things that people (might) need to know to work on the repo), but being opt-in mitigates that somewhat. Should we consider this an "experimental" thing in the repo for now, and eventually graduate it to a formal "this is a tool we're fully committed to leveraging/supporting" (even if it's still opt-in)? I think that's my main piece of feedback, I'd personally would go with docs (and comments in the `pk1` file) that suggest that vs "in this repo we use X".

alexvy86 (approved):

PR #26651, opened 2026-03-05
Git-ignore CLAUDE.local.md

alexvy86:
## Description

`CLAUDE.local.md` goes outside the `.claude/` folder so we needed another entry for it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,7 @@ build_output_archive
 # Claude Code
 # Local settings contain author-specific information and should not be shared.
 **/.claude/**/*.local.*
+CLAUDE.local.md
 # Session-specific task tracking
 **/.claude/todos/
 # Git worktrees created by Claude Code


jikim-msft (approved):

jason-ha:
It doesn't have to go outside. You can place it inside .claude. But this is okay too.

alexvy86:
> It doesn't have to go outside. You can place it inside .claude. But this is okay too.

The docs are pretty clear that `CLAUDE.md` can go inside `.claude/`, but don't say the same thing for `CLAUDE.local.md` so I assumed it wouldn't work. I haven't tried though. Also need to test whether files are loaded from `.claude/` in subdirectories or that's only a root thing and `CLAUDE.md` needs to go directly in the subdirectory (vs `<subdirectory>/.claude`).

I'll go test that now to see if we want any further tweaks before merging this.

alexvy86:
Doesn't look like `.claude/CLAUDE.local.md` files in particular are loaded at all, at any point in the folder hierarchy. They have to be outside `.claude/` :

image

Interestingly, though, if I have `azure/.claude/CLAUDE.local.md` open and focused in VSCode it gets loaded, but that doesn't happen for `.claude/CLAUDE.local.md`. Feels like there might be subtle bugs in the implementation of how `CLAUDE.local.md` is loaded. But ignoring them outside `.claude/` seems like the correct thing to do today.

jason-ha:
I think my memory (just from two days ago) was wrong. https://code.claude.com/docs/en/memory#choose-where-to-put-claude-md-files does say NOT in .claude for CLAUDE.local.md

.gitignore
```diff
@@ -58,6 +58,7 @@ build_output_archive
 # Claude Code
 # Local settings contain author-specific information and should not be shared.
 **/.claude/**/*.local.*
+CLAUDE.local.md
```
jason-ha, line 61: Probably should be `**/CLAUDE.local.md` in case someone wants to target a subfolder.

alexvy86: As-is it matches anywhere in the folder hierarchy (like all the entries at the top of `.gitignore`). I suspect that would also work for the other Claude-related entries (don't need leading `**` specifically, just useful to cover all the folder hierarchy inside `.claude/`).

jason-ha: Of course. Which is something I keep forgetting.
If there a `/` then it is exactly relative and
If there is no `/` then it matches everywhere (implicit `**/`).

PR #26655, opened 2026-03-05
Use independent view sandbox for tree-agent edit isolation

noencke:
TODO: History will be updated to use serialized changes rather than code snippets [once this PR goes in](https://github.com/microsoft/FluidFramework/pull/26662).

SharedTree branches share state/runtime, which means errors during LLM edits on a sub-branch can corrupt the parent query branch and even the outer tree. Previously, tree-agent relied on making new branches to handle error cases, but ST in general does not guarantee that an error on a branch does not corrupt other branches or the ST itself. So, this PR replaces the fork-per-edit checkpointing strategy with an isolated independentView "sandbox". Essentially, the LLM uses the sandbox as a scratchpad to "try out" its edits, and once it is happy with them, we take all those edits (as code) and finally apply them to the original ST branch for real.                                                                                 

  - LLM edits are executed against an independent view that shares no state with the outer tree
  - Successful edit code strings are tracked; on error, the sandbox is rebuilt from a snapshot and prior successful edits are replayed
  - When the query completes, all successful edits are replayed on the snapshot branch and merged into the outer tree.

In addition to being more robust against errors, this "run the code again" strategy allows us to put all final edits in a transaction, which is ideal, and which is currently impossible when using the previous branch-based-checkpointing approach. However, it's worth noting one downside to this strategy: if the LLM-generated edits are not pure, then the final edit application could differ from the sandbox edit application. In practice, I expect this to matter rarely if ever. Things like current time or random numbers will be different between the sandbox and output, but it would be surprising if this caused the output to error while not producing an error in the sandbox. If we observe that this does happen in the future, we can mitigate it by providing the LLM with a deterministic RNG across both environments.

A smattering of new tests have been added which flesh out the existing test coverage, and include a test for transactionality.

### DIFF
--- a/packages/framework/tree-agent/src/subtree.ts
+++ b/packages/framework/tree-agent/src/subtree.ts
@@ -5,18 +5,19 @@
 
 import { fail } from "@fluidframework/core-utils/internal";
 import { UsageError } from "@fluidframework/telemetry-utils/internal";
-import { TreeNode, Tree, NodeKind } from "@fluidframework/tree";
+import { TreeNode, Tree, NodeKind, TreeViewConfiguration } from "@fluidframework/tree";
 import type {
 	ImplicitFieldSchema,
 	TreeFieldFromImplicitField,
 	TreeMapNode,
 	TreeArrayNode,
 } from "@fluidframework/tree";
-import { TreeAlpha } from "@fluidframework/tree/alpha";
+import { independentView, TreeAlpha } from "@fluidframework/tree/alpha";
 import type {
 	ReadableField,
 	TreeRecordNode,
 	TreeBranchAlpha,
+	InsertableField,
 } from "@fluidframework/tree/alpha";
 
 import type { TreeView, ViewOrTree } from "./api.js";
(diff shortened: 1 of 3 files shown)


packages/framework/tree-agent/src/subtree.ts
```diff
@@ -120,3 +121,20 @@ export class Subtree<TRoot extends ImplicitFieldSchema> {
...
+ * Initializes a {@link Subtree} backed by an independent view.
+ */
+export function copyIndependentSubtree<TSchema extends ImplicitFieldSchema>(
+	view: ViewOrTree<TSchema>,
+): Subtree<TSchema> {
+	const subtree = new Subtree(view);
+	const config = new TreeViewConfiguration({ schema: subtree.schema });
+	const sandboxView = independentView(config);
+	const exported =
+		subtree.field === undefined ? undefined : TreeAlpha.exportVerbose(subtree.field);
```
noencke, line 135: I considered adding a flag for caching the verbose tree for a given view, since we know the view doesn't change in our scenario. Or something like that. But I think it's a premature optimization.

PR #26656, opened 2026-03-05
Tree Agent API Refactor

steffenloesch:
## Description

Introduces a new stateless API for the tree-agent packages that separates concerns between the agent (edit loop, history management) and the model (single LLM invocation). The previous API coupled these responsibilities within `SharedTreeChatModel.query()`, which owned the edit loop internally.

The new API consists of:
- **`TreeAgentChatMessage` / `TreeAgentChatResponse`** — role-tagged message and response types that model the conversation between agent and LLM in a framework-agnostic way.
- **`SharedTreeChatModel.invoke(history)`** — a stateless model method that receives the full message history and returns a single response (either an edit tool call or a final text response).
- **`createTreeAgent(model, tree, options)`** — a factory that creates a `TreeAgent` whose `message(prompt)` method manages the edit loop, branching, merging/rollback, and history externally.
- **`createLegacyLangchainChatModel()`** — the old stateful LangChain adapter, now deprecated and renamed to preserve backward compatibility.

The legacy `SharedTreeSemanticAgent` class and `SharedTreeChatModel.query()` / `appendContext()` methods are deprecated but still functional. Shared logic (tree tracking, header logging, prompt generation) has been extracted into reusable helpers.

## Breaking Changes

No breaking changes. The existing `SharedTreeSemanticAgent` class, `createLangchainChatModel()`, and all existing exports continue to work. New exports are additive. Deprecated members are marked with `@deprecated` JSDoc tags.

Note: `createLangchainChatModel()` now returns a model that implements `invoke()` (new API) instead of `query()` (legacy API). Code using `SharedTreeSemanticAgent` with `createLangchainChatModel()` should switch to `createLegacyLangchainChatModel()` or migrate to `createTreeAgent()`.

## Reviewer Guidance

- The core design question is whether the agent-owns-history / model-is-stateless split is the right boundary. The main motivation is making it easier to test, swap models, and inspect conversation history from outside.
- `createTreeAgent` replicates the edit loop logic from `SharedTreeSemanticAgent.query()` but drives it via `invoke()` calls instead of the callback-based `SharedTreeChatQuery.edit()` pattern. Review both paths to confirm behavioral parity.
- The `convertToLangchainMessages` function in `packages/framework/tree-agent-langchain/src/chatModel.ts` maps the new message types to LangChain's `BaseMessage` hierarchy — worth checking that `tool_call` → `AIMessage` with `tool_calls` and `tool_result` → `ToolMessage` are correct for the LangChain providers you care about.
- Tests in `packages/framework/tree-agent/src/test/agent.spec.ts` cover the new `createTreeAgent` factory (single edit, multi-edit, error recovery, rollback, edit limits, input validation). Existing legacy tests are unchanged.


packages/framework/tree-agent/src/agent.ts
```diff
@@ -43,53 +47,114 @@ import {
...
+): {
+	outerTree: Subtree<TSchema>;
+	refreshIfDirty: (onDirty: (text: string) => void, logger: Logger | undefined) => void;
+	resetDirty: () => void;
+} {
+	const outerTree = new Subtree(tree);
+	let isDirty = false;
+
+	if (tree instanceof TreeNode) {
+		Tree.on(tree, "treeChanged", () => {
```
noencke: These subscriptions should be cleaned up at some point.

packages/framework/tree-agent/src/api.ts
```diff
@@ -237,6 +238,147 @@ export interface SharedTreeChatQuery {
...
+ * @remarks Created via {@link createTreeAgent}.
+ * @alpha
+ */
+export interface TreeAgent {
+	/**
+	 * Process the user's prompt, potentially editing the tree, and return a response.
+	 * @param prompt - The user's question or edit instruction.
+	 * @returns The model's text response.
+	 */
+	invokeAgent(prompt: string): Promise<string>;
```
noencke: nit: just call this `invoke`, it's already being called on an "agent"

packages/framework/tree-agent/src/agent.ts
```diff
@@ -303,3 +366,147 @@ export function createContext<TSchema extends ImplicitFieldSchema>(
...
+	options?: TreeAgentOptions<TSchema>,
+): TreeAgent {
+	const editToolName = model.editToolName;
+	if (editToolName === undefined) {
+		throw new UsageError(
+			"The provided SharedTreeChatModel does not have an editToolName. Editing requires a model with editToolName set.",
+		);
+	}
+
+	const { outerTree, history, tracker } = createAgentState(model, tree, editToolName, options);
```
noencke: You have enough state here that I think it would be prudent to model the `TreeAgent` implementation as a class. We don't need to expose the class, but I expect it will make the construction and state management more organized and readable.

packages/framework/tree-agent/src/agent.ts
```diff
@@ -43,53 +47,114 @@ import {
...
+	return `The tree has changed since the last query. The new state of the tree is: \n\n\`\`\`JSON\n${stringified}\n\`\`\``;
+}
+
+/**
+ * Creates a tree tracker that monitors a tree for external changes and can notify consumers.
+ * @param tree - The tree or subtree to track.
+ * @returns An object with the subtree wrapper, and a `refreshIfDirty` function that calls the provided
+ * callback with the tree-changed notification text if the tree has been modified since the last check.
+ */
+function createTreeTracker<TSchema extends ImplicitFieldSchema>(
```
noencke: Was part of the goal of this abstraction to deduplicate code across the old and new implementations of the agent? I find it rather confusing and a strange abstraction (e.g. why does it own the creation of the Subtree `outerTree`?). 

Can we upgrade Subtree with our established eventing pattern? E.g. something like:

```ts
interface SubtreeEvents {
	treeChanged(): void;
}
```
```ts
// ... in Subtree ...

public readonly events: Listenable<SubtreeEvents>;

// In constructor
		this.events = {
			on(eventName: "treeChanged", listener: () => void) {
				return viewOrTree instanceof TreeNode
					? Tree.on(viewOrTree, eventName, listener)
					: viewOrTree.events.on("changed", listener);
			},
			off(eventName: "treeChanged", listener: () => void) {
				return viewOrTree instanceof TreeNode
					? Tree.off(viewOrTree, eventName, listener)
					: viewOrTree.events.off("changed", listener);
			},
		};
	}
```

Then just have each implementation use that as needed.

packages/framework/tree-agent/src/agent.ts
```diff
@@ -303,3 +366,147 @@ export function createContext<TSchema extends ImplicitFieldSchema>(
 		key: (child: TreeNode): string | number => Tree.key(child),
 	} satisfies Context<TSchema>;
 }
+
+// #region Factory functions
+
+/**
+ * Shared state for agents created via {@link createTreeAgent}.
+ */
+function createAgentState<TSchema extends ImplicitFieldSchema>(
```
noencke: IMO I would not even bother trying to share code between the old and new implementation, and just write the new implementation to be optimally clean/organized. It will have some duplicate code but oh well - we can kill the old implementation as soon as we switch to the new one in Boards.

packages/framework/tree-agent/src/agent.ts
```diff
@@ -303,3 +366,147 @@ export function createContext<TSchema extends ImplicitFieldSchema>(
...
+
+	const history: TreeAgentChatMessage[] = [{ role: "system", content: prompt }];
+
+	return { outerTree, history, tracker };
+}
+
+/**
+ * Begins a query by logging, refreshing tree state if dirty, and pushing the user message to history.
+ */
+function beginQuery(
```
noencke: Similar to other comment - I think this will be more intuitive as a method on an Agent class, rather than appearing functional but having to pass around a mutable "tracker".

noencke (commented):
I like this approach and I think we should switch to it ASAP! Regarding state management - I think it's good to make the _model_ abstraction stateless. I don't think we should try to make the _agent_ "stateless". It can't actually be stateless, you can either have a class with methods that mutate state, or you can have free functions that are passed mutable state (as is the current approach), and I think a class would be cleaner. Other than that, this is a big improvement IMO.

packages/framework/tree-agent/src/api.ts
steffenloesch: I'm OK with this. I had the different name, because SharedTreeChatModel also has an `invoke` function. It's probably clear enough with the object it's being called on. I was also considering `runEditLoop` or something like that.

packages/framework/tree-agent/src/agent.ts
steffenloesch: Yeah, you're right, the tracker was rather confusing. It came from a desire to make the tree agent implementation state-free, and worked around that. If we want the dirty tracking and extra message if we're dirty between invocation, it's probably better to just switch to state in a class. I've switched to that now.

packages/framework/tree-agent/src/agent.ts
steffenloesch: Good point. A few commits ago I had a second tree agent implementation that doesn't edit, and some of this is still legacy from that. I've now inlined some unnecessary function.

packages/framework/tree-agent/src/agent.ts
steffenloesch: We're now un-listening to the event, but that came at the cost of exposing a `dispose` method on the agent interface.

packages/framework/tree-agent/src/agent.ts
```diff
@@ -303,3 +325,153 @@ export function createContext<TSchema extends ImplicitFieldSchema>(
...
+		this.#offTreeChanged = this.#outerTree.onTreeChanged(() => {
+			this.#isDirty = true;
+		});
+	}
+
+	public dispose(): void {
+		this.#offTreeChanged();
+	}
+
+	public async invoke(prompt: string): Promise<string> {
```
noencke:
```suggestion
	public async message(prompt: string): Promise<string> {
```
Maybe? A little more concrete of a name?

noencke (approved):

packages/framework/tree-agent-langchain/src/chatModel.ts
```diff
@@ -3,29 +3,188 @@
...
 
 /**
- * Creates a `SharedTreeChatModel` that uses the LangChain library to connect to the underlying LLM.
+ * Creates a stateless {@link @fluidframework/tree-agent#SharedTreeChatModel} backed by LangChain.
+ * @remarks Use with {@link @fluidframework/tree-agent#createTreeAgent | createTreeAgent}.
  * @param langchainModel - The LangChain chat model to use.
  * @alpha
  */
 export function createLangchainChatModel(langchainModel: BaseChatModel): SharedTreeChatModel {
 	return new LangchainChatModel(langchainModel);
```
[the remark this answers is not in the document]
steffenloesch: tree-agent-langchain is only used in the repo and does not expose APIs via API Extractor. So this change is OK and we do not consider it breaking.

packages/framework/tree-agent/src/agent.ts
steffenloesch: Works for me. It's changed.

packages/framework/tree-agent/src/agent.ts
```diff
@@ -303,3 +325,153 @@ export function createContext<TSchema extends ImplicitFieldSchema>(
...
+				this.#editor,
+				this.#options?.logger,
+			);
+
+			rollbackEdits = editResult.type !== "success";
+			this.#history.push({
+				role: "tool_result",
+				toolCallId: response.toolCallId,
+				content: editResult.message,
+			});
```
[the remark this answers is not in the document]
steffenloesch: The exception catching to make sure dispose is called is now in place. I don't think the history needs to be re-written in such a situation.

packages/framework/tree-agent-langchain/src/chatModel.ts
```diff
@@ -3,29 +3,188 @@
...
+			}
+			case "tool_call": {
+				messages.push(
+					new AIMessage({
+						content: "",
+						tool_calls: [
+							{
+								id: msg.toolCallId,
+								name: msg.toolName,
+								args: { js: msg.code },
```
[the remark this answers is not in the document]
steffenloesch: Good catch. This is now changed to not be hardcoded, and unit tests are in place to make sure that the argument names are preserved when round-tripping through tree-agent.

packages/framework/tree-agent/src/agent.ts
```diff
@@ -155,15 +170,22 @@ export class SharedTreeSemanticAgent<TSchema extends ImplicitFieldSchema> {
 			return editResult;
 		};
 
+		if (this.client.query === undefined) {
+			throw new UsageError(
+				"The provided SharedTreeChatModel does not implement query(). Use createEditAgent or createAnalysisAgent instead.",
```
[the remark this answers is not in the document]
steffenloesch: Fixed.

packages/framework/tree-agent-langchain/src/test/utils.ts
```diff
@@ -99,6 +103,13 @@ export function createLlmClient(provider: LlmProvider): BaseChatModel {
 	}
 }
 
+/**
+ * The API implementation to use when querying the LLM.
+ * - `"new"`: Uses the new stateless `createLangchainChatModel` + `createEditAgent` + `edit()`.
```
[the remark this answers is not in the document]
steffenloesch: Fixed.

noencke (approved):

noencke (approved):

PR #26662, opened 2026-03-06
Improve `applySerializedChange` to support cross-session usage and transactions

noencke:
This change makes it possible to serialize and apply changes without worrying about conflicting Id compressor IDs, because no compressed IDs are serialized in the change object(s). It also allows serialized changes to be applied and squashed in a transaction, even if the original changes were not made in a transaction.                      
                                                                                                        
- Cross-session support: Removed the session ID check so serialized changes can be applied across different IdCompressor instances (e.g. from a sandbox process to the main runtime). The decode context uses the originator's session ID for correct revision tag decoding.
                                         
- Identifier safety: The encoding path for getChange() already uses uncompressed encoding (no schema passed), so identifier field values are stored as UUID strings rather than compressed integers, making them safe across compressor instances. This was actually previously an oversight, but is now the desired behavior, and so has been made explicit with a comment.

- Revision replacement: Serialized changes now get fresh revision tags when applied, preventing ID collisions in the changeset. Applying the same serialized change twice produces two distinct effects (not deduplicated).

- Transaction support: Serialized changes are now applied via the editor rather than directly to the branch, allowing them to be used within transactions (previously broken). The editor's applyExternalChange method replaces the revision, shifts local IDs to avoid collisions with other changes sharing the same transaction revision, and keeps the editor's ID allocator in sync. This allows serialized changes from different editors to be mixed with direct edits in the same transaction without local ID collisions. DefaultRevisionReplacer now optionally accepts an IdAllocator to coordinate reserved ID ranges.

### DIFF
--- a/packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
+++ b/packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
@@ -81,6 +81,7 @@ import {
 	getFirstFromCrossFieldMap,
 	setInCrossFieldMap,
 } from "./crossFieldQueries.js";
+import { DefaultRevisionReplacer } from "./defaultRevisionReplacer.js";
 import {
 	type FieldChangeHandler,
 	NodeAttachState,
@@ -2686,6 +2687,7 @@ export class ModularEditBuilder extends EditBuilder<ModularChangeset> {
 		private readonly fieldKinds: ReadonlyMap<FieldKindIdentifier, FlexFieldKind>,
 		changeReceiver: (change: TaggedChange<ModularChangeset>) => void,
 		codecOptions: CodecWriteOptions,
+		private readonly mintRevisionTag?: () => RevisionTag,
 	) {
 		super(family, changeReceiver);
 		this.idAllocator = idAllocatorFromMaxId();
@@ -2705,6 +2707,27 @@ export class ModularEditBuilder extends EditBuilder<ModularChangeset> {
 		if (this.transactionDepth === 0) {
 			this.idAllocator = idAllocatorFromMaxId();
 		}
+	}
+
+	/**
+	 * Apply a changeset that originated from a different editor.
+	 * @remarks
+	 * The revision for the change will be changed to a new revision (using `mintRevisionTag`) before application.
+	 * The change will also have its local IDs replaced to avoid collisions with any IDs produced by this editor.
+	 */
+	public applyExternalChange(change: ModularChangeset): void {
+		assert(
+			this.mintRevisionTag !== undefined,
+			"mintRevisionTag is required to apply external changes",
+		);
+		const revision = this.mintRevisionTag();
+		const replacer = new DefaultRevisionReplacer(
+			revision,
+			this.changeFamily.rebaser.getRevisions(change),
+			this.idAllocator,
+		);
+		const newChange = this.changeFamily.rebaser.changeRevision(change, replacer);
+		this.applyChange(tagChange(newChange, revision));
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


packages/dds/tree/src/shared-tree-core/branch.ts
```diff
@@ -124,7 +124,7 @@ export class SharedTreeBranch<TEditor extends ChangeFamilyEditor, TChange> {
 	public constructor(
 		private head: GraphCommit<TChange>,
 		public readonly changeFamily: ChangeFamily<TEditor, TChange>,
-		private readonly mintRevisionTag: () => RevisionTag,
+		public readonly mintRevisionTag: () => RevisionTag,
```
noencke: Rather than exposing this, I originally tried making `branch.apply` automatically change the revisions of changes to a newly minted revision if you didn't give them a revision tag. That is more general and perhaps more elegant. However, this module can't depend on the DefaultRevisionRewriter (or whatever it's called) because it creates a circular dependency, so that approach required some pretty ugly injection patterns and I said nah.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -715,6 +716,7 @@ export class TreeCheckout implements ITreeCheckoutFork {
 							idCompressor: this.idCompressor,
 							originatorId: this.idCompressor.localSessionId,
 							revision,
+							schema: undefined, // By not passing the schema, we avoid compressing identifiers in identifier fields
```
noencke: This doesn't change anything, I'm just adding this comment so now it's clear that it is an intentional omission.

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
```diff
@@ -2707,6 +2709,27 @@ export class ModularEditBuilder extends EditBuilder<ModularChangeset> {
...
+	 * @remarks
+	 * The revision for the change will be changed to a new revision (using `mintRevisionTag`) before application.
+	 * The change will also have its local IDs replaced to avoid collisions with any IDs produced by this editor.
+	 */
+	public applyExternalChange(change: ModularChangeset): void {
+		assert(
+			this.mintRevisionTag !== undefined,
+			"mintRevisionTag is required to apply external changes",
+		);
+		const revision = this.mintRevisionTag();
```
alex-pardes, line 2723: All the other editing methods take in a revision tag instead of minting one. Seems like we should be consistent here. Is it inconvenient to mint a revision at the call site?

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -775,19 +776,24 @@ export class TreeCheckout implements ITreeCheckoutFork {
 			throw new UsageError(`Cannot apply change. Invalid serialized change format.`);
 		}
 		const { revision, originatorId, change } = serializedChange;
-		if (originatorId !== this.idCompressor.localSessionId) {
```
alex-pardes: We still need to require that the local ID compressor knows about the IDs minted with the external ID compressor, right?

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -775,19 +776,24 @@ export class TreeCheckout implements ITreeCheckoutFork {
...
-			throw new UsageError(
-				`Cannot apply change. A serialized changed must be applied to the same SharedTree as it was created from.`,
-			);
-		}
 		const context: ChangeEncodingContext = {
 			idCompressor: this.idCompressor,
-			originatorId: this.idCompressor.localSessionId,
+			originatorId,
 			revision,
```
alex-pardes, line 792: This probably all works, but it's kludgy that we use revision without decoding it.

PR #26668, opened 2026-03-07
Fix memory test name collisions

CraigMacomber:
## Description

These tests would get the exact same fully qualified names for both table sizes, which results in overwritten data and confusing output.

This change disambiguates them by table size.

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/test/memory/tableTree.spec.ts
+++ b/packages/dds/tree/src/test/memory/tableTree.spec.ts
@@ -20,6 +20,7 @@ import {
 	type TableBenchmarkOptions,
 	createTableTree,
 } from "../tablePerformanceTestUtilities.js";
+import { configureBenchmarkHooks } from "../utils.js";
 
 /**
  * Note: These benchmarks are designed to closely match the benchmarks in SharedMatrix.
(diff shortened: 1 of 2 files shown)


packages/dds/tree/src/test/memory/tableTree.spec.ts
```diff
@@ -121,341 +126,575 @@ describe("SharedTree table APIs memory usage", () => {
 	for (const tableSize of tableSizes) {
 		// Filter counts to ensure they do not exceed tableSize
 		const validRemoveCounts = operationCounts.filter((count) => count <= tableSize);
+		describe(`Table size: ${tableSize}`, () => {
```
CraigMacomber, line 129: this line, and a lot of indents and the closing braces is the only change here.

packages/dds/tree/src/test/shared-tree/tableTree.bench.ts
```diff
@@ -107,176 +107,316 @@ describe("SharedTree table APIs execution time", () => {
 	let maxBenchmarkDurationSeconds: number;
 
 	for (const tableSize of tableSizes) {
-		maxBenchmarkDurationSeconds = tableSize === 50 ? 10 : 5;
-
-		// Filter counts to ensure remove operation do not exceed tableSize
-		const validRemoveCounts = operationCounts.filter((count) => count <= tableSize);
+		describe(`Table size: ${tableSize}`, () => {
```
CraigMacomber, line 110: this line, and a lot of indents and the closing braces is the only change here.

Abe27342 (approved):

PR #26669, opened 2026-03-07
test(client-import-testing): relocate typescript versions to host package

jason-ha:
This restores the typescript version used by import-testing build to 5.4.5 that is broadly used.

When the aliased packages are installed, the last one will set the bin for tsc (and tsserver), which was regularly observed to be 5.9.


WillieHabi (approved):
Makes sense! Interested in understanding the potential process.cwd() problem callout -- is this because if we invoke these tests from somewhere else other than import-testing directory then we'd have problems resolving the ts version host package? 

I see this code already did `fs.readFileSync("package.json");` (which looks like a similar bet?) so I guess this isn't a new thing being introduced

jason-ha:
> Makes sense! Interested in understanding the potential process.cwd() problem callout -- is this because if we invoke these tests from somewhere else other than import-testing directory then we'd have problems resolving the ts version host package?
> 
> I see this code already did `fs.readFileSync("package.json");` (which looks like a similar bet?) so I guess this isn't a new thing being introduced

This is a work around to dual-emitting CommonJS and ESM. We want to just pass `import.meta.url` to `createRequire`, but that is not supported for CommonJS. CommonJS could do other things that aren't supported in ESM. 
Since we expect to be executed as a test from local mocha, I think we are okay with `process.cwd()`. (Credit to AI for this solution that doesn't require having a CJS module in here.)

PR #26676, opened 2026-03-09
resolve dom on import issue by moving quill components into their own package

brrichards:
# Summary

Quill in the react package was polluting all users of that package with a need to have a dom on import time. Now the quill specific imports that caused this have been moved to the new package at `/packages/framework/quill-react`. 

Moved `quillFormattedView.tsx`, `quillView.tsx`, and `mochaHooks` to new package. Left `plainTextView.tsx` in the react package as it did not contain any quill components. Separated the plain text tests from quill view tests when moving them over to the new package. 

Removed the inventory-app JSDOM fix needed when importing `@fluidframework/react/internal` when it still had quill components.

Created `.changeset` with reasoning for why the package was created.

#### Breaking Change

Anyone importing quill components from `@fluidframework/react/internal` will now have to import from `@fluidframework/react/quill-react`. Since this is still internal and only used in text-editor, which has been updated with new path, this should be fine.


Josmithr:
Adding custom entrypoints like this is generally something we should avoid whenever possible IMO. Our standard entrypoints map to specific stability guarantees we make for our users, which we lose with a pattern like this.

I'm not completely up-to-speed on this issue - what is the failure mode users are seeing? What other options have we explored to resolve this?

.changeset/five-loops-sip.md
```diff
@@ -0,0 +1,10 @@
+---
+"@fluidframework/react": minor
+"@fluidframework/quill-react": minor
+"__section": other
+---
+New package created containing all examples for integrating Quill with Fluid Framework into React applications.
+
+Applications utilizing quill require DOM access at import time. This package contains all integrations of Fluid Framework with Quill/React. This package should only be imported in browser environments or test environments with JSDOM set up before import.
```
Josmithr: Nit: 
```suggestion
Applications utilizing Quill require DOM access at import time. This package contains all integrations of Fluid Framework with Quill/React. This package should only be imported in browser environments or test environments with JSDOM set up before import.
```

packages/framework/quill-react/src/plain/index.ts
```diff
@@ -0,0 +1,6 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export { MainView as QuillMainView, type MainViewProps } from "./quillView.js";
```
Josmithr: Should we similarly alias the props type to match the component?

packages/framework/quill-react/package.json
```diff
@@ -0,0 +1,108 @@
...
+	"sideEffects": false,
+	"type": "module",
+	"exports": {
+		".": {
+			"import": {
+				"types": "./lib/index.d.ts",
+				"default": "./lib/index.js"
+			}
+		}
+	},
```
Josmithr, line 40: It doesn't look like this package is set up with our API-Extractor infra. We should add that before merging and make sure we're generating the appropriate public/beta/alpha rollups. I can have context with this stuff if you have questions.

packages/framework/quill-react/README.md
```diff
@@ -0,0 +1,9 @@
+# @fluidframework/quill-react
```
Josmithr, line 1: I'd recommend adding our library package README content generation pragmas below your existing content here. You can look at one of the other package READMEs in the repo for reference. Will look something like:
```md
<!-- AUTO-GENERATED-CONTENT:START (LIBRARY_README_HEADER) -->
...
<!-- AUTO-GENERATED-CONTENT:END -->

<!-- AUTO-GENERATED-CONTENT:START (README_FOOTER) -->
...
<!-- AUTO-GENERATED-CONTENT:END -->
```

Then, running `npm run build:readme` will auto-generate a bunch of user-facing docs boilerplate.

Josmithr (commented):
Left a couple of suggestions, but I think the direction here looks good. Let me know if you have any questions / if I can help with any of the infra stuff 🙂

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -7,6 +7,7 @@ import type { TextAsTree } from "@fluidframework/tree/internal";
 
 /**
  * Sync `newText` into the provided `root` tree.
+ * @input @internal
```
alexvy86: Nit: I think `@input` would apply to a type but not really to a function.
```suggestion
 * @internal
```

.changeset/five-loops-sip.md
alexvy86, line 1: I'd argue that this change might not need a changeset... we as Fluid devs need to know how to use the new package in our examples, so having notes like these in its README would be good, but arguably we should not be encouraging any Fluid consumers to ever import from this package?

examples/data-objects/inventory-app/src/test/mochaHooks.ts
alexvy86, line 1: I thought inventory-app did use quill as a transitive dependency and thus needed this?

packages/framework/quill-react/package.json
brrichards: Updated with API-extractor infra

examples/data-objects/inventory-app/src/test/mochaHooks.ts
brrichards: Without these changes inventory-app was pulling in quill transitively through the react package, which re-exported the quill components. The example itself didn't use quill, so now it no longer has the dependency and I believe the JSDOM setup is no longer needed.

.changeset/five-loops-sip.md
brrichards: Yeah that makes sense to move this into the readme and get rid of the changeset. The main goal of this package is to separate quill from the react package so we don't have the JSDOM setup needed everytime the react package is imported even when quill is not needed. Not for Fluid consumers

Josmithr: Once we promote the APIs to alpha/beta we'll want a changeset. But until then, I agree.

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -8,7 +8,8 @@ import { createDevtoolsLogger, initializeDevtools } from "@fluidframework/devtoo
 import {
 	FormattedMainView,
 	QuillMainView as PlainQuillView,
-} from "@fluidframework/quill-react";
+	// eslint-disable-next-line import-x/no-internal-modules
```
Josmithr, line 14: We should probably leave a TODO comment explaining that while we're still prototyping / in the initial-development phase, we need access to the internal APIs here. Ideally, we'll update these imports to use more and more public exports as we promote the underlying APIs.

Josmithr (approved):
Looks great! 🚀

examples/data-objects/text-editor/src/app.tsx
brrichards: Good idea. Added the TODO comment

Josmithr (approved):

PR #26677, opened 2026-03-09
Quill.js style delta payload for arrayNode events

daesunp:
## Description

Adds quill.js delta event payload for arraynodes.


packages/dds/tree/src/core/tree/visitDelta.ts
```diff
@@ -372,6 +392,9 @@ function visitFieldMarks(
 	if (fields !== undefined) {
 		for (const [key, field] of fields) {
 			visitor.enterField(key);
+			if (config.isFirstPass) {
+				visitor.fieldMarks?.(key, field.marks);
```
alex-pardes: Instead of adding the `isFirstPass` flag, can we just put this line into the beginning of the `attachPass` function?

alex-pardes: I meant `detachPass`.

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -234,6 +269,43 @@ export const treeNodeApi: TreeNodeApi = {
...
+ * It is retained in case future encoder changes produce such marks.
+ */
+function deltaMarksToArrayOps(marks: readonly DeltaMark[]): ArrayNodeDeltaOp[] {
+	const ops: ArrayNodeDeltaOp[] = [];
+	for (const mark of marks) {
+		if (mark.attach !== undefined && mark.detach !== undefined) {
+			// Replacement: content removed and new content attached in the same position.
+			// The `Delta.Mark` format allows both to be set simultaneously (e.g. when a slot's
+			// content is atomically swapped), even though the sequence-field encoder does not
+			// currently emit such marks for array (EmptyKey) fields.  Handle it defensively.
```
alex-pardes: The terminology here doesn't make sense to me. It seems sufficient to say that this case should be unreachable (because sequence field never produces a replace delta mark), but is handled anyway.

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -234,6 +269,43 @@ export const treeNodeApi: TreeNodeApi = {
...
+			ops.push({ type: "remove", count: mark.count });
+			ops.push({ type: "insert", count: mark.count });
+		} else if (mark.attach !== undefined) {
+			ops.push({ type: "insert", count: mark.count });
+		} else if (mark.detach === undefined) {
+			// Neither attach nor detach: elements retained (may have nested changes in mark.fields).
+			ops.push({ type: "retain", count: mark.count });
+		} else {
+			ops.push({ type: "remove", count: mark.count });
+		}
```
alex-pardes: It looks like this could just have one block for `mark.detach !== undefined`, followed by a block for `mark.attach !== undefined`.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -441,16 +461,21 @@ class KernelEventBuffer implements Listenable<KernelEvents> {
 	}
 
 	#emit(
-		eventName: keyof KernelEvents,
-		arg?: {
-			changedFields: ReadonlySet<FieldKey>;
-		},
+		...args:
```
alex-pardes: What is the purpose of this change?

packages/dds/tree/src/core/tree/visitDelta.ts
daesunp: removed `isFirstPass` and added `key` in `Pass` function so `detachPass` already has key to pass into `visitor.fieldMarks?.(key, fieldChanges.marks)` directly (and `attachPass` ignores it)

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: simplified comment and made more clear

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: refactored to suggested change

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
daesunp: was originally to enforce `childrenChangedAfterBatch` to have args, and `subtreeChangedAfterBatch` to not require it, but reverted change and used an assert instead

packages/dds/tree/src/core/tree/anchorSet.ts
```diff
@@ -726,6 +736,12 @@ export class AnchorSet implements AnchorLocator {
 			 */
 			bufferedEvents: [] as BufferedEvent[],
 
+			/**
+			 * Field marks delivered via {@link DeltaVisitor.fieldMarks}, keyed by (PathNode, FieldKey).
+			 * Populated during the first (detach) pass and consumed during "free" when emitting events.
```
Josmithr: Nit (since this part describes how this is used as opposed to what it _is:)_
```suggestion
			 * @remarks Populated during the first (detach) pass and consumed during "free" when emitting events.
```

packages/dds/tree/src/simple-tree/api/treeBeta.ts
```diff
@@ -64,6 +65,21 @@ export interface NodeChangedData<TNode extends TreeNode = TreeNode> {
...
+	 * When present, the value may still be `undefined` in two cases:
+	 * - The array node is {@link Unhydrated} — unhydrated nodes are not visited by the delta
+	 * pipeline, so no field marks are available.
+	 * - The array was modified across multiple batches within a single flush (e.g. due to an
+	 * interleaved schema change) and the marks from those batches could not be composed.
+	 *
+	 * See {@link ArrayNodeDeltaOp} for op semantics.
+	 */
+	readonly delta?: readonly ArrayNodeDeltaOp[];
 }
```
Josmithr, line 67: Since the `delta` and `changedProperties` properties are (presumably) mutually exclusive, we can make things a bit more type-safe by splitting this into multiple types. E.g.

```typescript
export type NodeChangedData<TNode extends TreeNode = TreeNode> = {
	readonly changedProperties: ReadonlySet<...>;
} |  {
	readonly delta: readonly ArrayNodeDeltaOp[];
};
```

The interior types could be extracted into their own definitions too, if useful. Might clean up some of the type declarations used in the code below.

I'm assuming we now always return an object with either `delta` or `changedProperties`?

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -39,6 +39,32 @@ import { isArrayNodeSchema, isObjectNodeSchema } from "../node-kinds/index.js";
 
 import type { TreeChangeEvents } from "./treeChangeEvents.js";
 
+/**
+ * A single operation in an array node change delta, used to efficiently sync an external
```
Josmithr: Nit
```suggestion
 * A single operation in an array node change delta. Used to efficiently sync an external
```

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -39,6 +39,32 @@ import { isArrayNodeSchema, isObjectNodeSchema } from "../node-kinds/index.js";
...
+ * representation of an array (e.g. a text editor or virtual list) with tree changes, without
```
Josmithr: Nit
```suggestion
 * representation of an array (e.g. a text editor or virtual list) with tree changes without
```

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -39,6 +39,32 @@ import { isArrayNodeSchema, isObjectNodeSchema } from "../node-kinds/index.js";
...
+ * needing to snapshot the old state or diff the entire array. Each op describes a contiguous run
+ * of positions in the array before the change; for inserts, read the new element values from the
```
Josmithr: Nit
```suggestion
 * of positions in the array before the change. For inserts, read the new element values from the
```

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -39,6 +39,32 @@ import { isArrayNodeSchema, isObjectNodeSchema } from "../node-kinds/index.js";
...
+ * current tree at those positions.
+ *
+ * @remarks
+ * - `"retain"` — elements that were not added or removed (may have nested changes).
+ * - `"insert"` — elements added to the array.
+ * - `"remove"` — elements removed from the array.
+ *
+ * Moves are represented as remove + insert.
+ * There is no dedicated `"move"` op. When an element is moved within the same array it appears
```
Josmithr: Nit: I would probably swap the order of these first two sentences. I think it flows a bit better that way, personally.
> There is no dedicated "move" op. Moves are represented as "remove" + "insert".

Josmithr (commented):
Left a couple suggestions on the user-facing side of things, but overall the direction is looking good to me.

This PR should definitely also have a changeset describing the changes and guiding users through how to leverage the new functionality.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -362,6 +363,20 @@ class KernelEventBuffer implements Listenable<KernelEvents> {
...
+	 */
+	readonly #fieldMarksBuffer: Map<FieldKey, readonly DeltaMark[]> = new Map();
+
+	/**
+	 * Fields whose marks have been permanently invalidated within the current buffer window due to
+	 * two or more separate delta batches touching the same field.
+	 * Once a key is in this set it must never be re-added to the marks buffer, even if
+	 * a third (or later) batch arrives for that field.
+	 */
+	readonly #invalidatedFieldMarkKeys: Set<FieldKey> = new Set();
```
Josmithr, line 378: For my understanding, why is this needed?

packages/dds/tree/src/simple-tree/core/unhydratedFlexTree.ts
```diff
@@ -271,7 +271,11 @@ export class UnhydratedFlexTreeNode
 	}
 
 	public emitChangedEvent(key: FieldKey): void {
-		this._events.emit("childrenChangedAfterBatch", { changedFields: new Set([key]) });
+		this._events.emit("childrenChangedAfterBatch", {
+			changedFields: new Set([key]),
+			// Unhydrated nodes are not visited by the delta pipeline, so no field marks are available.
```
Josmithr, line 276: Is this something we plan to add support for in a future PR? Generally, I think we consider it a bug for unhydrated nodes to behave differently from hydrated ones.

Josmithr: If this is a temporary state, it would be good to call this out in the user-facing docs on `NodeChangedData`.

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -39,6 +39,32 @@ import { isArrayNodeSchema, isObjectNodeSchema } from "../node-kinds/index.js";
...
+ * as a `"remove"` at the source position followed by an `"insert"` at the destination position.
+ * When an element is moved across two different arrays, the source array's delta contains a
+ * `"remove"` and the destination array's delta contains an `"insert"` — they cannot be
+ * correlated without additional bookkeeping on the caller's side.
+ *
+ * @sealed @beta
+ */
+export type ArrayNodeDeltaOp =
```
Josmithr: I would recommend extracting these inner types out into their own definitions. That way we can document each in a more targeted way (rather than having the `@remarks` items for retain, insert, remove).

packages/dds/tree/src/core/tree/anchorSet.ts
daesunp: Updated with suggestion

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: Updated with suggestion

packages/dds/tree/src/simple-tree/api/treeBeta.ts
daesunp: Updated `NodeChangedData` to be a union of `NodeChangedDataProperties` and `NodeChangedDataDelta`, and simplified `nodeChanged` signature

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: Updated with suggestion

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: Updated with suggestion

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: Updated with suggestion

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
daesunp: Unfortunately we don't have a good way to compose deltas when we have multiple delta batches for the same field (due to the positions from the first batch's marks potentially no longer being valid relative to the second batch), so instead of providing an incorrect delta, we mark that field as invalidated for that buffer window and return an undefined delta. 

But this will only be an issue if withBufferedTreeEvents is used (which isn't user facing yet)

packages/dds/tree/src/simple-tree/core/unhydratedFlexTree.ts
daesunp: Updated the user facing doc mentioning that this will be addressed in a future release :)

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: Extracted inner types out into their own definitions

.changeset/goofy-keys-wait.md
```diff
@@ -0,0 +1,66 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": feature
+---
+
+Array node `nodeChanged` events now include a delta payload describing what changed
+
+The `nodeChanged` event for array nodes now provides a `delta` field — a sequence of `ArrayNodeDeltaOp` values that describe exactly what changed in the array. This lets you efficiently sync an external representation with tree changes, without snapshotting the old state or diffing the entire array.
```
Josmithr: This is only exposed via the `TreeBeta` or `TreeAlpha` API, right? We should probably be explicit about this in the title and description here.

Also, if we're currently adding this to the `TreeBeta` interface, we probably want to consider putting it on `TreeAlpha` instead (for now), so we can evaluate it before committing to longer-term support for our partners.

.changeset/goofy-keys-wait.md
```diff
@@ -0,0 +1,66 @@
...
+
+The delta follows Quill-style semantics: each op covers a contiguous run of positions in the array before the change.
+
+- `{ type: "retain", count: N }` — N elements were unchanged (and may have nested changes).
```
Josmithr: What does the "and may have nested changes" note mean? Is this indicating that the array change events don't track subtree changes within its elements?

.changeset/goofy-keys-wait.md
```diff
@@ -0,0 +1,66 @@
...
+The `NodeChangedData` type is now a discriminated union:
+- Object, map, and record nodes emit `NodeChangedDataProperties` (with a required `changedProperties` set).
+- Array nodes emit `NodeChangedDataDelta` (with a `delta` field).
+
+### Example: applying a delta to a plain array mirror
+
+```typescript
+const mirror: number[] = [1, 2, 3];
+
+TreeBeta.on(myArrayNode, "nodeChanged", (data) => {
```
Josmithr: Let's add a quick comment above this handler explaining what it's doing

.changeset/goofy-keys-wait.md
```diff
@@ -0,0 +1,66 @@
...
+	} else {
+		// Object/map/record node — data is NodeChangedDataProperties
+		console.log("properties changed:", data.changedProperties);
+	}
+});
+```
+
+> **Note:** The `delta` value may be `undefined` in two cases:
+> - The array node is unhydrated (a known temporary limitation).
+> - The array was modified across multiple batches within a single flush (e.g. due to an interleaved schema change).
```
Josmithr: Question: is this terminology we expect users of SharedTree to understand? I _think_ these are internal implementation nomenclature. It would be good to rephrase this in terms of user-facing behaviors that might cause this. (Please disregard if I am misinformed here 🙂)

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -414,7 +441,7 @@ export namespace FluidSerializableAsTree {
     export type _RecursiveArrayWorkaroundJsonArray = FixRecursiveArraySchema<typeof Array>;
     const // @system
     _APIExtractorWorkaroundArrayBase: TreeNodeSchemaClass_2<"com.fluidframework.serializable.array", NodeKind_2.Array, System_Unsafe_2.TreeArrayNodeUnsafe<readonly [() => typeof FluidSerializableObject, () => typeof Array, LeafSchema_2<"string", string>, LeafSchema_2<"number", number>, LeafSchema_2<"boolean", boolean>, LeafSchema_2<"null", null>, LeafSchema_2<"handle", IFluidHandle<unknown>>]> & WithType_2<"com.fluidframework.serializable.array", NodeKind_2.Array, unknown>, {
-    [Symbol.iterator](): Iterator<string | number | IFluidHandle<unknown> | System_Unsafe_2.InsertableTypedNodeUnsafe<LeafSchema_2<"boolean", boolean>, LeafSchema_2<"boolean", boolean>> | FluidSerializableObject | Array | null, any, undefined>;
+    [Symbol.iterator](): Iterator<string | number | IFluidHandle<unknown> | FluidSerializableObject | Array | System_Unsafe_2.InsertableTypedNodeUnsafe<LeafSchema_2<"boolean", boolean>, LeafSchema_2<"boolean", boolean>> | null, any, undefined>;
```
Josmithr: You have some unrelated changes here you'll need to revert by regenerating these files with a clean build of the package.

packages/dds/tree/src/simple-tree/api/treeBeta.ts
Josmithr: Looks great! Thanks!!

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
Josmithr: Sorry, I think I'm missing context. Can you elaborate on the scenarios that cause us problems? I don't know a lot about the underlying delta tracking logic.

What would this mean for users in the future if we exposed event batching options to users?

packages/dds/tree/src/simple-tree/core/unhydratedFlexTree.ts
Josmithr: Thanks!

packages/dds/tree/src/test/simple-tree/api/treeNodeApi.spec.ts
```diff
@@ -2720,7 +2735,133 @@ describe("treeNodeApi", () => {
 
 			view.checkout.merge(forkCheckout);
 
-			assert.deepEqual(eventLog, [undefined]);
+			assert.equal(eventCount, 1);
+		});
+
+		describe(`'nodeChanged' delta payload for array operations`, () => {
```
Josmithr, line 2742: We should probably leave a TODO comment here to use `describeWithHydration` once we've added delta event support for unhydrated nodes.

Josmithr: If we haven't filed a follow up work item for this yet, we probably should. Then, I would recommend updating our associated TODO comments to reference it. Will make it easier to discover / audit the places that need to be updated in the future.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
daesunp: Sorry I think I explained it poorly!

The delta is built from field marks which has the position information in a changed array. Each time a batch of changes is applied to the tree, the kernel stores the marks for that batch keyed by the field.

Unfortunately if we use withBufferedTreeEvents, and buffer two separate batches that both touch the same array field (before nodeChanged fires), each batch's field marks describe positions relative to the array state at the start of that batch. After batch 1 is applied, and the array has shifted, batch 2's marks may no longer have the correct positions.

Example:
Array = [A, B, C]

Batch1 removes index 0 -> [B,C]
marks: {remove at position 0} (relative to [A,B,C])

Batch2 inserts D at index 0 -> [D, B, C]
marks: {insert at position 0}  (relative to [B,C]

And we don't have a good way to compose these two deltas yet. If users get event batching in the future, they will get undefined in that scenario, which was what the eventing rewrite was aiming to fix (which may be on pause right now).

This feature is definitely not the most complete, but aims to give useful information some of the time (which can be a small win for things like text edit)

.changeset/goofy-keys-wait.md
daesunp: Updated to be in TreeAlpha and updated changeset

.changeset/goofy-keys-wait.md
daesunp: Updated comment to be more clear

.changeset/goofy-keys-wait.md
daesunp: Added comment to explain what it's doing

.changeset/goofy-keys-wait.md
daesunp: Updated with user-facing terminology :)

packages/dds/tree/src/test/simple-tree/api/treeNodeApi.spec.ts
daesunp: Filed a follow up work item and updated TODO comments to reference it

packages/dds/tree/api-report/tree.alpha.api.md
daesunp: regenerated doc files

.changeset/goofy-keys-wait.md
```diff
@@ -0,0 +1,71 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": feature
+---
+
+Array node `nodeChanged` events now include a delta payload (via `TreeAlpha`)
```
Josmithr: Our changeset guidelines call for no `code formatting` in changeset titles (to keep them looking clean and uniform).
```suggestion
Array node nodeChanged events now include a delta payload (via TreeAlpha)
```

.changeset/goofy-keys-wait.md
```diff
@@ -0,0 +1,71 @@
...
+
+The `nodeChanged` event for array nodes (accessed via `TreeAlpha.on`) now provides a `delta` field, a sequence of `ArrayNodeDeltaOp` values that describe exactly what changed in the array. This lets you efficiently sync an external representation with tree changes, without taking a snapshot of the old state or diffing the entire array.
+
+The delta follows Quill-style semantics: each op covers a contiguous run of positions in the array before the change.
```
Josmithr: Since this is the first place we reference Quill.js, let's link to their docs here for reference.
```suggestion
The delta follows [Quill](https://quilljs.com/docs/)-style semantics: each op covers a contiguous run of positions in the array before the change.
```

.changeset/goofy-keys-wait.md
```diff
@@ -0,0 +1,71 @@
...
+
+Trailing unchanged elements are not represented by a trailing `"retain"` op.
+
+Use `TreeAlpha.on` to subscribe to the richer alpha events. The data passed to the callback is typed as `NodeChangedDataAlpha<TNode>`:
+- Object, map, and record nodes receive `NodeChangedDataProperties` (with a required `changedProperties` set).
+- Array nodes receive `NodeChangedDataDelta` (with a `delta` field).
+
+`TreeBeta.on` is unchanged and does not include delta information.
+
+### Example: Applying a Delta to a Plain Array Mirror
```
Josmithr: In our changesets, sub-headings must be level 4 or higher (the tooling the generates our CHANGELOGs reserves the first 3 levels.
```suggestion
#### Example: Applying a Delta to a Plain Array Mirror
```

.changeset/goofy-keys-wait.md
```diff
@@ -0,0 +1,71 @@
...
+			writePos += op.count;
+			readPos += op.count;
+		} else if (op.type === "remove") {
+			mirror.splice(writePos, op.count);
+		}
+	}
+});
+```
+
+### Example: Narrowing the Union in a Generic Handler
```
Josmithr:
```suggestion
#### Example: Narrowing the Union in a Generic Handler
```

.changeset/goofy-keys-wait.md
```diff
@@ -0,0 +1,71 @@
...
+- Object, map, and record nodes receive `NodeChangedDataProperties` (with a required `changedProperties` set).
+- Array nodes receive `NodeChangedDataDelta` (with a `delta` field).
+
+`TreeBeta.on` is unchanged and does not include delta information.
+
+### Example: Applying a Delta to a Plain Array Mirror
+
+```typescript
+// Walk the delta to keep a plain JS array in sync with an array node.
+// retain = advance past unchanged elements, insert = splice in new elements,
```
Josmithr: Nit (for formatting consistency)
```suggestion
// retain = advance past unchanged elements,
// insert = splice in new elements,
```

.changeset/goofy-keys-wait.md
```diff
@@ -0,0 +1,71 @@
...
+	} else {
+		// Object/map/record node — data is NodeChangedDataProperties
+		console.log("properties changed:", data.changedProperties);
+	}
+});
+```
+
+> **Note:** The `delta` value may be `undefined` in two cases:
+> - The node was created locally and has not yet been inserted into a document tree (a known temporary limitation, tracked in AB#63261).
+> - The document was updated in a way that required multiple internal change passes in a single operation (for example, a data change combined with a schema upgrade).
```
Josmithr, line 72: Thanks for updating this note! This makes sense to me as someone mostly familiar with our public-facing API/layers 🙂

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -0,0 +1,88 @@
...
+			: NodeChangedDataProperties<TNode> | NodeChangedDataDelta;
+
+/**
+ * Extensions to {@link TreeChangeEventsBeta} which are not yet stable.
+ * @remarks
+ * Provides a richer `nodeChanged` event that includes a delta payload for array nodes.
+ * @sealed @alpha
+ */
+export interface TreeChangeEventsAlpha<TNode extends TreeNode = TreeNode>
+	extends TreeChangeEvents {
```
Josmithr, line 80: The comments note that this extends `TreeChangeEventsBeta`, but the code here doesn't match. Does this need to be updated?

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
Josmithr: I see. That makes sense. Thanks for the explanation! Would you mind updating the documentation here to provide more context and note that this is a temporary limitation we intend to address in the future?

.changeset/goofy-keys-wait.md
daesunp: Updated with suggested change

.changeset/goofy-keys-wait.md
daesunp: Updated with suggested change

.changeset/goofy-keys-wait.md
daesunp: Updated with suggested change

.changeset/goofy-keys-wait.md
daesunp: Updated with suggested change

.changeset/goofy-keys-wait.md
daesunp: Updated with suggested change

.changeset/goofy-keys-wait.md
daesunp: No problem! Thank you Josh :)

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: Removed incorrect comment and updated docs

Josmithr (approved):
User-facing aspects / simple-tree layer changes look good to me! But I'll defer to @alex-pardes on the lower-level changes since he's been reviewing those.

packages/dds/tree/src/core/tree/visitDelta.ts
```diff
@@ -317,6 +317,18 @@ export interface DeltaVisitor {
...
+	 * Optional hook for obtaining the complete ordered change description for a field, suitable
+	 * for producing positional deltas for external representations (e.g. a text editor or virtual
+	 * list) without needing to diff the old and new states.
+	 * @remarks
+	 * Called once per field per delta, during the first (detach) pass only, after `enterField` but
+	 * before any `attach`, `detach`, or `enterNode` calls for the field. Firing only during the
+	 * detach pass ensures marks are always relative to the original array positions rather than an
+	 * intermediate partially-transformed state.
+	 */
+	fieldMarks?(key: FieldKey, marks: readonly Delta.Mark[]): void;
```
alex-pardes: Unless it's inconvenient for the visitor implementations, I think it would make more sense to remove the `key` parameter, and just call this after `enterField`. The visitors probably already have the field key.

daesunp: Updated with suggestion

packages/dds/tree/src/core/tree/anchorSet.ts
```diff
@@ -934,6 +971,17 @@ export class AnchorSet implements AnchorLocator {
 			exitField(key: FieldKey): void {
 				this.parentField = undefined;
 			},
+			fieldMarks(marks: readonly Delta.Mark[]): void {
+				if (this.parent !== undefined && this.parentField !== undefined) {
```
alex-pardes: We could just assert that `this.parentField !== undefined`, as this shouldn't be called unless we're in a field.

packages/dds/tree/src/core/tree/visitDelta.ts
```diff
@@ -361,7 +373,12 @@ interface PassConfig {
 	readonly rootDestructions: Delta.DetachedNodeDestruction[];
 }
 
-type Pass = (delta: Delta.FieldChanges, visitor: DeltaVisitor, config: PassConfig) => void;
+type Pass = (
+	key: FieldKey,
```
alex-pardes: We should remove the `key` parameter since it isn't used.

packages/dds/tree/src/core/tree/anchorSet.ts
daesunp: Updated with suggested change

alex-pardes (approved):

packages/dds/tree/src/core/tree/visitDelta.ts
daesunp: Updated with suggested change

alex-pardes (approved):

PR #26680, opened 2026-03-09
build(client-test-utils): set mocha as devDep

jason-ha:
test-utils does not have a runtime dependency on mocha. Only types are imported for mocha and those were already set as a devDependency.

Add notes about the test timeout conditional side-effect of loading that is present in test-utils package.

Additionally refactor event and error logger away from testObjectProvider.ts to further isolate modules that depend on the side-effect related code. Nothing strictly relies on the side-effect, but many components reference the module with the side-effect. Per the barrel pattern, any package load will enact the side-effect at this time (assuming no tree shaking).

test-utils is not uniformly used throughout the client workspace which makes it non-trivial to relocate the test timeout logic to @fluid-internal/mocha-test-setup that is broadly used.

### DIFF
--- a/packages/test/test-utils/src/index.ts
+++ b/packages/test/test-utils/src/index.ts
@@ -3,8 +3,12 @@
  * Licensed under the MIT License.
  */
 
+export type { IEventAndErrorTrackingLogger } from "./eventAndErrorLogger.js";
+export {
+	EventAndErrorTrackingLogger,
+	getUnexpectedLogErrorException,
+} from "./eventAndErrorLogger.js";
 export { IProvideTestFluidObject, ITestFluidObject } from "./interfaces.js";
-export { LoaderContainerTracker } from "./loaderContainerTracker.js";
 export {
 	fluidEntryPoint,
 	LocalCodeLoader,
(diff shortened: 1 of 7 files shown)


tylerbutler (approved):
LGTM.

packages/test/test-utils/src/index.ts
```diff
@@ -30,25 +34,30 @@ export {
...
 	TestDataObjectKind,
 } from "./testFluidObject.js";
-export {
-	createDocumentId,
-	DataObjectFactoryType,
-	EventAndErrorTrackingLogger,
-	type IEventAndErrorTrackingLogger,
-	getUnexpectedLogErrorException,
+
+// #region Exports with load side-effect
```
alexvy86, line 38: Is the region correct? Importin from the type exports in here can have load side-effects?

jason-ha: Well, yes and no. You are correct that the `import type` statements won't cause the problem. But I thought we should keep them together with their runtime counterparts.

jason-ha: Comment cleanup: #26682

PR #26681, opened 2026-03-09
Remove browserLogin and downstream dependencies of it from OdspTokenManager

Abe27342:
## Description

Removes code relating to interactive browser login from OdspTokenManager. This code has likely not worked for a long time and adds additional complexity to various test/demo-only flows.

This is a precursor to work toward adopting a FIC solution from AuthN/Z team for test scenarios that has better short-lifespan properties for certain test accounts.


packages/utils/tool-utils/src/odspTokenManager.ts
```diff
@@ -45,17 +38,11 @@ export const getMicrosoftConfiguration = (): IPublicClientConfig => ({
...
-	| {
-			type: "browserLogin";
-			navigator: (url: string) => void;
-			redirectUriCallback?: (tokens: IOdspTokens) => Promise<string>;
-	  };
+export interface OdspTokenConfig {
+	type: "password";
+	username: string;
+	password: string;
+}
```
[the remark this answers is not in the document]
Abe27342: internal APIs do not need changesets and should not be used by consumers.

Internal API breaks only need to comply with our own layer compat guarantees. Since this is within tool-utils, there shouldn't be any concerns with just breaking it immediately.

anthony-murphy (approved):

packages/utils/tool-utils/src/odspTokenManager.ts
```diff
@@ -157,12 +144,10 @@ export class OdspTokenManager {
 	private static getCacheKey(
 		isPush: boolean,
 		tokenConfig: OdspTokenConfig,
-		server: string,
 	): IOdspTokenManagerCacheKey {
-		// If we are using password, we should cache the token per user instead of per server
 		return {
 			isPush,
-			userOrServer: tokenConfig.type === "password" ? tokenConfig.username : server,
+			userOrServer: tokenConfig.username,
```
ChumpChief: Optional nit, seems like this can be simplified to `user`.  Probably `IOdspTokenManagerCacheKey` doesn't need to be exported either.

packages/utils/tool-utils/src/odspTokenManager.ts
```diff
@@ -45,17 +38,11 @@ export const getMicrosoftConfiguration = (): IPublicClientConfig => ({
...
-			username: string;
-			password: string;
-	  }
-	| {
-			type: "browserLogin";
-			navigator: (url: string) => void;
-			redirectUriCallback?: (tokens: IOdspTokens) => Promise<string>;
-	  };
+export interface OdspTokenConfig {
+	type: "password";
```
ChumpChief, line 42: Optional nit, consider removing `type` field as it seems useless now.

packages/utils/tool-utils/src/odspTokenManager.ts
```diff
@@ -239,36 +223,16 @@ export class OdspTokenManager {
 		}
 
 		if (tokens) {
```
ChumpChief: Optional nit, this conditional return probably makes more sense if moved into the tail of the `if (!forceReauth)` block

ChumpChief (approved):
Some cleanup/stylistic nits if you like but looks good.

packages/utils/tool-utils/src/odspTokenManager.ts
Abe27342: Keeping it around as follow-up to this will involve a different shape for this object (rather than username+password, it has username+prefetched token) and keeping the type distinct allows avoiding things like sniffing the password field for eyJ.

I was thinking of renaming this to `LoginCredentials` or something to that effect. There are similarly named interfaces that kind of better fit this sort of name in OdspTestDriver. But will probably do in that follow-up if we want it.

packages/utils/tool-utils/src/odspTokenManager.ts
Abe27342: good call on the rename.

Don't want to bundle avoiding export into this PR. It's used in the constructor of OdspTokenManager and therefore pops up in a few places.

ChumpChief (approved):

jason-ha:
@Abe27342, @CraigMacomber, @alexvy86, anyone have guesses why this doesn't break CI?
The lockfile wasn't updated - it still has a reference to package that doesn't exist with this change.

alexvy86:
> @Abe27342, @CraigMacomber, @alexvy86, anyone have guesses why this doesn't break CI? The lockfile wasn't updated - it still has a reference to package that doesn't exist with this change.

Two reasons I can think why this should fail: 1) the "extraneous modified files" check after having done `pnpm i`, which I suspect doesn't fail because of the ["reset lockfile" step](https://dev.azure.com/fluidframework/public/_build/results?buildId=383390&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=28583298-6979-5448-745e-a78d643df4d6). 2) `pnpm i --frozen-lockfile` not liking the fact the lockfile still has those references... but it seems like it doesn't care? I removed the whole `examples/` folder locally and can still run `pnpm i` successfully. Maybe workspace packages aren't checked like external dependencies?

alexvy86:
`pnpm i --fix-lockfile` cleans up the removed packages but in my case it also removed some other bits:

image

I was going to suggest doing a `pnpm i --fix-lockfile` + an extraneous modified files check right after installing dependencies and before doing anything else, but based on that result on my local it looks like it could get flaky.

`pnpm i --no-frozen-lockfile` might work for this purpose though; tried it after removing the same package this PR removed, and it only cleaned up the package and its dependencies.

PR #26683, opened 2026-03-10
Add some utilities for explict typing of erased types

CraigMacomber:
## Description

These alpha utilities should be helpful for making erased schema types easy to use with isolateDeclrations.

## Reviewer Guidance


examples/utils/import-testing/src/test/build.spec.ts
```diff
@@ -37,10 +37,14 @@ const typescriptVersions = Object.entries(typescriptHostPackageJson.devDependenc
 
 const execFileAsync = promisify(execFile);
 
-async function compileTest(tscName: string, args: string[]): Promise<void> {
+async function compileTest(
```
Josmithr, line 46: Nit: docs for this and its params would probably be good

CraigMacomber: Done.

examples/utils/import-testing/src/test/isolatedDeclarations.ts
```diff
@@ -0,0 +1,162 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/*
+ * Users of Fluid may want to enable the TypeScript compiler flag `isolatedDeclarations`:
+ * https://www.typescriptlang.org/tsconfig/#isolatedDeclarations.
```
Josmithr, line 8: Nit
```suggestion
 * Users of Fluid may want to enable the TypeScript compiler flag {@link https://www.typescriptlang.org/tsconfig/#isolatedDeclarations | isolatedDeclarations.
```

examples/utils/import-testing/src/test/isolatedDeclarations.ts
```diff
@@ -0,0 +1,162 @@
...
+ * This flag is added in TypeScript 5.5 (which is newer that what we compile with by default)
```
Josmithr:
```suggestion
 * This flag is added in TypeScript 5.5 (which is newer than what we compile with by default)
```

examples/utils/import-testing/src/test/isolatedDeclarations.ts
```diff
@@ -0,0 +1,162 @@
...
+ * and thus requires special setup to validate.
+ * See ./build.spec.ts for the test which validates this.
+ *
+ * Note that as demonstrated in this file, while it is possible to use tree schema with `isolatedDeclarations`, its is not particularly nice to do so.
```
Josmithr:
```suggestion
 * Note that as demonstrated in this file, while it is possible to use tree schema with `isolatedDeclarations`, it is not particularly nice to do so.
```

examples/utils/import-testing/src/test/isolatedDeclarations.ts
```diff
@@ -0,0 +1,162 @@
...
+ * Consider using the `eraseSchemaDetails` pattern to encapsulate much of the schema as possible to reduce the amount of impacted code.
```
Josmithr:
```suggestion
 * Consider using the `eraseSchemaDetails` pattern to encapsulate as much of the schema as possible to reduce the amount of impacted code.
```

examples/utils/import-testing/src/test/isolatedDeclarations.ts
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * Another options is avoiding using `isolatedDeclarations` for schema.
```
Josmithr: Nit
```suggestion
 * Another option is to avoid using `isolatedDeclarations` for schema.
```

examples/utils/import-testing/src/test/isolatedDeclarations.ts
```diff
@@ -0,0 +1,162 @@
...
+ * This can be done by putting the schema in a separate TypeScript project with a ts-config file that disables `isolatedDeclarations`.
+ * This can be done within the same package, and roughly amounts to using TypeScripts ability to generate d.ts files
```
Josmithr:
```suggestion
 * This can be done within the same package, and roughly amounts to using TypeScript's ability to generate d.ts files
```

examples/utils/import-testing/src/test/isolatedDeclarations.ts
```diff
@@ -0,0 +1,162 @@
...
+	ErasedSchemaSubclassable,
+	TreeNodeSchema,
+	ObjectNodeSchema,
+	ObjectNodeSchemaWorkaround,
+} from "@fluidframework/tree/alpha";
+// eslint-disable-next-line import-x/no-internal-modules
+import type { requireAssignableTo } from "@fluidframework/tree/internal";
+
+// -------------------------------------- //
+// Show schema factories can be exported
```
Josmithr: Nit: `// #region Foo` `// #endregion` blocks might be a simpler alternative to these divider comments (and would make sections collapsable in the IDE)

packages/dds/tree/src/simple-tree/node-kinds/object/objectNodeTypes.ts
```diff
@@ -24,6 +24,34 @@ import type {
 	InsertableObjectFromSchemaRecord,
 } from "./objectNode.js";
 
+/**
+ * {@link (ObjectNodeSchema:interface)} with a workaround to avoid a specific known TypeScript issue which causes it to not be assignable to itself in some cases.
```
Josmithr, line 28: It would probably be good to document what the cases may look like where a user would need to use this. Also, it doesn't look like we link to this from any other APIs. I worry how discoverable this API will be in the cases where it is needed.

examples/utils/import-testing/src/test/isolatedDeclarations.ts
CraigMacomber: This is not a doc comment, its just a block comment, so I don't think tags like @link are helpful. As is, its already clickable in vscode

packages/dds/tree/src/simple-tree/node-kinds/object/objectNodeTypes.ts
CraigMacomber: Oops, I forgot to replace the inline type in schema factory alpha with this one. Done. Also added more details here. You should only need this when its already explicitly being referred to in a type so discoverability should be fine (now that I use it properly). I was getting annoyed at users of SharedTree which were copying the inline type with this workaround explicitly because it showed up in their type signatures and were losing the doc comment in the process making it very confusing. Now they should have a named type to reference so it won't be duplicated losing docs.

examples/utils/import-testing/src/test/isolatedDeclarations.ts
CraigMacomber: Done

packages/dds/tree/src/simple-tree/node-kinds/object/objectNodeTypes.ts
```diff
@@ -27,6 +27,9 @@ import type {
 /**
  * {@link (ObjectNodeSchema:interface)} with a workaround to avoid a specific known TypeScript issue which causes it to not be assignable to itself in some cases.
  * @remarks
+ * If dealing which a schema whose inferred type includes this workaround (because it was produced by a schema factory API which uses it),
```
Josmithr: (The view I'm in won't let me leave a suggestion) "which" should be "with" here.

Josmithr (approved):

PR #26684, opened 2026-03-10
fix: Clear snapshot cache if possible in case of redirection

lindsnguyen:
## Description

* Update epochTracker.ts to clear cache when we are in the scenario where a redirect error is about to be thrown.
* Prior to this fix, cache is kept even if the page was redirected therefore it would be stale the next time it's loaded.
* fixes https://dev.azure.com/fluidframework/internal/_workitems/edit/58855

## Reviewer Guidance

* Updated a unit test that is already going through this scenario and adds some additional validation to confirm the cache is indeed cleared.

### DIFF
--- a/packages/drivers/odsp-driver/src/epochTracker.ts
+++ b/packages/drivers/odsp-driver/src/epochTracker.ts
@@ -301,7 +301,7 @@ export class EpochTracker implements IPersistedFileCache {
 				await this.checkForEpochError(error, epochFromResponse, fetchType);
 				throw error;
 			})
-			.catch((error) => {
+			.catch(async (error) => {
 				// If the error is about location redirection, then we need to generate new resolved url with correct
 				// location info.
 				if (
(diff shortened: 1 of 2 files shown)


packages/drivers/odsp-driver/src/epochTracker.ts
```diff
@@ -320,6 +320,9 @@ export class EpochTracker implements IPersistedFileCache {
 							{ driverVersion, redirectLocation },
 						);
 						locationRedirectionError.addTelemetryProperties(error.getTelemetryProperties());
+						// Clear the cache for this file entry since the site/geo has moved.
+						// The cached snapshot was stored under the old siteUrl and is no longer valid.
+						await this.removeEntries();
```
jatgarg, line 325: failure in removeEntries() would mask the original locationRedirectionError. you can wrap the removeEntries() call in a try/catch so that cache-clearing failures don't prevent the redirect error from propagating.

lindsnguyen: `this.removeEntries` is calling into the async function at line 202 that already wraps the actual `cache.removeEntries` call in a try catch that just logs the error. I figured that would be good enough? 

I was using the other call to this at line 461 as an example too and I don't _think_ it's necessary but let me know! I'm not super familiar with typescript still so I might just not understand how errors propagate.

jatgarg (approved):

PR #26687, opened 2026-03-10
Add stateless function for Semantic Editing

steffenloesch:
## Description

Adds a stateless free function for semantic editing that can be used without creating a stateful tree agent.

More specifically:
- Adds a new `executeSemanticEdit()` stateless function that performs semantic editing on a SharedTree without requiring a full `TreeAgent` instance. This is useful for one-shot edit operations where maintaining conversation history isn't needed.
- Extracts shared edit-loop logic from `TreeAgentImpl.message()` into an internal `runEditLoop()` helper, reducing duplication between the stateful agent and the new stateless API

## Breaking Changes

None, only API additions.

## Test plan

- [x] Unit tests for single edit, multiple edits, edit failure recovery, max edit enforcement, bad tool args, missing `invoke()`, and missing `editToolName`
- [x] Existing `createTreeAgent` tests pass, plus new test for infinite-loop termination
- [x] Run `tree-agent-langchain` integration tests with `executeSemanticEdit()`


packages/framework/tree-agent/src/api.ts
```diff
@@ -369,6 +369,57 @@ export interface TreeAgentOptions<TSchema extends ImplicitFieldSchema> {
...
+	 * @remarks If not provided, the generated code will be executed using a simple JavaScript eval.
+	 */
+	editor?: SynchronousEditor<TSchema> | AsynchronousEditor<TSchema>;
+	/**
+	 * The maximum number of sequential edits the LLM can make before we assume it's stuck in a loop.
+	 */
+	maximumSequentialEdits?: number;
+	/**
+	 * An optional starting conversation history (including a system prompt) to use instead of generating one from the tree.
+	 * @remarks If provided, the history is cloned and used as-is. If not provided, a system prompt is generated from the tree schema.
```
noencke: This is underselling the fact that if you provide the history yourself, it needs to include (lead with?) a system prompt that was generated by a prior use of an agent or of `executeSemanticEdit`. Otherwise, the LLM is going to have no idea what it is supposed to be doing.

This is a weakness of having this degree of freedom. You _can_ pre-construct your message history however you want, but there are certain and strict requirements of how you can do it. Those need to be very clearly documented, and ideally validated - for example, check that the history contains a substring we know is uniquely present in one of our system prompts and throw an error if not.

packages/framework/tree-agent/src/api.ts
```diff
@@ -369,6 +369,57 @@ export interface TreeAgentOptions<TSchema extends ImplicitFieldSchema> {
...
+	/**
+	 * The full conversation history including system prompt, user message, tool calls, and the final assistant response.
+	 */
+	readonly history: TreeAgentChatMessage[];
+	/**
+	 * Whether the last edit attempt failed.
+	 * @remarks When using {@link executeSemanticEdit} directly, edits are applied to the input tree and this flag is informational.
+	 * When used internally by {@link TreeAgent.message}, this flag drives branch merge/dispose decisions.
+	 */
+	readonly lastEditFailed: boolean;
```
noencke: What is an "edit" here? As this doc alludes to, a single "edit" (via `executeSemanticEdit`) applies multiple "edits". Now that we're making these concepts more user facing, we need to be consistent with terminology. Maybe an edit is the outer thing, and a "change" is the inner thing? Or "mutation" is the inner thing? Or something else. Currently, it doesn't make sense for an "edit" to give you back a flag called "lastEditFailed".

In addition - this is kind of a weird thing to expose to a user anyway, is it not? What are they supposed to do with it? If it's true, does that mean that all of the inner changes except the last one applied, and so they should rollback or something? If that's not what it means, why expose it at all, vs. just a "the whole thing failed or the whole thing succeeded". I imagine this is just here because we're trying to re-use the same interface for both the external and internal logic.

packages/framework/tree-agent/src/agent.ts
```diff
@@ -500,4 +460,118 @@ export function createTreeAgent<TSchema extends ImplicitFieldSchema>(
 	return new TreeAgentImpl(model, tree, options);
 }
 
+/**
+ * Executes a stateless semantic edit loop on a tree.
+ * @remarks This function runs the same edit loop as {@link TreeAgent.message}, but without persistent state:
+ * - No change listener or "tree has changed" notifications.
```
noencke: Er, the tree _will_ emit change notifications. The "tree has changed" that this is talking about is in internal implementation detail of the agent, we shouldn't make users worry about it here.

packages/framework/tree-agent/src/agent.ts
```diff
@@ -500,4 +460,118 @@ export function createTreeAgent<TSchema extends ImplicitFieldSchema>(
...
+ * - Edits are applied directly to the input tree (no query-level branch isolation).
```
noencke: Once again, query-level branch isolation is an implementation detail of the agent (which might go away anyway). This shouldn't be mentioned here.

packages/framework/tree-agent/src/agent.ts
```diff
@@ -500,4 +460,118 @@ export function createTreeAgent<TSchema extends ImplicitFieldSchema>(
...
+ * - Per-edit branches still provide error isolation for individual edits.
```
noencke: Same thing. These must just be Claude-generated comments - I don't think they're appropriate for a user-facing API.

packages/framework/tree-agent/src/agent.ts
steffenloesch: That's fair, need to modify the comment.

packages/framework/tree-agent/src/agent.ts
steffenloesch: Yep they are indeed Claude-generated comment. Will change them before going out of draft.

packages/framework/tree-agent/src/api.ts
steffenloesch: For context: this was a late addition to allow for code re-use with `TreeAgent`. You bring up very good points. I think it would be better to remove this argument, and to use a private function for code re-use.

packages/framework/tree-agent/src/api.ts
steffenloesch: Fair points. It's a similar situation as passing in the history: it felt like a good idea in order to allow `TreeAgent` to use this directly. Will refactor.

steffenloesch: Not sure if we need to be super systematic about "edit". The return type is now just a string. It's important for the caller to know that the input tree is getting modified. The caller can then choose what to do with it, for example run some human-in-the-loop logic and then either discard or accept the changes.

packages/framework/tree-agent/src/agent.ts
steffenloesch: The comment is now updated.

packages/framework/tree-agent/src/agent.ts
steffenloesch: Done.

packages/framework/tree-agent/src/api.ts
steffenloesch: `executeSemanticEdit` now doesn't take the history as argument anymore.

noencke (approved):

noencke (approved):

PR #26697, opened 2026-03-11
tree: Improve RangeMap interface

alex-pardes:
## Description

This makes a few changes to the RangeMap utility class to make it easier to use.

### DIFF
--- a/packages/dds/tree/src/util/rangeMap.ts
+++ b/packages/dds/tree/src/util/rangeMap.ts
@@ -34,8 +34,8 @@ export class RangeMap<K, V> {
 	 * If `offsetValue` is left unspecified, all keys in a block will be given the same value.
 	 */
 	public constructor(
-		private readonly offsetKey: (key: K, offset: number) => K,
-		private readonly subtractKeys: (a: K, b: K) => number,
+		public readonly offsetKey: (key: K, offset: number) => K,
+		public readonly subtractKeys: (a: K, b: K) => number,
 		public readonly offsetValue: (value: V, offset: number) => V = defaultValueOffsetFn,
 	) {
 		this.tree = new BTree(undefined, subtractKeys);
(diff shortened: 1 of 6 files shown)


packages/dds/tree/src/util/rangeMap.ts
```diff
@@ -211,11 +198,23 @@ export class RangeMap<K, V> {
...
+		const result = new RangeMap<K, V>(this.offsetKey, this.subtractKeys, this.offsetValue);
+		for (const entry of this.entries()) {
+			result.set(mapKey(entry.start), entry.length, mapValue(entry.value));
+		}
+		return result;
+	}
+
 	/**
 	 * Returns a new map which contains the entries from both input maps.
 	 * Whenever both maps contain entires for the same keys, the value from map `b` is used in the returned map.
```
noencke: unless otherwise specified by the merg func

packages/dds/tree/src/util/rangeMap.ts
```diff
@@ -211,11 +198,23 @@ export class RangeMap<K, V> {
 		return cloned;
 	}
 
+	public mapEntries(mapKey: (key: K) => K, mapValue: (value: V) => V): RangeMap<K, V> {
```
noencke: Maybe just combine this with `clone`, and have the parameters be optional? Given that this returns another RangeMap, "clone" with map-like callbacks is more intuitive to me than "mapEntries" with map-like callbacks.

packages/dds/tree/src/util/rangeMap.ts
noencke: The choice of "b" by default is arbitrary. One way you could make it less arbitrary is by making it asymmetrical - have this be a method `a.union(b)`. Still does not mutate either `a` or `b`. This matches the pattern of JS's built-in 'array.concat' - which is also combining two like data structures into a third with the contents of both.

packages/dds/tree/src/util/rangeMap.ts
```diff
@@ -339,3 +350,11 @@ function subtractIntegers<K extends number>(a: K, b: K): number {
 function defaultValueOffsetFn<T>(value: T, offset: number): T {
 	return value;
 }
+
+export function areAdjacentIntegerRanges(
```
noencke: Is this valuable? Totally fine if you like it, just making sure it wasn't thrown in carelessly by Claude or something - seems kind of trivial.

noencke (approved):

packages/dds/tree/src/util/rangeMap.ts
alex-pardes: This is used on my attach-detach branch, but even there it's dubious that it belongs in this file. I'll remove it from this PR.

packages/dds/tree/src/util/rangeMap.ts
alex-pardes: I'll just remove this method for now, since we're not using it yet.

packages/dds/tree/src/util/rangeMap.ts
alex-pardes: I've updated the comment. I don't have a strong opinion about whether it makes more sense to have a static or a method.

noencke (approved):

PR #26698, opened 2026-03-11
Add configuration for consumer handling of retry on container creation

kian-thompson:
Currently, we continue to retry retriable errors indefinitely on container creation. However in case of certain consumers where there is limited time (30/60 seconds) for the create, we should not have infinite retries even on retriable errors.

If the consumer of `attach` wants to handle the retry mechanism, they can enable the `"Fluid.Container.DisableCloseOnAttachFailure"` flag. This flag will cause no internal retries to occur and let any error surface to the consumer of `attach`.

[AB#57593](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/57593)


anthony-murphy:
we built the ability to allow customers to retry attach if it should fail. did we consider moving retry for attach/create failures to the consumer when retriable attach is enabled: "Fluid.Container.RetryOnAttachFailure" with the eventually goal for enabling it by default, as the current behavior of just retrying forever isn't great for customers. Bascially, we can just set retries to 0 when Fluid.Container.RetryOnAttachFailure is on, and then the consumer owns retry, and whatever policy they want?

what i like about this approach is it lets use move away from exposing complex policy in our apis, and lets consumers have control.

packages/loader/container-loader/src/container.ts
```diff
@@ -1547,6 +1549,12 @@ export class Container
 				service.on("metadataUpdate", this.metadataUpdateHandler);
 			}
 		} else {
+			// When RetryOnAttachFailure is enabled, use no internal retries
+			// The consumer will own the retry policy
+			const retryOnAttachFailure =
+				this.mc.config.getBoolean("Fluid.Container.RetryOnAttachFailure") === true;
```
shlevari: Is this flag name already locked in? To me its naming is unintuitive -- it sounds like it's telling the Fluid Container to do retries, but instead its saying the caller will do the retries.

shlevari: Also, does it make sense to give the caller control over how many retries to do? If they want to disable it, set it to zero, but that way it gives them the opportunity to optimize for their specific startup time requirements

shlevari (commented):
Looks good in general, just some light comments about the naming and exposure surface for controlling the retries.

packages/loader/container-loader/src/container.ts
ChumpChief: +1, MaxRetriesOnAttachFailure seems nice, otherwise DisableRetryOnAttachFailure maybe

packages/loader/driver-utils/src/runWithRetry.ts
```diff
@@ -45,6 +46,12 @@ export interface IProgress {
 	 * @param error - error object returned from the call.
 	 */
 	onRetry?(delayInMs: number, error: unknown): void;
+
+	/**
+	 * Maximum number of retries before giving up on a retriable error.
+	 * If undefined, retries will continue indefinitely (default behavior).
+	 */
+	maxRetries?: number;
```
ChumpChief: Do we need this on IProgress?  I'd expect that if an AbortSignal and onRetry callback are provided that's enough to write logic to call the AbortSignal in response to the onRetry callback, and with a greater degree of control for the caller?

Not a hard pushback here, but just want to make sure we consider options that are lighter-weight by virtue of not baking in policy enforcement.

packages/loader/container-loader/src/container.ts
kian-thompson: @anthony-murphy Thoughts on changing the name to something like `Fluid.Container.DisableRetryOnAttachFailure`?

> Also, does it make sense to give the caller control over how many retries to do?

The important thing here is that the errors are bubbled up to the consumer of `attach`. I don't have any strong preference on how consumers can implement a retry mechanism, but retries are not necessary for this ask.

anthony-murphy: the flag primarily allows attach to be retired by the consumer, as without it the container will close when attach fails

anthony-murphy: the goal is to get out of the container level retry game here, and let the consumer decide what they want to do, retry calling, serialize the container and save it somewhere, or dispose the container throw it away

kian-thompson: I changed the flag name to `DisableCloseOnAttachFailure` to be extra clear about what it does

anthony-murphy (approved):

PR #26703, opened 2026-03-11
test(import-testing): Cross-package schema tests for tree export TypeScript type resolution bug 

WillieHabi:
## Description

Adds compile-time tests to `import-testing` that document a known TypeScript type resolution bug in tree's exports. When all API tiers (`.`, `./alpha`, `./beta`, etc.) point to the same JS module (`lib/index.js`), TypeScript ignores tree's exports type override and generates invalid import paths — resolving `import("@fluidframework/tree/alpha")` to `import("@fluidframework/tree")` in `.d.ts` output, losing alpha-only types like `ObjectNodeSchema`.

This causes: `TS2322: Type 'typeof AppState' is not assignable to type 'ImplicitFieldSchema'`

Tests use `@ts-expect-error` to document these failures. A follow-up PR (#26522) fixes tree's exports by giving each tier its own JS entrypoint, at which point the `@ts-expect-error` lines should be removed.

### Test structure

**Schema source** (`src/crossPackage/`):
- `schemaUtils.ts` — `SchemaFactoryAlpha` instance (must be in a separate file from definitions to trigger the bug)
- `schemaDefinitions.ts` — schema classes importing the factory

**Consumer tests** (`src/test/crossPackage/`):

| | Node16 | Bundler |
|---|---|---|
| **ESM** | `node16Importer.ts` | `bundlerImporter.ts` |
| **CJS** | `node16Importer.ts` (via `tsconfig.crossPackageConsumer.cjs.json`) | N/A |

The schema source is compiled three ways (Node16 ESM, Bundler ESM, Node16 CJS) to produce `.d.ts` files under `lib/crossPackage/` and `dist/crossPackage/`. The consumer tests then import these as a separate package via the new `exports` entries (`./crossPackageSchema/node16`, `./crossPackageSchema/bundler`).


examples/utils/import-testing/package.json
```diff
@@ -12,11 +12,23 @@
 	"license": "MIT",
 	"author": "Microsoft and contributors",
 	"type": "module",
+	"exports": {
+		"./crossPackageSchema": {
+			"import": "./lib/crossPackageSchemaExports.js",
+			"require": "./dist/crossPackageSchemaExports.js"
```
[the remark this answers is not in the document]
WillieHabi: Addressed. removed dead requrire condition from crossPackageSchema export

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "./tsconfig.json",
+	"include": ["src/test/crossPackageImporter.ts", "src/test/crossPackageBundlerImporter.ts"],
+	"exclude": [],
+	"compilerOptions": {
+		"outDir": "./lib/test/crossPackageConsumer",
+		"composite": false,
+		"skipLibCheck": true,
+		"noUnusedLocals": false,
+	},
```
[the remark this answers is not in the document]
WillieHabi: Addressed, added a build:test:esm:crossPackage script and wired it in

examples/utils/import-testing/src/crossPackage/schemaExports.ts
jason-ha, line 1: This file appears to be superfluous. If it is important add comments.

examples/utils/import-testing/src/crossPackage/schemaDefinitions.ts
```diff
@@ -0,0 +1,29 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/* eslint-disable jsdoc/require-jsdoc */
+
+import { sf } from "./crossPackageSchemaUtils.js";
+
+export class Position extends sf.objectAlpha("Position", {
```
jason-ha, line 18: Add some comments about what is important here.

examples/utils/import-testing/package.json
```diff
@@ -12,14 +12,26 @@
...
+			"import": "./lib/bundler/crossPackageSchemaExports.js"
+		}
+	},
 	"scripts": {
 		"build": "fluid-build . --task build",
 		"build:commonjs": "fluid-build . --task commonjs",
 		"build:compile": "fluid-build . --task compile",
-		"build:esnext": "tsc --project ./tsconfig.json",
+		"build:esnext": "npm run build:esnext:main && npm run build:esnext:crossPackage && npm run build:esnext:bundler",
+		"build:esnext:bundler": "tsc --project ./tsconfig.crossPackage.bundler.json",
```
jason-ha: Since this is about cross package, can we name it such?

examples/utils/import-testing/package.json
```diff
@@ -12,14 +12,26 @@
...
+		"build:esnext:crossPackage": "tsc --project ./tsconfig.crossPackage.json",
+		"build:esnext:main": "tsc --project ./tsconfig.json",
 		"build:test": "npm run build:test:esm && npm run build:test:cjs",
 		"build:test:cjs": "fluid-tsc commonjs --project ./src/test/tsconfig.cjs.json",
-		"build:test:esm": "tsc --project ./src/test/tsconfig.json",
+		"build:test:esm": "tsc --project ./src/test/tsconfig.json && npm run build:test:esm:crossPackage",
+		"build:test:esm:crossPackage": "tsc --project ./tsconfig.crossPackageConsumer.json",
```
jason-ha: structure should be consistent with other multi build steps (like with build:esnext above)

examples/utils/import-testing/tsconfig.test.json
jason-ha, line 1: I don't understand this file. Why does it include anything that is not under test?
Can we add some comments?

examples/utils/import-testing/src/test/tsconfig.json
```diff
@@ -19,6 +19,8 @@
 		"incremental": false,
 	},
 	"include": ["./**/*"],
+	// Cross-package importers are compiled by tsconfig.crossPackageConsumer.json (no project references).
+	"exclude": ["./crossPackageImporter.ts", "./crossPackageBundlerImporter.ts"],
```
jason-ha: Why do we need this complication? What is special that they need to be built separately?

examples/utils/import-testing/src/test/crossPackageImporter.ts
```diff
@@ -0,0 +1,29 @@
...
+/* eslint-disable import-x/no-internal-modules */
+import {
+	AppState,
+	Container,
+	Dimensions,
+	Position,
+} from "@fluid-example/import-testing/crossPackageSchema";
+/* eslint-enable import-x/no-internal-modules */
+
+// @ts-expect-error: typeof AppState is not assignable to ImplicitFieldSchema due to normalized import paths
```
jason-ha: Let's not say "normalized". That isn't what happens. There isn't normalization that happens.
tsc is resolving reference to types that result from declaration expression. It thinks they are exported from anything that references index.js (where the types are actually exported from), but it is wrong per the .d.ts for type condition.
You can say typescript type resolution bug that ignores tree's exports type override and generates invalid import path.

examples/utils/import-testing/src/test/crossPackageImporter.ts
```diff
@@ -0,0 +1,29 @@
...
+const _config = new TreeViewConfiguration({ schema: AppState });
+// @ts-expect-error: constructor args not recognized due to normalized import paths
+const _container = new Container({
+	id: "test",
+	// @ts-expect-error: constructor args not recognized due to normalized import paths
+	position: new Position({ x: 0, y: 0 }),
+	// @ts-expect-error: constructor args not recognized due to normalized import paths
+	dimensions: new Dimensions({ width: 100, height: 100 }),
```
jason-ha: It is often helpful to capture the actual error or a portion of it. Then we can see if something changes. (I wish you could ignore an exact error or set of them.)

examples/utils/import-testing/src/test/crossPackage/node16Importer.ts
jason-ha, line 1: Needs comment on current state - this is effectively a disabled test.

examples/utils/import-testing/package.json
```diff
@@ -12,14 +12,26 @@
 	"license": "MIT",
 	"author": "Microsoft and contributors",
 	"type": "module",
+	"exports": {
+		"./crossPackageSchema": {
+			"import": "./lib/crossPackageSchemaExports.js"
+		},
+		"./crossPackageSchemaBundler": {
+			"import": "./lib/bundler/crossPackageSchemaExports.js"
```
jason-ha: bundler is name qualified but node16 is not. They should both be qualified.

(An overall cleaner structure would have src/crossPackage to keep all in one place and not have so many individual file references around. Then you could generate /lib|dist/crossPackage/node and /lib|dist/crossPackage/bundler.
The issue happens with CommonJS too, right?)

examples/utils/import-testing/package.json
```diff
@@ -58,7 +69,32 @@
...
+		"tasks": {
+			"build:esnext:main": [
+				"^build:esnext"
+			],
+			"build:esnext:crossPackage": [
+				"build:esnext:main"
+			],
+			"build:esnext:bundler": [
+				"build:esnext:main"
+			],
```
jason-ha: These dependencies do not appear needed.
I would expect that these and build:esnext:main depend on `["^api-extractor:esnext", "^build:esnext"]`.

examples/utils/import-testing/package.json
```diff
@@ -58,7 +69,32 @@
...
+			"build:test:esm:crossPackage": [
+				"build:esnext"
+			],
+			"build:test:esm": [
+				"build:esnext"
```
jason-ha, line 106: Looks superfluous (if not currently incorrect)

examples/utils/import-testing/src/crossPackage/schemaExports.ts
WillieHabi: added comment explaining why three file split is important for deeper type resolution chain to trigger the bug for node16. Without this file Node16 preserves the correct paths

examples/utils/import-testing/src/crossPackage/schemaDefinitions.ts
WillieHabi: Added new comment

examples/utils/import-testing/package.json
WillieHabi: Split build:test:esm into build:test:esm:main and build:test:esm:crossPackage

examples/utils/import-testing/package.json
WillieHabi: Renamed to build:esnext:crossPackage:node16 and build:esnext:crossPackage:bundler

examples/utils/import-testing/tsconfig.test.json
WillieHabi: This file is pre-exisiting and I think it used by build.spec.ts to compile the package with multiple versions. I added some comments explaining this. The excludes we added prevent the cross-package .d.ts files, which have intentionally broken import paths.

examples/utils/import-testing/src/test/tsconfig.json
WillieHabi: This got caught in CI. The cross-package importer files use @ts-expect-error to document known failures and this tsconfig has project references which bypass the package export map (I believe this is the case) so the errors don't occur and the @ts-expect-error becomes unused (TS2578). The importers must be compiled by tsconfig.crossPackageConsumer.json (with no project references) to trigger the bug. Added a more detailed comment explaining this.

examples/utils/import-testing/src/test/crossPackageImporter.ts
WillieHabi: Updated to "TypeScript type resolution bug that ignores tree's exports type override and generates invalid import paths."

examples/utils/import-testing/src/test/crossPackageImporter.ts
WillieHabi: Added actual error codes: `TS2322: typeof AppState not assignable to ImplicitFieldSchema` and `TS2554: Expected 0 arguments.`

examples/utils/import-testing/src/test/crossPackage/node16Importer.ts
WillieHabi: "Currently a disabled test: @ts-expect-error documents known failures... Fixing tree's exports will resolve these — at which point the @ts-expect-error lines should be removed."

examples/utils/import-testing/package.json
WillieHabi: added `^api-extractor:esnext` to `build:esnext:main` dependencies alongside `^build:esnext`.

examples/utils/import-testing/package.json
WillieHabi: removed the build:test:esm to build:esnext dependency. 

build:test:esm:main and build:test:esm:crossPackage each declare their own build:esnext dependency.

examples/utils/import-testing/package.json
WillieHabi: Renamed exports to ./crossPackageSchema/node16 and ./crossPackageSchema/bundler and updated importers to match 

Will look to do teh refactor and look to add cjs variant

examples/utils/import-testing/src/crossPackage/schemaExports.ts
jason-ha: That doesn't make any sense to me.
The bug exists in schemaDefinitions.d.ts. Look in lib/crossPackage/node16|bundler.
If you see this line:
`declare const Position_base: import("@fluidframework/tree").ObjectNodeSchema<"cross-package-example.Position", {`
then the issue is there.

examples/utils/import-testing/src/test/crossPackageCjsImporter.ts
```diff
@@ -0,0 +1,34 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/*
+Compile-time test for cross-package schema consumption via CJS (require) resolution.
```
jason-ha: This is Node16 CJS.
We don't need to separate this from Node16 ESM.
It is the same content and can just be built twice.

examples/utils/import-testing/src/test/crossPackage/node16Importer.ts
jason-ha, line 1: rename to have Node16 in the name to match Bundler version (or combine into single file)

examples/utils/import-testing/tsconfig.test.json
jason-ha: Thanks. So, this can be temporary (sounds like).
Same ask for import test structure as there was for exports. Place them all under test/crossPackage folder to keep things organized, consistent, and easy to maintain.

examples/utils/import-testing/package.json
```diff
@@ -12,14 +12,30 @@
...
+		"./crossPackageSchema/bundler": {
+			"import": "./lib/crossPackage/bundler/schemaExports.js"
+		}
+	},
 	"scripts": {
 		"build": "fluid-build . --task build",
 		"build:commonjs": "fluid-build . --task commonjs",
 		"build:compile": "fluid-build . --task compile",
-		"build:esnext": "tsc --project ./tsconfig.json",
+		"build:esnext": "npm run build:esnext:main && npm run build:esnext:crossPackage:node16 && npm run build:esnext:crossPackage:bundler",
```
jason-ha: None of these actually depend on the other. So, we should add concurrently pattern here.

examples/utils/import-testing/package.json
```diff
@@ -58,7 +74,42 @@
...
+		"tasks": {
+			"build:esnext:main": [
+				"^api-extractor:esnext",
+				"^build:esnext"
+			],
+			"build:esnext:crossPackage:node16": [
+				"build:esnext:main"
+			],
+			"build:esnext:crossPackage:bundler": [
+				"build:esnext:main"
```
jason-ha: deps are incorrect - should be the same as "build:esnext:main"

examples/utils/import-testing/package.json
```diff
@@ -34,7 +50,8 @@
 		"test:mocha:cjs": "cross-env FLUID_TEST_MODULE_SYSTEM=CJS mocha",
 		"test:mocha:esm": "mocha",
 		"test:mocha:verbose": "cross-env FLUID_TEST_VERBOSE=1 npm run test:mocha",
-		"tsc": "fluid-tsc commonjs --project ./tsconfig.cjs.json && copyfiles -f ../../../common/build/build-common/src/cjs/package.json ./dist"
+		"tsc": "fluid-tsc commonjs --project ./tsconfig.cjs.json && copyfiles -f ../../../common/build/build-common/src/cjs/package.json ./dist",
+		"tsc:crossPackage": "fluid-tsc commonjs --project ./tsconfig.crossPackage.cjs.json"
```
jason-ha: Two issues:
1. build-common/src/cjs/package.json does not apply because the exports here are special (an alternative to special exports is to have "build:esnext:main" depend on the other two and export the module under rename. but it won't work out with the dual Node16 and Bundler emits - the barrel won't support custom outdirs. so don't worry if you don't fully follow this and just get the cjs/package.json correct)
2. structure should be similar to build:esnext - right now there isn't a script entry that references "tsc:crossPackage".
   - When restructured, use script entry "place:cjs:package-stub" for the copying (you'll see it in other places.) You'll also want to add `"place:cjs:package-stub": [], // no cross-package deps needed (without definition default is [^*])` to line 90 of fluidBuild.config.cjs. I discovered flaw yesterday. See https://github.com/microsoft/FluidFramework/commit/3cdee7c9fea57d7573d3e6490da6d13a45b436fa#diff-7e166e7d012639b8b2814714cc26841c485024bedd21cd73162b0cb14c628f77R90.

examples/utils/import-testing/src/crossPackage/schemaExports.ts
WillieHabi: Ahh see it now thanks, removed schemaExports.ts and pointed the exports directly to schemaDefinitions.js. will update the pr description as well

examples/utils/import-testing/src/test/tsconfig.json
```diff
@@ -19,6 +19,15 @@
 		"incremental": false,
 	},
 	"include": ["./**/*"],
+	// Cross-package importers must be compiled WITHOUT project references so TypeScript
+	// resolves types through the package export map (where the bug manifests).
+	// Project references bypass the export map, making @ts-expect-error directives unused.
```
jason-ha: This is bogus.
Project references are fine. The .d.ts's of the referenced project will be used and go through export map - if the export map weren't used the imported modules couldn't be found; tsc wouldn't know where to look.
These won't need to be separated out when the bug is fixed. They need to be separated out now because skipLibCheck must be true until the bug is fixed.
So, this exclusion is temporary.
There are two things that should happen to make sure things are good:
1. fix the cjs package stub
2. add the reference for the two other projects
And to test that things are good: build once without these exclusions but skipLibCheck set to true above.

examples/utils/import-testing/eslint.config.mts
```diff
@@ -7,6 +7,26 @@ import type { Linter } from "eslint";
...
+		// typescript-eslint's auto-discovery can't find.
+		files: ["**/*.ts", "**/*.tsx", "**/*.mts", "**/*.cts"],
+		languageOptions: {
+			parserOptions: {
+				projectService: false,
+				project: [
+					"./tsconfig.json",
+					"./tsconfig.crossPackage.json",
+					"./tsconfig.crossPackageConsumer.json",
+					"./tsconfig.crossPackageConsumer.cjs.json",
```
jason-ha: .cjs version won't be needed when crossPackageCjsImport.ts is removed.

examples/utils/import-testing/tsconfig.crossPackage.node16.json
jason-ha, line 1: This is the Node version and the filename should reflect that.

jason-ha (commented):
(I knew relocating tests here would be yucky)

examples/utils/import-testing/src/test/crossPackageCjsImporter.ts
WillieHabi: Ok removed crossPackageCjsImporter.ts. node16 importer is now built for both ESM and CJS via separate       
tsconfigs (tsconfig.crossPackageConsumer.json and tsconfig.crossPackageConsumer.cjs.json).

examples/utils/import-testing/src/test/crossPackage/node16Importer.ts
WillieHabi: renamed to node16Importer.ts and moved to src/test/crossPackage/ alongside bundlerImporter.ts

examples/utils/import-testing/package.json
WillieHabi: Think I addressed properly, added concurrently as devDep

examples/utils/import-testing/package.json
WillieHabi: ok all three build:esnext:* tasks now have the same deps: `["^api-extractor:esnext", "^build:esnext"]`

examples/utils/import-testing/package.json
```diff
@@ -25,9 +25,9 @@
 		"build": "fluid-build . --task build",
 		"build:commonjs": "fluid-build . --task commonjs",
 		"build:compile": "fluid-build . --task compile",
-		"build:esnext": "npm run build:esnext:main && npm run build:esnext:crossPackage:node16 && npm run build:esnext:crossPackage:bundler",
+		"build:esnext": "npm run build:esnext:main && concurrently \"npm:build:esnext:crossPackage:*\"",
```
jason-ha: None depend on the other; no ordering needed.
```suggestion
		"build:esnext": "concurrently \"npm:build:esnext:*\"",
```

examples/utils/import-testing/package.json
```diff
@@ -45,12 +45,13 @@
 		"format:biome": "biome check . --write",
 		"lint": "fluid-build . --task lint",
 		"lint:fix": "fluid-build . --task eslint:fix --task format",
+		"place:cjs:package-stub": "copyfiles -f ../../../common/build/build-common/src/cjs/package.json ./dist && copyfiles -f ../../../common/build/build-common/src/cjs/package.json ./dist/crossPackage/node16",
```
jason-ha: There should only be one copy into ./dist and it should not be the default one. It needs to be custom because the exports here are custom.

examples/utils/import-testing/tsconfig.crossPackageConsumer.cjs.json
```diff
@@ -1,7 +1,7 @@
 {
 	// This config must be used in a "type": "commonjs" environment. (Use fluid-tsc commonjs.)
 	"extends": "./tsconfig.crossPackageConsumer.json",
-	"include": ["src/test/crossPackageCjsImporter.ts"],
+	"include": ["src/test/crossPackage/node16Importer.ts"],
```
jason-ha: Do comment why exactly one file is here.

fluidBuild.config.cjs
```diff
@@ -86,6 +86,7 @@ module.exports = {
 		"typetests:gen": [],
 		"ts2esm": [],
 		"tsc": tscDependsOn,
+		"place:cjs:package-stub": [], // no cross-package deps needed (without definition default is [^*])
```
jason-ha, line 89: :p - I said line 90 (which would match the outstand commit I referenced) but this is fine. I am sure this will be first and I can resolve.

examples/utils/import-testing/src/crossPackage/schemaDefinitions.ts
jason-ha: Fair - it is also important that no tree import is in this file directly.

examples/utils/import-testing/src/test/crossPackage/node16Importer.ts
```diff
@@ -0,0 +1,34 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/*
+Compile-time test for cross-package schema consumption via Node16 resolution.
```
jason-ha: The comments could be misinterpreted. The important thing isn't consumption via Node16/Bundler (in both consumptions Node16 is used).
The important thing is that the exporter generated code for Bundler or Node16 resolution. The important thing are the conditions when schemaDefinitions.d.ts is generated.

examples/utils/import-testing/eslint.config.mts
```diff
@@ -7,6 +7,25 @@ import type { Linter } from "eslint";
...
+		// typescript-eslint's auto-discovery can't find.
+		files: ["**/*.ts", "**/*.tsx", "**/*.mts", "**/*.cts"],
+		languageOptions: {
+			parserOptions: {
+				projectService: false,
+				project: [
+					"./tsconfig.json",
+					"./tsconfig.crossPackage.node16.json",
+					"./tsconfig.crossPackageConsumer.json",
+					"./src/test/tsconfig.json",
```
jason-ha, line 24: `tsconfig.crossPackageConsumer.json` should really be under ./src/test like the regular test tsconfig.json is.
With it located in the package root it is sharing more with the "production" config than the test config.

examples/utils/import-testing/package.json
jason-ha: 1. The copied cjs/package.json needs to be custom. (I may have duplicated this comment now.)
2. The `place...` entry is good, but it still needs to be called when `tsc` script is invoked (not using fluid-build). The command structure should look like build:esnext with the addition of calling `place...`

examples/utils/import-testing/src/test/tsconfig.json
jason-ha: To test things you did:
> build once without these exclusions but skipLibCheck set to true above.

? (My spidey sense says "no".)

WillieHabi: Sorry had to step away for a bit, replying to unresolved now

You were right, I verified building with --skipLibCheck and no exclusions and it passes. I updated the comment after to say the excluision is temporary due to skipLibCheck:false checking the broke n.d.ts output

examples/utils/import-testing/eslint.config.mts
WillieHabi: Moved

examples/utils/import-testing/src/test/crossPackage/node16Importer.ts
WillieHabi: Updated the comments to focus on the exporter's module resolution when generating schemaDefinitions.d.ts

examples/utils/import-testing/tsconfig.crossPackageConsumer.cjs.json
WillieHabi: Added comment

examples/utils/import-testing/package.json
WillieHabi: Changed so there's one place:cjs:package-stub that copies custom src/cjs/package.json into dist/

examples/utils/import-testing/package.json
WillieHabi: changed so there's one place:cjs:package-stub that copies custom src/cjs/package.json into dist/ and tsc:crossPackage chains place:cjs:package-stub for non-fluid-build invocations

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.cjs.json
```diff
@@ -0,0 +1,9 @@
+{
+	// This config must be used in a "type": "commonjs" environment. (Use fluid-tsc commonjs.)
+	"extends": "./tsconfig.crossPackageConsumer.json",
+	// Only node16Importer.ts is included because bundler resolution is not applicable in a CJS environment.
+	"include": ["./crossPackage/node16Importer.ts"],
```
jason-ha: For best maintenance (assume something else is added), lets switch to an exclude for the special thing that doesn't exist.
```suggestion
	// bundler is only produced and tested for ESM, so exclude it from CJS compilation.
	"exclude": ["./crossPackage/bundlerImporter.ts"],
```

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.cjs.json
```diff
@@ -0,0 +1,9 @@
...
+	"compilerOptions": {
+		"outDir": "../../dist/test/crossPackageConsumer",
```
jason-ha: No need to further next the cross package stuff
```suggestion
		"outDir": "../../dist/test",
```

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.cjs.json
```diff
@@ -0,0 +1,9 @@
...
+	},
```
jason-ha, line 8: There should be references.
```suggestion
	},
	"references": [
		{
			"path": "../../tsconfig.crossPackage.node16.cjs.json",
		},
	],
```

examples/utils/import-testing/package.json
```diff
@@ -45,14 +45,14 @@
...
+		"place:cjs:package-stub": "copyfiles -f ./src/cjs/package.json ./dist",
 		"test": "npm run test:mocha",
 		"test:mocha": "npm run test:mocha:esm && npm run test:mocha:cjs",
 		"test:mocha:cjs": "cross-env FLUID_TEST_MODULE_SYSTEM=CJS mocha",
 		"test:mocha:esm": "mocha",
 		"test:mocha:verbose": "cross-env FLUID_TEST_VERBOSE=1 npm run test:mocha",
-		"tsc": "fluid-tsc commonjs --project ./tsconfig.cjs.json",
-		"tsc:crossPackage": "fluid-tsc commonjs --project ./tsconfig.crossPackage.cjs.json"
+		"tsc": "fluid-tsc commonjs --project ./tsconfig.cjs.json && npm run place:cjs:package-stub",
+		"tsc:crossPackage": "fluid-tsc commonjs --project ./tsconfig.crossPackage.cjs.json && npm run place:cjs:package-stub"
```
jason-ha: Really do suggest keeping parallel with build:next to keep it comparable. `tsc` is the placeholder for complete build; so, it should be using concurrently.
```suggestion
		"tsc": "concurrently \"npm:tsc:*\"",
		"tsc:crossPackage:node16": "fluid-tsc commonjs --project ./tsconfig.crossPackage.node16.cjs.json && npm run place:cjs:package-stub",
		"tsc:main": "fluid-tsc commonjs --project ./tsconfig.cjs.json && npm run place:cjs:package-stub"
```

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.json
```diff
@@ -1,9 +1,9 @@
 {
 	"extends": "./tsconfig.json",
-	"include": ["src/test/crossPackage/**/*"],
+	"include": ["./crossPackage/**/*"],
 	"exclude": [],
 	"compilerOptions": {
-		"outDir": "./lib/test/crossPackageConsumer",
+		"outDir": "../../lib/test/crossPackageConsumer",
```
jason-ha: No need to further next the cross package stuff. You can delete outDir.

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.json
jason-ha, line 10: References are missing.
```suggestion
	},
	"references": [
		{
			"path": "../../tsconfig.crossPackage.bundler.json",
		},
		{
			"path": "../../tsconfig.crossPackage.node16.json",
		},
	],
```

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.json
```diff
@@ -1,9 +1,9 @@
...
 		"composite": false,
```
jason-ha: Why? What I think you actually want is `"incremental": true`. This inherits from test tsconfig.json that explicitly turns incremental off - but I don't think that is needed here.

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.json
```diff
@@ -1,9 +1,9 @@
...
 		"skipLibCheck": true,
```
jason-ha, line 9: Comment skipLibCheck

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.json
```diff
@@ -1,9 +1,9 @@
...
 		"noUnusedLocals": false,
```
jason-ha: Not needed. Already inherited from tsconfig.json

examples/utils/import-testing/src/test/tsconfig.cjs.json
jason-ha, line 16: Missing new references
```suggestion
		},
		{
			"path": "../../tsconfig.crossPackage.node16.cjs.json",
		},
```

examples/utils/import-testing/src/test/tsconfig.cjs.json
```diff
@@ -4,6 +4,8 @@
 	"compilerOptions": {
 		"outDir": "../../dist/test",
 	},
+	// Cross-package importers are compiled separately by tsconfig.crossPackageConsumer*.json.
+	"exclude": ["./crossPackage/**/*"],
```
jason-ha:
```suggestion
	"exclude": [
		// Temporarily all cross-package importers are compiled separately by tsconfig.crossPackageConsumer*.json.
		"./crossPackage/**/*",
		// bundler is only produced and tested for ESM, so exclude it from CJS compilation.
		"./crossPackage/bundlerImporter.ts",
	],
```

examples/utils/import-testing/package.json
```diff
@@ -45,20 +63,61 @@
...
+			"build:test:esm:main": [
+				"build:esnext"
+			],
+			"build:test:esm:crossPackage": [
+				"build:esnext"
+			],
+			"tsc:crossPackage": [
+				"^tsc"
+			],
+			"place:cjs:package-stub": [],
```
jason-ha: already exists in base config - not needed here

examples/utils/import-testing/package.json
```diff
@@ -45,20 +63,61 @@
...
+			],
+			"build:test:esm:main": [
+				"build:esnext"
+			],
+			"build:test:esm:crossPackage": [
+				"build:esnext"
+			],
+			"tsc:crossPackage": [
+				"^tsc"
+			],
```
jason-ha: missing ^api-extractor dep.
With above parallelism suggestion
```suggestion
			"tsc:crossPackage:node16": [
				"^api-extractor:commonjs",
				"^tsc"
			],
			"tsc:main": [
				"^api-extractor:commonjs",
				"^tsc"
			],
```

examples/utils/import-testing/package.json
```diff
@@ -45,20 +63,61 @@
...
+			"place:cjs:package-stub": [],
+			"build:test:cjs:main": [
+				"tsc",
+				"place:cjs:package-stub"
+			],
+			"build:test:cjs:crossPackage": [
+				"tsc:crossPackage",
+				"place:cjs:package-stub",
+				"^tsc"
+			]
```
jason-ha: Don't need to reference `place:cjs:package-stub` as that is a part of the other referenced scripts.
```suggestion
			"build:test:cjs:main": [
				"tsc:main"
			],
			"build:test:cjs:crossPackage": [
				"tsc:crossPackage:node16"
			],
```

jason-ha (commented):
Mostly same make cjs parallel esm suggestions.
Plus test projects missing references.
Made suggestions this time for most everything.

examples/utils/import-testing/src/test/tsconfig.crossPackageConsumer.json
jason-ha: @WillieHabi, why not incremental?

examples/utils/import-testing/package.json
```diff
@@ -65,6 +85,43 @@
...
+				"^tsc"
+			],
+			"build:test:cjs:main": [
+				"tsc:main",
+				"tsc:crossPackage:node16"
+			],
+			"build:test:cjs:crossPackage": [
+				"tsc:crossPackage:node16"
+			]
+		}
```
jason-ha, line 123: The test builds for ESM and CJS are inconsistent.
Both should use the main (`build:esnext` | `tsc`) or both should use precise like "build:test:cjs:*" currently do.
Might be easier to notice if these were sorted alphabetically.

jason-ha (approved):
Two remaining possible improvements, but those can be fixed with the subsequent prod fix.

PR #26707, opened 2026-03-11
Fix tar 6.x vulnerabilities: upgrade ncu to v17 + override for @vvago/vale

frankmueller-msft:
## Summary
- **Upgrades `npm-check-updates` from v16 to v17** in `@fluid-tools/build-cli` — ncu@17 is fully bundled with zero dependencies, eliminating the `pacote → cacache → tar 6.x` transitive chain that was the sole source of tar 6.x across all non-root workspaces
- **Retains tar overrides** in non-root workspaces as an interim measure until the next `@fluid-tools/build-cli` release (published v0.64.0 still ships ncu@16). Override comments note they can be removed after a release includes ncu@17
- **Retains tar override in root workspace** because `@vvago/vale` directly depends on `tar ^6.x` and has no newer release
- Fixes js-yaml 3.14.1 → 3.14.2 in the build-infrastructure test fixture lockfile (was triggering CG alerts), including corrected integrity hash

## Approach
Rather than blindly overriding tar 6.x → 7.x everywhere, this PR addresses the root cause:
1. `npm-check-updates@16` → `17` eliminates tar 6.x from the build-cli dependency tree (ncu@17 has zero deps)
2. Overrides are only kept where the root cause can't be fixed yet (published build-cli, @vvago/vale)

## Test plan
- [x] All CI pipelines pass (build-tools, client, common-utils, protocol-definitions, routerlicious, gitrest, historian, policy checks)
- [ ] Confirm no tar 6.x remains in lockfiles after merge
- [ ] Monitor Component Governance for alert resolution after merge
- [ ] After next build-cli release with ncu@17: remove tar overrides from non-root workspaces

### DIFF
--- a/build-tools/package.json
+++ b/build-tools/package.json
@@ -173,7 +173,7 @@
 			"qs": "^6.15.0",
 			"simple-git": "^3.32.3",
 			"sharp": "^0.34.5",
-			"tar": ">=7.5.11"
+			"tar": "^7.5.11"
 		},
 		"updateConfig": {
 			"ignoreDependencies": [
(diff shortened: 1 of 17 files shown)


frankmueller-msft:
Split into two focused PRs:
- #26731 — tar overrides across all workspaces (security fix, config only)
- #26732 — npm-check-updates v16 → v17 upgrade in build-cli (code change)

PR #26708, opened 2026-03-12
SiteMove/Geo move not handled properly during container load

lindsnguyen:
## Description

When a site move or geo move triggers a location redirection during container load, the sharing link was not being redeemed against the new (redirected) URL. The redirect itself is handled earlier in the flow, but after the redirect the sharing link is not part of the resolved URL. This means on the subsequent retry at the new location, the sharing link can no longer be redeemed — causing the container load to fail for users who accessed via a sharing link.

fixes https://dev.azure.com/fluidframework/internal/_workitems/edit/58231

## Solution 
* On a location redirection error, redeem the sharing link against the redirected URL before re-throwing, so it's already redeemed when the retry happens at the new location.
* Only attempt redeem if a sharing link is actually present.
* Treat redeem as best-effort — catch failures and log telemetry (RedirectRedeemFallback / RedirectRedeemFallbackError), always re-throwing the original redirection error so the normal redirect flow continues.
* Add tests for the no-shareLink and redeem-failure paths.

### DIFF
--- a/packages/drivers/odsp-driver/src/fetchSnapshot.ts
+++ b/packages/drivers/odsp-driver/src/fetchSnapshot.ts
@@ -189,12 +189,31 @@ export async function fetchSnapshotWithRedeem(
 					putInCache,
 					loadingGroupIds,
 				);
-			} else if (isLocationRedirectionError(error)) {
-				const redirectedResolvedUrl: IOdspResolvedUrl = {
-					...getOdspResolvedUrl(error.redirectUrl),
-					shareLinkInfo: odspResolvedUrl.shareLinkInfo,
-				};
-				await redeemSharingLink(redirectedResolvedUrl, storageTokenFetcher, logger);
+			} else if (
+				isLocationRedirectionError(error) &&
+				odspResolvedUrl.shareLinkInfo?.sharingLinkToRedeem !== undefined
+			) {
+				try {
+					// The redirect itself is handled earlier, but we need to redeem the sharing link
+					// now against the redirected URL rather than waiting until the next API call retries
+					// with the redirect URL applied. After this point the sharing link is removed from
+					// the resolved URL, so we wouldn't be able to redeem during a later failure.
+					const redirectedResolvedUrl: IOdspResolvedUrl = {
+						...getOdspResolvedUrl(error.redirectUrl),
+						shareLinkInfo: odspResolvedUrl.shareLinkInfo,
+					};
+
+					await redeemSharingLink(redirectedResolvedUrl, storageTokenFetcher, logger);
+					logger.sendTelemetryEvent(
+						{
+							eventName: "RedirectRedeemFallback",
+							errorType: error.errorType,
+						},
+						error,
+					);
+				} catch (redeemError) {
+					logger.sendErrorEvent({ eventName: "RedirectRedeemFallbackError" }, redeemError);
+				}
 				throw error;
 			} else {
 				throw error;
@@ -808,8 +827,8 @@ function isLocationRedirectionError(error: unknown): error is ILocationRedirecti
 	return (
 		typeof error === "object" &&
 		error !== null &&
-		(error as Partial<ILocationRedirectionError>).errorType ===
-			DriverErrorTypes.locationRedirection
+		(error as Partial<IOdspError>).errorType === DriverErrorTypes.locationRedirection &&
+		(error as Partial<ILocationRedirectionError>).redirectUrl !== undefined
 	);
 }
 
--- a/packages/drivers/odsp-driver/src/fetchSnapshot.ts
+++ b/packages/drivers/odsp-driver/src/fetchSnapshot.ts
@@ -6,7 +6,12 @@
 import { fromUtf8ToBase64 } from "@fluid-internal/client-utils";
 import { assert } from "@fluidframework/core-utils/internal";
 import { getW3CData } from "@fluidframework/driver-base/internal";
-import type { ISnapshot, ISnapshotTree } from "@fluidframework/driver-definitions/internal";
+import {
+	DriverErrorTypes,
+	type ILocationRedirectionError,
+	type ISnapshot,
+	type ISnapshotTree,
+} from "@fluidframework/driver-definitions/internal";
 import {
 	type DriverErrorTelemetryProps,
 	NonRetryableError,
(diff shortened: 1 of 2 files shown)


packages/drivers/odsp-driver/src/fetchSnapshot.ts
```diff
@@ -183,6 +189,13 @@ export async function fetchSnapshotWithRedeem(
 					putInCache,
 					loadingGroupIds,
 				);
+			} else if (isLocationRedirectionError(error)) {
```
jatgarg: Please add a detailed comment explaining the scenario.

packages/drivers/odsp-driver/src/fetchSnapshot.ts
```diff
@@ -791,6 +804,15 @@ export const downloadSnapshot = mockify(
 	},
 );
 
+function isLocationRedirectionError(error: unknown): error is ILocationRedirectionError {
+	return (
+		typeof error === "object" &&
+		error !== null &&
+		(error as Partial<ILocationRedirectionError>).errorType ===
```
jatgarg: Since we don't know yet that it is ILocationRedirectionError, we should use the general error type like we are using IOdspError below when checking isRedeemSharingLinkError.

packages/drivers/odsp-driver/src/fetchSnapshot.ts
jatgarg: We should also check whether the shareLink was present or not and then only go for redeem.

packages/drivers/odsp-driver/src/fetchSnapshot.ts
```diff
@@ -183,6 +189,13 @@ export async function fetchSnapshotWithRedeem(
 					putInCache,
 					loadingGroupIds,
 				);
+			} else if (isLocationRedirectionError(error)) {
+				const redirectedResolvedUrl: IOdspResolvedUrl = {
+					...getOdspResolvedUrl(error.redirectUrl),
+					shareLinkInfo: odspResolvedUrl.shareLinkInfo,
+				};
+				await redeemSharingLink(redirectedResolvedUrl, storageTokenFetcher, logger);
```
jatgarg: Lets catch the error if it happens and throw the original redirection error otherwise we will swallow the original issue since redeem was just best effort flow. For now you can also add a telemetry event if we come into this flow.

packages/drivers/odsp-driver/src/fetchSnapshot.ts
```diff
@@ -791,6 +823,15 @@ export const downloadSnapshot = mockify(
 	},
 );
 
+function isLocationRedirectionError(error: unknown): error is ILocationRedirectionError {
+	return (
+		typeof error === "object" &&
+		error !== null &&
+		(error as Partial<IOdspError>).errorType === DriverErrorTypes.locationRedirection &&
+		(error as Partial<ILocationRedirectionError>).redirectUrl !== undefined
```
jatgarg: No need to check this as redirectUrl is not optional. So, if it is locationRedirection error, it must be present.

packages/drivers/odsp-driver/src/fetchSnapshot.ts
```diff
@@ -183,6 +189,32 @@ export async function fetchSnapshotWithRedeem(
...
+					// now against the redirected URL rather than waiting until the next API call retries
+					// with the redirect URL applied. After this point the sharing link is removed from
+					// the resolved URL, so we wouldn't be able to redeem during a later failure.
+					const redirectedResolvedUrl: IOdspResolvedUrl = {
+						...getOdspResolvedUrl(error.redirectUrl),
+						shareLinkInfo: odspResolvedUrl.shareLinkInfo,
+					};
+
+					await redeemSharingLink(redirectedResolvedUrl, storageTokenFetcher, logger);
+					logger.sendTelemetryEvent(
```
jatgarg, line 201: Maybe move this before the redeemSharingLink

packages/drivers/odsp-driver/src/fetchSnapshot.ts
```diff
@@ -183,6 +189,32 @@ export async function fetchSnapshotWithRedeem(
...
+				odspResolvedUrl.shareLinkInfo?.sharingLinkToRedeem !== undefined
+			) {
+				try {
+					// The redirect itself is handled earlier, but we need to redeem the sharing link
+					// now against the redirected URL rather than waiting until the next API call retries
+					// with the redirect URL applied. After this point the sharing link is removed from
+					// the resolved URL, so we wouldn't be able to redeem during a later failure.
+					const redirectedResolvedUrl: IOdspResolvedUrl = {
+						...getOdspResolvedUrl(error.redirectUrl),
+						shareLinkInfo: odspResolvedUrl.shareLinkInfo,
```
jatgarg: No need to do this as redirect URL is patched full IOdspResolvedUrl with new siteUrl and same rest of the props. So, it already has the shareLink.

lindsnguyen: Talked offline and we decided to leave this alone because while this works in production today because the epoch tracker and fetchSnapshotWithRedeem shares the same odspResolvedUrl object reference, this is an implementation detail and if someone refactors the service setup to clone the resolved URL, or a new caller constructs the epoch tracker differently, this would stop working as expected. The explicit copy would make it work regardless of the implementation.

packages/drivers/odsp-driver/src/fetchSnapshot.ts
```diff
@@ -183,6 +189,31 @@ export async function fetchSnapshotWithRedeem(
...
 				);
+			} else if (
+				isLocationRedirectionError(error) &&
+				odspResolvedUrl.shareLinkInfo?.sharingLinkToRedeem !== undefined
+			) {
+				try {
+					// The redirect itself is handled earlier, but we need to redeem the sharing link
+					// now against the redirected URL rather than waiting until the next API call retries
+					// with the redirect URL applied. After this point the sharing link is removed from
+					// the resolved URL, so we wouldn't be able to redeem during a later failure.
```
jatgarg: > After this point the sharing link is removed from
					// the resolved URL, so we wouldn't be able to redeem during a later failure.

Not sure what does this mean. Why is the sharingLink removed from resolvedURL?

lindsnguyen: changed to:
```
					// The redirect itself is handled earlier, but we need to redeem the sharing link
					// now against the redirected URL rather than waiting until the error reaches the
					// resolveWithLocationRedirectionHandling handler as it will call getAbsoluteURL
					// and would fail due to permission issues since it will not attempt to redeem.
```
to explain the issue better.

jatgarg (approved):

PR #26714, opened 2026-03-12
Support FIC-based authentication flow in OdspTestDriver

Abe27342:
## Description

Microsoft has some incoming changes that tighten security around test tenants. The general idea is that test users can be set up to be FIC-enabled (federated identity credentials). Without going into details on permissions, this allows tokens to be fetched for these users on behalf of "owning applications" rather than require the user's credentials. The federated identity (and source of trust relationship) ultimately comes from either a trusted internal developer (FF team member) or a service connection that the pipeline we run this context in is allowed to access.

Prior to this flow, the tenant checkout steps we had resulted in obtaining username/password combinations for test users. These username/password combinations were exchanged with the standard ROPC OAuth flow for tokens that were used in our e2e tests. In the FIC flow, only usernames are returned, and the same package that supports checking out/releasing test tenants now has a programmatic entrypoint which supports getting tokens for those users.

This PR implements support for using the FIC flow. It does not yet transition CI to use that flow (more testing is needed).

It includes some refactors/renames for clarity. Enumerating those here:

- `OdspTokenConfig` has been renamed to `LoginCredentials` reflecting its role as a discriminated union describing how we're planning on authenticating to ODSP
- Signature and implementation of `getOdspCredentials` has been generally slightly simplified (the "requestedUser" parameter is easily implemented after-the-fact and complicates the inline implementation)


packages/utils/odsp-doclib-utils/src/odspAuth.ts
```diff
@@ -15,7 +15,7 @@ import { unauthPostAsync } from "./odspRequest.js";
  */
 export interface IOdspTokens {
 	readonly accessToken: string;
-	readonly refreshToken: string;
+	readonly refreshToken?: string; // Refresh token is not used in federated credential flow, so it's optional.
 	readonly receivedAt?: number; // Unix timestamp in seconds
 	readonly expiresIn?: number; // Seconds from reception until the token expires
```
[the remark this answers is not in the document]
Abe27342: Don't want to make changes here. We already validate at runtime that refreshToken is populated where we need it. If we want better compile-time type safety, I'd want to restructure the types more wholesale (IMO incremental progress is not worth it)

packages/utils/tool-utils/src/odspTokenManager.ts
```diff
@@ -196,22 +203,29 @@ export class OdspTokenManager {
...
 			// check the cache again under the lock (if it is there)
 			const tokensFromCache = await this.getTokenFromCache(cacheKey);
 			if (tokensFromCache) {
 				if (forceRefresh || !isValidAndNotExpiredToken(tokensFromCache)) {
 					try {
-						// This updates the tokens in tokensFromCache
-						tokens = await refreshTokens(server, scope, clientConfig, tokensFromCache);
-						await this.updateTokensCacheWithoutLock(cacheKey, tokens);
+						if (credentials.type === "fic") {
+							const scopeEndpoint = isPush ? "push" : ("storage" as const);
```
alexvy86: Is the `as const` required?

packages/utils/tool-utils/src/odspTokenManager.ts
```diff
@@ -196,22 +203,29 @@ export class OdspTokenManager {
...
+							const newTokenData = await credentials.fetchToken(scopeEndpoint);
+							tokens = this.ficTokenToIOdspTokens(newTokenData, isPush);
+							await this.updateTokensCacheWithoutLock(cacheKey, tokens);
```
alexvy86: Nit: keep out of the branch?

packages/utils/tool-utils/src/odspTokenManager.ts
```diff
@@ -258,6 +285,52 @@ export class OdspTokenManager {
 		};
 		return fetchTokens(server, scope, clientConfig, credentials);
 	}
+
+	private ficTokenToIOdspTokens(token: string, isPush: boolean): IOdspTokens {
+		// eslint-disable-next-line unicorn/prefer-ternary -- using if statement for clarity
+		if (isPush) {
+			// Push tokens are not standard JWTs. With direct token exchange, the second leg includes information about expiry.
+			// This is not available in the FIC flow, but we request tokens with 1 hour expiry so default to that.
```
alexvy86: Makes it a bit clearer.
```suggestion
			// This is not available in the FIC flow, but in direct token exchange we request tokens with 1 hour expiry so default to that.
```

packages/test/test-drivers/src/odspTestDriver.ts
```diff
@@ -32,16 +32,44 @@ import { compare } from "semver";
...
+		const tokens = await testTenantCheckoutClient.fetchFicTokens([username], scopeEndpoint);
+		if (!Array.isArray(tokens)) {
+			// This error indicates a mismatch between the dynamically imported token fetcher package and this code.
+			// Double-check that the package specified in 'token__package__import__location' is up to date and its entrypoint
+			// matches the typing of `fetchFicTokens` as defined in `TestTenantCheckoutClient`.
+			throw new TypeError("Expected fetchFicTokens to return an array of tokens.");
+		}
+		const token = tokens.find((a) => a.UserPrincipalName === username);
+		if (!token) {
+			throw new Error(`Unable to fetch token for user ${username} and scope ${scopeEndpoint}`);
```
alexvy86: My personal peeve to make sure empty strings don't result in hard-to-interpret error messages 😄 .
```suggestion
			throw new Error(`Unable to fetch token for user '${username}' and scope '${scopeEndpoint}'`);
```

packages/test/test-drivers/src/odspTestDriver.ts
```diff
@@ -32,16 +32,44 @@ import { compare } from "semver";
...
 });
 
+const ficLoginCredentials = (username: string): LoginCredentials => {
+	const fetchToken = async (scopeEndpoint: "storage" | "push"): Promise<string> => {
+		const testTenantCheckoutClient = await getTestTenantCheckoutClient();
+		if (testTenantCheckoutClient === undefined) {
+			throw new Error(
+				'The FIC credential flow relies on a test tenant checkout client, but no client was found. Ensure that the environment variable "token__package__import__location" is set to the location of a package that exports a compatible client.',
+			);
+		}
```
alexvy86: Just throw inside the function instead of allowing an `undefined` return value? I think it makes it more maintainable to keep the error message with the variable name closer to where it's used.

packages/test/test-drivers/src/odspTestDriver.ts
```diff
@@ -90,41 +160,53 @@ export function assertOdspEndpoint(
...
 			? process.env.login__odsp__test__tenants
 			: process.env.login__odspdf__test__tenants;
 
-	if (loginTenants !== undefined) {
+	const ficAccounts = process.env.login__odsp__fic__test__users;
+	if (ficAccounts !== undefined) {
+		const { usernames } = JSON.parse(ficAccounts) as {
+			guid: string;
+			usernames: string[];
+		};
```
alexvy86, line 184: `guid` is for the tenant, and `usernames` the list of accounts within the tenant, I imagine? Documenting it here would be good.

packages/test/test-drivers/src/odspTestDriver.ts
```diff
@@ -90,41 +160,53 @@ export function assertOdspEndpoint(
...
+
+		if (usernames.length === 0) {
+			throw new Error(
+				"login__odsp__fic__test__users was defined does not have any valid usernames.",
```
alexvy86:
```suggestion
				"login__odsp__fic__test__users was defined but does not have any valid usernames.",
```

packages/test/test-drivers/src/odspTestDriver.ts
```diff
@@ -146,9 +228,7 @@ export function getOdspCredentials(
 			// Return the set of account to choose from a single tenant
 			for (let i = 0; i < range.count; i++) {
 				const username = `${range.prefix}${range.start + i}@${tenant}`;
-				if (requestedUserName === undefined || requestedUserName === username) {
-					creds.push({ username, password: range.password });
-				}
+				creds.push({ username, password: range.password });
```
alexvy86, line 230: Since the function is exported as `@internal`, do we need to check for codepaths outside the repo that use it and provide a requestedUsername? To see if they won't complain about potentially getting more than 1 credential now.

packages/test/test-drivers/src/odspTestDriver.ts
```diff
@@ -171,7 +251,7 @@ export function getOdspCredentials(
 		const passwords: { [user: string]: string } = JSON.parse(loginAccounts);
 
 		// Need to choose one out of the set as these account might be from different tenant
-		const username = requestedUserName ?? Object.keys(passwords)[0];
+		const username = Object.keys(passwords)[0];
```
alexvy86, line 253: Same concern as above, if several loginAccounts were being passed and `requestedUsernName` was set, this is potentially changing what that code will get.

packages/test/test-drivers/src/odspTestDriver.ts
Abe27342: Yeah that's nicer thanks for catching--original flow didn't make this possible but it now is easy to do after some other simplification I did.

packages/test/test-drivers/src/odspTestDriver.ts
Abe27342: the guid ends up unused with the current setup. I want to avoid documenting too much of the internal setup in OSS--assumption with the rest of this code is that there needs to be agreement between the formats for env variables. I'll just remove guid from the type here.

packages/test/test-drivers/src/odspTestDriver.ts
Abe27342: per [API support levels](https://fluidframework.com/docs/build/releases-and-apitags), such usages should not exist (or should be ok with breaking arbitrarily on version update).

I'm not aware of any in other code we own outside this repo or partner teams anyhow. The requestedUsername thing is almost certainly some random legacy feature from >5 years ago that used to be used from within the repo.

packages/test/test-drivers/src/odspTestDriver.ts
Abe27342: same response :)

alexvy86 (approved):

PR #26716, opened 2026-03-12
fix(build-tools): allow `generate entrypoints` using "import" resolution

jason-ha:
ts-morph does not infer import conditions when parsing sources and will have underlying typescript use CommonJS "require" even though a package has type "module". See [ts-morph issue #1667](https://github.com/dsherret/ts-morph/issues/1667).

Add option to specify project's `customConditions` through new argument `resolutionConditions`. Any arguments may be specified to allow supporting generating entrypoints that require custom conditions. When neither "import" nor "require" is given, mainEntrypoint extension and package.json "type" will be used to infer one of those.

Additionally add an error when using `generate entrypoints` and some export has no source instead of silently skipping. This was the case when entrypoints were generated for a package that reexported from another and that other packages CommonJS build was not done. `modify fluid-imports` is more conservative if export is not sourced and just emits a warning.

### DIFF
--- a/build-tools/packages/build-cli/src/library/commands/generateEntrypoints.ts
+++ b/build-tools/packages/build-cli/src/library/commands/generateEntrypoints.ts
@@ -447,6 +447,7 @@ export function readArgValues(
 	argQuery: Options,
 ): Pick<
 	Options,
+	// This selects only "out" prefixed keys whose values are `string | undefined`.
 	`out${string}` & keyof Options extends infer Key
 		? Options[Key & keyof Options] extends string | undefined
 			? Key
@@ -455,7 +456,7 @@ export function readArgValues(
 > {
 	const args = commandLine.split(" ");
 
-	const argValues: Record<string, string | undefined> = {};
+	const argValues: Record<`out${string}`, string> = {};
 	for (const argName of Object.keys(argQuery).filter(
 		(key): key is `out${string}` =>
 			key.startsWith("out") &&
--- a/build-tools/packages/build-cli/src/library/commands/generateEntrypoints.ts
+++ b/build-tools/packages/build-cli/src/library/commands/generateEntrypoints.ts
@@ -21,6 +21,7 @@ import { unscopedPackageNameString } from "./constants.js";
 
 interface Options {
 	readonly mainEntrypoint: string;
+	readonly resolutionConditions: string[];
 	readonly outDir: string;
 	readonly outFilePrefix: string;
 
@@ -71,6 +72,7 @@ interface Options {
  */
 export const optionDefaults = {
 	mainEntrypoint: "./src/index.ts",
+	resolutionConditions: [],
 	outDir: "./lib",
 	outFilePrefix: "",
 	outFileAlpha: "alpha",
@@ -96,6 +98,11 @@ export class GenerateEntrypointsCommand extends BaseCommand<
 			description: "Main entrypoint file containing all untrimmed exports.",
 			default: optionDefaults.mainEntrypoint,
 			exists: true,
+		}),
+		resolutionConditions: Flags.string({
+			description: `Import resolution conditions used while resolving imports. If neither "import" nor "require" are specified, one of those will be inferred based on mainEntrypoint file extension and package.json "type" field.`,
+			multiple: true,
+			default: optionDefaults.resolutionConditions,
 		}),
 		outDir: Flags.directory({
 			description: "Directory to emit entrypoint declaration files.",
@@ -148,7 +155,7 @@ export class GenerateEntrypointsCommand extends BaseCommand<
 	};
 
 	public async run(): Promise<void> {
-		const { mainEntrypoint, node10TypeCompat } = this.flags;
+		const { mainEntrypoint, resolutionConditions, node10TypeCompat } = this.flags;
 
 		const packageJson = await readPackageJson();
 
@@ -186,9 +193,32 @@ export class GenerateEntrypointsCommand extends BaseCommand<
 			);
 		}
 
+		const customConditions = [...resolutionConditions];
+		const conditionsPresent =
+			(customConditions.includes("import") ? 1 : 0) +
+			(customConditions.includes("require") ? 2 : 0);
+		if (conditionsPresent === 3) {
+			this.logger.warning(
+				'warning: both "import" and "require" --resolutionConditions flags given; results may be unstable',
+			);
+		} else if (conditionsPresent === 0) {
+			const inferredImportCondition = inferImportConditionFromFilePath(
+				mainEntrypoint,
+				packageJson,
+			);
+			this.logger.info(`Inferred import resolution condition "${inferredImportCondition}"`);
+			customConditions.push(inferredImportCondition);
+		}
+
 		const commandLine = `flub generate entrypoints${this.commandLineArgs()}`;
 		promises.push(
-			generateEntrypoints(mainEntrypoint, mapApiTagLevelToOutput, commandLine, this.logger),
+			generateEntrypoints(
+				mainEntrypoint,
+				customConditions,
+				mapApiTagLevelToOutput,
+				commandLine,
+				this.logger,
+			),
 		);
 
 		if (node10TypeCompat) {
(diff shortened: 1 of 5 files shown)


build-tools/packages/build-cli/src/test/library/commands/generateEntrypoints.test.ts
```diff
@@ -30,6 +30,18 @@ describe("generateEntrypoints", () => {
...
+		// Does not support multi-args (doesn't currently need to)
+		// Takes the first arg of the first multi-arg set.
+		expect(
+			readArgValues(
+				"--resolutionConditions first second --resolutionConditions third",
+				optionDefaults,
+			),
+		).to.deep.equal({
+			...optionDefaults,
+			resolutionConditions: "first",
```
frankmueller-msft: `readArgValues` will assign a plain `string` here, but `Options.resolutionConditions` is typed as `string[]`. TypeScript won't catch this because of the `Record<string, string | undefined>` intermediate in `readArgValues`.

More importantly, `readArgValues` is also called from `getGenerateEntrypointsOutput` (used by `modify fluid-imports`). If someone passes `--resolutionConditions` to `generate entrypoints` and that command line gets round-tripped through `getGenerateEntrypointsOutput`, the conditions will be silently truncated to the first value. Worth a follow-up to handle multi-value args there, or at minimum a TODO.

build-tools/packages/build-cli/src/library/typescriptApi.ts
```diff
@@ -159,7 +160,11 @@ function getNodeApiLevel(node: Node): ApiLevel | undefined {
  * Given a source file, extracts all of the named exports and associated support levels.
  * Named exports without a recognized tag are placed in unknown array.
  */
-export function getApiExports(sourceFile: SourceFile): ExportRecords {
+export function getApiExports(
+	sourceFile: SourceFile,
+	treatmentOfMissing: "throwForMissing" | "warnForMissing",
```
frankmueller-msft, line 165: This is an exported function — the signature change from `getApiExports(sourceFile)` to `getApiExports(sourceFile, treatmentOfMissing, log)` with two new required parameters is a breaking change for any external consumers. Consider making these optional with sensible defaults (e.g. default to `"warnForMissing"` and a no-op logger) to preserve backwards compatibility.

build-tools/packages/build-cli/src/library/commands/generateEntrypoints.ts
```diff
@@ -213,6 +243,21 @@ async function readPackageJson(): Promise<PackageJson> {
...
 }
 
+function inferImportConditionFromFilePath(
+	filePath: string,
+	packageJson: PackageJson,
+): "import" | "require" {
+	const ext = path.extname(filePath);
+	if (ext.match(/^\.m[jt]s/i) !== null) {
+		return "import";
+	}
```
frankmueller-msft, line 261: Two things on these regexes:
1. Missing `$` anchor — `/^\.m[jt]s/` will also match `.mjsx`, `.mts-backup`, etc. Should be `/^\.m[jt]s$/`.
2. The `/i` flag is unnecessary for file extensions, which are conventionally lowercase. Not harmful, but adds ambiguity about whether mixed-case extensions are intentionally supported.

build-tools/packages/build-cli/src/library/commands/generateEntrypoints.ts
```diff
@@ -186,9 +193,32 @@ export class GenerateEntrypointsCommand extends BaseCommand<
 			);
 		}
 
+		const customConditions = [...resolutionConditions];
+		const conditionsPresent =
+			(customConditions.includes("import") ? 1 : 0) +
+			(customConditions.includes("require") ? 2 : 0);
+		if (conditionsPresent === 3) {
+			this.logger.warning(
```
frankmueller-msft, line 204: If results are truly "unstable" when both conditions are present, should this be an error instead of a warning? Letting the tool proceed to produce unreliable output seems like a footgun.

frankmueller-msft (commented):
Solid fix for a real ts-morph limitation. The `customConditions` approach is the right call. A few things worth discussing below.

build-tools/packages/build-cli/src/library/typescriptApi.ts
```diff
@@ -224,7 +242,7 @@ export function getApiExports(sourceFile: SourceFile): ExportRecords {
 			const name = namespaceDecl.getName();
 			const unknownExported = records.unknown.get(name);
 			if (unknownExported !== undefined) {
-				console.log(
+				log.warning(
```
frankmueller-msft, line 245: Good cleanup (`console.log` → `log.warning`), but this is unrelated to the ESM resolution fix. Might be cleaner as a separate commit for bisectability.

build-tools/packages/build-cli/src/library/typescriptApi.ts
jason-ha: It is not exported outside of the package. The export here just means outside of this file. This package does have a handful of package exports. The are all found in src/index.ts. When externals are impacted we will see api-report/*.md files change. We're all good, but it is the right thing to keep an eye on.

build-tools/packages/build-cli/src/library/commands/generateEntrypoints.ts
jason-ha: I can remove the `i` - I do think strict is safe enough.
But the remainder is intentional. Any of the other patterns that use a .m or .c are most reasonably interpreted as being ESModule or CommonJS, respectively.
A best guess is okay as if there is a problem (which I don't expect), the solution is for caller to be explicit with resolutionConditions.
Since it is a best guess, it would also be okay to be more strict and use `/^\.m[jt]sx?$/` (note the `x?`). I will keep this unless you feel strongly.

build-tools/packages/build-cli/src/library/typescriptApi.ts
jason-ha: Opportunistic and added consistency with the other `log` use now in here. I wouldn't spend any resources on it otherwise. Planning to keep.

build-tools/packages/build-cli/src/library/commands/generateEntrypoints.ts
jason-ha: I had considered an error too. I will add a comment.
It is only might be unstable. It all depends on the context. I could see someone specifying both to deal with some funny case where a package is transitioning (seems rare). footgun should be avoided by having verifications post build.

build-tools/packages/build-cli/src/library/commands/generateEntrypoints.ts
```diff
@@ -186,9 +193,32 @@ export class GenerateEntrypointsCommand extends BaseCommand<
...
+				'warning: both "import" and "require" --resolutionConditions flags given; results may be unstable',
+			);
+		} else if (conditionsPresent === 0) {
+			const inferredImportCondition = inferImportConditionFromFilePath(
+				mainEntrypoint,
+				packageJson,
+			);
+			this.logger.info(`Inferred import resolution condition "${inferredImportCondition}"`);
```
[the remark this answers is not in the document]
jason-ha: Agreed but required test infrastructure has never been funded - continuing debt :/

build-tools/packages/build-cli/src/test/library/commands/generateEntrypoints.test.ts
jason-ha: Cleaned it up to just process --out* arguments and only those that are simple strings.
The use context for readArgValues doesn't require it to understand all of the options.
Now test now demonstrates the rails better.

frankmueller-msft (approved):
LGTM. Correct fix for the ts-morph import conditions gap — `customConditions` is the right lever, inference logic is sound, and the improved error handling for missing export declarations is a good addition. `console.log` → `log.warning` cleanup is appreciated.

One suggestion: `inferImportConditionFromFilePath` would benefit from a few dedicated unit tests (the regex matching + package.json fallback), but not blocking on that.

jason-ha:
Thanks for reviewing @frankmueller-msft!

> One suggestion: `inferImportConditionFromFilePath` would benefit from a few dedicated unit tests (the regex matching + package.json fallback), but not blocking on that.

Yes, to test coverage. There is a big test coverage hole for our build-tools as the best tests need a file system simulator or local setup. We could whitebox test (break encapsulation) that helper like was done with readValueArgs, but not currently seeing enough benefit.

PR #26717, opened 2026-03-12
fix(devcontainer): fix AI-enabled postCreateCommand and load agent aliases system-wide

tylerbutler:
## Summary

- Fixes the AI-enabled devcontainer `postCreateCommand` which wrapped commands in `bash -lc "..."` — scripts are 644 (not executable), so this silently failed. Switches to explicit `bash script.sh` invocations matching the other devcontainer configs.
- Loads agent aliases system-wide via `/etc/profile.d/` and `/etc/bash.bashrc` so they're available in all shell session types (login and interactive non-login) without modifying user dotfiles.
- Adds `copilot-kusto` alias for Kusto MCP integration.

### DIFF
--- a/scripts/codespace-setup/ai-setup.sh
+++ b/scripts/codespace-setup/ai-setup.sh
@@ -8,12 +8,30 @@ SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 #
 # Two hooks are needed because shells load different init files:
 #   - /etc/profile.d/*.sh  → sourced by login shells (ssh, `bash -l`)
-#   - /etc/bash.bashrc     → sourced by interactive non-login shells (VS Code terminal)
+#   - /etc/bash/bashrc.d/  → sourced by interactive non-login shells (VS Code terminal)
 #
-# VS Code / Codespaces terminals are non-login interactive shells, so
-# /etc/profile.d/ alone isn't enough — the /etc/bash.bashrc append covers that case.
-sudo cp "$SCRIPT_DIR/agent-aliases.sh" /etc/profile.d/agent-aliases.sh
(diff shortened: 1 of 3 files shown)


scripts/codespace-setup/ai-setup.sh
```diff
@@ -2,7 +2,18 @@
 set -euo pipefail
 
 SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
-source "$SCRIPT_DIR/agent-aliases.sh"
+
+# Install agent aliases system-wide so they're available in all shell sessions
+# without modifying any user dotfiles (~/.bashrc, ~/.profile, etc.).
```
alexvy86, line 7: On behalf of people with dotfiles repos, thank you 😄 .

scripts/codespace-setup/ai-setup.sh
```diff
@@ -2,7 +2,18 @@
...
+#
+# Two hooks are needed because shells load different init files:
+#   - /etc/profile.d/*.sh  → sourced by login shells (ssh, `bash -l`)
+#   - /etc/bash.bashrc     → sourced by interactive non-login shells (VS Code terminal)
+#
+# VS Code / Codespaces terminals are non-login interactive shells, so
+# /etc/profile.d/ alone isn't enough — the /etc/bash.bashrc append covers that case.
+sudo cp "$SCRIPT_DIR/agent-aliases.sh" /etc/profile.d/agent-aliases.sh
+echo "source /etc/profile.d/agent-aliases.sh" | sudo tee -a /etc/bash.bashrc > /dev/null
```
alexvy86: I'm not that familiar with these folders under `/etc/` but I can find references to a `bashrc.d/` folder. Could we use that so the approach is the same for profile and bashrc?

noencke (approved):

tylerbutler:
Thanks for the review, @noencke!

PR #26721, opened 2026-03-13
feat(experimental-tree): add ISharedTree interface and deprecate SharedTree class

tylerbutler:
## Summary

- Add a new `ISharedTree` interface extracted from the `SharedTree` class as the recommended type for referencing tree instances
- Mark the `SharedTree` class as `@deprecated`, directing users to `ISharedTree`
- Update examples to use `ISharedTree` for type annotations instead of the deprecated class
- Widen `Checkout`, `EagerCheckout`, and `LazyCheckout` to accept `ISharedTree` instead of the concrete `SharedTree` class
- Widen `SharedTree.equals()` and `SharedTree.mergeEditsFrom()` parameter types from `SharedTree` to `ISharedTree`
- Remove deprecated `SharedTree` cast-backs in example code now that `EagerCheckout` accepts `ISharedTree`

This is a **non-breaking** change — no APIs are removed or modified. `ISharedTree` extends `ISharedObject<ISharedTreeEvents>` to maintain full compatibility with the existing type hierarchy.

This prepares for a follow-up breaking change (#26722) that will decouple `ISharedTree` from `ISharedObject`.


examples/version-migration/tree-shim/src/model/inventoryList.ts
```diff
@@ -61,7 +63,7 @@ function migrate(legacyTree: LegacySharedTree, newTree: ITree): void {
 	NewTreeInventoryListController.initializeTree(newTree, initialTree);
 }
 
-const legacyTreeFactory = LegacySharedTree.getFactory();
+const legacyTreeFactory = new LegacySharedTreeFactory(WriteFormat.v0_1_1);
```
tylerbutler, line 66: Tree folks - can you confirm this is correct?

experimental/dds/tree/src/Checkout.ts
```diff
@@ -125,9 +126,10 @@ export abstract class Checkout extends EventEmitterWithErrorHandling<ICheckoutEv
 
 	public disposed: boolean = false;
 
-	protected constructor(tree: SharedTree, currentView: RevisionView, onEditCommitted: EditCommittedHandler) {
+	protected constructor(tree: ISharedTree, currentView: RevisionView, onEditCommitted: EditCommittedHandler) {
 		super((_event, error: unknown) => {
-			this.tree.emit('error', error);
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any
+			(this.tree as any).emit('error', error);
```
tylerbutler, line 132:
```suggestion
			(this.tree as SharedTree).emit('error', error);
```
Is this correct?

examples/version-migration/tree-shim/src/model/inventoryList.ts
CraigMacomber: I think pretty much any write format is fine here since:
1. This only impacts compat of data written by this, and our examples don't need to collab with old code
2. I don't think we actually write any data with this as mainly for reading.
3. I think the whole migration shim this is an example of was abandoned, so its doesn't really matter.

PR #26725, opened 2026-03-14
Add interactive tutorial playground to docs site

jikim-msft:
## Summary
  - Adds a Sandpack-based interactive tutorial system to the Fluid Framework docs site where users can write real SharedTree code in-browser with step-by-step guidance, validation checklists, and inline solutions
  - Includes two tutorial modules: **Dice Roller** (beginner, 5 steps) and **SharedTree Todo App** (intermediate, 5 steps), both teaching schema definition, tree creation, data binding, mutations, and multi-client sync


docs/src/components/playground/data/diceRollerTutorial.ts
```diff
@@ -0,0 +1,501 @@
...
+}
+
+h3 {
+  text-align: center;
+  color: #666;
+  margin-bottom: 16px;
+}
+`;
+
+// --- Tutorial-specific code fragments ---
```
Josmithr: Nit: I recommend using `// #region` / `// #endregion` comment blocks for this sort of region labeling comment. VSCode supports collapsing those regions, which is handy for big files like this.

docs/src/components/playground/TutorialPlayground.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { StepGuide } from "./StepGuide";
+import { modulesById } from "./data/modules";
+import type { ValidationPattern } from "./data/types";
+
+import "@site/src/css/playground.css";
+
+/**
+ * Path to the tutorial module index page (used for "Back to Tutorials" link).
+ */
+const MODULE_INDEX_PATH = "/docs/start/interactive-tutorial/";
```
Josmithr: Nit: would it make sense to update this code to accept this as a variable? Then each if the `dice-roller` and `shared-tree-todo` pages could specify themselves for the "back"/"return" behavior.

I worry that hard-coding this path here may be brittle to future changes / hard to discover.

docs/src/components/playground/TutorialPlayground.tsx
```diff
@@ -0,0 +1,197 @@
...
+}
+
+/**
+ * {@link TutorialPlayground} component props.
+ */
+export interface TutorialPlaygroundProps {
+	/**
+	 * The module to render (e.g. "dice-roller" or "shared-tree-todo").
+	 */
+	moduleId: string;
```
Josmithr: Nit: input properties should generally be `readonly`

docs/src/components/playground/TutorialPlayground.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * Runs validation patterns against code, stripping comments first.
+ * Returns false for any pattern that fails to compile as a regex.
+ */
+function runValidation(code: string, patterns: ValidationPattern[]): boolean[] {
+	const stripped = code
+		.replace(/\/\*[\s\S]*?\*\//g, "")
+		.replace(/\/\/.*$/gm, "");
+	return patterns.map((vp) => {
```
Josmithr: Nit: we should avoid abbreviations like this whenever possible. "validationPattern" would be preferrred.

docs/src/components/playground/TutorialPlayground.tsx
```diff
@@ -0,0 +1,197 @@
...
+ * given module. Module selection is handled at the page level via Docusaurus
+ * routing and the {@link ModuleSelector} component.
+ */
+export function TutorialPlayground({
+	moduleId,
+}: TutorialPlaygroundProps): React.ReactElement {
+	const [currentStepIndex, setCurrentStepIndex] = React.useState(0);
+	const [validationResults, setValidationResults] = React.useState<boolean[]>([]);
+	const [showSolution, setShowSolution] = React.useState(false);
+	const [resetCounter, setResetCounter] = React.useState(0);
```
Josmithr, line 77: Nit: I always recommend documenting state variables to make future maintainers' lives easier 😋

docs/src/css/custom.scss
```diff
@@ -54,6 +54,6 @@
 }
 
 // Applies text-decoration to all links in markdown/MDX content
-main a {
+main a:not(.breadcrumbs__link) {
```
Josmithr, line 59: We should probably update the comment above to explain the exception you added

docs/src/components/playground/TutorialPlayground.tsx
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * Path to the tutorial module index page (used for "Back to Tutorials" link).
+ */
+const MODULE_INDEX_PATH = "/docs/start/interactive-tutorial/";
+
+/**
+ * Runs validation patterns against code, stripping comments first.
+ * Returns false for any pattern that fails to compile as a regex.
+ */
+function runValidation(code: string, patterns: ValidationPattern[]): boolean[] {
```
Josmithr: So the code validation is RegExp based? How exactly does it work? What do we look for? Does sandpack do any sort of code compilation validation or anything?

Josmithr:
High level feedback:
- There is more code here than I was expecting. I had been assuming that sandpack handled more aspects of the view then it seems to. This might not have been a fair assumption on my part, and maybe this amount of code is okay. That said, it does add to our website maintenance costs, so it's something we should consider.
- We really need tests. The degree to which our website is untested is quite bad. I worry about adding large new features like this while we lack test coverage. At a minimum, I think we need to write component level tests for the new React components. But ideally, we'd have top-level end-to-end tests for the website's pages themselves. This isn't something for you to solve, but likely something that the team needs to invest in before we add much in the way of new features.

Questions:
1. How does code validation work? It looks like we have some regular expressions that we try to match against to determine if certain steps in the development process have been reached?
2. How do we evaluate whether or not the code is valid? Does sandpack do that for us?

PR #26733, opened 2026-03-16
build(client): update to build-tools 0.64 release

jason-ha:
pnpm-workspace.yaml, pnpm-lock.yaml:
- update build-tools packages

package.json_s:
- update CommonJS `flub generate entrypoints` calls to keep using `require` condition for resolution. (ESM version's now use `import` condition per build-tools support.)

type tests (per running `flub generate typetests`):
- now uses selected package exports (including self) rather than always using `/internal` or `../../index.ts`.
- header also corrected to reflect actual command.

packages/dds/tree:
- remove legacy.d.ts workaround
- clarify command to generate sources in header comment


alexvy86 (approved):
The bit about the export that type tests use seems like a fix, at least because it matches the declared intention better, but leaves me a bit uneasy. Like we might be mistakenly losing type test coverage somewhere. But since it matches the export declared in package.json, I think move forward and fix that if we encounter instances where we set it to the wrong thing.

jason-ha:
> The bit about the export that type tests use seems like a fix, at least because it matches the declared intention better, but leaves me a bit uneasy. Like we might be mistakenly losing type test coverage somewhere. But since it matches the export declared in package.json, I think move forward and fix that if we encounter instances where we set it to the wrong thing.

See #26440 for type test generation change details. Summary and extra explaination:
Before type test gen would read types from specific package export like `/legacy` to generate content. But it would always use `/internal` for the old import statement and `../../index.ts` for current. The type tests themselves should be unaffected by the tooling change (I didn't see any content change).
But the imports are now much better. The new pattern for example supports packages that don't export `/internal` (leaf packages never have a need for that). Also doesn't limit type tests to use `../../index.ts`. For the protocol compatibility workstream, the plan is to type test an alternate export like `/internal/protocol`.

jzaffiro (approved):

jason-ha:
There will be a conflict with #26522. As that is a release blocker, this PR will be held until that merges.

pnpm-workspace.yaml
```diff
@@ -23,17 +23,10 @@ packages:
...
-  - syncpack-linux-x64-musl
-  - syncpack-linux-arm64
-  - syncpack-linux-arm64-musl
-  - syncpack-darwin-x64
-  - syncpack-darwin-arm64
-  - syncpack-windows-x64
-  - syncpack-windows-arm64
+  # Allow our own packages to be ingested immediately.
+  - "@fluid-tools/*"
+  - "@fluidframework/*"
```
alexvy86: I'm not sure about this; I can see how it could help in public builds, but in internal ones where we pull from ADO feeds that might not even make the packages available because of their own policies, I suspect it could cause the kind of failures that we brought in `minimumReleaseAge` to address.

jason-ha: I needed it locally to start this work. Didn't want to wait 24h.
It has been long enough now and I can definitely remove these additions.
To not have this, either we'll need to update feeds to pull from ADO generally (dev boxes including non-msft authed accounts must have access) OR will need dev boxes to have override for this.
cc: @Abe27342

alexvy86 (approved):

PR #26740, opened 2026-03-16
Add alpha splice method to TreeArrayNode

brrichards:
## Summary

Adds `splice` method to a new alpha `TreeArrayNode` interface, `TreeArrayNodeAlpha`, that follows `Array.prototype.splice` semantics. Supports removing and inserting items in a single opertaion, with optional `deleteCount` and negative `start` indices. Wraps remove + insert in a transaction for hydrated nodes to enable atomic undo/redo. Users opt into using the new api by doing `asAlpha(<TreeArrayNode>)` which then allows them to use splice and also still interact with other non-alpha `TreeArrayNodes`


packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -997,6 +1009,46 @@ abstract class CustomArrayNodeBase<const T extends ImplicitAllowedTypes>
...
+	): TreeNodeFromImplicitAllowedTypes<T>[] {
+		const length = this.length;
+		const actualStart = start < 0 ? Math.max(length + start, 0) : Math.min(start, length);
+		const actualDeleteCount =
+			deleteCount === undefined
+				? length - actualStart
+				: Math.min(Math.max(deleteCount, 0), length - actualStart);
+		const removed: TreeNodeFromImplicitAllowedTypes<T>[] = [];
+		for (let index = actualStart; index < actualStart + actualDeleteCount; index++) {
+			removed.push(this.at(index) ?? fail(0xadc /* Index is out of bounds */));
```
Abe27342: `fail` & `assert` should generally indicate bugs in Fluid when users hit them. This looks like it should an input validation error instead. `UsageError` would be reasonable. Not sure if tree has other utilities built on top of that.

brrichards: I can swap it to `UsageError`, but I believe the input is clamped and bounded by actualStart + deleteCount which is guaranteed to be within range. The `fail()` is mostly used to appease typescript with it's possibly undefined rule. If `this.at(index)` returns undefined here (which it should never) it would be something wrong internally with fluid.

.changeset/plain-tires-melt.md
```diff
@@ -0,0 +1,27 @@
+---
+"@fluidframework/tree": minor
+"__section": feature
+---
+Add splice method on TreeArrayNode
+
+#### Why this change
+
+Adds a `splice` method on `TreeArrayNode` that supports removing and inserting items in a single operation. Returns the removed items as an array. Supports negative `start` indicies (wraps from end). Optional `deleteCount` (omitting removes everything from `start` onward). Requires safe integers values for `start` and `deleteCount`.
```
brrichards:
```suggestion
Adds a `splice` method on `TreeArrayNode` that supports removing and inserting items in a single operation. Returns the removed items as an array. Supports negative `start` indices (wraps from end). Optional `deleteCount` (omitting removes everything from `start` onward). Requires safe integers values for `start` and `deleteCount`.
```

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
brrichards: Changed `fail()` to `oob()`

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -997,6 +1011,42 @@ abstract class CustomArrayNodeBase<const T extends ImplicitAllowedTypes>
...
+			deleteCount === undefined
+				? length - actualStart
+				: Math.min(Math.max(deleteCount, 0), length - actualStart);
+		const removed: TreeNodeFromImplicitAllowedTypes<T>[] = [];
+		for (let index = actualStart; index < actualStart + actualDeleteCount; index++) {
+			removed.push(this.at(index) ?? oob());
+		}
+
+		const innerNode = getInnerNode(this);
+		if (innerNode.isHydrated()) {
```
noencke: nit: rather than checking isHydrated twice (two checks that will always give the same answer), you can save a few lines by doing:

```ts
const innerNode = getInnerNode(this);
const transaction = innerNode.isHydrated() ? innerNode.context.checkout.transaction : undefined;
transaction?.start();
// ...
transaction?.commit();
```

noencke (approved):

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
brrichards: updated to match this

.changeset/plain-tires-melt.md
```diff
@@ -0,0 +1,27 @@
+---
+"@fluidframework/tree": minor
+"__section": feature
+---
+Add splice method on TreeArrayNode
+
+#### Why this change
+
+Adds a `splice` method on `TreeArrayNode` that supports removing and inserting items in a single operation. Returns the removed items as an array. Supports negative `start` indices (wraps from end). Optional `deleteCount` (omitting removes everything from `start` onward). Requires safe integers values for `start` and `deleteCount`.
```
jzaffiro:
```suggestion
Adds a `splice` method on `TreeArrayNode` that supports removing and inserting items in a single operation. Returns the removed items as an array. Supports negative `start` indices (wraps from end). Optional `deleteCount` (omitting removes everything from `start` onward). Requires safe integer values for `start` and `deleteCount`.
```
nit

jzaffiro (approved):
Small nit but approved for docs

.changeset/plain-tires-melt.md
```diff
@@ -0,0 +1,27 @@
+---
+"@fluidframework/tree": minor
```
CraigMacomber, line 3: changes which impact tree's exports also impact the fluid-framework package which reexports them.

.changeset/plain-tires-melt.md
```diff
@@ -0,0 +1,27 @@
...
+"__section": feature
```
CraigMacomber: there is a section for tree stuff which we typically use for tree things, even if they are also features.

.changeset/plain-tires-melt.md
```diff
@@ -0,0 +1,27 @@
...
+---
+Add splice method on TreeArrayNode
+
+#### Why this change
```
CraigMacomber: The text in this section says what the change is, not why we made it. I think the why is to align with the built in Array APIs.

Personally I'd remove this heading, and add a note about this aligning with the JS array splice API to the first sentence below.

.changeset/plain-tires-melt.md
```diff
@@ -0,0 +1,27 @@
...
+
+Adds a `splice` method on `TreeArrayNode` that supports removing and inserting items in a single operation. Returns the removed items as an array. Supports negative `start` indices (wraps from end). Optional `deleteCount` (omitting removes everything from `start` onward). Requires safe integer values for `start` and `deleteCount`.
```
CraigMacomber: When working in markdown, personally I prefer if we line wrap ever sentence. It makes line based tooling like git and our review tools work better (behave more semantically, invalidate comments on unrelated sentences less etc) and doesn't impact the rendered results.

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -340,6 +340,58 @@ describe("ArrayNode", () => {
 				});
 			});
 
+			describe("splice", () => {
```
CraigMacomber, line 346: I think it would be better for these tests to compare the results to the build in array splice, since I assume they are supposed to match. That would result in more concise tests which are more clearly checking for the correct thing

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -340,6 +340,58 @@ describe("ArrayNode", () => {
...
+				it("splice first item", () => {
+					const list = init(schemaType, [0, 1, 2, 3]);
+					const removed = list.splice(0, 1, 4);
+					assert.deepEqual(removed, [0]);
+					assert.equal(list.length, 4);
```
CraigMacomber: This assert has negative value, since if it fails, you would have gotten a better error from the line below with more info, but you don't because this fails first.

CraigMacomber: Same applies to several cases below.

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -340,6 +340,58 @@ describe("ArrayNode", () => {
...
+					const list = init(schemaType, [0, 1, 2, 3]);
+					const removed = list.splice(2, 2, 4, 5, 6);
+					assert.deepEqual([...removed], [2, 3]);
+					assert.equal(list.length, 5);
+					assert.deepEqual([...list], [0, 1, 4, 5, 6]);
+				});
+				it("splice last three with negative start index", () => {
+					const list = init(schemaType, [0, 1, 2, 3]);
+					const removed = list.splice(-3);
+					assert.deepEqual([...removed], [1, 2, 3], `Expected [1,2,3] received ${removed}`);
```
CraigMacomber: providing a custom message here is worse than not, since the default message would be generated from the inputs and thus be more guaranteed to be accurate, and also easier to maintain.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -997,6 +1011,41 @@ abstract class CustomArrayNodeBase<const T extends ImplicitAllowedTypes>
...
+		start: number,
+		deleteCount?: number,
+		...items: Insertable<T>
+	): TreeNodeFromImplicitAllowedTypes<T>[] {
+		if (!Number.isSafeInteger(start)) {
+			throw new UsageError(`Expected a safe integer for start, got ${start}`);
+		}
+		if (deleteCount !== undefined && !Number.isSafeInteger(deleteCount)) {
+			throw new UsageError(`Expected a safe integer for deleteCount, got ${deleteCount}`);
+		}
```
CraigMacomber: Pleas reuse our existing input range validation APIs: they have better messages, and validate more cases.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -163,6 +163,20 @@ export interface TreeArrayNode<
 	 */
 	removeRange(start?: number, end?: number): void;
 
+	/**
+	 * Removes existing item(s) and/or adds new item(s).
+	 * @param start - The index at which to start changing the array. If negative, it is treated as `array.length + start`.
+	 * Must be a safe integer
```
CraigMacomber: I don't think we need to say or enforce this. Either we truncate the value to the length of the array, in which case unsafe integers, and infinites work fine, and we shouldn't really care if someone uses an index larger than 2^52, OR we require the index to be a valid in range value, which will always require it to be a safe integer since arrays largen enough to have unsafe indexes don't work in JS or fit in any existing computer.

From an API design perspective, I'd personally favor starting with the strictest version (only supporting in bounds positive indexes) since we can always relax it later.

packages/dds/tree/api-report/tree.public.api.md
```diff
@@ -479,6 +479,7 @@ export interface TreeArrayNode<TAllowedTypes extends System_Unsafe.ImplicitAllow
     push(...value: readonly (TNew | IterableTreeArrayContent<TNew>)[]): void;
     removeAt(index: number): void;
     removeRange(start?: number, end?: number): void;
+    splice(start: number, deleteCount?: number, ...items: readonly (TNew | IterableTreeArrayContent<TNew>)[]): T[];
```
CraigMacomber: Generally, we don't like adding something directly to the public APi surface without any stabilization period at alpha and beta.

CraigMacomber: I talked with @noencke and his approval on this was for the implementation, not API review. This still needs API review, and I think that API review will want some changes about the strictness of the inputs, and going straight to public skipping alpha and beta.

CraigMacomber (dismissed):
Despite having approval from an API Reviewer, this has not been api reviewed (GitHub doesn't let people approve for a subset of their review groups), so I'm going to use my requesting of changes as a placeholder for the need for an API review. If approving for API review, feel free to dismiss this.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
brrichards: Updated to validate range only

.changeset/plain-tires-melt.md
```diff
@@ -0,0 +1,31 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": tree
+---
+Adds an opt-in `TreeArrayNodeAlpha` with a new `splice` method.
+
+Adds an opt-in alpha `splice` method on `TreeArrayNodeAlpha` that supports removing and inserting items in a single operation to align with JavaScripts Array splice API.
```
CraigMacomber:
```suggestion
Adds a `splice` method on `TreeArrayNodeAlpha` that supports removing and inserting items in a single operation to align with JavaScripts Array splice API.
```

.changeset/plain-tires-melt.md
```diff
@@ -0,0 +1,31 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": tree
+---
+Adds an opt-in `TreeArrayNodeAlpha` with a new `splice` method.
```
CraigMacomber:
```suggestion
Adds TreeArrayNodeAlpha with a new splice method.
```

See https://github.com/microsoft/FluidFramework/wiki/Changesets#formatting

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNodeTypes.ts
```diff
@@ -37,6 +37,30 @@ export interface ArrayNodeCustomizableSchema<
 		>,
 		SimpleArrayNodeSchema<SchemaType.View, TCustomMetadata> {}
 
+/**
+ * A Schema for customizable {@link (TreeArrayNode:interface)}s with additional alpha APIs.
+ * @remarks
+ * Extends {@link ArrayNodeCustomizableSchema} with {@link TreeArrayNodeAlpha} as the node type,
```
noencke:
```suggestion
 * Duplicates {@link ArrayNodeCustomizableSchema} with {@link TreeArrayNodeAlpha} as the node type,
```

packages/dds/tree/src/simple-tree/api/schemaFactoryAlpha.ts
```diff
@@ -518,14 +519,24 @@ export class SchemaFactoryAlpha<
...
+		TCustomMetadata
+	> {
 		return arraySchema(
 			scoped<TScope, TName, Name>(this, name),
 			allowedTypes,
 			true,
 			true,
 			options ?? {},
-		);
+		) as unknown as ArrayNodeCustomizableSchemaAlpha<
```
noencke, line 534: Is there any way to avoid this cast? My guess is no, but curious what @CraigMacomber  thinks

noencke (approved):
Implementation looks good to me

Abe27342 (approved):
Did not review the tree code.

Approving for merge-tree, which doesn't actually have any changes we care about in this PR. Talked with Brennan offline and this is probably just a sticky GH check.

Also approving for API council as the PR contains only alpha API changes. I can see that this is also a sticky check from a previous revision that touched more broadly visible API files.

.changeset/plain-tires-melt.md
```diff
@@ -0,0 +1,41 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": tree
+---
+Adds TreeArrayNodeAlpha with a new splice method.
```
jzaffiro:
```suggestion
Adds TreeArrayNodeAlpha with a new splice method
```

.changeset/plain-tires-melt.md
```diff
@@ -0,0 +1,41 @@
...
+
+Adds a `splice` method on `TreeArrayNodeAlpha` that supports removing and inserting items in a single operation to align with JavaScripts Array splice API.
```
jzaffiro:
```suggestion
Adds a `splice` method on `TreeArrayNodeAlpha` that supports removing and inserting items in a single operation to align with JavaScript's Array splice API.
```

jzaffiro (approved):

PR #26748, opened 2026-03-16
Consolidate GC logs on boot into ContainerLoadStats event

markfields:
## Description

We're trying to reduce logging overhead/noise from high event volume for consumers. These two events occur on boot and the info can be consolidated into the canonical boot event "ContainerLoadStats":

* "GCFeatureMatrix"
* "GarbageCollectorLoaded"

### DIFF
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -219,7 +219,6 @@ import {
 	type IGCRuntimeOptions,
 	type IGCStats,
 	type IGarbageCollector,
-	gcGenerationOptionName,
 	type GarbageCollectionMessage,
 	type IGarbageCollectionRuntime,
 } from "./gc/index.js";
@@ -2084,11 +2083,8 @@ export class ContainerRuntime
 			summaryNumber: loadSummaryNumber,
 			summaryFormatVersion: metadata?.summaryFormatVersion,
 			disableIsolatedChannels: metadata?.disableIsolatedChannels,
-			gcDetails: JSON.stringify({
-				gcFeature: metadata?.gcFeature,
-				gcFeatureMatrix: JSON.stringify(metadata?.gcFeatureMatrix),
-				gcGeneration: runtimeOptions.gcOptions[gcGenerationOptionName],
-			}),
+			// This is useful even for interactive clients since they track unreferenced nodes and log errors.
+			gcConfigs: this.garbageCollector.serializedConfigs,
 			options: JSON.stringify(runtimeOptions),
 			idCompressorModeMetadata: metadata?.documentSchema?.runtime?.idCompressorMode,
 			idCompressorMode: this.sessionSchema.idCompressorMode,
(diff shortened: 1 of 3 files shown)


packages/runtime/container-runtime/src/gc/garbageCollection.ts
```diff
@@ -182,6 +184,7 @@ export class GarbageCollector implements IGarbageCollector {
 		});
 
 		this.configs = generateGCConfigs(this.mc, createParams);
+		this.serializedConfigs = JSON.stringify(this.configs);
```
agarwal-navin: `gcOptions` and the `GCFeatureMatrix` aren't part of configs. Don't we want those?

markfields: They're in `options` prop (next line)

packages/runtime/container-runtime/src/gc/garbageCollection.ts
```diff
@@ -100,6 +100,8 @@ export class GarbageCollector implements IGarbageCollector {
 
 	private readonly configs: IGarbageCollectorConfigs;
 
+	public readonly serializedConfigs: string;
```
agarwal-navin: Might be simpler to expose the configs and have it serialized at logging time in Container Runtime? That way we don't have to track the same config twice here.

markfields: It's to prevent tampering with the configs.  I didn't like exposing the config object directly

markfields: I guess downside of this is there's a bit of persistent memory bloat.  Maybe not worth it, I guess I don't see someone having a reason to modify it or accidentally doing it.  What do you think?

agarwal-navin: Make it a getter and readonly?

markfields: Yeah all the props are already readonly - it's pedantic but there's some nested objects in there which could be mucked with.

How about I switch `serializedConfig` to a getter that does the JSON.stringify, rather than storing the string.  Since we know it's only called once.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -2093,7 +2083,8 @@ export class ContainerRuntime
 			summaryNumber: loadSummaryNumber,
 			summaryFormatVersion: metadata?.summaryFormatVersion,
 			disableIsolatedChannels: metadata?.disableIsolatedChannels,
-			gcVersion: metadata?.gcFeature,
+			// This is useful even for interactive clients since they track unreferenced nodes and log errors.
+			gcConfigs: this.garbageCollector.serializedConfigs,
 			options: JSON.stringify(runtimeOptions),
```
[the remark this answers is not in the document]
markfields: logging is not a stable API

packages/runtime/container-runtime/src/gc/garbageCollection.ts
```diff
@@ -100,6 +100,10 @@ export class GarbageCollector implements IGarbageCollector {
 
 	private readonly configs: IGarbageCollectorConfigs;
 
+	public get serializedConfigs(): string {
+		return JSON.stringify(this.configs);
```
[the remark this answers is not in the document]
markfields: This was an intentional tradeoff given there's only 1 call-site (and others are not expected).  This reduces memory footprint of redundant config string.

agarwal-navin (approved):

agarwal-navin (approved):

PR #26749, opened 2026-03-16
Optimize deep compare for formatted text

brrichards:
## Summary

Optimize `buildDeltaFromTree` with cursor-based run detection. `buildDeltaFromTree` iterated through every character individually via `charactersWithFormatting()`, calling `formatToQuillAttributes` and `JSON.stringify` per character for format comparison. To help with operations like this two new cursor-based API's have been added to `FormattedTextAsTree.Tree`:

- getuniformRun(startIndex, endIndex?) - returns length of consecutive characters with identical formating using raw cursor comparison. Handles StringLineAtom by always returning 1 (Quill only supports one per line by default). 
- getString(startIndex, endIndex?) - returns a substring extracted via cursor, handling both `StringTextAtom` and `StringLineAtom` content types.

`buildDeltaFromTree` now walks the tree in runs rather than character-by-character, calling `formatToQuillAttributes()` once per run instead of once per character.

As part of this work a general tree comparison utility has been added to `packages/dds/tree/src/core/tree/cursor.ts`


packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -251,6 +265,113 @@ class StringArray extends sf.array("StringArray", [() => FormattedTextAsTree.Str
...
+	public textString(startIndex: number, length: number): string {
+		if (length <= 0 || startIndex >= this.length) return "";
+		const arrayLength = this.length;
+		return this.withBorrowedSequenceCursor((cursor) => {
+			const limit = Math.min(length, arrayLength - startIndex);
+			let result = "";
+
+			cursor.enterNode(startIndex);
+			for (let index = 0; index < limit; index++) {
+				if (index > 0 && !cursor.nextNode()) {
```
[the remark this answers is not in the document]
brrichards: I have to assume that `.join()` is slower for small runs (a sentence, or a paragraph) compared to `+=`. Maybe the overhead cost is worth it?

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -251,6 +265,113 @@ class StringArray extends sf.array("StringArray", [() => FormattedTextAsTree.Str
 	public fullString(): string {
 		return this.charactersCopy().join("");
 	}
+	public getUniformRun(startIndex: number, maxLength: number = this.length): number {
+		if (maxLength <= 0 || startIndex >= this.length) return 0;
+		const arrayLength = this.length;
+		return this.withBorrowedSequenceCursor((cursor) => {
+			cursor.enterNode(startIndex);
```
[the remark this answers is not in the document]
brrichards: clean errors are already thrown for index out of bounds and this is already consistent with other cursor based methods.

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -251,6 +265,113 @@ class StringArray extends sf.array("StringArray", [() => FormattedTextAsTree.Str
...
+			}
+
+			const formatValues: Value[] = [];
+			cursor.enterField(formatKey);
+			cursor.enterNode(0);
+			for (let inField = cursor.firstField(); inField; inField = cursor.nextField()) {
+				cursor.enterNode(0);
+				formatValues.push(cursor.value);
+				cursor.exitNode();
+			}
```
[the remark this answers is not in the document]
brrichards: An array has less overhead than a map for this. I believe CharacterFormat nodes all have the same schema in the same deterministic field order

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -251,6 +265,113 @@ class StringArray extends sf.array("StringArray", [() => FormattedTextAsTree.Str
...
+			cursor.exitNode();
+			cursor.exitField();
+
+			// Line atoms contain only a single value, so the uniform run length is always 1.
+			if (contentType === FormattedTextAsTree.StringLineAtom.identifier) {
+				cursor.exitNode();
+				return 1;
+			}
+
+			const formatValues: Value[] = [];
```
CraigMacomber: Comparing trees is a generally useful thing to do.

We should factor out the tree comparison logic such that it would be trivial to expose a customer facing API for comparing trees efficiently.

This will also make unit testing this easier as we can unit test the comparison logic easily.

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -251,6 +265,113 @@ class StringArray extends sf.array("StringArray", [() => FormattedTextAsTree.Str
...
+			cursor.enterField(formatKey);
+			cursor.enterNode(0);
+			for (let inField = cursor.firstField(); inField; inField = cursor.nextField()) {
```
CraigMacomber: I haven't gone through how you are doing this comparison in detail, but I think you are making some flawed assumptions.

Mainly in this case cursors do not make any guarantees about the order of fields.

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -251,6 +265,113 @@ class StringArray extends sf.array("StringArray", [() => FormattedTextAsTree.Str
...
+				cursor.enterNode(0);
+				formatValues.push(cursor.value);
```
CraigMacomber: I was thinking we would provide a general tree comparison utility: one that would work even with future configurations of formatted text where the formatting subtree is user controlled and might contain things like optional fields, and non-leaf values.

Rather than storing a flat array of values to compare against, the comparator could generate a flat array of commands to process (perhaps in the form of an array of closures)
Each command for nodes could be one of:
1. expect value: check for a specific value at the current node.
2. expect fields: expect a specific number of fields, and enter each one by name and apply the corresponding field compare operation to it.
And for fields:
1. expect nodes: array of node expectations (from above) to apply to each node in the field. Validates the length of the field matches array length.

This would be a little less efficient than what you have, but tolerate arbitrary field order and work on arbitrary trees.
I think the overhead would be small enough that it would be nearly as fast, but provide a much more general tool that will not be a problem if we allow for other format trees in the future, or cursor implementations show fields in other orders.

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -251,6 +265,113 @@ class StringArray extends sf.array("StringArray", [() => FormattedTextAsTree.Str
 	public fullString(): string {
 		return this.charactersCopy().join("");
 	}
+	public getUniformRun(startIndex: number, maxLength: number = this.length): number {
+		if (maxLength <= 0 || startIndex >= this.length) return 0;
```
CraigMacomber: Given the use case of this API, I think it might make more sense to spec anything that would not return at least 1 as an usage error rather than returning 0. I can't imagine a use case where someone would prefer 0 to a usage error, as 0 would cause infinite loops in most use cases for this function if they got the arguments wrong.

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -251,6 +265,113 @@ class StringArray extends sf.array("StringArray", [() => FormattedTextAsTree.Str
...
+			cursor.enterNode(startIndex);
+
+			// Line atoms are handled separately from text atoms
+			cursor.enterField(EmptyKey);
+			cursor.enterNode(0);
+			const contentType = cursor.type;
+			cursor.exitNode();
+			cursor.exitField();
+
+			// Line atoms contain only a single value, so the uniform run length is always 1.
```
CraigMacomber: This API should tell us how many atoms have the same formatting in a row. Both lines and string atoms have formatting, so this choice seems arbitrary to me.

I'd like to design this in a way that is less coupled to the exact set of atom types as we may want to allow that to be customized in the future.

At the very least this function should have a doc comment explaining what it does, including any atom specific polices, but I think making it less tied to specific atom types would be good. One way to do that would be to take in an optional set of atom types to treat as singletons (runs of 1) and the current use case could pass in the StringLineAtom.identifier.

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -427,6 +548,20 @@ export namespace FormattedTextAsTree {
...
+		 * @param maxLength - The maximum length of the run. Defaults to the length of the text.
+		 */
+		getUniformRun(startIndex: number, maxLength?: number): number;
+
+		/**
+		 * Returns a substring of the text starting at `startIndex` and spanning `length` characters.
+		 * @param startIndex - The starting index of the substring.
+		 * @param length - The number of characters to include in the substring.
+		 */
+		textString(startIndex: number, length: number): string;
```
CraigMacomber: I think "substring" would make more sense than "textString".

A better API might be to instead modify the existing "fullString" API, renaming it to something like "getString" and giving it optional start and end indexes (to match removeRange and formatRange)

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -427,6 +548,20 @@ export namespace FormattedTextAsTree {
 			endIndex: number | undefined,
 			format: Partial<CharacterFormat>,
 		): void;
+
+		/**
+		 * Returns the length of the run of characters starting at `startIndex` which have the same formatting, up to a maximum of `maxLength`.
+		 * @param startIndex - The starting index of the run.
+		 * @param maxLength - The maximum length of the run. Defaults to the length of the text.
+		 */
+		getUniformRun(startIndex: number, maxLength?: number): number;
```
CraigMacomber: Our other range APIs are in terms of start and end indexes, while this one takes a start index and a max length.

While semantically I think taking a length here makes sense, taking two numbers like the others and having the semantics be different seems pretty error prone.

It might be worth making this take an optional end index like the others, defaulting to the end of the string.

packages/dds/tree/src/text/textDomainFormatted.ts
CraigMacomber: Another option would be to have this just split if the atom type ever changes. That might be the simplest option for now.

packages/dds/tree/src/text/textDomainFormatted.ts
brrichards: fullString would probably be the better option. The problem with substring is our eslint confuses the substring method with a normal array method and gives an error for prefering slice.

packages/dds/tree/src/text/textDomainFormatted.ts
brrichards: Created a general cursor tree comparison logic in `packages/dds/tree/src/core/tree/cursor.ts`

packages/dds/tree/src/text/textDomainFormatted.ts
brrichards: You were right on this. The comparison now follows your suggestion method of a general tree comparison utility.

packages/dds/tree/src/text/textDomainFormatted.ts
brrichards: went the route of splitting on atom type change

packages/dds/tree/src/text/textDomainFormatted.ts
brrichards: update the API to be `endIndex`

packages/dds/tree/src/text/textDomainFormatted.ts
brrichards: ended up using `getString()`.

packages/dds/tree/src/text/textDomainFormatted.ts
brrichards: This worked great for a general comparison utility. Also the performance difference with what I had before was negligible which is nice.

packages/dds/tree/src/core/tree/cursor.ts
```diff
@@ -484,3 +484,78 @@ export function inCursorNode<T, TCursor extends ITreeCursor = ITreeCursor>(
 	cursor.exitNode();
 	return result;
 }
+export type NodeComparator = (cursor: ITreeCursorSynchronous) => boolean;
+
+type FieldComparator = (cursor: ITreeCursorSynchronous) => boolean;
+
+function buildFieldComparator(
+	expectedLength: number,
```
CraigMacomber: I think you can just use the nodeComparators's length, which would simplify the logic a bit as well since you would not need to handle out of bounds on it.

packages/dds/tree/src/core/tree/cursor.ts
```diff
@@ -484,3 +484,78 @@ export function inCursorNode<T, TCursor extends ITreeCursor = ITreeCursor>(
 	cursor.exitNode();
 	return result;
 }
+export type NodeComparator = (cursor: ITreeCursorSynchronous) => boolean;
+
+type FieldComparator = (cursor: ITreeCursorSynchronous) => boolean;
+
+function buildFieldComparator(
```
CraigMacomber: It would be good to document at least one of FieldComparator or buildFieldComparator, ideally both.

packages/dds/tree/src/core/tree/cursor.ts
```diff
@@ -484,3 +484,78 @@ export function inCursorNode<T, TCursor extends ITreeCursor = ITreeCursor>(
 	cursor.exitNode();
 	return result;
 }
+export type NodeComparator = (cursor: ITreeCursorSynchronous) => boolean;
```
CraigMacomber: There are enough things in this new code that a new file would make sense. Maybe cursorComparator.ts in feature-libraries to line up with where you put the tests?

packages/dds/tree/src/core/tree/cursor.ts
```diff
@@ -484,3 +484,78 @@ export function inCursorNode<T, TCursor extends ITreeCursor = ITreeCursor>(
...
+
+function buildFieldComparator(
+	expectedLength: number,
+	nodeComparators: NodeComparator[],
+): FieldComparator {
+	return (cursor: ITreeCursorSynchronous): boolean => {
+		if (cursor.getFieldLength() !== expectedLength) {
+			return false;
+		}
+		for (let index = 0; index < expectedLength; index++) {
```
CraigMacomber: its more efficient to use nextNode the enter node and exit node. Look at forEachNode as an example for how to do that cleanly. You can use cursor.fieldIndex for the current index inside the field to index into nodeComparators.

packages/dds/tree/src/core/tree/cursor.ts
```diff
@@ -484,3 +484,78 @@ export function inCursorNode<T, TCursor extends ITreeCursor = ITreeCursor>(
...
+		if (other.value !== expectedValue) {
+			return false;
+		}
+		for (const { key, compare } of fieldComparators) {
+			other.enterField(key);
+			if (compare(other) !== true) {
+				other.exitField();
+				return false;
+			}
+			other.exitField();
```
CraigMacomber: Using the `inCursorField` utility here would simplify this code slightly:
```suggestion
			if (!inCursorField(cursor, key, () => compare(other))) {
				return false;
			}
```

packages/dds/tree/src/core/tree/cursor.ts
```diff
@@ -484,3 +484,78 @@ export function inCursorNode<T, TCursor extends ITreeCursor = ITreeCursor>(
...
+	for (let inField = cursor.firstField(); inField; inField = cursor.nextField()) {
+		const key = cursor.getFieldKey();
+		const length = cursor.getFieldLength();
+		const nodeComparators: NodeComparator[] = [];
+
+		for (let index = 0; index < length; index++) {
+			cursor.enterNode(index);
+			nodeComparators.push(buildNodeComparator(cursor));
+			cursor.exitNode();
+		}
```
CraigMacomber: Use mapCursorField

packages/dds/tree/src/core/tree/cursor.ts
```diff
@@ -484,3 +484,78 @@ export function inCursorNode<T, TCursor extends ITreeCursor = ITreeCursor>(
...
+ * Missing fields in the compared node will cause a mismatch.
+ *
+ * The cursor must be in Nodes mode. After this call. the cursor is restored to its original position.
+ *
+ */
+export function buildNodeComparator(cursor: ITreeCursorSynchronous): NodeComparator {
+	const expectedValue: Value = cursor.value;
+	const fieldComparators: { key: FieldKey; compare: FieldComparator }[] = [];
+
+	for (let inField = cursor.firstField(); inField; inField = cursor.nextField()) {
```
CraigMacomber: Use mapCursorFields

packages/dds/tree/src/core/tree/cursor.ts
```diff
@@ -484,3 +484,78 @@ export function inCursorNode<T, TCursor extends ITreeCursor = ITreeCursor>(
...
+ *
+ * @remarks
+ * Fields are compared by key (not iteration order), so this is safe regardless of cursor field ordering.
+ * Missing fields in the compared node will cause a mismatch.
+ *
+ * The cursor must be in Nodes mode. After this call. the cursor is restored to its original position.
+ *
+ */
+export function buildNodeComparator(cursor: ITreeCursorSynchronous): NodeComparator {
+	const expectedValue: Value = cursor.value;
```
CraigMacomber: nodes can only ever have a fields OR a value.

Because of this, if you want, you could make an early out for the case where expectedValue !== undefined, which just returns `(other: ITreeCursorSynchronous) => Object.is(other.value, expectedValue)`

Note that comparing values should use Object.is incase there is a NaN since we want a node to be equal to itself and any structurally identical tree.

It should be impossible to get a NaN into the tree with the public APIs (we ban them since they don't Json serialize), so its not too important, but its the preferred pattern in my opinion when you want reflexive equality.

It does also impact -0, making it not compare equal to 0, which is good in my opinion here as they are different. Again, we ban -0 from making into trees though the public API and normalize them to 0 since thats what json does,

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -251,6 +265,103 @@ class StringArray extends sf.array("StringArray", [() => FormattedTextAsTree.Str
...
+				}
+				cursor.exitNode();
+				cursor.exitField();
+			}
+			cursor.exitNode();
+			return result;
+		});
+	}
+	public getUniformRun(startIndex: number, endIndex: number = this.length): number {
+		if (startIndex < 0 || startIndex >= this.length) {
```
CraigMacomber: there are some nice array index validation utils used in ArrayNode: use those to avoid duplicating the logic and for slightly nicer error text.

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -251,6 +265,103 @@ class StringArray extends sf.array("StringArray", [() => FormattedTextAsTree.Str
...
+			cursor.enterNode(0);
+			const formatComparator = buildNodeComparator(cursor);
+			cursor.exitNode();
+			cursor.exitField();
+
+			let runLength = 1;
+			const limit = Math.min(endIndex, arrayLength) - startIndex;
+
+			while (runLength < limit && cursor.nextNode()) {
+				cursor.enterField(EmptyKey);
```
CraigMacomber, line 337: A comment here to match below would be nice:
```suggestion
				// Compare atom type
				cursor.enterField(EmptyKey);
```

packages/dds/tree/src/core/tree/cursor.ts
brrichards: Moved new code to feature-libraries

packages/dds/tree/src/text/textDomainFormatted.ts
brrichards: I didn't see an index validation that doesn't include start !== end so I currently still have a check for that right after using a validation of range from our utils.

packages/dds/tree/src/core/tree/cursor.ts
CraigMacomber: I think you can still replace the outer loop with this

packages/dds/tree/src/text/textDomainFormatted.ts
CraigMacomber: I guess that implies zero length ranges are valid in our other APIs, but not here, which is inconsistent. Its worth considering making them consistent, but I think what you have done makes the most sense, and its better to error on the side of being too strict with the API's inputs as we can relax it later as a non breaking change.

packages/dds/tree/src/core/tree/cursor.ts
brrichards: replaced with mapcursorFields. My mistake on missing this comment beforehand

CraigMacomber (approved):

PR #26752, opened 2026-03-17
build(devcontainer): add repoverlay, agency install, and ff-oce overlay

tylerbutler:
## Summary

- Adds optional `agency` and `repoverlay` installation to the Dockerfile via build args, with user-local bin paths added to `PATH`.
- Adds the `ff-oce` repoverlay overlay: a Claude agent and Kusto skill for the Fluid Framework Client OCE rotation, with a comprehensive query reference covering Office Fluid telemetry tables, correlation IDs, and investigation patterns.
- Refactors shell alias setup to support both bash and zsh by adding a shell guard to the aliases script itself and replacing the bash-only `/etc/profile.d/` wrapper with a direct symlink.
- Adds `playwright-setup.sh` guard that skips gracefully when `playwright-cli` is not installed.
- Fixes `nvm` sourcing under `set -u` in `post-create.sh`.
- Enables SSH and ADO codespaces auth in the ai-agent devcontainer.


.devcontainer/ai-agent/devcontainer.json
```diff
@@ -16,6 +19,7 @@
 				"dbaeumer.vscode-eslint",
 				"esbenp.prettier-vscode",
 				"MAI-EngineeringSystems.mai-ai-telemetry",
+				"ms-codespaces-tools.ado-codespaces-auth",
```
alexvy86, line 22: From its docs: 

> This extension is not recommended to be installed by itself. You should instead use the [external-repository](https://github.com/microsoft/codespace-features/tree/main/src/external-repository) and [artifacts-helper](https://github.com/microsoft/codespace-features/tree/main/src/artifacts-helper) devcontainer features which will ensure this extension is preinstalled on your Codespace with proper configuration.

Worth consideration? Not sure the `external-repository` one matters in this case, but I think I've seen `artifacts-helper` in other places.

.repoverlay/library/ff-oce/.claude/skills/ff-oce-kusto/SKILL.md
```diff
@@ -0,0 +1,52 @@
...
+
+Key deployment ring field: `Loop_Audience` (FluidRuntime tables), `Release_AudienceGroup` (OWH/QoS tables)
+
+## Using the Full Reference
+
+Before writing any non-trivial Kusto query, read the full reference file:
+
+**`references/kusto-query-reference.md`**
+
+This 1,400-line reference contains:
```
alexvy86: Nit: agents like to provide details like this that aren't particularly helpful and prone to become outdated.
```suggestion
This reference contains:
```

.repoverlay/library/ff-oce/.claude/skills/ff-oce-kusto/SKILL.md
alexvy86, line 1: Not necessarily for this PR, but we'll probably want another reference doc for the Office Fluid Test database which is relatively similar, but also has tables that only exist there, e.g. the ones for our pipeline telemetry, perf benchmarks telemetry, bundle size telemetry. And in general, is the right database to query to investigate issues with our own engineering systems.

alexvy86 (commented):
Didn't go through the `ff-oce.md` or `kusto-query-reference.md` file but those are probably going to see iteration anyway.

.repoverlay/library/ff-oce/.claude/skills/ff-oce-kusto/SKILL.md
tylerbutler: @noencke Just FYI. I've accepted some of these changes, but I haven't yet integrated the latest version you sent me.

.repoverlay/library/ff-oce/.claude/skills/ff-oce-kusto/SKILL.md
tylerbutler: @noencke Feedback to consider.

.devcontainer/ai-agent/devcontainer.json
tylerbutler: It's not clear to me how any of this fits together. :) The extension was suggested to unblock agency's install - I'll add the other features as well.

.repoverlay/library/ff-oce/.claude/skills/ff-oce-kusto/SKILL.md
noencke: Agreed - ideally our kusto query reference files are factored into their most atomic chunks, and then various kusto-related skills can reference various query files as needed. Since some might overlap, and want to share, but not share everything - if that makes sense.

noencke (approved):

noencke (approved):

PR #26759, opened 2026-03-17
Sanitize user display names in dice roller example 

jzaffiro:
Instead of rendering the display names as HTML (which leaves the application vulnerable to injection attacks), render them as plain text.


steffenloesch (approved):
Looks good to me. Other than the div being rendered in the UI, I don't see any other usage, so change the shape like this shouldn't break anything.

PR #26763, opened 2026-03-18
Adds undoing of noChange constraint to treeCheckout

jumyhre:
This PR is a targeted fix for the noChange constraint. The problem we're fixing is that the noChange constraint was being triggered in cases where there was effectively no real change. The relevant scenario in Tablero was: simply insert 2 or more columns, then try to undo the inserts. Under the current system, only the first undo would be processed, and the others would claim the no change constraint was violated

The fix for this is a targeted solution: We've added logic to treeCheckout that, when a revert is processed, creates a diff of the old and new trees and reviews it for actual new changes.  If we don't detect any net changes to the tree, then we invoke the new `unviolateNoChangeConstraint` function to reset the noChange constraint when the result of undoing is no effective net change


packages/dds/tree/src/core/change-family/changeFamily.ts
```diff
@@ -16,6 +16,15 @@ export interface ChangeFamily<TEditor extends ChangeFamilyEditor, TChange> {
...
 	): TEditor;
 
+	/**
+	 * Produces a change equivalent to `change`, but with any violated "no change" constraint cleared.
+	 *
+	 * @remarks
+	 * This is primarily intended for undo/redo flows where an inverse is rebased over newer commits which net to no effect.
+	 * In those cases, the document state is effectively unchanged, and so a no-change constraint should not remain violated.
+	 */
+	unviolateNoChangeConstraint(change: TChange): TChange;
```
Josmithr: I don't know this code well, so please take this with a grain of salt.

Sanity check: I don't see any other constraint logic in this type. Is this the right place for this?

If so, does it need to know about specific constraint types (e.g. `NoChange`)? Or could it be more general and take in the kind of constraint being unmarked?

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
```diff
@@ -125,6 +125,21 @@ export class ModularChangeFamily
...
+	public unviolateNoChangeConstraint(change: ModularChangeset): ModularChangeset {
+		const constraint = change.noChangeConstraint;
+		if (constraint === undefined || constraint.violated === false) {
+			return change;
+		}
+
+		const updatedViolationCount = Math.max(0, (change.constraintViolationCount ?? 0) - 1);
+		return makeModularChangeset({
+			...change,
+			maxId: (change.maxId ?? -1) as number,
```
Josmithr: What does this do? Would someone more familiar with this code understand this easily? Also, why do we need a cast? (If any of this would be non-obvious to someone with more familiarity here, we should probably add a comment)

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1566,3 +1585,42 @@ function isSerializedChange(value: unknown): value is SerializedChange {
 		change.change !== undefined
 	);
 }
+
+function sharedTreeChangeHasNetEffect(change: SharedTreeChange): boolean {
```
Josmithr: Nit: while these functions are private, I think they are non-trivial. It would probably be good to add documentation to them.

packages/dds/tree/src/test/shared-tree/editing.spec.ts
```diff
@@ -3439,6 +3439,70 @@ describe("Editing", () => {
...
+
+				branch.transaction.start();
+				branch.editor.addNoChangeConstraintOnRevert();
+				branch.editor.sequenceField(rootField).insert(1, chunkFromJsonTrees(["W"]));
+				branch.transaction.commit();
+
+				undoStack.pop()?.revert();
+				undoStack.pop()?.revert();
+				undoStack.pop()?.revert();
+
```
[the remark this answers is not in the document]
Josmithr: Presumably not a problem because the subsequent `expectJsonTree` call would get different information if any of these calls didn't happen?

packages/dds/tree/src/test/shared-tree/editing.spec.ts
```diff
@@ -3439,6 +3439,88 @@ describe("Editing", () => {
...
+				branch.transaction.commit();
+
+				undoStack.pop()?.revert();
+				undoStack.pop()?.revert();
+
+				expectJsonTree(branch, [{ "X": 1, "B": 2, "C": 3 }]);
+				unsubscribe();
+			});
 		});
 	});
```
Josmithr, line 3568: Could we add a multi-change undo/redo test (or a couple) like this to the table schema tests as well? I'd love to have those locked in as regression tests.

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
jumyhre: Yeah this is a bit weird to read, but to break it down:

This is boilerplate:
`...change` gives us the original changeset
`maxId:` is being respecified here due to typescript typing weirdness. in theory this should work: `makeModularChangeset({...change})`, but in practice typescript has typing esoteric grievances that are worked around with providing the maxId, or default -1 (which you can find used elsewhere as a default in the modular change family code)

Then the important stuff:
`updatedViolationCount` is the remaining number of constraint violations after we've removed this instance from the total count, and
`noChangeConstraint` is returned to a default, unviolated form

jumyhre: also the cast as number is unfortunately needed here

packages/dds/tree/src/shared-tree/sharedTreeChangeFamily.ts
```diff
@@ -94,6 +94,24 @@ export class SharedTreeChangeFamily
...
+				);
+				if (updated !== inner.innerChange) {
+					changed = true;
+					return { type: "data", innerChange: updated };
+				}
+			}
+
+			return inner;
+		});
+		return changed ? { changes: newChanges } : change;
```
alex-pardes: This function would be a lot cleaner if we always returned `{ changes: newChanges }`.

packages/dds/tree/src/core/change-family/changeFamily.ts
alex-pardes: I would move this into `ChangeRebaser`, but it is still unfortunate to have to add a kludgy method to a high-level interface. I do think we should add some comments here about this being a kludge.

We want the rebaser to treat no-change constraints as violated only if the sequence of changes rebased over leave the document in some meaningfully different state. Because implementing that rebasing logic is more complex, this PR is adding a hack to manually get that behavior in a special case. We fundamentally want a special rebasing behavior for no-change constraint in particular, so I don't think there's any good way to generalize.

The only other architectural approach I see would be to instead pass a flag through the rebase method, indicating that we want to ignore no-change constraint violations (or even just ignore violations due to transient changes). That's still a kludgy parameter at the same interface level, but after looking through the changes in this PR, I'm starting to think that approach might be better, as it would be more contained.

packages/dds/tree/src/shared-tree/treeCheckout.ts
alex-pardes: Agreed, especially since there are multiple interpretations of an "effect."

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
Josmithr: Makes sense. And I was specifically concerned about this line:
> maxId: (change.maxId ?? -1) as number,

Seems probably worth a quick comment

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1566,3 +1585,42 @@ function isSerializedChange(value: unknown): value is SerializedChange {
...
+		}
+		const delta = intoDelta(tagChange(inner.innerChange, undefined));
+		if (deltaFieldMapHasNetEffect(delta.fields)) {
+			return true;
+		}
+	}
+	return false;
+}
+
+function deltaFieldMapHasNetEffect(fields: DeltaFieldMap | undefined): boolean {
```
alex-pardes: I think it makes sense to move these functions into `deltaUtil.ts`.
Also, we can drop the word "net" from the names of these functions

packages/dds/tree/src/test/shared-tree/editing.spec.ts
```diff
@@ -3439,6 +3439,88 @@ describe("Editing", () => {
 				expectJsonTree(branch, ["A", "B"]);
 				unsubscribe();
 			});
+
```
Josmithr, line 3442: The changes in this PR should also work in cases like the following right?
1. Insert A (with no change constraint on revert)
2. Insert B
3. Delete B
4. Revert Insert A

Probably worth adding tests for those kinds of scenarios too (rather than just undo/redo scenarios)

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1567,3 +1585,42 @@ function isSerializedChange(value: unknown): value is SerializedChange {
...
 }
+
+function sharedTreeChangeHasNetEffect(change: SharedTreeChange): boolean {
+	for (const inner of change.changes) {
+		if (inner.type === "schema") {
+			return true;
+		}
+		const delta = intoDelta(tagChange(inner.innerChange, undefined));
+		if (deltaFieldMapHasNetEffect(delta.fields)) {
+			return true;
```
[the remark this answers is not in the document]
jumyhre: nope, i dont think this is right

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1567,3 +1585,42 @@ function isSerializedChange(value: unknown): value is SerializedChange {
...
+		}
+	}
+	return false;
+}
+
```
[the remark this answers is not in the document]
jumyhre: nope, we're satisfied with the approach done by deltaFieldMapHasNetEffect

packages/dds/tree/src/test/shared-tree/editing.spec.ts
jumyhre: Done - added this test to editing.spec.ts. I didnt make an equivalent for table, just because we don't gave fine-grained control of the transaction and applying the constraint

packages/dds/tree/src/shared-tree/treeCheckout.ts
jumyhre: Good calls. I've moved their implementation and replaced the "NetEffect" to "VisibleChange" to address other feedback in the PR

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
jumyhre: Closing, this no longer exists post-refactor

packages/dds/tree/src/core/change-family/changeFamily.ts
jumyhre: I've removed this unviolate function in favor of the approach Alex suggested above

packages/dds/tree/src/test/shared-tree/editing.spec.ts
jumyhre: Yep - I've added a similar test that does this in tableSchema's tests

packages/dds/tree/src/shared-tree/treeCheckout.ts
jumyhre: Added comments, and changed the wording on the func a bit

packages/dds/tree/src/shared-tree/sharedTreeChangeFamily.ts
jumyhre: Closing, this no longer exists post-refactor

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1710,3 +1723,16 @@ function isSerializedChange(value: unknown): value is SerializedChange {
 		change.change !== undefined
 	);
 }
+
+function sharedTreeChangeHasVisibleChanges(change: SharedTreeChange): boolean {
```
Josmithr, line 1734: Nit: I think some docs here would be useful (even if that's primarily just done by linking to something else). "Visible" could mean a lot of different things.

packages/dds/tree/src/test/shared-tree/editing.spec.ts
```diff
@@ -3439,6 +3439,130 @@ describe("Editing", () => {
...
+			it("Should not be violated when there are multiple reverts", () => {
```
Josmithr: Nit (to match the new test below)?
```suggestion
			it("Should not be violated when there are multiple inserts reverted", () => {
```

packages/dds/tree/src/test/shared-tree/editing.spec.ts
```diff
@@ -3439,6 +3439,130 @@ describe("Editing", () => {
...
+				const undo1 = undoStack.pop() ?? assert.fail("Missing undo");
+				undo1.revert();
+				const undo2 = undoStack.pop() ?? assert.fail("Missing undo");
+				undo2.revert();
+
+				expectJsonTree(branch, [{ "X": 1, "B": 2, "C": 3 }]);
+				unsubscribe();
+			});
+
+			it("Should not be violated when a non-constrained edit is reverted by a later edit", () => {
```
Josmithr: Nit: in this case the non-constrained edit isn't "reverted" in the traditional sense. Not sure what the right terminology would be here, but might be worth differentiating the name from the other test cases that actually operate in terms of "reverts".

Josmithr (approved):
I primarily reviewed the test cases to make sure we have coverage for the primary scenarios we care about. Those look good to me.

Would be good to get @alex-pardes's review of the changes to the rebaser, though, since I don't know that space well.

packages/dds/tree/src/core/rebase/changeRebaser.ts
```diff
@@ -82,11 +82,19 @@ export interface ChangeRebaser<TChangeset> {
 	 * `compose([rebase(a, c), rebase(b, compose([inverse(a), c, rebase(a, c)])])`.
 	 * - `rebase(a, compose([]))` is equal to `a`.
 	 * - `rebase(compose([]), a)` is equal to `compose([])`.
+	 *
+	 * @param ignoreNoChangeViolation - When true, no-change constraint violations are suppressed during rebase.
+	 * This is a kludge to support undo/redo flows where an inverse is rebased over newer commits that have no net
```
alex-pardes: The term "inverse" has certain connotations, and I like to avoid using it to refer to undos.

```suggestion
	 * This is a kludge to support undo/redo flows where an undo is rebased over newer commits that have no net
```

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
```diff
@@ -882,6 +882,7 @@ export class ModularChangeFamily
 		taggedChange: TaggedChange<ModularChangeset>,
 		potentiallyConflictedOver: TaggedChange<ModularChangeset>,
 		revisionMetadata: RevisionMetadataSource,
+		ignoreNoChangeViolation?: boolean,
```
alex-pardes: Can we make this default to `false` instead of just being optional? It would be nice not to have to handle this being undefined.

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
```diff
@@ -943,9 +944,14 @@ export class ModularChangeFamily
...
-			noChangeConstraint = { violated: true };
-			constraintState.violationCount += 1;
+		if (noChangeConstraint !== undefined) {
+			if (!noChangeConstraint.violated && ignoreNoChangeViolation !== true) {
+				noChangeConstraint = { violated: true };
+				constraintState.violationCount += 1;
+			} else if (noChangeConstraint.violated && ignoreNoChangeViolation === true) {
+				noChangeConstraint = { violated: false };
+				constraintState.violationCount -= 1;
+			}
```
alex-pardes: I don't think we'll ever hit the second inner branch during undo rebasing (which is the only place we expect to ever set the `ignoreNoChangeViolation` flag), as I don't think we produce an undo for a commit which already has a constraint violation. But if we were rebasing a change with a constraint violation, we shouldn't remove the violation. Consider if we undid a conflicted commit. The undo would be marked as conflicted, and rebasing it over a no-op shouldn't change that.

```suggestion
		if (noChangeConstraint !== undefined && !noChangeConstraint.violated && ignoreNoChangeViolation !== true) {
			noChangeConstraint = { violated: true };
			constraintState.violationCount += 1;
```

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1339,13 +1340,25 @@ export class TreeCheckout implements ITreeCheckout {
...
+
 			change = tagChange(
 				rebaseChange(
 					this.changeFamily.rebaser,
 					change,
 					commitToRevert,
 					headCommit,
 					this.mintRevisionTag,
+					ignoreNoChangeViolation,
 				).change,
```
alex-pardes, line 1362: Since we already calculated the rebase path, it would be nice to reuse it here. Also good to make sure that we're actually rebasing over the same set of changes that we checked for visible effects.

```suggestion
rebaseChangeOverChanges(
this.changeFamily.rebaser,
change,
[diff],
),
```

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
```diff
@@ -943,7 +944,11 @@ export class ModularChangeFamily
 		);
 
 		let noChangeConstraint = change.noChangeConstraint;
-		if (noChangeConstraint !== undefined && !noChangeConstraint.violated) {
+		if (
+			noChangeConstraint !== undefined &&
+			!noChangeConstraint.violated &&
+			ignoreNoChangeViolation !== true
```
alex-pardes:
```suggestion
			!ignoreNoChangeViolation
```

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1721,6 +1724,12 @@ function isSerializedChange(value: unknown): value is SerializedChange {
 	);
 }
 
+/**
+ * Enumerates through a shared tree change, looking for schema change and field changes that result in visible changes to the tree (e.g. an insert, move, delete)
+ *
```
alex-pardes, line 1730: It would be nice to note that this function considers changes to detached roots to be visible changes, but not renames of roots or builds of new roots.

alex-pardes (approved):

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -439,7 +439,7 @@ export namespace FluidSerializableAsTree {
     export type Data = JsonCompatible<IFluidHandle>;
     const // @system
     _APIExtractorWorkaroundObjectBase: TreeNodeSchemaClass_2<"com.fluidframework.serializable.object", NodeKind_2.Record, TreeRecordNodeUnsafe_2<readonly [() => typeof FluidSerializableObject, () => typeof Array, LeafSchema_2<"string", string>, LeafSchema_2<"number", number>, LeafSchema_2<"boolean", boolean>, LeafSchema_2<"null", null>, LeafSchema_2<"handle", IFluidHandle<unknown>>]> & WithType_2<"com.fluidframework.serializable.object", NodeKind_2.Record, unknown>, {
-    readonly [x: string]: string | number | IFluidHandle<unknown> | System_Unsafe_2.InsertableTypedNodeUnsafe<LeafSchema_2<"boolean", boolean>, LeafSchema_2<"boolean", boolean>> | FluidSerializableObject | Array | null;
+    readonly [x: string]: string | number | IFluidHandle<unknown> | FluidSerializableObject | Array | System_Unsafe_2.InsertableTypedNodeUnsafe<LeafSchema_2<"boolean", boolean>, LeafSchema_2<"boolean", boolean>> | null;
```
Josmithr: @jumyhre Looks like these need to be regenerated

Josmithr (approved):

PR #26764, opened 2026-03-18
Stage removal of URL-based DDS types

Abe27342:
## Description

Deployed Fluid-based solutions can run into problems when their customers have certain reverse-proxy solutions which rewrite URLs to go through predefined endpoint (this allows routing traffic through known domains which can be used for various interception purposes). The core problem in those cases is that JS bundles containing Fluid libraries have rewritten "Type" values for DDSes pointing to the intercepted reverse proxy domain. Concretely, the problem resembles:

Customer has a reverse proxy that's configured to route https://github.com to `my-reverse-proxy.com/?path=github.com`, where the my-reverse-proxy.com domain serves assets from github.com (or whatever's in the path parameter) but replaces any links on that domain with an analogous link to my-reverse-proxy.com.

Fluid DDS types resemble URLs but they are not actually used as such, they function as registry keys. Furthermore, they are persisted in the document under the channel attributes blob. So, such auto-rewriting of URLs causes issues where documents created on a device subject to such a reverse proxy solution would only be usable as long as the reverse proxy has the exact same sort of configuration (and documents created on a device without that sort of thing would not be openable on one that does have such redirection). All of this looks effectively like document corruption to the user.

Historically, I believe URL-like values for DDSes were chosen when Fluid had more ambition for loading DDSes dynamically. We are at this point very far from that space (instead opting to close DDS development to within the repository for API reasons). There's no reason these need to be URLs.

This stages a redirect table in the DDS lookup process that will allow us to later update the type values to be non-URL based. The change cannot be performed immediately for compatibility reasons at rollout time (risk of users with newer version of the code writing documents that older versions do not understand).

### DIFF
--- a/packages/dds/cell/src/cellFactory.ts
+++ b/packages/dds/cell/src/cellFactory.ts
@@ -26,6 +26,9 @@ export class CellFactory implements IChannelFactory<ISharedCell> {
 	/**
 	 * {@inheritDoc CellFactory."type"}
 	 */
+	// New type string, to be activated once the migration has been fully shipped dark and is safe to flip.
+	// See legacyTypeRedirects in packages/runtime/datastore/src/channelContext.ts.
+	// public static readonly Type = "SharedCell";
 	public static readonly Type = "https://graph.microsoft.com/types/cell";
 
 	/**
--- a/packages/dds/cell/src/cellFactory.ts
+++ b/packages/dds/cell/src/cellFactory.ts
@@ -27,8 +27,8 @@ export class CellFactory implements IChannelFactory<ISharedCell> {
 	 * {@inheritDoc CellFactory."type"}
 	 */
 	// New type string, to be activated once the migration has been fully shipped dark and is safe to flip.
-	// See legacyTypeRedirects in packages/runtime/datastore/src/channelContext.ts.
-	// public static readonly Type = "SharedCell";
+	// See LegacyTypeAwareRegistry in packages/runtime/datastore/src/dataStoreRuntime.ts.
+	// public static readonly Type = "cell";
 	public static readonly Type = "https://graph.microsoft.com/types/cell";
 
 	/**
--- a/packages/dds/sequence/src/sequenceFactory.ts
+++ b/packages/dds/sequence/src/sequenceFactory.ts
@@ -16,8 +16,6 @@ import { pkgVersion } from "./packageVersion.js";
 import { SharedStringClass, SharedStringSegment, type ISharedString } from "./sharedString.js";
 
 export class SharedStringFactory implements IChannelFactory<ISharedString> {
-	// TODO rename back to https://graph.microsoft.com/types/mergeTree/string once paparazzi is able to dynamically
-	// load code (UPDATE: paparazzi is gone... anything to do here?)
 	// New type string, to be activated once the migration has been fully shipped dark and is safe to flip.
 	// See LegacyTypeAwareRegistry in packages/runtime/datastore/src/dataStoreRuntime.ts.
 	// public static Type = "mergeTree";
--- a/packages/dds/sequence/src/sequenceFactory.ts
+++ b/packages/dds/sequence/src/sequenceFactory.ts
@@ -18,6 +18,9 @@ import { SharedStringClass, SharedStringSegment, type ISharedString } from "./sh
 export class SharedStringFactory implements IChannelFactory<ISharedString> {
 	// TODO rename back to https://graph.microsoft.com/types/mergeTree/string once paparazzi is able to dynamically
 	// load code (UPDATE: paparazzi is gone... anything to do here?)
+	// New type string, to be activated once the migration has been fully shipped dark and is safe to flip.
+	// See legacyTypeRedirects in packages/runtime/datastore/src/channelContext.ts.
+	// public static Type = "SharedString";
 	public static Type = "https://graph.microsoft.com/types/mergeTree";
 
 	public static readonly Attributes: IChannelAttributes = {
--- a/packages/dds/sequence/src/sequenceFactory.ts
+++ b/packages/dds/sequence/src/sequenceFactory.ts
@@ -19,8 +19,8 @@ export class SharedStringFactory implements IChannelFactory<ISharedString> {
 	// TODO rename back to https://graph.microsoft.com/types/mergeTree/string once paparazzi is able to dynamically
 	// load code (UPDATE: paparazzi is gone... anything to do here?)
 	// New type string, to be activated once the migration has been fully shipped dark and is safe to flip.
-	// See legacyTypeRedirects in packages/runtime/datastore/src/channelContext.ts.
-	// public static Type = "SharedString";
+	// See LegacyTypeAwareRegistry in packages/runtime/datastore/src/dataStoreRuntime.ts.
+	// public static Type = "mergeTree";
 	public static Type = "https://graph.microsoft.com/types/mergeTree";
 
 	public static readonly Attributes: IChannelAttributes = {
--- a/packages/dds/tree/src/sharedTreeAttributes.ts
+++ b/packages/dds/tree/src/sharedTreeAttributes.ts
@@ -12,6 +12,11 @@ import { pkgVersion } from "./packageVersion.js";
  * @beta
  * @legacy
  */
+// New type string, to be activated once the migration has been fully shipped dark and is safe to flip.
+// See legacyTypeRedirects in packages/runtime/datastore/src/channelContext.ts.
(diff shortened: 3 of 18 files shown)


markfields:
@Abe27342 is the shape of the rewritten URLs consistent or predictable enough that we could also include in this fallback codepath some optimistic mapping from rewritten URLs to the known type?  To rescue those docs/tenants that have the reverse proxy problem.

packages/runtime/datastore/src/channelContext.ts
```diff
@@ -34,6 +34,35 @@ import type { ISharedObjectRegistry } from "./dataStoreRuntime.js";
...
+	"https://graph.microsoft.com/types/ink": "Ink",
+	"https://graph.microsoft.com/types/pact-map": "PactMap",
+	"https://graph.microsoft.com/types/task-manager": "TaskManager",
+	"https://graph.microsoft.com/types/consensus-queue": "ConsensusQueue",
+	"https://graph.microsoft.com/types/consensus-register-collection":
+		"ConsensusRegisterCollection",
+	"https://graph.microsoft.com/types/shared-summary-block": "SharedSummaryBlock",
+	"https://graph.microsoft.com/types/SharedArray": "SharedArray",
+	"https://graph.microsoft.com/types/signal": "SharedSignal",
+};
```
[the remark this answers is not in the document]
Abe27342: Going to reply here since it lets us keep the discussion to a comment thread, but fyi this applies to @markfields's comment [here](https://github.com/microsoft/FluidFramework/pull/26764#issuecomment-4084580036)

> @Abe27342 is the shape of the rewritten URLs consistent or predictable enough that we could also include in this fallback codepath some optimistic mapping from rewritten URLs to the known type? To rescue those docs/tenants that have the reverse proxy problem.

It's hard to say. From the telemetry we observed, the reverse proxy solutions mostly were using subdomain techniques in practice, so rather than map along the lines of what I put in the description, they were more like map github.com/foo to github.my-reverse-proxy.com/foo. So... we could consider something like "regex match on /types/{old-type-name} and forward" and I think this would work in practice but limit us flexibility for defining future DDSes. That's a reasonable tradeoff worth considering (ex: we commit to never including "/" in future DDS type names).

packages/runtime/datastore/src/channelContext.ts
```diff
@@ -168,7 +197,14 @@ export async function loadChannelFactoryAndAttributes(
 			}),
 		);
 	}
-	const factory = registry.get(channelFactoryType);
+	let factory = registry.get(channelFactoryType);
```
Abe27342: this redirect logic is actually not sufficient for rollout compat story. I'd fix that up before checking something like this in (and add some tests).

Specifically if the code here starts loading documents with the new Type fields, it doesn't map them to their correct DDSes, just errors. The code as presented is what needs to live long-term.

markfields: (mark as draft if not ready to merge?)

packages/runtime/datastore/src/channelContext.ts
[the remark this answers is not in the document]
anthony-murphy: won't this table just get re-written by the proxy? i'm fine to remove url like going forward. i wonder if we need to encode the old names? base64? so they can't be easily re-written by a proxy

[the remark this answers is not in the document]
anthony-murphy: i think all the customer problems are when they mix and match  proxy vs not. the solution today works if all usages go though the same proxy. it only fail when some users of a document proxy, and some do not

packages/dds/cell/src/cellFactory.ts
```diff
@@ -26,6 +26,9 @@ export class CellFactory implements IChannelFactory<ISharedCell> {
 	/**
 	 * {@inheritDoc CellFactory."type"}
 	 */
+	// New type string, to be activated once the migration has been fully shipped dark and is safe to flip.
+	// See legacyTypeRedirects in packages/runtime/datastore/src/channelContext.ts.
+	// public static readonly Type = "SharedCell";
```
ChumpChief: One feature of a URL is its uniqueness, which is an important property here too.  Maybe this would never be a problem worth worrying about, but we might consider:

1. Ensure uniqueness by appending a precomputed GUID as part of the type?
2. Include a versioning string of some sort?

I don't feel super-strongly about these at the moment though, it's not like these URLs do a great job of satisfying those requirements either.

packages/runtime/datastore/src/channelContext.ts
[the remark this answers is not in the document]
anthony-murphy: for the regex, couldn't we also only run that if we can find the type? so first is direct check? then encoded look up table? last we try a regex re-map. i think this doesn't limit the future, and the future will always pass direct check. regex re-map is just a last resort attempt and we would fail if we did nothing, so risk of being wrong isn't too bad

packages/dds/tree/src/sharedTreeAttributes.ts
```diff
@@ -12,6 +12,11 @@ import { pkgVersion } from "./packageVersion.js";
  * @beta
  * @legacy
  */
+// New type string, to be activated once the migration has been fully shipped dark and is safe to flip.
+// See legacyTypeRedirects in packages/runtime/datastore/src/channelContext.ts.
+// IMPORTANT: "SharedTree" cannot be used here because experimental/dds/tree already uses that string
```
ChumpChief: Maybe just call this "Tree"?

packages/dds/cell/src/cellFactory.ts
ChumpChief: Maybe don't add "Shared" in most cases - at least the ones that didn't already have it in the URL?
```suggestion
	// public static readonly Type = "Cell";
```

ChumpChief (commented):
I'm ok with this in concept, certainly as a fix for the issue you describe.

The probably-harder fix I prefer is to deprecate the `Type` static on the factory entirely, and instead have container and data object authors name the entries in their own registries.  I believe that last time I looked we depend on the static member internally though, so this is more a note for any future-thinking you're doing rather than a recommendation for this PR.

packages/dds/sequence/src/sequenceFactory.ts
```diff
@@ -18,6 +18,9 @@ import { SharedStringClass, SharedStringSegment, type ISharedString } from "./sh
 export class SharedStringFactory implements IChannelFactory<ISharedString> {
 	// TODO rename back to https://graph.microsoft.com/types/mergeTree/string once paparazzi is able to dynamically
 	// load code (UPDATE: paparazzi is gone... anything to do here?)
+	// New type string, to be activated once the migration has been fully shipped dark and is safe to flip.
```
Abe27342, line 19: if we're changing the type here anyway, do we want to address the pre-existing TODO at the same time? it will make the back-compat code a little more complex (need to have a lookup map rather than just prefix check), but might be worth it. Up to reviewer IMO; either way is ok with me.

packages/runtime/datastore/src/channelContext.ts
Abe27342: Probably fair on the first iteration, but seems like general response is pretty favorable and I'd like to go through with getting it or a variant of it merged. So I'll leave it as non-draft.

(this comment is now addressed in latest anyway)

anthony-murphy (approved):

packages/dds/sequence/src/sequenceFactory.ts
anthony-murphy: i'm fine with merge tree. these aren't really exposed, so i don't think clean up of an internal name is really worth it. i think its also fine to remove the comment

Abe27342: works for me. Removed.

packages/dds/cell/src/cellFactory.ts
Abe27342: using GUIDs from the start would be reasonable but I think introducing them at this point is not really worth it considering we now have effectively a closed ecosystem of identifiers we need to support.

In terms of versioning, my gut is to build something like that we wouldn't want to overload the function of `type` here and instead lean more heavily on other information in the attributes blob (we already write a "version" field which is pointless today)

packages/dds/tree/src/sharedTreeAttributes.ts
Abe27342: Yeah--latest version opts instead to just drop all of the initial part of the URLs. That keeps the update rule / logic simple and avoids need for a full lookup path.

steffenloesch (approved):
Looks good to me. We should have a work item to track this, so that we'll switch to the non-url `Type` in a few weeks.

noencke (approved):
Approved for 🌳

PR #26769, opened 2026-03-18
build(client): build:compile:esm:packages - fast production ESM build

jason-ha:
`pnpm build:compile:esm:packages` will do a minimal ESM only package build.

Can be used to check validity of production code without hitting test collateral (except for a few test-oriented packages).

`build:esnext` tasks (`fluid-build --task build:esnext`) are able to complete without needing to build CommonJS with these changes:
  1. Remove extraneous `^tsc` dependencies in examples.
  2. Reduce general `build:esnext` to not depend on `^tsc` and `^api`.
  3. Reduce presence `^api` dependencies to `^api-extractor:esnext` and `^api-extractor:commonjs`.
  4. PropertyDDS packages configured to use bundler resolution (less strict Node16 resolution) and have exports defined in property-common. Previously `property-properties` would build using CommonJS `property-changeset`.

Note general `tsc` dependency is left to depend on `^api` and `^build:esnext`.

### DIFF
--- a/examples/data-objects/multiview/constellation-view/package.json
+++ b/examples/data-objects/multiview/constellation-view/package.json
@@ -64,20 +64,6 @@
 			}
 		}
 	},
-	"fluidBuild": {
-		"tasks": {
-			"eslint": [
-				"@fluid-example/multiview-coordinate-interface#build:esnext",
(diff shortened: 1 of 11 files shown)


jason-ha:
Closing in favor of #27265 that also has option to build ESM tests

PR #26776, opened 2026-03-18
Ensure a summarizer stop request is respected after connecting

kian-thompson:
When taking a look at summarizer telemetry, I observed a weird race condition around connection and last summaries. If the parent disconnects and reconnects before the summarizer finishes loading then it won't stop as the parent expects. 

Example flow:

1. Parent is elected
2. Start loading summarizer
3. Parent disconnects
4. Attempt to call `summarizer?.stop(...)` does nothing since the summarizer has not finished loading (see `createSummarizerFn()`)
5. Parent reconnects
6. Summarizer finishes loading and returns instance to parent
7. Summarizer doesn't recognize that it was supposed to stop, so it continues to live as normal
8. Parent hits `"SummarizerStopTimeout"` since it expected its summarizer to stop

This PR adds a stop check right after step 6.

[AB#61863](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/61863)

### DIFF
--- a/packages/runtime/container-runtime/src/summary/summaryManager.ts
+++ b/packages/runtime/container-runtime/src/summary/summaryManager.ts
@@ -102,6 +102,7 @@ export class SummaryManager
 	private latestClientId: string | undefined;
 	private state = SummaryManagerState.Off;
 	private summarizer?: ISummarizer;
+	private pendingStopReason?: SummarizerStopReason;
 	private _disposed = false;
 	private summarizerStopTimeout?: ReturnType<typeof setTimeout>;
 	/**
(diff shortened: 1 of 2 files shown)


packages/runtime/container-runtime/src/summary/summaryManager.ts
```diff
@@ -287,6 +288,13 @@ export class SummaryManager
 				this.summarizer = summarizer;
 				this.setupForwardedEvents(summarizer);
 
+				// A stop may have been requested while we were awaiting summarizer creation.
+				// Replay it now so the summarizer can observe the stop intent and move to exit.
+				if (this.pendingStopReason !== undefined) {
+					summarizer.stop(this.pendingStopReason);
+					this.pendingStopReason = undefined;
```
shlevari, line 295: Should this be early exiting at this point? Are there any invariants the code below needs to be checking for now with the summarizer maybe being stopped?

kian-thompson: We still want the `summarizer.run(...)` call below to go through, but we want the summarizer to immediately enter the last summary flow when that method is called

shlevari: Nit: Feels worth adding a comment explaining the expectations there.

shlevari (approved):
Approved with suggestions for added comment

PR #26777, opened 2026-03-19
Edit some tree docs for conciseness

CraigMacomber:
## Description

Edits some docs for conciseness, and refer to node kinds instead of node types in a few places.

## Reviewer Guidance


docs/docs/data-structures/tree/node-types.mdx
```diff
@@ -7,41 +7,34 @@ Data on a [SharedTree](./index.mdx) is stored as a [node](./nodes.mdx) of one of
...
--   [**number**](../../api/fluid-framework/schemastatics-interface.md#number-propertysignature): Works identically to a JavaScript number except does not support -0, NaN or Infinity.
--   [**string**](../../api/fluid-framework/schemastatics-interface.md#string-propertysignature): Works identically to a JavaScript string except for not supporting unpaired UTF-16 surrogate pairs.
--   [**null**](../../api/fluid-framework/schemastatics-interface.md#null-propertysignature): Works identically to a JavaScript null.
--   [**FluidHandle**](../../api/fluid-framework/schemastatics-interface.md#handle-propertysignature): A handle to a Fluid DDS or Data Object in the current container. For more information about handles see [Handles](../../concepts/handles.mdx).
+-   [**boolean**](../../api/fluid-framework/schemastatics-interface.md#boolean-propertysignature): a JavaScript boolean.
+-   [**number**](../../api/fluid-framework/schemastatics-interface.md#number-propertysignature): a JavaScript number, but does not support `-0` (normalized to 0), `NaN`, or `Infinity` (normalized to `null` where allowed).
```
CraigMacomber, line 13: this is one place where I added info, in this case how numbers are normalized.

docs/docs/data-structures/tree/events.mdx
```diff
@@ -5,17 +5,15 @@ sidebar_position: 5
...
 
-The `TreeView` object exposes 2 events that communicate changes that affect the whole tree.
+`TreeView` exposes two events for tree-wide changes:
 
-`rootChanged` fires when the contents of the root field (the field that contains the root node) change.
-That is, if a new root node is assigned or the schema changes.
-This will not fire when the node itself changes.
+- `rootChanged` fires when the root field changes — i.e., a new root node is assigned or the schema changes.
+It does not fire when a property of the root node changes.
+- `commitApplied` fires when a change is applied outside a transaction, or when a transaction commits.
```
jenn-le: We should probably update this to prefer the `changed` event

jenn-le (approved):

docs/docs/data-structures/tree/events.mdx
```diff
@@ -37,17 +34,14 @@ on<K extends keyof TreeChangeEvents>(
...
--   `nodeChanged` fires whenever one or more properties of the specified node change.
--   `treeChanged` fires whenever one or more properties of the specified node or any node in its subtree, change.
+-   `nodeChanged` fires when one or more properties of `node` change.
+-   `treeChanged` fires when one or more properties of `node` or any node in its subtree changes.
 
-We recommend looking at the documentation of each of the events for more details.
+See the API docs for each event for further details.
 
-The `Tree.on()` method returns a function that unsubscribes the handler from the event. This method is typically called in clean up code when the node is being removed. For example:
+`Tree.on()` returns an unsubscribe function, typically called during cleanup:
```
jumyhre: Minor nit on the wording: It seems to me the important part of this is to tell the dev they should unsubscribe when they're done with the event handler, rather than suggest generally where this might be called.
```suggestion
`Tree.on()` returns an unsubscribe function. This function must be called once the event handler is no longer needed:
```

docs/docs/data-structures/tree/node-types.mdx
```diff
@@ -7,41 +7,34 @@ Data on a [SharedTree](./index.mdx) is stored as a [node](./nodes.mdx) of one of
...
 
 ### Object Nodes
 
-An [object node](../../api/fluid-framework/treeobjectnode-typealias.md) is a JavaScript object like TreeNode with zero or more named child [fields](../../api/fluid-framework/fieldschema-class.md) exposed as properties of the object.
-The object node's fields can be node types.
-A `SharedTree` can have many object nodes at various places in the tree and they do not all have to conform to the same schema.
-Your schema can specify different properties for different object nodes.
-The schema also specifies whether a child property is
+An [object node](../../api/fluid-framework/treeobjectnode-typealias.md) has zero or more named child [fields](../../api/fluid-framework/fieldschema-class.md) exposed as properties.
+Different object nodes in the same tree can have different schemas.
```
jumyhre, line 25: I'm wondering if we dropped important info here. One of the questions I have when reading this is "What's the difference between a map node and an object node?"

The old document noted 'The object node's fields can be node types.'. for object nodes and 'it only works for string keys' for map nodes. Was the original remark noting that object nodes could support anything as a key, including node objects? If so it seems like it might be good to add this back (though maybe reword it slightly, because i'm not 100% on what it means)

docs/docs/data-structures/tree/events.mdx
CraigMacomber: I feel like these docs exist to suggest usage patterns while the API docs exist to explain the usage contracts.

Also there is no requirement that you "must" unsubscribe. Regardless phrasing here isn't great. Here is a fix:

```suggestion
`Tree.on()` returns an unsubscribe function, which should be called any time the need for callback ends before the node is [deleted](https://fluidframework.com/docs/api/fluid-framework/treestatus-enum#deleted-enummember) (Not just [removed]](https://fluidframework.com/docs/api/fluid-framework/treestatus-enum#removed-enummember).
```

docs/docs/data-structures/tree/node-types.mdx
```diff
@@ -7,41 +7,34 @@ Data on a [SharedTree](./index.mdx) is stored as a [node](./nodes.mdx) of one of
...
+-   [**boolean**](../../api/fluid-framework/schemastatics-interface.md#boolean-propertysignature): a JavaScript boolean.
+-   [**number**](../../api/fluid-framework/schemastatics-interface.md#number-propertysignature): a JavaScript number, but does not support `-0` (normalized to 0), `NaN`, or `Infinity` (normalized to `null` where allowed).
+-   [**string**](../../api/fluid-framework/schemastatics-interface.md#string-propertysignature): a JavaScript string, but does not support unpaired UTF-16 surrogate pairs.
+-   [**null**](../../api/fluid-framework/schemastatics-interface.md#null-propertysignature): JavaScript `null`.
+-   [**FluidHandle**](../../api/fluid-framework/schemastatics-interface.md#handle-propertysignature): a handle to a Fluid DDS or Data Object. See [Handles](../../concepts/handles.mdx).
 
 ## TreeNodes
 
-The following subsections describe the available [TreeNode](../../api/fluid-framework/treenode-class.md) types.
+A [TreeNode](../../api/fluid-framework/treenode-class.md) represents an object or collection. The following kinds are available.
```
CraigMacomber:
```suggestion
There are multiple [kinds](https://fluidframework.com/docs/api/fluid-framework/nodekind-enum) of [TreeNode](../../api/fluid-framework/treenode-class.md), each of which manage and expose their children differently.
```

docs/docs/data-structures/tree/node-types.mdx
CraigMacomber: No. All field keys are strings. All supported property keys for object nodes are also strings (we do not support symbols, and JavaScript wouldn't let property keys be objects). The point here is that different fields can have different schema for the values, but the phrasing for it was unclear and misleading. I think the difference between object and map nodes is well stated in the map node section.

docs/docs/data-structures/tree/events.mdx
CraigMacomber: Replaced with links to API docs which now link to that.

jenn-le (approved):

jenn-le (approved):

PR #26784, opened 2026-03-19
improvement(id-compressor): Add resetUnfinalizedCreationRange API and use in CR.replayPendingStates

markfields:
### Background

Managing ID Allocation ops/batches in the ContainerRuntime has been a perennial source of difficulty, including some current efforts around Batch tracking, and potentially Staging Mode as well.

One special case that we can simplify is what happens for "resubmit".  The current flow submits an ID Allocation op before replaying pending ops.  This is divergent from the typical semantics of ID Allocation ops which is that they're only submitted right before an op that uses a new compressed ID.

### The new API

So we are introducing `IdCompressor.releaseUnfinalizedCreationRange`.  It's similar to `takeUnfinalizedCreationRange`, but instead of returning the range, it merely resets the internal state `nextRangeBaseGenCount` to be before any unfinalized ranges, since we know they won't be finalilzed (since the connection the ID Allocations were sent on closed without those acks).

Now, the next call to `takeNextCreationRange` will include those unfinalized ranges, and we don't need to interject an extra op before replay.

`takeUnfinalizedCreationRange` itself becomes a two-line shortcut.  Maybe we would want to deprecate it since additionally, it's unused internally?

### Reconnect Flow

By the time we get to `replayPendingStates`, we know everything pending has not been sequenced, and never will be until we resubmit.  I.e. we've finalized ALL successfully submitted ranges.  So we can safely move `nextRangeBaseGenCount` back according to the local clusters - those IDs will not be finalized by any in-flight ops, and should be included in the next creation range, triggered whenever the resubmitted main content (or subsequent changes) demands.

### Testing

Updated a test that was added for that `replayPendingStates` codepath to demonstrate that this works (if you comment out the new call, the test fails with `Ranges Finalized Out Of Order`).

---
---

_Here's the summary of the design discussion Claude and I had, comparing this ("Approach B") to 2 other alternatives (Approach A was very bad so omitted here):_

# Approaches for Deferring ID Allocation Op During Replay

## Problem

During `replayPendingStates`, the container runtime calls `takeUnfinalizedCreationRange` and immediately submits an IdAllocation op. Submitting this op during replay is problematic. We want to defer the allocation so it is included in the next naturally-submitted IdAllocation op instead.

## Background

- `takeNextCreationRange` returns IDs generated since the last range was taken, starting from an internal cursor (`nextRangeBaseGenCount`), and advances the cursor forward.
- `takeUnfinalizedCreationRange` returns ALL unfinalized IDs (going back to the last finalized cluster), and also advances the cursor forward.
- `generateCompressedId` does not interact with the range-taking cursor at all -- it only touches `localGenCount`, cluster state, and the normalizer.

## Approach B: `releaseUnfinalizedCreationRange` (IdCompressor change)

Add a new void method to `IIdCompressorCore` / `IdCompressor`:

```ts
public releaseUnfinalizedCreationRange(): void {
    // Reset nextRangeBaseGenCount back to the start of the unfinalized region
}
```

The container runtime calls this during replay instead of submitting an op. The next `takeNextCreationRange` call naturally produces a range covering both the old unfinalized IDs and any new IDs generated in the interim.

### Pros

- State tracking is consolidated inside the IdCompressor
- No merge logic needed -- `takeNextCreationRange` handles everything
- Only a single range is ever produced, so no ordering/overlap issues
- `normalizer.getRangesBetween` works correctly for the expanded range

### Cons

- New method on `IIdCompressorCore` (a `@legacy @beta` public interface)
- `nextRangeBaseGenCount` going backward is a new pattern (currently it only advances)
- The reserved range is opaque -- caller cannot inspect it for logging/debugging

## Approach C: Boolean flag in container runtime (no IdCompressor change)

Add a boolean flag (`needsUnfinalizedResubmit`) to the container runtime. Set it during replay instead of submitting an op. In `submitIdAllocationOpIfNeeded`, check the flag:

```ts
const idRange = this.needsUnfinalizedResubmit
    ? this._idCompressor.takeUnfinalizedCreationRange()
    : this._idCompressor.takeNextCreationRange();
this.needsUnfinalizedResubmit = false;
```

### Pros

- Zero changes to IdCompressor or its public API
- Uses existing, well-tested methods (`takeUnfinalizedCreationRange`)
- No new invariants -- `nextRangeBaseGenCount` continues to only advance
- Simpler to review and reason about

### Cons

- State is split across two components (boolean in container runtime, range state in compressor)
- Container runtime must know about the distinction between the two take methods (though it already does today)

## Recommendation

Both approaches are functionally equivalent and correct. Approach C is the lower-risk path (no API change, uses existing methods). Approach B is the more principled one (compressor owns its own state).


packages/runtime/id-compressor/src/idCompressor.ts
```diff
@@ -244,35 +244,18 @@ export class IdCompressor implements IIdCompressor, IIdCompressorCore {
...
-		return this.updateToRange(range);
+	public releaseUnfinalizedCreationRange(): void {
+		const lastLocalCluster = this.localSession.getLastCluster();
+		this.nextRangeBaseGenCount =
+			lastLocalCluster === undefined
+				? 1
+				: genCountFromLocalId(
+						(lastLocalCluster.baseLocalId - lastLocalCluster.count) as LocalCompressedId,
+					);
 	}
```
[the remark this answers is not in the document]
markfields: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/FluidFramework/pull/26784#discussion_r2962915930)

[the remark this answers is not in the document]
markfields: Never mind I'll take care of it

anthony-murphy (approved):
this change seems like a good direction to me, but i don't have deep id-compressor knowledge

packages/runtime/id-compressor/src/types/idCompressor.ts
```diff
@@ -97,6 +97,18 @@ export interface IIdCompressorCore {
 	 */
 	takeUnfinalizedCreationRange(): IdCreationRange;
```
anthony-murphy, line 98: you mention deprecating this, which seems like the correct approach if it is unused

packages/runtime/id-compressor/api-report/id-compressor.legacy.beta.api.md
```diff
@@ -48,6 +48,7 @@ export interface IIdCompressor {
 export interface IIdCompressorCore {
     beginGhostSession(ghostSessionId: SessionId, ghostSessionCallback: () => void): void;
     finalizeCreationRange(range: IdCreationRange): void;
+    releaseUnfinalizedCreationRange(): void;
```
anthony-murphy: i'm not really sure why so much the of the idcompressor interface is exposed? probably out of scope for this pr,  but if we don't think internal callers should use the new method, we could add an internal interface with internal methods that don't get exposed to the legacy layers

markfields: I believe that's the idea of the "Core" interface that this lives on.  Recently @taylorsw04 mentioned he was surprised to see that it's not `@internal`.  I think it's just exposed through the mocks, as per usual.

packages/runtime/id-compressor/src/types/idCompressor.ts
```diff
@@ -97,6 +97,18 @@ export interface IIdCompressorCore {
...
 
+	/**
+	 * Resets the next creation range to include all unfinalized IDs.
+	 * After calling this, the next call to `takeNextCreationRange` will produce a range
```
Josmithr: Nit: our guidelines call for the summary component of comments like this to be short and semantic. Other aspects like usage, side-effects, etc. should go in a `@remarks` block (or a more specific block as applicable).

```suggestion
	 * Resets the next creation range to include all unfinalized IDs.
	 * @remarks
	 * After calling this, the next call to `takeNextCreationRange` will produce a range
```

Josmithr: Nit (for improved API docs experience, and to make the docs slightly more robust to future refactoring)
```suggestion
	 * Resets the next creation range to include all unfinalized IDs.
	 * After calling this, the next call to {@link IIdCompressorCore takeNextCreationRange} will produce a range
```

Josmithr: (Same suggestion for the other API references here)

packages/runtime/id-compressor/api-report/id-compressor.legacy.beta.api.md
taylorsw04: My sharding PR (which got reverted, and we no longer need) included all the work to make it internal and hide it from consumers. Maybe we should revisit that since we did the work to remove usages from partner repos.

markfields: I might try to get this queued up for 2.100.  If I have time this week to get the announcement in

packages/runtime/id-compressor/src/types/idCompressor.ts
markfields: I'm not going to mess with it here, in favor of trying to deprecate this whole `IIdCompressorCore` interface for moving to `@internal` in 2.100.

anthony-murphy (approved):

taylorsw04:
@markfields I thought you were going to deprecate (or, honestly, just remove--I've audited all external users repos) takeUnfinalizedCreationRange? I'd like to not let the surface grow.

markfields:
> deprecate (or, honestly, just remove...)

@taylorsw04 - I'm hoping to deprecate this whole IIdCompressorCore interface for moving to `@internal` in 2.100.  We'd need to get the announcement together by Friday.

I'll either do that, or have a 1-liner follow-up PR that deprecates the new thing, and then remove it in 2.100.

taylorsw04 (approved):

PR #26788, opened 2026-03-19
feat: add fluid-controlled map interfaces

jenn-le:
Adds fluid-controlled map interfaces, `FluidMap` and `FluidReadonlyMap` for extension by interfaces that want APIs that behave closely to TypeScript maps. These are provided to reduce the risk of changes to the built in map breaking downstream customers. 

`TreeIndex` is changed to use `FluidReadonlyMap` and `TreeMapNodeAlpha` is added which extends `FluidReadonlyMap` instead of `ReadonlyMap`.


packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @remarks
+ * Fluid Framework uses this type instead of the built-in `ReadonlyMap` to insulate
+ * against breaking changes introduced by the TypeScript standard library across versions.
+ * Interfaces that need map-like read behavior should extend this type.
+ *
+ * @sealed @alpha
+ */
+export interface FluidReadonlyMap<K, V>
+	extends Omit<
+		FluidMapLegacy<K, V>,
```
ChumpChief: Prefer to have FluidMapLegacy extend the non-legacy types - define the new/preferred options without any reference to legacy stuff that we'll eventually want to shed.  This way any eventual deprecation/removal of legacy will have no impact on the new stuff.

packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -0,0 +1,149 @@
...
+		"clear" | "delete" | "set" | "forEach" | typeof Symbol.toStringTag
+	> {
+	/**
+	 * Executes the provided function once per each key/value pair in the map.
+	 */
+	forEach(
+		callbackfn: (value: V, key: K, map: FluidReadonlyMap<K, V>) => void,
```
ChumpChief: I wonder if someone better than me at TS knows a clever way to define the map type here to be self-referential so it doesn't have to be defined twice.

packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -0,0 +1,149 @@
...
+ * Fluid Framework uses this type instead of the built-in `ReadonlyMap` to insulate
+ * against breaking changes introduced by the TypeScript standard library across versions.
+ * Interfaces that need map-like read behavior should extend this type.
+ *
+ * @sealed @alpha
+ */
+export interface FluidReadonlyMap<K, V>
+	extends Omit<
+		FluidMapLegacy<K, V>,
+		"clear" | "delete" | "set" | "forEach" | typeof Symbol.toStringTag
```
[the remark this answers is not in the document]
ChumpChief: I also had this comment :)

packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -0,0 +1,149 @@
...
+ * which avoids covariant `this` return type issues in interface hierarchies.
+ * New Fluid interfaces should prefer extending this type over {@link FluidMapLegacy}.
+ *
+ * @sealed @alpha
+ */
+export interface FluidMap<K, V> extends FluidReadonlyMap<K, V> {
+	/**
+	 * Removes all entries from the map.
+	 */
+	clear(): void;
```
ChumpChief: Do we want to keep clear in the non-legacy FluidMap?  I have not thought through it at all really but we frequently comment on how annoying its been to support in SharedMap.  @noencke @anthony-murphy might have opinions.

packages/dds/map/src/index.ts
```diff
@@ -25,6 +25,8 @@ export type {
 	ISharedMapEvents,
 	IValueChanged,
 } from "./interfaces.js";
+
+export type { FluidMapLegacy } from "@fluidframework/core-interfaces";
```
ChumpChief: We don't usually re-export types like this (rather encouraging folks to import them from their origination packages directly).  I'd probably omit unless we have a strong need to.

packages/common/core-interfaces/src/fluidMap.ts
ChumpChief: This also might allow you to put FluidMapLegacy directly in the map package rather than core-interfaces (which will discourage proliferation elsewhere).

ChumpChief (commented):
Thanks for picking this up, it will be nice to stabilize this type :)

packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -0,0 +1,149 @@
+/*!
```
CraigMacomber, line 1: I would like to see tests which check the assignability of these types to the built in types and vise-versa (omitting fields if required)

Also I think the documentation for these types should indicate how they related (variance/assignability wise) to JS maps.

packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -0,0 +1,149 @@
...
+ * but using `IterableIterator` for its iterator methods.
+ *
+ * @remarks
+ * Fluid Framework uses this type instead of the built-in `Map` to insulate against
+ * breaking changes introduced by the TypeScript standard library across versions.
+ * The `set` method returns `this` for compatibility with existing Fluid interfaces
+ * such as `IDirectory` and `ISharedMap`.
+ *
+ * @sealed @public
+ */
```
CraigMacomber, line 35: I think we want something like:

```suggestion
 * Like TypeScript's built in `Map` type, except unaffected by TypeScript's version and configuration options.
 * @remarks
 * This exists so that Fluid has a `Map` type which is safe to implement that cannot be broken by changes to TypeScript's default Map type.
 * All behavior exposed through this interface should be compatible with the corresponding behavior of JavaScript Maps,
 * but it may lack some of the newer APIs,
 * and might express the type slightly different from how TypeScript does in its `Map` type.
 * @sealed @beta @legacy
 */
```

Most importantly, I think this should be beta-legacy, and we can't really document exactly how this different from typescript's map since the whole problem is there is no stable typescript map type which we could compare to.

packages/common/core-interfaces/src/fluidMap.ts
CraigMacomber: I agree: we want to invert the dependency so legacy can be tagged legacy. It is also cleaner that way as the legacy type is a lot bigger and makes more sense to be the one specified in terms of the smaller more public one.

jenn-le: We chose to do it this way purely because the other way means that FluidMapLegacy will have to also be alpha which prevents it from being used by IDirectory and ISharedMap. Any ideas on how we could solve that?

packages/dds/map/src/index.ts
jenn-le: I did it because api extractor complained otherwise. Is there a better way to fix that?

packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -0,0 +1,149 @@
...
+export interface FluidMapLegacy<K, V> {
+	/**
+	 * Returns an iterable of entries in the map.
+	 */
+	[Symbol.iterator](): IterableIterator<[K, V]>;
```
noencke: Should we also decouple ourselves from `IterableIterator`, just in case? If we're going halfway, we may as well go the rest of the way, perhaps? It's cheap to make our own iterable and iterator interfaces - they don't need much of the "extra stuff" that's on the built-in iterator/iterable. [Here's an example.](https://www.typescriptlang.org/play/?ts=5.9.3#code/PTAEBUHcHtQFwJYDsCeAaUBnBBbADgDYqjJwCmATgGYCGAxmZgFwBQpltDoAYgQK4IAJgElyFGgCMCZADzgAfKADeLUGtABtAMoocE6AQB0CMTTjQKAXQAUASiY9+Q0ZUnSX48xTnyA3CwBfFjYkMU4yRwERUykyDzMLH1AyAA9yJEFMSOcY6SSVNSRUuDsHJVAANxp+MgB+B3BfUEFoIvrQfQMyGiRQIKCWEFAtMwRMKgQyQQ7iaoJ4AAsIiQECOABaZHgUPEYSXv04BYgtUABWQwAWUB7p8FOLgE4WOlbMOFAcGjwHXij42IyDTvCjIADmGCQfD0lEsigAvKAipBQABZb4yEHgyHQiSUeR2fyvJDvT7fADSZBQzGy0VcsXiXkxcFBSDBCKRZBR6LwzNZEKRuPxdkMAGsqZg7C83h8vngAGrVPiMX5OOniBmmJlQmEUDnItEYrFsnG6gm2QxVGqS2zSkmy74AUVCoJVtIB7i1iWBLOxgt1cNAiINPL5fp1eL1IrILsmNv8dtJAH1MGQ4Kr-rlZMb2UHOSitGmw2zzUSZVg05TqRmcvTPa4mTn9VzhkWmyLxdTCYmPqm4IrrTX1W44l7vE28wbC3Bi+yRVblfGexW4M7fW6-rWNfXPN6cxgc4Hgy3p7PzYYY+ul8uaBRxCghx7s76S5OWwBBO80FBn7vE0m3veVY0puw6ag2iQRvib4op+96-hanZLv+HyAd+A6Lo+WaMokE7HrBX4-u2FoLowf7lmhKBrq6IFqk+OHeBoUEUAeL5gke+agHB34IResZkbYCZAA)

packages/common/core-interfaces/src/fluidMap.ts
noencke: I think we should have `clear`. This type isn't necessarily for shared objects - one of the use cases in this PR (tree indexing) is not a shared object. If shared objects are logically "A map, but we can't support some traditional operations", then it makes sense to represent that as `Omit<FluidMap, "clear">`.

packages/common/core-interfaces/src/fluidMap.ts
noencke: Then the only built-in we are coupled to is the iterator symbol.

packages/common/core-interfaces/src/fluidMap.ts
jenn-le: Sounds good! Although I have the same question as above about making it so that IDirectory and ISharedMap can still use it

packages/common/core-interfaces/src/fluidMap.ts
jenn-le: It looks like this is the way it's done by TS's ReadonlyMap so I'm inclined to say there probably isn't a better way :(

packages/common/core-interfaces/src/fluidMap.ts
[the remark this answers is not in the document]
jenn-le: So the weird thing here is that Map has the symbol but ReadonlyMap does not, because they're defined as separate interfaces. I removed the omission, but that means that a ReadonlyMap can't be assigned to FluidReadonlyMap. How much do we care about that? @CraigMacomber, thoughts?

.changeset/bright-bananas-yell.md
```diff
@@ -0,0 +1,18 @@
+---
+"@fluidframework/core-interfaces": minor
+"fluid-framework": minor
+"@fluidframework/map": minor
+"@fluidframework/tree": minor
+"__section": feature
+---
+Add Fluid-controlled FluidMapLegacy, FluidReadonlyMap, and FluidMap interfaces to insulate against breaking changes in TypeScript's built-in Map types
```
CraigMacomber: I think this is a bit long for a title.

Maybe 
```suggestion
Fix type errors related to Maps in some TypeScript configurations
```

.changeset/bright-bananas-yell.md
```diff
@@ -0,0 +1,18 @@
...
+
+Introduces `FluidMapLegacy`, `FluidReadonlyMap`, and `FluidMap` interfaces in `@fluidframework/core-interfaces` as, Fluid-controlled alternatives to TypeScript's built-in `Map` and `ReadonlyMap` types.
```
CraigMacomber: I think it would read better if we don't lead with the strange edge case that is FluidMapLegacy, and instead entirely leave that to the sentence below about it as this make more sense without mentioning it until then.

.changeset/bright-bananas-yell.md
```diff
@@ -0,0 +1,18 @@
...
+These interfaces use `IterableIterator` for iterator methods instead of the built-in `MapIterator`, which has changed across TypeScript versions (e.g., gaining `[Symbol.dispose]`), causing type incompatibilities for downstream consumers.
+
+`FluidMapLegacy` is the `@public` mutable map interface, where `set` returns `this` for backward compatibility with existing interfaces like `IDirectory` and `ISharedMap`.
```
CraigMacomber: Given that we have a legacy API surface, and its tagged beta legacy, it seems off to tag this as public. In fact the only think public in the entire shared Map package is this new interface.

I think this type should only be as public/stable as necessary given is users (which I think are all beta/legacy, but are at the very least legacy public not normal public)

packages/common/core-interfaces/src/fluidMap.ts
CraigMacomber: Why is FluidMap alpha? Its alpha because you don't want to make a change to the public API without staging it as alpha first, correct?

So why is it ok to violate that for FluidMapLegacy? If we are confident making FluidMapLegacy public as is and supporting it for ever, then we should be confident in supporting the less screwy MapReadonly to the same level (I don't think we should have such confidence, but if we do, then we should have it for all these types right?)

I think the right thing to do here is if we want to introduce new types, and we want to start them as alpha, then we should start them as alpha.

You can add a way to use shared map using the new alpha FluidMapLegacy if you want, or you could leave FluidMapLegacy out of this PR entirely (and add it later as alpha, or wait for FluidMap to make it to beta, and add FluidMapLegacy after that).

packages/common/core-interfaces/src/fluidMap.ts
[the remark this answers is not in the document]
CraigMacomber: FluidReadonlyMap is sealed, meaning that only we can implement it. If someone has a read-only map and assigns it to a FluidReadonlyMap, that is a bug as they are violating the sealed tag, and their code is likely to stop compiling if typescript changes how it types maps some more (defeating the point of this PR).

Additionally, as a sealed interface, we reserve the right to add things to it in the future. Therefor its legal to assign a FluidReadonlyMap to something else, but the reverse is not true as that could break when we make what we claim are non breaking changes.

So personally I think we should not try and make ReadonlyMap assignable to FluidReadonlyMap: it might work, it might not, but either way users shouldn't do it.

That said, FluidReadonlyMap is sealed, so we can always add more properties later, but can't remove them.

Thus I'm of the opinion that we can omit Symbol.toStringTag if we want, and can add it layer as a new feature in a non breaking change. We can also include it now: I personally don't think it matters much, but causing ReadonlyMap to be unassignable to FluidReadonlyMap might help catch misuse of the sealed type which could be a feature (there are more robust ways to do that, like adding members no type except our implementations could have, but they are messy).

packages/common/core-interfaces/src/fluidMap.ts
CraigMacomber: As this interface is sealed, we should err on the side of having less APIs, as we can always add more later as a non breaking change.

.changeset/bright-bananas-yell.md
```diff
@@ -0,0 +1,15 @@
...
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": feature
+---
+Add Fluid-controlled map and iterator interfaces
+
+Introduces `FluidReadonlyMap`, `FluidMap`, `FluidIterable`, and `FluidIterableIterator` interfaces as alpha APIs in `@fluidframework/core-interfaces`.
+These exist so that Fluid has types which are safe to implement and cannot be broken by changes to TypeScript's default types. All behavior exposed through these interfaces should be compatible with the corresponding behavior of the built-in types, but they may lack some of the newer APIs.
+
+New APIs looking to extend Map or ReadonlyMap should extend these interfaces instead.
```
CraigMacomber: changesets are for consumers of fluid: these new APIs are sealed so the intended audience of this document are not allowed to do this. I think this line should be removed.

.changeset/bright-bananas-yell.md
```diff
@@ -0,0 +1,15 @@
...
+
+`TreeIndex` now extends `FluidReadonlyMap` instead of the built-in `ReadonlyMap`, and `TreeMapNodeAlpha` which extends `FluidReadonlyMap` instead of the built-in `ReadonlyMap` has been added.
+This works to uncouple Fluid's public API surface to the TypeScript standard library's map types, preventing future breakage when those types change.
```
CraigMacomber, line 10: This seems like the thing relevant to FF client users.
 Just keeping this paragraph might make sense.

.changeset/bright-bananas-yell.md
```diff
@@ -0,0 +1,15 @@
...
+"__section": feature
+---
+Add Fluid-controlled map and iterator interfaces
+
+Introduces `FluidReadonlyMap`, `FluidMap`, `FluidIterable`, and `FluidIterableIterator` interfaces as alpha APIs in `@fluidframework/core-interfaces`.
+These exist so that Fluid has types which are safe to implement and cannot be broken by changes to TypeScript's default types. All behavior exposed through these interfaces should be compatible with the corresponding behavior of the built-in types, but they may lack some of the newer APIs.
+
+New APIs looking to extend Map or ReadonlyMap should extend these interfaces instead.
+
+`TreeIndex` now extends `FluidReadonlyMap` instead of the built-in `ReadonlyMap`, and `TreeMapNodeAlpha` which extends `FluidReadonlyMap` instead of the built-in `ReadonlyMap` has been added.
```
CraigMacomber, line 9: There is no change to any package exported TreeMapNodeAlpha types (based on the API report), so this does not seem accurate.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -152,6 +153,47 @@ export interface TreeMapNode<T extends ImplicitAllowedTypes = ImplicitAllowedTyp
...
+ * {@link TreeMapNode} with FluidReadonlyMap-based iteration.
+ *
+ * @remarks
+ * This is the same as {@link TreeMapNode} except that it extends FluidReadonlyMap
+ * instead of the built-in `ReadonlyMap`, insulating against breaking changes
+ * in TypeScript's standard library iterator types.
+ *
+ * @sealed @alpha
+ */
+export interface TreeMapNodeAlpha<T extends ImplicitAllowedTypes = ImplicitAllowedTypes>
```
CraigMacomber, line 166: This type is not included in the API report, so it must not be exported, and if that's not erroring, it must also be unused.

If adding this, I'd expect it to be used similar to how ArrayNodeAlpha is used (in asAlpha and schemaFasctoryAlpha, see the recent PR adding it for refence)

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -152,6 +153,47 @@ export interface TreeMapNode<T extends ImplicitAllowedTypes = ImplicitAllowedTyp
...
+	 * deleted.
+	 *
+	 * @privateRemarks
+	 * Regarding the choice to not return a boolean: Since this data structure is distributed in nature, it isn't
+	 * possible to tell whether or not the item was deleted as a result of this method call. Returning a "best guess"
+	 * is more likely to create issues / promote bad usage patterns than offer useful information.
+	 *
+	 * @param key - The key of the element to remove from the map.
+	 */
+	delete(key: string): void;
```
CraigMacomber: It is odd for these members to exist here.

If they are unchanged from MapNode, please just pull them in from there (With Pick or something like that) rather than duplicating them. If they do not match MapNode, then I don't think we have a good path forward stabilizing this and have a problem.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
jenn-le: It wasn't included in the API report because I didn't realize I had to explicitly generate the entrypoints files as of #26522

I also fixed the parts in this file where I had forgotten to replace uses of TreeMapNode with the alpha version

packages/dds/tree/src/entrypoints/beta.ts
```diff
@@ -115,7 +115,7 @@ export {
 	JsonCompatible, 
 	JsonCompatibleObject, 
 	KeyEncodingOptions, 
-	NodeChangedData,
+	NodeChangedData,
```
CraigMacomber, line 118: this seems wrong

packages/dds/tree/src/entrypoints/legacy.ts
```diff
@@ -102,10 +102,6 @@ export {
 	AnnotatedAllowedTypeUnsafe, 
 	AnnotatedAllowedTypes, 
 	AnnotatedAllowedTypesUnsafe, 
-	ArrayNodeDeltaOp, 
-	ArrayNodeInsertOp, 
-	ArrayNodeRemoveOp, 
-	ArrayNodeRetainOp,
```
CraigMacomber: this seems wrong

jenn-le: that it does

jenn-le: So it actually seems like these are correct. ArrayNodeDeltaOp and the others are being removed from the legacy entrypoint and they aren't legacy so that seems correct. There are also a bunch of things moved to the alpha region in the alpha entrypoint from the beta region and I confirmed that these things actually are alpha. Seems like the current state is just stale or wrong?

@WillieHabimana for confirmation that I'm not royally messing things up

WillieHabi: Think @jason-ha can speak to this

WillieHabi: I thnk you're right, the current state is wrong from merged #26677. I think these might've been incorrectly added to legacy there and your regeneration is fixing that

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
CraigMacomber: > I'd expect it to be used similar to how ArrayNodeAlpha is used (in asAlpha and schemaFasctoryAlpha, see the recent PR adding it for refence)

That part of my comment still applies. There are still no uses of this type in the API report, so the only thing anyone could do with this is blindly down cast a node. We did a nicer job for arrays for the same thing, and should do that here.

packages/dds/tree/src/entrypoints/legacy.ts
CraigMacomber: Would be preferable to do a regeneration in another PR to clean up main and make this diff more logical.

.changeset/bright-bananas-yell.md
jenn-le: Resolving since this is now exported with the updated entrypoint

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
jenn-le: I see, that makes sense. Added it.

CraigMacomber: The `schemaFasctoryAlpha` portion of my comment still applies.

Look at https://github.com/microsoft/FluidFramework/commit/f2b0cf917609b84952db2b9492867e70e0d57981#diff-8c55d1bff49ecdf910cd13e6284c77a8151acbd49cc95297e66a6d702eb19c22R1062

.changeset/bright-bananas-yell.md
CraigMacomber: What I meant was this paragraph:

```markdown
`TreeIndex` now extends `FluidReadonlyMap` instead of the built-in `ReadonlyMap`, and `TreeMapNodeAlpha` which extends `FluidReadonlyMap` instead of the built-in `ReadonlyMap` has been added.
This works to uncouple Fluid's public API surface to the TypeScript standard library's map types, preventing future breakage when those types change.
```

seems like the thing relevant to FF client users.

It might make sense to delete the other paragraphs from the changeset, and only have this one. This paragraph states why changes, then why that matters for customers. The existence of the new Map and iterator types is not directly relevant to anyone, but this is: its users of TreeIndex and TreeMapNode that should care about this change as they are the people impacted.

jenn-le: Ah I see, that makes more sense

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
jenn-le: Oh I see, I misunderstood the comment. Thanks for the example!

CraigMacomber (approved):
Looks good to me.

PR #26789, opened 2026-03-19
Log Level values Update

MarioJGMsoft:
## Description
This change adds the new `LogLevel.info` and `LogLevel.essential` values to the `LogLevel` constant (alongside the existing deprecated default and error values) with proper documentation. It then starts passing LogLevel.info explicitly at telemetry call sites (blobManagerSnapSum and SampledTelemetryHelper) that were previously relying on the default.

The goal is to begin annotating telemetry events with an appropriate log level so that consumers can filter and prioritize events accordingly — for example, distinguishing `info` events (which can be omitted in some sessions to reduce telemetry volume) from `essential` events (which should always be collected for diagnostic purposes). This is a first step toward broader adoption of explicit log levels across the codebase.

## Reviewer Guidance


packages/common/core-interfaces/src/logger.ts
```diff
@@ -55,8 +55,35 @@ export interface ITelemetryBaseEvent extends ITelemetryBaseProperties {
...
+	 * these should always be collected, even in production, for diagnostic purposes.
+	 */
+	essential: 30,
+
+	/**
+	 * Value to assume when not otherwise given.
+	 * @see {@link (LogLevel:variable).info}
+	 * @deprecated Use {@link (LogLevel:variable).info} or {@link (LogLevel:variable).essential}
+	 */
 	default: 20, // Default log level
```
lindsnguyen, line 82: Does this comment make sense if we were saying we'll make everything that doesn't have a level "essential" instead? (ie 30 not 20)

Can we validate that essential is the default level we want to set when someone logging doesn't set the level?

packages/framework/client-logger/app-insights-logger/src/fluidAppInsightsLogger.ts
```diff
@@ -180,6 +180,7 @@ class FluidAppInsightsLogger implements ITelemetryBaseLogger {
 	 * Routes Fluid telemetry events to the trackEvent App Insights API.
 	 * This method also uses the provided {@link FluidAppInsightsLoggerConfig} to
 	 * determine whether an event should be sent or not.
+	 * PR COMMMENT: Should this be updated to also use logLevel?
```
lindsnguyen: yes, for all implementations of `ITelemtryBaseLogger` we need to update send's implementation to actually take in and use logLevel otherwise it just gets ignored even if the caller calls send with LogLevel.

packages/runtime/container-runtime/src/blobManager/blobManagerSnapSum.ts
```diff
@@ -57,11 +58,15 @@ export const toRedirectTable = (
...
-		count: blobManagerLoadInfo.ids?.length ?? 0,
-		redirectTable: blobManagerLoadInfo.redirectTable?.length,
-	});
+	logger.sendTelemetryEvent(
+		{
+			eventName: "AttachmentBlobsLoaded",
+			count: blobManagerLoadInfo.ids?.length ?? 0,
+			redirectTable: blobManagerLoadInfo.redirectTable?.length,
+		},
+		undefined,
```
lindsnguyen: what's this `undefined` being passed in for?

lindsnguyen (commented):
I'm not sure where it would be but we should make sure to document that if the logLevel is not set, clients should assume those logs are essential and to not sample them.

Also we should look in ChildLogger to see if the filtering (not sampling) is being done correctly. I don't think it is anymore because I think in scenarios where the caller sets minLogLevel to essential, while we would have assumed all logs that don't have a logLevel set are still sent (not filtered) the current code would set it to "default" and then filter them. take a look at logger.ts#shouldFilterOutEvent (lines 461-466)

Reason is in this implementation, if logLevel is empty it sets it to LogLevel.default which i think we had decided that is no longer the case?

(lmk if you want to chat about this directly for clarfication!)

packages/runtime/container-runtime/src/blobManager/blobManagerSnapSum.ts
MarioJGMsoft: `sendTelemetryEvent` has `error` as it's second parameter, and from my understanding if I don't add this undefined, then the LogLevel would be sent as the error and not the actual LogLevel

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: I think this comment does make sense because we want to offer guidance as to what level to use instead of default. This is meant more for clients that are using `LogLevel.default`, they need to know what is going to be supported from now on since what they're using is deprecated.

I think we can validate it, through the sendTelemetryEventCore function in logger.ts in telemetry-utils, I think that might be a good choke point to verify that if we are receiving a log without a LogLevel we can add it there. 

I'm not sure how necessary it is because we want to use this as a signal to let consumers know what telemetry from Fluid they can omit (sample), so we don't have any real say on what to do with logs that we are receiving from the outside.

lindsnguyen: Did we not agree that any current logs that have no LogLevel set is going to be set to essential or is that just were doing internally, and the plan is that once we've updated all our internally sent logs to essential manually, anything in the future that doesn't have LogLevel set we'll assume is default/info/20

MarioJGMsoft: Any current logs that Fluid is sending that have no LogLevel are going to be essential, and yeah, that's also my understanding of the plan for the future.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -57,13 +57,14 @@ import type {
...
 } from "@fluidframework/core-interfaces";
-import type {
-	IFluidHandleContext,
-	IFluidHandleInternal,
-	IProvideFluidHandleContext,
-	ISignalEnvelope,
-	OpaqueJsonDeserialized,
-	TypedMessage,
+import {
+	LogLevel,
```
jason-ha: Keep value imports in separate import statement

packages/utils/telemetry-utils/api-report/telemetry-utils.legacy.beta.api.md
```diff
@@ -32,8 +32,8 @@ export interface ITelemetryGenericEventExt extends ITelemetryPropertiesExt {
 // @beta @legacy
 export interface ITelemetryLoggerExt extends ITelemetryBaseLogger {
     sendErrorEvent(event: ITelemetryErrorEventExt, error?: unknown): void;
-    sendPerformanceEvent(event: ITelemetryPerformanceEventExt, error?: unknown, logLevel?: typeof LogLevel.verbose | typeof LogLevel.default): void;
-    sendTelemetryEvent(event: ITelemetryGenericEventExt, error?: unknown, logLevel?: typeof LogLevel.verbose | typeof LogLevel.default): void;
+    sendPerformanceEvent(event: ITelemetryPerformanceEventExt, error?: unknown, logLevel?: typeof LogLevel.verbose | typeof LogLevel.info | typeof LogLevel.essential): void;
```
jason-ha, line 35: Are there essential performance events?

packages/utils/telemetry-utils/src/config.ts
```diff
@@ -3,10 +3,11 @@
  * Licensed under the MIT License.
  */
 
-import type {
-	ConfigTypes,
-	IConfigProviderBase,
-	ITelemetryBaseLogger,
+import {
+	LogLevel,
```
jason-ha: Keep value imports separate from type imports

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -187,7 +187,7 @@ export abstract class TelemetryLogger implements ITelemetryLoggerExt {
 	public sendTelemetryEvent(
 		event: ITelemetryGenericEventExt,
 		error?: unknown,
-		logLevel: typeof LogLevel.verbose | typeof LogLevel.default = LogLevel.default,
+		logLevel?: typeof LogLevel.verbose | typeof LogLevel.info | typeof LogLevel.essential,
```
jason-ha, line 190: So, before there was a default and now we let it be `undefined`. What is the difference in generated .js code for this?
(I want to know if the default was encoded at the call site or in the implementation.)

packages/utils/telemetry-utils/src/sampledTelemetryHelper.ts
```diff
@@ -4,7 +4,11 @@
  */
 
 import { performanceNow } from "@fluid-internal/client-utils";
-import type { IDisposable, ITelemetryBaseProperties } from "@fluidframework/core-interfaces";
+import {
+	LogLevel,
```
jason-ha: Keep value imports separate from type imports. (Assume I comment that for all remaining instances. :) )

jason-ha: Keep value imports separate from type imports. (Assume I comment that for all remaining instances. :) )

packages/utils/telemetry-utils/src/utils.ts
```diff
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import type { ITelemetryBaseEvent } from "@fluidframework/core-interfaces";
+import { LogLevel, type ITelemetryBaseEvent } from "@fluidframework/core-interfaces";
```
jason-ha: heh - I think there was only one more

packages/utils/telemetry-utils/src/utils.ts
jason-ha, line 1: I feel like you don't want to change these now or at least not default to essential. Probably you do want to preserve the log levels that callers are passing.

packages/common/core-interfaces/api-report/core-interfaces.beta.api.md
```diff
@@ -321,6 +321,8 @@ export type Listeners<T extends object> = {
 // @public
 export const LogLevel: {
     readonly verbose: 10;
+    readonly info: 20;
+    readonly essential: 30;
```
jason-ha, line 325: I don't see `// @deprecated` here which makes me think that the deprecation mechanism may not work for these. You should generate documentation and see how the docs look.
Also see what IntelliSense say when these are used.

packages/common/core-interfaces/src/logger.ts
jason-ha: I think the guidance can be updated. It could say use .info to preserve prior treatment (same value).

I think we probably want to state that handling of missing logLevel should be done in two steps. First, we want all missing to be treated as "essential" by any consumers. Second, after we have gone through and audited / updated, then it is okay for missing to be treated as "info". Ideally (Jason's opinion) we don't have any default (missing) cases, but I don't think everyone is on board with that and they want to let it be optional.
(In Jason's really ideal case the internal interfaces are updated to have logLevel be the very first argument.)

Important: logging interface likely spans compatibility boundaries. This means that getting to the second long-term default treatment takes time. And altering method parameter order might be tricky. I don't think we have great clarity on when `@internal` types cross the compat boundaries.

packages/runtime/container-runtime/src/blobManager/blobManagerSnapSum.ts
jason-ha: That doesn't sound right. I would take a look at the before and after .js for the caller and callee to make sure know see what is actually going on.

packages/utils/telemetry-utils/src/config.ts
MarioJGMsoft: applied change

packages/utils/telemetry-utils/src/sampledTelemetryHelper.ts
MarioJGMsoft: Applied change

MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/utils.ts
MarioJGMsoft: Applied change

packages/runtime/container-runtime/src/containerRuntime.ts
MarioJGMsoft: Applied change

packages/common/core-interfaces/api-report/core-interfaces.beta.api.md
MarioJGMsoft: Intellisense does detect the the parameters are deprecated properly but it isn't detected properly in the docs

packages/utils/telemetry-utils/src/utils.ts
MarioJGMsoft: What code are you referring to?

packages/utils/telemetry-utils/api-report/telemetry-utils.legacy.beta.api.md
MarioJGMsoft: I'm not sure, I think because we haven't done the audit it's better to have them be sent as essential rather than as info.

packages/runtime/container-runtime/src/blobManager/blobManagerSnapSum.ts
MarioJGMsoft: I took a look at the .js callers and it still made sense to me to add that undefined. `sendTelemetryEvent` has the signature `(event, error?, logLevel?)`, so if I didn't have that undefined then the logLevel would be passed as the error argument. 

(Copilot even called me out on a few spots I missed it in the comments below)

packages/framework/client-logger/app-insights-logger/src/fluidAppInsightsLogger.ts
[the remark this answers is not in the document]
MarioJGMsoft: Applied change

packages/loader/container-loader/src/debugLogger.ts
```diff
@@ -73,6 +73,7 @@ export class DebugLogger implements ITelemetryBaseLogger {
 	 * Send an event to debug loggers
 	 *
 	 * @param event - the event to send
+	 * PR COMMENT: Should this be updated to also use logLevel?
```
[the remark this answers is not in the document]
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -459,8 +459,8 @@ export class ChildLogger extends TelemetryLogger {
 	}
 
 	private shouldFilterOutEvent(event: ITelemetryBaseEvent, logLevel?: LogLevel): boolean {
-		const eventLogLevel = logLevel ?? LogLevel.default;
-		const configLogLevel = this.baseLogger.minLogLevel ?? LogLevel.default;
+		const eventLogLevel = logLevel ?? LogLevel.essential;
```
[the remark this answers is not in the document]
MarioJGMsoft: Won't apply change.

packages/utils/telemetry-utils/src/utils.ts
```diff
@@ -96,7 +97,7 @@ export function createSampledLogger(
 				if (isSamplingDisabled && (skipLoggingWhenSamplingIsDisabled ?? false)) {
 					return;
 				}
-				logger.sendPerformanceEvent(event);
+				logger.sendPerformanceEvent(event, LogLevel.essential);
```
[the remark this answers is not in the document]
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/utils.ts
```diff
@@ -72,15 +73,15 @@ export function createSampledLogger(
...
 			}
 		},
 		sendTelemetryEvent: (event: ITelemetryGenericEventExt): void => {
 			if (isSamplingDisabled || eventSampler === undefined || eventSampler.sample()) {
 				if (isSamplingDisabled && (skipLoggingWhenSamplingIsDisabled ?? false)) {
 					return;
 				}
-				logger.sendTelemetryEvent(event);
+				logger.sendTelemetryEvent(event, LogLevel.essential);
 			}
```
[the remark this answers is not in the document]
MarioJGMsoft: Applied change

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -2066,35 +2067,41 @@ export class ContainerRuntime
...
+				sessionRuntimeSchema: JSON.stringify(this.sessionSchema),
+				featureGates: JSON.stringify({
+					...featureGatesForTelemetry,
+					closeSummarizerDelayOverride,
+				}),
+				telemetryDocumentId: this.telemetryDocumentId,
+				groupedBatchingEnabled: this.groupedBatchingEnabled,
+				initialSequenceNumber: this.deltaManager.initialSequenceNumber,
+				minVersionForCollab: this.minVersionForCollab,
+			},
```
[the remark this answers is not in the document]
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: The default was encoded in the implementation.

It went from:                                                                                                               
event, error, logLevel = core_interfaces_1.LogLevel.default                                                                                                                                                                                                                  
To:                                                                                                                                                                                                                                                                          
event, error, logLevel

packages/common/core-interfaces/src/logger.ts
```diff
@@ -55,8 +55,35 @@ export interface ITelemetryBaseEvent extends ITelemetryBaseProperties {
...
+	 * They need not be collected in production.
+	 */
+	verbose: 10,
+
+	/**
+	 * Information about the session. These logs could be omitted in some sessions
+	 * if needed (e.g. to reduce overall telemetry volume). If any are collected
+	 * from a particular session, all should be.
+	 */
+	info: 20,
```
ChumpChief, line 69: Naming nit: all telemetry is "info".  Perhaps try to find a word that better communicates the level of optionality between verbose and essential.  Maybe "helpful"?

packages/utils/telemetry-utils/api-report/telemetry-utils.legacy.beta.api.md
ChumpChief: If I understood our prior conversation correctly, all "default" events become "essential" unless/until we audit them individually to determine they are good candidates to get dropped.  In which case all performance events are also essential by default.

packages/runtime/container-runtime/src/blobManager/blobManagerSnapSum.ts
jason-ha: okay - the comment didn't capture the `LogLevel.info` argument addition below. Yes, this is okay.
However, the call site should be updated with comment to say what argument the `undefined` is for. `/* error */ undefined` or `/* error - none */ undefined` or `/* error: no error */ undefined` will work.

packages/common/core-interfaces/api-report/core-interfaces.beta.api.md
jason-ha: The text is kept but not the `@deprecated` annotation?

packages/framework/client-logger/app-insights-logger/src/fluidAppInsightsLogger.ts
```diff
@@ -181,7 +182,7 @@ class FluidAppInsightsLogger implements ITelemetryBaseLogger {
 	 * This method also uses the provided {@link FluidAppInsightsLoggerConfig} to
 	 * determine whether an event should be sent or not.
 	 */
-	public send(event: ITelemetryBaseEvent): void {
+	public send(event: ITelemetryBaseEvent, logLevel?: LogLevel): void {
```
jason-ha, line 185: Isn't this a compile or lint complaint that `logLevel` is unused?
This change doesn't seem to be needed. Why is it needed?

packages/loader/container-loader/src/debugLogger.ts
```diff
@@ -73,8 +74,9 @@ export class DebugLogger implements ITelemetryBaseLogger {
 	 * Send an event to debug loggers
 	 *
 	 * @param event - the event to send
+	 * @param logLevel - the log level of the event
 	 */
-	public send(event: ITelemetryBaseEvent): void {
+	public send(event: ITelemetryBaseEvent, logLevel?: LogLevel): void {
```
jason-ha, line 79: Same as fluidAppInsightsLogger.ts - arg is unused. Why add?

packages/runtime/container-runtime/src/containerRuntime.ts
jason-ha: Does not appear to be applied.

packages/utils/telemetry-utils/api-report/telemetry-utils.legacy.beta.api.md
jason-ha: There were lots of discussions. In another there was a question about performance events and if all of those could be optional.
I just want to understand the plan. If we don't have an answer, then okay, but we need comments.
Also, if we aren't yet changing anything to pass "essential" to `sendPerformanceEvent`, then leave "essential" out until it is used.

packages/utils/telemetry-utils/src/logger.ts
jason-ha: Not flagged as deprecated? `LogLevel.error` should be `LogLevel.essential`.

jason-ha, line 209: Based on callers, `logLevel?: LogLevel` should be retyped to at least `logLevel?: LogLevel | undefined`.
This isn't an error here because `exactOptionalPropertyTypes` isn't yet enabled. Since this is protected, the better signature is:
```typescript
	protected sendTelemetryEventCore(
		event: ITelemetryGenericEventExt & { category: TelemetryEventCategory },
		error: unknown,
		logLevel: LogLevel | undefined,
```

Since it is protected and not exported the best signature might be:
```typescript
	protected sendTelemetryEventCore(
		logLevel: LogLevel | undefined,
		event: ITelemetryGenericEventExt & { category: TelemetryEventCategory },
		error?: unknown,
```
In either case the docs above need updated too.

packages/utils/telemetry-utils/src/logger.ts
jason-ha, line 462: Per Teams discussion we shouldn't just use `LogLevel.essential` here.
Could use a constant like `defaultLogLevelForMissingEventValue`. See Teams to help with the right comment for such a constant.

packages/utils/telemetry-utils/src/telemetryTypes.ts
```diff
@@ -117,12 +117,12 @@ export interface ITelemetryLoggerExt extends ITelemetryBaseLogger {
...
 	 * @param event - Event to send.
 	 * @param error - Optional error object to log.
-	 * @param logLevel - Optional level of the log. Default: {@link @fluidframework/core-interfaces#LogLevel.default}.
+	 * @param logLevel - Optional level of the log. Default: {@link @fluidframework/core-interfaces#LogLevel.essential}.
 	 */
 	sendTelemetryEvent(
 		event: ITelemetryGenericEventExt,
 		error?: unknown,
-		logLevel?: typeof LogLevel.verbose | typeof LogLevel.default,
+		logLevel?: typeof LogLevel.verbose | typeof LogLevel.info | typeof LogLevel.essential,
```
jason-ha, line 125: This is a legacy breaking change since this interface is not sealed.
You either need to avoid this change now or justify it is okay to break outside of the regular process.
Probably we'll want to mark this `@sealed`, but currently it isn't.

jason-ha (changes requested):

packages/utils/telemetry-utils/src/utils.ts
jason-ha: All of the changes in this file.

packages/utils/telemetry-utils/api-report/telemetry-utils.legacy.beta.api.md
ChumpChief: I might have missed that other discussion then.  Is there a spec written down that we could refer to (if not would probably be a good idea to start one)?

In general I'm deferring to the two of you since you have more context on the conversations with the customer, my feedback should only be taken as based on the portions I've observed.

packages/runtime/container-runtime/src/blobManager/blobManagerSnapSum.ts
```diff
@@ -57,11 +58,14 @@ export const toRedirectTable = (
...
-		count: blobManagerLoadInfo.ids?.length ?? 0,
-		redirectTable: blobManagerLoadInfo.redirectTable?.length,
-	});
+	logger.sendTelemetryEvent(
+		{
+			eventName: "AttachmentBlobsLoaded",
+			count: blobManagerLoadInfo.ids?.length ?? 0,
+			redirectTable: blobManagerLoadInfo.redirectTable?.length,
+		},
+		LogLevel.info,
```
jason-ha, line 67: Looks like another `LogLevel` sent as `error` parameter. `error` parameter being `unknown` makes it very hard to make sure we are making the right changes.
I think real consideration should be given to reauthoring the internal interfaces we use to make sure we are doing the right things.
Set a new pattern where logLevel is first and required. Then go through and use that pattern where updating. Everything using the old pattern should be considered not visited.

packages/service-clients/end-to-end-tests/azure-client/src/test/multiprocess/childClient.tool.ts
```diff
@@ -86,7 +86,7 @@ function selectiveVerboseLog(event: ITelemetryBaseEvent, logLevel?: LogLevel): v
 	if (interest === "details") {
 		content.details = event.details;
 	}
-	log(`[${logLevel ?? LogLevel.default}]`, content);
+	log(`[${logLevel ?? LogLevel.essential}]`, content);
```
jason-ha, line 89: I know this one. It doesn't need to be a number. `LogLevel.default` can be replaced with `"unspecified"`.
Also, the signature is wrong (my past mistake). It should be
`function selectiveVerboseLog(event: ITelemetryBaseEvent, logLevel: LogLevel | undefined): void`

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
```diff
@@ -257,10 +257,10 @@ describe("ChildLogger", () => {
 				}
 				sent = true;
 			},
-			minLogLevel: LogLevel.default,
+			minLogLevel: LogLevel.essential,
```
jason-ha, line 260: I think we want to leave this at "info" and test that "info" events are sent.

packages/utils/telemetry-utils/src/test/multiSinkLogger.spec.ts
```diff
@@ -86,16 +86,16 @@ describe("MultiSinkLogger", () => {
...
 		const logger2 = new MockLogger();
 		const multiSink = createMultiSinkLogger({
 			loggers: [createChildLogger({ logger: logger1 }), logger2],
 		});
 		assert.strictEqual(
 			multiSink.minLogLevel,
-			LogLevel.default,
-			"Min loglevel should be set correctly to default",
+			LogLevel.essential,
+			"Min loglevel should be set correctly to essential",
```
jason-ha, line 98: This looks wrong. minLogLevel should not be increasing to filter out more.

packages/utils/telemetry-utils/src/logger.ts
jason-ha: Okay, so below (with this change), it means that we'll now be calling `sendTelemetryEventCore` with `undefined` whereas in the past there was always a defined `logLevel`.
Is that the intention?

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -262,7 +262,7 @@ export abstract class TelemetryLogger implements ITelemetryLoggerExt {
 		this.sendTelemetryEventCore(
 			perfEvent,
 			error,
-			perfEvent.category === "error" ? LogLevel.error : logLevel,
+			perfEvent.category === "error" ? LogLevel.essential : logLevel,
```
jason-ha, line 265: This is now sometimes passing `undefined` instead of a value. Intended?

packages/common/core-interfaces/api-report/core-interfaces.beta.api.md
MarioJGMsoft: Yes, that appears to be the case.

packages/runtime/container-runtime/src/containerRuntime.ts
MarioJGMsoft: Properly applied change

PR #26799, opened 2026-03-20
Add `isViewOf` API and clean up deprecations

noencke:
Closing because I don't like the new name quite enough. Will rethink it.

- Deprecated hasRootSchema in favor of the new isViewOf method on TreeBranchAlpha. `isViewOf` has the same behavior when called with a schema, but also supports being called without one to check whether a branch is a view without strongly typing the result. The name is also more intuitive given that `TreeCheckout` now implements `TreeBranchAlpha` (when `hasRootSchema` was written, only `TreeViewAlpha` implemented `TreeBranchAlpha`)                                                                     
- Delegated hasRootSchema implementations to isViewOf in SchematizingTreeView and TreeCheckout so the deprecated method is just a thin wrapper.                             
- Did the same for `context` and `branch` - the deprecated is now implemented in terms of the supported rather than the other way around.
- Updated tests to use supported rather than deprecated APIs

### DIFF
--- a/.changeset/every-foxes-dig.md
+++ b/.changeset/every-foxes-dig.md
@@ -1,0 +1,17 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": tree
+---
+Replaced `hasRootSchema` with `isViewOf`
+
+`TreeBranchAlpha.hasRootSchema` has been deprecated in favor of `TreeBranchAlpha.isViewOf`.
+Migrating to the new method is trivial:
+
+```ts
+view.hasRootSchema(MySchema);
+// becomes:
+view.isViewOf(MySchema);
+```
+
+The schema parameter may be omitted from `isViewOf` in order to check if a branch is a view without knowing the schema.
--- a/packages/dds/tree/src/shared-tree/schematizingTreeView.ts
+++ b/packages/dds/tree/src/shared-tree/schematizingTreeView.ts
@@ -161,6 +161,15 @@ export class SchematizingSimpleTreeView<
 		);
 	}
 
+	public isViewOf<TSchema extends ImplicitFieldSchema = ImplicitFieldSchema>(
+		schema?: TSchema,
+	): this is TreeViewAlpha<TSchema> {
+		if (schema !== undefined) {
+			return areImplicitFieldSchemaEqual(this.rootFieldSchema, schema);
+		}
+		return true;
+	}
+
 	public isBranch(): this is TreeBranchAlpha {
 		return true;
 	}
@@ -172,7 +181,7 @@ export class SchematizingSimpleTreeView<
 	public hasRootSchema<TSchema extends ImplicitFieldSchema>(
 		schema: TSchema,
 	): this is TreeViewAlpha<TSchema> {
-		return areImplicitFieldSchemaEqual(this.rootFieldSchema, schema);
+		return this.isViewOf(schema);
 	}
 
 	public get schema(): ReadSchema<TRootSchema> {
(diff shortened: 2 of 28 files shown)


.changeset/every-foxes-dig.md
```diff
@@ -0,0 +1,17 @@
...
+`TreeBranchAlpha.hasRootSchema` has been deprecated in favor of `TreeBranchAlpha.isViewOf`.
+Migrating to the new method is trivial:
+
+```ts
+view.hasRootSchema(MySchema);
+// becomes:
+view.isViewOf(MySchema);
+```
+
+The schema parameter may be omitted from `isViewOf` in order to check if a branch is a view without knowing the schema.
```
jenn-le, line 17: nit: naming it `view` but then calling it a branch and saying we can check if it's a view is kinda confusing

.changeset/every-foxes-dig.md
```diff
@@ -0,0 +1,17 @@
...
+---
+Replaced `hasRootSchema` with `isViewOf`
+
+`TreeBranchAlpha.hasRootSchema` has been deprecated in favor of `TreeBranchAlpha.isViewOf`.
+Migrating to the new method is trivial:
+
+```ts
+view.hasRootSchema(MySchema);
+// becomes:
+view.isViewOf(MySchema);
```
jenn-le, line 14: times like these make me wish TS had swift style parameter naming so we could do 

`view.isView(of: MySchema);`

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
```diff
@@ -161,6 +161,15 @@ export class SchematizingSimpleTreeView<
 		);
 	}
 
+	public isViewOf<TSchema extends ImplicitFieldSchema = ImplicitFieldSchema>(
+		schema?: TSchema,
+	): this is TreeViewAlpha<TSchema> {
+		if (schema !== undefined) {
+			return areImplicitFieldSchemaEqual(this.rootFieldSchema, schema);
+		}
+		return true;
```
jenn-le, line 170: It feels weird to me to even allow schema to be undefined just to always return true. When would a user want this? Is there any case where what they're calling this function on isn't a view?

jenn-le: Oh I see, because it can be called on TreeCheckout. Which, still feels weird but fine i guess?

jenn-le: Maybe the naming can be better? It's part of the branch apis but it reads as part of the view apis so it's kind of confusing. Maybe it's worth making a separate `isView` since that more clearly implies that the thing you're calling it on might not be a view. And then `isViewOf` can be part of the view apis.

jzaffiro (approved):

jenn-le (approved):

.changeset/every-foxes-dig.md
noencke: Ugh yes I very much agree.

.changeset/every-foxes-dig.md
noencke: As discussed offline - a view _is_ a branch, but we may be able to do some renaming of the interfaces in the future to make it all more clear.

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
noencke: Hmm, well, we already have `hasRootSchema`, which is basically that.

Would it be better to just do 'isView` - it still works well enough when passed a schema, and it's not awkward when not passed anything.

Or we could do `isView({ of: MySchema })`

PR #26814, opened 2026-03-23
Add linetags and indent support for formatted text

brrichards:
## Summary

- Add support for ordered list, check list, code-block, block-quote on `LineTag`
- Add optional indent field for `StringLineAtom` - Quill intercepts tab input and converts it to a `"\t"` delta insert which works already. When intercepting a tab on a line with a line tag Quill sends the indent with the line tag change. This requires `StringLineAtom` to hold indent value
- Updated the currentViewSchema to reflect new fields. Updated the minversion for collab test to reflect new schema. 

Files Changed
- `textDomainFormatted.spec.ts` - updated compatibility test
- `textDomainFormatted` - add new fields onto `StringLineAtom`
- `textEditor.test.tsx` - add tests for indent
- `quillFormattedView.tsx` - update `parseLineTag` and delta logic to deal with new optional `StringLineAtom.indent` field


packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -431,6 +436,7 @@ export namespace FormattedTextAsTree {
 	 */
 	export class StringLineAtom extends sf.object("StringLineAtom", {
 		tag: LineTag.schema,
+		indent: sf.optional(SchemaFactory.number),
```
CraigMacomber: I think this would benefit from some docs.

Probably update the StringLineAtom docs to clarify that this line formatting applies to the line before the line break (which is counter intuitive for indent), then restate that explicitly specifically for this indent property. Would be nice in indent wored the other way so we could make "content" below include the tabs.

Since I don't think there is a difference between 0 and undefined here, it might be better to just make this required (and delete the old schema compatibility snapshot to enable that).

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -77,6 +77,13 @@ const headerToLineTag = {
...
 } as const satisfies Readonly<Record<number, LineTagValue>>;
+/** Quill list tags → LineTag values. */
+const listToLineTag = {
+	bullet: "li",
+	ordered: "ol",
+	checked: "checked",
+	unchecked: "unchecked",
+} as const satisfies Readonly<Record<string, LineTagValue>>;
 /** LineTag values → Quill attributes. Used by buildDeltaFromTree (tree → Quill). */
 const lineTagToQuillAttributes = {
```
CraigMacomber, line 90: If I understand correctly, it seems like some of these are not mutually exclusive in quill. Like you could have a list thats also a header and a block quote and code block? I had assumed otherwise when I setup the LineTag design.

Not in this PR, but in the future we might want to adjust our schema to better model this, adding `heading`, `list`, `blockquote` etc. properties to StringLineAtom like you did for indent.

brrichards: From messing around in their [formatting playground](https://quilljs.com/docs/formats) it seems only one `Block` can be applied at the same time. Maybe quill can support more than one, but it seems by default they clear all previous block formatting when applying a new one through the delta

brrichards: On second look it seems indent, alignment, and direction can coexist with a block blot. Those might benefit from being added to StringLineAtom in the future

packages/dds/tree/src/text/textDomainFormatted.ts
brrichards: updated docs and made it required

packages/framework/react/src/test/text/textEditor.test.tsx
```diff
@@ -1066,7 +1069,7 @@ describe("textEditor", () => {
 					(op) => op.insert === "\n" && op.attributes?.list === "ordered",
 				);
 				assert(lineOp !== undefined);
-				assert.equal(lineOp.attributes?.indent, undefined);
+				assert.equal(lineOp.attributes?.indent, 0);
```
CraigMacomber: You have code which checks `"indent" in op.attributes`, which I think is broken now. I think you need to either make the view code treat 0 as no indent in the attributes, OR you need to make the code generating the attributes omit 0 and have this test validate the ``"indent" lineOpop.attributes` is false.

brrichards: changed to have attributes omit 0

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -332,8 +332,9 @@ export function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp
 			const lineTag = atom.content.tag.value;
 			Object.assign(currentAttributes, lineTagToQuillAttributes[lineTag]);
 
-			// If the line has an indent property, include that as well
-			if (atom.content.indent !== undefined) {
+			// If the line has a nonzero indent, include that as well.
+			// Omit indent 0 so that "indent" in op.attributes is false for unindented lines.
+			if (atom.content.indent) {
```
CraigMacomber, line 337: exploiting the fact that 0 is falsy here makes the code a bit hard to understand. An explicit `!==0` would be more clear in my opinion (0 is not the only falsy number: I think NaN would get skipped here as well, and we want to make it clear to the reader they are only suppose to be thinking about the zero case.

That said, the comment here is good.

CraigMacomber (approved):

PR #26815, opened 2026-03-24
Add ff-oce agent dashboard skill

noencke:
This moves OCE "dashboard" logic into its own skill for the oce copilot agent. This makes generation of the dashboard more optional and easily repeatable. Also:

* The agent will suggest generating a dashboard in the absence of a specific request.
* The dashboard spins up a sub-agent for each service. This not only "jumpstarts auth" for all the services at once, but is much faster because they are run in parallel. It is also tolerant of services which are slow or down, because it sets timeouts on each service and will generate "the best dashboard it can" even without all information available.
* The `copilot-oce` command now invokes `copilot` rather than `copilot-kusto`, because `copilot-kusto` is redundant (its MCP config is a subset of the ff-oce agent).
* `copilot-oce` now properly boots up the ff-oce agent automatically. Before, it was running ff-oce as a _sub_-agent, not as the primary agent. So users of `copilot-oce` would not actually be running the ff-oce agent directly, but indirectly.
* Small optimization to kusto database querying.


.repoverlay/library/ff-oce/.claude/skills/ff-oce-dashboard/SKILL.md
```diff
@@ -0,0 +1,69 @@
...
+### Step 0: Confirm with the user
+
+Before gathering any data, use the `ask_user` tool:
+
+**Question:** "Would you like me to generate a shift status dashboard? This will query IcM, ADO, Kusto, Teams, and WorkIQ and may trigger auth prompts. Make sure you're connected to **VPN** first — it's required for Kusto and other internal services."
+
+**Choices:** "Yes", "Yes, and write it to a file", "No"
+
+- If **No**: respond "No problem — what can I help you with?" and stop.
+- If **Yes** or **Yes, and write it to a file**: proceed to Step 1. Remember the choice for Step 3.
```
tylerbutler: This is interesting! If you don't tell the agent to explicitly remember this, does it not?

scripts/codespace-setup/agent-aliases.sh
```diff
@@ -13,5 +13,5 @@ alias opus="agency claude --model opus"
 alias copilot="agency copilot"
 alias copilot-ado="agency copilot --mcp 'ado --org fluidframework'"
 alias copilot-kusto="agency copilot --mcp 'kusto --service-uri https://kusto.aria.microsoft.com'"
-alias copilot-oce="repoverlay switch ff-oce && copilot-kusto -i 'use the ff-oce agent for this session'"
+alias copilot-oce="repoverlay switch ff-oce && agency copilot"
```
tylerbutler: This effectively defaults it to the settings you provide in the skill? You may want to use just "copilot" since it's an alias to agency copilot. I can't remember exactly how alias expansion works but it'll probably cause problems.

.repoverlay/library/ff-oce/.claude/agents/ff-oce.md
```diff
@@ -40,6 +40,16 @@ You are an expert at the Fluid Framework Client OCE rotation. You instruct and a
 
 ---
 
+## Session Initialization
```
tylerbutler, line 43: How do we ensure that this is the agent loaded for the session?

tylerbutler: `copilot --agent 'ff-oce'`

scripts/codespace-setup/agent-aliases.sh
tylerbutler:
```suggestion
alias copilot-oce="repoverlay switch ff-oce && copilot --agent 'ff-oce'"
```

tylerbutler (approved):

scripts/codespace-setup/agent-aliases.sh
noencke: Good idea to use the alias. However, I can't get the `--agent` to work. From copilot:

>  1. --agent ff-oce (CLI startup): This is handled by the agency wrapper, which resolves agents from its own hardcoded search paths: ~/.copilot/agents/ (personal) and .github/agents/ (repo). It doesn't know about .claude/agents/.
> 2. /agent (interactive runtime): This is handled by the Copilot CLI engine itself (the inner copilot binary that agency wraps). It scans .claude/agents/, .github/agents/, and probably other paths. That's why it finds ff-oce at runtime.
>
> So agency and copilot have different agent discovery logic — the wrapper is more limited than the runtime. Feels like a bug or oversight in agency. Might be worth flagging to the agency team: "make --agent search the same paths that /agent searches."

Can we perhaps move, duplicate, or sym link the agent file into .copilot/agents?

.repoverlay/library/ff-oce/.claude/agents/ff-oce.md
noencke: I'm not sure yet. Currently the user has to remember to do /agent ff-oce. I was planning on addressing that in a followup PR, though if we can figure it out now, that's great too. [See other comment](https://github.com/microsoft/FluidFramework/pull/26815/changes/BASE..71cd2e70f9f1d944caf2ac100f211041c5f73d76#r2982596813).

scripts/codespace-setup/agent-aliases.sh
tylerbutler: We can! Repoverlay supports this. I'll help you wire it up later this morning. (At the vet right now :))

.repoverlay/library/ff-oce/.claude/skills/ff-oce-dashboard/SKILL.md
noencke: It does/should, but this is just a little "forward link" to make it even more probable. It's hard to know when we should or shouldn't do things like this, but I have a general sense that if you "cross-link" stuff in prompts (things reference other things to "remind" the LLM about them) you increase reliability. Especially when it's cheap to do.

Regardless I inlined the information here in the latest version of this prompt.

scripts/codespace-setup/agent-aliases.sh
noencke: Awesome! Thanks - looking forward to it, this will improve the dev experience a lot.

noencke: Working. We simply needed an extra indirection when passing the command line args (extra "--"). Good find Tyler, thanks.

tylerbutler (approved):
Looks good once CI passes.

PR #26820, opened 2026-03-24
fix(property-changeset): Prevent duplicate array insertions during rebase

bhavin121:
<h2 data-pm-slice="1 3 []">Summary</h2><ul><li><p>Fixes the <strong>Invalid Array Length</strong> crash (<code>RangeError: Invalid array length</code>) that occurs during array changeset rebasing in PropertyDDS concurrent editing scenarios.</p></li><li><p>The root cause is duplicate INSERT operations being emitted by <code>_rebaseArrayChangeSetForProperty</code>. This fix adds an <code>advanceRebaseIteratorB</code> skip mechanism that prevents a paired INSERT from being emitted twice — once from the <code>splitOverlapping</code> special case and again from the iterator's standalone yield.</p></li><li><p>Mirrors the existing <code>advanceIteratorB</code> skip logic already proven correct in <code>_performApplyAfterOnPropertyArray</code>.</p></li></ul><h2>Problem</h2><p>When two clients concurrently edit the same array — one replacing elements (remove + insert at the same position) and another inserting elements — the rebased changeset contains <strong>duplicate insertions</strong>. These duplicate INSERT operations at the same position then flow into <code>mergeWithLastIfPossible</code>, where two Map entries end up pointing to the <strong>same</strong> <code>metaInformation</code> object (and thus the same <code>rebasedRemoveInsertRanges</code> array). The merge loop pushes elements from <code>nextRange</code> into <code>previousRange</code>, but since they are the same array reference, each push grows the array being iterated — creating an infinite loop that runs until the array exceeds JavaScript's maximum size (~4.3 billion elements), crashing with:</p><pre><code>RangeError: Invalid array length
  at Array.push (&lt;anonymous&gt;)
  at mergeWithLastIfPossible (array.ts)
  at pushOp (array.ts)
  at applyRebaseSegment (array.ts)
  at _rebaseArrayChangeSetForProperty (array.ts)</code></pre><p>This crashes <code>SharedPropertyTree.processCore</code>, disconnecting the client from the collaborative session.</p><p><strong>Reproduction scenario:</strong></p><pre><code>Initial array: ["A", "B", "C", "D", "E"]

Client 1 (local):  replace first 3 → { remove: [[0, ["A","B","C"]]], insert: [[0, ["X","Y","Z"]]] }
Client 2 (remote): insert "P" at 0, "Q" at 4 → { insert: [[0, ["P"]], [3, ["Q"]]] }

Expected rebased output: { remove: [[1, ["A","B","C"]]], insert: [[1, ["X","Y","Z"]]] }
Actual rebased output:   { remove: [[1, ["A","B","C"]]], insert: [[1, ["X","Y","Z"]], [1, ["X","Y","Z"]]] }
                                                                                       ^^^^^^^^^^^^^^^^^^
                                                                                       duplicate INSERT
                                                          → triggers infinite loop in mergeWithLastIfPossible
                                                          → crashes with "Invalid array length"</code></pre><h2>Root Cause</h2><p>In <code>_rebaseArrayChangeSetForProperty</code> (<code>array.ts</code>), when the <code>splitOverlapping</code> special case emits a paired INSERT (because its position fell behind the advancing REMOVE), it sets the segment flag to <code>completeB</code>. This triggers <code>iteratorB.next()</code>, which yields the <strong>same INSERT again</strong> (since the iterator only <em>peeked</em> at it, not consumed it). Without a skip mechanism, the INSERT gets emitted a second time.</p><p>The duplicate INSERT at the same position is the upstream root cause. The infinite loop in <code>mergeWithLastIfPossible</code> (via shared <code>metaInformation</code> references) is a downstream consequence — fixing the duplicate emission eliminates the crash entirely without needing any changes to <code>mergeWithLastIfPossible</code> or <code>getRangeForAppliedOperation</code>.</p><h2>Fix</h2><p>Added ~15 lines of new code — an <code>advanceRebaseIteratorB</code> closure with a two-stage skip:</p><ol><li><p><strong>Stage 1 (record):</strong> When the just-emitted segment matches the paired <code>removeInsertOperation</code> (by reference equality), record it for skipping and do <em>not</em> advance the iterator.</p></li><li><p><strong>Stage 2 (skip):</strong> On the next advancement, call <code>next()</code>, detect the duplicate INSERT via the recorded reference, call <code>next()</code> again to skip past it, then clear the skip state.</p></li></ol><p>Replaced the two direct <code>iteratorB.next()</code> call sites (<code>completeB</code>/<code>completeBpartOfA</code> and <code>completeAcompleteB</code>) with <code>advanceRebaseIteratorB()</code>.</p><h2>Changed File</h2>
File | Change
-- | --
experimental/PropertyDDS/packages/property-changeset/src/changeset_operations/array.ts | Added advanceRebaseIteratorB helper in _rebaseArrayChangeSetForProperty; replaced raw iteratorB.next() calls at the completeB/completeBpartOfA and completeAcompleteB advancement sites

<h2>Test Plan</h2><ul><li><p>Run existing <code>array.spec.ts</code> rebase tests to verify no regressions</p></li><li><p>Add a test case for the reproduction scenario: concurrent replace (remove+insert at position 0) vs. multi-position insert, verifying the rebased output contains exactly one copy of the INSERT and no <code>Invalid array length</code> crash</p></li><li><p>Verify convergence: apply Client 2's change then the rebased Client 1 change, and confirm the final array matches the expected state</p></li></ul>


bhavin121:
Hi @anthony-murphy, When you get a chance, could you please take a look at this PR?

bhavin121:
Hi @anthony-murphy, any update on this PR.

CraigMacomber:
PR description includes:

>Add a test case for the reproduction scenario: concurrent replace (remove+insert at position 0) vs. multi-position insert, verifying the rebased output contains exactly one copy of the INSERT and no Invalid array length crash

That sound like a good idea (a regression test for this bug). There however does not seem to be one included in this PR.

> Verify convergence: apply Client 2's change then the rebased Client 1 change, and confirm the final array matches the expected state

I'm not sure if that's intended to be part of the above mentioned test (which is missing) or some separate verification, but might be good to ensure that is done as well (and clarified in the PR description)

experimental/PropertyDDS/packages/property-changeset/src/changeset_operations/array.ts
```diff
@@ -1761,6 +1761,26 @@ export namespace ChangeSetArrayFunctions {
...
 		const segment: OperationRangeRemove | OperationRangeInsert = {};
+		let skipIteratorBOperation;
+		const opB = iteratorB.opDescription;
+
+		const advanceRebaseIteratorB = () => {
+			if (
+				(opB as any).removeInsertOperation &&
+				segment.op !== undefined &&
+				skipIteratorBOperation === undefined &&
+				segment.op.operation === (opB as any).removeInsertOperation
```
CraigMacomber: These casts to `any` are extremally non-type safe.

Please avoid introducing new code which defeats type safety.

If it is impractical to be type safe here, please document why that is the case, why its actually safe, and do something less unsafe (like cast to some specific type which exposes removeInsertOperation with the correct tying instead of it also being just `any`.

experimental/PropertyDDS/packages/property-changeset/src/changeset_operations/array.ts
```diff
@@ -1761,6 +1761,26 @@ export namespace ChangeSetArrayFunctions {
 
 		let currentIndexOffset = 0;
 		const segment: OperationRangeRemove | OperationRangeInsert = {};
+		let skipIteratorBOperation;
+		const opB = iteratorB.opDescription;
+
+		const advanceRebaseIteratorB = () => {
```
CraigMacomber, line 1794: This code is relatively complex, and, given the bug, somewhat fragile. I think it would be a good idea to include some comments on these new locals (skipIteratorBOperation, opB, advanceRebaseIteratorB) to explain their semantics.

Without such details, it's quite hard to know if this code is correct, both for current reviewers, and future readers and editors of this code.

experimental/PropertyDDS/packages/property-changeset/src/changeset_operations/array.ts
bhavin121: Thanks for catching this. I've replaced `(opB as any)` with `(opB as NoneNOPOperation)` in the new `advanceRebaseIteratorB`, and also updated the pre-existing casts in `advanceIteratorB` (line ~1568) to match.

`NoneNOPOperation` is the union `RemoveOperation | InsertOperation | ModifyOperation` — all three subtypes define `removeInsertOperation?`, so the cast is type-safe. The only member of `GenericOperation` that lacks this property is `NOPOperation`, but in that case `removeInsertOperation` evaluates to `undefined`, and the `if` condition short-circuits harmlessly at the first check.

Commit: 87206eb

experimental/PropertyDDS/packages/property-changeset/src/changeset_operations/array.ts
bhavin121: Agreed — the two-stage skip mechanism is non-obvious without context. I've added doc comments on all three locals:

- **`opB`**: Reference to iteratorB's mutable operation descriptor (shared object updated in-place on each `next()` call).
- **`skipIteratorBOperation`**: Two-stage skip state — records the INSERT's operation reference when `splitOverlapping`'s special case has already emitted it, so the next advancement knows to skip past the duplicate yield.
- **`advanceRebaseIteratorB`**: The skip-aware advancement helper. Stage 1 records, Stage 2 skips.

I've also added equivalent comments to the existing `advanceIteratorB` in `_performApplyAfterOnPropertyArray` for consistency, and noted that the two closures are intentionally kept separate (they capture different local scopes).

Commit: 87206eb

experimental/PropertyDDS/packages/property-changeset/src/changeset_operations/array.ts
```diff
@@ -1761,6 +1761,26 @@ export namespace ChangeSetArrayFunctions {
...
+				skipIteratorBOperation = segment.op.operation;
+			} else {
+				iteratorB.next();
+				if (skipIteratorBOperation && opB.operation === skipIteratorBOperation) {
+					iteratorB.next();
+				}
+				skipIteratorBOperation = undefined;
+				getRangeForAppliedOperation(opB, rangeB, undefined, in_options);
+			}
+		};
```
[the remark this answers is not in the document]
bhavin121: Good observation. The two helpers (`advanceIteratorB` and `advanceRebaseIteratorB`) are structurally identical, but they close over different local variables in different function scopes — extracting a shared helper would require passing ~6 mutable references via a parameter object, which would add complexity and indirection for minimal benefit (~15 lines of code).

Additionally, the two functions operate in different contexts (apply-after vs. rebase) that could theoretically diverge in the future. Refactoring `advanceIteratorB`, which has been stable for years, carries more risk than the duplication itself.

I've added a comment in both helpers noting the relationship and explaining why they are kept separate.

experimental/PropertyDDS/packages/property-changeset/src/changeset_operations/array.ts
```diff
@@ -1761,6 +1761,26 @@ export namespace ChangeSetArrayFunctions {
...
+		const opB = iteratorB.opDescription;
+
+		const advanceRebaseIteratorB = () => {
+			if (
+				(opB as any).removeInsertOperation &&
+				segment.op !== undefined &&
+				skipIteratorBOperation === undefined &&
+				segment.op.operation === (opB as any).removeInsertOperation
+			) {
+				skipIteratorBOperation = segment.op.operation;
```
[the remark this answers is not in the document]
bhavin121: Added two regression tests in `array.spec.ts`:

1. **"Concurrent replace (remove+insert) vs multi-position insert should not produce duplicate inserts"** — directly verifies that the rebased output contains exactly one INSERT (not two), and that positions are correctly shifted.

2. **"Convergence: concurrent replace (remove+insert) vs multi-position insert"** — uses the existing `testRebasedApplies` helper to verify that both paths (local-then-delta vs. remote-then-rebased-local) converge to the same final state.

Both tests pass (ESM and CJS). All 268 existing tests continue to pass with no regressions.

Commit: 87206eb

bhavin121:
Hi @CraigMacomber, I have addressed the review comments. Could please check now? And let me know if anything is needed from my side.

CraigMacomber:
CI seems rather stuck, I'm going to merge main and see if that unsticks it.

CraigMacomber (approved):
The code looks to be decent quality, and has regression tests. I am not truly confident that the fix is correct as I have not understood it in sufficient detail, but given the fix is coming from the as far as I know only users of this, seems maintainable and there is nothing harmful in it, I think it should be fine to merge.

PR #26826, opened 2026-03-25
fix(build-cli): remove flaky parallel changeset test

tylerbutler:
## Summary

- Removes the "should write all changesets in parallel" test from `canonicalizeChangesets.test.ts` that was causing intermittent CI failures due to concurrent git operations in the test environment.

### DIFF
--- a/build-tools/packages/build-cli/src/test/library/canonicalizeChangesets.test.ts
+++ b/build-tools/packages/build-cli/src/test/library/canonicalizeChangesets.test.ts
@@ -210,17 +210,14 @@ describe("canonicalizeChangesets", () => {
 	});
 
 	it("should write all changesets in parallel", async () => {
-		// Create multiple changesets
-		const writePromises: Promise<void>[] = [];
+		// Create multiple changesets sequentially (git operations can't be parallel)
 		for (let i = 0; i < 10; i++) {
-			writePromises.push(
-				writeChangesetFile(
-					`change-${i}.md`,
-					`---\n"@fluid/package-${i}": patch\n"__custom": "metadata"\n---\n\nChange ${i}\n\nDetailed description.\n`,
-				),
+			// eslint-disable-next-line no-await-in-loop
+			await writeChangesetFile(
+				`change-${i}.md`,
+				`---\n"@fluid/package-${i}": patch\n"__custom": "metadata"\n---\n\nChange ${i}\n\nDetailed description.\n`,
 			);
 		}
-		await Promise.all(writePromises);
 
 		await canonicalizeChangesets(testDir);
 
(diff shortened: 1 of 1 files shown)


build-tools/packages/build-cli/src/test/library/canonicalizeChangesets.test.ts
```diff
@@ -210,17 +210,14 @@ describe("canonicalizeChangesets", () => {
 	});
 
 	it("should write all changesets in parallel", async () => {
-		// Create multiple changesets
-		const writePromises: Promise<void>[] = [];
+		// Create multiple changesets sequentially (git operations can't be parallel)
```
tylerbutler:
```suggestion
		// Create multiple changesets sequentially because `writeChangesetFile` calls git operations and can't be parallellized
```

build-tools/packages/build-cli/src/test/library/canonicalizeChangesets.test.ts
```diff
@@ -210,17 +210,14 @@ describe("canonicalizeChangesets", () => {
...
 		for (let i = 0; i < 10; i++) {
-			writePromises.push(
-				writeChangesetFile(
-					`change-${i}.md`,
-					`---\n"@fluid/package-${i}": patch\n"__custom": "metadata"\n---\n\nChange ${i}\n\nDetailed description.\n`,
-				),
+			// eslint-disable-next-line no-await-in-loop
```
tylerbutler:
```suggestion
			// eslint-disable-next-line no-await-in-loop -- needs to be sequential because the method calls git operations
```

alexvy86 (commented):
Given the test name, feels like more than changing the way the test works we should just delete it? The actual functionality seems to be covered by other tests in the file; not sure what this one was doing other than trying to verify that making those writes in parallel worked... which apparently didn't?  Feels like I'm missing something because based on the PR description, this test should have never been able to pass.

build-tools/packages/build-cli/src/test/library/canonicalizeChangesets.test.ts
```diff
@@ -210,17 +210,14 @@ describe("canonicalizeChangesets", () => {
 	});
 
 	it("should write all changesets in parallel", async () => {
```
jason-ha:
```suggestion
	it("should write all changesets sequentially", async () => {
```

jason-ha: I think I agree with Alex's suggestion. Just delete this. The multiple case appears handled by "should process multiple changeset files".
Otherwise, the implementation should be smartened to use atomic git operations and handle this test case as is.

tylerbutler: The impl in question is just atest helper so not worth any more effort. Deleted the test.

tylerbutler:
> Given the test name, feels like more than changing the way the test works we should just delete it? The actual functionality seems to be covered by other tests in the file; not sure what this one was doing other than trying to verify that making those writes in parallel worked... which apparently didn't? Feels like I'm missing something because based on the PR description, this test should have never been able to pass.

Test is now deleted.

alexvy86 (approved):

PR #26830, opened 2026-03-25
Allow skipping connection retries on load

kian-thompson:
Currently, we continue to retry retriable errors indefinitely on container load. However in case of certain consumers where there is limited time (30/60 seconds) for the load, we should not have infinite retries even on retriable errors.

If the consumer of `load` wants to handle the retry mechanism (i.e. via the `loadExistingContainer` free function), they can enable the `"Fluid.Container.DisableLoadConnectionRetries"` flag. This flag will cause no internal retries to occur and let any error surface to the consumer of `load`.

[AB#42777](https://dev.azure.com/fluidframework/internal/_workitems/edit/42777)

### DIFF
--- a/packages/loader/container-loader/src/connectionManager.ts
+++ b/packages/loader/container-loader/src/connectionManager.ts
@@ -344,6 +344,8 @@ export class ConnectionManager implements IConnectionManager {
 		reconnectAllowed: boolean,
 		private readonly logger: ITelemetryLoggerExt,
 		private readonly props: IConnectionManagerFactoryArgs,
+		private readonly maxConnectionAttempts?: number,
+		private readonly allowReconnect: boolean = true,
 	) {
 		this.clientDetails = this.client.details;
 		this.defaultReconnectionMode = this.client.mode;
@@ -621,6 +623,17 @@ export class ConnectionManager implements IConnectionManager {
 				);
 
 				lastError = origError;
+
+				// When maxConnectionAttempts is set, do not retry beyond the allowed attempts.
+				// The consumer will own the retry policy.
+				if (
+					this.maxConnectionAttempts !== undefined &&
+					connectRepeatCount >= this.maxConnectionAttempts
+				) {
+					const error = normalizeError(origError, { props: fatalConnectErrorProp });
+					this.props.closeHandler(error);
+					throw error;
+				}
 
 				// We will not perform retries if the container disconnected and the ReconnectMode is set to Disabled or Never
 				// so break out of the re-connecting while-loop after first attempt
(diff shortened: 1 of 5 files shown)


packages/loader/container-loader/src/connectionManager.ts
```diff
@@ -970,6 +983,17 @@ export class ConnectionManager implements IConnectionManager {
 
 		this.disconnectFromDeltaStream(reason);
 
+		// When allowReconnect is false, do not attempt to reconnect after a disconnect.
+		// Surface the error to the consumer who owns the retry policy.
+		if (!this.allowReconnect) {
+			if (reason.error === undefined) {
```
jatgarg: nit: maybe use ternary operator.

packages/loader/container-loader/src/container.ts
```diff
@@ -2007,6 +2013,8 @@ export class Container
 					this._canReconnect,
 					createChildLogger({ logger: this.subLogger, namespace: "ConnectionManager" }),
 					props,
+					disableLoadConnectionRetries ? 1 : undefined,
```
jatgarg: nit: put the argument name in comment from the Deltamanager constrcutor

packages/loader/container-loader/src/container.ts
```diff
@@ -2007,6 +2013,8 @@ export class Container
...
+					!disableLoadConnectionRetries,
```
jatgarg: nit: put the argument name in comment from the Deltamanager constrcutor

packages/loader/container-loader/src/container.ts
dannimad: This is redundant. We could just send disableLoadContecctionRetries and deal with this loci inside the connection manager.

dannimad:
I know at least one workflow (loading a container offline with blobs attached) that relies on indefinite retries and I don't know if there could be more of them that could be affected by this change. I would leave the current behavior as default and raise the flag on known affected consumers. I'm sure @anthony-murphy would have a say on this.

anthony-murphy:
> I know at least one workflow (loading a container offline with blobs attached) that relies on indefinite retries and I don't know if there could be more of them that could be affected by this change. I would leave the current behavior as default and raise the flag on known affected consumers. I'm sure @anthony-murphy would have a say on this.

it out of scope for this pr, but overall, i think the load flow is broken in that there isn't a way for a consumer to have visibility into it until it completes, which is what leads to problems like this.  Ideally consumers would have objects first, like document service, then container, and then instantiate actions on them to transition states, load, connect, etc. this would allow user controlled retry on those state transitions, and they could give up and dispose at any time.

the above it not how it works, and not how it expect to work soon, but maybe helpful to keep in mind as we make changes in this area.

I don't see a problem with this change, given the behavior is opt-in, so load will fail in more cases. retry will need to retry the whole expensive flow without any way to save partial progress, but that's just how it is. to get any kind of partial progress we have to start moving in the direction i laid out above, which is a much larger and more costly change.

packages/loader/container-loader/src/container.ts
```diff
@@ -1610,7 +1610,11 @@ export class Container
 			this.connectToDeltaStream(connectionArgs);
 		}
 
-		this.storageAdapter.connectToService(this.service);
+		// When DisableLoadConnectionRetries is enabled, use no internal retries.
+		// The consumer will own the retry policy.
+		const disableLoadRetries =
+			this.mc.config.getBoolean("Fluid.Container.DisableLoadConnectionRetries") === true;
```
shlevari, line 1616: Since the control internally is for maxRetries, does it make sense to pass that through config as well?

shlevari (approved):

dannimad (approved):

PR #26834, opened 2026-03-25
Port EditManager codec to ClientVersionDispatchingCodecBuilder

CraigMacomber:
## Description

Port EditManager codec to ClientVersionDispatchingCodecBuilder.

This is the second to last of the version dispatched codecs to be ported to the new builder system.
This deduplicates logic related to schema validation, version dispatching etc.

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/shared-tree-core/editManagerCodecs.ts
+++ b/packages/dds/tree/src/shared-tree-core/editManagerCodecs.ts
@@ -30,16 +30,28 @@ import { makeV1toV4andV6CodecWithVersion } from "./editManagerCodecsV1toV4.js";
 import { makeSharedBranchesCodecWithVersion } from "./editManagerCodecsVSharedBranches.js";
 import { EditManagerFormatVersion } from "./editManagerFormatCommons.js";
 
+/**
+ * Context required for encoding/decoding the {@link EditManager}'s {@link SummaryData}.
+ */
 export interface EditManagerEncodingContext {
 	idCompressor: IIdCompressor;
 	readonly schema?: SchemaAndPolicy;
 }
 
+/**
+ * Codec name used to identify the {@link EditManager} codec, see {@link makeEditManagerCodecBuilder}.
+ */
 export const editManagerCodecName = "EditManager";
 
+/**
+ * Options for constructing an {@link EditManager} codec, see {@link makeEditManagerCodecBuilder}.
+ */
 interface EditManagerCodecOptions<TChangeset> extends ICodecOptions {
+	/** Codecs for encoding changesets. */
 	changeCodecs: ICodecFamily<TChangeset, ChangeEncodingContext>;
+	/** Maps each EditManager format version to the corresponding changeset format version. */
 	dependentChangeFormatVersion: DependentFormatVersion<EditManagerFormatVersion>;
+	/** Codec for encoding revision tags within changesets. */
 	revisionTagCodec: IJsonCodec<
 		RevisionTag,
 		EncodedRevisionTag,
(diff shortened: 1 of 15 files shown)


packages/dds/tree/src/shared-tree-core/editManagerCodecs.ts
```diff
@@ -3,176 +3,131 @@
...
 	revisionTagCodec: IJsonCodec<
 		RevisionTag,
 		EncodedRevisionTag,
 		EncodedRevisionTag,
 		ChangeEncodingContext
-	>,
-	options: EditManagerCodecOptions & CodecWriteOptions,
-): IJsonCodec<
+	>;
+}
```
Josmithr, line 61: Nit: anything interesting we should document here?

packages/dds/tree/src/shared-tree-core/editManagerCodecs.ts
```diff
@@ -3,176 +3,131 @@
...
+
+export function makeEditManagerCodecBuilder<
```
Josmithr, line 66: Nit: docs?

Josmithr: Would also probably be good to add some inline comments to the `versions` entries

packages/dds/tree/src/shared-tree-core/editManagerCodecsV1toV4.ts
```diff
@@ -23,7 +27,15 @@ export interface EditManagerEncodingContext {
 	readonly schema?: SchemaAndPolicy;
 }
 
-export function makeV1CodecWithVersion<TChangeset>(
+/**
+ * Create the provided version of the EditManager codec.
```
Josmithr: Nit: should we replace "EditManager" with a link tag?

packages/dds/tree/src/shared-tree-core/editManagerCodecsVSharedBranches.ts
```diff
@@ -43,107 +44,83 @@ export function makeSharedBranchesCodecWithVersion<TChangeset>(
...
-					originator: json.originator,
-				};
-
-				if (json.branches !== undefined) {
-					const branches = new Map<BranchId, SharedBranchSummaryData<TChangeset>>();
-					for (const branch of json.branches) {
-						const decodedBranch = decodeSharedBranch(
+	const codec: CodecAndSchema<SummaryData<TChangeset>, EditManagerEncodingContext> = {
+		schema,
+		encode: (data: SummaryData<TChangeset>, context: EditManagerEncodingContext) => {
```
Josmithr, line 53: Nit: given how lengthy the implementations of `encode` and `decode` are here, do you think it would enhance readability to extract them into module-private functions (with docs)?

CraigMacomber: Yes, I think it would be good, but not in this PR: this change is not changing them at all, just lowering their indent by 1.

It sucks this diff is such a mess despite that, but I'd rather not make it worse.

packages/dds/tree/src/shared-tree-core/editManagerCodecsV1toV4.ts
CraigMacomber: If we did, the link wouldn't work in IntelliSense since it's not in scope, and nothing else would process it, so it would never work. Linking SummaryData works, so I've added a link to both.

packages/dds/tree/src/shared-tree-core/editManagerCodecs.ts
CraigMacomber: Add note to see EditManagerFormatVersion for version specific docs.

Josmithr (approved):

PR #26835, opened 2026-03-25
Add CI readiness skill

noencke:
This adds a skill to the repo which checks whether a branch is ready to be pushed to CI without obvious problems. This skill is meant to catch the annoying things you forget to do, like format, lint, policy check, and which will cause your CI build to fail. This is otherwise not reviewing the contents of the PR. There are two steps:

1. A hardcoded script which runs all the low hanging fruit (format, policy, etc.). This should complete in under 10 seconds.
2. An LLM reasoning step which looks at more complicated things like, "This package changed, and it changed an API, and it's a dependency of the @fluid-framework package, so you'll need to build API extractor in that package as well". This step can optionally run tests for changed packages, but won't unless instructed to. The skill will prompt the user at the start with "how strong" of a validation they'd like.

The hardcoded script portion could also become a git push hook or similar, if we like.


noencke:
Incorporated into https://github.com/microsoft/FluidFramework/pull/26849

PR #26840, opened 2026-03-25
test(tree): Add simple plain text insert/remove character op-size benchmark tests

Josmithr:
Adds `TextAsTree.Tree` op size benchmark integration tests.

[AB#68388](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/68388) 

The following report breaks down the current results.

# Op Size Benchmark Results

All measurements taken in `--perfMode`. Other axes are held at their default values (depth=1, characterCount=1, keyLength=1) while the axis under test varies.

---

### Op size in relation to Character Count

| Characters inserted | Op size (bytes) |
|--------------------:|----------------:|
|                   1 |             676 |
|                  10 |             752 |
|                 100 |           1,294 |
|               1,000 |           6,696 |

**Relationship: ** Linear — approximately **6 bytes per character** inserted, with a ~670-byte baseline.

---

### Op size in relation to Tree Depth

| Tree depth | Op size (bytes) |
|-----------:|----------------:|
|          1 |             676 |
|          5 |           1,044 |
|         25 |           2,885 |
|        125 |          12,086 |

**Relationship: ** Linear — approximately **92 bytes per level** of tree depth, with a ~584-byte baseline.

---

### Op size in relation to Property Key Length

| Key length (chars) | Op size (bytes) |
|-------------------:|----------------:|
|                  1 |             676 |
|                 10 |             685 |
|                100 |             775 |
|              1,000 |           1,675 |

**Relationship: ** Linear — approximately **1 byte per character** of key length, with a ~675-byte baseline.

---

### Op size in relation to Character Count

| Characters removed | Op size (bytes) |
|-------------------:|----------------:|
|                  1 |             505 |
|                 10 |             507 |
|                100 |             509 |
|              1,000 |             511 |

**Relationship: ** Effectively constant — total range is only **6 bytes** across a 1000x change in removed character count. Remove ops encode a `(start, count)` range rather than the removed content, so op size is independent of how many characters are removed.

---

### Op size in relation to Tree Depth

| Tree depth | Op size (bytes) |
|-----------:|----------------:|
|          1 |             505 |
|          5 |             873 |
|         25 |           2,714 |
|        125 |          11,915 |

**Relationship: ** Linear — approximately **92 bytes per level** of tree depth, with a ~413-byte baseline. The per-level cost is identical to the insert case.

---

### Op size in relation to Property Key Length

| Key length (chars) | Op size (bytes) |
|-------------------:|----------------:|
|                  1 |             505 |
|                 10 |             514 |
|                100 |             604 |
|              1,000 |           1,504 |

**Relationship: ** Linear — approximately **1 byte per character** of key length, with a ~504-byte baseline. The per-character key cost is identical to the insert case.

---

# Encoding Benchmark Results

## Encoded size in relation to String Length

| String length (chars) | Encoded size (bytes) |
|----------------------:|---------------------:|
|                     1 |                  122 |
|                    10 |                  158 |
|                   100 |                  518 |
|                 1,000 |                4,118 |

**Relationship: ** Linear — approximately **4 bytes per character**, with a ~118-byte baseline for the tree structure overhead.


packages/dds/tree/src/test/shared-tree/opBenchmarkUtilities.ts
```diff
@@ -0,0 +1,73 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import type { ISequencedDocumentMessage } from "@fluidframework/driver-definitions/internal";
```
Josmithr, line 9: Note for reviewers: most of this code was extracted from `opSize.bench.ts` for reuse. `assertLinear` and `assertApproximatelyConstant` are the only new additions.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,278 @@
...
+
+					const encoded = TreeAlpha.exportVerbose(view.root);
+					// TextAsTree nodes never contain IFluidHandle, so this cast is safe.
+					const encodedSize = utf8Length(encoded as unknown as JsonCompatibleReadOnly);
+
+					reporter.addMeasurement("Encoded Size (Bytes)", encodedSize);
+				},
+			});
+		}
+	});
```
Josmithr, line 527:
```suggestion
	});

	// TODO: formatted text benchmarks.
```

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,280 @@
...
+	/**
+	 * Numbers of characters to insert or remove in each benchmark.
+	 */
+	const charCounts = [1, 10, 100] as const;
+
+	/**
+	 * One entry per wrapper schema variant. Each bundle captures the view config, a factory for
+	 * building wrapper trees of a given depth, and a helper for finding the leaf text node.
+	 *
+	 * `getLeaf` accepts `unknown` so that the heterogeneous array can be iterated uniformly;
```
CraigMacomber: Seems like it should probably accept `TreeNode | string` or something like that.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,280 @@
...
+
+	const key1 = "a" as const;
+	const key10 = "a".repeat(10) as A10;
+	const key100 = "a".repeat(100) as A100;
+
+	// #region Schema definitions
+	// Each wrapper schema places a TextAsTree.Tree node under a property key of a fixed length.
+	// The key length affects op size because property names appear in encoded tree paths.
+
+	class WrapperKeyLen1 extends schemaFactory.objectRecursive(
```
CraigMacomber: The reports in the PR summary don't mention key size all. It would be good to clarify which key size those tables are for, and have a table for how key size impacts something.

CraigMacomber: If you use a map node instead, then you can have a single schema for all key sizes, and type the key as "string".

Seems like that would make things a bit cleaner.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,280 @@
...
+	});
+
+	// TODO: formatted text benchmarks.
+});
+
+describe("TextDomain encoding benchmarks", () => {
+	/**
+	 * String lengths used as the test axis.
+	 */
+	const stringLengths = [1, 10, 100, 1000] as const;
```
CraigMacomber: We are pretty sure the relationship is linear.

What we really care about is the constant factor and the slope. 2 data points should be enough for that (1 and something larger).

This does remind we, for cases like this, we used to (intyernally) have a util like "expectLinearRelationship" which could take in an array of x,y pairs, thro if they were not linear, and then report the offset and slope.

Using such a setup here could let is take three data points, and do an assert (which will serve as a correctness test) they are linear, and report the exact offset and slope as performance measurements. I think that would be ideal, though in practice not too much better than just taking 1 data points for 1 and 1000.

CraigMacomber: Note the compressed format (if/when we test it) might not be exactly linear for, but maybe the simpler setup and a few data points makes sense.

At the very least though, I think we should probably mark the middle sizes as "BenchmarkType.Perspective" tests and also skip them when running in correctness mode to keep the test count down as they are not really measuring something we care about as a actual metric and are instead mostly redundant)

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,280 @@
...
+
+	configureBenchmarkHooks();
+
+	const viewConfig = new TreeViewConfiguration({ schema: TextAsTree.Tree });
+
+	describe("TextAsTree.Tree node encoded size", () => {
```
CraigMacomber: We try and have one top level describe per file to keep the total list of top level describe blocks from getting too big (its annoying in the test browser for example) and to keep a 1:1 between files and describe which makes test a bit easier to find.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,280 @@
...
+				only: false,
+				type: BenchmarkType.Measurement,
+				title: `exportVerbose encoded size for string of length ${length}`,
+				run: async (reporter) => {
+					const independentTree = createIndependentTreeAlpha({});
+					const view = independentTree.viewWith(viewConfig);
+					view.initialize(TextAsTree.Tree.fromString("a".repeat(length)));
+
+					const encoded = TreeAlpha.exportVerbose(view.root);
+					// TextAsTree nodes never contain IFluidHandle, so this cast is safe.
```
CraigMacomber: Rather than this comment and a unchecked cast, I believe we have a convertHandles util which you can call and provide "fail" as the conversion function to convert them to never, and use that to do a safe checked conversion.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: Yeah, that seems an obvious thing to do... no idea why that didn't occur to me initially... Updated.

Josmithr: There were some details, but they were a bit buried. Should be more obvious now.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: I've updated the tests to only use 2 configurations for each test axis for performance measurements; the others are now perspective. Each axis now also only has 1 configuration tested in correctness mode.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: This ended up being moot with the conversion from recursive object schema to recursive map schema.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,195 @@
...
+							benchmarkCustom({
+								only: false,
+								type: benchmarkType,
+								title: `insert ${charCount} character(s) into empty string at depth ${depth} with key length ${keyLength}`,
+								run: async (reporter) => {
+									const tree = createConnectedTree();
+									registerOpListener(tree, currentTestOps);
+									const view = tree.viewWith(viewConfig);
+									view.initialize(makeTree(depth, key, ""));
+									currentTestOps.length = 0; // discard initialization ops
```
CraigMacomber: This seems much cleaner. Applies to other places as well.

```suggestion
									const view = tree.viewWith(viewConfig);
									view.initialize(makeTree(depth, key, ""));
									registerOpListener(tree, currentTestOps);
```

packages/dds/tree/src/test/shared-tree/opBenchmarkUtilities.ts
```diff
@@ -0,0 +1,170 @@
...
+	/**
+	 * The data points to test, where `x` is the axis value and `y` is the measured op size.
+	 */
+	readonly points: readonly { x: number; y: number }[];
+	/**
+	 * The minimum acceptable R² value.
+	 * @defaultValue 0.999
+	 */
+	readonly r2Threshold?: number;
+}): void {
```
CraigMacomber: When I wrote a utility like this, I made some rather different choices, which I think would make sense to use here as well.

1. (This one is a bit subjective, and I don't have a strong opinion): I excepted the relationship to be exact, like you would get with integers. For ops this would make sense. Thus the threshold isn't needed: just assert all points with the same x value have the same y, then dedupe the points and form a line from the first two, and assert the others land exactly on it.
2. Return the slope and offset so callers can either assert they have the expected values or use them as performance measurements.
3. Document this is for deterministic data only: we don't want randomness getting involved.

packages/dds/tree/src/test/shared-tree/opBenchmarkUtilities.ts
```diff
@@ -0,0 +1,189 @@
...
+			units: "count",
+		},
+	];
+}
+
+/**
+ * Asserts that the given (x, y) points lie on a line, using R² ≥ `r2Threshold`.
+ * @throws Throws an error when fewer than 3 points are provided, since 2 points always define a perfect line.
+ */
+export function assertLinear({
```
CraigMacomber, line 142: this function is easy to test, and has a few edge cases easy to mess up. A simple test suite for it would be good.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,526 @@
...
+								const opStats = getOperationsStats(localOps);
+								opSizeByCharCount.push({
+									x: characterCount,
+									y: opStats["Total Op Size (Bytes)"],
+								});
+								measurements.addOpStats(opStats, `characterCount=${characterCount}`);
+							}
+							if (isInCorrectnessTestingMode) {
+								assertLinear({ points: opSizeByCharCount });
+							}
```
CraigMacomber, line 189: I think you should call this in all modes.

If you adjust assertLinear as I suggested, then you can return the per character size (the slope) as the measurement primary for the benchmark

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,526 @@
...
+		const filteredKeyConfigs = keyConfigs.filter(
+			(config) => isInPerformanceTestingMode || config.runInCorrectnessMode,
+		);
+
+		/**
+		 * Creates a collector for op-size sweeps.
+		 * Extends {@link createMeasurementCollector} with {@link addOpStats}, which records the
+		 * three standard op statistics (Total Op Size, Max Op Size, Total Ops) for a single config.
+		 */
+		function createCollector(): {
```
CraigMacomber: I don't think this is a good pattern. I think each test should prabably directly report the thing its supposed to measure, like the bytes per character in an insert, fixed op overhead, bytes per level in path to tree etc.

We get much less but much more meaningful data that way.

Also I don't think this code is using our units options very well (given that I see (Bytes) inline in a string here, not listed as the units.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,526 @@
...
+							const measurements = createCollector();
+
+							for (const { characterCount } of filteredCharacterCountConfigs) {
+								const localOps: ISequencedDocumentMessage[] = [];
+
+								const key = getPropertyKey(defaultKeyLength);
+
+								const tree = createConnectedTree();
+								const view = tree.viewWith(viewConfig);
+								view.initialize(makeTree(defaultTreeDepth, key, "a".repeat(1000)));
```
CraigMacomber: I think this 1000 here is arbitrary, but must be the same as the one below. Seems like it must be larger than characterCount: is there a reason not to just use characterCount, or maybe characterCount+1 or something here?

CraigMacomber: Regardless of what we do use, we should dedupe the two numbers which must match

CraigMacomber: This applies to other places hard coding 1000 multiple times in other tests

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,526 @@
...
+								const textNode = getLeaf(view.root, key);
+								textNode.removeRange(0, defaultCharacterCount);
+								assert.equal(textNode.characterCount(), 1000 - defaultCharacterCount);
+
+								const opStats = getOperationsStats(localOps);
+								opSizeByDepth.push({ x: depth, y: opStats["Total Op Size (Bytes)"] });
+								measurements.addOpStats(opStats, `depth=${depth}`);
+							}
+							if (isInCorrectnessTestingMode) {
+								assertLinear({ points: opSizeByDepth });
```
CraigMacomber: we might want to assert we have at least 3 data points here, since it would be easy to edit how we list and filter the options such that this check does not actually enforce anything.

Josmithr: `assertLinear` asserts that there are 3+ data points.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: As I noted in my Teams message, we are grouping these so we can assert linear _inside_ the test, rather than in an after block or something that will cause less than ideal error output. Possible alternatives I can see:
- Assert in an `after` block and live with unfortunate error output
- Separate tests for measurements vs linearity assertion

We can take the other approach, but it has downsides.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: > If you adjust assertLinear as I suggested, then you can return the per character size (the slope) as the measurement primary for the benchmark

This isn't possible if we extract each of the inner benchmarks into their own test cases as you suggested [here](https://github.com/microsoft/FluidFramework/pull/26840#discussion_r3089438580). Unless we want to re-run the scenarios for a separate relationship benchmark test.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
CraigMacomber: I think we want to (as noted in https://github.com/microsoft/FluidFramework/pull/26840/changes/b7eab12f79dfcd44ce50aa239b8e8198df792d49?new_files_changed=true#r2067821389 ) report the data computed by the linearity logic as the stats for the test.

I agree we should not do asserts in an after block: I'm not suggesting that.

I'm suggesting that in each test, we directly create a CollectedData that is what that specific test wants to report, and return it, rather than using any sort of builder.

We want to report something like:

```
[
			{
				name: "size per character",
				value: linearData.slope,
				units: "bytes",
				type: ValueType.SmallerIsBetter,
				significance: "Primary",
			},
]
```

or for another test:

```
[
			{
				name: "fixed op overhead",
				value: linearData.offset,
				units: "bytes",
				type: ValueType.SmallerIsBetter,
				significance: "Primary",
			},
]
```

For possibly both in the same test if you prefer.

CraigMacomber: Ok, maybe I can't link to that comment I posted (links are hard I guess?), but I mean the one thats outdated that you didn't respond to on src/test/shared-tree/opBenchmarkUtilities.ts

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
CraigMacomber: I don't think rerunning them would be too bad, but you can report multiple primary measurements from a single benchmark if you want to. All of them will get included in the geometric mean and highlighted in the console output.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,518 @@
...
+							return [
+								{
+									name: "bytes per key character (remove)",
+									value: slope,
+									units: "bytes",
+									type: ValueType.SmallerIsBetter,
+									significance: "Primary",
+								},
+								{
+									name: "base remove op size (by key)",
```
Josmithr:
```suggestion
									name: "base remove op size (measured by variable key length)",
```

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,518 @@
...
+							return [
+								{
+									name: "bytes per path level (remove)",
+									value: slope,
+									units: "bytes",
+									type: ValueType.SmallerIsBetter,
+									significance: "Primary",
+								},
+								{
+									name: "base remove op size",
```
Josmithr:
```suggestion
									name: "base remove op size (measured by variable path length)",
```

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,518 @@
...
+								opSizes.push(totalOpSize);
+							}
+
+							// Remove ops encode a (start, count) range, not the removed characters,
+							// so op size should be essentially independent of character count.
+							assertApproximatelyConstant({ sizes: opSizes, maxDeltaBytes: 20 });
+							const avgOpSize = opSizes.reduce((a, b) => a + b, 0) / opSizes.length;
+							return [
+								{
+									name: "fixed remove op size",
```
Josmithr:
```suggestion
									name: "fixed remove op size (measured by variable character count)",
```

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,518 @@
...
+							return [
+								{
+									name: "bytes per key character (insert)",
+									value: slope,
+									units: "bytes",
+									type: ValueType.SmallerIsBetter,
+									significance: "Primary",
+								},
+								{
+									name: "base insert op size (by key)",
```
Josmithr:
```suggestion
									name: "base insert op size (measured by variable key length)",
```

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,518 @@
...
+							return [
+								{
+									name: "bytes per path level (insert)",
+									value: slope,
+									units: "bytes",
+									type: ValueType.SmallerIsBetter,
+									significance: "Primary",
+								},
+								{
+									name: "base insert op size",
```
Josmithr:
```suggestion
									name: "base insert op size (measured by variable path length)",
```

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,518 @@
...
+							return [
+								{
+									name: "bytes per inserted character",
+									value: slope,
+									units: "bytes",
+									type: ValueType.SmallerIsBetter,
+									significance: "Primary",
+								},
+								{
+									name: "fixed insert op overhead",
```
Josmithr:
```suggestion
									name: "fixed insert op overhead (measured by variable character count)",
```

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: Updated

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: Updated

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: I went with the recommendation you made [here](https://github.com/microsoft/FluidFramework/pull/26840#discussion_r3103592956)

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -218,36 +170,44 @@ describe("TextDomain benchmarks", () => {
...
-							}
-							if (isInCorrectnessTestingMode) {
-								assertLinear({ points: opSizeByCharCount });
+								const { "Total Op Size (Bytes)": totalOperationSize } =
+									getOperationsStats(localOperations);
+								points.push({ x: characterCount, y: totalOperationSize });
 							}
 
-							return measurements.build();
+							// The encoding of individual character nodes uses variable-length IDs
```
CraigMacomber: I suspect results would be exact if we stick to character counts of at least 2 (so we always deduplicate the string identifier).

"com.fluidframework.leaf.string" is 30 characters, which should go fully missing when n = 0 skewing things a bit, in addition to it saving the indirection to deduplicate it in the n = 1 case.

At least I think that's what happening.

Might be nicer to filter those out and assert its exactly linear, and use that data for the report. Optionally you could include an assertLinear with them and a maxDeviation to ensure things arn't too crazy.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -258,33 +218,43 @@ describe("TextDomain benchmarks", () => {
...
-							if (isInCorrectnessTestingMode) {
-								assertLinear({ points: opSizeByDepth });
+								const { "Total Op Size (Bytes)": totalOperationSize } =
+									getOperationsStats(localOperations);
+								points.push({ x: depth, y: totalOperationSize });
 							}
 
-							return measurements.build();
+							// Depth values with more digits in JSON encoding (e.g. 25 vs 5) can
+							// cause 1-byte deviations from exact linearity.
```
CraigMacomber, line 240: I'm curious how this happens. Offhand, I can't think of why this would be the case, so I wonder if the cause of it is a bug.

Maybe add in a TODO to investigate this and confirm it the encoding is working as intended (no reason to block this PR for fixing it, as this will help the in investigation)

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,526 @@
...
+								},
+							] as CollectedData;
+						},
+					});
+				});
+
+				describe(`Op size by tree depth`, () => {
+					benchmarkIt({
+						only: false,
+						type: BenchmarkType.Measurement,
```
CraigMacomber: these are now properly documented defaults, and in my opinion should not  be included (I removed them from all existing tests while this branch was outstanding)

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,526 @@
...
+									type: ValueType.SmallerIsBetter,
+									significance: "Primary",
+								},
+								{
+									name: "base insert op size (measured by variable path length)",
+									value: intercept,
+									units: "bytes",
+									type: ValueType.SmallerIsBetter,
+								},
+							] as CollectedData;
```
CraigMacomber: Casts like this are bad for type safety. Remove it (and optionally use an explicit return type), use "satisfies" or a typed variable. See also other tests doing the same.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,526 @@
...
+							const { slope, intercept } = assertLinear({ points });
+							return [
+								{
+									name: "bytes per key character (insert)",
+									value: slope,
+									units: "bytes",
+									type: ValueType.SmallerIsBetter,
+									significance: "Primary",
+								},
+								{
```
CraigMacomber, line 295: IIRC, I think having two `significance: "Primary",` measurements is actually legal, and might be a good idea here so this suite will include both in the computed geometric mean in the console output (really that's the main things its sued for currently). If you don't do that these values will get treated like things like margin or error or batch size, and reported, but not considered something we are directly trying to measure/optimize

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -0,0 +1,526 @@
...
+
+		// TODO: formatted text benchmarks.
+	});
+
+	describe("TextDomain encoding benchmarks", () => {
+		const testConfigurations = [
+			{
+				stringLength: 1,
+				benchmarkType: BenchmarkType.Measurement,
+				runInCorrectnessMode: true,
```
CraigMacomber, line 462: our benchmarks recently gained a "skip" option which lets you set a mode to skip them in. Setting that to undefined or BenchmarkMode.Correctness might be a cleaner way to handle that now.

Josmithr: Since we run everything together in a single test, I don't think this is applicable. If we were running each configuration separately, mapping that way would make more sense. Unless I'm misunderstanding your suggestion.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: Done — Claude removed all `only: false` fields from every `benchmarkIt` call in the file.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: Done — Claude replaced all `as CollectedData` casts with `satisfies CollectedData` throughout the file.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: Done (Claude) — added `significance: "Primary"` to all intercept/base-size measurements throughout the file, so both the slope and intercept from each linearity test contribute to the geometric mean in the console output.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: Done (Claude) — added a TODO comment above the `assertApproximatelyConstant` call to flag this for investigation.

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: This ended up being insufficient. I'll leave a TODO here to investigate.

CraigMacomber (approved):

PR #26849, opened 2026-03-26
build: add ci readiness skill

tylerbutler:
## Summary

Improves the ci-readiness-check skill based on a systematic evaluation using the skill-creator framework. Ran 18 test runs across 3 iterations comparing with-skill vs baseline performance on a test branch with deliberate CI issues (bad formatting, missing copyright header, new public API export without API reports, missing changeset).

### Changes

**Shell script (`ci-readiness-check.sh`):**
- Replaced per-package `pnpm run format` + `flub check policy --fix` loops with `fluid-build --task checks:fix` scoped to changed packages — also covers syncpack and version consistency checks that the old approach missed
- Added `pnpm install --frozen-lockfile` guard for environments without `node_modules` (e.g. fresh git worktrees)
- Added verification step (`fluid-build --task checks`) after auto-fix to confirm all checks pass

**Skill instructions (`SKILL.md`):**
- Clarified Quick mode stops entirely at step 3, even for already-built packages (previously ambiguous, causing agents to overstep into API report regeneration)
- Added TSDoc release tag guidance in step 5: when `ae-missing-release-tag` occurs, add the appropriate tag, rebuild, and retry
- Added "(Full and Thorough only)" headers to steps 5-7 for clarity

### Eval Results

| Mode | With Skill | Without Skill (baseline) |
|------|-----------|------------------------|
| Quick mode | **100%** pass rate (auto-fixes formatting, policy; respects mode boundaries) | 56% (detects issues but never auto-fixes; uses Prettier instead of Biome in 4/6 runs) |
| Full mode | **100%** pass rate (builds, fixes release tags, regenerates API reports, checks cascade) | 59% (detects issues but doesn't fix, build, or regenerate) |

The skill's primary value is **auto-fixing** — both with-skill and baseline runs detected all planted issues, but only the skill actually fixed them. The skill also ensures the correct repo-specific tools are used (Biome, flub, fluid-build) rather than generic alternatives.

### Evaluation Methodology

Created a test branch with 4 deliberate CI issues in `@fluidframework/map`:
1. Bad formatting in `mapKernel.ts` (inconsistent import spacing)
2. Missing copyright header in a new `utils2.ts` file
3. New public export (`countEntries`) added to `index.ts` without API report regeneration
4. No changeset

Each test prompt was run both with the skill and without (baseline), in isolated git worktrees. Assertions checked detection, auto-fixing, use of correct repo tools, mode boundary compliance, and report quality.

## Test plan

- [x] Evaluated with 3 test prompts across Quick and Full modes
- [x] Verified `pnpm install` guard resolves fresh worktree failures (iteration 3)
- [x] Verified Quick mode respects step 3 boundary (iteration 2+)
- [x] Verified Full mode handles `ae-missing-release-tag` with new guidance (iteration 2+)


.claude/skills/ci-readiness-check/SKILL.md
```diff
@@ -0,0 +1,155 @@
...
+- Any file that defines an exported type, interface, class, or function whose **signature** changed (not just the implementation body)
+- The package's `package.json` `exports` field changed
+
+**API did not change — skip the check:**
+- Only test files changed (`src/test/`, `*.spec.ts`, `*.test.ts`)
+- Only internal implementation changed with no exported signature differences
+- Only comments, documentation, or non-code files changed
+- Only a function body changed (not its signature)
+
+**Why be conservative:** There is a known bug in API Extractor where it non-deterministically reorders some generated types. The output differs between local and CI, producing bogus diffs that look like real changes but aren't. Running API Extractor unnecessarily can introduce these confusing phantom diffs. Only run it when you're confident the public API actually changed.
```
noencke: I think this only happens in the tree package? Is that right? If so, we can make this tighter.

.claude/skills/ci-readiness-check/SKILL.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Step 5: API reports and cross-package cascade (Full and Thorough only)
```
noencke, line 95: This is good here for now. Later, I think we should try to write an "API Change" skill, which this section would delegate to, and which would include all the information here. It would also include additional information about API conventions from our best practices, and remind the user to contact the API Council for approval if the API change is above @alpha.

.claude/skills/ci-readiness-check/SKILL.md
```diff
@@ -0,0 +1,155 @@
...
+</required>
+
+# Step 1: Confirm with the user
+
+Before doing anything, ask the user:
+
+> I'll run a CI readiness check on your branch. Pick a mode (fastest to slowest):
+>
+> 1. **Cancel** — never mind, don't run checks
+> 2. **Quick** — auto-fix formatting, policy, and lint; skip all build-dependent checks
```
noencke: We should time the different modes on some changes to common packages. I think it'd be helpful to say something like:
2. **Quick** - ... (takes < 10 seconds)
3. **Full** - ... (takes up to a minute)
4. **Thorough** - ... (fast for small packages, can take minutes for changes to large packages)

.claude/skills/ci-readiness-check/SKILL.md
```diff
@@ -0,0 +1,155 @@
...
+# Step 8: Run tests (Thorough mode only)
+
+If the user chose **Thorough** mode, run tests in each built changed package. Check the package's `package.json` for available test scripts and run whichever exist:
+
+```bash
+cd <package-dir> && pnpm run test:mocha
+cd <package-dir> && pnpm run test:jest
+```
+
+Some packages have both, some have one, some have neither. Only run what exists. If tests fail, report the failures but continue with remaining packages — collect all results for the final report.
```
noencke: "Skip performance tests and real service tests"

Maybe? I'm not sure how many packages it's relevant in but I wouldn't want the agent to get the impression that it needs to run ALL mocha tests it can find, no matter what, in order to complete this step

.claude/skills/ci-readiness-check/SKILL.md
```diff
@@ -0,0 +1,155 @@
...
+---
+
+<required>
+*CRITICAL* Add the following steps to your Todo list using TodoWrite:
+
+1. Confirm with the user before proceeding.
+2. Run the CI readiness script.
+3. Review script output and report results to the user.
+4. If the user chose Full or Thorough mode and there are unbuilt packages, build them.
+5. For each built changed package where the public API surface changed, regenerate API reports and check for downstream cascade.
```
noencke:
```suggestion
5. For each built changed package where the user-facing API surface changed, regenerate API reports and check for downstream cascade.
```

.claude/skills/ci-readiness-check/SKILL.md
```diff
@@ -0,0 +1,155 @@
...
+6. For each built changed package, run ESLint auto-fix.
+7. For each built changed package where the public API surface changed, regenerate type tests.
```
noencke:
```suggestion
7. For each built changed package where the user-facing API surface changed, regenerate type tests.
```

.claude/skills/ci-readiness-check/SKILL.md
noencke: I'm running some timing tests to get some rough data

noencke: Hmm - "quick" mode is taking 100+ seconds consistently, even for small changes. I think we want it to be quicker than that. Does `pnpm checks` normally take that long? Or is it another step in there? I'm asking claude for a breakdown - we ought to move something in there to one of the longer modes.

noencke: looks like it was the biome formatting - even in quick mode, it is formatting every package. Tweaking the script now  to only format _changed_ files and then I'll rerun the timing tests.

tylerbutler: That is one advantage to using git hooks - it is scoped to only changes files.

tylerbutler: But biome formatting is also very fast in general

.claude/skills/ci-readiness-check/ci-readiness-check.sh
```diff
@@ -0,0 +1,255 @@
+#!/usr/bin/env bash
```
tylerbutler, line 1: Since this would presumably run outside codespaces, it would be best if it was cross platform. Maybe use zx? https://google.github.io/zx/getting-started

.claude/skills/ci-readiness-check/SKILL.md
tylerbutler: pnpm checks is designed to run repo-wide checks - the sorts of checks that isolated builds don't catch. Maybe we can make this a pre-push hook? I'm continuing to explore that, but I think we can go either way. The git hook would be a little more deterministic.

noencke: # CI Readiness Check — Benchmark Results

Each cell is the average of 3 runs. All times include the fixed biome-on-changed-files optimization (see [Format Fix](#format-fix) below).

## Change Scenarios

| Scenario | Packages | Description |
|----------|----------|-------------|
| **Small** | `@fluidframework/driver-definitions` (1 pkg) | A leaf-level interface-only package with ~21 source files, no mocha tests, and no downstream re-exports. Represents a minimal, isolated change. |
| **Medium** | `@fluidframework/counter` + `@fluidframework/register-collection` (2 pkgs) | Two small DDS packages (~10 source files each) with full infrastructure: API reports, type tests, ESLint, and mocha tests. Represents a typical multi-package PR. |
| **Large** | `@fluidframework/tree` (1 pkg) | The largest package in the repo — 595 source files, extensive mocha test suite, multiple API extractor configs, and re-exported through `fluid-framework`. Represents a major change to a core package. |

## Total Time by Mode (seconds)

```
                    Small (1 pkg)    Medium (2 pkgs)    Large (tree)
                   ──────────────   ────────────────   ─────────────
Quick                   ~16s             ~16s              ~16s
Full                    ~28s             ~44s              ~99s
Thorough                ~27s             ~46s             ~270s
```

## Bar Chart

```
Quick/Small    ████████ 16s
Quick/Medium   ████████ 16s
Quick/Large    ████████ 16s

Full/Small     ██████████████ 28s
Full/Medium    ██████████████████████ 44s
Full/Large     █████████████████████████████████████████████████ 99s

Thor./Small    ██████████████ 27s
Thor./Medium   ███████████████████████ 46s
Thor./Large    ████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████ 270s
```

## Per-Step Breakdown (seconds, avg of 3 runs)

| Step | Small | Medium | Large | Notes |
|------|------:|-------:|------:|-------|
| **Script (Quick)** | 16 | 16 | 16 | Fixed overhead — does not scale with change size |
| ↳ biome format | 1 | 1 | 1 | Single invocation on changed files only |
| ↳ policy check | 7 | 7 | 7 | `flub check policy --fix` + verify |
| ↳ changeset check | 3 | 3 | 3 | `flub check changeset` |
| ↳ pkg detection | 5 | 5 | 5 | git diff + `get_pkg_name` × N packages |
| **Build (tsc)** | 2 | 7 | 10 | Scales with package size |
| **API Reports** | 3 | 5 | 5 | API Extractor, multiple configs for tree |
| **ESLint fix** | 4 | 10 | 63 | Tree's 595 files dominate here |
| **Type tests** | 2 | 5 | 3 | `typetests:gen` |
| **Tests (mocha)** | 0* | 3 | 173 | Tree's test suite is ~3 minutes |

\* `driver-definitions` has no mocha tests.

## Key Findings

1. **Quick mode is constant ~16s** regardless of change size. The cost is dominated by `flub check policy` (~7s) and per-package name resolution (~5s). Biome formatting on only changed files adds <1s.

2. **Full mode scales with package complexity.** The main variable is ESLint — tree alone takes 63s. Small packages add only ~12s on top of Quick.

3. **Thorough mode is dominated by test execution for large packages.** Tree's `test:mocha` takes ~173s (~3 min). For small packages with no tests, Thorough ≈ Full.

4. **The ~16s Quick overhead is baked into every Full and Thorough run** as the baseline script cost.

## Suggested Time Estimates

| Mode | Estimate | What it does |
|------|----------|--------------|
| **Quick** | ~15–20 seconds | Format, policy, changeset check. Constant time. |
| **Full** | ~30s – 2 minutes | Quick + build + API reports + ESLint + type tests. Scales with package count and size. |
| **Thorough** | ~30s – 5+ minutes | Full + run tests. Dominated by test suite size for large packages. |

## Format Fix

The original script ran `pnpm run format` per-package, spawning a separate
node/pnpm process for each. With 160 changed packages, this took
**88 seconds** — 85% of the total Quick mode time.

The fix runs `biome check --write` once on just the changed files list,
completing in **<1 second**. This reduced Quick mode from ~103s to ~16s.

### Before

```bash
# For each changed package that has a "format" script in its package.json,
# run it. This invokes Biome to auto-fix formatting issues in-place.
FORMATTED=0
FORMAT_ERRORS=0
for pkg in "${PACKAGES[@]}"; do
    pkg_dir="${REPO_ROOT}/${pkg}"
    has_format=$(cd "${pkg_dir}" && \
        node -p "Boolean(require('./package.json').scripts?.format)" \
        2>/dev/null || echo "false")
    if [ "${has_format}" = "true" ]; then
        if (cd "${pkg_dir}" && pnpm run format 2>&1) >/dev/null 2>&1; then
            ok "Formatted ${pkg}"
            FORMATTED=$((FORMATTED + 1))
        else
            fail "Format failed in ${pkg}"
            FORMAT_ERRORS=$((FORMAT_ERRORS + 1))
        fi
    fi
done
```

### After

```bash
# Run Biome directly on just the changed files instead of per-package.
# This avoids 160+ sequential pnpm/node invocations and completes in <1s.
FORMAT_FILES=$(echo "${CHANGED_FILES}" \
    | grep -E '\.(ts|tsx|js|jsx|json|css|jsonc)$' || true)
if [ -n "${FORMAT_FILES}" ]; then
    FORMAT_COUNT=$(echo "${FORMAT_FILES}" | wc -l | tr -d ' ')
    if echo "${FORMAT_FILES}" \
        | xargs pnpm exec biome check --write 2>&1 | tail -1; then
        ok "Checked ${FORMAT_COUNT} changed files"
    else
        warn "Biome reported issues on some files (auto-fixed what it could)"
    fi
else
    ok "No formattable files changed"
fi
```

noencke: These results are pretty great - just what I was hoping for, more or less. Lots of flexibility about how much validation you want to do, and you can tailor it depending on whether you're working in a Beefy package or not.

noencke: @tylerbutler the only thing I don't like about the pre-push hook is that it's not optional. When I'm e.g. making a small change outside of an agent, and I'm confident in it, I'd rather not have to wait 15 seconds to open a PR.

.claude/skills/ci-readiness-check/ci-readiness-check.sh
noencke: If we use it as a git hook, then yea. If not, then probably not - I don't think you can run claude (or copilot? not sure actually) in an environment that doesn't have bash.

I vote that we decide if we want to make it a git hook as a followup later. If so, at that time, we can also convert it to a cross-platform script and test it in both environments.

noencke (approved):

PR #26855, opened 2026-03-26
unhydrated array node delta payload fix

daesunp:
## Description

Fixes delta being `undefined` for unhydrated array nodes in `nodeChanged` events.

Populates the delta field in `NodeChangedDataDelta` for unhydrated array nodes.

### DIFF
--- a/.changeset/good-dryers-strive.md
+++ b/.changeset/good-dryers-strive.md
@@ -1,0 +1,7 @@
+---
+"@fluidframework/tree": minor
+"__section": fix
+---
+Array node nodeChanged delta payload now populated for unhydrated nodes
+
+The `delta` field in `NodeChangedDataDelta` was previously `undefined` for array nodes that had not yet been inserted into a document tree (unhydrated nodes). It now correctly reflects the operation that was performed, using the same [Quill](https://quilljs.com/docs/)-style semantics as hydrated nodes.
--- a/packages/dds/tree/src/simple-tree/core/unhydratedFlexTree.ts
+++ b/packages/dds/tree/src/simple-tree/core/unhydratedFlexTree.ts
@@ -11,6 +11,8 @@ import { UsageError } from "@fluidframework/telemetry-utils/internal";
 import {
 	type AnchorEvents,
 	dummyRoot,
+	type DeltaDetachedNodeId,
+	type DeltaMark,
 	type FieldKey,
 	type FieldKindIdentifier,
 	type ITreeCursorSynchronous,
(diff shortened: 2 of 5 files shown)


.changeset/good-dryers-strive.md
```diff
@@ -0,0 +1,7 @@
+---
+"@fluidframework/tree": minor
+"__section": fix
+---
+Array node nodeChanged delta payload now populated for unhydrated nodes
```
Josmithr, line 5: Have we released since your last changeset went in? If not, it might make more sense to just update the existing changeset file to omit any mention of this limitation. Otherwise, the resulting changelog might give readers some whiplash 😜

packages/dds/tree/src/simple-tree/core/unhydratedFlexTree.ts
```diff
@@ -527,6 +531,53 @@ class UnhydratedRequiredField
 	}
 }
 
+// Dummy detached node ID used when building unhydrated delta marks.
+// deltaMarksToArrayOps only checks presence (not value) of attach/detach, so any value works.
+const dummyDetachedNodeId = { minor: 0 } satisfies DeltaDetachedNodeId;
+
+/**
+ * Builds delta marks for an intra-field move expressed as sequential Quill-style ops.
```
Josmithr: Nit: I'm not sure I like us referencing Quill in internal implementation details like this. I think it's fine to just reference (and maybe link to) delta mark types, which are hopefully documented well enough to explain the scenario.

packages/dds/tree/src/simple-tree/core/unhydratedFlexTree.ts
```diff
@@ -527,6 +531,53 @@ class UnhydratedRequiredField
...
+ *
+ * @remarks
+ * A forward move (`sourceIndex < destIndex`) is expressed as:
+ * `retain(sourceIndex), remove(count), retain(between), insert(count)`
+ *
+ * A backward move (`destIndex < sourceIndex`) is expressed as:
+ * `retain(destIndex), insert(count), retain(between), remove(count)`
```
Josmithr: Nit: these could be expressed as two separate `@example` blocks

packages/dds/tree/src/simple-tree/core/unhydratedFlexTree.ts
```diff
@@ -527,6 +531,53 @@ class UnhydratedRequiredField
...
+ */
+function buildUnhydratedMoveMarks(
+	sourceIndex: number,
+	count: number,
+	destIndex: number,
+): readonly DeltaMark[] {
+	const marks: DeltaMark[] = [];
```
Josmithr, line 559: Should this function be asserting any invariants for the input params?

.changeset/good-dryers-strive.md
Josmithr: We have since released a new version.

packages/dds/tree/src/simple-tree/core/unhydratedFlexTree.ts
```diff
@@ -536,20 +536,26 @@ class UnhydratedRequiredField
...
+ * ```
  */
 function buildUnhydratedMoveMarks(
 	sourceIndex: number,
 	count: number,
 	destIndex: number,
 ): readonly DeltaMark[] {
+	assert(sourceIndex >= 0, 0xbb9 /* sourceIndex must be non-negative */);
+	assert(count > 0, 0xbba /* count must be positive */);
+	assert(destIndex >= 0, 0xbbb /* destIndex must be non-negative */);
```
Josmithr, line 558: New asserts generally shouldn't have codes already. These are generated later. These should presumably just use simple string messages.

PR #26862, opened 2026-03-27
Move non-tree dependent tree-agent functions to new package

brrichards:
# Summary

As part of the work to expose text APIs to tree-agent, relevant tree-agent functions and others have been moved to a separate package to work around a circular dependency that arose between tree-agent and tree. 

### Moved to tree-agent-types

| Export | Source file |
|---|---|
| `MethodKeys`, `Ctor`, `Arg`, `FunctionDef`, `buildFunc`, `exposeMethodsSymbol`, `ExposedMethods`, `IExposedMethods` | `methodBinding.ts` |
| `exposePropertiesSymbol`, `ExposedProperties`, `IExposedProperties` | `propertyBinding.ts` |
| `TypeFactoryType`, all TypeFactory interfaces, `isTypeFactoryType`, `typeFactory` (base) | `treeAgentTypes.ts` |
| Base type factory tests | `typeFactory.spec.ts` |

### Stayed in tree-agent

| Export | Reason |
|---|---|
| `BindableSchema`, `isBindableSchema`, `FunctionWrapper`, `getExposedMethods`, `ExposedMethodsI` | Depends on `TreeNodeSchema`/`NodeKind` |
| `PropertyDef`, `getExposedProperties`, `ExposedPropertiesI` | Depends on `TreeNodeSchema` |
| `TypeFactoryInstanceOf`, `typeFactory` (extended with `instanceOf`) | Depends on `ObjectNodeSchema` |
| `ArgsTuple` | Stayed in tree-agent |

### Signature changes

| Change | Before | After | Reason |
|---|---|---|---|
| `ExposedMethods.expose` `S extends` | `BindableSchema & Ctor & IExposedMethods` | `Ctor & IExposedMethods` | Can't reference `BindableSchema` from tree-agent-types; semantically equivalent since `getExposedMethods` gates on `BindableSchema` |
| `ExposedProperties.exposeProperty` `S extends` | `BindableSchema & Ctor` | `Ctor` | Same reason |
| `ExposedMethodsI.expose` equality check | `schema !== this.schemaClass` | `(schema as unknown) !== (this.schemaClass as unknown)` | Types no longer overlap after loosening, needs cast |
| `ExposedPropertiesI.exposeProperty` equality check | same direct comparison | same `as unknown` cast | Same reason |

### Tag changes

| Export | Before | After | Reason |
|---|---|---|---|
| `typeFactory` in tree-agent-types | `@alpha` | `@internal` | Single public alpha export from tree-agent only |

### Layer override

- `tree-agent-types` lives in `packages/framework/` but is explicitly listed in the `Runtime` layer in `layerInfo.json`. Normally `packages/framework/` maps to the `Framework` layer, which would create a layer violation since `@fluidframework/tree` (Runtime layer) depends on it. Placing it in Runtime is safe because tree-agent-types has no Framework-layer dependencies. Its only runtime dep is `@fluidframework/telemetry-utils` which sits in a lower layer.

### Design decisions

- The `BindableSchema` constraint loosening on `ExposedMethods`/`ExposedProperties` is semantically equivalent because `getExposedMethods`/`getExposedProperties` still require `BindableSchema` at the entry point
- tree-agent-types only runtime dependency is `@fluidframework/telemetry-utils` (for `UsageError` in typeFactory validation)


packages/framework/tree-agent-types/src/methodBinding.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * An interface for exposing methods of schema classes to an agent.
+ * @alpha
+ */
+export interface ExposedMethods {
+	/**
+	 * Expose a method with type factory types.
+	 */
+	expose<
```
noencke: This might be a good opportunity to change this to `exposeMethod` rather than `expose` - then it has symmetry with `exposeProperty`. It's not that way currently because we wrote the method API first and didn't think about properties. What do you think?

packages/framework/tree-agent-types/src/methodBinding.ts
```diff
@@ -0,0 +1,120 @@
...
+		const K extends string & keyof MethodKeys<InstanceType<S>>,
+		S extends Ctor & IExposedMethods,
+		Z extends FunctionDef<readonly Arg[], TypeFactoryType, TypeFactoryType | null>,
+	>(schema: S, methodName: K, tfFunction: Z): void;
```
noencke: nit: anywhere that it says "schema" in these APIs we can change to just be "class" since it's no longer (technically) specific to ST classes.

packages/framework/tree-agent/api-report/tree-agent.alpha.api.md
```diff
@@ -236,150 +161,15 @@ export const typeFactory: {
     object(shape: Record<string, TypeFactoryType>): TypeFactoryObject;
     record(keyType: TypeFactoryType, valueType: TypeFactoryType): TypeFactoryRecord;
     map(keyType: TypeFactoryType, valueType: TypeFactoryType): TypeFactoryMap;
-    tuple(items: readonly TypeFactoryType[], rest?: TypeFactoryType): TypeFactoryTuple;
+    tuple(items: readonly TypeFactoryType[], rest?: TypeFactoryType | undefined): TypeFactoryTuple;
```
noencke, line 164: What's the significance of this change?

noencke (approved):
I think we can market this package as a generic tool since you had to decouple it from tree. Instead of `tree-agent-types`, it could be `type-factory` perhaps? Or something else?

It's possible we might even want to use it outside of tree-agent. It's a nice way to define methods that we could use for an LLM library other than tree-agent.

 Overall this looks good though.

packages/framework/tree-agent/api-report/tree-agent.alpha.api.md
brrichards: I was getting typescript errors in a previous change, but I'll relook since I've done more changes to this function since then

packages/framework/tree-agent-types/src/methodBinding.ts
brrichards: I'll go ahead and update this

packages/framework/tree-agent-types/src/methodBinding.ts
brrichards: That makes sense to me! I can go ahead and update it to match exposeProperty

packages/framework/tree-agent-types/src/propertyBinding.ts
```diff
@@ -0,0 +1,62 @@
...
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import type { Ctor } from "./methodBinding.js";
+import type { TypeFactoryType } from "./treeAgentTypes.js";
+
+/**
+ * A symbol used to expose properties to the LLM.
+ * @alpha
+ */
+export const exposePropertiesSymbol: unique symbol = Symbol.for(
```
brrichards: I noticed that this is different from `exposeMethodSymbol` which doesn't use `Symbol.for()` and only uses `Symbol("run")`. Curious if this is another thing we want to make symmetrical to exposeProperties, or if there a needed difference between the two. Curious what you think @noencke

noencke: Hmm good question. I'm not sure. `symbol.for` lets you re-create the symbol from the string. I don't think we expect our users to do that. I'd say just make them Symbol("Type factory expose method/symbol").

packages/framework/tree-agent/api-report/tree-agent.alpha.api.md
brrichards: I was wrong and this change was not related to those errors. I think this happened because of how the API extractor renders optional params when they come from a spread object. That being `(...baseTypeFactory)`.

packages/framework/tree-agent-types/src/methodBinding.ts
brrichards: unfortunately class can't be used as a parameter name in js/ts and I don't know if theres something better than schema. we could use cls or ctor, but Ctor is already used

noencke (approved):

PR #19193, opened 2024-01-10
TableBench: Basic comparisons between SharedTree and SharedMatrix

DLehenbauer:
Basic comparisons between SharedTree and SharedMatrix

PR #26866, opened 2026-03-27
Update benchmark package in client

CraigMacomber:
## Description

Update code and scripts for new benchmark package.

## Breaking Changes

This changes the report format for benchmark reports, as well as the scripts for running them.

## Reviewer Guidance


CraigMacomber:
I think this is ready, but doing a round of AI assisted review first, and will also be running the new pipeline to confirm that works. This may require moving to an upstream test branch.

packages/dds/tree/src/test/memory/tableTree.spec.ts
```diff
@@ -42,50 +40,36 @@ function runBenchmark({
...
-				this.cleanUp();
-				this.undoRedoStack = undefined;
-				this.cleanUp = undefined;
-			}
-		})(),
-	);
+					afterOperation?.(table, undoRedoStack);
+					cleanUp();
+				}
+			},
```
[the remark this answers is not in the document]
CraigMacomber: Its not needed, and doesn't seem to help here.

CraigMacomber:
Internal benchmark collection pipeline is passing for this branch, see run at https://dev.azure.com/fluidframework/internal/_build/results?buildId=390932&view=results

Note that the new pipeline is over 3 times faster the the onld one, drastically less code, and no longer has to wait for the client build before starting.

These savings come from some parallelization, test tweaks, better memory testing, and skipping perspective tests.

examples/benchmarks/tablebench/src/test/table.bench.spec.ts
```diff
@@ -87,29 +94,29 @@ describe("Table", () => {
...
-
-					const totalRevenue = unitsSold * unitPrice;
-					const totalCost = unitsSold * unitCost;
-					const totalProfit = totalRevenue - totalCost;
-
-					row["Total Revenue"] = totalRevenue;
-					row["Total Cost"] = totalCost;
-					row["Total Profit"] = totalProfit;
-				}
+				// Batching these updates in a transaction gives a about a 3x performance boost
```
CraigMacomber, line 97: In made this change in this PR because I kepy running perf tests to test things, and these took a lot less time to run after this change. Seems fine to keep.

tools/pipelines/test-perf-benchmarks.yml
```diff
@@ -70,547 +70,167 @@ variables:
...
-              tar --extract --verbose --file $TAR_PATH ./lib/test
-              tar --extract --verbose --file $TAR_PATH ./dist/test
-              tar --extract --verbose --file $TAR_PATH ./src/test
-
-        # Install package dependencies (this gets devDependencies installed so we can then run the package's tests).
-        - template: /tools/pipelines/templates/include-copy-dev-dependencies.yml@self
-          parameters:
-            sourcePackageLocation: ${{ variables.testWorkspace }}/node_modules/${{ testPackage }}
-            destPackageLocation: $(Pipeline.Workspace)/$(FFPipelineHostDirectory)
+              pnpm exec fluid-build --nolint --task build:test:esm --worker
```
CraigMacomber, line 170: This takes just under 2 minutes on CI.

It is not ideal to build this in every job, but doing so is way simpler than the old setup of installing pre build stuff from the client build then mergin that with the test code. Given the new pipeline is still way faster than the old one, and this build is only less than 10% of the time, I think its worth the simplification.

tools/pipelines/test-perf-benchmarks.yml
```diff
@@ -70,547 +70,167 @@ variables:
...
+        # Install FluidFramework dependencies
+        - template: /tools/pipelines/templates/include-install.yml@self
           parameters:
-            artifactBuildId: $(artifactBuildId)
-            artifactPipeline: $(artifactPipeline)
-            testPackageName: ${{ testPackage }}
-            installPath: $(testWorkspace)
+            packageManager: pnpm
+            buildDirectory: $(FluidFrameworkDirectory)
+            packageManagerInstallCommand: pnpm install --frozen-lockfile
```
Abe27342: nit: be consistent about whether you specify the extra flag here and above. I could see arguments for both options given in each workspace we already declare frozen-lockfile at the .npmrc level. Personally would lean toward omitting it.

packages/dds/sequence/src/test/memory/sharedString.spec.ts
```diff
@@ -27,147 +29,138 @@ function createLocalSharedString(id: string) {
...
-						this.sharedString.annotateMarker(this.simpleMarker, { color: "blue" });
+				},
+			}),
+		});
+
+		benchmarkIt({
+			title: `Annotate marker ${x} times with same options`,
+			...benchmarkMemoryUse({
+				benchmarkFn: async (state) => {
+					while (state.continue()) {
```
Abe27342: This test seems like it intends to measure memory impact of just the marker operations in L155-L157. I haven't been following all the implementation changes in benchmark, but from the API it doesn't seem like the placement of `beforeAllocation()` in the new diff respects that, seems like it would be measuring the size of the sharedstring in addition to those markers.

From what I can tell every memory test follows this boilerplate:

```
async (state) => {
	while (state.continue()) {
		await state.beforeAllocation();
		{
			// do stuff
			await state.whileAllocated();
			// maybe do some cleanup
		}
		await state.afterDeallocation()
	}
}
```

which feels wrong, I would have expected the setup here to be more like

```
async (state) => {
	while (state.continue()) {
		// make sharedstring, generally the things previously in beforeIteration
		await state.beforeAllocation();
		{
			// call annotateMarker a bunch
			await state.whileAllocated();
			// cleanup
		}
		await state.afterDeallocation()
	}
}
```

CraigMacomber: I honestly can't tell what the old code was trying to test as the old setup was very confusing and didn't really do what it looked like it did IIRC. If I ignore the old code, and just assume your understanding is correct and just try and make this test make sense, I can adjust it accordingly.

Abe27342: I think even looking at structure of old code without thinking of implementation details of what that meant in benchmark gives pretty clear intent. There are a fair number of other memory tests in DDS packages which:

- Are named in a way which reflects intent to measure a particular operation on that DDS
- Have their "run()" in the old code do that operation and only that operation
- Have their "beforeIteration()" in the old code set up the rest of the boilerplate to create the DDS

I think all of these should be adjusted similarly.

tools/pipelines/test-perf-benchmarks.yml
CraigMacomber: omitted.

packages/dds/sequence/src/test/memory/sharedString.spec.ts
CraigMacomber: I have cleaned them up. Should be better now. I'll give another look to the other packages as well. Clearly I didn't give ll of them a good enough look initially.

examples/benchmarks/tablebench/src/test/table.bench.spec.ts
Josmithr, line 97: I hadn't seen these benchmarks prior to just now. These seem to be testing a bespoke table implementation that lives in this package. I'm assuming these are fairly old?

packages/dds/matrix/src/test/matrixBenchmarks.ts
```diff
@@ -207,47 +203,28 @@ function runMemoryBenchmark({
...
+		...benchmarkMemoryUse({
+			benchmarkFn: async (state) => {
+				while (state.continue()) {
+					await state.beforeAllocation();
+					{
+						const { matrix, undoRedoStack, cleanUp } = createTestMatrix({
+							matrixSize,
+							initialCellValue,
+						});
+						beforeOperation?.(matrix, undoRedoStack);
```
Josmithr: Maybe I'm misunderstanding the way the new APIs work here, but shouldn't these lines occur before the `await state.beforeAllocation();` call? The goal is to measure the memory footprint of what happens in `operation` only, excluding the before and after steps.

packages/dds/tree/src/test/memory/tableTree.spec.ts
```diff
@@ -42,50 +40,36 @@ function runBenchmark({
...
+		}),
+	});
 }
 
 describe("SharedTree table APIs memory usage", () => {
-	// When run as benchmarks, these tests are very slow.
-	// This should speed them up a little, and make them a little more production like for better data.
+	// This is more because these tests are so slow than to make the data more production like.
```
Josmithr: Nit: I don't think this is very descriptive. More detail would be useful.

packages/dds/tree/src/test/memory/tree.spec.ts
```diff
@@ -113,117 +113,117 @@ function createLocalSharedTree<TSchema extends ImplicitFieldSchema>(
...
-
-				public async run(): Promise<void> {
-					assert(this.sharedTree?.root.child !== undefined);
-
-					for (let i = 0; i < numberOfEntries; i++) {
-						this.sharedTree.root.child.propertyOne = numberOfEntries;
+		benchmarkIt({
+			title: `Set an integer property ${numberOfEntries} times in a local SharedTree`,
+			...benchmarkMemoryUse({
+				keepIterations: 4,
```
Josmithr: In the table benchmark suite above, you left a nice comment explaining why we were setting these. Would be nice to add a similar comment here.

packages/test/mocha-test-setup/mocharc-common.cjs
```diff
@@ -64,29 +64,52 @@ function getFluidTestMochaConfig(packageDir, additionalRequiredModules, testRepo
 		spec: process.env.MOCHA_SPEC ?? defaultSpec,
 	};
 
-	if (process.env.FLUID_TEST_TIMEOUT !== undefined) {
-		config["timeout"] = process.env.FLUID_TEST_TIMEOUT;
+	if (process.env.FLUID_TEST_PERF_MODE !== undefined) {
```
Josmithr, line 71: Nit: seems unlikely, but theoretically this could be set explicitly to false. A truthy check would probably be better?

package.json
```diff
@@ -100,16 +100,15 @@
 		"syncpack:versions:fix": "syncpack fix",
 		"test": "npm run test:mocha && npm run test:jest && npm run test:realsvc && npm run test:stress",
 		"test:bail": "npm run test:mocha:bail && npm run test:jest:bail",
-		"test:benchmark:copyresults": "copyfiles --exclude \"**/node_modules/**\" \"**/benchmarkOutput/**\" benchmarkOutput",
-		"test:benchmark:report": "pnpm run -r --no-sort --stream --no-bail test:benchmark:report --reporterOptions reportDir=./benchmarkOutput",
+		"test:benchmark:copyresults": "copyfiles --exclude \"**/node_modules/**\" \"**/benchmark*Output.json\" benchmarkOutput",
+		"test:benchmark:duration:report": "cross-env FLUID_TEST_VERBOSE=1 pnpm run -r --workspace-concurrency=4 --no-sort --stream --no-bail bench -- --fail-zero=false --grep \"@Benchmark @Measurement @ExecutionTime\" --reporterOptions reportFile=./benchmarkDurationOutput.json",
+		"test:benchmark:other:report": "cross-env FLUID_TEST_VERBOSE=1 pnpm run -r --no-sort --stream --no-bail bench -- --fail-zero=false --grep \"@Benchmark @Measurement(?!.*@ExecutionTime)\" --reporterOptions reportFile=./benchmarkOtherOutput.json",
```
alexvy86, line 105: Wondering if this has to do with the fact that some e2e memory tests didn't seem to run (maybe the ones from map as well). I recall that way back, Nicholas wrote some wrapper so a single test definition could work for both memory and duration benchmarking, so maybe they're ending up with this tag in their name and getting excluded even when we want to run memory tests?

packages/dds/shared-object-base/package.json
```diff
@@ -82,7 +83,7 @@
 		"format:biome": "biome check . --write",
 		"lint": "fluid-build . --task lint",
 		"lint:fix": "fluid-build . --task eslint:fix --task format",
-		"perf": "cross-env FLUID_TEST_VERBOSE=1 mocha --perfMode --fgrep @Benchmark --reporter @fluid-tools/benchmark/dist/MochaReporter.js --timeout 30000",
+		"perf": "cross-env FLUID_TEST_VERBOSE=1 FLUID_TEST_PERF_MODE=1 mocha --fgrep @Benchmark --timeout 30000",
```
alexvy86: Is there a need to keep `perf` here? We removed it in other packages, and `bench` seems to be enough going forward, right?

packages/dds/merge-tree/package.json
```diff
@@ -94,11 +95,8 @@
 		"lint": "fluid-build . --task lint",
 		"lint:fix": "fluid-build . --task eslint:fix --task format",
 		"pack:tests": "tar -cvf ./merge-tree.test-files.tar ./src/test ./dist/test ./lib/test",
-		"perf": "cross-env FLUID_TEST_VERBOSE=1 mocha \"dist/**/*.spec.*js\" --node-option unhandled-rejections=strict,expose-gc --exit --perfMode --fgrep @Benchmark --fgrep @ExecutionTime --reporter @fluid-tools/benchmark/dist/MochaReporter.js --timeout 30000",
-		"perf:measure": "npm run perf -- --fgrep @Measurement",
-		"perf:profile": "node --inspect-brk ./node_modules/mocha/bin/mocha.js \"dist/**/*.spec.*js\" --node-option unhandled-rejections=strict,expose-gc --exit --perfMode --fgrep @Benchmark --fgrep @ExecutionTime --reporter @fluid-tools/benchmark/dist/MochaReporter.js --timeout 30000",
+		"perf:profile": "cross-env FLUID_TEST_PERF_MODE=1 node --inspect-brk ./node_modules/mocha/bin/mocha.js \"dist/**/*.spec.*js\" --node-option unhandled-rejections=strict,expose-gc --exit --fgrep @ExecutionTime",
```
alexvy86, line 98: Another `perf` which I think can go away?

packages/dds/matrix/package.json
```diff
@@ -98,10 +99,7 @@
 		"lint:fix": "fluid-build . --task eslint:fix --task format",
 		"pack:tests": "tar -cf ./matrix.test-files.tar ./src/test ./dist/test ./lib/test",
 		"test": "npm run test:mocha",
-		"test:benchmark:report": "mocha --config src/test/time/.mocharc.cjs",
 		"test:coverage": "c8 npm test",
-		"test:memory": "mocha --config src/test/memory/.mocharc.cjs",
```
alexvy86, line 103: We kept `test:memory` in sequence and in map. If `bench` runs everything, like I think it does, seems fine to remove it, but then we can also do it in the other packages, no?

packages/common/core-utils/package.json
```diff
@@ -60,8 +60,8 @@
 		"api": "fluid-build . --task api",
 		"api-extractor:commonjs": "flub generate entrypoints --resolutionConditions require --outFileLegacyBeta legacy --outDir ./dist",
 		"api-extractor:esnext": "flub generate entrypoints --outFileLegacyBeta legacy --outDir ./lib --node10TypeCompat",
-		"bench": "mocha --timeout 999999 --perfMode --parentProcess --fgrep @Benchmark --fgrep @ExecutionTime --reporter @fluid-tools/benchmark/dist/MochaReporter.js",
-		"bench:profile": "mocha --v8-prof --v8-logfile=profile.log --v8-no-logfile-per-isolate --timeout 999999 --perfMode --fgrep @Benchmark --fgrep @ExecutionTime --reporter @fluid-tools/benchmark/dist/MochaReporter.js && node --prof-process profile.log > profile.txt && rimraf profile.log && echo See results in profile.txt",
+		"bench": "cross-env FLUID_TEST_PERF_MODE=1 mocha --parentProcess",
+		"bench:profile": "cross-env FLUID_TEST_PERF_MODE=1 mocha --v8-prof --v8-logfile=profile.log --v8-no-logfile-per-isolate --timeout 999999 --fgrep @ExecutionTime --reporter @fluid-tools/benchmark/dist/mocha/Reporter.js && node --prof-process profile.log > profile.txt && rimraf profile.log && echo See results in profile.txt",
```
alexvy86: Those should come for free with FLUID_TEST_PERF_MODE, right?
```suggestion
		"bench:profile": "cross-env FLUID_TEST_PERF_MODE=1 mocha --v8-prof --v8-logfile=profile.log --v8-no-logfile-per-isolate --fgrep @ExecutionTime && node --prof-process profile.log > profile.txt && rimraf profile.log && echo See results in profile.txt",
```

packages/dds/map/src/test/memory/directory.spec.ts
```diff
@@ -21,76 +23,66 @@ function createLocalDirectory(id: string): ISharedDirectory {
...
-					for (let i = 0; i < x; i++) {
-						this.dir.set(i.toString().padStart(6, "0"), i);
+		benchmarkIt({
+			title: `Add ${x} integers to a local directory`,
+			...benchmarkMemoryUse({
+				benchmarkFn: async (state) => {
+					while (state.continue()) {
+						await state.beforeAllocation();
+						{
+							const dir = createLocalDirectory("testDirectory");
```
alexvy86: Note: I think the test as written before tried to measure the effect just of the `set()` calls, and now will include the creation of the directory itself. Given our memory tests are still overall quite immature, I don't mind the change, but calling it out in case it informs any further updates to the benchmark tool.

packages/dds/matrix/src/test/matrixBenchmarks.ts
alexvy86: I made a similar comment in my review about other memory tests.

packages/dds/sequence/src/test/memory/sharedString.spec.ts
alexvy86: Ah, just now ran into this one; Josh and I also pointed it out for other tests.

packages/dds/tree/package.json
```diff
@@ -124,12 +125,8 @@
 		"lint:fix": "fluid-build . --task eslint:fix --task format",
 		"pack:tests": "tar -cf ./tree.test-files.tar ./src/test ./dist/test ./lib/test",
 		"test": "npm run test:mocha",
-		"test:benchmark:report": "mocha --exit --perfMode --parentProcess --fgrep @Benchmark --fgrep @ExecutionTime --reporter @fluid-tools/benchmark/dist/MochaReporter.js --timeout 60000",
 		"test:coverage": "c8 npm test",
-		"test:customBenchmarks": "mocha --config ./.mocharc.customBenchmarks.cjs",
-		"test:customBenchmarks:verbose": "cross-env FLUID_TEST_VERBOSE=1 npm run test:customBenchmarks",
-		"test:memory": "mocha --perfMode --config ./src/test/memory/.mocharc.cjs",
-		"test:memory-profiling:report": "mocha --config ./src/test/memory/.mocharc.cjs",
+		"test:memory": "cross-env FLUID_TEST_PERF_MODE=1 mocha --fgrep @Memory --parallel --jobs 6",
```
alexvy86, line 129: Another `test:memory` that I think should go, if `bench` runs them all already, or should stay across all packages. In `tree` in particular it also feels like if we keep it, we might want to keep one for custom benchmarks too?

tools/pipelines/test-perf-benchmarks.yml
```diff
@@ -3,65 +3,65 @@
 
 name: $(Build.BuildId)
 
-trigger: none
+trigger:
```
alexvy86, line 6: The change from `resources.pipelines.pipeline.trigger` to this means that the pipeline will now run in parallel to the build one for every commit. I get that's possible because this one doesn't have to download artifacts from the build pipeline anymore, which is great. I'd raise the question if we want to keep triggering this one only after the build pipeline succeeds, so we don't waste the compute if the build is going to fail (although that shouldn't be common in the internal ADO project). Personally, I'm fine giving the new approach a try so fine to merge by me

tools/pipelines/test-perf-benchmarks.yml
```diff
@@ -70,547 +70,167 @@ variables:
...
-
-              if [[ $memoryTestsExitCode -ne 0 ]]; then
-                echo "##vso[task.logissue type=error]Exit code for memory tests execution = $memoryTestsExitCode ${{ endpointObject.endpointName }}"
-              fi
-
-              if [[ $executionTimeTestsExitCode -ne 0 ]] || [[ $memoryTestsExitCode -ne 0 ]]; then
-                exit 1;
-              fi
-          env:
-            FLUID_TEST_LOGGER_PKG_SPECIFIER: '@ff-internal/aria-logger' # Contains createTestLogger impl to inject
```
alexvy86, line 543: Without the `FLUID_TEST_LOGGER_PKG_SPECIFIER` we don't get standard Fluid telemetry from when the tests are running. I confirmed that the test run of the new setup didn't push any Fluid telemetry to Kusto (benchmark results telemetry is not affected by this). And with aria-logger now installed in a folder/package/workspace other than the one where the tests run, I don't know off the top of my head how we could specify this so things work.

That said, I think any investigation of benchmark tests is more probably going to happen locally, so maybe losing this is not terrible. Interested in more thoughts.

Abe27342: It's still available on disk with the new pipeline setup, so there shouldn't be any technical blockers for sending telemetry while tests are running assuming we still want that. Just need to make sure we set the right env variables (and import the package from that location) from within the test processes.

alexvy86: My concern is due to having dealt with this variable myself in the past and changing it from being a path to explicitly being just a package name, which IIRC is expected to be found by the standard module resolution algorithm (so in node_modules somewhere in the path from `.` to the root of the drive). At this point I don't recall the reason off the top of my head, just that it probably had to do with resolution/import sheananigans. I'm sure it could be fixed, just not sure if trivially so.

CraigMacomber: This change was intentional. Module resolution fails if we enable this, and without it we fall back to the default (null logger IIRC) which seems better IMO anyway. To get the custom logger work this work how it was requires some nontrivial hacks to get the aria logger into node_modules. I had that working, but it was messy and I realized that I don't think we need it.

As the performance tests now also run in shorter versions in correctness mode, having them log end to end test telemetry seemed like a possibly confusing thing to do as it would add a bunch of telemetry for the same tests in when running in performance mode which already do it in correctness mode, which might be confusing.

Given we now include error results robustly in the performance telemetry, it seems pretty safe only include the other telemetry in non-performance runs. That said, I'm not super clear on the implications of including or not including this telemetry.

tools/pipelines/test-perf-benchmarks.yml
CraigMacomber: Generally build on post merge main should be green most of the time, and most of the errors that happen when its in a bad state shouldn't make running perf tests worthless (ex: a lint error, or an error in one package).

Also currently this pipeline is probable cheaper (or at least similar) in resource cost than that the client-build pipeline, as its just 4 jobs which are shorter than client build. That fact that we only build esm, and don't lint saves a ton of time.

CraigMacomber: Thus I think leaving it as is makes sense, though maybe I should tweak its trigger paths to match client.

examples/benchmarks/tablebench/src/test/table.bench.spec.ts
CraigMacomber: Daniel wrote then ~ 2 years ago: https://github.com/microsoft/FluidFramework/pull/19193

packages/dds/map/src/test/memory/directory.spec.ts
CraigMacomber: I think before it might have measured the wrong thing, but either way I have now fixed these to be much nicer and more clear and do what they should. Should be better now: let me know if you see any other similar such issues.

packages/dds/matrix/src/test/matrixBenchmarks.ts
CraigMacomber: Should be fixed now.

packages/dds/tree/package.json
CraigMacomber: This is useful if a dev want to run just the memory tests for tree locally. Tree's full test suite is quite slow, and has a lot of active perforance work which might want to use this, so it can be handy to have more specific options

packages/test/mocha-test-setup/mocharc-common.cjs
CraigMacomber: "false" would be true, even with truthyness, since enviroment variables are strings.

This style of check is consistant with FLUID_TEST_TIMEOUT and FLUID_TEST_FORBID_ONLY, and I think is the more clear pattern in this case.

I've added a comment about this.

packages/dds/shared-object-base/package.json
CraigMacomber: removed.

packages/dds/tree/src/test/memory/tableTree.spec.ts
CraigMacomber: Added.

packages/dds/matrix/package.json
CraigMacomber: removed from sequence and map.

packages/dds/sequence/src/test/memory/sharedString.spec.ts
CraigMacomber: SHould be even better now.

CraigMacomber: All fixed, I think.

packages/dds/merge-tree/package.json
CraigMacomber: The profile scripts only ever existed for direct developer use locally. It should still work, and profiling is mainly used for execution time so this still makes sense. I figure I migth as well leave it.

tools/pipelines/test-perf-benchmarks.yml
CraigMacomber: I have added path filteing: as a lot less files impact perf than the full client CI, it should now runn less often.

package.json
CraigMacomber: The issue was there was logic (isMemoryTest) which tried to detect which mode was being run, when we should just emit both test unconditionally. This has been fixed.

CraigMacomber: also note end to end tests do not use these scripts (they have no bench script) and instead have their own entries in the pipeline.

packages/test/test-version-utils/src/describeE2eDocs.ts
```diff
@@ -501,43 +490,4 @@ export const describeE2EDocsRuntime: DescribeE2EDocSuite =
...
- * @internal
- */
-export const getCurrentBenchmarkType = (currentType: DescribeE2EDocSuite): BenchmarkType => {
-	return currentType === describeE2EDocsMemory ? "MemoryUsage" : "ExecutionTime";
-};
-
-function createE2EDocsDescribeRun(): DescribeE2EDocSuite {
-	return isMemoryTest() === true ? describeE2EDocsMemory : describeE2EDocsRuntime;
-}
+export const describeE2EDocRun: DescribeE2EDocSuite = describeE2EDocsRuntime;
```
alexvy86: With `describeE2EDocsMemory` gone, seems like we want to update `describeE2EDocsRuntime` to not say "with type Runtime benchmarks"? Maybe even get rid of it too.

CraigMacomber:
New pipeline test run: https://dev.azure.com/fluidframework/internal/_build/results?buildId=392185&view=results

That failed with end to end test issues. This one should be fixed: https://dev.azure.com/fluidframework/internal/_build/results?buildId=392412&view=results

Fixed timeouts: https://dev.azure.com/fluidframework/internal/_build/results?buildId=392412&view=results

E2E fixed: https://dev.azure.com/fluidframework/internal/_build/results?buildId=392675&view=results

packages/dds/matrix/src/test/matrixBenchmarks.ts
```diff
@@ -207,47 +208,55 @@ function runMemoryBenchmark({
...
-				this.undoRedoStack = undoRedoStack;
-				this.cleanUp = cleanUp;
-
-				beforeOperation?.(this.matrix, this.undoRedoStack);
-			}
-
-			afterIteration(): void {
-				assert(this.matrix !== undefined, "matrix is not initialized");
-				assert(this.undoRedoStack !== undefined, "undoRedoStack is not initialized");
-				assert(this.cleanUp !== undefined, "cleanUp is not initialized");
```
alexvy86, line 239: @Josmithr do you recall if these caught something we were doing wrong with the test? I.e. measuring when nothing had been initialized, or similar? We're not keeping them in the PR and I don't recall if they were really valuable or not.

examples/benchmarks/tablebench/src/test/table.bench.spec.ts
Josmithr: Question (not to be addressed in this PR): do we still need these? Now that we have a user-facing table API with its own suite of performance tests, do these provide value?

packages/test/test-end-to-end-tests/src/test/benchmark/SimpleTest.all.spec.ts
```diff
@@ -1,75 +0,0 @@
...
-			iteration = 0;
-
-			async run(): Promise<void> {
-				this.container = undefined;
-				this.iteration++;
-				assert(this.iteration > 0, "testTitle needs to be defined");
-				console.log(`this will run for each iteration ${this.iteration}`);
-			}
-			async before(): Promise<void> {
-				console.log(`this will run before the measurements`);
```
CraigMacomber, line 41: this file, which adds tests that spam console output for every iteration (so hundreds of lines) is annoying, and seems like a poor way to document this. I replaced it with doc comments of the members of IBenchmarkParameters.

packages/test/test-end-to-end-tests/src/test/benchmark/summarizationDocument.all.spec.ts
```diff
@@ -102,10 +90,10 @@ describeE2EDocRun(scenarioTitle, (getTestObjectProvider, getDocumentInfo) => {
 				this.container = undefined;
 				this.summarizerClient = undefined;
 				if (isInPerformanceTestingMode) {
-					// TODO: this should be removed, or document why it exists (probably a workaround for memory measurement issues in current version of benchmark).
```
CraigMacomber, line 105: I think the memory tests didn't actually run this, just every batch of the duration tests, so this guess was wrong. Maybe its to avoid throttling on real service tests? Leaving this for now, but removed the probably wrong part of the comment (which I previously wrote when I had less information).

packages/dds/matrix/src/test/matrixBenchmarks.ts
CraigMacomber: these asserts were required for the code below to type check as they narrow the types to exclude undefined. That is no longer needed due to storing them in places with a better lifetime.

examples/benchmarks/tablebench/src/test/table.bench.spec.ts
CraigMacomber: if you don't use this package, then we can probably delete it.

packages/test/test-version-utils/src/describeE2eDocs.ts
CraigMacomber: deleted, along with a lot of related stuff.

alexvy86 (approved):
Looks good to me! Verified that the number of benchmarks run in the pipeline before/after this makes sense (some packages run the same number, others run less for known reasons like Diagnostic/Perspective benchmarks not running by default now).

Abe27342 (approved):

alexvy86:
Linking AB#6200 just for reference.

PR #26867, opened 2026-03-27
Deduplicate logic in charactersCopy and getString

CraigMacomber:
## Description

Deduplicate logic in charactersCopy and getString.

Add new forEachNodeSubsequence utility.

Make getString more strict about its inputs (and give nice errors).

Fix that there was a duplicated assert short code (blocking assert tagging for release).

## Reviewer Guidance

The review process is outlined on [this wiki page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

PR #26869, opened 2026-03-27
Validate asserts

CraigMacomber:
## Description

Add `flub generate assertTags --all --validate` to "checks" command to block PRs which duplicate assert short codes or  include non constant assert mmessages which can not be tagged.

This should prevent merging code which will block assert tagging for releases.

## Reviewer Guidance


CraigMacomber:
Until https://github.com/microsoft/FluidFramework/pull/26867 merges, this should fail on CI, detecting the duplicated assert.

CraigMacomber:
Validation confirmed to work: expected error in https://dev.azure.com/fluidframework/public/_build/results?buildId=387930&view=logs&j=eefa678c-f484-5837-78d3-627f1635913b&t=a53dad46-1a42-5abb-a3cd-1368a15b6673&l=259:

> ERROR: Completed with 1 errors.
> ERROR: Duplicate shortcode 0xcde detected
> 	/mnt/vss/_work/1/s/FluidFramework/packages/dds/tree/src/text/textDomainFormatted.ts:254
> 	/mnt/vss/_work/1/s/FluidFramework/packages/dds/tree/src/text/textDomainFormatted.ts:298

Looks like this works.

package.json
```diff
@@ -45,7 +45,7 @@
 		"check:format:repo": "biome check .",
 		"check:versions": "flub check buildVersion -g client --path .",
 		"check:versions:fix": "flub check buildVersion -g client --path . --fix",
-		"checks": "fluid-build --task checks",
+		"checks": "fluid-build --task checks && flub generate assertTags --all --validate",
```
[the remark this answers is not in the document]
CraigMacomber: Assert tagging is a bit slow, and not clearly more important than the other checks. Running it second was intended.

markfields: I interpret `fluid-build --task checks` to imply that there's some set of tasks defined in config or code elsewhere.  Can we add this assert task there, rather than having a special case here?

CraigMacomber: We could, but that would be part of build tools, need to go through a package publish, then be integrated into client. If someone bothers to do that, they could remove this. However, since assert tagging is real strange and has hard coded client specific things (like the assert mapping file path), having this be client specific (and thus included here) doesn't seem too bad.

CraigMacomber: Or I guess maybe there is some way to do that in the fluid-build config? I know what I did here works: there likely is a better way though.

CraigMacomber: I made this work like the other checks via the config below in this file.

markfields (approved):

PR #26883, opened 2026-03-30
chore(lint): move inline type imports to import type statements in drivers

ChumpChief:
## Description

In `driver-base/documentDeltaConnection.ts`, move `ISentSignalMessage` and `IFluidErrorBase` from inline `type` positions in value import statements to their corresponding `import type` statements from the same source. This keeps type-only imports cleanly separated from value imports.

## Reviewer Guidance

### DIFF
--- a/packages/drivers/driver-base/src/documentDeltaConnection.ts
+++ b/packages/drivers/driver-base/src/documentDeltaConnection.ts
@@ -15,21 +15,20 @@ import type {
 	IConnect,
 	IConnected,
 	IDocumentMessage,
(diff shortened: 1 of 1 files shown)


jason-ha (approved):
:)
It was just this file? And I happened to see per merge conflict. Easy then. =)

ChumpChief:
> :) It was just this file? And I happened to see per merge conflict. Easy then. =)

I made Claude triple-check and did some spot-checking of my own and didn't see anything else :-D

PR #26888, opened 2026-03-30
fix: pnpm overrides for batch CG CVE fixes (lodash, yaml, picomatch, brace-expansion, langsmith, path-to-regexp)

frankmueller-msft:
## Summary

Resolves multiple CG security alerts by updating the lockfile to pull in patched transitive dependency versions, plus one pnpm override for langsmith (the only case where the consumer's declared range doesn't reach the patched version).

| Package | Old Version(s) | New Version | Override needed? |
|---------|---------------|-------------|:---:|
| lodash | 4.17.21 | 4.17.23+ | No — lockfile only |
| lodash-es | 4.17.21/4.17.22 | 4.18.1 | No — lockfile only |
| yaml@1.x | 1.10.2 | 1.10.3 | No — lockfile only |
| yaml@2.x | 2.8.1, 2.8.2 | 2.8.3 | No — lockfile only |
| brace-expansion@1.x | 1.1.11, 1.1.12 | 1.1.13 | No — lockfile only |
| brace-expansion@2.x | 2.0.1, 2.0.2 | 2.0.3 | No — lockfile only |
| brace-expansion@5.x | 5.0.2 | 5.0.5 | No — lockfile only |
| langsmith | 0.3.73 | ^0.5.15 | **Yes** — consumer declares ^0.3.x |
| path-to-regexp@8.x | 8.2.0 | 8.4.2 | No — lockfile only |

## Notes

- Most packages are resolved to patched versions naturally by pnpm without overrides — the transitive dependency specs already allow the newer versions
- Only `langsmith` requires an override because the consumer declares `^0.3.x` and the patched version (0.5.15) crosses a minor version boundary in 0.x semver
- Picomatch and node-forge overrides were already merged separately and are now in main

## Test plan

- [x] `pnpm install` succeeds with no errors
- [x] `syncpack lint` passes
- [x] `biome check` passes
- [ ] CI passes (builds, policy checks, all tests)
- [ ] Verify flagged package versions no longer appear in lockfile after merge


package.json
```diff
@@ -401,7 +406,17 @@
 			"serialize-javascript@>=6 <7": "^7.0.4",
 			"express@>=4 <5": "^4.22.1",
 			"picomatch@>=2 <3": "^2.3.2",
-			"picomatch@>=4 <5": "^4.0.4"
+			"picomatch@>=4 <5": "^4.0.4",
```
Abe27342, line 405: how many of these actually force the use of overrides? if the transitive dep specification allows just resolving to the newer version, it would be better to update the lockfile without adding the overrides. That is usually accomplished by doing what you have already done (add the overrides first and `pnpm install --no-frozen-lockfile`), but then removing them before committing just the lockfile changes.

For ones where we actually do need them, these are fine. But I'd guess most of these aren't necessary.

frankmueller-msft: Good call. Tested by removing all overrides and running `pnpm install --no-frozen-lockfile` — pnpm naturally resolves to the patched versions for everything except **langsmith** (consumer declares `^0.3.x`, so pnpm stays at 0.3.87 and can't reach the patched 0.5.15).

Removed all unnecessary overrides (lodash, lodash-es, yaml, brace-expansion, path-to-regexp) and their syncpackrc entries. Kept only the langsmith override. The lockfile still resolves to the patched versions for the rest.

Abe27342 (approved):

PR #26892, opened 2026-03-30
Update test tenant acquisition for real service e2e test pipeline 

jzaffiro:
Use federated tokens as the credential method for authentication in the e2e and stress pipeline templates. 

[AB#52057](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/52057)


Abe27342 (approved):
approving so we meet the FIC deadline. Could you do a cleanup pass on env variables that aren't necessary anymore? e.g. I don't think login__odsp__test__tenants is.

PR #26779, opened 2026-03-19
Remove ID Allocation Batch, instead JIT generating ID Allocations right before submitting other batch types

markfields:
## _PROTOTYPE / PROOF OF CONCEPT_

This gets rid of ID Allocation Batches in the batch stream which makes detecting forked containers more robust and simple.

(Incidental rebirth of #24510 a year later)

PR #26898, opened 2026-03-31
Remove ID allocation batches and JIT generate Id alloc batches.

dannimad:
Using prototype https://github.com/microsoft/FluidFramework/pull/26779/changes/a244a1f0056b244abb8428ebc786d86587a771d4 as base. 

This branch removes the dedicated idAllocationBatch from the Outbox and replaces it with just-in-time (JIT) ID allocation op generation at flush time.

### Key Changes
**Core change — JIT ID allocation ([outbox.ts](vscode-webview://0setsnuqdgpe6inajv22i20l1mo21fcect4fpqbjjmsc3q1f9bv8/packages/runtime/container-runtime/src/opLifecycle/outbox.ts), [containerRuntime.ts](vscode-webview://0setsnuqdgpe6inajv22i20l1mo21fcect4fpqbjjmsc3q1f9bv8/packages/runtime/container-runtime/src/containerRuntime.ts)):**

- Previously: submitIdAllocationOpIfNeeded() was called eagerly before each non-staged op submission, pushing an ID alloc op into a separate idAllocationBatch in the Outbox.
- Now: A generateIdAllocationOp callback is passed to the Outbox via IOutboxParameters. At flush time (after any rebase), the outbox calls this JIT to generate the ID alloc op and prepends it to the batch being sent.

The ID alloc op is generated after rebase, so its refSeq always matches the rest of the batch.
ID ranges are no longer lost during rebase (previously reSubmit would drop IdAllocation ops and needed special handling with takeUnfinalizedCreationRange).
Simplifies the Outbox from 3 batch managers down to 2 (mainBatch + blobAttachBatch).

By removing ID allocation batches, we are effectively solving the inherent problem of not being able to track forking containers due to these batches being untraceable via batchId.

### DIFF
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -2064,6 +2064,28 @@ export class ContainerRuntime
 			}),
 			reSubmit: this.reSubmit.bind(this),
 			opReentrancy: () => this.dataModelChangeRunner.running,
+			generateIdAllocationOp: (
+				useUnfinalizedRange: boolean,
+			): LocalBatchMessage | undefined => {
+				if (this._idCompressor === undefined) {
+					return undefined;
+				}
+				const idRange = useUnfinalizedRange
+					? this._idCompressor.takeUnfinalizedCreationRange()
+					: this._idCompressor.takeNextCreationRange();
+				if (idRange.ids === undefined) {
+					return undefined;
+				}
+				const idAllocationMessage: ContainerRuntimeIdAllocationMessage = {
+					type: ContainerMessageType.IdAllocation,
+					contents: idRange,
+				};
+				return {
+					runtimeOp: idAllocationMessage,
+					referenceSequenceNumber: this.deltaManager.lastSequenceNumber,
+					staged: false,
+				};
+			},
 		});
 
 		this._quorum = quorum;
@@ -3682,9 +3704,6 @@ export class ContainerRuntime
 
 						this.stageControls = undefined;
 
-						// During Staging Mode, we avoid submitting any ID Allocation ops (apart from resubmitting pre-staging ops).
-						// Now that we've exited, we need to submit an ID Allocation op for any IDs that were generated while in Staging Mode.
-						this.submitIdAllocationOpIfNeeded({ staged: false });
 						const batchInfos = discardOrCommit();
 						event.reportProgress({
 							details: {
@@ -4728,33 +4747,6 @@ export class ContainerRuntime
 		return this.blobManager.lookupTemporaryBlobStorageId(localId);
 	}
 
-	private submitIdAllocationOpIfNeeded({
-		resubmitOutstandingRanges = false,
-		staged,
-	}: {
-		resubmitOutstandingRanges?: boolean;
-		staged: boolean;
-	}): void {
-		if (this._idCompressor) {
-			const idRange = resubmitOutstandingRanges
-				? this._idCompressor.takeUnfinalizedCreationRange()
-				: this._idCompressor.takeNextCreationRange();
-			// Don't include the idRange if there weren't any Ids allocated
-			if (idRange.ids !== undefined) {
-				const idAllocationMessage: ContainerRuntimeIdAllocationMessage = {
-					type: ContainerMessageType.IdAllocation,
-					contents: idRange,
-				};
-				const idAllocationBatchMessage: LocalBatchMessage = {
-					runtimeOp: idAllocationMessage,
-					referenceSequenceNumber: this.deltaManager.lastSequenceNumber,
-					staged,
-				};
-				this.outbox.submitIdAllocation(idAllocationBatchMessage);
-			}
-		}
-	}
-
 	private submit(
 		containerRuntimeMessage: LocalContainerRuntimeMessage,
 		localOpMetadata: unknown = undefined,
@@ -4797,11 +4789,6 @@ export class ContainerRuntime
 				!staged || canStageMessageOfType(type),
 				0xbba /* Unexpected message type submitted in Staging Mode */,
 			);
-
-			// Before submitting any non-staged change, submit the ID Allocation op to cover any compressed IDs included in the op.
-			if (!staged) {
-				this.submitIdAllocationOpIfNeeded({ staged: false });
-			}
 
 			// Allow document schema controller to send a message if it needs to propose change in document schema.
 			// If it needs to send a message, it will call provided callback with payload of such message and rely
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -2064,15 +2064,11 @@ export class ContainerRuntime
 			}),
 			reSubmit: this.reSubmit.bind(this),
 			opReentrancy: () => this.dataModelChangeRunner.running,
-			generateIdAllocationOp: (
-				useUnfinalizedRange: boolean,
-			): LocalBatchMessage | undefined => {
+			generateIdAllocationOp: (): LocalBatchMessage | undefined => {
 				if (this._idCompressor === undefined) {
 					return undefined;
 				}
-				const idRange = useUnfinalizedRange
-					? this._idCompressor.takeUnfinalizedCreationRange()
-					: this._idCompressor.takeNextCreationRange();
+				const idRange = this._idCompressor.takeNextCreationRange();
 				if (idRange.ids === undefined) {
 					return undefined;
 				}
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -4282,7 +4282,6 @@ export class ContainerRuntime
 					outboxLength: this.outbox.messageCount,
 					mainBatchLength: this.outbox.mainBatchMessageCount,
 					blobAttachBatchLength: this.outbox.blobAttachBatchMessageCount,
-					idAllocationBatchLength: this.outbox.idAllocationBatchMessageCount,
 				},
 			);
 		}
@@ -4845,7 +4844,7 @@ export class ContainerRuntime
 		// Incoming ops still break the batch via direct this.flush() calls elsewhere
 		// (deltaManager "op" handler, process(), connection changes, getPendingLocalState,
 		// exitStagingMode). Those all bypass scheduleFlush(), so they're unaffected by this check.
-		// Additionally, outbox.maybeFlushPartialBatch() (called on every submit) detects
+		// Additionally, outbox.outboxSequenceNumberCoherencyCheck() (called on every submit) detects
 		// sequence number changes and throws if unexpected changes are detected.
 		if (
 			this.inStagingMode &&
--- a/packages/runtime/container-runtime/src/opLifecycle/batchManager.ts
+++ b/packages/runtime/container-runtime/src/opLifecycle/batchManager.ts
@@ -20,6 +20,7 @@ import { serializeOp } from "./opSerialization.js";
 import type { BatchStartInfo } from "./remoteMessageProcessor.js";
 
 export interface IBatchManagerOptions {
+	readonly disableGroupedBatching: boolean;
 	readonly compressionOptions?: ICompressionRuntimeOptions;
 }
 
--- a/packages/runtime/container-runtime/src/opLifecycle/batchManager.ts
+++ b/packages/runtime/container-runtime/src/opLifecycle/batchManager.ts
@@ -26,11 +26,6 @@ export interface IBatchManagerOptions {
 	 * If true, the outbox is allowed to rebase the batch during flushing.
 	 */
 	readonly canRebase: boolean;
-
-	/**
-	 * If true, don't compare batchID of incoming batches to this. e.g. ID Allocation Batch IDs should be ignored
-	 */
-	readonly ignoreBatchId?: boolean;
 }
 
 export interface BatchSequenceNumbers {
@@ -127,8 +122,9 @@ export class BatchManager {
 
 	/**
 	 * Gets the pending batch and clears state for the next batch.
+	 * The caller is responsible for calling {@link addBatchMetadata} after any modifications (e.g. prepending messages).
 	 */
-	public popBatch(batchId?: BatchId): LocalBatch {
+	public popBatch(): LocalBatch {
 		assert(this.pendingBatch[0] !== undefined, 0xb8a /* expected non-empty batch */);
 		const batch: LocalBatch = {
 			messages: this.pendingBatch,
@@ -141,7 +137,7 @@ export class BatchManager {
 		this.clientSequenceNumber = undefined;
 		this.hasReentrantOps = false;
 
-		return addBatchMetadata(batch, batchId);
+		return batch;
 	}
 
 	/**
--- a/packages/runtime/container-runtime/src/opLifecycle/batchManager.ts
+++ b/packages/runtime/container-runtime/src/opLifecycle/batchManager.ts
@@ -21,11 +21,6 @@ import type { BatchStartInfo } from "./remoteMessageProcessor.js";
 
 export interface IBatchManagerOptions {
 	readonly compressionOptions?: ICompressionRuntimeOptions;
-
-	/**
-	 * If true, the outbox is allowed to rebase the batch during flushing.
-	 */
-	readonly canRebase: boolean;
 }
 
 export interface BatchSequenceNumbers {
--- a/packages/runtime/container-runtime/src/opLifecycle/outbox.ts
+++ b/packages/runtime/container-runtime/src/opLifecycle/outbox.ts
@@ -212,8 +212,8 @@ export class Outbox {
 	constructor(private readonly params: IOutboxParameters) {
 		this.logger = createChildLogger({ logger: params.logger, namespace: "Outbox" });
 
-		this.mainBatch = new BatchManager({});
-		this.blobAttachBatch = new BatchManager({});
+		this.mainBatch = new BatchManager({ disableGroupedBatching: false });
+		this.blobAttachBatch = new BatchManager({ disableGroupedBatching: true });
 	}
 
 	public get messageCount(): number {
@@ -233,10 +233,7 @@ export class Outbox {
 	}
 
 	public containsUserChanges(): boolean {
-		return (
-			this.mainBatch.containsUserChanges() || this.blobAttachBatch.containsUserChanges()
-			// ID Allocation ops are not user changes
-		);
+		return this.mainBatch.containsUserChanges() || this.blobAttachBatch.containsUserChanges();
 	}
 
 	/**
@@ -374,7 +371,6 @@ export class Outbox {
 
 		this.flushInternal({
 			batchManager: this.blobAttachBatch,
-			disableGroupedBatching: true,
 			resubmitInfo,
 		});
 		this.flushInternal({
@@ -414,10 +410,9 @@ export class Outbox {
 
 	private flushInternal(params: {
 		batchManager: BatchManager;
-		disableGroupedBatching?: boolean;
 		resubmitInfo?: BatchResubmitInfo; // undefined if not resubmitting
 	}): void {
-		const { batchManager, disableGroupedBatching = false, resubmitInfo } = params;
+		const { batchManager, resubmitInfo } = params;
 		if (batchManager.empty) {
 			return;
 		}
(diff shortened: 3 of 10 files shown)


packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -65,6 +66,17 @@ export interface IOutboxParameters {
...
+	/**
+	 * JIT callback to generate an ID allocation op at flush time.
+	 * Called after rebase (if any), so the returned message has the correct refSeq.
+	 *
+	 * @param useUnfinalizedRange - If true, use `takeUnfinalizedCreationRange()` (for reconnect).
+	 * Otherwise use `takeNextCreationRange()` (normal flush).
+	 * @returns A LocalBatchMessage for the ID allocation op, or undefined if no IDs need allocating.
+	 */
+	readonly generateIdAllocationOp: (
+		useUnfinalizedRange: boolean,
```
markfields: Shouldn't need this param anymore

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -189,7 +201,7 @@ export class Outbox {
 	private readonly logger: ITelemetryLoggerExt;
 	private readonly mainBatch: BatchManager;
 	private readonly blobAttachBatch: BatchManager;
-	private readonly idAllocationBatch: BatchManager;
+	private useUnfinalizedRangeOnNextFlush = false;
```
markfields: This should not be needed either

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -456,16 +465,32 @@ export class Outbox {
...
-			this.rebase(rawBatch, batchManager);
+			this.rebase(rawBatch, batchManager, generateJIT);
 			return;
 		}
 
+		// Generate ID Allocation ops just-in-time, after rebase (if any).
+		// This ensures the refSeq is correct (matching the rest of the batch) and that
+		// ID ranges aren't lost during rebase (since reSubmit drops IdAllocation ops).
+		// Only generate for non-staged batches — ID alloc ops are always non-staged.
+		if (generateJIT === true && !staged) {
```
markfields: I think this block should move down into the "shouldSend and not staged" block

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -225,9 +228,11 @@ export class Outbox {
...
 	}
 
-	public get idAllocationBatchMessageCount(): number {
-		return this.idAllocationBatch.length;
-	}
+	/**
+	 * @remarks With JIT ID allocation, pending IDs are tracked by the compressor, not the outbox.
+	 * This always returns 0. ID allocation ops are generated at flush time.
+	 */
+	public readonly idAllocationBatchMessageCount: number = 0;
```
markfields: Can we just get rid of usages of this and delete it?

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -254,19 +259,16 @@ export class Outbox {
 	private maybeFlushPartialBatch(): void {
```
markfields: Unrelated: We should rename this, want to do that in this PR?  No pressure.

Name idea if so:  ` outboxSequenceNumberCoherencyCheck`

markfields: (this was a miss from a recent PR that removed the codepath that would actually flush, it had been disabled via feature gate for months)

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -351,8 +347,9 @@ export class Outbox {
 	 */
 	public flush(resubmitInfo?: BatchResubmitInfo): void {
 		// We have nothing to flush if all batchManagers are empty, and we we're not needing to resubmit an empty batch placeholder
+		// Note that it's possible that there are unfinalized ranges in the ID Compressor,
```
markfields: nit: I'd move this down a few lines right above the return statement

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -207,14 +214,10 @@ export class Outbox {
 
 		this.mainBatch = new BatchManager({ canRebase: true });
 		this.blobAttachBatch = new BatchManager({ canRebase: true });
-		this.idAllocationBatch = new BatchManager({
-			canRebase: false,
```
markfields, line 211: We could probably get rid of `canRebase` in BatchManager as well

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -456,6 +445,7 @@ export class Outbox {
 			groupingEnabled
 		) {
 			assert(!this.rebasing, 0x6fa /* A rebased batch should never have reentrant ops */);
+			// Rebase the current batch (resubmit the ops one-by-one) and then reinvoke flushInternal.
```
markfields, line 431: Side note - thanks for adding this. I remember being confused and surprised that rebase calls back into flushInternal.  If we ever touch this code again maybe it's worth updating rebase to return a bool or something and have the reentrant call from here directly so the control flow is more obvious.

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -466,11 +456,22 @@ export class Outbox {
...
-		if (this.params.shouldSend() && !staged) {
+		const shouldSendNow = this.params.shouldSend() && !staged;
+		let clientSequenceNumber: number | undefined;
+		if (shouldSendNow) {
+			// Generate ID Allocation op just-in-time, after rebase (if any), and before addBatchMetadata,
+			// so that the prepended idAllocMsg is correctly marked as the first op in the batch.
+			// This ensures the refSeq is correct (matching the rest of the batch) and that
+			// ID ranges aren't lost during rebase (since reSubmit drops IdAllocation ops).
+			// Only generate for non-staged batches — ID alloc ops are always non-staged.
+			const idAllocMsg = this.params.generateIdAllocationOp();
```
markfields, line 453: I wonder if we should call this `generateUnstagedIdAllocationOp`? Could be too much (kind of distracting when not thinking about staging mode) but the benefit is it makes it easier to reason over the fact that ID Allocation ops are never submitted as "staged".

Just thinking out loud, totally up to you what you choose.

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -466,11 +456,22 @@ export class Outbox {
...
+			if (idAllocMsg !== undefined) {
+				rawBatch = { ...rawBatch, messages: [idAllocMsg, ...rawBatch.messages] };
+			}
+			addBatchMetadata(rawBatch, resubmitInfo?.batchId);
 			const virtualizedBatch = this.virtualizeBatch(rawBatch, groupingEnabled);
```
markfields, line 458: Optional: What do you think about moving the previous couple lines into `virtualizeBatch`?  Could rename to `finalize` or `prepare`.

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -480,11 +481,12 @@ export class Outbox {
 			);
 		}
 
+		addBatchMetadata(rawBatch, resubmitInfo?.batchId);
```
markfields: May not be practical, but it might be nice if `addBatchMetadata` changed the type of the input (could give it a return value) so it's impossible to accidentally flush a batch to PSM without calling this first.

packages/runtime/container-runtime/src/test/opLifecycle/batchManager.spec.ts
markfields: Seems like these tests could be moved/simplified to just be testing `addBatchMetadata` and not include anything with BatchManager.  What do you think?

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -466,11 +456,22 @@ export class Outbox {
...
+		let clientSequenceNumber: number | undefined;
+		if (shouldSendNow) {
+			// Generate ID Allocation op just-in-time, after rebase (if any), and before addBatchMetadata,
+			// so that the prepended idAllocMsg is correctly marked as the first op in the batch.
+			// This ensures the refSeq is correct (matching the rest of the batch) and that
+			// ID ranges aren't lost during rebase (since reSubmit drops IdAllocation ops).
+			// Only generate for non-staged batches — ID alloc ops are always non-staged.
+			const idAllocMsg = this.params.generateIdAllocationOp();
+			if (idAllocMsg !== undefined) {
+				rawBatch = { ...rawBatch, messages: [idAllocMsg, ...rawBatch.messages] };
```
markfields, line 455: nit: I think we could do `const rawBatch` above and here just do `rawBatch.messages = [idAllocMsg, ...rawBatch.messages]` or even `rawBatch.messages.unshift(idAllocMsg)`.

The spreading protects against modifying an input in case the caller doesn't want it modified, but we own this variable's lifetime so it feels fine to just modify it in place.

packages/runtime/container-runtime/src/test/opLifecycle/remoteMessageProcessor.spec.ts
```diff
@@ -372,6 +373,7 @@ describe("RemoteMessageProcessor", () => {
 				false /* reentrant */,
 			);
 			const batchA = batchManager.popBatch();
+			addBatchMetadata(batchA);
```
markfields, line 368: Check if we need BatchManager for these tests anymore or if we can just call addBatchMetadata on a const array of ops?

packages/runtime/container-runtime/src/containerRuntime.ts
markfields, line 3685: Think about if we should have a comment _somewhere_ about how we didn't generate ID Allocations during Staging Mode but will include one if needed when submitting the staged changes?

markfields (commented):
Looking good!  Left some comments to consider, I still need to review all the tests.

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
dannimad: There are some nuances there since addBatchMetadata changes rawBatch in place. I agree there are some improvements here but I'd prefer to do those ones in a follow up since it's a bit more of a targeted change.

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
dannimad: This can be discussed and better tackled in a follow up.

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
dannimad: Follow up change

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
dannimad: Maybe we could find a way to assert the staged state, I'll leave this as it is right now.

packages/runtime/container-runtime/src/opLifecycle/batchManager.ts
```diff
@@ -21,16 +21,6 @@ import type { BatchStartInfo } from "./remoteMessageProcessor.js";
 
 export interface IBatchManagerOptions {
 	readonly compressionOptions?: ICompressionRuntimeOptions;
```
markfields, line 24: For future reference - Apparently this is unused too.  We can/should remove IBatchManagerOptions IMO

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -450,12 +433,9 @@ export class Outbox {
 
 		const groupingEnabled =
 			!disableGroupedBatching && this.params.groupingManager.groupedBatchingEnabled();
-		if (
-			batchManager.options.canRebase &&
-			rawBatch.hasReentrantOps === true &&
-			groupingEnabled
-		) {
+		if (rawBatch.hasReentrantOps === true && groupingEnabled) {
```
markfields: We could have a problem here, I'm not sure.  If grouping is not enabled, then we will not rebase, so the batch could start with ops with a lower refSeq.  Then we pre-pend an ID Allocation op with a higher refSeq, which means refSeq is not monotonically increasing. Is that violating an invariant of the op stream?

We can ask around, but also should write a test highlighting this case to just find out for ourselves.  Even if the test passes we may still want to do something different based on input from others.

markfields: Ok as discussed, should be safe to just get rid of the groupingEnabled check.  HOWEVER we'll also need to ensure `disableGroupedBatching` gets plumbed through properly. Since that's only used by Blob Attach, we can just add that as a property on the BatchManager which is plumbed through.

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -450,12 +433,9 @@ export class Outbox {
...
 			assert(!this.rebasing, 0x6fa /* A rebased batch should never have reentrant ops */);
```
markfields, line 430: Let's also assert that resumbitInfo is undefined. This is important since it's not plumbed through rebase.

(btw, this subtlety is another reason why having rebase return info and have the recursive call to flushInternal here in flushInternal would be better - more obvious that you need to pass through all params.)

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -225,10 +228,6 @@ export class Outbox {
 		return this.blobAttachBatch.length;
 	}
 
-	public get idAllocationBatchMessageCount(): number {
-		return this.idAllocationBatch.length;
-	}
-
 	public get isEmpty(): boolean {
 		return this.messageCount === 0;
 	}
```
markfields, line 233: nit: containsUserChanges (next fn) has an obsolete commment now

packages/runtime/container-runtime/src/test/opLifecycle/batchManager.spec.ts
```diff
@@ -47,50 +52,40 @@ const smallMessage = (size: number = 100): LocalBatchMessage => {
...
-			assert.deepEqual(
-				singleOpBatch.messages.map((m) => m.metadata as IBatchMetadata),
-				[
-					includeBatchId ? { batchId } : undefined, // batchId propertly should be omitted (v. set to undefined) if not provided
-				],
-			);
+	describe("addBatchMetadata", () => {
+		const makeLocalBatch = (...ops: string[]): LocalBatch => ({
+			messages: ops.map((data, i) => ({ runtimeOp: op(data), referenceSequenceNumber: i })),
+			referenceSequenceNumber: ops.length - 1,
```
markfields, line 60: nit: I'd do undefined here, since it's irrelevant for the logic.  (Looking at the BatchManager anbd Outbox code, I don't feel like this property should exist on a LocalBatch at all, it should be computed when virtualizing the batch to an Outbound batch, at which point we could assert all ops have the same refseq)

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -480,11 +471,12 @@ export class Outbox {
 			);
 		}
```
markfields, line 468: Missed this the first time through - let's put this `addBatchMetadata` in an else block!  It's idempotent but still we shouldn't be calling it twice.

packages/runtime/container-runtime/src/test/opLifecycle/outbox.spec.ts
```diff
@@ -325,53 +327,64 @@ describe("Outbox", () => {
 	});
 
 	it("Sending batches", () => {
-		const outbox = getOutbox({ context: getMockContext() });
+		const idAllocMessage = createMessage(ContainerMessageType.IdAllocation, "idAlloc");
+		let idAllocReturned = false;
```
markfields, line 331: Nit: Tests would be more readable IMO if this bit were flipped alongside the submit calls, so you can easily see its current state at any point and reason over whether an ID Allocation will be included in the next batch

packages/runtime/container-runtime/src/test/opLifecycle/outbox.spec.ts
```diff
@@ -325,53 +327,64 @@ describe("Outbox", () => {
...
+				}
+				return undefined;
+			},
+		});
 		const messages = [
 			createMessage(ContainerMessageType.FluidDataStoreOp, "0"),
 			createMessage(ContainerMessageType.FluidDataStoreOp, "1"),
-			createMessage(ContainerMessageType.IdAllocation, "2"),
-			createMessage(ContainerMessageType.IdAllocation, "3"),
 			createMessage(ContainerMessageType.FluidDataStoreOp, "4"),
```
markfields, line 346: nit: 2, 3 NOT 4, 5

packages/runtime/container-runtime/src/test/opLifecycle/outbox.spec.ts
```diff
@@ -573,45 +579,39 @@ describe("Outbox", () => {
 		const messages = [
 			createMessage(ContainerMessageType.FluidDataStoreOp, "0"),
 			createMessage(ContainerMessageType.FluidDataStoreOp, "1"),
-			createMessage(ContainerMessageType.IdAllocation, "2"),
-			createMessage(ContainerMessageType.FluidDataStoreOp, "3"),
+			createMessage(ContainerMessageType.FluidDataStoreOp, "2"),
```
markfields, line 582: Can be follow-up:  I think we'll have more robust test coverage if by default at least one (or for simplicity could probaly do all) batch gets ID Allocation prepended.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4758,11 +4778,10 @@ describe("Runtime", () => {
 
 				it("IdAllocation + reconnect while in staging mode does not hit coherency check", async () => {
 					// This is the highest-risk scenario from Mark's test plan.
```
markfields: nit: Remove :)

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4758,11 +4778,10 @@ describe("Runtime", () => {
...
-					// The fix in b4e1fd1dd25 added scheduleFlush() after submitIdAllocationOpIfNeeded
-					// during replayPendingStates. With threshold suppression, that scheduleFlush()
-					// returns early if in staging mode and under threshold. But the "op" handler
-					// calls flush() directly, so the IdAllocation op still gets flushed before
-					// new ops with different refSeqs are submitted.
+					// With JIT ID allocation, the ID alloc op is generated at flush time
+					// (via generateIdAllocationOp callback), so it naturally gets the correct refSeq.
+					// The "op" handler calls flush() directly, so the IdAllocation op still gets
+					// flushed before new ops with different refSeqs are submitted.
```
markfields, line 4783: Check other comments in this test, some are wrong (e.g. L4809).

Actually this may be another test I should I looked at with my PR.  We can follow-up after

markfields (approved):
LGTM! We can take my remaining comments in follow-ups

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
markfields: @dannimad - This is the only other comment that you might want to take in this PR as opposed to a follow-up.

markfields (changes requested):
Oops I forgot about the rebase thing we talked about this morning. Will approve after that's fixed

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -449,13 +420,15 @@ export class Outbox {
...
 		const groupingEnabled =
-			!disableGroupedBatching && this.params.groupingManager.groupedBatchingEnabled();
-		if (
-			batchManager.options.canRebase &&
-			rawBatch.hasReentrantOps === true &&
-			groupingEnabled
-		) {
+			!batchManager.options.disableGroupedBatching &&
+			this.params.groupingManager.groupedBatchingEnabled();
+		if (rawBatch.hasReentrantOps === true) {
```
markfields, line 425: @ChumpChief -- FYI that "reentrant blob attach batches" will now be rebased. i.e. if processing an incoming op triggers the submission of a blob attach op, then that op will go through the resubmit fn here during flush.

I'm not really sure how common (possible?) this would be.  I looked at the resubmit code and it seems fine to do, it'll always be within the same JS turn as the original submit.

markfields (approved):
LGTM! 🚢

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
ChumpChief: In general I don't expect blob attach ops to be particularly sensitive to timing/reentrancy/rebase - whenever the blob manager is asked to resubmit an attach op it will reevaluate whether the attach op is still valid and either repeat it or go back through the upload flow.  Shouldn't matter what anything else is doing at the time afaik.

You probably can't ever get a blob attach op synchronously triggered from the processing of an incoming op, since we don't give direct control over submitting blob attach ops (at least with current functionality).  Rather the customer might start an upload in response to op processing and later we'll send a corresponding attach op.

PR #26899, opened 2026-03-31
Revert "Remove isomorphic-fetch from odsp-doclib-utils"

alexvy86:
Reverts microsoft/FluidFramework#26842.

Seems like the removal of isomorphic-fetch broke the integration pipeline with Loop, where we see complaints like `fetch is not defined` during e2e tests.


ChumpChief (approved):
Approving, will get the downstream fix and then revert-revert

PR #26900, opened 2026-03-31
Update guidance for changeset issue linting

CraigMacomber:
## Description

Update guidence for changeset authoring regarding linking to clarify what is done automatically and that links to public issues can be included (as required by our deprecation process)

## Reviewer Guidance

### DIFF
--- a/.changeset/README.md
+++ b/.changeset/README.md
@@ -73,7 +73,7 @@ The sections are configured in [fluidBuild.config.cjs](https://github.com/micros
 
 > [!NOTE]
 > For deprecations, follow the [deprecation process](https://github.com/microsoft/FluidFramework/wiki/API-Deprecation).
-> Ensure both the changeset when doing the deprecation and the changeset when removing the deprecated API link the tracking issues on GitHub if one was created.
+> Ensure that both the deprecation changeset and the removal changeset include a link to the GitHub tracking issue, when one exists.
 
 
 ## Changesets that apply to no packages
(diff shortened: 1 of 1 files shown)


.changeset/README.md
```diff
@@ -71,6 +71,11 @@ The sections are configured in [fluidBuild.config.cjs](https://github.com/micros
 > [!NOTE]
 > Client releases with breaking _legacy_ changes should use the `legacy` section, not `breaking`. The `breaking` section is reserved for major releases, which practically means server.
 
+> [!NOTE]
+> For deprecations, follow the [deprecation process](https://github.com/microsoft/FluidFramework/wiki/API-Deprecation).
+> Ensure both the changeset when doing the deprecation and the changeset when removing the deprecated API link the tracking issues on GitHub if one was created.
```
jason-ha, line 76: There really should be just one issue, not multiple. The API Deprecation issue is the one that should exist. A Breaking Change issue might happen when things break down and there wasn't an API Deprecation issue.

I would use slightly stronger language and assume issue exists:
```suggestion
> Ensure both the changeset when doing the deprecation and the changeset when removing the deprecated API link the tracking issue on GitHub.
```

jason-ha (approved):
Heh. You updated before I could submit.
I was going to suggest dropping the conditional and just assume that one exists, because one should exist.
Can ignore the outdated item below.

Josmithr (approved):

PR #26901, opened 2026-03-31
arraynode event integration in text editor

daesunp:
## Description

Integrates the array node delta event payload into the text-editor app


examples/data-objects/text-editor/src/app.tsx
```diff
@@ -38,23 +38,12 @@ import { type FC, useEffect, useMemo, useState } from "react";
...
-	if (hostname.endsWith(".app.github.dev")) {
-		const match = /^(.+)-\d+\.app\.github\.dev$/.exec(hostname);
-		if (match) {
-			const codespaceName = match[1];
-			return `https://${codespaceName}-${tinyliciousPort}.app.github.dev`;
-		}
-	}
-
-	return `http://localhost:${tinyliciousPort}`;
+	return window.location.origin;
```
Josmithr: Why are we changing this?

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1582,17 +1582,36 @@ export namespace TableSchema {
...
 // @alpha
 export namespace TextAsTree {
+    // @sealed
     export interface Members {
         characterCount(): number;
         characters(): Iterable<string>;
         charactersCopy(): string[];
         fullString(): string;
         insertAt(index: number, additionalCharacters: string): void;
+        onCharactersChanged(callback: (ops: readonly TextOp[] | undefined) => void): () => void;
```
Josmithr, line 1602: Nit: I think it would probably be fine to just call this `onChange`. It isn't (currently) possible to put anything except string leaf nodes in a text tree, so we don't need to worry about differentiating between deep/shallow changes here. But this is just an opinion. @CraigMacomber curious what you think.

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -6,7 +6,11 @@
 import type { TextAsTree } from "@fluidframework/tree/internal";
 
 /**
- * Sync `newText` into the provided `root` tree.
+ * Sync `newText` into the provided `root` tree by applying the minimal remove + insert pair
```
Josmithr, line 9: Thanks for the docs updates! These are very useful!

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -6,7 +6,11 @@
...
+ * needed to transform the tree's current content into `newText`.
+ * @remarks
+ * The diff is computed by finding the longest shared prefix/suffix between current and new content
```
Josmithr, line 12: So this doesn't attempt to break up isolated ranges of changes? It just replaces everything except leading/trailing unchanged ranges?

packages/framework/quill-react/src/plain/quillView.tsx
```diff
@@ -27,90 +28,113 @@ export interface MainViewProps {
...
-		if (quillText !== treeTextWithNewline) {
 			isUpdatingRef.current = true;
-
-			const selection = quillRef.current.getSelection();
-			quillRef.current.setText(treeTextWithNewline);
-			if (selection) {
-				const length = quillRef.current.getLength();
-				const newPosition = Math.min(selection.index, length - 1);
-				quillRef.current.setSelection(newPosition, 0);
+			try {
```
Josmithr: I'm guessing this logic is basically the same as in `quillFormattedView` and `plainTextView`? Can we share any of this?

packages/dds/tree/api-report/tree.alpha.api.md
daesunp: Makes sense! Would we want to have a separate `onCharactersChanged` for rich text though? Since in those cases we would have deep changes (bold, italics, etc.)

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: No problem :)

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: Yes, I've updated the docs to be more clear!

packages/framework/quill-react/src/plain/quillView.tsx
daesunp: Unfortunately the smallest shared piece I found that doesn't add useCallback overhead is the isUpdating + try/finally + early-return boilerplate (which is now updated)

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -38,23 +38,12 @@ import { type FC, useEffect, useMemo, useState } from "react";
...
 }
```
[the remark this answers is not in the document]
daesunp: reverted

examples/data-objects/text-editor/src/app.tsx
daesunp: reverted

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -177,6 +186,22 @@ class TextNode
...
+		);
+	}
+
+	public onContentChanged(
+		callback: (ops: readonly TextAsTree.TextOp[] | undefined) => void,
+	): () => void {
+		return TreeAlpha.on(this.content, "treeChanged", ({ delta }) =>
+			processCharactersChangedDelta(delta, (i) => this.content[i]?.content.content, callback),
+		);
+	}
```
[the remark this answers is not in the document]
daesunp: Extracted an atomChar() helper

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -186,6 +200,60 @@ class StringArray extends sf.array("StringArray", SchemaFactory.string) {
...
+				const ch = getChar(readPos);
+				if (ch === undefined) {
+					// Tree is out of sync with the delta — fall back to full re-read.
+					callback(undefined);
+					return;
+				}
+				text += ch;
+				readPos++;
+			}
+			ops.push({ type: "insert", text });
```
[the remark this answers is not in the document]
daesunp: insert-loop now accumulates characters into an array and joins at the end

packages/dds/tree/src/text/codePointUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Returns the number of Unicode code points in `str`.
+ * @remarks
+ * Use this to translate a JavaScript string length (which is in UTF-16 code units) into
+ * the atom/code-point space used by {@link TextAsTree}.
+ * @internal
+ */
+export function codePointCount(str: string): number {
+	return [...str].length;
```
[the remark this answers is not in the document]
daesunp: updated with suggestion

packages/dds/tree/src/text/codePointUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function utf16LengthForCodePoints(str: string, start: number, count: number): number {
+	assert(start >= 0 && start <= str.length, "start must be within str bounds");
+	assert(count >= 0, "count must be non-negative");
+	let utf16 = 0;
+	let counted = 0;
+	while (counted < count && start + utf16 < str.length) {
+		utf16 += (str.codePointAt(start + utf16) ?? 0) > 0xffff ? 2 : 1;
+		counted++;
+	}
```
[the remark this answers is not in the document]
daesunp: updated with assert

packages/dds/tree/api-report/tree.alpha.api.md
Josmithr: Yeah, I think separate / more explicit methods names makes sense in the formatted case.

packages/framework/react/src/text/plain/plainUtils.ts
Josmithr: Thanks!

packages/dds/tree/src/text/codePointUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+import { assert } from "@fluidframework/core-utils/internal";
+
+/**
+ * Returns the number of Unicode code points in `str`.
+ * @remarks
+ * Use this to translate a JavaScript string length (which is in UTF-16 code units) into
+ * the atom/code-point space used by {@link TextAsTree}.
+ * @internal
+ */
+export function codePointCount(str: string): number {
```
Josmithr: Nit: our guidelines call to avoid abbreviations when reasonable. What do you think about `value` instead of `str` here?

Josmithr: (same for the function below)

packages/dds/tree/src/text/codePointUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Use this to translate {@link TextAsTree}-space counts (code points) into JavaScript string indices (UTF-16).
+ * One code point outside the Basic Multilingual Plane (e.g. most emoji) occupies two UTF-16 code units.
+ *
+ * Asserts that the requested `count` code points are fully consumable from `start`; silent truncation
+ * would misalign delta offsets applied to strings rather than surface the drift to the caller.
+ * @param str - The string to measure.
+ * @param start - The UTF-16 index in `str` to start measuring from. Must be in `[0, str.length]`.
+ * @param count - The number of Unicode code points to measure. Must be non-negative, and there must
+ * be at least `count` code points available in `str` starting at `start`.
+ * @internal
```
Josmithr, line 56: Nit: I think some examples (using `@example` comment blocks) would likely be useful here, especially if we're planning to eventually export this as a part of the user-facing API.

packages/dds/tree/src/text/codePointUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export function utf16LengthForCodePoints(str: string, start: number, count: number): number {
+	assert(start >= 0 && start <= str.length, "start must be within str bounds");
+	assert(count >= 0, "count must be non-negative");
+	let utf16 = 0;
+	let counted = 0;
+	while (counted < count && start + utf16 < str.length) {
+		utf16 += (str.codePointAt(start + utf16) ?? 0) > 0xffff ? 2 : 1;
```
Josmithr: Nit: I think a quick comment explaining this line would be useful.

packages/dds/tree/src/text/codePointUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Asserts that the requested `count` code points are fully consumable from `start`; silent truncation
+ * would misalign delta offsets applied to strings rather than surface the drift to the caller.
+ * @param str - The string to measure.
+ * @param start - The UTF-16 index in `str` to start measuring from. Must be in `[0, str.length]`.
+ * @param count - The number of Unicode code points to measure. Must be non-negative, and there must
+ * be at least `count` code points available in `str` starting at `start`.
+ * @internal
+ */
+export function utf16LengthForCodePoints(str: string, start: number, count: number): number {
+	assert(start >= 0 && start <= str.length, "start must be within str bounds");
```
Josmithr: Is the intention to export this from the package? If so, we should update these asserts to throw `UsageErrors`, in which case we can also provide specific numbers in the error message.

packages/dds/tree/src/text/codePointUtils.ts
```diff
@@ -0,0 +1,51 @@
+/*!
```
Josmithr, line 1: We should add a suite of unit tests for these functions.

packages/dds/tree/src/index.ts
```diff
@@ -403,5 +403,10 @@ export { FluidSerializableAsTree } from "./serializableDomainSchema.js";
 export { TableSchema, type System_TableSchema } from "./tableSchema.js";
 export { asAlpha, asBeta } from "./api.js";
 
-export { TextAsTree, FormattedTextAsTree } from "./text/index.js";
+export {
+	TextAsTree,
+	FormattedTextAsTree,
+	codePointCount,
+	utf16LengthForCodePoints,
```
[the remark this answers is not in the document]
Josmithr: `@internal` *is* a valid release tag. It indicates that something is only intended for use within the repo. This is correct.

packages/framework/quill-react/src/shared/runOnce.ts
```diff
@@ -0,0 +1,29 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import type { MutableRefObject } from "react";
+
+/**
+ * Runs `fn` exactly when the shared `isUpdatingRef` is not already set, with the ref
```
Josmithr:
```suggestion
 * Runs `fn` exactly once when the shared `isUpdatingRef` is not already set, with the ref
```

packages/framework/quill-react/src/plain/quillView.tsx
```diff
@@ -94,34 +91,54 @@ const TextEditorView = withMemoizedTreeObservations(({ root }: { root: TextAsTre
...
-
-			isUpdatingRef.current = false;
-		}
-	}
+	// Subscribe to incremental tree changes — sync tree → Quill.
+	// Kept in a separate effect so it re-subscribes correctly after React strict mode cleanup,
+	// independent of the Quill initialization guard above.
+	useEffect(() => {
+		return root.onCharactersChanged((ops) => {
+			runOnce(isUpdatingRef, () => {
```
Josmithr: I'm not following this. Don't we only want to run this when *not* `isUpdatingRef`? Doesn't the presence of that flag indicate that the update is happening as a result of a quill state update, rather than a tree update?

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -308,6 +316,168 @@ function formatToQuillAttributes(
...
+ */
+function formatToFullQuillAttributes(
+	format: FormattedTextAsTree.CharacterFormat,
+): Record<string, unknown> {
+	// Quill uses `null` (not `undefined`) to clear attributes, so we must use null
+	// for default-valued properties rather than omitting them.
+	// eslint-disable-next-line unicorn/no-null
+	const off = null;
+
+	const sizeValue =
```
Josmithr: Nit: would it make sense to hoist this out into a helper function?

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -308,6 +316,168 @@ function formatToQuillAttributes(
...
 
+/**
+ * Convert a CharacterFormat to Quill attributes that fully describe the formatting state.
+ * Unlike {@link formatToQuillAttributes}, this includes `null` for default-valued properties
+ * so that Quill clears any previously-set attributes. Used when applying formatting changes
+ * to retained (already-present) content via `updateContents`.
+ */
+function formatToFullQuillAttributes(
+	format: FormattedTextAsTree.CharacterFormat,
+): Record<string, unknown> {
```
Josmithr: Is there a more specific type we can use here? Or does Quill not have anything for this?

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -308,6 +316,168 @@ function formatToQuillAttributes(
...
+ * display matches the new tree state.
+ *
+ * `preEditContent` is the Quill document text (via `quill.getText()`) captured
+ * before applying these ops. It is needed to compute the UTF-16 width of removed
+ * atoms, which are no longer present in the tree after the edit.
+ *
+ * Returns `undefined` when the tree is out of sync with the delta (e.g., a concurrent edit raced with event delivery).
+ * The caller should fall back to a full diff in that case.
+ */
+function contentOpsToQuillDelta(
```
Josmithr, line 128: Nit: this seems like something that might be worth adding unit tests for. It might make sense to pull some of these conversion utilities out into their own module and adding a unit test suite for that module.

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -434,134 +604,129 @@ const FormattedTextEditorView = forwardRef<
...
-						const retainedStr = content.slice(utf16Pos, utf16Pos + op.retain);
-						const cpCount = codepointCount(retainedStr);
-
-						if (op.attributes) {
+			if (source !== "user") return;
+			runOnce(isUpdating, () => {
+				// Wrap all tree mutations in a transaction so they undo/redo as one atomic unit.
+				// If the node is not part of a branch (e.g. unhydrated), apply edits directly.
+				const branch = TreeAlpha.branch(root);
+				const applyDelta = (): void => {
```
Josmithr: Can we hoist this logic out into a helper function? Would be good to unit test the Quill deltas -> ArrayNode deltas conversion

Josmithr (commented):
Left some comments. Mostly about code factoring and unit tests.

Overall, looking good. But there is a lot of complexity (in particular in the array-node delta <-> quill delta conversions) that I think we should ensure is factored out / tested.

packages/dds/tree/src/text/codePointUtils.ts
daesunp: renamed str -> value

packages/dds/tree/src/text/codePointUtils.ts
daesunp: added example blocks to both functions

packages/dds/tree/src/text/codePointUtils.ts
daesunp: Updated with comment

packages/dds/tree/src/text/codePointUtils.ts
daesunp: Added `codePointUtils.spec.ts` with test suite

packages/framework/quill-react/src/shared/runOnce.ts
daesunp: Updated name to `runGuarded` and updated docs to be more clear

packages/framework/quill-react/src/plain/quillView.tsx
daesunp: The early return on the `if(isUpdatingRef.current) return is inside the `runGuarded` itself, so the body is skipped when an outgoing direction sync is in progress. But I think the name was a bit misleading to I renamed to `runGuarded`

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
daesunp: Updated to use `AttributeMap`

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
daesunp: added test suite for `applyQuillDeltaToTree` in `applyQuillDeltaToTree.tests.ts`.

I noticed that the conversion utilities are only used in `quillFormattedView.tsx` though, should we still move these into their own modules in case we want to reuse them for the future? Happy to make this change, just wanted to confirm!

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
daesunp: Moved to a `applyQuillDeltaToTree` function and also has test suite now

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
daesunp: moved to `sizeToQuillAttribute` function

packages/dds/tree/src/test/text/textDomain.spec.ts
```diff
@@ -80,6 +80,132 @@ describe("textDomain", () => {
 		});
 	});
 
+	describeHydration("onCharactersChanged", (_init, hydrated) => {
```
Josmithr, line 83: Nit: it would probably be worth adding tests for insertion / removal of empty ranges. I'm assuming we get no change notification in those cases, but it would be good to capture that explicitly.

packages/dds/tree/src/test/text/codePointUtils.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+		it("starts measuring from `start`", () => {
+			// "a😀b": skip 'a' (1 unit), measure 1 code point ('😀') = 2 units
+			assert.equal(utf16LengthForCodePoints("a😀b", 1, 1), 2);
+			// skip 'a' + '😀' (3 units), measure 1 code point ('b') = 1 unit
+			assert.equal(utf16LengthForCodePoints("a😀b", 3, 1), 1);
+		});
+
+		it("throws UsageError when start or count is out of range", () => {
+			assert.throws(() => utf16LengthForCodePoints("abc", -1, 1), UsageError); // negative start
```
Josmithr: We actually have a `validateUsageError` utility function you can use here to ensure the error has the right message, etc. I would recommend using that for these failure mode test cases.

packages/dds/tree/src/test/text/textDomain.spec.ts
Josmithr: (Same for the formatted text tests)

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -186,6 +200,62 @@ class StringArray extends sf.array("StringArray", SchemaFactory.string) {
 	}
 }
 
+/**
+ * Processes an array-node delta into a {@link TextAsTree.TextOp}[] and calls `callback`.
+ * Shared by both the plain `onCharactersChanged` (from `nodeChanged`) and formatted `onContentChanged`
```
Josmithr, line 206: Nit
```suggestion
 * @remarks
 * Shared by both the plain `onCharactersChanged` (from `nodeChanged`) and formatted `onContentChanged`
```

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -186,6 +200,62 @@ class StringArray extends sf.array("StringArray", SchemaFactory.string) {
...
+ * retain ops include an explicit `formattingChanged: boolean`. Otherwise `formattingChanged` is omitted.
+ * @param getChar - Returns the character string at the given array index in the **post-edit** tree.
+ * Only invoked for insert ops, where it must read the inserted character at the given index of the tree
+ * after the edit has been applied. Passing an accessor that reads pre-edit content will silently produce wrong text.
+ * Return `undefined` if the tree is out of sync with the delta; this triggers a full-reread fallback.
+ * @param callback - The user-supplied callback to invoke with the translated ops.
+ */
+export function processCharactersChangedDelta(
+	delta: readonly (ArrayNodeDeltaOp | ArrayNodeTreeChangedDeltaOp)[] | undefined,
+	getChar: (index: number) => string | undefined,
```
Josmithr: Nit: we generally prefer to avoid abbreviations when possible. Can we call this `getCharacter`?

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -186,6 +200,62 @@ class StringArray extends sf.array("StringArray", SchemaFactory.string) {
...
+	callback: (ops: readonly TextAsTree.TextOp[] | undefined) => void,
+): void {
+	if (delta === undefined) {
+		callback(undefined);
+		return;
+	}
+	let readPos = 0;
```
Josmithr: Nit: `readPosition`?

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -249,6 +319,69 @@ class StringArray extends sf.array("StringArray", SchemaFactory.string) {
...
+		/**
+		 * Discriminator identifying this op as a retain.
+		 */
+		readonly type: "retain";
+		/**
+		 * The number of Unicode code points to retain.
+		 */
+		readonly count: number;
+		/**
+		 * Present only on retain ops delivered by {@link @fluidframework/tree#FormattedTextAsTree.Members.onContentChanged};
```
Josmithr, line 340: Nit: we should reformat this to state what it _is_ first, then add any deeper notes in a `@remarks` block.

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -263,6 +396,7 @@ export namespace TextAsTree {
 
 	/**
 	 * Interface for a text node.
+	 * @sealed
```
Josmithr: Nit: we usually put `@sealed` near the bottom by the release tag.

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -325,6 +459,22 @@ export namespace TextAsTree {
...
+		 * or `undefined` when a delta could not be computed (e.g. during a schema upgrade).
+		 * @returns A cleanup function that unsubscribes the callback when called.
+		 * @remarks
+		 * Only fires on shallow changes — inserts and removes.
+		 * It does not fire on deep changes such as formatting property updates on existing characters.
+		 * For formatted text, use {@link @fluidframework/tree#FormattedTextAsTree.Members.onContentChanged} to also receive deep changes.
+		 *
+		 * All counts in the delivered ops are in Unicode code points, not UTF-16 code units.
+		 * For characters outside the Basic Multilingual Plane (e.g. emoji), one code point
+		 * corresponds to two UTF-16 code units — convert before using the counts as string indices.
```
Josmithr, line 478: Great docs! Thanks!!

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -169,6 +173,11 @@ class TextNode
 			// If this node does not have a corresponding branch, then it is unhydrated.
 			// I.e., it is not part of a collaborative session yet.
 			// Therefore, we don't need to run the edits as a transaction.
+			// Note: for unhydrated nodes each atom edit fires a separate `treeChanged` event,
```
Josmithr, line 176: Can we mitigate this somehow? Could we leverage `withBufferedEvents` for this?

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -177,6 +186,34 @@ class TextNode
 			});
 		}
 	}
+	/**
+	 * Returns the character string at the given atom index, or `undefined` if out of bounds.
+	 * Line atoms expand to `"\n"`; text atoms return their underlying code point(s).
+	 */
+	private atomChar(i: number): string | undefined {
```
Josmithr: Nit: we generally recommend naming functions/methods using verbs. Also recommended to avoid abbreviations where possible. How about `getAtomCharacterAt` (or something)?

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -285,36 +308,324 @@ function quillAttributesToPartial(
...
-				: `${format.size}px`;
+		attributes.size = sizeToQuillAttribute(format.size);
 	}
 	if (format.font !== defaultFont) attributes.font = format.font;
 	return attributes;
 }
 
+/**
+ * Convert a CharacterFormat to Quill attributes that fully describe the formatting state.
+ * Unlike {@link formatToQuillAttributes}, this includes `null` for default-valued properties
```
Josmithr: Nit:
```suggestion
 * @remarks
 * Unlike {@link formatToQuillAttributes}, this includes `null` for default-valued properties
```

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
Josmithr: I still think it would probably make sense, yeah. These are complex enough that they deserve unit test coverage, in my opinion. And then we can rename the test module after the module we create for this stuff.

packages/dds/tree/src/text/codePointUtils.ts
daesunp: Updated with suggestion

packages/dds/tree/src/text/codePointUtils.ts
daesunp: Updated with explanation comment

packages/dds/tree/src/text/codePointUtils.ts
daesunp: Updated to throw usage errors

packages/dds/tree/src/text/codePointUtils.ts
daesunp: Added test suite

packages/dds/tree/src/test/text/codePointUtils.spec.ts
daesunp: Updated to use `validateUsageError`

packages/dds/tree/src/text/textDomain.ts
daesunp: Updated with suggestion

packages/dds/tree/src/text/textDomain.ts
daesunp: Updated with suggestion

packages/dds/tree/src/text/textDomain.ts
daesunp: Updated with suggestion

packages/dds/tree/src/text/textDomain.ts
daesunp: Updated with suggestion

packages/dds/tree/src/text/textDomain.ts
daesunp: Updated with suggestion

packages/dds/tree/src/text/textDomainFormatted.ts
daesunp: Updated with suggestion

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
daesunp: Updated with suggestion

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
daesunp: Moved to quillAttributeUtils.ts and added unit tests

packages/dds/tree/src/text/textDomainFormatted.ts
daesunp: Unfortunately when using `withBufferedEvents` and we touch two fields, we invalidate and return undefined, so we can't use it as a workaround.

Updated the docs mentioning this

Josmithr: We should file a bug for this and probably reference it here. I think any time the hydrated vs unhydrated behavior differs, that constitutes a bug that we need to (eventually) fix.

Josmithr: Do we have test coverage for this? If not, we should add it and similarly document the bug and what the intended behavior should be.

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -244,77 +105,272 @@ function createLineAtom(
...
-	if (format.italic) attributes.italic = true;
-	if (format.underline) attributes.underline = true;
-	if (format.size !== defaultSize) {
-		// Convert pixel value back to Quill size name if possible
-		attributes.size =
-			format.size in sizeReverse
-				? sizeReverse[format.size as keyof typeof sizeReverse]
-				: `${format.size}px`;
+export function applyQuillDeltaToTree(root: FormattedTextAsTree.Tree, delta: Delta): void {
+	const branch = TreeAlpha.branch(root);
```
brrichards: Should this be using `TreeAlpha.context().isBranch()` with a transaction label here? The plain text view passes `label: editLabelRef.current ` through `context.runTransaction()`, but this function uses TreeAlpha.branch() and drops the label. I think without the label, formatted editor edits won't be reachable by the scoped undo button since they'd be unlabeled and only show up in the global undo stack.

daesunp: Added optional label parameter to `applyQuillDeltaToTree` and routed through `TreeAlpha.context().runTransaction(applyDelta, { label })`

brrichards (approved):
LGTM!

PR #26902, opened 2026-03-31
Server release reports

kekachmar:
## Description

- Uploads the release reports for server packages (historian, gitrest, routerlicious).
- Publishes exact versions for the selected release group.

Reviewer Guidance
- Creates a new file for the upload steps `include-upload-release-reports.yml`
- Server files are uploaded to blob storage with the path `/server/prerelease/<filename>.json`


build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
```diff
@@ -10,9 +10,15 @@ import type { Logger } from "@fluidframework/build-tools";
...
 import { formatISO } from "date-fns";
 
-import { semverFlag } from "../../flags.js";
-import { BaseCommand } from "../../library/commands/base.js";
-import { type ReleaseReport, toReportKind } from "../../library/release.js";
+import { releaseGroupFlag, semverFlag } from "../../flags.js";
+import { BaseCommand, type ReleaseReport, toReportKind } from "../../library/index.js";
+import type { ReleaseGroup } from "../../releaseGroups.js";
+
+const releaseGroupsWithTestVersionsInManifest = new Set<ReleaseGroup>([
```
Abe27342: I don't follow flub particularly much, but from what I've gathered on its architecture it seems to try to not be configured by default to be specific to FluidFramework. Maybe this should be something that gets specified in the fluid build config file?

build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
```diff
@@ -41,6 +47,11 @@ export class UnreleasedReportCommand extends BaseCommand<typeof UnreleasedReport
 				"Branch name. For release branches, the manifest file is uplaoded by build number and not by current date.",
```
tylerbutler: nit, preexisting

```suggestion
				"Branch name. For release branches, the manifest file is uploaded by build number and not by current date.",
```

build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
```diff
@@ -79,14 +91,31 @@ async function generateReleaseReport(
 	version: string,
 	outDir: string,
 	branchName: string,
+	releaseGroup: ReleaseGroup | undefined,
```
tylerbutler, line 118: This function is already quite complex with a lot of inputs. Can we create a separate function that takes a `ReleaseReport`, updates/reshapes it as needed, and then call writeReport on the result? It might be a little bit of duplication but since release group is basically a filter, the logic should be simpler if you first filter out the packages you don't care about _then_ call generateReleaseReport. This function should ideally just be about writing the reports out; filtering can be done beforehand.

build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
```diff
@@ -110,6 +139,29 @@ async function generateReleaseReport(
...
+			.filter(([, details]) => details?.releaseGroup === releaseGroup)
+			.map(([packageName]) => packageName),
+	);
+
+	if (packageNames.size === 0) {
+		throw new Error(`No packages found for release group '${releaseGroup}' in the full report.`);
+	}
+
+	return packageNames;
+}
```
tylerbutler: You should be able to replace this with something like this within the command class:

```ts
const packageNames = new Set(
    this.context.packagesInReleaseGroup(releaseGroup).map(pkg => pkg.name)
  );
```

build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
```diff
@@ -156,8 +208,17 @@ async function updateReportVersions(
 	report: ReleaseReport,
 	ignorePackageList: Set<string>,
 	version: string,
+	releaseGroup: ReleaseGroup | undefined,
```
tylerbutler, line 199: Needs to be added to the docstring.

build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
tylerbutler: Yeah this shouldn't be hardcoded. If it's required to explicitly define this rather than inferring it from some other property of a release group, then it needs to be provided in config (a little tricky since this command has no config today) or as command line options.

build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
```diff
@@ -156,8 +208,17 @@ async function updateReportVersions(
...
 	log: Logger,
 ): Promise<void> {
+	const useTestVersionForReleaseGroupPackages =
+		releaseGroup !== undefined &&
+		releaseGroupsWithTestVersionsInManifest.has(releaseGroup) &&
+		isInternalTestVersion(version);
+
+	const packageNamesForReleaseGroup =
+		releaseGroup === undefined ? undefined : getPackageNamesForReleaseGroup(report, releaseGroup);
```
tylerbutler: Can be simplified using `this.context` within the command class.

build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
```diff
@@ -187,6 +248,12 @@ async function updateReportVersions(
 
 		const packageInfo = report[packageName];
 
+		if (useTestVersionForReleaseGroupPackages && packageNamesForReleaseGroup?.has(packageName)) {
+			report[packageName].ranges.caret = version;
+			report[packageName].version = version;
+			continue;
+		}
```
tylerbutler: It seems like this should be centralized someplace else. If I understand correctly, the basic overall flow is:

1. Generate a release report "as usual"
2. Filter the report to a release group
3. Update its versions
4. Write out the report

Putting this in its own named function should make it clearer how it fits into the overall flow.

build-tools/packages/build-cli/src/commands/release/report.ts
```diff
@@ -128,6 +128,7 @@ export abstract class ReleaseReportBaseCommand<
 		mode: ReleaseSelectionMode = this.defaultMode,
 		releaseGroupOrPackage?: ReleaseGroup | ReleasePackage,
 		includeDependencies = true,
+		useCurrentVersion = false,
```
tylerbutler, line 131: I don't love this additional flag - why is it needed?

build-tools/packages/build-cli/src/commands/release/report.ts
```diff
@@ -273,15 +282,26 @@ export abstract class ReleaseReportBaseCommand<
...
-					const [, previousMinor] = getPreviousVersions(repoVersion);
-					this.info(
-						`The in-repo version of ${chalk.blue(releaseGroupOrPackage)} is ${chalk.yellow(
-							repoVersion,
-						)}, but there's no release for that version. Picked previous minor version instead: ${chalk.green(
-							previousMinor ?? "undefined",
-						)}. If you want to create a report for a specific version, check out the tag for the release and re-run this command.`,
-					);
-					latestReleasedVersion = sortedByVersion[0];
+					if (useCurrentVersion) {
```
tylerbutler, line 285: Maybe we can just get rid of the old behavior? Does anything depend on it?

build-tools/packages/build-cli/src/commands/release/report.ts
```diff
@@ -332,7 +361,7 @@ export default class ReleaseReportCommand extends ReleaseReportBaseCommand<
 
     The command operates in two modes: "whole repo" or "release group." The default mode is "whole repo." In this mode, the command will look at the git tags in the repo to determine the versions, and will include all release groups and packages in the repo. You can control which version of each package and release group is included in the report using the --interactive, --mostRecent, and --highest flags.
 
-    The "release group" mode can be activated by passing a --releaseGroup flag. In this mode, the specified release group's version will be loaded from the repo, and its immediate Fluid dependencies will be included in the report. This is useful when we want to include only the dependency versions that the release group depends on in the report.`;
+	    The "release group" mode can be activated by passing a --releaseGroup flag. In this mode, the specified release group's version will be loaded from the repo, and its immediate Fluid dependencies will be included in the report. This is useful when we want to include only the dependency versions that the release group depends on in the report.`;
```
tylerbutler: Whitespace issue.

build-tools/packages/build-cli/src/commands/release/report.ts
```diff
@@ -349,6 +378,11 @@ export default class ReleaseReportCommand extends ReleaseReportBaseCommand<
 				"Generate a release report for each package and release group in the repo interactively.",
 			command: "<%= config.bin %> <%= command.id %> -i",
 		},
+		{
+			description:
+				"Generate a release report for the historian release group at the in-repo version.",
+			command: "<%= config.bin %> <%= command.id %> -g historian",
```
tylerbutler: Wouldn't this need `--useCurrentVersion`?

tools/pipelines/templates/include-publish-npm-package.yml
```diff
@@ -89,9 +105,9 @@ stages:
       tagName: ${{ parameters.tagName }}
 
 # Only upload release manifest files for runs in the internal project (i.e. CI runs, not PR runs), for the main, test, and release branches, and
-# for the build of the client release group (we don't need manifests for anything else).
+# for builds of release groups that currently publish manifests.
```
tylerbutler, line 92: At this point aren't we publishing manifests for every release group?

tools/pipelines/templates/include-publish-npm-package.yml
```diff
@@ -89,9 +105,9 @@ stages:
...
 # Enabling this template for every PR run risks overwriting existing manifest files uploaded to Azure blobs. Therefore, it's crucial to restrict this template to commits merged into the main and release branches.
-- ${{ if and(eq(variables['System.TeamProject'], 'internal'), eq(parameters.tagName, 'client'), or(eq(variables['Build.SourceBranch'], 'refs/heads/main'), startsWith(variables['Build.SourceBranch'], 'refs/heads/release/'), startsWith(variables['Build.SourceBranch'], 'refs/heads/test/'))) }}:
+- ${{ if and(eq(variables['System.TeamProject'], 'internal'), or(eq(parameters.tagName, 'client'), eq(parameters.tagName, 'server'), eq(parameters.tagName, 'historian'), eq(parameters.tagName, 'gitrest')), or(eq(variables['Build.SourceBranch'], 'refs/heads/main'), startsWith(variables['Build.SourceBranch'], 'refs/heads/release/'), startsWith(variables['Build.SourceBranch'], 'refs/heads/test/'))) }}:
```
tylerbutler: Ouch... this is almost impossible to grok.

build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
kekachmar: We will need to select packages for each release group to replace `clientPackageVersion` in `updateReportVersions()` because the `fluid-framework` package will not be included in the filtered reports for routerlicious, historian, and gitrest.
https://github.com/microsoft/FluidFramework/blob/258aa332a9ab237e8a72b6747968669be9d4db20/build-tools/packages/build-cli/src/commands/release/report-unreleased.ts#L155

For example:
    "client": "fluid-framework",
    "server": "@fluidframework/server-routerlicious",
    "gitrest": "@fluidframework/gitrest",
    "historian": "@fluidframework/historian",

tylerbutler (approved):
I'm OK with the changes if they unblock you. It's not your responsibility to clean up all the mess. Most of my objections are kind of required because the existing code is in bad shape. Just because you walked into a messy room to get something doesn't mean you're expected to clean the whole room while you're there. :)

tools/pipelines/templates/include-publish-npm-package.yml
kekachmar: `build-tools` is also considered a release group. Release reports are not published for build-tools

kekachmar: https://github.com/microsoft/FluidFramework/blob/f7ff4ecc9f2bd9e51596adaa197c87812aff6eff/tools/pipelines/build-build-tools.yml#L111

build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
tylerbutler: That makes sense to me - basically picking a representative "main" package for each release group. You can hard code that if needed. Ideally it would come from config but that's additional work.

build-tools/packages/build-cli/src/commands/release/report.ts
kekachmar: useCurrentVersion allows versions that have not been released to be included in the release reports.

tools/pipelines/templates/include-publish-npm-package.yml
Abe27342: using [in](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#in) might make the portion of the expression being changed a bit simpler here.

Probably the more important thing which would make it more readable would be extracting parts of the logic into meaningful units. That could either be in a variables block within the template if they are local or somewhere more common, e.g. we already expose `testBuild` in tools\pipelines\templates\include-vars.yml.

I think you can also multiline template expressions in this context. You definitely can in the variables context, see above file.

If I read the clauses right, the logic here is
```
${{ if and(
  eq(variables['System.TeamProject'], 'internal'),
  in(parameters.tagName, 'client', 'server', 'historian', 'gitrest'),
  or(isMainBranch, isTestBranch, isReleaseBranch)
) }}
```

where the last 3 are pretty reasonable variables to stick in a common place (we already have `testBuild`, though I'd prefer the naming I used)

tools/pipelines/templates/include-publish-npm-package.yml
```diff
@@ -115,6 +115,16 @@ stages:
     variables:
     - group: ado-feeds
     jobs:
-    - template: /tools/pipelines/templates/upload-dev-manifest.yml@self
-      parameters:
-        buildDirectory: ${{ parameters.buildDirectory }}
+    - ${{ if or(eq(parameters.tagName, 'server'), eq(parameters.tagName, 'historian'), eq(parameters.tagName, 'gitrest')) }}:
```
Abe27342: use `in` here as well

tools/pipelines/templates/include-publish-npm-package.yml
Abe27342: (also that whole expression could be a variable local to this template `shouldUploadReleaseManifests` or something like that, and if for some weird reason multiline template expressions are only supported in variable contexts then IMO it should for readability)

tools/pipelines/server-gitrest.yml
```diff
@@ -42,6 +42,7 @@ trigger:
     - main
     - next
     - lts
+    - test/*
```
Abe27342: test branch triggers should go away before merging. We don't want pipelines triggering by default on such branches.

tools/pipelines/templates/upload-server-manifest.yml
```diff
@@ -0,0 +1,79 @@
...
+jobs:
+  - job:
+    displayName: Upload Release Reports
+    variables:
+    - group: storage-vars
+    - name: version
+      value: $[stageDependencies.build.build.outputs['SetVersion.version']]
+    - name: isReleaseBuild
+      value: $[eq(variables['release'], 'release')]
+    steps:
```
Abe27342, line 20: other template has an explicit checkout--any reason to not have that here?

```
    - checkout: self
      path: $(FluidFrameworkDirectory)
```

tools/pipelines/templates/include-publish-npm-package.yml
```diff
@@ -115,6 +115,16 @@ stages:
...
+      - template: /tools/pipelines/templates/upload-server-manifest.yml@self
+        parameters:
+          buildDirectory: '${{ parameters.buildDirectory }}'
+          releaseGroup: ${{ parameters.tagName }}
```
Abe27342, line 122: `releaseReportCategory` is a pure function of tagName (which === releaseGroup apparently?). Can you remove that parameter and move this computation into the template?

tools/pipelines/templates/upload-server-manifest.yml
```diff
@@ -0,0 +1,79 @@
...
+        targetType: 'inline'
+        workingDirectory: ${{ parameters.buildDirectory }}
+        script: |
+          set -eu -o pipefail
+          mkdir upload_release_reports
+          flub release report-unreleased --version $(version) --fullReportFilePath generate_release_reports/manifest.full.json --outDir upload_release_reports --branchName '$(Build.SourceBranch)' --releaseGroup ${{ parameters.releaseGroup }}
+
+    - ${{ if ne(parameters.releaseReportCategory, '') }}:
+      - task: Bash@3
+        displayName: Rename release report files
```
Abe27342: maybe a me problem (not particularly fluent in bash when it gets into these sorts of manipulations), but the intent of this step is not very clear. Probably could use an explanatory comment

tools/pipelines/templates/upload-server-manifest.yml
kekachmar: I've added it to this template

tools/pipelines/templates/include-publish-npm-package.yml
```diff
@@ -89,9 +89,9 @@ stages:
       tagName: ${{ parameters.tagName }}
 
 # Only upload release manifest files for runs in the internal project (i.e. CI runs, not PR runs), for the main, test, and release branches, and
-# for the build of the client release group (we don't need manifests for anything else).
+# for builds of release groups that currently publish manifests.
 # Enabling this template for every PR run risks overwriting existing manifest files uploaded to Azure blobs. Therefore, it's crucial to restrict this template to commits merged into the main and release branches.
-- ${{ if and(eq(variables['System.TeamProject'], 'internal'), eq(parameters.tagName, 'client'), or(eq(variables['Build.SourceBranch'], 'refs/heads/main'), startsWith(variables['Build.SourceBranch'], 'refs/heads/release/'), startsWith(variables['Build.SourceBranch'], 'refs/heads/test/'))) }}:
+- ${{ if and(eq(variables['System.TeamProject'], 'internal'), in(parameters.tagName, 'client', 'server', 'historian', 'gitrest'), in('true', variables.isMainBranch, variables.isReleaseBranch, variables.testBuild)) }}:
```
Abe27342: nit: `or` the last few args rather than `in('true', ...)`?. This looks so much better though, thanks!

Abe27342 (approved):

PR #26911, opened 2026-04-01
Expose text alpha APIs to LLMs

brrichards:
## Summary

Exposes the alpha plain text APIs using typeFactory ExposeMethod so tree agent can see them when encountering a TextNode

### DIFF
--- a/packages/dds/tree/src/text/textDomain.ts
+++ b/packages/dds/tree/src/text/textDomain.ts
@@ -4,6 +4,14 @@
  */
 
 import { compareArrays, debugAssert } from "@fluidframework/core-utils/internal";
+import {
+	buildFunc,
+	exposeMethodsSymbol,
+	type ExposedMethods,
+	type IExposedMethods,
+	// eslint-disable-next-line import-x/no-internal-modules
+} from "@fluidframework/type-factory/alpha";
+import { typeFactory as tf } from "@fluidframework/type-factory/internal";
 
 import { EmptyKey, mapCursorField, type ITreeCursorSynchronous } from "../core/index.js";
 import {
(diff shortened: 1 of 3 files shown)


packages/dds/tree/src/text/textDomain.ts
```diff
@@ -26,8 +34,71 @@ class TextNode
...
+			buildFunc({
+				description: "Gets the number of characters currently in the text.",
+				returns: tf.number(),
+			}),
+		);
+		methods.exposeMethod(
+			TextNode,
+			"characters",
+			buildFunc({
+				description: "Gets an iterable over the characters currently in the text.",
```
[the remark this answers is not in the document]
brrichards: fixed

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -26,8 +34,63 @@ class TextNode
 	extends sf.object("Text", {
 		content: SchemaFactory.required([() => StringArray], { key: EmptyKey }),
 	})
-	implements TextAsTree.Members
+	implements TextAsTree.Members, IExposedMethods
 {
+	public static [exposeMethodsSymbol](methods: ExposedMethods): void {
```
Josmithr, line 39: Nit: let's be sure to document public methods

brrichards: Added docs!

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -26,8 +34,76 @@ class TextNode
 	extends sf.object("Text", {
 		content: SchemaFactory.required([() => StringArray], { key: EmptyKey }),
 	})
-	implements TextAsTree.Members
+	implements TextAsTree.Members, IExposedMethods
 {
+	/**
+	 * Exposes the following methods to an LLM agent:
```
noencke: I don't think this doc comment needs to enumerate all the functions it exposes. If it needs to exist for lint or something, a short line will do, otherwise just delete it.

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -26,8 +34,76 @@ class TextNode
...
+				["index", tf.number()],
+				["additionalCharacters", tf.string()],
+			),
+		);
+		methods.exposeMethod(
+			TextNode,
+			"removeRange",
+			buildFunc(
+				{
+					description: "Remove a range of characters from the text by character index.",
```
noencke: You could mention in here that startIndex defaults to 0 and endIndex defaults to the length. Ideally we'd have a way to put comments on parameters but we don't have that yet.

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -26,8 +34,76 @@ class TextNode
...
+				},
+				["startIndex", tf.union([tf.number(), tf.undefined()])],
+				["endIndex", tf.union([tf.number(), tf.undefined()])],
+			),
+		);
+		methods.exposeMethod(
+			TextNode,
+			"fullString",
+			buildFunc({
+				description: "Copy the content of this text node into a string.",
```
noencke:
```suggestion
				description: "Return a copy of this text node's content as a string.",
```

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -26,8 +34,76 @@ class TextNode
...
+				returns: tf.string(),
+			}),
+		);
+		methods.exposeMethod(
```
noencke, line 74: FYI you can also expose getter/setter properties to the LLM in case that influences your API design decisions

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -26,8 +34,76 @@ class TextNode
...
+			buildFunc({
+				description: "Gets the number of characters currently in the text.",
+				returns: tf.number(),
+			}),
+		);
+		methods.exposeMethod(
+			TextNode,
+			"charactersCopy",
+			buildFunc({
+				description: "Optimized way to get a copy of the characters in an array.",
```
noencke: I'd elaborate just a bit here. This is the only function where it isn't pretty obvious what it does from the surface.

brrichards: Elaborated, and also added that data is stored as Unicode code points. It'll probably be useful for LLMs to know

packages/dds/tree/src/text/textDomain.ts
brrichards: Updated!

packages/dds/tree/src/text/textDomain.ts
brrichards: removed. Lint doesn't mind

packages/dds/tree/src/text/textDomain.ts
brrichards: Added clarifying comments for defaults

noencke (approved):

PR #26912, opened 2026-04-01
api(client-telemetry-utils): deprecate ITelemetryLoggerExt methods and related types

jason-ha:
## Deprecated API

The external-facing `ITelemetryLoggerExt` interface's methods and its associated telemetry event types are deprecated. The `ITelemetryLoggerExt` interface will continue to exist such that it can be passed across FF API surface as needed.

The following types and methods from `@fluidframework/telemetry-utils` are deprecated:

| API | Kind | Package Export |
|-----|------|----------------|
| `TelemetryEventCategory` | Type alias | `@legacy` `@beta` |
| `ITelemetryGenericEventExt` | Interface | `@legacy` `@beta` |
| `ITelemetryErrorEventExt` | Interface | `@legacy` `@beta` |
| `ITelemetryPerformanceEventExt` | Interface | `@legacy` `@beta` |
| `ITelemetryLoggerExt.sendTelemetryEvent()` | Method | `@legacy` `@beta` |
| `ITelemetryLoggerExt.sendErrorEvent()` | Method | `@legacy` `@beta` |
| `ITelemetryLoggerExt.sendPerformanceEvent()` | Method | `@legacy` `@beta` |

### Context

`ITelemetryLoggerExt` was an internal interface that leaked into the public API surface. While method access continues the internal implementation cannot safely evolve.

### Approach

The type will be "erased" so that internals can evolve. The erased type will continue to inherit from `ITelemetryBaseLogger` and its `send` method will remain usable.

#### Internals

The internal interface is now named `TelemetryLoggerExt`.
`@fluidframework/telemetry-utils/internal` exports `ITelemetryLoggerExt` as `@internal` and exactly is an alias for `TelemetryLoggerExt`.

Any uses of `ITelemetryLoggerExt` exposed through the `@legacy` API surface must use `ITelemetryLoggerExt` imported from `@fluidframework/telemetry-utils/legacy`. `toITelemetryLoggerExt` and `extractTelemetryLoggerExt` helpers have been added to `@fluidframework/telemetry-utils/internal` exports to aid in conversion as needed.

### Compatibility Concerns

There are no runtime changes at this time.
If externally `ITelemetryLoggerExt` interface has been implemented to pass to FF, instead implement `ITelemetryBaseLogger` from `@fluidframework/core-interfaces` and call `@fluidframework/telemetry-utils`'s `createChildLogger({ logger: your_logger_impl })` to safely create an expected `ITelemetryLoggerExt`.

### Expected Future Timeline

Removal of methods and related interfaces are expected in v2.110.


.changeset/lovely-groups-relate.md
```diff
@@ -0,0 +1,17 @@
...
+`ITelemetryLoggerExt` is only for internal Fluid Framework use. The following deprecations help formalize that. See [issue #26910](https://github.com/microsoft/FluidFramework/issues/26910) for details.
+
+| API | Kind | Package Export |
+|-----|------|----------------|
+| `TelemetryEventCategory` | Type alias | `@legacy` `@beta` |
+| `ITelemetryGenericEventExt` | Interface | `@legacy` `@beta` |
+| `ITelemetryErrorEventExt` | Interface | `@legacy` `@beta` |
+| `ITelemetryPerformanceEventExt` | Interface | `@legacy` `@beta` |
+| `ITelemetryLoggerExt.sendTelemetryEvent()` | Method | `@legacy` `@beta` |
+| `ITelemetryLoggerExt.sendErrorEvent()` | Method | `@legacy` `@beta` |
```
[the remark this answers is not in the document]
jason-ha: There don't appear to be any double leading pipes.

packages/utils/telemetry-utils/api-report/telemetry-utils.legacy.beta.api.md
```diff
@@ -18,22 +18,25 @@ export class EventEmitterWithErrorHandling<TEvent extends IEvent = IEvent> exten
...
 export interface ITelemetryLoggerExt extends ITelemetryBaseLogger {
+    // @deprecated
     sendErrorEvent(event: ITelemetryErrorEventExt, error?: unknown): void;
-    sendPerformanceEvent(event: ITelemetryPerformanceEventExt, error?: unknown, logLevel?: typeof LogLevel.verbose | typeof LogLevel.default): void;
-    sendTelemetryEvent(event: ITelemetryGenericEventExt, error?: unknown, logLevel?: typeof LogLevel.verbose | typeof LogLevel.default): void;
+    // @deprecated
+    sendPerformanceEvent(event: ITelemetryPerformanceEventExt, error?: unknown, logLevel?: LogLevel): void;
+    // @deprecated
+    sendTelemetryEvent(event: ITelemetryGenericEventExt, error?: unknown, logLevel?: LogLevel): void;
 }
```
jason-ha, line 40: restore logLevel argument tyoe

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -369,15 +407,25 @@ export class TaggedLoggerAdapter implements ITelemetryBaseLogger {
...
+ * @internal
+ *
+ * @privateRemarks
+ * Return type is both TelemetryLoggerExt and ITelemetryLoggerExt to allow for
+ * easier internal usage without needing to type erase or un-type-erase the
+ * logger.
+ *
+ * If always creating a child logger for direct external exposure, consider
+ * using {@link createChildLogger} from /legacy API instead, which returns
+ * exactly an {@link ITelemetryLoggerExt}.
```
lindsnguyen, line 430: Question: Just for my own understanding, the reason this works is because while we want the returned object to be useable as the internal object, by the time it gets passed outside of fluid it can only be used as a `ITelemetryLoggerExt` since `TelemetryLoggerExt` isn't visable?

experimental/dds/tree/src/SharedTree.ts
```diff
@@ -552,11 +555,12 @@ export class SharedTree extends SharedObject<ISharedTreeEvents> implements NodeI
 		const historyPolicy = this.getHistoryPolicy(options);
 		this.summarizeHistory = historyPolicy.summarizeHistory;
 
-		this.logger = createChildLogger({
+		this.loggerInternal = createChildLogger({
 			logger: runtime.logger,
 			namespace: 'SharedTree',
 			properties: sharedTreeTelemetryProperties,
 		});
+		this.logger = this.loggerInternal;
```
lindsnguyen, line 569: super nit: Took me a bit to understand why we have both `logger` and `loggerInternal` but I see now that logger is public while `loggerInternal` is private (duh, hence the name) But later on we sometimes use `logger` even though `loggerInternal` could be used. I was confused wondering if there was some meaningful reason to use one over the other when I'm pretty sure (?) it doesn't matter. Could we use the internal one for all places where it's passed around in this file? (lines 565 and line 988).

lindsnguyen (approved):
In general, I understand the change so approving but it might make sense to wait for others approval since I'm relatively new to the codebase and the changes touch a lot of files.

one more question: For files that are at the boundary and has access to both the external `logger` and internal `loggerInternal` I think they're used interchangeably but I can't tell if there's a reason for this or if it truly doesn't matter. If it actually doesn't matter, would it make sense to always use/pass around `loggerInternal` just to keep it consistent/clear? (for example, in `epochTracker.ts` has both, although based on your private remark maybe it should only have an internal logger? but at line 619 we're using `logger` instead of `loggerInternal`) Could be I'm reading into it too much but want to make sure I understand the different parts.

packages/utils/telemetry-utils/src/logger.ts
jason-ha: Correct.

experimental/dds/tree/src/SharedTree.ts
jason-ha: In some places they would be interchangeable. But they won't always be. Once the breaking change comes in `ITelemetryLoggerExt` (public version) isn't usable (callable) except where `ITelemetryBaseLogger` is the only actual need. Both `ITelemetryLoggerExt` and `TelemetryLoggerExt` extend `ITelemetryBaseLogger`.
If you were browsing the code and types, you'd want to look at [fix(client): updates to accommodate ITelemetryLoggerExt erasing](https://github.com/microsoft/FluidFramework/pull/26912/commits/aa352aee860bcfec009c652d1f6f3361bb88db49) file state because that is before I rollback `ITelemetryLoggerExt` being an `ErasedType`.

.changeset/lovely-groups-relate.md
```diff
@@ -0,0 +1,17 @@
+---
+"@fluidframework/telemetry-utils": minor
+"__section": deprecation
+---
+Deprecate ITelemetryLoggerExt methods and related types
```
alexvy86, line 5: I know it feels a bit redundant when the PR talks about it and the changeset already points to it, but I think the changeset should have its own copy of the guidance for consumers on what to do (if anything) about this change, specifically the part about "if you were implementing X, now you should implement Y instead and combine with calling Z".

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -24,15 +24,44 @@ import {
...
+ */
+export function toITelemetryLoggerExt(logger: TelemetryLoggerExt): ITelemetryLoggerExt {
+	return logger as unknown as ITelemetryLoggerExt;
+}
+
+/**
+ * Un-type-erase the {@link ITelemetryLoggerExt}.
+ * @internal
+ */
+export function extractTelemetryLoggerExt<
```
alexvy86, line 63: Curious: do we know if TS is smart enough to inline/remove calls to functions like these (that just change the type of something) during compilation? Or do they introduce a bit of extra latency in the logging paths that need to call them?

packages/utils/telemetry-utils/src/telemetryTypes.ts
```diff
@@ -104,15 +112,65 @@ export interface ITelemetryPerformanceEventExt extends ITelemetryGenericEventExt
 	duration?: number;
 }
 
+/**
+ * An extended {@link @fluidframework/core-interfaces#ITelemetryBaseLogger} which allows for more lenient event types.
+ *
+ * @remarks
+ * `ITelemetryBaseLogger` should be used when loggers are passed between layers. This interface exists
+ * to patch cases where `ITelemetryLoggerExt` leaked out.
```
alexvy86: The second part feels more like `@privateremarks` to me, but also I'm not sure I get it. "`ITelemetryLoggerExt` exists to patch cases where _itself_ leaked out"?
```suggestion
 * @remarks
 * When loggers are passed between layers, `ITelemetryBaseLogger` should be used instead of this. 
 * @privateremarks
 * This interface exists to patch cases where `ITelemetryLoggerExt` leaked out.
```

packages/utils/telemetry-utils/src/telemetryTypesUndeprecated.ts
```diff
@@ -0,0 +1,51 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/* eslint-disable import-x/no-deprecated -- This file specifically works with deprecated types to undeprecate them for internal use */
+
+// This file contains exports that "un-deprecate" certain types that we want to
+// continue using internally while exposing deprecated version that will be
+// removed in a future release.
```
alexvy86, line 10: Neat :)

packages/utils/telemetry-utils/package.json
```diff
@@ -16,47 +16,48 @@
...
 			"require": {
-				"types": "./dist/index.d.ts",
-				"default": "./dist/index.js"
+				"types": "./dist/internal.d.ts",
+				"default": "./dist/internal.js"
 			}
 		}
 	},
-	"main": "lib/index.js",
+	"main": "lib/api.js",
```
alexvy86, line 47: I think there's an argument for sticking with ìndex.js` for consistency across the repo (and to avoid the need for specifying it explicitly in the npm scripts below + api-extractor.json?), but I personally don't dislike this. Just bringing it up in case it sparks discussion.

packages/utils/telemetry-utils/package.json
```diff
@@ -16,47 +16,48 @@
...
 	"types": "lib/public.d.ts",
 	"scripts": {
 		"api": "fluid-build . --task api",
-		"api-extractor:commonjs": "flub generate entrypoints --resolutionConditions require --outFileLegacyBeta legacy --outDir ./dist",
-		"api-extractor:esnext": "flub generate entrypoints --outFileLegacyBeta legacy --outDir ./lib --node10TypeCompat",
+		"api-extractor:commonjs": "flub generate entrypoints --resolutionConditions require --mainEntrypoint ./src/api.ts --outFileLegacyBeta legacy --outDir ./dist",
+		"api-extractor:esnext": "flub generate entrypoints --mainEntrypoint ./src/api.ts --outFileLegacyBeta legacy --outDir ./lib",
```
alexvy86, line 52: For my education, can you elaborate a bit on why `--node10TypeCompat` goes away here and is replaced with the new npm script?

alexvy86 (commented):
A couple of docs-related suggestions, and an unrelated question.

jzaffiro (approved):
Approved for merge-tree!

experimental/dds/tree/src/SharedTree.ts
Josmithr: This sounds like useful context to document in the code 😉

packages/utils/telemetry-utils/src/api.ts
```diff
@@ -0,0 +1,44 @@
...
+import { createChildLogger as createChildLoggerInternal } from "./logger.js";
+import type { ITelemetryLoggerExt } from "./telemetryTypes.js";
+
+/**
+ * Create a child logger based on the provided props object.
+ *
+ * @remarks
+ * Passing in no props object (i.e. undefined) will return a logger that is effectively a no-op.
+ *
+ * @param props - logger is the base logger the child will log to after it's processing, namespace will be prefixed to all event names, properties are default properties that will be applied events.
```
Josmithr, line 33: nit: Rather than using an `@param` block here with entries for each inner property, you could just add TSDoc comments to the inner properties. I think the intellisense comes out better that way.

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -24,15 +24,44 @@ import {
...
+/**
+ * Type erase a {@link TelemetryLoggerExt} to an {@link ITelemetryLoggerExt}.
+ * @internal
+ */
+export function toITelemetryLoggerExt(logger: TelemetryLoggerExt): ITelemetryLoggerExt {
+	return logger as unknown as ITelemetryLoggerExt;
+}
+
+/**
+ * Un-type-erase the {@link ITelemetryLoggerExt}.
```
Josmithr, line 49: nit: I think this needs more documentation explaining why this pattern is needed, etc.

packages/utils/telemetry-utils/package.json
```diff
@@ -16,47 +16,48 @@
 		".": {
```
Josmithr, line 16: I think we should either add notes describing the infra pattern deviations we've made in this package and why. Probably a good excuse to add a `DEV.md` file to the package.

Josmithr: And (obviously not for this PR), I think we should add similar documentation to our other packages that have weird deviations from our standard patterns.

packages/utils/telemetry-utils/src/telemetryTypes.ts
markfields: I think it's a ghost copy of `TelemetryLoggerExt` which is the renamed now-internal version of this.  But yeah I agree that the wording is a bit confusing as-is

packages/drivers/odsp-driver/src/epochTracker.ts
```diff
@@ -90,6 +94,9 @@ export const Odsp409Error = "Odsp409Error";
  * then it also clears all the cached entries for the given container.
  * @legacy
  * @beta
+ *
+ * @privateRemarks
+ * This class should be hidden and an interface exposed to better manage internal types like telemetry logger.
```
markfields, line 99: I have a branch that does this.  Maybe I'll get it done in time for 2.110 after I'm back

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -24,15 +24,44 @@ import {
...
+		PossiblyUndefined?: true;
+	} = // eslint-disable-next-line @typescript-eslint/no-empty-object-type
+	{},
+>(
+	input:
+		| ITelemetryLoggerExt
+		| TelemetryLoggerExt
+		| (options["PossiblyUndefined"] extends true ? undefined : never),
+): TelemetryLoggerExt | (options["PossiblyUndefined"] extends true ? undefined : never) {
+	return input as unknown as TelemetryLoggerExt;
```
markfields, line 74: Maybe toss a comment here saying "if PossiblyUndefined is true, then this will also pass-through an undefined value"

markfields:
Thanks for sharing, thanks for doing this, it'll be great to have the freedom to iterate on the API we use for instrumentation!  Meanwhile we should work on reducing usages of this type shim to the point we can remove it.

MarioJGMsoft (approved):

.changeset/lovely-groups-relate.md
jason-ha: Actually, the guidance is to not do that duplication. The issue should be the best place to look, and it can be revised if things change. We can duplicate guidance when it is simple and we are very confident.

I wouldn't say I am super confident in the guidance for these. So really want readers to hit the issue.

packages/utils/telemetry-utils/src/telemetryTypes.ts
jason-ha: It is for both inside and out. External folks that will stitch layers together may have been exposed to this.
I will try to make better comments.

packages/utils/telemetry-utils/src/api.ts
jason-ha: That block is an exact duplicate of comment in `logger.ts`. Here I am preserving content as it was. I would inheritDoc except that rest needs to different.

packages/utils/telemetry-utils/package.json
jason-ha: 1. index.ts/js is meant for use when there is a single common entrypoint. That is not the case here as we have main.ts that has all of the common stuff, but api.ts and internal.ts are different. I think the unique names help make that clear.
2. Unfortunately, we have some tooling that makes assumptions about index.ts beyond a default. It is best to just avoid it if we aren't following the very precise pattern.

packages/utils/telemetry-utils/package.json
jason-ha: The Node10 support in `generate entrypoints` is limited. (One of the places that assumes /internal maps to index.ts.)
We will be moving all to `generate node10Entrypoints` in the future as we will abandon `generate entrypoints` altogether and use .ts entrypoint generation that has some advantages. `tree` has a crude form of it to work around a tsc bug.

packages/utils/telemetry-utils/src/logger.ts
jason-ha: TS definitely won't remove these. But packers/minimizers, at least webpack, can and does remove these.
There are a lot of type guards that do nothing when using TypeScript. I have to imagine runtimes are good at handling no-op pass-thru routines.
Also I don't think this will be used in high volume callstacks. Typically will be at boundaries where things are getting glued together.

packages/utils/telemetry-utils/DEV.md
```diff
@@ -0,0 +1,31 @@
+# Internal entrypoint source structure for `@fluidframework/telemetry-utils`
+
+Note: index.ts is avoided as the external and internal entrypoints are reasonably different.
```
Josmithr: Nit: we should elaborate on _why_ they're different. Why we decided we needed to use a non-standard pattern here.

packages/utils/telemetry-utils/DEV.md
```diff
@@ -0,0 +1,31 @@
+# Internal entrypoint source structure for `@fluidframework/telemetry-utils`
```
Josmithr, line 1: Thanks for the docs!

packages/utils/telemetry-utils/DEV.md
Josmithr: Also would be good to clarify the point at which we could consider going back to the standard pattern.

Josmithr (approved):

Josmithr (approved):

alexvy86 (approved):
Seems good to me.

Abe27342 (approved):

PR #26918, opened 2026-04-01
feat(tree): add SchemaFactoryAlpha.stagedOptional for incremental required→optional field migration

noencke:
## Summary

- Adds `SchemaFactoryAlpha.stagedOptional(T)` to enable incremental migration of required fields to optional fields without a coordinated deployment
- During the staged rollout period, the field is Optional in the view schema (TypeScript type `T | undefined`) but the stored schema stays Required, so old clients are unaffected and writing `undefined` is blocked at runtime
- Mirrors the existing `SchemaStaticsBeta.staged(T)` pattern for allowed types, but for field optionality

Migration path: `sf.required(T)` → `sf.stagedOptional(T)` → `sf.optional(T)`


packages/dds/tree/src/simple-tree/fieldSchema.ts
```diff
@@ -412,6 +421,10 @@ export class FieldSchemaAlpha<
 		return this.propsAlpha?.persistedMetadata;
 	}
 
+	public get isOptionalStaged(): false | SchemaUpgrade {
```
noencke: Is this more idiomatic than e.g. `get optionalStagedUpgrade(): SchemaUpgrade | undefined`? Is the way it's currently written analogous to how staged allowed types do it? If so, we should keep it this way for consistency.

I guess it also interacts better with the new `isOptionalStaged` on SimpleSchema as well.

packages/dds/tree/src/simple-tree/toStoredSchema.ts
```diff
@@ -433,9 +440,27 @@ function filterFieldAllowedTypes(
...
+ * Returns the field kind to use in the stored schema for the given view field schema.
+ * @remarks
+ * For {@link SchemaFactoryAlpha.stagedOptional | staged optional} fields, the stored field kind is Required
+ * when the staged optional is not being included (i.e., restrictive options).
+ */
+function getStoredFieldKind(
+	f: SimpleFieldSchema,
+	options: SimpleSchemaTransformationOptions,
+): FieldKind {
+	if (!isStoredFromView(options)) return f.kind;
```
noencke: We should use braces {} even for one-line conditionals

packages/dds/tree/src/simple-tree/api/discrepancies.ts
```diff
@@ -362,8 +362,16 @@ function* getFieldDiscrepancies(
 		convertFieldKind.get(view.kind) ??
 		fail(0xbef /* A conversion from a FieldKind to a FlexFieldKind should exist */);
 
+	// A staged optional field (Optional in the view) is compatible with a Required stored field
+	// during the rollout period — suppress the kind discrepancy in that case.
+	const isOptionalStaged = view.isOptionalStaged;
```
noencke: Ehh, don't bother pulling this out.

packages/dds/tree/src/simple-tree/api/storedSchema.ts
```diff
@@ -53,7 +53,7 @@ export function extractPersistedSchema(
 	minVersionForCollab: MinimumVersionForCollab,
 	includeStaged: (upgrade: SchemaUpgrade) => boolean,
 ): JsonCompatible {
-	const stored = toStoredSchema(schema, { includeStaged });
+	const stored = toStoredSchema(schema, { includeStaged, includeStagedOptional: () => false });
```
noencke: Why is this just `() => false` here? Do we want to hook this up to `includeStaged`? Seems like staged optional field should be able to follow the same flow as a staged type, since it's also a `SchemaUpgrade`?

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1236,6 +1243,7 @@ export interface SimpleArrayNodeSchema<Type extends SchemaType = SchemaType, out
 
 // @alpha @sealed
 export interface SimpleFieldSchema<Type extends SchemaType = SchemaType> {
+    readonly isOptionalStaged?: Type extends SchemaType.Stored ? undefined : false | SchemaUpgrade;
```
jenn-le: nit: it's throwing me off a little that this is OptionalStaged but everything else is stagedOptional

packages/dds/tree/src/simple-tree/api/discrepancies.ts
```diff
@@ -363,7 +363,12 @@ function* getFieldDiscrepancies(
 		fail(0xbef /* A conversion from a FieldKind to a FlexFieldKind should exist */);
 
 	// This checks if the field kind in the view schema is not compatible with the stored schema.
-	if (viewKind.identifier !== stored.kind) {
+	// A staged optional field (Optional in the view) is compatible with a Required stored field
```
jenn-le: nit: are these supposed to be specifically capitalized? it feels kinda off

packages/dds/tree/src/test/simple-tree/api/stagedSchemaUpgrade.spec.ts
jenn-le:
```suggestion
describe("staged allowed type upgrade", () => {
```

packages/dds/tree/src/simple-tree/simpleSchema.ts
```diff
@@ -286,6 +286,22 @@ export interface SimpleFieldSchema<Type extends SchemaType = SchemaType> {
...
 	 */
 	readonly persistedMetadata?: JsonCompatibleReadOnlyObject | undefined;
+
+	/**
+	 * {@link SchemaUpgrade} if this field is included as a {@link SchemaStaticsAlpha.stagedOptional | staged optional} field,
+	 * allowing the view schema to be compatible with stored schema where the field is required (pre-upgrade) or optional (post-upgrade).
+	 * New documents and schema upgrades will keep the field as required in the stored schema.
+	 *
+	 * `false` if this is a view schema field that is not staged optional.
+	 * Absent (`undefined`) if derived from a stored schema (where staged optional has no meaning).
```
jenn-le, line 296: Why have both false and undefined as options rather than just undefined?

noencke: Good question. Updated the comment. It's consistent with how staged allowed types do it. But there's even a comment on that one that it's kind of weird. But we should be consistent, and if we want to change them later, change them together.

packages/dds/tree/api-report/tree.alpha.api.md
noencke: reversed

packages/dds/tree/src/simple-tree/api/discrepancies.ts
noencke: true

packages/dds/tree/src/simple-tree/simpleSchema.ts
jenn-le: Oh I see, I was thinking this was supposed to be analogous to `AllowedTypeMetadata.stagedSchemaUpgrade`. Since it's not, agreed we should be consistent.

jenn-le (approved):

.changeset/fruity-cars-tell.md
```diff
@@ -0,0 +1,18 @@
+---
+"@fluidframework/tree": minor
+"__section": feature
+---
+
+Add `SchemaFactoryAlpha.stagedOptional` for incremental required→optional field migrations
```
Josmithr: Changeset *titles* should not include `inlineCodeSyntax` per [our guidelines](https://github.com/microsoft/FluidFramework/blob/main/.changeset/README.md#formatting). Everywhere else is good (and recommended when appropriate).

Josmithr:
```suggestion
Add SchemaFactoryAlpha.stagedOptional for incremental required→optional field migrations
```

.changeset/fruity-cars-tell.md
```diff
@@ -0,0 +1,18 @@
...
+
+`SchemaFactoryAlpha.stagedOptional(T)` creates a field that is optional in the view schema but
+stored as required in the stored schema until all clients have been upgraded. This enables
+rolling out an optional field without coordinating a simultaneous deployment across all clients.
```
Josmithr: This is about migrating a field from *required* to *optional*, right? I don't think the way this is currently stated makes that very clear.

.changeset/fruity-cars-tell.md
```diff
@@ -0,0 +1,18 @@
...
+
+Migration path:
+1. Start with `sf.required(T)` — all clients require the field.
+2. Deploy `sf.stagedOptional(T)` — new clients see the field as optional and can read documents
+   whether the field is present or absent, but the stored schema stays required so old clients
+   are not broken. Writing `undefined` is blocked at runtime during this phase.
```
Josmithr, line 17: I assume there's no reasonable way for us to make it this type-safe? I.e., reading the value returns `T | undefined`, but writing the value forbids `undefined`?

packages/dds/tree/src/simple-tree/api/schemaFactoryAlpha.ts
```diff
@@ -143,6 +144,42 @@ export interface SchemaStaticsAlpha {
...
+	 * Creates a field schema that is Optional in the view but behaves as Required in the stored schema
+	 * during the rollout period of an optional-field migration.
+	 *
+	 * @remarks
+	 * Use this to incrementally migrate a required field to optional without a coordinated deployment.
+	 * The migration path is:
+	 *
+	 * 1. Start with `sf.required(T)` — all clients require the field.
+	 * 2. Deploy `sf.stagedOptional(T)` — new clients can read documents where the field is present or absent, but the stored schema stays Required (so old clients are unaffected). Writing `undefined` is blocked at runtime.
+	 * 3. Deploy `sf.optional(T)` once all clients support the optional field — the stored schema becomes Optional and the field can be cleared.
```
Josmithr: Nit
```suggestion
	 * 3. Deploy `sf.optional(T)` once all clients support the staged optional field — the stored schema becomes Optional and the field can be cleared.
```

packages/dds/tree/src/simple-tree/api/schemaFactoryAlpha.ts
```diff
@@ -143,6 +144,42 @@ export interface SchemaStaticsAlpha {
 		TCustomMetadata,
 		FieldPropsAlpha<TCustomMetadata> & { defaultProvider: DefaultProvider }
 	>;
+	/**
```
Josmithr, line 147: Great docs! Thank you!

packages/dds/tree/src/simple-tree/fieldSchema.ts
```diff
@@ -182,6 +183,14 @@ export interface FieldPropsAlpha<TCustomMetadata = unknown>
 	 * Sets {@link SimpleFieldSchema.persistedMetadata}.
 	 */
 	readonly persistedMetadata?: JsonCompatibleReadOnlyObject | undefined;
+
+	/**
+	 * If set, marks this field as a {@link SchemaStaticsAlpha.stagedOptional | staged optional} field.
+	 * @remarks
+	 * Do not set directly; use {@link SchemaStaticsAlpha.stagedOptional} instead.
```
Josmithr: Is there not a way for us to hide this property from the public-facing type? E.g. add an internal `FieldProps` type that unions the public type with a bucket of internal-only properties?

packages/dds/tree/src/simple-tree/simpleSchema.ts
```diff
@@ -286,6 +286,26 @@ export interface SimpleFieldSchema<Type extends SchemaType = SchemaType> {
 	 * Set via {@link FieldPropsAlpha.persistedMetadata}.
 	 */
 	readonly persistedMetadata?: JsonCompatibleReadOnlyObject | undefined;
+
+	/**
+	 * {@link SchemaUpgrade} if this field is included as a {@link SchemaStaticsAlpha.stagedOptional | staged optional} field,
```
Josmithr: Nit: this reads as an action to me. I.e. "Performs a schema upgrade if...", but that doesn't make sense with the boolean typing.

Would it make more sense to express this as something like "Indicates that..."?

packages/dds/tree/src/simple-tree/fieldSchema.ts
noencke: Maybe, I'm not sure. I specifically tried to make this PR mirror the same patterns that we already employ for staged allowed types, since they are similar mechanisms. Staged allowed types have a `stagedSchemaUpgrade` property on a visible beta type as well. Maybe we could hide them, but we should do both at once.

What I did for now is actually make it more visible - it's technically fine for a user to make this object manually (it's just that there is a helper for it for convenience). So I removed the warning.

.changeset/fruity-cars-tell.md
noencke: That's correct, unfortunately. Same problem we always have with compile time checking of fields - we use TS mapped types to generate the API and mapped types can't have different get and set types.

.changeset/fruity-cars-tell.md
noencke: Updated, thanks.

jenn-le (approved):

jenn-le (approved):

.changeset/fruity-cars-tell.md
Josmithr: Womp womp. Makes sense.

Josmithr (approved):

PR #26924, opened 2026-04-01
Improve ci-readiness-check skill: build:docs check, phantom diff guidance, post-check change warning

noencke:
## Summary

Improvements to the `ci-readiness-check` skill based on real usage against `@fluidframework/tree` changes.

### New: TaskCreate-based step tracking to prevent skipped steps

The skill now instructs the agent to create tasks (via `TaskCreate`) for every applicable step immediately after the user picks a mode — before doing any other work. Each task is marked `in_progress` when started and `completed` when done. This prevents steps from being silently dropped as context window fills up during a long check. The task list is mode-aware: Quick mode gets 3 tasks, Full/Thorough get the full set, and the tree cascade task is only added when `@fluidframework/tree` is among the changed packages.

### New: `build:docs` step to catch `ae-unresolved-link` errors

`build:api-reports` inherits a config that suppresses `ae-unresolved-link`, but CI catches these via `build:docs` which uses the unfiltered `api-extractor.json`. Added an explicit step 6b to run `build:docs` after API reports, with guidance on fixing ambiguous `{@link}` tags (the TSDoc `:instance`/`:static` selectors are unsupported in this version of API Extractor).

### New: `generate:entrypoint-sources` guidance for tree — extracted to `tree-api-checks.md`

`@fluidframework/tree` uses committed files in `src/entrypoints/` that explicitly list every named export by API tier. Run `generate:entrypoint-sources` any time `src/index.ts` is touched — this covers new exports, removals, and tier changes (e.g. promoting from `@alpha` to `@public`). The script clobbers `lib/entrypoints/` with wrong import paths, requiring a `build:esnext` follow-up.

### Refactored: tree-specific guidance extracted to `tree-api-checks.md`

All three tree-specific sections (`generate:entrypoint-sources`, phantom key-reorder, cascade to `fluid-framework`/`azure-client`) moved to a new sub-document. The main `SKILL.md` now has a single prominent trigger at the top of step 6: "If `@fluidframework/tree` is among the changed packages, read `tree-api-checks.md` now." This keeps the main skill clean for non-tree work while ensuring the complex tree-specific logic isn't missed when it matters.

### Corrected: phantom key-reorder guidance

Initial guidance said "restore the file, don't commit it" — which is exactly backwards. Corrected: always commit what `build:api-reports` produces. Local fresh builds and CI agree on the same ordering; restoring the old order causes CI to fail.

### ESLint step ordering

ESLint (step 5) runs before API report regeneration (step 6) so that any auto-fixes affecting exported signatures are in place before API Extractor runs. Added a note to re-run relevant checks after any further code changes.

### DIFF
--- a/.claude/skills/ci-readiness-check/SKILL.md
+++ b/.claude/skills/ci-readiness-check/SKILL.md
@@ -24,9 +24,9 @@ Before doing anything, ask the user:
 > I'll run a CI readiness check on your branch. Pick a mode (fastest to slowest):
 >
 > 1. **Cancel** — never mind, don't run checks
-> 2. **Quick** — auto-fix formatting only (Biome on changed files); skip all build-dependent checks (~15–20 seconds)
-> 3. **Full** — Quick + build unbuilt packages + regenerate API reports and type tests (~30 seconds – 2 minutes depending on package size)
-> 4. **Thorough** — Full + run tests in changed packages (~30 seconds – 5+ minutes; dominated by test suite size)
+> 2. **Quick** — auto-fix formatting, policy, and syncpack on changed packages; skip build-dependent checks (~30–60 seconds)
+> 3. **Full** — Quick + build unbuilt packages + ESLint + regenerate API reports and type tests (~1–3 minutes depending on package size)
+> 4. **Thorough** — Full + run tests in changed packages (~5+ minutes; dominated by test suite size)
 
 Wait for the user's response. If they say cancel (or anything clearly negative), stop here. Otherwise, note their choice and **immediately create tasks for all remaining steps** as described in the `<required>` block above before proceeding.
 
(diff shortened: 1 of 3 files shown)


.claude/skills/ci-readiness-check/SKILL.md
```diff
@@ -75,9 +75,20 @@ If the script reports unbuilt packages, build them:
...
+# Step 5: ESLint auto-fix (Full and Thorough only)
+
+For each built changed package, check its `package.json` for the available lint fix script and run it:
+```bash
+# Check which script exists — packages use different names
+cd <package-dir> && node -p "Object.keys(require('./package.json').scripts || {}).filter(s => /eslint.fix|lint.fix/.test(s))"
+cd <package-dir> && pnpm run <script-name>
+```
+
+Common script names: `eslint:fix`, `lint:fix`. Only run what exists. If the script fails due to non-auto-fixable errors, note them but do not block — CI will catch those.
```
tylerbutler: I think we should use fluid-build for this. `pnpm exec fluid-build . -t lint:fix` If that doesn't run the right script for the package, then something's messed up. Bottom line we should try to use fluid-build's native understanding of the dependency graph and what scripts map to what rather than sending the agent to find it.

.claude/skills/ci-readiness-check/SKILL.md
```diff
@@ -102,7 +115,32 @@ cd <package-dir> && pnpm run build:api-reports
...
+
+```bash
+cd <package-dir> && pnpm run build:docs
+```
+
+If you see `ae-unresolved-link` errors, the `{@link}` or `{@inheritdoc}` tag references an ambiguous name (most commonly a member that exists as both a static and instance declaration). Fix by linking to an unambiguous target — for example, link to the interface that declares the member (which has exactly one declaration) rather than the class that exposes it as both static and instance. The TSDoc `:instance`/`:static` system selectors are **not** supported by this version of API Extractor.
+
+## 6c. Cross-package cascade
+
+After regenerating API reports for a package, check if any "aggregator" packages re-export from it. If so, their API reports are now stale too.
```
tylerbutler: Another place we might use fluid-build. But I guess you're trying to avoid running it here? It's a little weird since my guidance to people is to run `pnpm build` locally before opening a PR, relying on fluid-build's caching to make that fast most of the time. Any reason to make the agent do more complex gymnastics?

.claude/skills/ci-readiness-check/SKILL.md
noencke: Interesting, okay. So flub will dispatch `lint:fix` to the proper lint command (which might be e.g. `eslint:fix`)?

tylerbutler (approved):

.claude/skills/ci-readiness-check/SKILL.md
tylerbutler: Yes, it should.

tylerbutler (approved):
LGTM!

taylorsw04 (approved):

taylorsw04 (approved):

PR #26927, opened 2026-04-02
array node treeChanged delta payload and unhydrated node fix

daesunp:
## Description

Adds support for delta payload for array node events in `treeChanged`, and addresses the getting `undefined` deltas for unhydrated nodes.


packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -298,34 +331,148 @@ export const treeNodeApi: TreeNodeApi = {
...
+		// TreeNodes and are handled by the else branch below.
+		return isArrayNodeSchema(childSchema)
+			? { kind: "array", ops: [] }
+			: { kind: "object", changedProperties: new Map() };
+	} else {
+		if (fieldMark.fields !== undefined && fieldMark.detach === undefined) {
+			// A retain mark with nested `fields` implies a tree node was retained, not a leaf.
+			// Reaching here indicates an unexpected internal state.
+			fail("Expected a tree node for a retain mark with nested fields.");
+		}
```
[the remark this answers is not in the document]
daesunp: Updated to have `kind: "deleted"`

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -298,34 +331,148 @@ export const treeNodeApi: TreeNodeApi = {
...
+	} else if (isArrayNodeSchema(nodeSchema)) {
+		// Nested array: its sequence field changes are under EmptyKey.
+		const emptyKeyChanges =
+			fields.get(EmptyKey) ?? fail("Expected EmptyKey changes in array node delta.");
+		return { kind: "array", ops: deltaMarksToArrayOps(emptyKeyChanges.marks, node) };
+	} else {
+		// Map or record node: stored field keys are the property keys.
+		// Map entries are accessed via .get() — direct property access does not work on map node proxies.
+		assert(isMapNodeSchema(nodeSchema), "expected map or record schema in map branch");
+		const mapNode = node as unknown as TreeMapNode;
```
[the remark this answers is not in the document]
daesunp: Added section for recode nodes

packages/dds/tree/src/simple-tree/api/treeChangeEvents.ts
```diff
@@ -36,7 +36,10 @@ export interface TreeChangeEvents {
 	 *
 	 * - Object nodes define a change as being when the value of one of its properties changes (i.e., the property's value is set, including when set to `undefined`).
 	 *
-	 * - Array nodes define a change as when an element is added, removed, moved or replaced.
+	 * - Array nodes define a change as when an element is added, removed, moved, or replaced,
+	 * or when a property of an element changes (i.e., a nested change within an element).
+	 * In the latter case, the {@link ArrayNodeRetainOp.childDelta} field on the corresponding
+	 * retain op describes what changed within the element.
```
Josmithr: I don't think we want to report nested changes as a part of the `nodeChanged` event. Presumably we only want to do this for the `treeChanged` event to be consistent with other node types. `nodeChanged` is a shallow check, and `treeChanged` is a deep check.

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -857,6 +858,20 @@ export interface NodeChangedDataProperties<TNode extends TreeNode = TreeNode> {
     readonly changedProperties: ReadonlySet<TNode extends WithType<string, NodeKind.Object, infer TInfo> ? string & keyof TInfo : string>;
 }
 
+// @alpha @sealed
+export type NodeDelta = {
```
Josmithr: How does `treeChanged` represent deep tree changes? Seems like we would want to leverage the same tree structure (presumably just adding support for arrays).

packages/dds/tree/src/simple-tree/api/treeChangeEvents.ts
daesunp: Updated to be in `treeChanged` and now just provides a boolean instead of childDelta

packages/dds/tree/api-report/tree.alpha.api.md
daesunp: Moved to `treechanged` :)

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -131,7 +131,7 @@ export interface ArrayNodeRemoveOp {
 
 // @alpha @sealed
 export interface ArrayNodeRetainOp {
-    readonly childDelta?: NodeDelta;
+    readonly contentChanged?: true;
```
Josmithr: Nit: is there a reason we don't want to just use a boolean here? Would it be unreasonable to specify `false` (rather than `undefined`) to indicate no changes?

packages/dds/tree/src/shared-tree/treeAlpha.ts
```diff
@@ -810,6 +814,19 @@ export const TreeAlpha: TreeAlpha = {
 		eventName: K,
 		listener: NoInfer<TreeChangeEventsAlpha<TNode>[K]>,
 	): () => void {
+		// For array nodes, override treeChanged to fire via childrenChangedAfterBatch (which
```
Josmithr: Special-casing this here seems odd to me. Is there a reason we can't implement this in the actual `treeChanged` event implementation? I think we already surface an alpha/beta-only variant(s) of the events. Presumably we could enable this as alpha-only the same way.

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -70,22 +71,65 @@ export type NodeChangedDataAlpha<TNode extends TreeNode = TreeNode> =
...
  * @sealed @alpha
  */
 export interface TreeChangeEventsAlpha<TNode extends TreeNode = TreeNode>
 	extends TreeChangeEvents {
 	/**
-	 * Like `TreeChangeEventsBeta.nodeChanged`, but for array nodes the event data includes
-	 * a {@link NodeChangedDataDelta.delta | delta} payload describing the changes as a sequence
-	 * of {@link ArrayNodeDeltaOp} values.
+	 * Emitted when a direct structural change occurs on the node (an element was inserted,
+	 * removed, or moved). For array nodes the event data includes a
```
Josmithr: Nit: I think everything after the first sentence here should go into the `@remarks` block

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -70,22 +71,65 @@ export type NodeChangedDataAlpha<TNode extends TreeNode = TreeNode> =
...
+	 * Does not fire for pure nested-content changes (e.g. a property of an array element changed
+	 * without any structural array change). Subscribe to {@link TreeChangeEventsAlpha.treeChanged}
+	 * on the array to receive a delta for those cases as well.
+	 *
 	 * This defines a property which is a function instead of using the method syntax to avoid function bi-variance issues with the input data to the callback.
 	 */
 	nodeChanged: (data: NodeChangedDataAlpha<TNode>) => void;
+
+	/**
+	 * For array nodes: emitted when any change occurred within the array, including both
```
Josmithr: All of these notes really belong in the `@remarks` block. And we should add a brief, general description above it (like you have for `nodeChanged`).

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -40,50 +39,32 @@ import {
...
  */
 export interface ArrayNodeRetainOp {
 	readonly type: "retain";
 	readonly count: number;
 	/**
-	 * When present, describes what changed within the single retained element.
+	 * When `true`, one or more properties of this retained element changed (e.g. an object field
+	 * was updated, or a map/record entry was added, removed, or replaced).
+	 * When absent, the element itself was not modified — only its position in the array may have
+	 * changed due to surrounding inserts, removes, or moves.
```
Josmithr: I think you could re-express this more simply by saying something like "If there were any changes anywhere in the subtree rooted under the retained element." (This is more similar to the TreeChangeEvent docs terminology.

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -284,18 +265,27 @@ export const treeNodeApi: TreeNodeApi = {
 				} else if (isArrayNodeSchema(nodeSchema)) {
 					return kernel.events.on("childrenChangedAfterBatch", ({ fieldMarks }) => {
 						const marks = fieldMarks.get(EmptyKey);
+						// nodeChanged fires only for structural changes (insert, remove, move).
```
Josmithr: Nit: I'm not sure "structural change" is the right terminology here. Maybe "shallow change"?

packages/dds/tree/src/test/tableSchema.spec.ts
```diff
@@ -1642,21 +1642,24 @@ describe("TableFactory unit tests", () => {
 			});
 			assert.equal(eventCount, 1);
 
-			// Update column props
+			// Update column props.
```
Josmithr: Nit: do we need these changes?

Josmithr (commented):
Looking good! I left a few more minor comments.

One piece of high-level feedback: I think it would be worth aligning on consistent terminology for shallow (nodeChanged) vs deep (treeChanged) changes.

I don't think "structural" is the right term to use for shallow changes. I suspect we want to use "shallow".

When describing deeper changes, I think we've used both "deep" and "nested". Not sure which we prefer.

Might be a good parking lot discussion for stand-up :)

packages/dds/tree/api-report/tree.alpha.api.md
daesunp: My thought was that we only care about the retain ops that have `contentChanged: true`, and it would just look cleaner if it didn't show up at all for the other cases. That being said, I would be happy to update it to a boolean as well :)

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: Updated with suggestion

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: updated with suggestion

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: Simplified doc comment

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: Updated "structural" and "nested" to "shallow" and "deep"

packages/dds/tree/src/test/tableSchema.spec.ts
daesunp: No they are not, removed :)

packages/dds/tree/api-report/tree.alpha.api.md
Josmithr: I guess my take is that we're emulating `true | false` with `true | undefined`, which has slightly less clear semantics. If we were serializing this, I think I would feel differently (since we could save space in the common case).

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1733,6 +1734,7 @@ export interface TreeChangeEvents {
 // @alpha @sealed
 export interface TreeChangeEventsAlpha<TNode extends TreeNode = TreeNode> extends TreeChangeEvents {
     nodeChanged: (data: NodeChangedDataAlpha<TNode>) => void;
+    treeChanged: TNode extends WithType<string, NodeKind.Array> ? (data: NodeChangedDataDelta) => void : () => void;
```
Josmithr: `TreeChangeEventsBeta.treeChanged` supports returning some change details for non-array nodes. This change degrades that experience for everything but array nodes. We should make sure not to remove any functionality relative to the beta events. It would probably make sense to update `TreeChangeEventsAlpha` to extend `TreeChangeEventsBeta` to enforce this.

packages/dds/tree/src/core/tree/anchorSet.ts
```diff
@@ -983,6 +983,31 @@ export class AnchorSet implements AnchorLocator {
 							this.parentField,
 							marks,
 						);
+						// Queue childrenChangedAfterBatch for array sequence fields that have
+						// pure retain-with-fields marks (element property changes with no
```
Josmithr: These docs could use a terminology audit based on our discussion during stand-up this morning.

packages/dds/tree/src/shared-tree/treeAlpha.ts
```diff
@@ -810,6 +818,19 @@ export const TreeAlpha: TreeAlpha = {
...
```
Josmithr: (Sorry, I deleted this before, but I think my original concern here was valid after all).

Is there a reason we can't implement this directly in the implementation of `TreeChangeEventsAlpha.treeChange`?

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -35,26 +35,31 @@ export interface NodeChangedDataProperties<TNode extends TreeNode = TreeNode> {
...
-	 * - The document was updated in a way that required multiple internal change passes in a single
-	 * operation (for example, a data change combined with a schema upgrade).
+	 * The value may be `undefined` when the document was updated in a way that required multiple
+	 * internal change passes in a single operation (for example, a data change combined with a
+	 * schema upgrade).
+	 *
+	 * @remarks
+	 * Note: for unhydrated array nodes (not yet inserted into a document), deep changes
+	 * on elements do not fire `treeChanged` at all — not even with `delta: undefined`. Only direct
+	 * changes (insert, remove, move) fire the event on unhydrated nodes (tracked in AB#63261).
```
Josmithr: Nit: we should avoid including ADO item numbers in user-facing docs. We should extract the tracking note into a `@privateRemarks` block.

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -35,26 +35,31 @@ export interface NodeChangedDataProperties<TNode extends TreeNode = TreeNode> {
 }
 
 /**
- * Data included for {@link TreeChangeEventsAlpha.nodeChanged} when the node is an array node.
+ * Data carried by array-node events from {@link TreeChangeEventsAlpha}:
+ * both {@link TreeChangeEventsAlpha.nodeChanged} and {@link TreeChangeEventsAlpha.treeChanged}.
  * @sealed @alpha
  */
 export interface NodeChangedDataDelta {
```
Josmithr, line 44: Where all is this type used? Is it _only_ used in `treeChanged`? If so, I wonder if we should rename it to "TreeChangedDataDelta" instead.

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -69,22 +74,68 @@ export type NodeChangedDataAlpha<TNode extends TreeNode = TreeNode> =
...
  * Use via `TreeAlpha.on`.
  * @sealed @alpha
  */
 export interface TreeChangeEventsAlpha<TNode extends TreeNode = TreeNode>
 	extends TreeChangeEvents {
 	/**
-	 * Like `TreeChangeEventsBeta.nodeChanged`, but for array nodes the event data includes
-	 * a {@link NodeChangedDataDelta.delta | delta} payload describing the changes as a sequence
-	 * of {@link ArrayNodeDeltaOp} values.
+	 * Emitted when a shallow change occurs on this node.
```
Josmithr: Nit: since this is user-facing, I think we should probably elaborate on what we mean by "shallow". Then subsequent notes can just reference "shallow" without elaborating. Something like:
```suggestion
	 * Emitted when a shallow change occurs on this node. I.e. when the node's own set of properties changes.
```

packages/dds/tree/api-report/tree.alpha.api.md
Josmithr: On that note: we should really be able to compose these. Since the `treeChanged` events for non-array nodes actually report specific details about their deep changes, we should be able to include that information in the array node deltas for `treeChanged` events.

It probably makes sense to do that in a separate PR, since this is only alpha-level.

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -69,22 +74,68 @@ export type NodeChangedDataAlpha<TNode extends TreeNode = TreeNode> =
...
+	 * element nodes.
+	 *
+	 * This event only fires on the array node whose elements changed — it does not propagate to
+	 * ancestor nodes. To receive a delta for a nested array, subscribe directly to that array.
+	 * Ancestor nodes still receive the base (no-payload) `treeChanged` via the normal subtree
+	 * propagation.
+	 *
+	 * For unhydrated array nodes (not yet inserted into a document), only direct changes
+	 * (insert, remove, move) fire this event. Deep changes on elements of unhydrated
+	 * arrays are not detected (tracked in AB#63261).
```
Josmithr: Let's move the tracking detail into a `@privateRemarks` block to avoid polluting our public/user-facing docs.

packages/dds/tree/api-report/tree.alpha.api.md
daesunp: I tried to extend it, but unfortunately ran into an issue where `NodeChangedData<TNode>` (beta) needs to be assignable to `NodeChangedDataAlpha<TNode>`, and fails for array nodes because beta's data is `changedProperties?: ReadonlySet<string>` and alpha is `delta: ArrayNodeDeltaOp[]`.

Should we update alpha's NodeChangedData to have changedProperties (but empty) to satisfy the typing issue? 

Or can we reference the beta type directly in alpha instead? Something like:
```
treeChanged: TNode extends WithType<string, NodeKind.Array> 
? (data: NodeChangedDataDelta) => void 
: TreeChangeEventsBeta<TNode>["treeChanged"]
```

regarding the composition, I agree and would be happy to address it in a follow up :)

packages/dds/tree/api-report/tree.alpha.api.md
daesunp: Makes sense! Updated to boolean

packages/dds/tree/src/shared-tree/treeAlpha.ts
daesunp: It was easy to do for the `nodeChanged` case because both the stable and alpha fire on the node directly, so we can subscribe to the node's own kernel and pass the data along.

For `treeChanged` the stable version uses `subtreeChangedAfterBatch` which propagates up through all the ancestors (with no payload). The delta comes from `childrenChangedAfterBatch` which we special cased for array nodes.

I tried moving this into `treeNodeApi.on`, but it broke the unhydrated array node case, because `childrenChangedAfterBatch` does not propagate up for unhydrated nodes (which may be an issue for the stable version, but this limitation is documented for alpha).

I was originally planning to address it in the follow up pr for fixing the unhydrated case, but currently seeing if there is an easy fix for this that I can just include for this PR

daesunp: Fixed unhydrated case, and moved to treeNodeApi.ts

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: removed ADO after fix for unhydrated case

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: The type is also used in `nodeChanged` so I thought it may be misleading if we named it `TreeChangedDataDelta`

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: Updated to suggested change

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: removed after fix

packages/dds/tree/api-report/tree.alpha.api.md
Josmithr: Yeah, I think the pattern you suggested there looks good!

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -69,22 +69,64 @@ export type NodeChangedDataAlpha<TNode extends TreeNode = TreeNode> =
...
 /**
- * Extension of {@link TreeChangeEvents} with a richer `nodeChanged` event.
+ * Extension of {@link TreeChangeEvents} with a richer `nodeChanged` event and a
+ * delta-carrying `treeChanged` event for array nodes.
  * @remarks
  * Provides a `nodeChanged` event that includes a delta payload for array nodes and
  * requires `changedProperties` for object, map, and record nodes.
+ * Also provides a `treeChanged` event that, for array nodes, carries a {@link NodeChangedDataDelta}
+ * payload describing both shallow and deep changes.
+ * For non-array nodes the `treeChanged` signature is the same as the base event.
```
Josmithr: Nit
```suggestion
 * For non-array nodes, the `treeChanged` signature is the same as the base event.
```

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -69,22 +69,64 @@ export type NodeChangedDataAlpha<TNode extends TreeNode = TreeNode> =
...
 	extends TreeChangeEvents {
 	/**
-	 * Like `TreeChangeEventsBeta.nodeChanged`, but for array nodes the event data includes
-	 * a {@link NodeChangedDataDelta.delta | delta} payload describing the changes as a sequence
-	 * of {@link ArrayNodeDeltaOp} values.
+	 * Emitted when a shallow change occurs on this node, i.e., when the node's direct children change.
 	 *
 	 * @remarks
+	 * For array nodes: the event data includes a {@link NodeChangedDataDelta.delta | delta} payload
+	 * as a sequence of {@link ArrayNodeDeltaOp} values; retain ops with
```
Josmithr: Nit
```suggestion
	 * as a sequence of {@link ArrayNodeDeltaOp} values. Retain ops with
```

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -69,22 +69,64 @@ export type NodeChangedDataAlpha<TNode extends TreeNode = TreeNode> =
...
+	 * changed). The event data carries a {@link NodeChangedDataDelta.delta | delta} payload
+	 * describing what changed. The delta uses {@link ArrayNodeRetainOp.contentChanged} to flag
+	 * elements that have deep changes, without describing the details of those deep changes.
+	 * To inspect deep changes, subscribe to `nodeChanged` or `treeChanged` on the individual
+	 * element nodes.
+	 *
+	 * This event only fires on the array node whose elements changed — it does not propagate to
+	 * ancestor nodes. To receive a delta for a nested array, subscribe directly to that array.
+	 * Ancestor nodes still receive the base (no-payload) `treeChanged` via the normal subtree
+	 * propagation.
```
Josmithr: This is true only for now, right? We plan on composing events in a future change such that this will no longer be the case? Would probably be good to clarify that this is just a current limitation.

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -69,22 +69,64 @@ export type NodeChangedDataAlpha<TNode extends TreeNode = TreeNode> =
...
+	 *
+	 * For non-array nodes: same as the base {@link TreeChangeEvents.treeChanged}.
+	 *
+	 * The listener type is conditional on `TNode`. If `TNode` is the base {@link TreeNode} type
+	 * (i.e. the node's schema is not known statically), the listener is typed as `() => void`
+	 * and no delta payload is provided. Use a more specific schema type to get the delta.
+	 *
+	 * This defines a property which is a function instead of using the method syntax to avoid function bi-variance issues with the input data to the callback.
```
Josmithr, line 148: Nit: this feels like it probably belongs in a `@privateRemarks` block (I don't think end-users need this information).

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
```diff
@@ -69,22 +69,64 @@ export type NodeChangedDataAlpha<TNode extends TreeNode = TreeNode> =
...
+	 * This event only fires on the array node whose elements changed — it does not propagate to
+	 * ancestor nodes. To receive a delta for a nested array, subscribe directly to that array.
+	 * Ancestor nodes still receive the base (no-payload) `treeChanged` via the normal subtree
+	 * propagation.
+	 *
+	 * For non-array nodes: same as the base {@link TreeChangeEvents.treeChanged}.
+	 *
+	 * The listener type is conditional on `TNode`. If `TNode` is the base {@link TreeNode} type
+	 * (i.e. the node's schema is not known statically), the listener is typed as `() => void`
+	 * and no delta payload is provided. Use a more specific schema type to get the delta.
```
Josmithr: This will be obsolete when you restore the beta-level non-array node handling we've discussed elsewhere. Just leaving this comment as a reminder to clean up this comment when we fix that :)

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -41,12 +41,23 @@ import type { TreeChangeEvents } from "./treeChangeEvents.js";
 
 /**
  * A `"retain"` op in an {@link ArrayNodeDeltaOp} sequence.
- * Represents elements that were not added or removed (though they may have nested changes).
+ * Represents elements that were neither inserted nor removed from the array.
```
Josmithr: Nit (sorry, I can be pedantic when it comes to documentation 😅)
```suggestion
 * Represents elements that were neither inserted into nor removed from the array.
```

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -41,12 +41,23 @@ import type { TreeChangeEvents } from "./treeChangeEvents.js";
...
 	readonly count: number;
+	/**
+	 * Whether there are any changes in the subtree rooted at this retained element.
+	 * `true` if the element's subtree changed; `false` if nothing changed within it.
+	 * @remarks
+	 * Subscribe to `nodeChanged` or `treeChanged` on the element node itself for details.
+	 * This field appears in deltas from both {@link TreeChangeEventsAlpha.nodeChanged} and
+	 * {@link TreeChangeEventsAlpha.treeChanged}. It is not present in
+	 * {@link TreeChangeEventsBeta.nodeChanged}, which provides
+	 * {@link NodeChangedData.changedProperties} for non-array nodes instead.
```
Josmithr: Nit: I don't know that these details are necessary. The docs on the event interfaces give the necessary context here.

packages/dds/tree/src/test/simple-tree/api/treeNodeApi.spec.ts
```diff
@@ -2850,25 +2851,316 @@ describe("treeNodeApi", () => {
...
 				assert.deepEqual(deltas, [
 					[
-						{ type: "retain", count: 1 },
+						{ type: "retain", count: 1, contentChanged: true },
 						{ type: "remove", count: 1 },
 					],
 				]);
 			});
+
+			describe(`unhydrated array — nodeChanged delta`, () => {
```
Josmithr: Can we use `describeHydration` so we can use the same tests to evaluate both the hydrated and unhydrated cases automatically? We expect the behavior to be the same now right?

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
```diff
@@ -41,12 +41,23 @@ import type { TreeChangeEvents } from "./treeChangeEvents.js";
...
+	 */
+	readonly contentChanged: boolean;
```
Josmithr: Sorry to keep complicating things for you. Feel free to hold off on addressing this feedback until a future PR (though we may want to capture some of this in documentation at least).

1. For the `nodeChanged` event, this flag doesn't make sense. We don't track deep changes for `nodeChanged`, so neither `true` nor `false` is really accurate to say. We just don't know one way or another in that case.
    For the `treeChanged` event, this makes perfect sense as is. We know whether or not there were any changes beneath this level in the tree.
    - This dichotomy makes me think we may want separate types. The `nodeChanged` variant would simply not have this property.

2. Once we are only applying this to `treeChanged` events, it would probably rename this to be a bit more explicit. Something like `subtreeChanged` or `containsSubtreeChanges`.

daesunp: Removed `contentChanged` from nodeChanged event, and updated name to `subtreeChanged`

packages/dds/tree/src/test/simple-tree/api/treeNodeApi.spec.ts
daesunp: Updated to `describeHydration`

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: Updated to suggestion

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: Updated to suggestion

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: moved to privateRemarks

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: Updated with suggestion :)

packages/dds/tree/src/simple-tree/api/treeNodeApi.ts
daesunp: Updated with `contentChanged` refactor

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: The documentation was a bit inaccurate. Updated to be more clear

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1734,7 +1756,7 @@ export interface TreeChangeEvents {
 // @alpha @sealed
 export interface TreeChangeEventsAlpha<TNode extends TreeNode = TreeNode> extends TreeChangeEvents {
```
Josmithr, line 1757: Nit: Can this extend `TreeChangeEventsBeta` now, or does that still cause problems?

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
Josmithr: @daesunp I think that comment section can be removed now, right?

packages/framework/presence/api-report/presence.alpha.api.md
```diff
@@ -395,8 +395,8 @@ export interface RawValueAccessor<T> {
 
 // @public
 export const StateFactory: {
-    readonly latest: LatestFactory_2;
-    readonly latestMap: LatestMapFactory_2;
+    readonly latest: LatestFactory;
```
Josmithr: These presence API report changes should be reverted.

packages/dds/tree/api-report/tree.alpha.api.md
Josmithr: Synced via Teams chat. This is more complicated than it seems. We'll worry about this later.

packages/dds/tree/src/simple-tree/api/treeAlpha.ts
daesunp: removed comment section

packages/framework/presence/api-report/presence.alpha.api.md
daesunp: Reverted presence api changes

Josmithr (approved):

PR #26930, opened 2026-04-02
Fixup TypeTest configs

CraigMacomber:
## Description

Type test configuration was denromalized such that `pnpm exec flub typetests -g client --normalize` was not a no-op, and tryp tests updates after releases would cause some issues.

This gets everything back into a normalized state, adding missing test scripts to id-compressor as part of that.

## Reviewer Guidance


packages/runtime/id-compressor/package.json
```diff
@@ -161,7 +163,11 @@
 		"typescript": "~5.4.5"
 	},
 	"typeValidation": {
-		"broken": {},
+		"broken": {
+			"Interface_IIdCompressorCore": {
+				"forwardCompat": false
```
CraigMacomber, line 168: Looks like there was an API break for IdCompressorCore last release, which is now detected.

taylorsw04 (approved):

packages/runtime/id-compressor/package.json
jason-ha: That package went without type test for 2.5 years while giving the impression that it did have coverage all that time.

PR #26933, opened 2026-04-02
Bump client patch 2.92.1 and update type tests

CraigMacomber:
## Description

Bump patch version and update type tests.

Includes a cherry-pick of 19a716b6250d945acb514205d6f452bd524b4a57 to mitigate issues with type test updates.

## Reviewer Guidance


packages/framework/oldest-client-observer/package.json
```diff
@@ -104,5 +104,8 @@
 		"jiti": "^2.6.1",
 		"rimraf": "^6.1.3",
 		"typescript": "~5.4.5"
+	},
+	"typeValidation": {
+		"disabled": true
```
TommyBrosman, line 109: Is there a plan to reenable type validation? I looked at #26930 but didn't see any mention of whether it is a permanent change.

packages/tools/fetch-tool/package.json
```diff
@@ -50,7 +50,6 @@
 	"devDependencies": {
 		"@biomejs/biome": "~2.4.5",
 		"@fluid-tools/build-cli": "catalog:buildTools",
-		"@fluid-tools/fetch-tool-previous": "npm:@fluid-tools/fetch-tool@2.91.0",
```
TommyBrosman, line 53: Was this removed because type validation was turned off? (Why is it different from the other packages?)

TommyBrosman (changes requested):

packages/framework/oldest-client-observer/package.json
CraigMacomber: This package lacks the scripts to generate type tests, so if we do enable it here, it does nothing but add an unused dev dependency.

The change here is to fix the fact that this package.json is denormalized, so updating the type test version (even to the same version it already was) makes changes here, which don't even work due to the missing script. This did not have type testing before thig change: this simply adjust the configuration to inform our tools of that.

If someone thinks this package should have type testing, they can fix it (on main) by adding the proper script, removing this disable, selecting which entry point they want to test and rerunning the type test preparation.

TommyBrosman: Makes sense. This is a case where I really wish JSON supported comments.

packages/tools/fetch-tool/package.json
CraigMacomber: This package had no type tests, due to lacking the typetest:gen script. Having it claim to have type tests when it does not, and maintain an unused dependency on a previous version is misleading.

Also having type tests against internal APIs is a real strange thing to do. As since this is a console tool, I suspect type tests don't actually enforce CLI interface API stability.

TommyBrosman (approved):

frankmueller-msft (approved):

jzaffiro (approved):

taylorsw04 (approved):

PR #26934, opened 2026-04-03
chore: Upgrade development and CI to Node 22

ChumpChief:
[AB#68708](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/68708)
[AB#68709](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/68709)
[AB#68710](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/68710)

## Description

Node 20 is approaching EOL. This upgrades development tooling, CI pipelines, and type
definitions from Node 20 to Node 22 (target runtime version: 22.22.2).

Changes:
- `.nvmrc` updated to Node 22
- ADO pipeline template default updated to 22.22.2
- GitHub Actions `website-validation` workflow updated to Node 22
- `engines` fields updated to `>=22.22.2` across 5 package.json files
- `@types/node` standardized to `~22.19.17` across all workspaces
- Added a `types` pnpm catalog for `@types/node` in the client workspace (128 packages)
- All lockfiles regenerated

This does **not** update the server Dockerfiles — that will follow in a separate PR after
this change has soaked in CI.

## Breaking Changes

Developers should update to Node 22.22.2 or newer. This is enforced via `engines` fields
in the package.json. See [Breaking-vs-Non-breaking-Changes](https://github.com/microsoft/FluidFramework/wiki/Breaking-vs-Non-breaking-Changes).

## Reviewer Guidance

The review process is outlined on [this wiki page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

- The bulk of the diff is mechanical `@types/node` version bumps and lockfile regeneration.
- The `types` pnpm catalog is a new pattern — client workspace packages now use
  `"@types/node": "catalog:types"` so future version bumps only require updating
  `pnpm-workspace.yaml`. Independent workspaces (build-tools, server, docs, tools)
  retain explicit versions since catalogs don't span workspaces.
- Node 22.22.2 must be baked into ADO CI build agent images before this PR can merge.
- Watch for OOM issues in end-to-end tests (this occurred during the Node 20 upgrade).

PR #26935, opened 2026-04-03
fix(gitrest): Update code for @types/node@22 compatibility

ChumpChief:
[AB#71216](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/71216)

## Description

Updates gitrest code to be compatible with `@types/node@22`, which introduces two breaking type changes:

- **`fs.Stats` constructor is now private**: `fileSystemHelper.ts` was calling `new fs.Stats()` to construct a Stats object. Replaced with a plain object literal containing the required properties and `is*()` methods, cast to `fs.Stats`.

- **`fsPromises.readdir` gained new overloads**: `redisFsManager.ts` had manually-written overload signatures for `readdirCore` that no longer matched the base class (`Parameters<typeof fsPromises.readdir>`). Replaced the manual overloads with the same `Parameters`/`ReturnType` pass-through pattern used by `filesystems.ts`.

## Reviewer Guidance


server/gitrest/packages/gitrest-base/src/utils/redisFs/redisFsManager.ts
```diff
@@ -225,33 +225,9 @@ export class RedisFs extends FsPromisesBase {
...
-		options: ObjectEncodingOptions & { withFileTypes: true },
-	): Promise<Dirent[]>;
-	public async readdirCore(
-		folderpath: PathLike,
-		options?: any,
-	): Promise<string[] | Buffer[] | Dirent[]> {
+		...args: Parameters<typeof fsPromises.readdir>
+	): ReturnType<typeof fsPromises.readdir> {
+		const [folderpath] = args;
 		const folderpathString = folderpath.toString();
```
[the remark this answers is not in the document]
ChumpChief: Preexisting, and not necessary for the isomorphic-git contract per comment above.

server/gitrest/packages/gitrest-base/src/utils/fileSystemHelper.ts
```diff
@@ -149,31 +149,46 @@ export function filepathToString(filepath: PathLike | FileHandle): string {
...
+		size: size ?? 1,
+		uid: 0,
+		gid: 0,
+		ino: 0,
+		dev: 0,
+		mtime,
+		mtimeMs,
+		ctime: mtime,
+		ctimeMs: mtimeMs,
+		isFile: () => type === "file" || type === undefined,
```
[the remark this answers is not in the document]
ChumpChief: Updated in latest

server/gitrest/packages/gitrest-base/src/utils/redisFs/redisFsManager.ts
```diff
@@ -264,7 +240,10 @@ export class RedisFs extends FsPromisesBase {
 			},
 		);
 
-		return result;
+		// Cast through unknown because keysByPrefix returns string[], which doesn't
+		// overlap with Dirent<Buffer>[] (a return type variant added in @types/node@22).
+		// This is safe because isomorphic-git always expects string[] results from readdir.
+		return result as unknown as Awaited<ReturnType<typeof fsPromises.readdir>>;
```
jason-ha: Can you put this in a type-guard helper? (and keep comments)
```typescript
function safeCastNameHere(in: FromTypeExpected): ToTypeOkay {
   return in as unknown as ...;
}
```
sidenote: why does this `await` and not just return? (helper signature could let Promise be part of type)

jason-ha: It could be safer to name the type rather than match whatever fsPromises.readdir changes to. Right now there is some judgement that result is okay to be reinterpreted as XYZ. Name that XYZ so that if there is a change in the future, we can make sure the cast is still okay.

ChumpChief: Can def drop the await - updated in latest.

Re: "`safeCastNameHere`": We can't truthfully type check here, as this is in fact not a "safe" cast per the interface.  It's only safe to lie in this way in the context of the knowledge that isomorphic-git will only ever call expecting a `string[]` return.  So `fsPromises.readdir` changing isn't really a threat so much as isomorphic-git's calling patterns changing (which is runtime-only knowledge, not something we can detect in the type system).

Prior to this change, we effectively do the unsafe cast by declaring method overloads that we in fact do not correctly support (e.g. not returning `Buffer[]` even when `encoding: "buffer"` is passed).  I think this explicit cast is a little easier to spot and understand the danger as compared to the untruthful overloads?

Re: having a helper function wrapping the unsafe cast: I think I prefer to keep the cast inline here to (1) keep the cast/comment close to the context in which it can be understood (2) discourage reuse of such a helper and (3) avoid implying that real type checking might be happening somewhere else.

That said, I think it is somewhat of a matter of preference.  I'd be interested in @znewton and @pradeepvairamani 's opinions in particular since they will likely be the ones maintaining this :)

ChumpChief: Addressed in other comment - the threat is isomorphic-git rather than readdir changes.

znewton (approved):
:shipit:

PR #26936, opened 2026-04-03
fix(mocha-test-setup): Remove legacy module path resolution

ChumpChief:
[AB#71217](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/71217)

## Description

Removes path resolution logic in `mocharc-common.cjs` that converted bare package specifiers (e.g. `@fluid-internal/mocha-test-setup`) to filesystem directory paths (e.g. `/path/to/node_modules/@fluid-internal/mocha-test-setup`).

This was a workaround for Lerna's hoisting behavior, where dependencies could be placed in a parent `node_modules` folder and not be resolvable by mocha. Under pnpm this is unnecessary since all dependencies appear in the package's own `node_modules` (symlinked to the actual location).

Additionally, resolving to directory paths breaks ESM imports in Node 22+, which no longer supports directory imports (`ERR_UNSUPPORTED_DIR_IMPORT`). This was causing test failures in `import-testing`, `app-insights-logger`, and `container-runtime`.

## Reviewer Guidance

- The same directory-to-bare-specifier change is applied to the `FLUID_TEST_LOGGER_PKG_SPECIFIER` code path as well, along with removing the now-unused `existsSync` import and `moduleDir` variable.


packages/test/mocha-test-setup/mocharc-common.cjs
```diff
@@ -30,34 +27,9 @@ function getFluidTestMochaConfig(packageDir, additionalRequiredModules, testRepo
...
-		return mod;
-	});
-
 	if (process.env.FLUID_TEST_LOGGER_PKG_SPECIFIER) {
-		const modulePath = path.join(moduleDir, process.env.FLUID_TEST_LOGGER_PKG_SPECIFIER);
 		// Inject implementation of createTestLogger, put it first before mocha-test-setup
-		if (existsSync(modulePath)) {
-			requiredModulePaths.unshift(modulePath);
-		}
+		requiredModules.unshift(process.env.FLUID_TEST_LOGGER_PKG_SPECIFIER);
```
alexvy86, line 32: Can we double check that this injection still worked as expected? I don't recall if we should expect to see test telemetry from runs of the build pipeline, but the test pipelines definitely send it, so maybe we want to do manual runs of build + those on a test/ branch.

alexvy86 (commented):
I agree with the general idea, especially if it blocks the move to Node22. Left one comment below though, and want to point out [this bit of an open PR](https://github.com/microsoft/FluidFramework/pull/26926/changes#diff-4c15f94828a94ba3c0176187dfe5835b02505268fdd7be919fb10dc91217d0b6R306-R310) just because it relates to hoisting behavior... although I don't think it matters since that's just for the legacy version installs, and we're not running tests that end up using `mocharc-common.cjs` from inside those.

packages/test/mocha-test-setup/mocharc-common.cjs
ChumpChief: Did a check - produced client build 389713, then real service e2e tests 389741 (using pipeline artifacts from 389713), and I see telemetry coming in for that run.

alexvy86 (approved):

PR #26938, opened 2026-04-03
fix(container-runtime): Fix navigator assignment in Hardware Stats tests

ChumpChief:
[AB#71218](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/71218)

## Description

The `getDeviceSpec()` function reads `navigator.deviceMemory` and `navigator.hardwareConcurrency` to report hardware info via telemetry when a container loads. The Hardware Stats test mocks `navigator` to simulate three states: a normal object with values, `null`, and `undefined`.

On Node 20, `navigator` wasn't a built-in global, so direct assignment worked. Node 22 added a built-in `navigator` (for web platform alignment) exposed as a read-only getter, so direct assignment now throws `TypeError: Cannot set property navigator of #<Object> which has only a getter`.

Fix: use `Object.defineProperty` to override the property instead of direct assignment. This is the standard way to replace a getter-backed property in JavaScript, so this is the correct approach going forward, not a temporary workaround. The `restoreNavigator` cleanup in `afterEach` ensures the override doesn't leak into other test suites.

[AB#68707](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/68707) tracks several potential simplifications once we've upgraded to Node 22:

- The `delete` branch in `restoreNavigator` can be removed since Node 22 always provides a descriptor to restore.
- The `null`/`undefined` test cases could be dropped since no supported environment (Node 22+ or browsers) has a missing navigator.
- We'd still need the mock to test the browser-like case where `deviceMemory` is present (Node's navigator doesn't have it). We could also add an unmocked test verifying behavior with Node 22's real navigator (`hardwareConcurrency` set, `deviceMemory` undefined), covering both real environments without testing impossible states.

## Reviewer Guidance

### DIFF
--- a/packages/runtime/container-runtime/src/test/hardwareStats.spec.ts
+++ b/packages/runtime/container-runtime/src/test/hardwareStats.spec.ts
@@ -34,7 +34,7 @@ function setNavigator(
 
 function restoreNavigator() {
 	if (originalNavigatorDescriptor === undefined) {
-		// eslint-disable-next-line @typescript-eslint/no-dynamic-delete -- this branch can be removed after upgrading to Node 22, which always provides a built-in navigator descriptor
+		// eslint-disable-next-line @typescript-eslint/no-dynamic-delete -- this branch can be removed after upgrading to Node 22, which always provides a built-in navigator descriptor (AB#68707)
 		delete (globalThis as Record<string, unknown>).navigator;
 	} else {
 		Object.defineProperty(globalThis, "navigator", originalNavigatorDescriptor);
--- a/packages/runtime/container-runtime/src/test/hardwareStats.spec.ts
+++ b/packages/runtime/container-runtime/src/test/hardwareStats.spec.ts
@@ -17,6 +17,8 @@ import {
 import { ContainerRuntime, getDeviceSpec } from "../containerRuntime.js";
 import { FluidDataStoreRegistry } from "../dataStoreRegistry.js";
 
+const originalNavigatorDescriptor = Object.getOwnPropertyDescriptor(globalThis, "navigator");
+
 function setNavigator(
 	// eslint-disable-next-line @rushstack/no-new-null -- testing behavior with global
 	navigator: Partial<Navigator & { deviceMemory?: number }> | undefined | null,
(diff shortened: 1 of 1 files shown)


packages/runtime/container-runtime/src/test/hardwareStats.spec.ts
```diff
@@ -21,7 +21,13 @@ function setNavigator(
...
 	navigator: Partial<Navigator & { deviceMemory?: number }> | undefined | null,
 ) {
-	global.navigator = navigator as Navigator;
+	// In Node 22+, globalThis.navigator is a read-only getter, so direct
+	// assignment throws. Use Object.defineProperty to override it.
+	Object.defineProperty(globalThis, "navigator", {
+		value: navigator,
+		writable: true,
+		configurable: true,
+	});
```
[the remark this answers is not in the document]
ChumpChief: Updated in latest

packages/runtime/container-runtime/src/test/hardwareStats.spec.ts
```diff
@@ -17,11 +17,28 @@ import {
...
+}
+
+function restoreNavigator() {
+	if (originalNavigatorDescriptor === undefined) {
+		// eslint-disable-next-line @typescript-eslint/no-dynamic-delete -- this branch can be removed after upgrading to Node 22, which always provides a built-in navigator descriptor
```
jason-ha: Got a work item?

jason-ha (approved):

shlevari (commented):
This code seems right for doing what you're doing, but, the fix feels a little weird to me. Like it just seems odd to add code to get around the fact that you can't set something because there is only a getter. Could you add more description to the PR on why this is necessary and the right fix and if there is any plan to remove it once the update happens?

packages/runtime/container-runtime/src/test/hardwareStats.spec.ts
ChumpChief: Just opened AB#68707 and added it to the inline comment

ChumpChief:
@shlevari Good question — here's some more context.

The `getDeviceSpec()` function reads `navigator.deviceMemory` and `navigator.hardwareConcurrency` to report hardware info via telemetry when a container loads. The test mocks `navigator` to simulate three states: a normal object with values, `null`, and `undefined`.

On Node 20, `navigator` wasn't a built-in global, so direct assignment worked. Node 22 added a built-in `navigator` (for web platform alignment) exposed as a read-only getter. `Object.defineProperty` is the standard way to replace a getter-backed property, so this is the correct approach going forward, not a temporary workaround. The `restoreNavigator` cleanup in `afterEach` ensures the override doesn't leak into other test suites.

AB#68707 tracks several potential simplifications once we've upgraded to Node 22:

- The `delete` branch in `restoreNavigator` can be removed since Node 22 always provides a descriptor to restore.
- The `null`/`undefined` test cases could be dropped since no supported environment (Node 22+ or browsers) has a missing navigator.
- We'd still need the mock to test the browser-like case where `deviceMemory` is present (Node's navigator doesn't have it). We could also add an unmocked test verifying behavior with Node 22's real navigator (`hardwareConcurrency` set, `deviceMemory` undefined), covering both real environments without testing impossible states.

jason-ha (approved):

PR #26939, opened 2026-04-03
Revert "feat: add fluid-controlled map interfaces"

alexvy86:
Reverts microsoft/FluidFramework#26788 because I suspect it broke the Loop integration pipeline.


jason-ha (approved):
That change is confirmed to have caused the break in copilot-board.

CraigMacomber (approved):

PR #26940, opened 2026-04-04
fix: Disposal in fluid statics container now disposes instead of closes

taylorsw04:
## Description

This PR fixes a bug where the dispose function in the fluid-static FluidContainer called close on the container it wraps rather than dispose (note: dispose calls close). This would result in memory leaks because summarizer timers and any other top-level closures would retain transitive content.

It also ensures that the dispose event fires for both underlying close and dispose events, as the public API makes no distinction between them.

### DIFF
--- a/packages/framework/fluid-static/src/fluidContainer.ts
+++ b/packages/framework/fluid-static/src/fluidContainer.ts
@@ -390,7 +390,7 @@ class FluidContainer<TContainerSchema extends ContainerSchema = ContainerSchema>
 	}
 
 	public dispose(): void {
-		this.container.close();
+		this.container.dispose();
 		this.container.off("connected", this.connectedHandler);
 		this.container.off("closed", this.disposedHandler);
 		this.container.off("disconnected", this.disconnectedHandler);
--- a/packages/framework/fluid-static/src/fluidContainer.ts
+++ b/packages/framework/fluid-static/src/fluidContainer.ts
@@ -98,11 +98,11 @@ export interface IFluidContainerEvents extends IEvent {
 	(event: "dirty", listener: () => void): void;
 
 	/**
-	 * Emitted when the {@link IFluidContainer} is closed, which permanently disables it.
+	 * Emitted when the {@link IFluidContainer} is disposed, which permanently disables it.
 	 *
 	 * @remarks Listener parameters:
 	 *
-	 * - `error`: If the container was closed due to error (as opposed to an explicit call to
+	 * - `error`: If the container was disposed due to error (as opposed to an explicit call to
 	 * {@link IFluidContainer.dispose}), this will contain details about the error that caused it.
 	 */
 	(event: "disposed", listener: (error?: ICriticalContainerError) => void);
@@ -138,7 +138,7 @@ export interface IFluidContainer<TContainerSchema extends ContainerSchema = Cont
 	 * A container is considered dirty in the following cases:
 	 *
 	 * 1. The container has been created in the detached state, and either it has not been attached yet or it is
-	 * in the process of being attached (container is in `attaching` state). If container is closed prior to being
+	 * in the process of being attached (container is in `attaching` state). If container is disposed prior to being
 	 * attached, host may never know if the file was created or not.
 	 *
 	 * 2. The container was attached, but it has local changes that have not yet been saved to service endpoint.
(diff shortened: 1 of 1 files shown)


packages/framework/fluid-static/src/fluidContainer.ts
```diff
@@ -331,6 +331,7 @@ class FluidContainer<TContainerSchema extends ContainerSchema = ContainerSchema>
 		this.acquireExtension = extensionStore.acquireExtension.bind(extensionStore);
 		container.on("connected", this.connectedHandler);
 		container.on("closed", this.disposedHandler);
```
agarwal-navin, line 333: Interesting that there is no "close" event / API on fluid container. I wonder why that's the case.
@ChumpChief Do you know?

packages/framework/fluid-static/src/fluidContainer.ts
```diff
@@ -390,9 +391,10 @@ class FluidContainer<TContainerSchema extends ContainerSchema = ContainerSchema>
 	}
 
 	public dispose(): void {
-		this.container.close();
+		this.container.dispose();
```
ChumpChief, line 394: Disposal has different characteristics than close(), which could technically make this a breaking change depending on what the customer is doing post-close today.  Have you considered whether this is a concern and/or mitigations or customer guidance if so?

packages/framework/fluid-static/src/fluidContainer.ts
```diff
@@ -390,9 +391,10 @@ class FluidContainer<TContainerSchema extends ContainerSchema = ContainerSchema>
...
 		this.container.off("connected", this.connectedHandler);
 		this.container.off("closed", this.disposedHandler);
+		this.container.off("disposed", this.disposedHandler);
```
ChumpChief, line 397: Emitting the same single event for both "closed" and "disposed" may not align well with the underlying system - depending on whether the underlying container closed/disposed the user will observe inconsistent results to further interaction with the FluidContainer, e.g. subsequent calls to retrieve data objects fail or not.  Would be good for the event to give a deterministic signal about container state and capabilities.

packages/framework/fluid-static/src/fluidContainer.ts
ChumpChief: I believe all of this predates any intentionality about the term "close" vs. "dispose", at the time I think we just used both inconsistently throughout the repo.

ChumpChief (commented):
Although I'm not a huge fan of the dual closed/disposed states, I think we probably need to either:
1. Mirror the underlying Container and expose both here (along with documentation, customer guidance, etc.).
2. Treat it like a name alignment - everything becomes "disposed", stop saying "closed" anywhere.

Either way, changing "disposed" to actually mean "disposed" instead of "closed" is a technically-breaking change, and so the safest option would be to align with a major release.  But I think an argument could be made that it's not a real threat in practice?  Probably raise it with API council to get more diverse opinions.

packages/framework/fluid-static/src/fluidContainer.ts
taylorsw04: @ChumpChief is there any documentation anywhere describing the difference? If not, could you explain it? AFAICT, customers of this API have no knowledge of "close" as a concept, although once I understand the difference we can perhaps figure out if they will be impacted by this change.

packages/framework/fluid-static/src/fluidContainer.ts
taylorsw04: Ya, I think I'd like to move to _only_ registering/emitting for disposed, but I wasn't sure how closed works so I'll wait until I do.

packages/framework/fluid-static/src/fluidContainer.ts
taylorsw04: Depending on how observable this is, customers may prefer us to fix the memory leak over preserving the existing behavior?

ChumpChief: I believe @kian-thompson did the work to define the difference and might have some good pointers.  My basic understanding is that on close, the container remains readable, just not writable.  Whereas on dispose, we start to throw for various read operations.

I have no pushback for making one of the two changes I list above here eventually, my point is more about "when" such a change is safe outside of a major release.

kian-thompson: I imagine to preserve previous behavior, you could call both `close` and `dispose`

ChumpChief: Since we don't dispose today, wouldn't this start to cause errors thrown if e.g. handles are deref'd post-dispose?

kian-thompson: From when I made this change, old `close` = new `close` + `dispose`. Not sure what new issues would arise, but the behavior should be preserved

PR #26945, opened 2026-04-06
chore(build-cli): deprecate release report commands

tylerbutler:
## Summary

The release report commands were never intended to be used for anything other than reporting after a release. They have long been used for other purposes. Now, I want to delete them so I don't have to maintain them or code review them any more. There should be no need for these commands in 2026. They were created as a "fun addition" to the build tools in what was perhaps my greatest mistake in creating the build-tools. If we are relying on them, then we need to stop.

## Changes

- Marks `flub release report` and `flub release report-unreleased` as deprecated using oclif's built-in deprecation mechanism
- Adds a prominent warning banner (12 lines of "DO NOT RELY ON THIS COMMAND") printed on every invocation
- Release reports will no longer be generated in the future and these commands will be removed

## Test plan

- [x] Run `flub release report --help` and verify it shows deprecated status
- [x] Run `flub release report` and verify the warning banner is printed
- [x] Run `flub release report-unreleased --help` and verify it shows deprecated status

### DIFF
--- a/build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
+++ b/build-tools/packages/build-cli/src/commands/release/report-unreleased.ts
@@ -53,18 +53,18 @@ export class UnreleasedReportCommand extends BaseCommand<typeof UnreleasedReport
 	};
 
 	public async run(): Promise<void> {
+		const { flags } = this;
+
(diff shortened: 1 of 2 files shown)


tylerbutler:
Closing since these reports are apprently still needed.

PR #26948, opened 2026-04-06
Fixed `--parentProcess` mode partial match

CraigMacomber:
## Description

Fixed `--parentProcess` mode incorrectly matching tests whose full title is a substring of another test's full title. The child process filter now uses an exact-match regex (`--grep ^title$`) instead of substring matching (`--fgrep title`).

This bug was detected by the improved validation of output added in 0.53 which reported errors due to this bug when run on preexisting tests which previously silently produced incorrect data ( In this case "findOverlappingIntervals on string of length 25000 with 200 equally spaced intervals and 25000 segments" vs "findOverlappingIntervals on string of length 25000 with 200 equally spaced intervals and 25000 segments with endpoint resolution"

Also refactors argument handling code for more testability.

## Reviewer Guidance

### DIFF
--- a/tools/benchmark/CHANGELOG.md
+++ b/tools/benchmark/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.54.0
 
+-   Fixed `--parentProcess` mode incorrectly matching tests whose full title is a substring of another test's full title. The child process filter now uses an exact-match regex (`--grep ^title$`) instead of substring matching (`--fgrep title`).
+
 ## 0.53.0
 
 -   Memory benchmarks now have much more aggressive GC to collect more stable results.
--- a/tools/benchmark/CHANGELOG.md
+++ b/tools/benchmark/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.54.0
 
--   Fixed `--parentProcess` mode incorrectly matching tests whose full title is a substring of another test's full title. The child process filter now uses an exact-match regex (`--grep ^title$`) instead of substring matching (`--fgrep title`).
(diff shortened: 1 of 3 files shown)


tools/benchmark/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 0.54.0
 
+-   Fixed `--parentProcess` mode incorrectly matching tests whose full title is a substring of another test's full title. The child process filter now uses an exact-match regex (`--grep ^title$`) instead of substring matching (`--fgrep title`).
```
Josmithr: Nit: an example might be nice (but I doubt too many folks are reading these, so probably not critical).

tools/benchmark/src/mocha/runner.ts
```diff
@@ -62,8 +61,13 @@ async function runTestInChildProcess(testFullTitle: string): Promise<CollectedDa
 		}
 	}
 
-	// Add test filter so child process only run the current test.
-	childArgs.push("--fgrep", testFullTitle);
+	// Add test filter so child process only runs the current test.
+	// Use --grep with an anchored regex for an exact match, since --fgrep does substring matching
+	// and would incorrectly match tests whose full title contains another test's full title.
+	// Escape any special regex characters in the test title to avoid changing the meaning of the regex.
+	// TODO: once only supporting NodeJS 24+, we can use the new RegExp.escape function here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape
```
Josmithr, line 76: Nice!

tools/benchmark/src/mocha/runner.ts
```diff
@@ -73,6 +77,27 @@ async function runTestInChildProcess(testFullTitle: string): Promise<CollectedDa
...
+
+/**
+ * Runs the specified test in a child process and returns the results.
+ * @remarks
+ * The provided test must write a {@link BenchmarkResult} to stdout.
+ * See {@link recordTestResult} which does this for child processes.
+ */
+async function runTestInChildProcess(testFullTitle: string): Promise<CollectedData> {
+	// Instead of running the benchmark in this process, create a new process.
+	// See {@link isParentProcess} for why.
```
Josmithr: Nit: `{@link }` syntax doesn't have meaning in `//` comments. Not necessarily harmful, but probably unecessary.

CraigMacomber: These docs were out of date anyway, and that was a dead link. I have moved updated text to the doc comment instead.

Josmithr (approved):

PR #26951, opened 2026-04-06
feat(tree): reapply "feat: add fluid-controlled map interfaces"

jenn-le:
This reapplies #27688 with a fix to the `FluidIterableIterator` interface that was causing downstream call sites to infer the type with undefined.


packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -0,0 +1,132 @@
...
+	[Symbol.iterator](): FluidIterableIterator<T>;
+}
+
+/**
+ * Like TypeScript's built-in iterable iterator type, except unaffected by TypeScript's version and configuration options.
+ *
+ * @privateRemarks
+ * The done branch uses `any` rather than `undefined` to match TypeScript's built-in IteratorReturnResult, which defaults to any for the same reason.
+ * Using `undefined` causes TypeScript to include it in type inference at call sites like `Array.from` which would infer `T | undefined` instead of `T`.
+ * The done value is never meaningful to callers since values are only consumed when done is false, so `any` is both correct and safe here.
```
CraigMacomber: Personally, I don't think this is "safe" and it can be meaningful.

if I have an iterator, it lets me do:

```typescript
while (true) {
const x = iterator.next();
if (x.done) {
 // x.value has type any here!
myNode.requiredFieldFoo = x.value; // Put undefined into a required field!
x.value.iThoughtThisThisExistedButItDoesNotAndTheCompilerWontHelpMe()
}
}
```

In my opinion this is all unsafe hacks due to TypeScript's built in types doing a poor job of capturing the iterator protocol (mainly defaulting the return type to "any", which was a bad choice.

```suggestion
 * The done value is rarely meaningful to callers since values are mainly consumed when done is false.
 * Thus `any` is not particularly harmful here, and the unsafety is worth it as
 * it interacts better with the existing TypeScript `IteratorResult` interfaces where `TReturn` defaults to any.
```

packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -0,0 +1,134 @@
...
+ * Like TypeScript's built-in `Iterable` type, except unaffected by TypeScript's version and configuration options.
+ *
+ * @sealed @alpha
+ */
+export interface FluidIterable<T> {
+	[Symbol.iterator](): FluidIterableIterator<T>;
+}
+
+/**
+ * Like TypeScript's built-in iterable iterator type, except unaffected by TypeScript's version and configuration options.
```
noencke:
```suggestion
 * Like TypeScript's built-in iterable iterator type.
```
I don't think we need to justify to customers why we have this - it should go in private remarks instead

noencke (approved):

noencke (approved):

taylorsw04 (approved):

jzaffiro (approved):

PR #26952, opened 2026-04-06
improvement(client-core-interfaces): remove unused impls

jason-ha:
`ErasedType` is never to be instantiated nor checked for `instanceof`. Thus, don't need implementations. Remove those implementations.

### DIFF
--- a/packages/common/core-interfaces/src/erasedType.ts
+++ b/packages/common/core-interfaces/src/erasedType.ts
@@ -92,8 +92,8 @@ export declare abstract class ErasedType<out Name = unknown> {
  * Implement interfaces which extend this by sub-classing {@link ErasedTypeImplementation}.
  *
  * This class should only be a `type` package export, preventing users from extending it directly.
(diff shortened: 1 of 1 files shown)


packages/common/core-interfaces/src/erasedType.ts
```diff
@@ -95,6 +92,8 @@ export abstract class ErasedType<out Name = unknown> {
  * Implement interfaces which extend this by sub-classing {@link ErasedTypeImplementation}.
  *
  * This class should only be a `type` package export, preventing users from extending it directly.
+ * But since {@link ErasedTypeImplementation} does extend it an implementation
+ * for ctor must be provided, unlike {@link ErasedType}.
```
CraigMacomber:
```suggestion
 * for `constructor` must be provided, unlike {@link ErasedType}.
```

See https://github.com/microsoft/FluidFramework/wiki/Coding-Guidelines#-avoid-using-abbreviations

CraigMacomber (approved):

PR #26955, opened 2026-04-07
Opt into chunked forest for text

brrichards:
# Summary 

Based of findings found while testing text memory and performance with `objectForest` vs `chunkedForest` it shows that plain text, but especially formatted, would benefit from using chunked forest. This PR enables `chunkedForest` for text and the `text-editor` app. Until the memory regression for plain text is fixed/improved this PR will remain in draft

Test findings are found on this [draft PR](https://github.com/microsoft/FluidFramework/pull/26950)

### DIFF
--- a/examples/data-objects/text-editor/src/app.tsx
+++ b/examples/data-objects/text-editor/src/app.tsx
@@ -26,10 +26,17 @@ import {
  */
 // eslint-disable-next-line import-x/no-internal-modules
 import { InsecureTokenProvider } from "@fluidframework/test-runtime-utils/internal";
-import { SchemaFactory, TreeViewConfiguration, type TreeView } from "@fluidframework/tree";
+import { SchemaFactory, type TreeView } from "@fluidframework/tree";
+import {
+	configuredSharedTreeAlpha,
+	ForestTypeOptimized,
+	TreeCompressionStrategy,
+	incrementalEncodingPolicyForAllowedTypes,
+	TreeViewConfigurationAlpha,
+	FluidClientVersion,
+} from "@fluidframework/tree/alpha";
 // eslint-disable-next-line import-x/no-internal-modules
 import { FormattedTextAsTree, TextAsTree } from "@fluidframework/tree/internal";
-import { SharedTree } from "@fluidframework/tree/legacy";
 import type { IFluidContainer } from "fluid-framework";
 // eslint-disable-next-line import-x/no-internal-modules, import-x/no-unassigned-import
 import "quill/dist/quill.snow.css";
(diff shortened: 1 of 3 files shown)


packages/dds/tree/src/text/textDomain.ts
```diff
@@ -24,7 +25,12 @@ const sf = new SchemaFactoryAlpha("com.fluidframework.text");
 
 class TextNode
 	extends sf.object("Text", {
-		content: SchemaFactory.required([() => StringArray], { key: EmptyKey }),
+		content: sf.required(
+			sf.types([{ type: () => StringArray, metadata: {} }], {
+				custom: { [incrementalSummaryHint]: true },
+			}),
+			{ key: EmptyKey },
```
Josmithr, line 32: Nit: I realize this was already here, but it would probably be a good idea to leave a quick comment here explaining why we're setting the persisted key as "" here.

Josmithr: (same for the schema in `textDomainFormatted`)

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -24,7 +25,12 @@ const sf = new SchemaFactoryAlpha("com.fluidframework.text");
 
 class TextNode
 	extends sf.object("Text", {
-		content: SchemaFactory.required([() => StringArray], { key: EmptyKey }),
+		content: sf.required(
+			sf.types([{ type: () => StringArray, metadata: {} }], {
+				custom: { [incrementalSummaryHint]: true },
```
Josmithr, line 30: Nit: Let's document what this is doing. (Same with the schema in `textDomainFormatted`)

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -57,20 +64,27 @@ function getTinyliciousEndpoint(): string {
...
 }) {}
 
-export const treeConfig = new TreeViewConfiguration({ schema: TextEditorRoot });
+export const treeConfig = new TreeViewConfigurationAlpha({ schema: TextEditorRoot });
+
+const SharedTree = configuredSharedTreeAlpha({
+	forest: ForestTypeOptimized,
+	treeEncodeType: TreeCompressionStrategy.CompressedIncremental,
+	shouldEncodeIncrementally: incrementalEncodingPolicyForAllowedTypes(treeConfig),
+	minVersionForCollab: FluidClientVersion.v2_74,
```
Josmithr, line 80: Nit: Let's leave a comment explaining why we are setting this min version.

brrichards:
Going to keep this PR in draft until the memory regression for plain text is fixed/improved

PR #26959, opened 2026-04-07
Forward ODSP endpoint to token fetch and make token manager injectable 

jzaffiro:
Pass the ODSP environment variable to the token fetch methods to ensure that the credentials received are for the correct endpoint. In addition, update the OdspTestDriver's token manager to allow callers to inject a token manager if they do not want to use the provided file cache token manager. The stress tests will inject a memory-cache-based token manager to avoid file lock bottlenecks when resolving credentials. 

[AB#65073](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/65073)


packages/test/test-drivers/src/odspTestDriver.ts
```diff
@@ -265,22 +273,35 @@ export function getOdspCredentials(
...
+
 /**
  * @internal
  */
 export class OdspTestDriver implements ITestDriver {
-	// Share the tokens and driverId across multiple instance of the test driver
-	private static readonly odspTokenManager = new OdspTokenManager(odspTokensCache);
+	// Share the driveId across multiple instances of the test driver.
 	private static readonly driveIdPCache = new Map<string, Promise<string>>();
+
```
jzaffiro:
```suggestion
```

Abe27342 (approved):

Abe27342 (approved):

PR #26960, opened 2026-04-07
feat(tree-agent): support staged allowed types in schema type generation

jenn-le:
## Description

 Adds support for staged allowed types in the tree-agent's TypeScript schema rendering. Staged types are types that can be read from a node but cannot yet be written —
  they represent a forward-compatible schema upgrade that has not been fully committed.

  Previously, the type generator treated all allowed types uniformly for reads and writes. With this change:

   - Array and map nodes with staged types render as TreeArray<TRead, TWrite> / TreeMap<TRead, TWrite>, where TRead includes staged types and TWrite excludes them.
   - Object node fields with staged types render as getter/setter pairs (get foo(): string | number; set foo(value: string);) instead of simple properties, enforcing
  write-time restrictions.
   - Record nodes include a warning comment noting which types are staged and not writeable.
   - The TreeArray and TreeMap prompt documentation is updated to reflect the dual type parameter pattern.

## Reviewer Guidance


packages/framework/tree-agent/src/renderSchemaTypeScript.ts
```diff
@@ -49,6 +54,7 @@ interface BindingIntersectionResult {
 }
 
 interface TypeExpression {
+	staged?: boolean;
```
noencke: IMO just make this required (true or false). It simplifies the space state and will clean up some of the code when reading and writing. No need for the "convenience" of an optional property when it's a module-private internal type, not an "API".

packages/framework/tree-agent/src/renderSchemaTypeScript.ts
```diff
@@ -49,6 +54,7 @@ interface BindingIntersectionResult {
 }
 
 interface TypeExpression {
```
noencke, line 56: A TypeExpression can be a type, or it can be a union of types, correct? In the latter case, what does `staged` mean? _Every_ type is staged? Or _any_ type is staged? I could deduce by reading the code below but it'd be good if it were documented here.

noencke (approved):

noencke (approved):

PR #26963, opened 2026-04-07
Make typetest normalize apply to the gen scripts

CraigMacomber:
## Description

Make typetest normalize apply to the gen scripts.
This should help mitigate confusion/risk around type tests appring enabled, but doing nothing due to missing gen scripts.

At least every release this will get normalized and cleaned up.
Mopre enforcement of consistancy is possible, but not included here.

## Reviewer Guidance

### DIFF
--- a/build-tools/packages/build-cli/src/commands/typetests.ts
+++ b/build-tools/packages/build-cli/src/commands/typetests.ts
@@ -111,6 +111,7 @@ If targeting prerelease versions, skipping versions, or using skipping some alte
 
 			if (this.flags.normalize) {
 				json.typeValidation = normalizeConfig(json.typeValidation);
+				normalizeTypeTestScript(json);
 			}
 		});
 	}
@@ -148,6 +149,24 @@ export function normalizeConfig(
 		normalized.broken = { ...defaultTypeValidationConfig.broken };
 	}
 	return normalized;
+}
+
+const typetestsGenScript = "flub generate typetests --dir . -v";
+
+/**
+ * Adds, removes, or replaces the `typetests:gen` script in the package.json `scripts` section
+ * based on whether type validation is enabled or disabled.
+ */
+export function normalizeTypeTestScript(pkgJson: PackageWithTypeTestSettings): void {
+	if (pkgJson.typeValidation?.disabled === true) {
+		if (pkgJson.scripts?.["typetests:gen"] !== undefined) {
+			// eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+			delete pkgJson.scripts["typetests:gen"];
+		}
+	} else if (pkgJson.typeValidation !== undefined) {
+		pkgJson.scripts ??= {};
+		pkgJson.scripts["typetests:gen"] = typetestsGenScript;
+	}
 }
 
 export enum VersionOptions {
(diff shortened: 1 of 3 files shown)


build-tools/packages/build-cli/src/commands/typetests.ts
```diff
@@ -150,6 +151,24 @@ export function normalizeConfig(
...
+export function normalizeTypeTestScript(pkgJson: PackageWithTypeTestSettings): void {
+	if (pkgJson.typeValidation?.disabled === true) {
+		if (pkgJson.scripts?.["typetests:gen"] !== undefined) {
+			// eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+			delete pkgJson.scripts["typetests:gen"];
+		}
+	} else if (pkgJson.typeValidation !== undefined) {
+		pkgJson.scripts ??= {};
+		pkgJson.scripts["typetests:gen"] = typetestsGenScript;
+	}
```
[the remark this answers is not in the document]
CraigMacomber: roots have an undefined typetesting block, so this does not impact them.

build-tools/packages/build-cli/src/commands/typetests.ts
```diff
@@ -111,6 +111,7 @@ If targeting prerelease versions, skipping versions, or using skipping some alte
 
 			if (this.flags.normalize) {
 				json.typeValidation = normalizeConfig(json.typeValidation);
+				normalizeTypeTestScript(json);
 			}
```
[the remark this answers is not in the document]
CraigMacomber: fixed

build-tools/packages/build-cli/src/commands/typetests.ts
```diff
@@ -150,6 +151,28 @@ export function normalizeConfig(
...
+ */
+export function normalizeTypeTestScript(pkgJson: PackageWithTypeTestSettings): void {
+	if (pkgJson.typeValidation?.disabled === true) {
+		if (pkgJson.scripts?.["typetests:gen"] !== undefined) {
+			// eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+			delete pkgJson.scripts["typetests:gen"];
+		}
+	} else if (pkgJson.typeValidation !== undefined) {
+		pkgJson.scripts ??= {};
+		pkgJson.scripts["typetests:gen"] = typetestsGenScript;
```
jason-ha, line 172: It would be nice to have an out here.
As part of the work to check protocol compatibility, I want to leverage generating special typetests files.
The current plan is to have something like
```json
   "typetests:gen": "concurrently \"npm:typetests:gen:*\"",
   "typetests:gen:api": "flub generate typetests --dir . -v",
   "typetests:gen:protocol": "flub generate typetests --dir . -v --entrypoint internal/protocol --outFile validate{@unscopedPackageName}PreviousProtocol.generated.ts",
```
Also seems like we would want to do away with more command arguments if they can't be used (without getting annoyed).

jason-ha (approved):
I have a request - to avoid future headache, but approve either way.

build-tools/packages/build-cli/src/commands/typetests.ts
CraigMacomber: I'd rather have extra configuration be part of the typetest configuration object then as extra scripts. it already lets you select the entry point: it could be made to take an array of them.

jason-ha: That could be okay. Whatever the solution this change requires extra work to be done to complete the other planned work. Or there will be hiccup in the normalize when lines like above or some workaround needs unstaged as there is no escape hatch.

PR #24877, opened 2025-06-19
(compat) Update layer compat validation tests to be generic so they can be re-used across all layer combinations

agarwal-navin:
This change updates the layer compat validation e2e tests to be generic so they can be re-used across all layer combinations. Currently, these only validate Loader / Driver compatibility. Also, added separate cases for create and load flows as these go down different paths.

Similarly, the Runtime layer's tests have been updated to be re-used across the various layers it supports compatibility with rather then duplicating them for each layer combinations.

Other changes:
- Added a try .. catch block around the container runtime initiliazation logic during container creation / load. Without this, the container is not disposed and the error is essentially lost.
- Renamed some of the compat properties in various layers to be able to differentiate between them. For instance, both Loader and DataStore layers have a property `runtimeSupportRequirements`. Appended the layer name to it to differentiate between them for better readability and usage in tests.

PR #16131, opened 2023-06-23
Log container dispose error telemetry if not already closed

kian-thompson:
Currently if container is disposed with an error, the error will always be sent as generic telemetry. The telemetry event should be an error if we haven't already closed the container (to prevent double counting the error).

[AB#4611](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/4611)

PR #14482, opened 2023-03-08
Do not log containerClose event if lifeCycleState is loading

chensixx:
## Description

right now in our logs, a lot of container close errors are not actual bugs, but permission issue, making it hard to differentiate with actual errors during lifetime of a file. so, we will no longer log container close errors if the container is in loading state
we can still rely on Container:Load_cancel events for these errors

PR #26964, opened 2026-04-07
fix(container-loader): log ContainerClose and ContainerDispose as errors when closed due to an error

MarioJGMsoft:
## Description

When container load fails, the `Container.load()` static method calls `container.close(err)` then `container.dispose(err)`. Both `ContainerClose` and `ContainerDispose` telemetry events were incorrectly logged as category `"generic"` instead of `"error"` due to two broken conditions:

- `closeCore()`: guarded on `_lifecycleState !== "loading"` — during a load failure, the container is still in `"loading"` state, so this always evaluated to `false` → `"generic"`
- `disposeCore()`: guarded on `!this.closed` — `close()` sets `this.closed = true` before `dispose()` runs, so this always evaluated to `false` → `"generic"`

This caused partners (e.g. Loop) to not receive monitor alerts when data corruption errors caused container load to fail, since their alerts filter on `category: "error"` events.

**Fix:** Remove both guards. Log as `"error"` whenever an error is passed, regardless of lifecycle state.

Note: PR #24877 previously fixed the narrow case of layer-compat validation failures during runtime instantiation (by calling `dispose()` directly without `close()` first). This PR covers the broader general load failure path.

**Verification:** 4 new unit tests in `container-loader/src/test/container.spec.ts` verify:
- `ContainerClose` is `"error"` when closed with an error during loading
- `ContainerDispose` is `"error"` when disposed with an error after close
- Both remain `"generic"` when called without an error

## Reviewer Guidance

- Note that removing the `_lifecycleState !== "loading"` guard means ALL load-phase errors (including network/auth errors) will now emit `ContainerClose` as `"error"`. This is intentional per the task requirements, but may increase error telemetry volume during load.

Fixes [AB#43473](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/43473)

### DIFF
--- a/packages/loader/container-loader/src/container.ts
+++ b/packages/loader/container-loader/src/container.ts
@@ -95,7 +95,6 @@ import {
 	createChildLogger,
 	createChildMonitoringContext,
 	formatTick,
-	isFluidError,
 	normalizeError,
 	raiseConnectedEvent,
 	wrapError,
@@ -1056,11 +1055,7 @@ export class Container
 				this.mc.logger.sendTelemetryEvent(
 					{
 						eventName: "ContainerClose",
-						category:
-							error === undefined ||
-							(isFluidError(error) && error.getTelemetryProperties().alreadyLogged === true)
-								? "generic"
-								: "error",
+						category: error === undefined ? "generic" : "error",
 					},
 					error,
 				);
@@ -1108,11 +1103,7 @@ export class Container
 				this.mc.logger.sendTelemetryEvent(
 					{
 						eventName: "ContainerDispose",
-						category:
-							error === undefined ||
-							(isFluidError(error) && error.getTelemetryProperties().alreadyLogged === true)
-								? "generic"
-								: "error",
+						category: !this.closed && error !== undefined ? "error" : "generic",
 					},
 					error,
 				);
--- a/packages/loader/container-loader/src/container.ts
+++ b/packages/loader/container-loader/src/container.ts
@@ -723,16 +723,14 @@ export class Container
 		loadProps?: Pick<IContainerLoadProps, "pendingLocalState">,
 	) {
 		super((name, error) => {
-			const normalizedError = normalizeError(error);
 			this.mc.logger.sendErrorEvent(
 				{
 					eventName: "ContainerEventHandlerException",
 					name: typeof name === "string" ? name : undefined,
 				},
-				normalizedError,
+				error,
 			);
-			normalizedError.addTelemetryProperties({ alreadyLogged: true });
-			this.close(normalizedError);
+			this.close(normalizeError(error));
 		});
 
 		const {
--- a/packages/loader/container-loader/src/container.ts
+++ b/packages/loader/container-loader/src/container.ts
@@ -1050,8 +1050,9 @@ export class Container
 			try {
 				// Raise event first, to ensure we capture _lifecycleState before transition.
 				// This gives us a chance to know what errors happened on open vs. on fully loaded container.
-				// Log as error whenever an error is present — load-phase errors (e.g. data corruption causing
-				// load failure) are legitimate errors that should surface in monitoring.
+				// Log as error whenever an error is present. Some unrelated errors might get caught during load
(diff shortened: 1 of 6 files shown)


packages/loader/container-loader/src/container.ts
```diff
@@ -1048,13 +1048,12 @@ export class Container
 			try {
 				// Raise event first, to ensure we capture _lifecycleState before transition.
 				// This gives us a chance to know what errors happened on open vs. on fully loaded container.
-				// Log generic events instead of error events if container is in loading state, as most errors are not really FF errors
-				// which can pollute telemetry for real bugs
+				// Log as error whenever an error is present — load-phase errors (e.g. data corruption causing
+				// load failure) are legitimate errors that should surface in monitoring.
```
markfields: Hmm what changed here? Did you look at the PR that introduced this comment you're changing?  I know it was years ago, but it might have useful context as to why we _wanted_ to exclude events during loading.

I did read the task and I see that there are cases that we've been missing, but let's just be sure to have eyes wide open on the complete expected fallout of this change.

Maybe we would also inspect error type...?  Sounds kind of messy, not sure.

MarioJGMsoft: The two PRs that introduced the changes are [PR#14482](https://github.com/microsoft/FluidFramework/pull/14482) and [PR#16131](https://github.com/microsoft/FluidFramework/pull/16131). I'm going through them trying to get more context for the changes

MarioJGMsoft: PR#14482:
It mentions that the change was necessary because permission issues were being caught as if they were actual errors. I don't have context as to what permission errors it's referring to, but I do wonder if their proposed solution relying on `Container:Load_cancel events` might still work.
PR#16131:
For the second change, I'm starting to see the issue that if I remove the `!this.closed` condition, we might actually start double logging the same errors, so we need to see if I should revert that part of the PR

kian-thompson: Yes, the `!this.closed` condition was explicitly included to not double count the errors, but maybe we can think of something better here. There are a few internal places that call `close(error)` then `dispose(error)` in our code, but maybe we can remove the error from the second part? The intention of allowing `dispose` to take an `error` argument was so consumers of the container could pass in whatever error they wanted when choosing to dispose.

packages/loader/container-loader/src/deltaManager.ts
```diff
@@ -425,22 +425,29 @@ export class DeltaManager<TConnectionManager extends IConnectionManager>
...
 			this.logger.sendErrorEvent(
 				{
 					eventName: "DeltaManagerEventHandlerException",
 					name: typeof name === "string" ? name : undefined,
 				},
-				error,
+				normalizedError,
 			);
-			this.close(normalizeError(error));
+			normalizedError.addTelemetryProperties({ alreadyLogged: true });
```
kian-thompson: nit: I'm not sure how I feel about this `alreadyLogged` thing. Feels a bit hacky to me.

MarioJGMsoft: Removed that logic

packages/loader/container-loader/src/container.ts
MarioJGMsoft: After discussion, the `dispose` event will remain the same, but now we're going to expand the errors we send when we close so that we always send an error, even while the container is still loading.

kian-thompson (approved):

packages/loader/container-loader/src/test/container.spec.ts
```diff
@@ -66,6 +79,101 @@ class MockContainer
...
+		);
+		assert(closeEvent !== undefined, "ContainerClose event should be logged");
+		assert.strictEqual(
+			closeEvent.category,
+			"generic",
+			"ContainerClose should be category 'generic' when closed without an error",
+		);
+	});
+
+	it("ContainerDispose is logged as generic when dispose is called without an error", () => {
```
markfields, line 158: One other case would be dispose called with an error (container not closed)

markfields (approved):

PR #26967, opened 2026-04-07
chore(eslint-config-fluid): clean up stale minimal-deprecated references

ChumpChief:
## Description

Removes remaining references to the `minimal-deprecated` eslint configuration that was collapsed into `base` in #26801. These are stale comments, docs, and code paths that were missed in that PR.

- **`recommended.js`**: Fixed broken `extends` from `./minimal-deprecated.js` (deleted file) to `./base.js`
- **`README.md`**: Removed `minimal-deprecated.mts` from the directory tree listing (file no longer exists)
- **`library/configs/base.mts`**: Removed `minimal-deprecated` from doc comment listing config tiers
- **`library/configs/overrides.mts`**: Removed stale `minimal-deprecated.js` line-number references in comments
- **`eslint-plugin-fluid/DEV.md`**: Removed `minimal-deprecated.js` from the config list
- **`generate-flat-eslint-configs.ts`**: Removed dead `minimalDeprecated` variant from type, detection heuristic, and comments

## Reviewer Guidance

The broken `extends` in `recommended.js` (pointing at the deleted `minimal-deprecated.js`) suggests the legacy `.js` configs in the package root (`base.js`, `recommended.js`, `strict.js`, `strict-biome.js`) may no longer be exercised by any consumer. No `.eslintrc` files in the repo reference them. If that's the case, we may want to delete them entirely in a follow-up rather than maintaining them — would appreciate reviewer input on that.

### DIFF
--- a/common/build/eslint-config-fluid/README.md
+++ b/common/build/eslint-config-fluid/README.md
@@ -28,7 +28,6 @@ eslint-config-fluid/
 │   ├── settings.mts            # Plugin settings (import-x, jsdoc)
 │   ├── rules/
 │   │   ├── base.mts            # Base rules from eslint:recommended, typescript-eslint, etc.
-│   │   ├── minimal-deprecated.mts  # Additional rules for minimal-deprecated config
 │   │   ├── recommended.mts     # Rules for recommended config (unicorn, type safety)
 │   │   └── strict.mts          # Rules for strict config (jsdoc requirements, explicit access)
 │   └── configs/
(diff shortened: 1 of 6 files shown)


alexvy86 (approved):

tylerbutler:
Do we need to release these fixes since 9.0.0 is released already?

ChumpChief:
> Do we need to release these fixes since 9.0.0 is released already?

These changes were based on code inspection alone - I think I must have had a bad merge or two, and/or just missed some bits somehow.

I'm not aware of anything impacted by the recommended.js error (I'm not aware if anything actually uses that or if we can just delete it).  And the generator change should be a no-op, generally.  And the rest is just documentation.

So my take would probably be to defer any release unless/until we know there is real impact.

PR #7322, opened 2021-09-01
feat: optional property in PropertyDDS

evaliyev:

PR #26971, opened 2026-04-08
fix(client-PropertyDDS): correct MSG constants

jason-ha:
PR #7322 "feat: optional property in PropertyDDS"
introduced use of several `MSG` constants that were never defined.
 - `OVERRIDEN_PROP_MUST_HAVE_SAME_CONTEXT_AS_BASE_TYPE` use was replaced by `OVERRIDEN_PROP_MUST_HAVE_SAME_FIELD_VALUES_AS_BASE_TYPE` (but never defined).
 - `CANNOT_INSERT_UNKNOWN_PROPERTY`, `MISMATCHING_PROPERTY_TYPEID`, and `CANNOT_REMOVE_NON_OPTIONAL_PROP` uses were added without defining them.

Tests improperly validated that those were thrown, but since they were `undefined` in test code too, the tests didn't fail.

`checkJs` was enabled for `property-properties` test build to find those issues and several more.
 - Some additional production typing changes were made where no runtime changes were required.
 - `global.d.ts` updates with the additional globals configured by `setup.cjs` (`@types/chai` devDep added)
 - Some incorrect test code was updated for proper execution and build validation.
   - Removed unused variables / imports.
   - Corrected duplicate declarations.
   - Fixed `BaseProperty.MODIFIED_STATE_FLAGS.DIRTY` references.
   - Fixed `ValueMapProperty` add/remove test `expect`s.
   - Workaround `console.warn` typing to access spy `.callCount` property.
   - Use intermediate `SerializedChangeSet` variable in `StringProperty` Rebasing tests.


lindsnguyen (approved):
Walked through the code with Jason, lgtm once build passes

experimental/PropertyDDS/packages/property-changeset/src/utils.ts
```diff
@@ -1620,7 +1620,7 @@ export namespace Utils {
 	export function getChangesByType(
 		in_typeid: string,
 		in_changeSet: SerializedChangeSet,
-		in_excludeTypeids: boolean,
+		in_excludeTypeids?: boolean,
 	): { insert?: object; modify?: object } {
```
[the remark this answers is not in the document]
jason-ha: This type change is made to accommodate existing test call pattern in properties-properties.

experimental/PropertyDDS/packages/property-common/src/constants.ts
```diff
@@ -1476,10 +1482,10 @@ const PropertyFactoryError = {
 	/**
 	 * methods: PropertyFactory.create
 	 *
-	 * Overridden properties must have same context as the base type.
+	 * Overridden properties must have same field values as the base type.
 	 */
-	OVERRIDEN_PROP_MUST_HAVE_SAME_CONTEXT_AS_BASE_TYPE:
-		"PF-067: Overridden properties must have same context as the base type: ",
+	OVERRIDEN_PROP_MUST_HAVE_SAME_FIELD_VALUES_AS_BASE_TYPE:
+		"PF-067: Overridden properties must have same field values as the base type: ",
```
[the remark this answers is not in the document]
jason-ha: Any external uses are expected to be meaningless as there were no internal uses of the prior name. The risk is minimal.

experimental/PropertyDDS/packages/property-properties/src/test/properties/baseProperty.spec.js
```diff
@@ -612,7 +612,10 @@ describe("BaseProperty", function () {
 	describe("Ancestry relations should be resolved correctly", function () {
 		/**
 		 * creates a workspace
-		 * @return {property-properties.Workspace} workspace
+		 * @return {Promise<import("../../index.js").NodeProperty>} workspace
+		 * @privateRemarks
+		 * The return type can be `Workspace` once package is updated to support modules. Then
+		 * {Promise<import("@fluid-experimental/property-properties").Workspace>} may be used.
```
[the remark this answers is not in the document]
jason-ha: checkJs is happy

experimental/PropertyDDS/packages/property-properties/src/test/utils.spec.js
```diff
@@ -3056,7 +3057,11 @@ describe("Utils", function () {
 		};
 
 		it("should exclude single given path", () => {
-			let res = Utils.excludePathsFromChangeSet(changeset, "assets[Prop3]");
+			let res = Utils.excludePathsFromChangeSet(
+				changeset,
+				// @ts-expect-error - testing the case when in_paths argument is a string instead of an array
+				"assets[Prop3]",
+			);
```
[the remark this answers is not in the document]
jason-ha: Prefer no runtime changes in general. This demonstrates expectations from improper usage - which is a nice feature.

experimental/PropertyDDS/packages/property-properties/src/test/utils.spec.js
```diff
@@ -3076,8 +3081,17 @@ describe("Utils", function () {
...
+				// @ts-expect-error - testing the case when in_paths argument is not passed
+				Utils.excludePathsFromChangeSet(changeset),
+			).to.be.deep.equal(changeset);
+			expect(
+				Utils.excludePathsFromChangeSet(
+					changeset,
+					// @ts-expect-error - testing the case when in_paths argument is an empty string
+					"",
+				),
+			).to.be.deep.equal(changeset);
```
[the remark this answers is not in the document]
jason-ha: Prefer no runtime changes in general. This demonstrates expectations from improper usage - which is a nice feature.

experimental/PropertyDDS/packages/property-changeset/src/changeset.ts
```diff
@@ -1006,7 +1006,7 @@ export class ChangeSet {
 	 * in a later release.
 	 * @param in_withoutRoot - Bypass a fix where the root of a changeset is cleaned
 	 */
-	public _stripReversibleChangeSet(in_withoutRoot: boolean) {
+	public _stripReversibleChangeSet(in_withoutRoot?: boolean) {
 		const callback = function (in_context) {
```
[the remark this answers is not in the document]
jason-ha: Would still prefer to have value specified as documented. This is an accommodation for existing testing calls given that production code functions fine with optional arg.

.changeset/easy-ravens-spend.md
```diff
@@ -0,0 +1,12 @@
+---
+"@fluid-experimental/property-common": minor
+"__section": fix
+---
+Add and correct MSG error constants
```
CraigMacomber: I agree with vale. 'MSG' has no definition. As much as I like Monosodium glutamate, our policy on abbreviations is to spell them out the first time.

CraigMacomber: Remember that this description will be both in the changelog for property-common (where this might make sense) but also the release notes, where I don't think this will be clear enough for people who have never used property DDS to realize they can ignore it.

.changeset/easy-ravens-spend.md
```diff
@@ -0,0 +1,12 @@
...
+
+Several error constants referenced in other PropertyDDS packages did not exist and would produce errors with "undefined".
```
CraigMacomber: Is this referring to a string "undefined" or the JS value `undefined`? Might be good to explicitly clarify.

.changeset/easy-ravens-spend.md
```diff
@@ -0,0 +1,12 @@
...
+In the past:
+ - OVERRIDEN_PROP_MUST_HAVE_SAME_CONTEXT_AS_BASE_TYPE use was replaced by OVERRIDEN_PROP_MUST_HAVE_SAME_FIELD_VALUES_AS_BASE_TYPE (but never defined).
```
CraigMacomber: Are these all caps things JavaScript properties or strings (or something else?). Quotes or code formatting would make it more clear.

.changeset/easy-ravens-spend.md
```diff
@@ -0,0 +1,12 @@
+---
+"@fluid-experimental/property-common": minor
+"__section": fix
+---
+`@fluid-experimental/property-common`: Add and correct `MSG` error constants
```
CraigMacomber: https://github.com/microsoft/FluidFramework/blob/main/.changeset/README.md#formatting

> The summary line should not contain code formatting (backticks). Too much code formatting in headings becomes unreadable, and there's no good way to measure "too much," so we avoid it altogether.

experimental/PropertyDDS/packages/property-changeset/src/utils.ts
```diff
@@ -1620,7 +1620,7 @@ export namespace Utils {
 	export function getChangesByType(
 		in_typeid: string,
 		in_changeSet: SerializedChangeSet,
-		in_excludeTypeids: boolean,
+		in_excludeTypeids?: boolean,
```
CraigMacomber, line 1623: Based on the use, it seems like this defaults to false: maybe making it explicitly do that would be better? Alternatively you can document the default.
```suggestion
		in_excludeTypeids = false,
```

CraigMacomber: Same for getChangesByPath below

.changeset/easy-ravens-spend.md
jason-ha: I know the backtick rule, but it is in conflict with the tooling that will treat `@` as an at mention. Last I saw on Teams thread was that that was not handled and backticks were okay.

experimental/PropertyDDS/packages/property-changeset/src/utils.ts
jason-ha: These changes are very explicitly only trying to change the typing (without touching production code if possible).
The only production runtime changes are some constants.

.changeset/easy-ravens-spend.md
jason-ha: There was one [Msft Teams thread](https://teams.microsoft.com/l/message/19:428af2427cbf4c91a834e0b69d0ee5f9@thread.skype/1767817966981?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=9ce27575-2f82-4689-abdb-bcff07e8063b&parentMessageId=1767817966981&teamName=Fluid%20Framework&channelName=Releases&createdTime=1767817966981) that I could find.
There was longer discussion at https://github.com/microsoft/FluidFramework/pull/26399#discussion_r2790425952.

CraigMacomber: My understanding of those discussions:
1. The people involved didn't seem to have super strong opinions on if they agree with out currently policy around use of back ticks in titles
2. Maybe we want to change the policy due to the issue of tagging github accounts

Given the policy was not changed, it seems to me that `MSG` should not get code formatting here, and neither should `@fluid-experimental/property-common` despite that one being similar to the problematic tagging case (I don't think there is a user with that name).

I don't really care about what the policy is, but I'm supposed to review things to compliance with it, and this fails.

At least change the MSG case, maybe use "" for it? If you really want the package name with the `@` code formatted, I guess I can ok that, but I'll note we should revisit the policy and update it if it has cases where we disagree with it (and I'd prefer we do that before we do the things which violate it)

experimental/PropertyDDS/packages/property-changeset/src/utils.ts
CraigMacomber: If you don't want to change the runtime behavior than the other suggested solution I had would be good, and adjust the doc comment for the parameter to indicate the default there and leave the code alone?

.changeset/easy-ravens-spend.md
jason-ha: I didn't list the package, but was asked to :/
I think backticks are needed to deal with the mentioning.
There was also question about MSG being a thing. Backticks are the way we callout code.
In my opinion the policy should go. I've nudged the Teams thread.

experimental/PropertyDDS/packages/property-changeset/src/utils.ts
jason-ha: What is the other suggested solution? Setting a default value (only suggestion I see) is a runtime `.js` change.
Updating comments to match `getChangesToTokenizedPaths` style.

.changeset/easy-ravens-spend.md
CraigMacomber: I'd go with something like `Correct some PropertyDDS error strings`

That makes it clear this is only relevant to users of PropertyDDS without having to list the package name in the title, and avoids referring to the MSG abbreviation which is not defined in said title, and avoids phrasing which is ambiguous without code quoting, and has not `@` signs.

experimental/PropertyDDS/packages/property-changeset/src/utils.ts
CraigMacomber: > Alternatively you can document the default.

And I see you have done that.

.changeset/easy-ravens-spend.md
jason-ha: done

CraigMacomber (approved):

PR #26976, opened 2026-04-08
fix(container-runtime): generate IdAlloc op only once per flush, prepended to first non-empty batch

dannimad:
The changes fix a bug where `generateIdAllocationOp()` could be called multiple times in a single flush — once for the blobAttach batch and once for the main batch — potentially producing duplicate or misplaced ID allocation ops.

- `flush() `now only calls flushInternal for non-empty batches (previously it called it unconditionally and let flushInternal return early).
- A `sendIdAllocOp: boolean` parameter is threaded through flushInternal and rebase. It is true only for the first non-empty batch flushed, ensuring `generateIdAllocationOp()` is invoked at most once per flush cycle — and only after any rebase is complete.
- (misc) Removing unused `compressionOptions` from `IBatchManagerOptions`
- (misc) Improve some tests readability.

### DIFF
--- a/packages/runtime/container-runtime/src/opLifecycle/outbox.ts
+++ b/packages/runtime/container-runtime/src/opLifecycle/outbox.ts
@@ -368,9 +368,14 @@ export class Outbox {
 			}
 			return;
 		}
-
-		this.flushInternal(this.blobAttachBatch, resubmitInfo);
-		this.flushInternal(this.mainBatch, resubmitInfo);
+		let sendIdAllocOp = true;
+		if (!this.blobAttachBatch.empty) {
+			this.flushInternal(this.blobAttachBatch, sendIdAllocOp, resubmitInfo);
+			sendIdAllocOp = false; // Only send ID alloc op for the first batch flushed, if needed. This ensures correct refSeq and prevents duplicate ID alloc ops.
+		}
+		if (!this.mainBatch.empty) {
+			this.flushInternal(this.mainBatch, sendIdAllocOp, resubmitInfo);
+		}
 	}
 
 	private flushEmptyBatch(
@@ -404,12 +409,10 @@ export class Outbox {
 
 	private flushInternal(
 		batchManager: BatchManager,
+		sendIdAllocOp: boolean, // undefined if not needed
 		resubmitInfo?: BatchResubmitInfo, // undefined if not resubmitting
 	): void {
-		if (batchManager.empty) {
-			return;
-		}
-
+		assert(!batchManager.empty, "Cannot flush an empty batch");
 		let rawBatch = batchManager.popBatch();
 
 		// On resubmit we use the original batch's staged state, so these should match as well.
@@ -435,7 +438,7 @@ export class Outbox {
 			// Note: Since this is happening in the same turn the ops were originally created with,
 			// and they haven't gone to PendingStateManager yet, we can just let them respect
 			// ContainerRuntime.inStagingMode.  So we do not plumb local 'staged' variable through here.
-			this.rebase(rawBatch, batchManager);
+			this.rebase(rawBatch, batchManager, sendIdAllocOp);
 			return;
 		}
 
@@ -450,7 +453,7 @@ export class Outbox {
 			// This ensures the refSeq is correct (matching the rest of the batch) and that
 			// ID ranges aren't lost during rebase (since reSubmit drops IdAllocation ops).
 			// Only generate for non-staged batches — ID alloc ops are always non-staged.
-			const idAllocMsg = this.params.generateIdAllocationOp();
+			const idAllocMsg = sendIdAllocOp ? this.params.generateIdAllocationOp() : undefined;
 			if (idAllocMsg !== undefined) {
 				rawBatch = { ...rawBatch, messages: [idAllocMsg, ...rawBatch.messages] };
 			}
@@ -478,8 +481,13 @@ export class Outbox {
 	 * they will end up back in the same batch manager they were flushed from and subsequently flushed.
 	 *
 	 * @param rawBatch - the batch to be rebased
+	 * @param sendIdAllocOp - whether to send ID allocation op
 	 */
-	private rebase(rawBatch: LocalBatch, batchManager: BatchManager): void {
+	private rebase(
+		rawBatch: LocalBatch,
+		batchManager: BatchManager,
+		sendIdAllocOp: boolean,
+	): void {
 		assert(!this.rebasing, 0x6fb /* Reentrancy */);
 
 		this.rebasing = true;
@@ -507,7 +515,9 @@ export class Outbox {
 			this.batchRebasesToReport--;
 		}
 
-		this.flushInternal(batchManager);
+		if (!batchManager.empty) {
+			this.flushInternal(batchManager, sendIdAllocOp);
+		}
 		this.rebasing = false;
 	}
 
--- a/packages/runtime/container-runtime/src/opLifecycle/outbox.ts
+++ b/packages/runtime/container-runtime/src/opLifecycle/outbox.ts
@@ -371,7 +371,7 @@ export class Outbox {
 		let sendIdAllocOp = true;
 		if (!this.blobAttachBatch.empty) {
 			this.flushInternal(this.blobAttachBatch, sendIdAllocOp, resubmitInfo);
-			sendIdAllocOp = false; // Only send ID alloc op for the first batch flushed, if needed. This ensures correct refSeq and prevents duplicate ID alloc ops.
+			sendIdAllocOp = false; // Only send ID alloc op for the first batch flushed, if needed. Prevents duplicate ID alloc ops.
 		}
 		if (!this.mainBatch.empty) {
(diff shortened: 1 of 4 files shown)


packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -450,7 +453,7 @@ export class Outbox {
 			// This ensures the refSeq is correct (matching the rest of the batch) and that
 			// ID ranges aren't lost during rebase (since reSubmit drops IdAllocation ops).
 			// Only generate for non-staged batches — ID alloc ops are always non-staged.
-			const idAllocMsg = this.params.generateIdAllocationOp();
+			const idAllocMsg = sendIdAllocOp ? this.params.generateIdAllocationOp() : undefined;
```
markfields, line 456: I looked at the doc comment for `takeNextCreationRange` and it says this:

> The range will include all IDs generated via calls to `generateCompressedId` since the last time a
> range was taken (via this method or `takeUnfinalizedCreationRange`).

For me, this assuages any fear of calling it twice in a row causing an invalid state.  So I'm having a hard time justifying adding the extra branching here.

anthony-murphy:
## Deep Review

*Reviewed commit `23641a2` on 2026-04-27.*

The PR's intent — fire `generateIdAllocationOp()` at most once per flush and only after rebase — is sound, and the threading of `sendIdAllocOp` through `flushInternal`/`rebase` is wired correctly in the symmetric case. One latent regression remains around how the toggle is flipped after the blobAttach call, and there is an unresolved area-owner question about whether the extra branching is needed at all.

### Findings

- **`sendIdAllocOp` is flipped to `false` regardless of whether `flushInternal` actually emitted the alloc op** — packages/runtime/container-runtime/src/opLifecycle/outbox.ts:374 — (inline thread)

### Path to Ready

Items to move from **5/10 (MAKING PROGRESS) → 7–8 (ALMOST READY)**:

- [ ] **Resolve the toggle-vs-emission gap** (see inline thread on `outbox.ts:374`). Either prove that `blobAttachBatch.staged` and `mainBatch.staged` cannot diverge within a single `flushAll` (and pin it with an assertion or comment), or change the toggle so it reflects actual emission rather than call site — for example, have `flushInternal` return a `didEmitIdAllocOp` boolean and only flip `sendIdAllocOp = false` when that returns `true`. Add a test that pins the `blobAttach staged + main non-staged` case so this can't regress silently.
- [ ] **Close out the open thread at `outbox.ts:456`.** The area owner's question — whether the extra branching is needed at all, given that `takeNextCreationRange`'s doc states it returns only IDs generated *since the last call* — deserves an explicit answer. Either add a code/PR-description note articulating the specific wire-protocol or refSeq failure mode that motivates the once-per-flush invariant, or, if a no-op fix is acceptable, adjust the PR scope.

Items to move from **7–8 → 9/10** (optional now, required for sign-off):

- [ ] Confirm the `compressionOptions` removal from `IBatchManagerOptions` via api-report; add a changeset entry if the type is exported externally.
- [ ] Add a one-line comment near the simulated `reSubmit` in the new "Rebase path…" test (`outbox.spec.ts` ~lines 271–325) pointing at the real `ContainerRuntime.reSubmit` and noting that real `reSubmit` *drops* IdAllocation ops (which is why the test only emits once after rebase). Briefly explain the `referenceSequenceNumber: Number.POSITIVE_INFINITY` sentinel.
- [ ] The new `assert(!batchManager.empty, "Cannot flush an empty batch")` at `outbox.ts:415` is untagged in a file where surrounding asserts are hex-tagged. Flag-only — the release-time bulk-tag pass will pick this up, but you may prefer to tag it in-PR for consistency.

### Context for Reviewers

- **JIT IdAllocation generation was introduced specifically to survive rebase.** The comment block at `outbox.ts:451-455` explains that idAlloc is generated just-in-time "after rebase (if any)… ID ranges aren't lost during rebase (since reSubmit drops IdAllocation ops). Only generate for non-staged batches — ID alloc ops are always non-staged." The latent bug this PR fixes — `generateIdAllocationOp()` firing twice when both batch managers are non-empty — has been present since the JIT idAlloc design was added; it surfaces only when blobAttach and main are both non-empty in the same flush. This is why the fix threads a flag through `flushInternal`/`rebase` rather than hoisting `generateIdAllocationOp()` into `flush()`.
- **Reference-sequence coherence across the two batch managers is a core invariant.** The implementation maintains independent `BatchManager`s and asserts their refSeqs match before submit; the generated IdAllocation op must carry a refSeq matching the rest of the batch. This PR changes which non-empty batch receives the single generated alloc op, so reviewers should sanity-check that the refSeq invariant still holds when both blobAttach and main are non-empty.
- **Batch metadata on resubmit is shaped for fork dedupe.** `idAllocMsg` is now prepended *before* `addBatchMetadata()` runs, so the generated IdAllocation op can become the first message carrying batch-start metadata (including `batchId`). Worth confirming that making the alloc op the batch-start message is intended for resubmit/batchId paths.
- **The `referenceSequenceNumber: undefined` change in `batchManager.spec.ts:60`** is functionally inert — `addBatchMetadata` doesn't read that field. It aligns the test data with the post-refSeq-on-runtime model.
- **Suggested reviewers**: **markfields** (densest recent ownership of `outbox.ts`/`batchManager.ts`; already has the open thread on `outbox.ts:456`), **andre4i** (rebase / refSeq design owner — directly relevant since this PR threads `sendIdAllocOp` through the rebase path), and **vladsud** (authored the precedent fix for spurious-op-in-disconnected-state, which is the same `shouldSend()`/`staged` interaction surfaced by the inline finding above).

packages/runtime/container-runtime/src/opLifecycle/outbox.ts
```diff
@@ -368,9 +368,14 @@ export class Outbox {
 			}
 			return;
 		}
-
-		this.flushInternal(this.blobAttachBatch, resubmitInfo);
-		this.flushInternal(this.mainBatch, resubmitInfo);
+		let sendIdAllocOp = true;
+		if (!this.blobAttachBatch.empty) {
+			this.flushInternal(this.blobAttachBatch, sendIdAllocOp, resubmitInfo);
+			sendIdAllocOp = false; // Only send ID alloc op for the first batch flushed, if needed. Prevents duplicate ID alloc ops.
```
anthony-murphy, line 374: Deep Review: Possible silent drop of the alloc op in the asymmetric case.

In `flushAll`:

```ts
let sendIdAllocOp = true;
if (!this.blobAttachBatch.empty) {
    this.flushInternal(this.blobAttachBatch, sendIdAllocOp, resubmitInfo);
    sendIdAllocOp = false; // unconditional
}
if (!this.mainBatch.empty) {
    this.flushInternal(this.mainBatch, sendIdAllocOp, resubmitInfo);
}
```

Inside `flushInternal` (around lines 448–456) the alloc op is only generated when `shouldSendNow = this.params.shouldSend() && !staged` **and** `sendIdAllocOp` is true:

```ts
const idAllocMsg = sendIdAllocOp ? this.params.generateIdAllocationOp() : undefined;
```

The header comment immediately above explicitly states *"Only generate for non-staged batches — ID alloc ops are always non-staged."* `staged` is computed per-batch from `rawBatch.staged` (~line 419), so the two batch managers can in principle have divergent `staged` values within a single `flushAll`.

If the blobAttach batch is staged (or `shouldSend()` flips between calls), `generateIdAllocationOp()` is **never** invoked for blobAttach — but `sendIdAllocOp` is still flipped to `false`, and the subsequent non-staged `mainBatch` is silently denied its alloc op, contradicting the invariant the comment describes ("ID alloc ops are always non-staged" — i.e. they should ride with whichever batch is non-staged).

The new `JIT IdAllocation` describe block in `outbox.spec.ts` does not appear to cover the divergent-staged case (blobAttach non-empty + staged, main non-empty + non-staged).

**Two cheap fixes:**

1. **(Smallest delta)** Have `flushInternal` return a boolean indicating whether it actually emitted the alloc op, and only flip `sendIdAllocOp = false` when that returns `true`:

   ```ts
   if (!this.blobAttachBatch.empty) {
       const emitted = this.flushInternal(this.blobAttachBatch, sendIdAllocOp, resubmitInfo);
       if (emitted) sendIdAllocOp = false;
   }
   ```

2. Hoist `generateIdAllocationOp()` into `flushAll`, call it once after the relevant rebase has settled, and pass the resulting message down. (You may have already considered and rejected this on refSeq grounds — option 1 is strictly smaller and preserves the threaded design.)

Either way, please add a test that pins the `blobAttach staged + main non-staged` case (and the inverse), so this can't regress silently.

**Question for the author:** can `blobAttachBatch.staged` and `mainBatch.staged` ever diverge within a single `flushAll`? If they cannot, an assertion or a comment pinning that invariant would be enough to close this out — the current code reads as if they can, and the toggle relies on them being symmetric.

For context, PR #20111 (vladsud — *"Compression code could send spurious ops in disconnected state"*) is direct precedent for the `shouldSend()`/`staged` branch being a known source of side-effect/caller-assumption bugs in this area.

PR #26978, opened 2026-04-08
Tweak value logic in UniformChunk cursor

CraigMacomber:
## Description

If compressor is required but missing, assert instead of returning a number where the string should be.

Avoid reading `info.shape.hasValue` multiple times.
Avoid computing `this.chunk.values[info.valueOffset]` in multiple places.

## Reviewer Guidance


packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -550,14 +550,16 @@ class Cursor extends SynchronousCursor implements ChunkedCursor {
 	public get value(): Value {
 		const idCompressor = this.chunk.idCompressor;
```
daesunp: nit: If this is only being used to check the assert and decompress the value, can we move it to the line just before the assert in line 558?

daesunp (approved):

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
CraigMacomber: Good point.

CraigMacomber:
```suggestion
```

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -550,14 +550,16 @@ class Cursor extends SynchronousCursor implements ChunkedCursor {
...
-		// If the maybeDecompressedStringAsNumber flag is set to true, we check if the value is a number.
-		// This flag can only ever be set on string leaf nodes, so if the value is a number, we can assume it is a compressible, known stable id.
-		if (info.shape.hasValue && info.shape.maybeDecompressedStringAsNumber) {
+		if (info.shape.hasValue) {
 			const value = this.chunk.values[info.valueOffset];
-			if (typeof value === "number" && idCompressor !== undefined) {
+			// If the maybeDecompressedStringAsNumber flag is set to true, we check if the value is a number.
+			// This flag can only ever be set on string leaf nodes, so if the value is a number, we can assume it is a compressible, known stable id.
+			if (info.shape.maybeDecompressedStringAsNumber && typeof value === "number") {
+				assert(idCompressor !== undefined, "chunk required but did not provide idCompressor");
```
CraigMacomber:
```suggestion
				const idCompressor = this.chunk.idCompressor;
				assert(idCompressor !== undefined, "chunk required idCompressor but did not provide it");
```

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
CraigMacomber: Fixed.

daesunp (approved):

daesunp (approved):

PR #26982, opened 2026-04-09
fix(tree): Don't apply nodeInDocument constraint to cell deletion when part of column deletion

Josmithr:
There is a known issue with nested transactions that causes constraints to not compose correctly ([AB#55763](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/55763)).

In the case of table cell removal, we apply an on-revert constraint requiring the associated column still exist in the document in order to restore the cell. This prevents cells from becoming "orphaned" nodes within their row. This is important for individual cell removal via the `removeCell` API.

But when deleting cells as a part of a column deletion operation (via the `removeColumns` API), applying this on-revert constraint currently causes reverts to be dropped, despite the fact that the associated column would be restored as a part of the same transaction.

To mitigate this, cell deletion has been refactored to not apply this on-revert constraint when occurring as part of column deletion. This is fundamentally safe, and also likely yields a performance win, since we no longer need to incur needless constraint validation overhead.

### DIFF
--- a/packages/dds/tree/src/test/tableSchema.spec.ts
+++ b/packages/dds/tree/src/test/tableSchema.spec.ts
@@ -2313,6 +2313,194 @@ describe("TableFactory unit tests", () => {
 
 			unsubscribe();
 		});
+
+		// removeColumns internally calls removeCell for each populated cell in every row.
+		// Each removeCell call creates its own nested transaction carrying:
+		//   preconditionsOnRevert: [{ type: "nodeInDocument", node: column }]
+		// That constraint means: "when reverting this cell removal, the column must still
+		// be in the tree."
+		//
+		// This is correct for the concurrent case (you don't want to restore a cell whose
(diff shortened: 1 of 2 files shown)


packages/dds/tree/src/test/tableSchema.spec.ts
```diff
@@ -2507,61 +3001,5 @@ describe("TableFactory unit tests", () => {
 
 			unsubscribe();
 		});
-
-		it("multiple inserts should be undoable if no concurrent modifications occurred", () => {
```
Josmithr, line 2511: Note for reviewers: this was moved up into the new "Undo/redo" block (and split into separate tests for column/row inserts)

jumyhre (approved):

jumyhre (approved):

PR #26988, opened 2026-04-09
Add cache for chunkShapes

brrichards:
# Summary

Adds a `Map<number, ChunkShape>` cache to `TreeShape` so that repeated calls to `withTopLevelLength(n)` with the same small value (n < 8) return the same `ChunkShape` instance instead of creating a new one each time. `Values >= 8` are not cached to prevent unbounded map growth. 

Full heap byte size difference between `objectForest` vs `chunkedForest` tested with plain text after changes. The heap snapshots are taken after building and materializing a fresh plain text tree:

| Text characters | `objectForest` | `chunkedForest`| diff |
|----|----| ---- | ---- |
| 10 | 355,336 | 359,560 | 1.19% |
| 100 | 401,224 | 375,664 | -6.37% |
| 1,000 | 943,520 | 706,652 | -25.10% |
| 10,000 | 5,334,648 | 2,807,992 | -47.36% |


packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -92,6 +92,13 @@ export type FieldShape = readonly [FieldKey, TreeShape, number];
  * Note that since this requires fields to have uniform shapes (see `FieldShape`),
  * not all trees can have their shape described using this type.
  */
+/**
+ * Maximum topLevelLength value (exclusive) for which {@link TreeShape.withTopLevelLength}
+ * caches the resulting {@link ChunkShape}. Values at or above this threshold always
+ * create a new instance to prevent unbounded cache growth.
+ */
+const chunkShapeCacheLimit = 8;
```
CraigMacomber: don't put this inbetween TreeShape and it doc coment

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -101,7 +108,13 @@ export class TreeShape {
 	public readonly positions: readonly NodePositionInfo[];
 
 	/**
-	 *
+	 * Cache for ChunkShape instances created by {@link withTopLevelLength}.
+	 * Only caches shapes with topLevelLength less than {@link chunkShapeCacheLimit} to prevent unbounded growth.
+	 * Lazy-initialized to avoid allocating a Map for TreeShape instances that never use caching.
+	 */
+	private chunkShapeCache: Map<number, ChunkShape> | undefined;
```
CraigMacomber: There should not be all that many TreeShape instances (it scales with schema, not data), so I'd prefer the smaller code size, and faster hot path (in withTopLevelLength) of just allocating the map unconditonally over saving this tiny ammount of memory.

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -161,6 +174,15 @@ export class TreeShape {
...
 
 	public withTopLevelLength(topLevelLength: number): ChunkShape {
+		if (topLevelLength < chunkShapeCacheLimit) {
+			const cached = this.chunkShapeCache?.get(topLevelLength);
+			if (cached !== undefined) {
+				return cached;
+			}
+			const shape = new ChunkShape(this, topLevelLength);
+			(this.chunkShapeCache ??= new Map()).set(topLevelLength, shape);
+			return shape;
```
CraigMacomber: if we make the cache always exist, then we can just do `return getOrCreate(this.chunkShapeCache, (length) =>new ChunkShape(this, length));

packages/dds/tree/src/test/feature-libraries/chunked-forest/uniformChunk.spec.ts
```diff
@@ -66,6 +66,26 @@ describe("uniformChunk", () => {
 				validateShape(tree.dataFactory().shape);
 			});
 		}
+		it("withTopLevelLength caches ChunkShapes for small topLevelLength values", () => {
+			const shape = new TreeShape(brand(numberSchema.identifier), true, []);
+			// Small values (< 8) should return the same cached instance
+			const a1 = shape.withTopLevelLength(1);
+			const a2 = shape.withTopLevelLength(1);
+			assert.equal(a1, a2, "expected same instance for topLevelLength 1");
```
CraigMacomber: For asserts like this in tests, omitting the custom error message is generally preferred since then you get the auto genrted one, which us usally pretty useful (like saying what the values are)

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
brrichards: I'll update to remove the lazy init

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -83,6 +83,13 @@ export class UniformChunk extends ReferenceCountedBase implements TreeChunk {
  */
 export type FieldShape = readonly [FieldKey, TreeShape, number];
 
+/**
+ * Maximum topLevelLength value (exclusive) for which {@link TreeShape.withTopLevelLength}
+ * caches the resulting {@link ChunkShape}. Values at or above this threshold always
+ * create a new instance to prevent unbounded cache growth.
+ */
+const chunkShapeCacheLimit = 8;
```
Josmithr, line 101: Nit: can we document how we landed on `8` here? That may help future maintainers determine whether or not it is appropriate to modify this.

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -111,6 +118,12 @@ export class TreeShape {
 
 	/**
 	 *
+	 * Cache for ChunkShape instances created by {@link withTopLevelLength}.
+	 * Only caches shapes with topLevelLength less than {@link chunkShapeCacheLimit} to prevent unbounded growth.
+	 */
+	private readonly chunkShapeCache: Map<number, ChunkShape> = new Map();
```
Josmithr, line 134: Question: what are the expectations for `topLevelLength`? Is it expected to always be a positive integer? Would be good to document the expectations here.

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -111,6 +118,12 @@ export class TreeShape {
 
 	/**
 	 *
+	 * Cache for ChunkShape instances created by {@link withTopLevelLength}.
+	 * Only caches shapes with topLevelLength less than {@link chunkShapeCacheLimit} to prevent unbounded growth.
```
Josmithr: Nit
```suggestion
	 * Only caches shapes with `topLevelLength` less than {@link chunkShapeCacheLimit} to prevent unbounded growth.
```

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
brrichards: Yeah, in uniform chunk it should always be a positive integer. If `topLevelLength` is 0 it should be an `emptyChunk`, but it should never be negative. I can update the docs to reflect this

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -176,6 +199,13 @@ export class TreeShape {
 	}
 
 	public withTopLevelLength(topLevelLength: number): ChunkShape {
```
Josmithr, line 204: Nit: seems like we should probably assert that `topLevelLength` is positive (unless the `ChunkShape` constructor will already do that.

Josmithr (approved):
Looks good to me! It sounded like you had done some perf analysis on this? Might be nice to include a summary of those results in the PR description for historical context.

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
brrichards: Yeah, `ChunkShape` has what you described: 
`assert(topLevelLength > 0, 0x4c6 /* topLevelLength must be greater than 0 */);`

PR #26990, opened 2026-04-09
(tree) Fix incremental summary stale handle paths

agarwal-navin:
## Bug
When a parent incremental chunk was re-encoded with a new `referenceId`, child chunks that became handles still held a `summaryPath` string referencing the old `referenceId`. On the next summary, those handle URLs pointed to keys that no longer existed in the preceding summary tree.

## Fix
Replaced `summaryPath: string` in `ChunkSummaryProperties` and `ChunkLoadProperties` with `parentReferenceId: ChunkReferenceId | undefined`. The full handle path is now computed on demand by walking up the parent chain via `latestSummaryRefIdMap` — a `Map<ChunkReferenceId, ChunkSummaryProperties>` built at the start of each summary and scoped to `TrackedSummaryProperties`. This keeps every entry correct regardless of ancestor re-encodings.

## Tests
Added parameterized test cases covering the stale-path scenarios (`[1,1]`, `[2,2]`, `[1,0,1]`, `[2,0,2]`, etc.), a multi-handle test (simultaneous handles at depth 3 and 4), and a `fullTree=true` test.


CraigMacomber:
This seems like it could impact users, so it should have a changeset.

It seems like this bug could have corrupted documents: we should document in a user facing way (the changeset) what the symptoms of this corruption would be and in what cases they might have occurred. Ideally we should also note how people can recover (I guess just roll back to an earlier version of the document?). Might be worth calling out https://fluidframework.com/docs/api/fluid-framework/#foresttypeexpensivedebug-variable as way to validate a given document is not corrupted such that it's out of schema.

agarwal-navin:
> This seems like it could impact users, so it should have a changeset.
> 
> It seems like this bug could have corrupted documents: we should document in a user facing way (the changeset) what the symptoms of this corruption would be and in what cases they might have occurred. Ideally we should also note how people can recover (I guess just roll back to an earlier version of the document?). Might be worth calling out https://fluidframework.com/docs/api/fluid-framework/#foresttypeexpensivedebug-variable as way to validate a given document is not corrupted such that it's out of schema.

Added a changeset. The document corruption will not happen because an incorrect summary was uploaded because summary upload would fail. What may result in document corruption is multiple falied summaries leading to ops accumulation which can reach a threshold after which service will reject future ops.
As such, I did not mention https://fluidframework.com/docs/api/fluid-framework/#foresttypeexpensivedebug-variable because I don't suppose it's applicable. Please let me know if it still is and I can add it.

packages/dds/tree/src/feature-libraries/forest-summary/incrementalSummaryBuilder.ts
```diff
@@ -319,17 +355,54 @@ export class ForestIncrementalSummaryBuilder implements IncrementalEncoderDecode
...
+	 * Each {@link ChunkSummaryProperties.parentReferenceId} points to the chunk's parent as it
+	 * appeared in the summary where the entry was last written. Walking up the chain from the
+	 * chunk to the root produces the full path that can be used in a summary handle path.
+	 */
+	private computeHandlePathInLatestSummary(chunkProperties: ChunkSummaryProperties): string {
+		const { latestSummaryRefIdMap } = this.requireTrackingSummary();
+		const pathSegments: string[] = [];
+		let current: ChunkSummaryProperties | undefined = chunkProperties;
+		while (current !== undefined) {
+			pathSegments.unshift(`${current.referenceId}`);
```
taylorsw04: This causes this method to be N^2. Can we push to an array and then reverse it?

agarwal-navin: Good catch. Updated it.

taylorsw04 (approved):

taylorsw04 (approved):

jzaffiro (approved):

PR #26995, opened 2026-04-10
Enable multi-node uniform Chunks and fix forest navigation

brrichards:
# Summary

Previously, chunks were only single nodes. Changing this to allow multi-node chunks (specifically tested with plain text) triggered unused codepaths that deal with indexing these multi-node chunks. This caused new errors in how indexing the nodes is done. The edit operations were conflating flat node indices with chunk array indices. Two new methods, `splitFieldAtIndex` and `isolateNodeAt` in `chunkedForest.ts` resolve a flat node index to the right chunk, expand only that chunk when needed, and return the actual chunk array position for the splice/edit to operate on. 

### Field/node changes for basicChunkCursor
The main changes involve when we push/pop chunkIndex. Previously this was done during a field -> node transition, but the index would be stale by 1 level. Previously this wasn't a problem since this stack was used to quickly get the node depth. Now it gets pushed/popped during node -> field with the correct chunkIndex, which getField now uses to find the parent in the sibling stack. The changes are located in `basicChunk.ts` since chunked forests cursor extends this class and would need overrides for many of the base methods if the logic was moved there. 

### Testing
All existing edit, splice, and fuzz tests pass. They were already testing these code paths, but with single node chunks they never hit the multi-node branches.


packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -51,6 +50,69 @@ function makeRoot(): BasicChunk {
...
+/**
+ * Split the chunk containing `nodeIndex` so that the node falls at a chunk boundary.
+ * Only the single chunk that spans the boundary is split; all other chunks are untouched.
+ * Returns the chunk-array index where `nodeIndex` starts.
+ */
+function splitFieldAtIndex(
+	chunks: TreeChunk[],
+	nodeIndex: number,
+	forest: ChunkedForest,
+): number {
```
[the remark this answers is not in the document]
brrichards: Added index check

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -51,6 +50,69 @@ function makeRoot(): BasicChunk {
...
+ * Isolate the node at `nodeIndex` into its own single-node chunk.
+ * Only the chunk containing that node is split; all other chunks are untouched.
+ * Returns the chunk-array index of the isolated single-node chunk.
+ */
+function isolateNodeAt(chunks: TreeChunk[], nodeIndex: number, forest: ChunkedForest): number {
+	let remaining = nodeIndex;
+	for (let i = 0; i < chunks.length; i++) {
+		const chunk = chunks[i] ?? oob();
+		if (remaining < chunk.topLevelLength) {
+			if (chunk.topLevelLength === 1) {
```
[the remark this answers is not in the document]
brrichards: same as above

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -454,6 +486,7 @@ export function chunkRange(
 	chunkCompressor: ChunkCompressor,
 	length: number,
 	skipLastNavigation: boolean,
+	maxNodesPerChunk?: number,
```
Josmithr: Let's be sure to add a corresponding `@param` comment above. We should also note there what the default policy is.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -417,6 +435,12 @@ export interface ChunkPolicy {
 	 * Returns information about the shapes trees of type `schema` can take.
 	 */
 	shapeFromSchema(schema: TreeNodeSchemaIdentifier): ShapeInfo;
+
+	/**
+	 * Returns true if the given field should produce single-node chunks
+	 * for incremental summary tracking.
+	 */
+	isFieldIncremental?(parentType: TreeNodeSchemaIdentifier, fieldKey: FieldKey): boolean;
```
Josmithr, line 452: Since this method is optional, we should document what it means for it to be missing.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
Josmithr, line 500: Do we have unit tests for this? Since we're modifying its behavior, it would be good to accompany these changes with new unit tests.

brrichards: I'm going to add new tests for most of changes on this PR

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
brrichards: Good call. I'll document the tradeoffs of having it enabled/disabled. For plain text it's pretty impactful if its enabled/disabled.

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
```diff
@@ -166,16 +166,12 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
...
-		// eslint-disable-next-line no-bitwise
-		const halfHeight = (this.siblingStack.length + 1) >> 1;
 		assert(
-			this.indexOfChunkStack.length === halfHeight,
+			this.indexOfChunkStack.length === this.siblingStack.length,
 			0x51c /* unexpected indexOfChunkStack */,
 		);
 		assert(
-			this.indexWithinChunkStack.length === halfHeight,
+			this.indexWithinChunkStack.length === this.siblingStack.length,
```
CraigMacomber: This change does not make sense to me.

First, why are we changing logic for basic chunks as part of this change? Isn't this change about uniform chunks?
It sems particularly odd to be modifying the non-nested case as that case should be totally unaffected by changes to other chunk types.


Secondly, the actual changes here don't make sense to me.

Lastly, this seems like its making these asserts rather different than they were before. If this code did have a bug, I'd expect to see a regression test for the bug added to go with this bug fix, and a note about the bug in the PR description (and maybe a separate PR for it).

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
```diff
@@ -322,6 +313,10 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 		assert(this.mode === CursorLocationType.Nodes, 0x528 /* must be in nodes mode */);
 		this.siblingStack.push(this.siblings);
 		this.indexStack.push(this.index);
+		// Save the chunk index of the current node so getField can find
+		// the parent BasicChunk even when the field contains multi-node chunks.
```
CraigMacomber: you can't be in a multi node chunk. This is the non nested cursor for basicChunk, and basic chunks only every have one node.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -134,9 +139,21 @@ export class Chunker implements IChunker {
...
 		);
 	}
 
+	public isFieldIncremental(
+		parentType: TreeNodeSchemaIdentifier,
+		fieldKey: FieldKey,
+	): boolean {
+		try {
+			return this.shouldEncodeIncrementally(parentType, fieldKey);
+		} catch {
```
CraigMacomber: do not catch exceptions (unless you are specifically calling an APi documented to throw a specific exception, and you confirm its exactly that exception, and its documented to leave things in a valid state afterwartds). What if that exception is indicating a bug in the code, and/or leaving the code in an invalid state.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -463,6 +496,8 @@ export function chunkRange(
 		(cursor.mode === CursorLocationType.Nodes) === length > 0,
 		0xb59 /* Should be in nodes mode if not past end */,
 	);
+	const effectiveMaxPerChunk =
```
CraigMacomber: So this is just an APi for optionally overriding the policy which is also passed in to this function? That seems pointless as the caller could just adjust the chunkCompressor.policy they provide instead. Why do we want two wants to tell this function a limit on uniform chunk size?

CraigMacomber: Please give this a better name and/or a comment to indicate if its a number of nodes (deep), or a count of subtrees or something else. To me this looks like its a number of nodes in a chunk (based on what you got it from, but below you compare it to the top level length of a chunk (a number of subtrees), which only seems correct if length of TreeShape.positionsfor the chunk is 1.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -473,7 +508,10 @@ export function chunkRange(
 		// return existing chunk with a increased ref count if possible.
 		if (start === cursor.fieldIndex) {
 			const chunkLength = cursor.chunkLength;
-			if (chunkLength <= remaining) {
+			if (
+				chunkLength <= remaining &&
+				(chunkLength <= effectiveMaxPerChunk || chunkLength === 1)
```
CraigMacomber: Generally the heuristics around uniform chunk size limits are to avoid costs related to creating really large chunks which we will just have to split layer when editing, since the benefits of really large chunks are pretty small, and those costs are significant.

If we already have an existing reusable chunk which we can just add a reference to, we should just do that, and if we need to we can split it later.
I believe this is what the code before you change would do.

Now if we have an existing large chunk, I think we split it eagerly? I'm unclear what this change is trying to accomplish/fix. I can tell though that what ever is supposed to be changed by this does not have testing, as there are not yest changes in this PR.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -51,6 +51,71 @@ function makeRoot(): BasicChunk {
...
 
+/**
+ * Split the chunk containing `nodeIndex` so that the node falls at a chunk boundary.
+ * Only the single chunk that spans the boundary is split; all other chunks are untouched.
+ * Returns the chunk-array index where `nodeIndex` starts.
+ */
+function splitFieldAtIndex(
+	chunks: TreeChunk[],
+	nodeIndex: number,
+	forest: ChunkedForest,
```
CraigMacomber: this should not depend on a forest, it should be usable without any specific forest. API should look more like basicChunkTree.

Additionally this should have tests if we are going to add it (the above suggested API change will make testing much easier)

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -51,6 +51,71 @@ function makeRoot(): BasicChunk {
 	return new BasicChunk(aboveRootPlaceholder, new Map());
 }
 
+/**
+ * Split the chunk containing `nodeIndex` so that the node falls at a chunk boundary.
+ * Only the single chunk that spans the boundary is split; all other chunks are untouched.
+ * Returns the chunk-array index where `nodeIndex` starts.
+ */
+function splitFieldAtIndex(
```
CraigMacomber, line 67: Since most chunk related stuff does copy on write, its probably a good ideal to note in the docs for this that it modified in place. This is clear from the signature (which is great), but probably worth calling out anyway.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -51,6 +51,71 @@ function makeRoot(): BasicChunk {
 	return new BasicChunk(aboveRootPlaceholder, new Map());
 }
 
+/**
+ * Split the chunk containing `nodeIndex` so that the node falls at a chunk boundary.
+ * Only the single chunk that spans the boundary is split; all other chunks are untouched.
+ * Returns the chunk-array index where `nodeIndex` starts.
```
CraigMacomber: This can return a number that is not a valid array index (the length of the array). That case seems odd, and maybe should not be supported? (Aka: maybe these docs are right, and the implementation should be made to assert in that case)

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -51,6 +51,71 @@ function makeRoot(): BasicChunk {
...
+): number {
+	assertNonNegativeSafeInteger(nodeIndex);
+	let remaining = nodeIndex;
+	for (let i = 0; i < chunks.length; i++) {
+		if (remaining === 0) {
+			return i;
+		}
+		const chunk = chunks[i] ?? oob();
+		if (remaining < chunk.topLevelLength) {
+			const expanded = mapCursorField(chunk.cursor(), (cursor) =>
```
CraigMacomber: when splitting a chunk, forcing one chunk per top level node doesn't seem right.

if I want to split a chunk thats 100 nodes long in half, I don't need to split it into 100 separate chunks, two would be sufficient.

Also we don't need to force the result to be a basicChunk.

I think this should use two calls to chunkRange when trying to split a chunk (one for the first half, and one for the second)

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -51,6 +51,71 @@ function makeRoot(): BasicChunk {
...
+	assert(remaining === 0, "node index out of bounds");
+	return chunks.length;
+}
+
+/**
+ * Isolate the node at `nodeIndex` into its own single-node chunk.
+ * Only the chunk containing that node is split; all other chunks are untouched.
+ * Returns the chunk-array index of the isolated single-node chunk.
+ */
+function isolateNodeAt(chunks: TreeChunk[], nodeIndex: number, forest: ChunkedForest): number {
```
CraigMacomber: It is unclear if this is intended to force the node to be part of a chunk with a top level length of once, or to be a basic chunk.

The name and docs for this should probably indicate which is intended.

Also, if practical, I think this should be deduplicated with splitFieldAtIndex. Maybe this can just call splitFieldAtIndex twice, or splitFieldAtIndex can optionally take multiple places to split at? I don't think we should need both, and only having one should save on code size and give is less to have to test/maintain/review

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
brrichards: The changes are in `basicChunk.ts` because that's where BasicChunkCursor lives, and that's the class `ChunkedForest`s cursor extends. That subclass is a thin wrapper from what I saw for forest specific concerns that didn't override navigation, and moving these fixes there would mean overriding `enterField/firstField/exitField/getField/getNode` and duplicating what this base class could already be doing. If thats where we want this logic I can move it. 

For the point of these changes, the point is to make `indexOfChunkStack` grow at every level instead of only at node levels. This lets `getField` find the parent chunk by its chunk array position and not the flat node index. This is necessary when the sibling chunks have `topLevelLength > 1`. The `halfheight` assertion is removed because now that it grows at every level, `indexOfChunkStack.length === siblingStack.length`

I'll add regression tests for this, especially around the assertion changes. I can also separate this into another PR

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -173,8 +238,9 @@ export class ChunkedForest implements IEditableForest {
 
 				const parent = this.getParent();
 				const destinationField = getOrAddEmptyToMap(parent.mutableChunk.fields, parent.key);
+				const destChunkIndex = splitFieldAtIndex(destinationField, destination, this.forest);
```
CraigMacomber: https://github.com/microsoft/FluidFramework/wiki/Coding-Guidelines#-avoid-using-abbreviations

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
brrichards: You're right and these docs do give the wrong impression. This logic isnt about navigating inside multi-node chunks, It's about BasicChunkCursor correctly skipping past them when walking a field's chunk array, so the flat node index and the chunk array index stay in sync. This is to make sure that getNode/getField and the stack read the right slot when multi-node chunks sit between the cursor and the target position. Again these changes are in basicChunk since this is the class the forest cursor extends.I'll update the comments for better clarity

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
brrichards: I can update it to catch usage error which is what the policy contract documents

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
brrichards: Yeah that makes sense. I'll update it so that its only in one place. this code is mainly for enforcing the single node chunks for incremental fields. This wasnt a problem before since every uniform chunk was a single node

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
brrichards: Since Im removing the `effectiveMaxPerNode` and `maxNodesPerChunk` to put the incremental enforcement in one place this guard can be removed as well.  So the reuse condition will return to what it was before these changes. Also I'll add more regression tests in general to this PR.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
brrichards: I'll update it to take ChunkCompressor and add unit tests to verify the split and isolate behavior

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
brrichards: Yeah It should be splice position and not chunk array here. Since we allow inserting at end which is the edge case of where the returned position will be outside the arrays length

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
brrichards: I agree. This was kind of a naive approach. I'll try to get this working with the two chunkRange calls

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
brrichards: I'll update this to remove the duplication. I think having two calls to splitFieldAtIndex could allow us to just keep that method and not need this one at all. Also I'll add tests for this as well

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -45,12 +45,54 @@ import {
...
 	return new BasicChunk(aboveRootPlaceholder, new Map());
 }
 
+/**
+ * Split the chunk containing `nodeIndex` so that the node falls at a chunk boundary.
+ * Only the single chunk that spans the boundary is split; all other chunks are untouched.
+ * Each half is re-chunked via `chunkRange` so the chunker's policy decides the shapes:
+ * Returns the splice index corresponding to `nodeIndex`.
+ */
+function splitFieldAtIndex(
```
CraigMacomber, line 67: as this is a new function which rather non-trivial logic, it should have tests

brrichards: There are tests that pass through this method in `chunkEncodingEndToEnd.spec.ts` on line 270. It tests splitting a 10 node length uniform chunk in multiple places asserting the intended chunks after the split

PR #26996, opened 2026-04-10
feat(eslint-config-fluid): Add `no-only-tests/no-only-tests` rule

Josmithr:
Adds [no-only-tests/no-only-tests](https://github.com/levibuzolic/eslint-plugin-no-only-tests) rule to ensure tests configured with `.only` are not checked in.

This PR also adds some basic testing infrastructure so that we can verify that configured rules are behaving the way we expect. Added a couple basic tests to ensure the new rule behaves correctly.


common/build/eslint-config-fluid/.mocharc.json
```diff
@@ -1,6 +1,7 @@
 {
-	"recursive": true,
+	"import": "tsx/esm",
+	"spec": "src/test/**/*.test.mts",
 	"reporter": "mocha-multi-reporters",
 	"reporter-options": "configFile=mocha-multi-reporter-config.json",
-	"timeout": 5000
+	"timeout": 60000
```
alexvy86: Is this necessary? Feels overkill for linting tests.

common/build/eslint-config-fluid/package.json
```diff
@@ -46,11 +48,13 @@
 	},
 	"devDependencies": {
 		"@fluidframework/build-common": "^2.0.3",
+		"@types/mocha": "~10.0.10",
 		"@types/node": "~22.19.17",
 		"@types/sort-json": "^2.0.3",
 		"concurrently": "^9.2.1",
 		"eslint": "~9.39.1",
 		"jiti": "^2.6.1",
+		"mocha": "^10.8.2",
```
alexvy86: 11.7.5 to be in line with the client release group?

common/build/eslint-config-fluid/recommended.js
```diff
@@ -239,13 +239,22 @@ module.exports = {
...
 				// file-system logic more cumbersome.
 				"unicorn/prefer-module": "off",
+
+				/**
+				 * Disallow `.only()` in tests (e.g. `describe.only`, `it.only`) to prevent accidentally
+				 * committing focused tests that would skip the rest of the suite in CI.
+				 *
+				 * @see https://github.com/levibuzolic/eslint-plugin-no-only-tests
+				 */
+				"no-only-tests/no-only-tests": "error",
```
alexvy86, line 257: So we'll have to add a disable when using `.only` for a build to pass (or just `build:compile`), right?

common/build/eslint-config-fluid/src/test/test-cases/tsconfig.json
alexvy86, line 1: I can't quite tell if this file is having any effect or how. I would have expected something like TS projects and some reference between tsconfig files but I don't see one. If it's actually necessary can we document (in the file itself?) how things work?

common/build/eslint-config-fluid/.mocharc.json
```diff
@@ -1,6 +1,7 @@
 {
-	"recursive": true,
+	"import": "tsx/esm",
```
alexvy86, line 2: What is this for?

common/build/eslint-config-fluid/package.json
Josmithr: Ah, yes. Will update.

common/build/eslint-config-fluid/.mocharc.json
Josmithr: Fair question. This was Claude, haha. I'll test if we can lower this.

common/build/eslint-config-fluid/recommended.js
Josmithr: Yes. We could configure this as a warning, if we think that's too aggressive. But I'd really love to ensure we don't accidentally check in `.only`s.

Josmithr: This is how office-bohemia is configured (though with a jest-specific rule, rather than a general purpose one).

common/build/eslint-config-fluid/.mocharc.json
Josmithr: It lets us run the tests directly against the typescript source, rather than build output (which this package doesn't generate). Leverages tsx for this.

common/build/eslint-config-fluid/src/test/test-cases/tsconfig.json
Josmithr: The tests are using this directory as though it were a TypeScript project to be linted. Eslint expects a tsconfig file for context and fails if this isn't here. I'll add a comment

common/build/eslint-config-fluid/.mocharc.json
Josmithr: Seems unneeded. No idea why it did this. Reverted.

common/build/eslint-config-fluid/package.json
```diff
@@ -46,12 +48,15 @@
...
 		"@fluidframework/build-common": "^2.0.3",
+		"@types/mocha": "~10.0.10",
 		"@types/node": "~22.19.17",
 		"@types/sort-json": "^2.0.3",
 		"concurrently": "^9.2.1",
 		"eslint": "~9.39.1",
 		"jiti": "^2.6.1",
+		"mocha": "^11.7.5",
 		"mocha-multi-reporters": "^1.5.1",
+		"tsx": "^4.19.4",
```
Josmithr: @tylerbutler I wasn't able to find a reasonable way to get mocha to work with jiti, but it works with tsx pretty seamlessly. So I restored this here for now. Let me know if this is a problem.

common/build/eslint-config-fluid/.mocharc.json
tylerbutler: This is interesting in light of https://github.com/microsoft/FluidFramework/pull/27011, where I am removing tsx and replacing it with jiti because tsx has some weird behaviors with ESM-only packages. Not necessarily a change needed here - if this merges I'll need to fix up my other PR anyway.

Josmithr: I wasn't able to find a straightforward way to get mocha and jiti to play nice together, which I noted [here](https://github.com/microsoft/FluidFramework/pull/26996#discussion_r3081954243). I'm definitely open to alternatives, just not sure what the best options are.

Your above PR already merged, and I resolved the conflicts here. Unless you have another PR in the works that you're referring to.

tylerbutler: Duhhh, #27011 already merged and is in this PR... but could you use jiti here instead?

Josmithr: Not out of the box. I couldn't find any solutions online. Claude churned on this for a while and found a solution that involved a bunch of custom test setup logic. My gut inclination was that the extra dep was probably better than a bunch of custom test setup boilerplate. But I will defer to you on this. (I could probably have it regenerate those changes too if you would like to compare)

tylerbutler (approved):

alexvy86 (approved):
Seems good to me. I'll let @tylerbutler comment on the tsx change.

Tangent: now that we're on eslint9, when do we merge esilnt-plugin-fluid into eslint-config-fluid? 😄

Josmithr:
> Seems good to me. I'll let @tylerbutler comment on the tsx change.
> 
> Tangent: now that we're on eslint9, when do we merge esilnt-plugin-fluid into eslint-config-fluid? 😄

That's right! I forgot we could do that starting with v9. Have we removed v8 support yet? Presumably we can't do that until that support has been removed.

PR #26983, opened 2026-04-09
deps: Update api-extractor

alexvy86:
## Description

Update `api-extractor` across the repo, including updating the patches we maintain for it. Partly motivated for the need to update `lodash` to a more recent version to address a CVE, which I'm trying to do by updating the packages that bring it in rather than through a forced override.

`api-extractor` seems to have fixed some bugs and detected that `fluid-framework` was not exporting `ISharedObjectKind` which it should have been doing (thanks @jason-ha for the explanation and suggested fix).

Also note that a couple of API report files were updated with what seem to be reasonable changes (fewer `_2` types, and some de-structured args now aren't). I suspect that has to do with the fact that [api-extractor now uses TypeScript 5.9.3](https://github.com/microsoft/rushstack/blob/main/apps/api-extractor/CHANGELOG.md#7580).

Finally, we had to extend the use of the workaround that imports unused types into a file, just so api-extractor generates reports without inline `import()` statements to places that arguably already needed it, but for some reason weren't causing the reports to get those `import()`s.

Github Copilot summary of what it did:

```
Here's a summary of what was done:

Updated @microsoft/api-extractor from various versions to 7.58.1 across all workspaces:

┌──────────────────────┬────────────────────────┬──────────────────────┬──────────┐
│ Workspace            │ Old Version            │ New Version          │ Packages │
├──────────────────────┼────────────────────────┼──────────────────────┼──────────┤
│ Root (client)        │ 7.52.11 / ^7.52.11     │ 7.58.1 / ^7.58.1     │ 98       │
├──────────────────────┼────────────────────────┼──────────────────────┼──────────┤
│ build-tools          │ ^7.55.1                │ ^7.58.1              │ 5        │
├──────────────────────┼────────────────────────┼──────────────────────┼──────────┤
│ server/routerlicious │ ^7.45.1                │ ^7.58.1              │ 6        │
└──────────────────────┴────────────────────────┴──────────────────────┴──────────┘

Patches updated (paths changed from lib/ → lib-commonjs/ due to package restructuring):

 - patches/@microsoft__api-extractor@7.52.11.patch → @7.58.1.patch (ValidationEnhancer + ApiReportGenerator)
 - server/routerlicious/patches/@microsoft__api-extractor@7.45.1.patch → @7.58.1.patch (ValidationEnhancer only)

Note: Used 7.58.1 instead of 7.58.2 because the latter was published <24h ago and fails the minimumReleaseAge constraint. The server lockfile
[...]

PR #26997, opened 2026-04-10
test(client-presence-definitions): enable API reports

jason-ha:
- enable api-extractor reporting and linting (subset of what `presence` already does)

Additionally, remove commented out lines in `presenceStatesTypes.ts`


jason-ha:
Note: will be in conflict with #26983

WillieHabi (approved):
LGTM

packages/framework/presence-definitions/api-extractor.json
```diff
@@ -0,0 +1,8 @@
+{
+	"$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+	"extends": "../../../common/build/build-common/api-extractor-base.esm.no-legacy.json",
+	// Disable docModel as customer facing API surface is from `presence` package.
```
Josmithr, line 4: Nit: seems fine, but also doesn't really seem necessary?

Josmithr (approved):

packages/framework/presence-definitions/api-extractor.json
jason-ha: I don't want any smart AI looking to `@fluid-internal/presence-definitions` for inspiration.

packages/framework/presence-definitions/package.json
```diff
@@ -86,6 +93,16 @@
...
+	"fluidBuild": {
+		"tasks": {
+			"check:exports:cjs:index": [
+				"tsc"
+			],
+			"check:exports:esm:index": [
+				"build:esnext"
+			]
+		}
+	},
```
alexvy86, line 105: Question: why don't all packages need these? And is it correct that test-loader-utils has something similar but not quite identical to this (and is the only one that has something like it, other than client-utils which has a slightly more contrived setup).

jason-ha: There aren't any default rules for `check:exports:*` tasks (when things are complex). There is one for "check:exports" that shields us from most cases where there could be a problem. I have been working on getting a faster isolated esm only build. In there, I will add rules for the more common versions of these, but the one-offs like these will need their own specifications.

PR #26998, opened 2026-04-10
chore(server): Upgrade server Dockerfiles to Node 22

ChumpChief:
## Description

Updates the base image in the routerlicious, historian, and gitrest Dockerfiles from
`node:20.19.0-bookworm-slim` to `node:22.22.2-bookworm-slim` with a pinned sha256 digest.

This is the follow-up to #26934, which upgraded development tooling and CI to Node 22. The
gitssh Dockerfile uses Alpine (no Node) and does not require changes.

## Reviewer Guidance

- All three images were built locally with `docker build` to verify the Dockerfiles are valid.
- The Node 22.22.2 base image must be baked into CI build agents before this PR can merge.


alexvy86 (approved):
Approving with one caveat: I don't think the docker image for 22.22.2 is the one we have baked in the agents. I added the Node22 and 24 docker images on 2/20, when only 22.22.0 was out. I'd suggest we hold merging this until I put the newer one in the agents.

ChumpChief:
> Approving with one caveat: I don't think the docker image for 22.22.2 is the one we have baked in the agents. I added the Node22 and 24 docker images on 2/20, when only 22.22.0 was out. I'd suggest we hold merging this until I put the newer one in the agents.

Will hold until your greenlight.  Alternatively, is updating the agents something that I should be looking at?  I have not really ever looked at the agent images.

alexvy86:
> Will hold until your greenlight. Alternatively, is updating the agents something that I should be looking at? I have not really ever looked at the agent images.

There's documentation on how to do it in our internal repo, but I'm already in the middle of the process so don't worry about it this time.

alexvy86:
Agent pools updated with these docker images baked in, as well as Node v22.22.2 on the agents themselves. Merge away!

PR #27000, opened 2026-04-10
Add utilities for easier memory benchmarking.

CraigMacomber:
## Description

It is very easy to screw up writing memeory benchmarks.

These changes make it much easier to get them right.

## Reviewer Guidance


tools/benchmark/src/test/memoryBenchmarking/memoryUtils.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+		it("setter can overwrite an existing value", () => {
+			const box = Box.full(1);
+			box.value = 2;
+			assert.equal(box.value, 2);
+		});
+	});
+
+	describe("memoryAddedBy", () => {
+		benchmarkIt({
+			title: "inserting 1000 numbers into an array",
```
alexvy86: To match the other one; the operation in this case seems irrelevant, I think.
```suggestion
			title: "sync setup and modify",
```

tools/benchmark/src/test/memoryBenchmarking/memoryUtils.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+						}
+					},
+				}),
+			),
+		});
+	});
+
+	describe("memoryUseOfValue", () => {
+		benchmarkIt({
+			title: "array with 1000 numbers",
```
alexvy86: Same as above
```suggestion
			title: "sync factory",
```

tools/benchmark/src/memoryBenchmarking/configuration.ts
```diff
@@ -33,8 +33,11 @@ export interface MemoryUseCallbacks {
  * @remarks
  * Use with {@link benchmarkMemoryUse} to measure the memory usage of the operation.
  * Only supported in Node.js environments with `--expose-gc`.
+ *
+ * {@link memoryUseOfValue} and {@link memoryAddedBy} can be used to create this for common cases.
+ * When implementing other cases, consider using {@link Box} to manage memory retention issues due to local variables.
+ *
  * @public
- * @input
```
alexvy86, line 37: I forget the exact semantics that `@input` intends to convey... this type does not satisfy them anymore because free functions return it, even if the functions are controlled by us and the returned object is only supposed to work as an input to other APIs in the package?

tools/benchmark/src/memoryBenchmarking/memoryUtils.ts
```diff
@@ -0,0 +1,161 @@
...
+	/**
+	 * The value contained in the box.
+	 * @throws if the value is undefined.
+	 */
+	public get value(): T {
+		assertProperUse(this.inner !== undefined, "Box is empty");
+		return this.inner;
+	}
+
+	public set value(v: T) {
```
alexvy86, line 68:
```suggestion
	/**
	 * The value to store in the box.
	 * @throws if the value is undefined.
	 */
	public set value(v: T) {
```

CraigMacomber: Do that and API extractor will complain the documentation is supposed to be on the getter instead.

Both share the same doc comment, in intellisense and API extractor.

tools/benchmark/src/memoryBenchmarking/configuration.ts
CraigMacomber: > This type is "input," meaning that code outside of the library defining it should not read from it. Future versions of this type may add optional members or make typing of members more general.

Since we never do non-breaking releases of benchmark, it doesn't really matter, so I figured we might as well allow people to read from the output of these new functions. I don't think that's really required if they are only passed right back into our own APIs as the same type or used with ... into the same type, but as we don't do non breaking releases, relaxing the rules seems fine.

tools/benchmark/src/memoryBenchmarking/memoryUtils.ts
alexvy86: Huh, interesting. Ok, pretty minor.

alexvy86 (approved):

PR #27004, opened 2026-04-13
 feat(client-fluid-framework): expose getPresence from fluid-framework

jason-ha:
- Move support to new `@fluid-internal/presence-static`, re-export from `fluid-framework`, and deprecate from `@fluidframework/presence`
- `fluid-framework` only exports `getPresence*`. Callers can access remaining surface from `@fluidframework/presence`.
- Update example and test uses


jason-ha:
Original plan merged - this version is not needed

PR #27008, opened 2026-04-13
Reduce number of intervals in endpointInRangeIndex.spec.ts fuzz test

Abe27342:
## Description

This test regularly hovers near the timeout, and recently caused a PR failure due to going over it.

Given the test config, it roughly makes sense that it would take a while: pre-PR, it adds 800 intervals to a SharedString with 1k characters, then performs 100 queries at random positions comparing the actual API with a brute-force test implementation which isn't guaranteed to be particularly fast. I'd estimate the original test calls `localReferencePositionToPosition` on this sharedstring somewhere on the order of 500k times among other bookkeeping. Additionally, the API cannot hope to ever be more efficient than O(# of intervals in overlap), which will tend to be rather large given the density of intervals in the string and the test strategy of generating random positions.

I've reduced the number of intervals in the string by a factor of 2 to reduce test time. This should still provide some decent smoke coverage.


noencke (approved):

Abe27342:
closing - this ended up being node 22 perf regressions in coverage which we have since done alternate mitigations for.

PR #27007, opened 2026-04-13
test(app-insights-logger): Skip flaky test

alexvy86:
## Description

Skips a test that's been timing out in PR builds lately. Opened [AB#69025](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/69025) to track it.

## Reviewer Guidance

The review process is outlined on [this wiki page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

PR #27015, opened 2026-04-13
fix(fluid-telemetry,app-insights-logger): Fix flaky ApplicationInsights test timeouts

ChumpChief:
[AB#71221](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/71221)

## Description

Tests in `@fluidframework/fluid-telemetry` and `@fluidframework/app-insights-logger` were
intermittently timing out in CI. Both had the same root cause.

Without calling `loadAppInsights()`, the first `trackEvent` call on an `ApplicationInsights`
instance triggers an expensive fallback initialization path (~250ms locally, potentially much
worse under CI resource contention). Previously, `beforeEach` created a new instance per test,
so every test paid this cost within its own timeout budget.

The fix for both packages:
- Moves `ApplicationInsights` construction from `beforeEach` to `before` (suite-level)
- Calls `loadAppInsights()` to eagerly initialize the SDK, avoiding the expensive fallback
  path entirely (reduces first `trackEvent` from ~250ms to ~1ms)
- Adds `afterEach` to restore the sinon spy between tests (necessary since the instance is
  now shared)
- Uses named functions for mocha hooks for better debuggability in stack traces, with an
  eslint override (`allowNamedFunctions`) scoped to `*.spec.ts` files

The `app-insights-logger` change also unskips the test that was skipped in #27007 ([AB#69025](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/69025)).

## Reviewer Guidance

### DIFF
--- a/packages/framework/client-logger/fluid-telemetry/src/test/containerTelemetry.spec.ts
+++ b/packages/framework/client-logger/fluid-telemetry/src/test/containerTelemetry.spec.ts
@@ -49,7 +49,11 @@ describe("container telemetry via", () => {
 	let trackEventSpy: Sinon.SinonSpy;
 	let telemetryConfig: TelemetryConfig;
 
-	beforeEach(() => {
+	// Create the ApplicationInsights instance once for the suite and eagerly initialize it.
+	// Without loadAppInsights(), the first trackEvent call triggers an expensive fallback
+	// initialization path (~250ms locally, potentially much worse in CI). Calling
+	// loadAppInsights() up front avoids that path entirely, reducing first trackEvent to ~1ms.
+	before(() => {
 		appInsightsClient = new ApplicationInsights({
 			config: {
 				connectionString:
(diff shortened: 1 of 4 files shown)


packages/framework/client-logger/fluid-telemetry/src/test/containerTelemetry.spec.ts
```diff
@@ -49,15 +49,22 @@ describe("container telemetry via", () => {
 	let trackEventSpy: Sinon.SinonSpy;
 	let telemetryConfig: TelemetryConfig;
 
-	beforeEach(() => {
+	// Create the ApplicationInsights instance once for the suite and eagerly initialize it.
+	// Without loadAppInsights(), the first trackEvent call triggers an expensive fallback
+	// initialization path (~250ms locally, potentially much worse in CI). Calling
+	// loadAppInsights() up front avoids that path entirely, reducing first trackEvent to ~1ms.
+	before(() => {
```
jason-ha: suggestion: It isn't common use yet, but if you do `before(function loadAppInsights(): void => {` then function name will be used whenever there is a problem in there and things are much easier to debug. Applies to beforeEach as well. A nice descriptive name will distinguish those hooks.

jason-ha (approved):

jason-ha (approved):

packages/framework/client-logger/fluid-telemetry/src/test/containerTelemetry.spec.ts
ChumpChief: Updated in latest

jason-ha (approved):
You're missing return types, but if we aren't linting that in these, then we're good.

jason-ha (approved):

PR #27016, opened 2026-04-13
feat(core-interfaces): Add info and essential log levels to LogLevel …

MarioJGMsoft:
## Description

Adds two new named log levels to the `LogLevel` constant in `@fluidframework/core-interfaces`:                                           
                                                                                                                                         
- **`info` (20)** — Session-level information that may be omitted in some sessions if needed (e.g. to reduce overall telemetry volume), but if any are collected in a session, all should be.                                                                                    
- **`essential` (30)** — Essential information about the operation of Fluid that should always be collected, even in production, for diagnostic purposes.                                                                                                                                                                                           
                                                                   
This PR also introduces a new public interface called `LogLevelConst` that types the `LogLevel` constant. This was necessary because API Extractor does not capture per-property JSDoc comments  on `as const` object literals — it treats them as a single variable. By extracting the properties into a named interface, tags on the object of the type are properly surfaced in the API reports.   
                                                                                                                                         
## Reviewer Guidance

- No logger implementations are updated in this PR — only the `LogLevel` constant is extended.
- The new `LogLevelConst` interface is a public export solely to allow API Extractor to document the individual members of `LogLevel`. It is not intended as a general-purpose type for consumers to implement. 

[AB#58991](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/58991)


packages/common/core-interfaces/src/index.ts
```diff
@@ -60,6 +60,7 @@ export type {
 export { IFluidHandleContext, IFluidHandle, fluidHandleSymbol } from "./handles.js";
 
 export type {
+	LogLevelConst,
```
jason-ha: nit: sort please

packages/common/core-interfaces/src/logger.ts
```diff
@@ -54,11 +54,51 @@ export interface ITelemetryBaseEvent extends ITelemetryBaseProperties {
  * Specify levels of the logs.
  * @public
  */
-export const LogLevel = {
-	verbose: 10, // To log any verbose event for example when you are debugging something.
-	default: 20, // Default log level
-	error: 30, // To log errors.
-} as const;
+export interface LogLevelConst {
```
jason-ha, line 62: Let's give this guy a privateRemarks for why we can't just use implicit const typing.

packages/common/core-interfaces/src/logger.ts
```diff
@@ -54,11 +54,51 @@ export interface ITelemetryBaseEvent extends ITelemetryBaseProperties {
...
+
+	/**
+	 * Essential information about the operation of Fluid.
+	 * @remarks It is recommended that these should always be collected, even in production, for diagnostic purposes.
+	 * If an event does not have a log level specified, it should be treated as if it were LogLevel.essential.
+	 */
+	readonly essential: 30;
+
+	/**
+	 * @deprecated Use {@link LogLevelConst.info}. If an event does not have a log level specified, it should be treated as if it were LogLevel.essential.
```
jason-ha: `{@link LogLevelConst.info}` feels wrong. 
We would want someone to be able to copy from the text and so it should read `{@link LogLevel.info}`. If that is unhappy and `variable` qualifier doesn't work. Then use `|` alt-text syntax.

packages/common/core-interfaces/src/logger.ts
```diff
@@ -75,13 +115,13 @@ export interface ITelemetryBaseLogger {
 	/**
 	 * Log a telemetry event, if it meets the appropriate log-level threshold (see {@link ITelemetryBaseLogger.minLogLevel}).
 	 * @param event - The event to log.
-	 * @param logLevel - The log level of the event. Default: {@link (LogLevel:variable).default}.
+	 * @param logLevel - The log level of the event. Default: {@link LogLevelConst.default}.
```
jason-ha: `LogLevel.info` isn't the default?

packages/common/core-interfaces/src/logger.ts
```diff
@@ -75,13 +115,13 @@ export interface ITelemetryBaseLogger {
...
 	 */
 	send(event: ITelemetryBaseEvent, logLevel?: LogLevel): void;
 
 	/**
 	 * Minimum log level to be logged.
-	 * @defaultValue {@link (LogLevel:variable).default}
+	 * @defaultValue {@link LogLevelConst.default}
```
jason-ha: `LogLevel.info`

packages/utils/telemetry-utils/src/mockLogger.ts
```diff
@@ -41,7 +41,8 @@ export class MockLogger implements ITelemetryBaseLogger {
 	public readonly minLogLevel: LogLevel;
 
 	public constructor(minLogLevel?: LogLevel) {
-		this.minLogLevel = minLogLevel ?? LogLevel.default;
+		//  The default value for logLevel will be updated to {@link @fluidframework/core-interfaces#LogLevel.essential} once issue #26910 is resolved.
```
jason-ha: I think this comment is incorrect. Won't this need to remain `info`?

packages/utils/telemetry-utils/src/mockLogger.ts
```diff
@@ -59,7 +60,8 @@ export class MockLogger implements ITelemetryBaseLogger {
 	 * {@inheritDoc @fluidframework/core-interfaces#ITelemetryBaseLogger.send}
 	 */
 	public send(event: ITelemetryBaseEvent, logLevel?: LogLevel): void {
-		if ((logLevel ?? LogLevel.default) >= this.minLogLevel) {
+		//  The default value for logLevel will be updated to {@link @fluidframework/core-interfaces#LogLevel.essential} once issue #26910 is resolved.
```
jason-ha: This doesn't need to wait for that, right?
I support changing to essential in a separate PR, but I think it can be done right away.

.changeset/tender-poems-float.md
```diff
@@ -0,0 +1,22 @@
...
+Deprecated default and error in LogLevel
+
+Two new `LogLevel` members have been added to `@fluidframework/core-interfaces` to better express intent:
+
+- `LogLevel.info` — replaces `LogLevel.default` (same numeric value). Use this for informational session logs that could be omitted at high telemetry volumes.
+- `LogLevel.essential` — replaces `LogLevel.error` (same numeric value). Use this for critical operational events that should always be collected.
+
+`LogLevel.default` and `LogLevel.error` are now `@deprecated` and will be removed in a future release (see [issue #26969](https://github.com/microsoft/FluidFramework/issues/26969) for the removal timeline).
+
+If an event does not have a log level specified, it should be treated as if it were `LogLevel.essential`.
```
jason-ha: If that is true, then we should be doing that in this code too, yeah? Okay to be in a separate PR as noted elsewere.

packages/common/core-interfaces/src/index.ts
MarioJGMsoft: Applied change

.changeset/tender-poems-float.md
MarioJGMsoft: I added this comment to provide more guidance as to what to do if a client receives a telemetry event without a LogLevel. I don't think it implies any code changes.

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: Applied docs

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: Applied change

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: It is

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/mockLogger.ts
MarioJGMsoft: You're right, I'll update the comment accordingly

packages/utils/telemetry-utils/src/mockLogger.ts
MarioJGMsoft: I had thought of updating it to essential with the change we are also doing to `MockLogger.send()`, but since the MockLogger is meant to only be for testing, we can leave it as `info`

noencke (commented):
Have you run the integration pipeline against this branch?

noencke (approved):
Integration pipeline has passed - https://office.visualstudio.com/OC/_build/results?buildId=47481008&view=results

Approving for API.

packages/common/core-interfaces/src/logger.ts
jason-ha: Based on the current change, I am guessing you can't get a link to `LogLevel.info`? I would have thought that you could. (Wasn't the prior link to LogLevel.info?)
Please explain for my understanding.

packages/common/core-interfaces/src/logger.ts
jason-ha: Still referencing `LogLevelConst`.

packages/common/core-interfaces/src/logger.ts
jason-ha: Still referencing `LogLevelConst`.

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
```diff
@@ -271,6 +271,7 @@ describe("ChildLogger", () => {
 
 		childLogger1.send({ category: "generic", eventName: "testEvent" }, LogLevel.info);
 		assert(sent, "info event should be sent");
+		sent = false;
 
 		childLogger1.send({ category: "generic", eventName: "testEvent" }, LogLevel.essential);
```
jason-ha: Add a comment for the sent reset. Probably better to separate subcases from each other. (Even better when individual test cases have their own test case.)

I recommend `// Setup`, `// Act`, and `// Verify` blocks for most test cases and especially when it isn't obvious.
For this it could be
```suggestion
		assert(sent, "info event should be sent");
		
		// Test - essential should be sent
		// Setup
		sent = false;
		// Act
		childLogger1.send({ category: "generic", eventName: "testEvent" }, LogLevel.essential);
		// Verify
```

packages/utils/telemetry-utils/src/mockLogger.ts
jason-ha: Tracking work item?

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
MarioJGMsoft: Applied changes

.changeset/tender-poems-float.md
jason-ha: I don't know what you mean it doesn't imply any code changes.
Perhaps you mean that the introduction of these two names and the deprecation doesn't require a client code change.
There are sort of two issues tied together here. 1) new symbolic names and 2) change in treatment of missing LogLevel.
The first says `.default` -> `.info` and `.error` -> `.essential`.
The second says some defaults go to `20` and others to `30` (and is code change*)
The combination means some `.default` -> `.info` (where used as a min) and others -> `.essential` (where used for categorization of an event) and `.error` -> `.essential`.

Either this PR does both or it does one of the two issues. If this is only doing new symbolic names, then since `.default` has semantic meaning to the change to come, I would leave it not deprecated. (`.error` is fine to deprecate since there is not special semantics that are important for planned changes.)

`*` - if at the time we do (2), we ensure that all cases where we call `send` that we are sending a level, then there are no cases for a customer where they would need to apply a runtime logic change. I am sure this can be accomplished without requiring the breaking change.

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: It took me a sec to fix but I was able to get it working with the `|` syntax. What happened was that I needed to specify that it's the `LogLevel:variable`. I thought I had applied these changes but I missed this specific change. It should all be fixed now

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: I mistook this for the comment in the `LogLevelConst` interface, fixed it.

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: I mistook this for the comment in the `LogLevelConst` interface, fixed it.

packages/utils/telemetry-utils/src/mockLogger.ts
MarioJGMsoft: It's being tracked by this [item](https://dev.azure.com/fluidframework/internal/_workitems/edit/68968)

.changeset/tender-poems-float.md
MarioJGMsoft: I think I understand better now, because of my comment I am saying that if there is no LogLevel, it should be considered `.essential (30)`, but currently if there is no LogLevel, then it is considered `.default (20)`. I think both issues should be a part of this PR, so I'll apply the changes to try to cover the case you mentioned in the `*` section

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
jason-ha: I feel like you mostly applied. There is a missing vertical space (blank line) :p

packages/dds/merge-tree/src/snapshotChunks.ts
```diff
@@ -5,7 +5,7 @@
 
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 
-import type { IFluidHandle } from "@fluidframework/core-interfaces";
+import { type IFluidHandle, LogLevel } from "@fluidframework/core-interfaces";
```
jason-ha: Keep type and value imports separate (you should review for this yourself - I have said it decent number of times now in past reviews)

packages/dds/merge-tree/src/snapshotChunks.ts
```diff
@@ -93,12 +93,15 @@ export function serializeAsMinSupportedVersion(
...
 	if (chunk.version !== undefined) {
-		logger.send({
-			eventName: "MergeTreeChunk:serializeAsMinSupportedVersion",
-			category: "generic",
-			fromChunkVersion: chunk.version,
-			toChunkVersion: undefined,
-		});
+		logger.send(
+			{
+				eventName: "MergeTreeChunk:serializeAsMinSupportedVersion",
```
jason-ha: Should add notes in PR for all of the events that are getting a specific treatment. Just commenting on this one.

packages/loader/container-loader/src/deltaManager.ts
```diff
@@ -11,10 +11,11 @@ import type {
...
-import type {
-	IEventProvider,
-	ITelemetryBaseEvent,
-	ITelemetryBaseProperties,
+import {
+	type IEventProvider,
+	type ITelemetryBaseEvent,
+	type ITelemetryBaseProperties,
+	LogLevel,
 } from "@fluidframework/core-interfaces";
```
jason-ha: keep type and value imports separate

packages/loader/container-loader/src/deltaManager.ts
jason-ha, line 132: Add note that all events sent are marked as essential

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -2141,11 +2142,14 @@ export class ContainerRuntime
...
-			eventName: "DeviceSpec",
-			...getDeviceSpec(),
-		});
+		this.baseLogger.send(
+			{
+				category: "generic",
+				eventName: "DeviceSpec",
+				...getDeviceSpec(),
+			},
+			LogLevel.info,
```
jason-ha: Are we sure? Who is making calls per event?
I would have guessed that this could be information that we always want.

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
```diff
@@ -257,10 +257,10 @@ describe("ChildLogger", () => {
 				}
 				sent = true;
 			},
-			minLogLevel: LogLevel.default,
+			minLogLevel: LogLevel.info,
 		};
 		const multiSinkLogger = createMultiSinkLogger({
-			loggers: [logger1, new MockLogger(LogLevel.error)],
+			loggers: [logger1, new MockLogger(LogLevel.info)],
```
jason-ha: Why switch if this was explicitly min level 30 before?

packages/utils/telemetry-utils/src/config.ts
```diff
@@ -3,10 +3,11 @@
...
-import type {
-	ConfigTypes,
-	IConfigProviderBase,
-	ITelemetryBaseLogger,
+import {
+	type ConfigTypes,
+	type IConfigProviderBase,
+	type ITelemetryBaseLogger,
+	LogLevel,
 } from "@fluidframework/core-interfaces";
```
jason-ha: keep type and value imports separate

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -183,16 +183,18 @@ export abstract class TelemetryLogger implements ITelemetryLoggerExt {
...
 	 * @param logLevel - optional level of the log. It category of event is set as error,
 	 * then the logLevel will be upgraded to be an error.
+	 *
+	 * @remarks The default value for logLevel will be updated to {@link @fluidframework/core-interfaces#LogLevel.essential}.
 	 */
 	public sendTelemetryEvent(
 		event: ITelemetryGenericEventExt,
 		error?: unknown,
-		logLevel: typeof LogLevel.verbose | typeof LogLevel.default = LogLevel.default,
+		logLevel: typeof LogLevel.verbose | typeof LogLevel.info = LogLevel.info,
```
jason-ha: What are we waiting on here?
Why can't it be defaulted to essential (like the guidance says)?

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -459,8 +463,9 @@ export class ChildLogger extends TelemetryLogger {
 	}
 
 	private shouldFilterOutEvent(event: ITelemetryBaseEvent, logLevel?: LogLevel): boolean {
-		const eventLogLevel = logLevel ?? LogLevel.default;
-		const configLogLevel = this.baseLogger.minLogLevel ?? LogLevel.default;
+		// The default value for eventLogLevel will be updated to LogLevel.essential in the future, see work item #68968 for more information.
+		const eventLogLevel = logLevel ?? LogLevel.info;
```
jason-ha: Why are some things changed, but not this?

jason-ha (commented):
more changes; more questions :)

packages/common/core-interfaces/src/logger.ts
jason-ha: Doesn't look changed from when I last asked. The link is now `(LogLevel:variable) | LogLevel.info`. Before it was `LogLevel.info`. Why can't the reference be to `.info`? Why can you only get to the constant now?

packages/common/core-interfaces/src/logger.ts
jason-ha: Now has the `(LogLevel:variable) | LogLevel.info` question like above

packages/common/core-interfaces/src/logger.ts
```diff
@@ -52,13 +52,55 @@ export interface ITelemetryBaseEvent extends ITelemetryBaseProperties {
...
+
+	/**
+	 * Default LogLevel
+	 */
+	readonly default: 20;
+
+	/**
+	 * To log errors.
+	 */
+	readonly error: 30;
```
jason-ha, line 92: While not actually deprecating, you can still add the comments that info and essential should be used instead.
These were really undocumented before. Now that they are documented, you don't want them to look like recommendations.

packages/common/core-interfaces/src/logger.ts
```diff
@@ -52,13 +52,55 @@ export interface ITelemetryBaseEvent extends ITelemetryBaseProperties {
...
+}
+
+/**
+ * {@inheritDoc LogLevelConst}
```
jason-ha: The only thing this inherits is "* Specify levels of the logs."
Better to have its own description. Perhaps
```suggestion
 * Provides runtime {@link (type:LogLevel)} values via symbolic names
 * @see {@link LogLevelConst} type.
```

packages/common/core-interfaces/src/logger.ts
```diff
@@ -75,13 +117,13 @@ export interface ITelemetryBaseLogger {
 	/**
 	 * Log a telemetry event, if it meets the appropriate log-level threshold (see {@link ITelemetryBaseLogger.minLogLevel}).
 	 * @param event - The event to log.
-	 * @param logLevel - The log level of the event. Default: {@link (LogLevel:variable).default}.
+	 * @param logLevel - The log level of the event. Default: {@link (LogLevel:variable) | LogLevel.info}.
```
jason-ha: These appear to be unchanged at this time. Does `{@link (LogLevel:variable).default` not work anymore?

MarioJGMsoft: It doesn't, but I decided to change it to point to LogLevelConst, because I realized that once the docs are actually processed, they might not point to the right documentation, so pointing it directly to LogLevelConst seems like the right move here

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: Applied changes

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: I tried communicating that through a new `@remarks` block

packages/common/core-interfaces/src/logger.ts
jason-ha: In the prior reviews there were notes about avoiding references to `LogLevelConst` because that is a type and not the thing that anyone should really reference.
Why can't this be `(variable:LogLevel).info` or `(variable:LogLevel).default` as it previously was?

jason-ha (commented):
Except for line 97 of logger.ts there should be no TS Doc references to `LogLevelConst`. Use `(variable:LogLevel)` or LMK why that doesn't work.

packages/common/core-interfaces/src/logger.ts
jason-ha: We don't want LogLevelConst to be referenced if we can help it. Do generate the docs and try it out.

MarioJGMsoft:
> Except for line 97 of logger.ts there should be no TS Doc references to `LogLevelConst`. Use `(variable:LogLevel)` or LMK why that doesn't work.


It doesn't work anymore because I checked the docs and even if I use `{@link (LogLevel:variable)| LogLevel.default}`, it now throws an error when building saying that it's unable to resolve.

packages/common/core-interfaces/src/logger.ts
```diff
@@ -52,13 +52,58 @@ export interface ITelemetryBaseEvent extends ITelemetryBaseProperties {
...
+
+	/**
+	 * Essential information about the operation of Fluid.
+	 * @remarks It is recommended that these should always be collected, even in production, for diagnostic purposes.
+	 */
+	readonly essential: 30;
+
+	/**
+	 * Default LogLevel
+	 * @remarks Prefer {@link LogLevelConst.info} when selecting a level explicitly since this will be deprecated and removed in a future release.
```
jason-ha: + `| LogLevel.info` here and line 90

jason-ha (approved):

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: Applied change

jason-ha (approved):

PR #27021, opened 2026-04-13
feat(container-runtime): flatten staging mode API 

sonalideshpandemsft:
This change removes `StageControlsAlpha` and `StageControlsInternal` interfaces entirely, flattening the staging mode API so that `enterStagingMode()` returns `void` and consumers call `exitStagingMode(action)` directly on the runtime.

Previously, entering staging mode returned a controls object:
```ts
const controls = runtime.enterStagingMode();
controls.commitChanges(); // or controls.discardChanges();
```

After this change, the API is flat:
```ts
runtime.enterStagingMode();
runtime.exitStagingMode("commit"); // or "discard"
```


packages/runtime/container-runtime-definitions/src/containerRuntime.ts
```diff
@@ -204,6 +204,51 @@ export interface IContainerRuntime
...
+	enterStagingMode(): void;
+
+	/**
+	 * Exit staging mode and either commit or discard the staged changes.
+	 *
+	 * @param action - `"commit"` sends the buffered ops to the ordering service.
+	 * `"discard"` rolls back all changes made while in staging mode.
+	 * @throws If not currently in staging mode.
+	 */
+	exitStagingMode(action: "commit" | "discard"): void;
```
anthony-murphy: i would take an object union {action: "commit""} | {action:"discard"} this maintains strong typing but allows for extensitibily, like squashing for commit

PR #27041, opened 2026-04-14
test(tree): add buildRegistry to compat test config and validate custom factories in convertRegistry, tests for incremental summary bug

justus-camp-microsoft:
## Description

In cross-version compat tests, `convertRegistry` remaps channel factories by type string. When a test uses a custom/configured factory (e.g. `configuredSharedTree`),
  the remap silently drops the custom configuration — the test runs with a default factory without any indication that something is wrong.

  This PR:
  - Adds `ICompatTestContainerConfig` with a `buildRegistry` callback that receives the per-side data-runtime API, letting tests with custom factory config produce
  correct factories without going through the type-based remap.
  - Adds a constructor identity check in `convertRegistry` that throws when a non-default factory is passed via the legacy `registry` field, instead of silently dropping
  config.
  - Passes `CompatApis` to `describeInstallVersions` test suites, matching `describeCompat` behavior.
  - Migrates several tests (`compression`, `entryPointCompat`, `treeCompat`, `legacyChunking`) to pull factories from `apis.dds`/`apis.dataRuntime` instead of direct
  imports.
  - Adds `summarizeIncrementalSharedTree.spec.ts` — validates that SharedTree incremental summary handle paths remain valid across multiple summaries, using
  `buildRegistry` with `configuredSharedTree`.

  ## Reviewer Guidance

  The core logic change is in `compatUtils.ts` — the constructor identity check in `convertRegistry` and the new `buildRegistry` path. The test file changes are mostly
  mechanical (switching from direct imports to `apis.dds`). The new `summarizeIncrementalSharedTree.spec.ts` is the motivating test that requires `buildRegistry` with
  `configuredSharedTree`.'

### DIFF
--- a/packages/test/test-version-utils/src/compatUtils.ts
+++ b/packages/test/test-version-utils/src/compatUtils.ts
@@ -105,9 +105,30 @@ export interface ITestDataObject extends IFluidLoadable {
 	_root: ISharedDirectory;
 }
 
+/**
+ * Extension of {@link ITestContainerConfig} that lets a test build its channel factory registry
+ * from a per-side data-runtime api rather than passing a pre-built one.
+ *
+ * @remarks
+ * Prefer `buildRegistry` over `registry` when the registry contains factories that encode more
+ * than a DDS kind (e.g. `configuredSharedTree`, which bakes options into the factory class). In
+ * cross-client compat tests the framework invokes the builder once per side with that side's
+ * api, producing per-version factories directly and bypassing the type-based remap in
+ * `convertRegistry`. This prevents the silent-drop footgun where custom factory configuration
+ * gets discarded when the load side remaps by type string alone.
+ *
+ * `registry` and `buildRegistry` are mutually exclusive; setting both will throw.
+ *
+ * @internal
+ */
+export interface ICompatTestContainerConfig extends ITestContainerConfig {
+	buildRegistry?: (
+		dataRuntime: ReturnType<typeof getDataRuntimeApi>,
+	) => ChannelFactoryRegistry;
+}
+
 function createGetDataStoreFactoryFunction(
 	api: ReturnType<typeof getDataRuntimeApi>,
-	isCrossClientLoadPath = false,
 ): (containerOptions?: ITestContainerConfig) => IFluidDataStoreFactory {
 	class TestDataObject extends api.DataObject implements ITestDataObject {
 		public get _context(): IFluidDataStoreContext {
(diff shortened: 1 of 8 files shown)


packages/test/test-version-utils/src/compatUtils.ts
```diff
@@ -105,6 +105,24 @@ export interface ITestDataObject extends IFluidLoadable {
...
 }
 
+/**
+ * Maps DDS type string to the constructor of the current version's default factory.
+ * `makeSharedObjectKind` creates a new class per `SharedObjectKind`, so
+ * custom-configured variants have a different constructor than the default. The identity
+ * check in {@link convertRegistry} uses this to let those factories pass through
+ * without being swapped to the compat version.
+ */
+const currentDefaultFactoryCtors: Map<string, unknown> = new Map();
```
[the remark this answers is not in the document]
justus-camp-microsoft: I think `unknown` is better here

packages/test/test-version-utils/src/compatUtils.ts
```diff
@@ -139,7 +157,14 @@ function createGetDataStoreFactoryFunction(
...
 				throw Error(`Invalid or unimplemented channel factory: ${factory.type}`);
 			}
-			oldRegistry.push([key, oldFactory]);
+			const currentDefaultCtor = currentDefaultFactoryCtors.get(factory.type);
+			if (currentDefaultCtor !== undefined && factory.constructor === currentDefaultCtor) {
+				// Factory matches a known current-version default — convert to compat version.
+				oldRegistry.push([key, oldFactory]);
+			} else {
+				// Custom-configured factory — preserve as-is.
+				oldRegistry.push([key, factory]);
```
agarwal-navin: This should work fine. However, the disadvantage here is that any test using a custom factory won't be testing compat. This is problematic because the test authors may assume that since its running in different compat modes, they are getting coverage but they are actually not (this has happened in the past).

agarwal-navin: I think we should remove this workaround altogether and use the pattern where the dds factory is retrieved from the `apis: CompatApis` that is provided as part of the `describeCompat` block.

agarwal-navin: Most of the tests have already been converted to use the above pattern. However, there may be some left so updating them, removing the registry conversation should be an alternative.
It would also be great if we can instead add validation that the version of the factory matches the current compat version the test is running in so that tests don't accidentally use the non-comat APIs.

PR #27043, opened 2026-04-14
chore: move pnpm overrides from package.json to pnpm-workspace.yaml

alexvy86:
## Description

pnpm v10 supports `overrides` natively in `pnpm-workspace.yaml`, which is their canonical location going forward. This moves all `pnpm.overrides` entries and their associated comment arrays from `package.json` to the corresponding `pnpm-workspace.yaml` across all 14 workspaces.

- YAML comments replace the JSON comment arrays (`comments`, `commentsOverrides`, `overrideComments`, `overridesComments`) so override documentation is preserved inline next to each override
- Where `onlyBuiltDependencies` was already duplicated between `package.json` and `pnpm-workspace.yaml`, the `package.json` copy is removed since the yaml is the authoritative source
- Each workspace was independently validated with `pnpm install --frozen-lockfile` — all no-ops, no lockfile changes

28 files changed across 14 workspaces (14 `package.json` + 14 `pnpm-workspace.yaml`), 298 override entries moved total.

## Reviewer Guidance

This is a pure config migration — no code changes. The best way to verify is to confirm `pnpm install --frozen-lockfile` is a no-op in each workspace after the change.


alexvy86:
I'd like to move forward with this to get better support for comments for overrides, but right now Syncpack doesn't support looking at versions in `pnpm-workspace.yaml`, so we would be losing that bit of validation, so it's not a trivial trade-off.

Issues tracking this or related functionality:
- https://github.com/JamieMason/syncpack/issues/304
- https://github.com/JamieMason/syncpack/issues/231

PR #27045, opened 2026-04-14
test(client-test-version--utils): import SharedTee via /internal

jason-ha:
- add preferred import entrypoint to loading logic
- set "./internal" as tree packages preferred import which preserves prior behavior of runtime imports. (In 2.92.0 tree removed the non-internal runtime APIs from non-internal entrypoints.)

[AB#69118](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/69118)

### DIFF
--- a/packages/test/test-version-utils/src/testApi.ts
+++ b/packages/test/test-version-utils/src/testApi.ts
@@ -73,7 +73,7 @@ export interface PackageToInstall {
 	/**
 	 * Entrypoint to load from, if available. Otherwise, root entrypoint will be used.
 	 */
-	preferredEnrypoint?: "." | `./${string}`;
+	preferredEntrypoint?: "." | `./${string}`;
 }
 
 // List of driver API packages to install.
@@ -107,7 +107,7 @@ const dataRuntimePackageEntries = [
 	{ pkgName: "@fluidframework/register-collection", minVersion: "0.56.0" },
 	{ pkgName: "@fluidframework/sequence", minVersion: "0.56.0" },
 	{ pkgName: "@fluidframework/agent-scheduler", minVersion: "0.56.0" },
-	{ pkgName: "@fluidframework/tree", minVersion: "2.0.0", preferredEnrypoint: "./internal" },
+	{ pkgName: "@fluidframework/tree", minVersion: "2.0.0", preferredEntrypoint: "./internal" },
 ] satisfies PackageToInstall[];
 
 /**
@@ -248,7 +248,7 @@ async function loadIfCompatible(
 ): Promise<any> {
 	// Check if the requested version satisfies the minVersion requirement
 	if (semver.gte(versionToInstall, pkgEntry.minVersion)) {
-		return loadPackage(modulePath, pkgEntry.pkgName, pkgEntry.preferredEnrypoint);
+		return loadPackage(modulePath, pkgEntry.pkgName, pkgEntry.preferredEntrypoint);
 	}
 	return undefined;
 }
--- a/packages/test/test-version-utils/src/testApi.ts
+++ b/packages/test/test-version-utils/src/testApi.ts
@@ -59,6 +59,7 @@ import {
 export interface PackageToInstall {
 	/** The name of the package to install. */
 	pkgName: string;
+
 	/**
 	 * The minimum version where the package should be installed.
 	 * If the requested version is lower than this, the package will not be installed. This enables certain level of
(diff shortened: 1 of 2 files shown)


packages/test/test-version-utils/src/testApi.ts
```diff
@@ -101,20 +107,20 @@ const dataRuntimePackageEntries: PackageToInstall[] = [
 	{ pkgName: "@fluidframework/register-collection", minVersion: "0.56.0" },
 	{ pkgName: "@fluidframework/sequence", minVersion: "0.56.0" },
 	{ pkgName: "@fluidframework/agent-scheduler", minVersion: "0.56.0" },
-	{ pkgName: "@fluidframework/tree", minVersion: "2.0.0" },
-];
+	{ pkgName: "@fluidframework/tree", minVersion: "2.0.0", preferredEnrypoint: "./internal" },
```
Abe27342: fine with this as-is, but I feel like we probably want `./legacy` here (and in maybe all of our DDSes) as e2e test package is largely written against the legacy API surface, and people probably don't want to be dealing with things marked internal (and likely subject to more change) in a compat-aware fashion.

Abe27342 (approved):

packages/test/test-version-utils/src/testApi.ts
jason-ha: I went with /internal as that is what we import from current. :/

Abe27342: yeah totally understand picking `internal`--it matches prior behavior and from what I know entrypoint behavior of our other packages at runtime. Just saying I feel like better state long-term would be for e2e tests to just have access to legacy API and not internal. That's a bit more aspirational though and I expect probably opens up some other cans of worms :)

packages/test/test-version-utils/src/testApi.ts
```diff
@@ -243,7 +248,7 @@ async function loadIfCompatible(
 ): Promise<any> {
 	// Check if the requested version satisfies the minVersion requirement
 	if (semver.gte(versionToInstall, pkgEntry.minVersion)) {
-		return loadPackage(modulePath, pkgEntry.pkgName);
+		return loadPackage(modulePath, pkgEntry.pkgName, pkgEntry.preferredEnrypoint);
```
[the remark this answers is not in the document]
jason-ha: This doesn't seem to be a problem.

packages/test/test-version-utils/src/testApi.ts
agarwal-navin: Can't we make all packages use the internal or alpha entry point? That way we can avoid this new parameter. We can add it if some package doesn't really want to do that in the tests but we don't need it for now.
Unless, not all packages support different entry points.

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -415,21 +419,29 @@ export const loadPackage = async (modulePath: string, pkg: string): Promise<any>
...
 			primaryExport = pkgJson.exports;
 		} else {
-			const exp: any | undefined = pkgJson.exports["."];
+			const exp: any | undefined = pkgJson.exports[importPath] ?? pkgJson.exports["."];
 			primaryExport =
 				typeof exp === "string"
 					? exp
 					: exp.require !== undefined
 						? exp.require.default
 						: exp.default;
```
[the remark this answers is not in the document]
jason-ha: Yes, this whole pattern should go away. :)

packages/test/test-version-utils/src/testApi.ts
```diff
@@ -68,28 +69,33 @@ export interface PackageToInstall {
 	 * versions prior to that, the package will not be installed and the test should skip testing these versions.
 	 */
 	minVersion: string;
+
+	/**
+	 * Entrypoint to load from, if available. Otherwise, root entrypoint will be used.
+	 */
+	preferredEnrypoint?: "." | `./${string}`;
```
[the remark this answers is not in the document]
jason-ha: To fix in follow-up, or here if another change is required.

packages/test/test-version-utils/src/testApi.ts
jason-ha: Not all packages support all entrypoints.
It is best if most packages don't have /internal exports.
Though one could be forgiven for aggressive use of /internal throughout the codebase because our lint rules.
We really need to establish better rules for layers and entrypoints. Use of /internal across layers is tricky, particularly when we don't check for type compatibility for `@internal` changes.
I'd be curious to hear why `tree` has any non-test `@internal` APIs.

I have zero opposition to transitioning to /legacy or other non-/internal entrypoints. Just think that should be an explicit revision apart from this repair work.

alexvy86 (approved):

PR #27049, opened 2026-04-15
Report Mocha errors in becnmarks

CraigMacomber:
## Description

Error messages from Mocha, including timouts, are now included to the console and output files.

## Reviewer Guidance

### DIFF
--- a/tools/benchmark/src/test/mocha/integration.spec.ts
+++ b/tools/benchmark/src/test/mocha/integration.spec.ts
@@ -24,49 +24,54 @@ const sampleResult: CollectedData = [
 const perfModeArgs = ["--perfMode", "--reporter", "./dist/mocha/Reporter.js"];
 
 describe("mocha integration", () => {
-	function integrationTest(testArgs: string[], shouldFail: boolean): string {
+	function integrationTest(
+		testArgs: string[],
+		shouldFail: boolean,
+		options?: {
+			/** Defaults to "mocha-integration-inner". */
+			testName?: string;
+			env?: NodeJS.ProcessEnv;
+		},
+	): { stdout: string; stderr: string } {
 		// Disable colors with "--no-color" does not seem to work on CI (perhaps due to an environment variable),
 		// but enabling colors does seem to work locally, so we enable colors to get consistent results for these tests.
-		const args = [...testArgs, "--fgrep", "mocha-integration-inner", "--color"];
-		if (shouldFail) {
-			args.push("--integrationFail");
-		}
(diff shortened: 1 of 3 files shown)


tools/benchmark/src/test/mocha/integration.spec.ts
```diff
@@ -24,49 +24,54 @@ const sampleResult: CollectedData = [
...
-		const args = [...testArgs, "--fgrep", "mocha-integration-inner", "--color"];
-		if (shouldFail) {
-			args.push("--integrationFail");
-		}
-		const result = childProcess.spawnSync("mocha", args, { encoding: "utf8" });
+		const testName = options?.testName ?? "mocha-integration-inner";
+		const args = [...testArgs, "--fgrep", testName, "--color"];
+		const result = childProcess.spawnSync("mocha", args, {
+			encoding: "utf8",
+			env: options?.env,
```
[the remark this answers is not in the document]
CraigMacomber: Making these tests less dependent on how they were run is prabably an improvement not a regression here (assuming they do still pass), so this seems fine.

tools/benchmark/src/test/mocha/integration.spec.ts
```diff
@@ -81,6 +86,43 @@ describe("mocha integration", () => {
...
+			true,
+			{
+				testName: "timeout-integration-inner",
+				// FLUID_TEST_VERBOSE prevents @fluid-internal/mocha-test-setup from suppressing
+				// console.error during test execution, so the timeout error reaches stderr.
+				env: { ...process.env, FLUID_TEST_VERBOSE: "1" },
+			},
+		);
+		// The table in stdout truncates the error, so check stderr:
+		assert.match(stderr, /Timeout of 1ms exceeded/);
```
alexvy86, line 110: Isn't the point of the test to validate that the timeout error makes it through the custom reporter?

alexvy86 (commented):
The feature seems fine to me but I'm not sure the test accomplishes what it should.

tools/benchmark/src/test/mocha/integration.spec.ts
CraigMacomber: yes. This validates that the timeout error is reported to stderr. Before this change, the error message made no mention of the timeout, now it does, enabling this test to pass.

alexvy86: Ah, we're validating the code path that writes the error message to stderr, I thought we wanted to ensure that the error in the output table also talked about the timeout, but I guess the truncation you mention above complicates that. 👍

alexvy86 (approved):

PR #27054, opened 2026-04-15
Modernize benchmarks

CraigMacomber:
## Description

Avoid deprecated benchmark API, and reduce use of mutable statics in describe blocks.

## Reviewer Guidance


experimental/dds/tree/src/test/Forest.perf.tests.ts
```diff
@@ -26,63 +26,56 @@ describe('Forest Perf', () => {
...
 			title: `${count} random inserts in TreeView`,
-			benchmarkFn: () => {
-				buildRandomTree(testTree, count);
-			},
+			run: async () =>
+				collectDurationData({
+					benchmarkFn: () => {
+						buildRandomTree(testTree, count);
+					},
+				}),
```
alexvy86: Does this one really need the extra boilerplate? Can't it use `...benchmarkDuration()`?

experimental/dds/tree/src/test/AppendOnlySortedMap.perf.tests.ts
```diff
@@ -30,21 +30,29 @@ function runAppendOnlyMapPerfTests(mapBuilder: () => AppendOnlySortedMap<number,
...
 	};
 
-	benchmark({
+	benchmarkIt({
 		type,
 		title: `lookup a key`,
-		before,
-		benchmarkFn: () => {
-			map.get(keyChoices[localChoice++ % keyChoices.length]);
+		run: async () => {
```
Josmithr, line 34: When should someone writing a benchmark prefer this (`collectDurationData`) pattern over the `...benchmarkDuration(...)` pattern used in tests above? Is this needed because of the `setup` function called before we begin measurements? Does `benchmarkDuration` not have a `before` hook?

CraigMacomber: It doesn't really matter.

If you can use benchmarkDuration and benchmarkFn or benchmarkFnAsync, its likely simpler to do so.

If you need some before code, but no per iteration stuff, collectDurationData + benchmarkFn is likely easiest but benchmarkDuration and benchmarkFnCustom also work: it doesn't really matter.

If you want to control the inner timing loop and do per iteration or per patch stuff you need to use benchmarkFnCustom, which can be used with collectDurationData or benchmarkDuration.

If you want access to the measurements to alter or inject your own, then you must use collectDurationData.

CraigMacomber: I deprecated all the hooks since they are a crappy pattern (bad for type safety) and there is at least one good way to do everything without using them now (often more than one way).

CraigMacomber: The built in `setup` support is only a thing in the memory test helpers btw. We have multiple ways to do everything without needing it for duration tests (as shown here)

CraigMacomber: This file is still using the Side effects anti pattern though:

setup should return stuff which the benchmarkFn closes over... let me fix that

Josmithr: The main reason that I ask is that it seems like it would be generally preferable to do things a single way consistently across the repo (as much as possible).

I personally find this pattern easier to parse than the `...benchmarkDuration(...)` pattern, likely even for the cases that don't need any before/after logic or anything else. It also makes it more obvious how you would _modify_ the benchmark if you need to add before/after logic (where with the other pattern you would have to deconstruct the helper function call to do so).

I'd be interested in others' thoughts on this though.

CraigMacomber: Oh, one thing I missed (and Claude missed). I hadn't really intended this pattern to get used much, just left it as a possibility, and forgot that doing it this way fails to tag the benchmark as a duration benchmark. Gotta update everything using this pattern to do better tagging

CraigMacomber: Fixed.

I'll also make some tweaks to the benchmark tool in a seperate change to make it easier to pick a pattern that doesn't need that, then we can make the tests more consistant.

packages/dds/merge-tree/src/test/Insertion.perf.spec.ts
```diff
@@ -26,78 +26,95 @@ function constructTree(numOfSegments: number): MergeTree {
...
+	benchmarkIt({
+		type: BenchmarkType.Measurement,
+		title: "insert at start of large tree",
+		...benchmarkDuration({
+			benchmarkFnCustom: async (state) => {
+				let startTree = constructTree(TREE_SIZE);
+				let running: boolean;
+				do {
+					startTree = constructTree(TREE_SIZE);
+					running = state.timeBatch(() => {
```
alexvy86: Nit (needs the `while (running)` updated too). I think it makes the flow a bit easier to understand.
```suggestion
					keepRunning = state.timeBatch(() => {
```

packages/dds/tree/src/test/feature-libraries/chunked-forest/uniformChunk.spec.ts
```diff
@@ -200,43 +205,48 @@ describe("uniformChunk", () => {
...
+							benchmarkFn: () => {
+								sum(cursor);
+							},
+						});
 					},
 				});
 			}
 
-			benchmark({
+			benchmarkIt({
```
alexvy86, line 223: Since it doesn't use `...benchmarkDuration` this one wants a `testType`, right?

packages/dds/merge-tree/src/test/Removal.perf.spec.ts
```diff
@@ -16,175 +20,180 @@ import {
 import { TestString, loadSnapshot } from "./snapshot.utils.js";
 
 describe("MergeTree remove", () => {
-	let summary: ISummaryTree;
-
-	benchmark({
+	benchmarkIt({
```
alexvy86, line 23: This file seems to be missing `testType` in the new versions.

experimental/dds/tree/src/test/AppendOnlySortedMap.perf.tests.ts
alexvy86: I'll say that like Josh, I'm not the biggest fan of the `...benchmarkDuration()` pattern, but do appreciate that it tags the test as execution time instead of having to do it manually, particularly because that property `testType` is not required and is easy to miss. If we made it required then I think I would also prefer the more manual pattern here; but not sure if we then make it `testType: TestType | string` or something along those lines so tests that are neither execution time nor memory can be defined.

packages/runtime/id-compressor/src/test/idCompressor.perf.spec.ts
```diff
@@ -111,127 +112,148 @@ describe("IdCompressor Perf", () => {
...
 
 	for (const clusterSize of [1, 10, 500, 1000]) {
 		const numIds = 7;
 		const session1 = "8150a099-5302-4672-b5f3-7a4492b59418" as SessionId;
 		const session2 = "f2ded886-92da-4248-967b-eb96ee04cf51" as SessionId;
-		let session: SessionId = session1;
-		let nextFirstFinalizedGenCount1 = 1;
-		let nextFirstFinalizedGenCount2 = 1;
-		benchmark({
+		benchmarkIt({
```
alexvy86, line 148: Missing `testType` here and in a few more below.

experimental/dds/tree/src/test/IdCompressor.perf.tests.ts
```diff
@@ -120,139 +125,159 @@ describe('IdCompressor Perf', () => {
 			let lastFinalizedLocalId1 = 0 as LocalCompressedId;
 			let lastFinalizedLocalId2 = 0 as LocalCompressedId;
 			let overrideIndex = 0;
-			benchmark({
+			benchmarkIt({
```
alexvy86, line 128: This file also has several missing `testType`s.

experimental/dds/tree/src/test/Forest.perf.tests.ts
CraigMacomber: Fixed.

experimental/dds/tree/src/test/IdCompressor.perf.tests.ts
CraigMacomber: Fixed.

experimental/dds/tree/src/test/AppendOnlySortedMap.perf.tests.ts
CraigMacomber: https://github.com/microsoft/FluidFramework/pull/27092 should help cleanup the `...benchmarkDuration` cases.

I'm also open to if someone want to do a follow-up to try and further simplify the API as long as we keep the mocha specific API surface minimal (benchmarkDuration was introduced to provide a non-mocha specific way to express the duration specific portion of the old benchmark API).

Note that currently this testType tag is only used for clarity of what its testing in the generated test name, and for splitting the CI test run. Someday I'd like to some more specific to duration tests for more stable data (like tweak the CPU settings to disable boost, reduce concurrency etc), but right now its not really very important.


When I redid the benchmark tool package, I didn't have a great sense of how client used testType, so I mostly left it alone, but I think we might want to just remove it instead have a more general tagging system. Really all I plan to use test type for now is flagging which tests are sensitive to runtime, which would be better done with a boolean flag rather that a type enum.

Having flags for runtime sensitivity, requiring Node's GC API etc would make more sense.

alexvy86 (approved):
Looks good from my end. Maybe @Josmithr has further thoughts.

Abe27342 (approved):
Merge-tree and id-compressor look good.

Josmithr (approved):

PR #27055, opened 2026-04-15
ci(codespaces): add prebuild drift detection for setup script changes

tylerbutler:
## Description

Codespace prebuilds in "on configuration change" mode only watch `devcontainer.json` and the `Dockerfile` — not the setup scripts they reference (`scripts/codespace-setup/`). This means setup script changes silently drift, leaving users with stale prebuilt environments.

This PR adds:

1. **`// prebuild-version: N` comments** in each `devcontainer.json`. Bumping the version forces a byte change that triggers a prebuild rebuild.
2. **A CI workflow** (`devcontainer-prebuild-check.yml`) that fails PRs touching `scripts/codespace-setup/**` without also modifying a `devcontainer.json` or `Dockerfile`.

The convention is documented in `.devcontainer/README.md`.

## Reviewer Guidance

- The approach uses a JSONC comment rather than a JSON field — any byte change to `devcontainer.json` triggers the prebuild, so comments work fine.
- The CI workflow is lightweight and read-only, following the same pattern as other validation workflows.
- No breaking changes or code changes — only devcontainer config, a CI workflow, and docs.

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -35,8 +35,19 @@ The devcontainer lifecycle hooks are structured for prebuild optimization:
 | --- | --- | --- |
 | `onCreateCommand` | Node.js setup (nvm install, corepack enable) | Yes |
 | `postCreateCommand` | Welcome notice (all profiles); AI tooling setup (AI-enabled profiles) | Yes |
+| `postStartCommand` | bwrap/sandbox setup (AI profiles only) | No |
 
 Heavy setup work runs in `onCreateCommand` so it is captured by Codespace prebuilds, making the user-connect experience faster.
+
+## Prebuild configuration
+
+Prebuilds use the **"on configuration change"** trigger, which only fires when `devcontainer.json` or the referenced `Dockerfile` changes. It does **not** detect changes to files those configs depend on, such as scripts in `scripts/codespace-setup/`.
+
+### The `prebuild-version` comment
+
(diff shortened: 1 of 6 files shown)


.devcontainer/README.md
```diff
@@ -35,9 +35,20 @@ The devcontainer lifecycle hooks are structured for prebuild optimization:
 | --- | --- | --- |
 | `onCreateCommand` | Node.js setup (nvm install, corepack enable) | Yes |
 | `postCreateCommand` | Welcome notice (all profiles); AI tooling setup (AI-enabled profiles) | Yes |
+| `postStartCommand` | bwrap/sandbox setup (AI profiles only) | No |
```
tylerbutler, line 38: Is this addition correct?

.devcontainer/README.md
```diff
@@ -35,9 +35,20 @@ The devcontainer lifecycle hooks are structured for prebuild optimization:
...
 
 Heavy setup work runs in `onCreateCommand` so it is captured by Codespace prebuilds, making the user-connect experience faster.
 
+## Prebuild configuration
+
+Prebuilds use the **"on configuration change"** trigger, which only fires when `devcontainer.json` or the referenced `Dockerfile` changes. It does **not** detect changes to files those configs depend on, such as scripts in `scripts/codespace-setup/`.
+
+### The `prebuild-version` comment
+
+Each `devcontainer.json` contains a `// prebuild-version: N` comment. **Bump this value whenever you change files in `scripts/codespace-setup/`** — this ensures the devcontainer.json file is modified, which triggers a prebuild rebuild.
```
tylerbutler: Worth noting that oif you make other changes to the dev container config anyway, then you don't need to bump the version necessarily. the field exists to help do necessary but not version-relevant changes.

jumyhre (approved):

PR #27072, opened 2026-04-16
Tweaks to improve error output for benchmarks

CraigMacomber:
## Description

Slightly nicer error output for benchmarks from reporter.

## Reviewer Guidance

### DIFF
--- a/tools/benchmark/src/mocha/Reporter.ts
+++ b/tools/benchmark/src/mocha/Reporter.ts
@@ -97,18 +97,20 @@ module.exports = class {
 
 				let benchmark: BenchmarkResult | undefined = this.testData.get(test.id)?.data;
 
-				if (!("on" in test)) {
-					// In parallel mode, we can not subscribe to events on the test,
-					// but the event we are in is delayed until after the test ran so we can get the results off of it.
-					// To make this work, the emit code (emitResultsMocha) crammed the results in the test body, so parse that.
+				if (benchmark === undefined) {
+					if (!("on" in test)) {
+						// In parallel mode, we can not subscribe to events on the test,
+						// but the event we are in is delayed until after the test ran so we can get the results off of it.
+						// To make this work, the emit code (emitResultsMocha) crammed the results in the test body, so parse that.
 
-					// The if above narrows test to `never` here, so undo that:
-					const test2 = test as Test;
-					const body = test2.body;
-					try {
-						benchmark = parseBenchmarkResult(body);
-					} catch {
-						// If the body isn't json, then the event was not put into the body, and so treat it like no data was reported.
+						// The if above narrows test to `never` here, so undo that:
+						const test2 = test as Test;
+						const body = test2.body;
+						try {
+							benchmark = parseBenchmarkResult(body);
+						} catch {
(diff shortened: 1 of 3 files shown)


tools/benchmark/src/mocha/Reporter.ts
```diff
@@ -97,18 +97,20 @@ module.exports = class {
...
-					// To make this work, the emit code (emitResultsMocha) crammed the results in the test body, so parse that.
-
-					// The if above narrows test to `never` here, so undo that:
-					const test2 = test as Test;
-					const body = test2.body;
-					try {
-						benchmark = parseBenchmarkResult(body);
-					} catch {
-						// If the body isn't json, then the event was not put into the body, and so treat it like no data was reported.
+				if (benchmark === undefined) {
```
CraigMacomber, line 100: For parallel mode, if we had an error already, don't overwrite it. I'm not aware of this happening in any cases, but this change makes it more robust and works better with the recent fix for timeout errors.

tools/benchmark/src/mocha/Reporter.ts
```diff
@@ -125,12 +127,10 @@ module.exports = class {
...
 
 				if (benchmark === undefined) {
-					// Mocha test completed without emitting benchmark data and without a recorded failure.
+					// Mocha test completed without emitting benchmark data and without a recorded error.
 					// This can happen if the test passed without calling emitResultsMocha.
 					benchmark = {
-						error: `Test ${test.fullTitle()} completed with status '${
-							test.state
-						}' without reporting any data or an error.`,
+						error: `Test completed with status '${test.state}' without reporting any data or an error.`,
```
CraigMacomber, line 133: The benchmark result when displayed is already paired with info about which test it was, so including the title here was redundant and tended to get the real error truncated out of the results table.

tools/benchmark/src/test/mocha/integration.spec.ts
```diff
@@ -39,7 +39,7 @@ describe("mocha integration", () => {
 		const args = [...testArgs, "--fgrep", testName, "--color"];
 		const result = childProcess.spawnSync("mocha", args, {
 			encoding: "utf8",
-			env: options?.env,
+			env: { ...process.env, ...options?.env },
```
CraigMacomber, line 42: we need the existing env for things like path, so this now overrides members of it instead of replacing it.

tools/benchmark/src/reporterUtilities.ts
```diff
@@ -132,8 +132,8 @@ export function reportTable(heading: string, reports: readonly ReportEntry[]): s
 		if (isResultError(result)) {
 			table.cell("Status", status(0, 1));
 			table.cell("Name", name);
-			// Full error should be included outside of table, so limit text in table to avoid breaking formatting.
-			const errorColumns = 50;
+			// Full error should be included outside of table (in standard error), so limit text in table to avoid breaking formatting.
+			const errorColumns = 60;
```
CraigMacomber, line 136: In practice this often ended up being annoyingly short, and as our test config throws away standard err by default (a questionable choice), including a bit more here is nice.

alexvy86 (approved):

PR #27077, opened 2026-04-17
Cleanup benchmark deps, including updating mocha to 11

CraigMacomber:
## Description

Update mocha to 11, and cleanup unneeded deps.

## Reviewer Guidance


tools/benchmark/pnpm-workspace.yaml
```diff
@@ -21,5 +21,23 @@ trustPolicy: no-downgrade
...
 # package (including later major versions) which had better provenance information.
 # ALWAYS REVIEW CAREFULLY BEFORE ADDING SOMETHING TO THIS LIST.
-trustPolicyExclude: []
+trustPolicyExclude:
+  # chokidar@4.0.3 is a transitive dependency of mocha@11.x. The trust downgrade fires
+  # because this version lacks provenance attestation while earlier 4.x releases had it.
+  # No supply chain incident is known.
+  # This version is heavily used (~40 million downloads) and a year old at this point,
+  # and also used by other workspaces in this repo, so allowing it seems like a minimal risk.
+  - "chokidar@4.0.3"
```
Josmithr, line 30: FWIW, I ran into this when I was updating mocha in eslint-config-fluid - I was able to pnpm override this to `^5` without issue. Don't know if that will prove true here or not, but might be worth a shot.

Josmithr (approved):
I left a small suggestion, and Copilot left some typo-fix suggestions. Otherwise looks good to me.

tools/benchmark/pnpm-workspace.yaml
CraigMacomber: the client lock file already uses 4 different copies, two of 4.0.2 and two 3.6.0, so I figured I'd leave it.

CraigMacomber: Mocha 12 does update to ^5 though, so we have that to look forward to.

alexvy86 (approved):

PR #27078, opened 2026-04-17
More support and use of Unified ServiceClient API

CraigMacomber:
## Description

See https://github.com/microsoft/FluidFramework/pull/27693 for initial creation of this API.

This adds a bunch of, mostly AI driver, implementations, integrations and adoption of this new API surface.

## Reviewer Guidance

This PR is a collection of use, which will be split up and merged independently. It serves as an example of how things like look end to end when its well integrated, but is not intended as a location for review of the details of the implementations included.


ChumpChief:
Quick note - it would be helpful to reviewers if you can point to a specific A/B comparison of what you're trying to improve (i.e. "before it would look like <code snippet> but now it's simpler like <code snippet>").  I see a lot of stuff moving around but it's hard to tell what specifically is expected to be improved.

tylerbutler:
I love this direction. I want to give it a longer look and maybe build a test app against it. Exciting stuff!

PACKAGES.md
```diff
@@ -65,11 +65,17 @@ The dependencies between layers are enforced by the layer-check command._
 | --- | --- |
 | - [@fluidframework/telemetry-utils](/packages/utils/telemetry-utils)</br>&nbsp;</br>&nbsp;</br>&nbsp; | - [Core-Interfaces](#Core-Interfaces)</br>- [Driver-Definitions](#Driver-Definitions)</br>- [Core-Utils](#Core-Utils)</br>- [Client-Utils](#Client-Utils) |
 
+### Runtime-Definitions
```
CraigMacomber: Changes to this file are due to changes in layerinfo.json: discuss them there instead of here.

packages/runtime/runtime-definitions/src/serviceClient.ts
```diff
@@ -0,0 +1,460 @@
...
+	IFluidDataStoreChannel,
+} from "./dataStoreContext.js";
+import type { IFluidDataStoreFactory } from "./dataStoreFactory.js";
+import { registryLookup, type Registry, type RegistryKey } from "./registry.js";
+
+/*
+ * This file defines the public API for the ServiceClient and related types.
+ *
+ * TODO:
+ * Currently this API surface expect all code using it together to be using a singly copy of the Fluid Framework client packages.
```
CraigMacomber: Am I correct that many (or maybe most) of our customers (especially future customers) shouldn't need to do the package version mixing/duplication that layer compat allows (And thus having a nice API like this without support for it has some real value)?

Also, for our customers that do rely on our layer compat story, what is the long term vision? Do we expect the layer compat for mixing/duplicating package versions to eventually to away (perhaps on the same timeline as migrating off legacy+beta APIs)? Or do we expect our non-legacy APIs to eventually provide support for the package version mixing as a prerequisite for apps getting off legacy APIs? Or perhaps such apps will be forever supported with legacy APIs for doing the layer version mixing, and we will try and make the divergence needed to dop that from non-legacy APIs as small as possible?

layerInfo.json
```diff
@@ -47,7 +47,13 @@
 			},
 			"Driver-Utils": {
 				"packages": ["@fluidframework/driver-utils"],
-				"deps": ["Protocol-Utils", "Driver-Definitions", "Telemetry-Utils", "Core-Utils"]
+				"deps": [
+					"Protocol-Utils",
+					"Driver-Definitions",
+					"Telemetry-Utils",
+					"Core-Utils",
+					"Runtime-Definitions"
```
CraigMacomber: I intentionally made a change to our layering.

Clearly, runtime should not depend on specific drivers: that would make no sense as no driver is used in all cases.

Therefor it seems reasonable to me to allow driver specific code to depend on runtime (Such a dependency could never cause a cycle, and since they are versioned together, it doesn't impact when what changes require a version bump.)

Allowing these extra dependencies make it practical to provide high level APIs like those in fluid-static and the service client packages, but do so in the driver layer.

Them preexisting fluid-static and service client style APIs at their existing locations in the dependency graph made them unusable in SharedObject packages (like SharedTree, SharedMap etc.) as doing so would create cycles.

The new setup makes it possible for SharedObjects to use this new customer facing APIs for their own tests with is both really nice for testing actual customer use cases, but also nice as it provides a nice API surface to use for testing which makes things like collab easy to test. It also ensures that our new customer facing API surface is good for writing tests, which is something our customers will need to so with it as well (this is a major usability gap in the preexisting public API surface, and even in the legacy one which motivated customers to use a lot of our test-utils, many of which can be obsoleted by this new API surface).

This all sounds like a big win to me, and justifies the little bit of extra coupling. I however do not have a great understanding of the implications of this with regard to layer compat. If our layer compat allows using drivers with a different version of the runtime (for example newer drivers with older runtime, or vise versa), then this additional coupling is likely introducing a lot of layer-compat risk, and a better solution would be to introduce new service-client versions of each driver which have these dependencies, and leave the existing driver packages unchanged. That would still not support mixing driver and runtime versions when using the new API, but it currently isn't designed to handle that anyway (See https://github.com/microsoft/FluidFramework/pull/27078/files#r3174454103 )

CraigMacomber: After reviewing our layer documentation, I'm pretty sure this change is no ok, and I'll need to take the approach of introducing new packages.

PR #27084, opened 2026-04-17
fix(tree): track chunkIndex on field entry so getField finds parent

brrichards:
# Summary

basicChunkCursor currently doesn't handle tracking the chunk index around multi-node chunks correctly. This problem shows up in chunkedForest, which extends this cursor. 

### Files changed
| File | Changes |
| ---- | ---- |
| `basicChunk.ts` | Change location of when `chunkIndex` is pushed onto the stack and `getField` now uses `chunkIndex` to find the parent in `siblingStack`|
| `basicChunk.spec.ts` | Added two unit tests that manually construct a field with a `UniformChunk` followed by a `BasicChunk` with a subfield | 

### What surfaces the bug
<pre>                                                                                                  
  Root field                                                                     
  ├─ chunks[0]  UniformChunk  (nodes 0, 1)
  │    ├─ 10                                                                                             
  │    └─ 20
  └─ chunks[1]  BasicChunk    (node 2)                                                                   
       └─ subField                                                               
            └─ 42     
   
siblingStack when getField is called after navigating to subField: [ UniformChunk(10,20) , BasicChunk "Trailing" ]              
 </pre>
When a field contains multi-node chunks followed by a `BasicChunk` containing a subfield, entering that subfield with `enterField` will call `getField` which searches for the parent of this subfield in the `siblingStack`. Currently this uses the flat node index, which equals the count of all nodes in the field. The problem is that `siblingStack` contains the chunks and not the flat nodes. So when `getField` is called using an index based off the flat nodes it will incorrectly index into the siblings array on that stack.  In the case above we would attempt to get the parent BasicChunk at index 2, but that would be `oob`. 

#### chunkIndex
The proposed changes are to swap the use of a flat node index for `getField` and use `chunkIndex`. This was being tracked before, but was a stale reference for what this fix requires. The flow for the current chunk index is: 
`start at 0 -> increment as we change chunks -> enterField(subField) -> getField -> enterNode/firstNode -> push chunkIndex onto the stack`
This causes `chunkIndex` to not be on the stack when we need it. We would be looking at a stale reference for 1 level up. The change needed was pushing `chunkIndex` onto the stack during enterField/firstField so we have the needed `chunkIndex` on the stack when `getField` needs it. 
#### halfHeight
halfHeight balances `indexOfChunkStack.length` against `siblingStack.length`. The current code pushed once per node level. The new code pushes once per Field level, so the rounding isn't needed like before. The assertion has been moved to its own private method `assertChunkStacksMatchNodeDepth()` since this is what the assertion was originally doing. Checking that the chunk stack indices were half the height of sibling stack, which would also correctly match the node depth. 

### New Test Coverage
Two unit tests in `basicChunk.spec.ts` that manually create a multi-node uniform chunk(not currently possible by the chunker) and then navigating into the subfield of a trailing basic chunk throws under the old indexing logic.

### DIFF
--- a/packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
+++ b/packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
@@ -555,14 +555,16 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 	}
 
 	/**
-	 * Returns the chunks that make up the field the cursor is currently in.
+	 * Resolves the chunks that make up the field the cursor is currently in. At the root, this is
+	 * {@link root} directly. Otherwise, the cursor must be in {@link CursorLocationType.Fields} mode,
+	 * and the result is looked up on the parent node using the current field key.
 	 *
-	 * @remarks At the root, this is {@link root} directly. Otherwise, the cursor must be in
-	 * {@link CursorLocationType.Fields} mode, and the result is looked up on the parent node using
-	 * the current field key. The parent node is the {@link BasicChunk} in the node array at the top
-	 * of {@link siblingStack} while we are in {@link CursorLocationType.Fields} mode. We need the
-	 * parent since a field's chunks are stored on the parent node's {@link BasicChunk.fields} map,
-	 * not on the cursor itself.
+	 * @remarks The parent node is the {@link BasicChunk} in the node array at the top of
+	 * {@link siblingStack} while we are in {@link CursorLocationType.Fields} mode. We need the parent
+	 * since a field's chunks are stored on the parent node's {@link BasicChunk.fields} map, not on
+	 * the cursor itself.
+	 *
+	 * @returns The chunks that make up the field the cursor is currently in.
 	 */
 	private getField(): readonly TreeChunk[] {
 		if (this.siblingStack.length === 0) {
--- a/packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
+++ b/packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
@@ -220,7 +220,15 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 		return (this.siblingStack[height] as readonly TreeChunk[])[index] as BasicChunk;
 	}
 
-	// Returns the floor of half the height passed in, or half of `siblingStack.length` if no height is provided.
+	/**
+	 * Converts a {@link height}, which contains field and node levels, into the corresponding depth/index
+	 * for the node-only stacks ({@link indexOfChunkStack} and {@link indexWithinChunkStack}), which are
+	 * only pushed on node-level transitions.
+	 *
+	 * @param height - A depth in {@link siblingStack} to convert. Defaults to {@link siblingStack}'s
+	 * current length, which gives the current depth of the node-only stacks.
+	 * @returns `floor(height / 2)` — the number of node levels at or below the given stack height.
+	 */
 	private getNodeOnlyHeightFromHeight(height: number = this.siblingStack.length): number {
 		// The bitwise shift computes the floor, which is valid assuming the depth is less than 2^31, which seems safe.
 		// eslint-disable-next-line no-bitwise
@@ -546,6 +554,16 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 		return (this.siblings as TreeChunk[])[this.indexOfChunk] as BasicChunk;
 	}
 
+	/**
+	 * Returns the chunks that make up the field the cursor is currently in.
+	 *
+	 * @remarks At the root, this is {@link root} directly. Otherwise, the cursor must be in
+	 * {@link CursorLocationType.Fields} mode, and the result is looked up on the parent node using
+	 * the current field key. The parent node is the {@link BasicChunk} in the node array at the top
+	 * of {@link siblingStack} while we are in {@link CursorLocationType.Fields} mode. We need the
+	 * parent since a field's chunks are stored on the parent node's {@link BasicChunk.fields} map,
+	 * not on the cursor itself.
+	 */
 	private getField(): readonly TreeChunk[] {
 		if (this.siblingStack.length === 0) {
 			return this.root;
--- a/packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
+++ b/packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { assert, oob, fail } from "@fluidframework/core-utils/internal";
+import { assert, oob, fail, debugAssert } from "@fluidframework/core-utils/internal";
 
 import {
 	CursorLocationType,
@@ -217,7 +217,9 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 
 	private getStackedChunk(height: number): BasicChunk {
 		const index = this.getStackedChunkIndex(height);
-		return (this.siblingStack[height] as readonly TreeChunk[])[index] as BasicChunk;
+		const chunk = (this.siblingStack[height] as readonly TreeChunk[])[index];
+		debugAssert(() => chunk instanceof BasicChunk || "only basic chunks are expected");
+		return chunk as BasicChunk;
 	}
 
 	/**
@@ -551,18 +553,15 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 
 	private getNode(): BasicChunk {
 		assert(this.mode === CursorLocationType.Nodes, 0x52f /* can only get node when in node */);
-		return (this.siblings as TreeChunk[])[this.indexOfChunk] as BasicChunk;
+		const chunk = (this.siblings as TreeChunk[])[this.indexOfChunk];
+		debugAssert(() => chunk instanceof BasicChunk || "only basic chunks are expected");
+		return chunk as BasicChunk;
 	}
 
 	/**
 	 * Resolves the chunks that make up the field the cursor is currently in. At the root, this is
 	 * {@link root} directly. Otherwise, the cursor must be in {@link CursorLocationType.Fields} mode,
 	 * and the result is looked up on the parent node using the current field key.
-	 *
-	 * @remarks The parent node is the {@link BasicChunk} in the node array at the top of
-	 * {@link siblingStack} while we are in {@link CursorLocationType.Fields} mode. We need the parent
-	 * since a field's chunks are stored on the parent node's {@link BasicChunk.fields} map, not on
-	 * the cursor itself.
 	 *
 	 * @returns The chunks that make up the field the cursor is currently in.
 	 */
@@ -574,6 +573,10 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 			this.mode === CursorLocationType.Fields,
 			0x530 /* can only get field when in fields */,
 		);
+		// The parent node is the `BasicChunk` in the node array at the top of
+		// `siblingStack` while we are in `CursorLocationType.Fields` mode. We need the parent
+		// since a field's chunks are stored on the parent node's `BasicChunk.fields` map, not on
+		// the cursor itself.
 		const parent = this.getStackedChunk(this.siblingStack.length - 1);
 		const key: FieldKey = this.getFieldKey();
 		const field = parent.fields.get(key) ?? [];
--- a/packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
+++ b/packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
@@ -169,7 +169,7 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 		// Compute the number of nodes deep the current depth is.
 		// We want the floor of the result, which can computed using a bitwise shift assuming the depth is less than 2^31, which seems safe.
 		// eslint-disable-next-line no-bitwise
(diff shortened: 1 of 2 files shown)


packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
```diff
@@ -486,6 +498,10 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 		this.siblings =
 			this.siblingStack.pop() ?? fail(0xaf0 /* Unexpected siblingStack.length */);
 		this.index = this.indexStack.pop() ?? fail(0xaf1 /* Unexpected indexStack.length */);
+		this.indexOfChunk =
+			this.indexOfChunkStack.pop() ?? fail("Unexpected indexOfChunkStack.length");
+		this.indexWithinChunk =
+			this.indexWithinChunkStack.pop() ?? fail("Unexpected indexWithinChunkStack.length");
```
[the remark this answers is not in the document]
brrichards: Since I am just moving this in the file I don't know if I should be re-using the fail codes.

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
```diff
@@ -203,8 +203,15 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 		return this.indexStack[height] ?? oob();
 	}
 
-	private getStackedNode(height: number): BasicChunk {
-		const index = this.getStackedNodeIndex(height);
+	private getStackedChunkIndex(height: number): number {
+		assert(height % 2 === 1, "must be node height");
+		assert(height >= 0, "must not be above root");
+		// eslint-disable-next-line no-bitwise
```
CraigMacomber: When suppressing the linter, its a good idea to document why its ok to suppress it in this case, like we do in https://github.com/microsoft/FluidFramework/pull/27084/changes#diff-e637d8b2f919c742dad1d0d98d18b101019ade1f3b748623aed966ba4b1119d7R169-R170

That said, we should probably deduplicate that code with this.

Introducing a `getNodeOnlyHeightFromHeight` method, which takes an optional height defaulting to this.siblingStack.length seems like it would be a good way to deduplicate this logic, and make it a bit more clear what is going on in it.

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
CraigMacomber: All of this seems like its just checking some invariants about the state of this cursor.

Randomly throwing invariant validation code in the middle of some method like this isn't a good pattern (I assume it was me who did this btw).

Factoring this out into something like an "assertValid" method, then calling it from a few key places might be a better approach.

You could probably then add a couple more checks which detect the kinds of bugs you found and are fixing here to help reduce the risk of regressions even more.

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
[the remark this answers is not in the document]
CraigMacomber: It is ok to reuse codes when moving the assert/fail around (so long as you don't introduce another copy of it), but it's also ok to remove them.

packages/dds/tree/src/test/feature-libraries/chunked-forest/basicChunk.spec.ts
```diff
@@ -155,6 +164,75 @@ describe("basic chunk", () => {
...
+			0,
+			0,
+			0,
+			undefined,
+		);
+
+		// Logical node index 2 maps to chunk array index 1 (the trailing BasicChunk).
+		cursor.enterNode(2);
+
+		// enterField drives getField, which on main looks up the parent at the
```
CraigMacomber: this code is going to be on main soon. If you want to comment on some specific comparison between this branch and main which will no longer apply after the PR merges, you can do so in a PR comment, or if putting it in the code, you need to phrase it differently.

Anyway, thanks for the explanation.

packages/dds/tree/src/test/feature-libraries/chunked-forest/basicChunk.spec.ts
```diff
@@ -155,6 +164,75 @@ describe("basic chunk", () => {
...
+		// logical node index (2) instead of the chunk array index (1). siblings[2]
+		// is out of bounds in the 2-chunk array, so the cast returns undefined and
+		// the subsequent field read throws.
```
CraigMacomber: When you say "looks up the parent", that's a bit confusing.

I think this is coming from the use of the variable name "parent" in get field, which is also rather poor.

Our trees and cursors do provide the ability to look up the parents of nodes, which sounds like what is being done here, but isn't.

Its instead simply looking up the current node, which it happens to call parent, since it's the parent of the field that method is operating on.

Also, you should make it clear why it was wrong to use a node index.

```suggestion
		// enterField uses getField, which must looks up the current node (who's field should be gotten).
		// This must handle when the index of the node does not match the index of the chunk containing that node.
		// This test is a node index 2 and chunk index 1.
```

packages/dds/tree/src/test/feature-libraries/chunked-forest/basicChunk.spec.ts
```diff
@@ -155,6 +164,75 @@ describe("basic chunk", () => {
...
+		cursor.enterField(subField);
+		assert.doesNotThrow(() => cursor.getFieldLength());
```
CraigMacomber: Rather than simply asserting it doesn't throw, it would be better to assert that enter field actually worked correctly (its easy to have bugs that don't throw).

Asserting the field length is 1, and that if you enter that node you get a value of 42 should do it I think.

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
```diff
@@ -355,6 +367,8 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 
 		this.siblingStack.push(this.siblings);
 		this.indexStack.push(this.index);
+		this.indexOfChunkStack.push(this.indexOfChunk);
+		this.indexWithinChunkStack.push(this.indexWithinChunk);
```
CraigMacomber, line 394: The places in the code where we edit these parallel arrays could call our validation method afterward ( see https://github.com/microsoft/FluidFramework/pull/27084/changes#r3103181969 ) which could assert they are properly in sync to catch future similar bugs.

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
```diff
@@ -169,7 +169,7 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 		// Compute the number of nodes deep the current depth is.
 		// We want the floor of the result, which can computed using a bitwise shift assuming the depth is less than 2^31, which seems safe.
 		// eslint-disable-next-line no-bitwise
-		const halfHeight = (this.siblingStack.length + 1) >> 1;
+		const halfHeight = this.siblingStack.length >> 1;
```
CraigMacomber: Was the old code wrong, but just never getting called in cases where that caused an assert before?

If so, https://github.com/microsoft/FluidFramework/pull/27084/changes#r3103181969 to make it easy to run this in more places seems extra valuable. I think that might have caught this bug.

brrichards: The old code was right for when we were pushing onto `siblingStack` and `chunkStack` The description was a little off since previously this was getting `ceil(siblingStack.length/2)` and not the floor. `this.siblingStack.length + 1)` gets the floor, but then "rounds up" by adding the 1. Since we changed when we push onto `chunkStack` we want the floor without rounding up, causing the equation change. 

Old Stack Growth 
| Transition | `siblingStack.length` | `chunkStack.length` |
| ---- | ---- | ---- | 
| fields (root) | 0 | 0 |
| After firstNode | 1 | 1 |
| After enterField | 2 | 1 | 
| After firstNode | 3 | 2 | 

New Stack Growth
| Transition | `siblingStack.length` | `chunkStack.length` |
| ---- | ---- | ---- | 
| fields (root) | 0 | 0 |
| After firstNode | 1 | 0 |
| After enterField | 2 | 1 | 
| After firstNode | 3 | 1 |

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
brrichards: I added the validation everytime we push/pop from the staacks

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
```diff
@@ -203,11 +209,24 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
...
+		assert(height >= 0, "must not be above root");
+		return this.indexOfChunkStack[this.getNodeOnlyHeightFromHeight(height)] ?? oob();
+	}
+
+	private getStackedChunk(height: number): BasicChunk {
+		const index = this.getStackedChunkIndex(height);
 		return (this.siblingStack[height] as readonly TreeChunk[])[index] as BasicChunk;
 	}
 
+	// Returns the floor of half the height passed in, or half of `siblingStack.length` if no height is provided.
```
CraigMacomber: Use doc comment syntax. Also phrasing there sounds like it returns siblingStack.length/2 if nothing is provided rather than using siblingStack.length as the default height (which gets floored after divide)

packages/dds/tree/src/test/feature-libraries/chunked-forest/basicChunk.spec.ts
brrichards: updated. Also stopped referring to the "parent" node as "parent" and described its relationship better with the subfield

packages/dds/tree/src/test/feature-libraries/chunked-forest/basicChunk.spec.ts
brrichards: Updated tests

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
```diff
@@ -502,16 +534,16 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
...
+		this.assertChunkStacksMatchNodeDepth();
 	}
 
 	private getNode(): BasicChunk {
 		assert(this.mode === CursorLocationType.Nodes, 0x52f /* can only get node when in node */);
-		return (this.siblings as TreeChunk[])[this.index] as BasicChunk;
+		return (this.siblings as TreeChunk[])[this.indexOfChunk] as BasicChunk;
 	}
 
 	private getField(): readonly TreeChunk[] {
```
CraigMacomber, line 568: This method is a bit confusing. Could you add a doc comment on it, and maybe a comment in the implementation about what that "parent" is the parent of, and why we need it?

CraigMacomber: Note that generally the mostly the public methods here are part of an interface and thus have docs, but these private ones are missing them.

brrichards: Update to have proper doc comments, and a remarks section explaining what a "parent" is and why we need it.

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
```diff
@@ -522,7 +574,7 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
 			this.mode === CursorLocationType.Fields,
 			0x530 /* can only get field when in fields */,
 		);
-		const parent = this.getStackedNode(this.indexStack.length - 1);
+		const parent = this.getStackedChunk(this.siblingStack.length - 1);
```
CraigMacomber: Details about the parent should be a code comment inside the method, since its not relevant to callers.

Also, now that I see this explanation (thanks for that!), I think this code could be simplified.

this.siblingStack should, at the correct level, contain the TreeChunk array we want, so we can skip having to get the parent node in the first place.

packages/dds/tree/src/feature-libraries/chunked-forest/basicChunk.ts
```diff
@@ -203,11 +209,32 @@ export class BasicChunkCursor extends SynchronousCursor implements ChunkedCursor
...
-		const index = this.getStackedNodeIndex(height);
+	private getStackedChunkIndex(height: number): number {
+		assert(height % 2 === 1, "must be node height");
+		assert(height >= 0, "must not be above root");
+		return this.indexOfChunkStack[this.getNodeOnlyHeightFromHeight(height)] ?? oob();
+	}
+
+	private getStackedChunk(height: number): BasicChunk {
+		const index = this.getStackedChunkIndex(height);
 		return (this.siblingStack[height] as readonly TreeChunk[])[index] as BasicChunk;
```
CraigMacomber: This should always be true, but for clarity/safety, can we:

```suggestion
		const chunk = this.siblingStack[height] as readonly TreeChunk[])[index];
		debugAssert(() => chunk instanceof BasicChunk && "Only basic chunks are expected");
		return chunk as BasicChunk;
```

CraigMacomber: Note that as this are is rather performance sensitive, and very unlikely to fail due to some specific runtime data, I opted for debugAssert which can be compiled out of production builds. We will likely want to do the same for some other places in here.

brrichards: I found one other place with the `as BasicChunk` cast and updated it to match this

CraigMacomber (approved):

PR #27086, opened 2026-04-17
refactor(build-cli): improve flub ai alias handling and launcher prompt

tylerbutler:
## Summary

Simplifies and improves the `flub ai` launcher's alias handling without adding configurable alias loading machinery:

- Extracts `buildLauncherPrompt` helper to inject the hardcoded allowed alias list into the launcher prompt template, giving the LLM a deterministic list to choose from
- Makes `assertSafeAliasSelection` accept an explicit `Set<string>` for testability
- Adds an "Allowed Aliases for This Session" section to `launcher-prompt.md` so the LLM sees a clear allow-list separate from the full alias definitions script
- Consolidates redundant ONLY-recommend guidelines into one clear rule (addresses review feedback from #27079)
- Fixes `this.warn` → `this.warning` deprecation
- Adds a "Please wait..." status message after agency installation

This is a simpler alternative to #27079 — it keeps the alias allow-list hardcoded in code rather than making it configurable via frontmatter or shell script manifest. The deterministic allow-list serves as a security gate between LLM output and shell execution. Configurability can be added later if needed.

## Test plan

- [x] `npm run build` passes
- [x] All 491 tests pass (`npm test`)
- [x] CI readiness check passes (checks:fix clean)
- [x] `assertSafeAliasSelection` tests cover custom alias sets
- [x] `buildLauncherPrompt` test verifies template variable injection

### DIFF
--- a/build-tools/packages/build-cli/src/test/commands/ai.test.ts
+++ b/build-tools/packages/build-cli/src/test/commands/ai.test.ts
@@ -14,20 +14,6 @@ import {
 } from "../../commands/ai.js";
 
 describe("ai command", () => {
-	it("allows aliases present in the provided set", () => {
-		const aliasSet = new Set(["my-alias", "other-alias"]);
-		expect(() =>
-			assertSafeAliasSelection({ alias: "my-alias", explanation: "custom alias" }, aliasSet),
-		).to.not.throw();
-	});
-
-	it("rejects aliases not in the provided set", () => {
-		const aliasSet = new Set(["my-alias", "other-alias"]);
-		expect(() =>
-			assertSafeAliasSelection({ alias: "claude", explanation: "not in set" }, aliasSet),
-		).to.throw(/Unsupported AI alias selection: claude/);
-	});
-
 	it("allows all supported aliases", () => {
 		const aliasSet = new Set<string>(SUPPORTED_ALIASES);
 		for (const alias of SUPPORTED_ALIASES) {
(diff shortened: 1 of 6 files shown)


build-tools/packages/build-cli/src/test/commands/ai.test.ts
```diff
@@ -8,23 +8,46 @@ import { describe, it } from "mocha";
...
+	buildLauncherPrompt,
 	normalizePromptAnswer,
-	supportedAliases,
+	SUPPORTED_ALIASES,
 } from "../../commands/ai.js";
 
 describe("ai command", () => {
-	it("allows known agent aliases", () => {
-		for (const alias of supportedAliases) {
+	it("allows aliases present in the provided set", () => {
```
noencke: IMO we don't need three unit tests for this trivial function - we're basically unit testing `Set.has`. I don't think we need to unit test it at all, but if we want to keep any of the tests, I guess the one that checks the error string is the most useful.

noencke (approved):

build-tools/packages/build-cli/src/test/commands/ai.test.ts
tylerbutler: yeah I felt similarly but... they were here. I really wish we had a faster CI for "trivial PR updates." I would be much more amenable to small changes if it didn't cause 45 mins of waiting :)

noencke (approved):

PR #27091, opened 2026-04-18
Disable code coverage for server Docker builds

ChumpChief:
[AB#71226](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/71226)

## Description

Adds a `testCoverage` parameter to the `build-docker-service` template (defaults to `false`)
to control whether tests run with code coverage analysis. Previously code coverage was
hardcoded to run for non-PR `ci:test` builds. The new parameter aligns with the existing
`testCoverage` parameter in `build-npm-client-package.yml`.

Code coverage is disabled for routerlicious because it regressed in Node 22, causing test
timeouts that have failed several CI runs (e.g. 392547, 3925124). This wasn't caught earlier
because code coverage only runs for CI builds, not PR validation builds.

## Reviewer Guidance

- The `testCoverage` parameter defaults to `false`, so no existing pipeline behavior changes
  unless a consumer explicitly opts in.
- The commented-out `testCoverage` line in `server-routerlicious.yml` preserves the original
  condition for when coverage can be re-enabled.

### DIFF
--- a/tools/pipelines/server-routerlicious.yml
+++ b/tools/pipelines/server-routerlicious.yml
@@ -137,6 +137,9 @@ extends:
     pack: true
     lint: true
     test: ci:test
+    # Disabled: coverage performance regressed in Node 22, causing test timeouts
+    # that need mitigation before coverage can be reenabled.
+    # testCoverage: ${{ ne(variables['Build.Reason'], 'PullRequest') }}
     docs: true
     containerBaseDir: /usr/FluidFramework/server/routerlicious
     checks:
--- a/tools/pipelines/templates/build-docker-service.yml
+++ b/tools/pipelines/templates/build-docker-service.yml
@@ -22,6 +22,11 @@ parameters:
     type: string
     default:
 
+  # Indicates if tests should be run with code coverage analysis.
+  - name: testCoverage
+    type: boolean
+    default: false
+
   - name: docs
     type: boolean
     default: false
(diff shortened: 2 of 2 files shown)


tools/pipelines/server-routerlicious.yml
```diff
@@ -137,6 +137,9 @@ extends:
     pack: true
     lint: true
     test: ci:test
+    # Disabled: coverage performance regressed in Node 22, causing test timeouts
+    # that need mitigation before coverage can be reenabled.
```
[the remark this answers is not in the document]
ChumpChief: won't fix

alexvy86, line 141: Do we need a similar note for the historian and gitrest pipelines?

ChumpChief: Historian and gitrest currently don't run coverage for two reasons:
* They [pass "test"](https://github.com/microsoft/FluidFramework/blob/main/tools/pipelines/server-historian.yml#L115) as the `test` parameter, but the template only runs coverage [for a value of "ci:test"](https://github.com/microsoft/FluidFramework/blob/main/tools/pipelines/templates/build-docker-service.yml#L444).
* They don't have "ci:test" scripts (and historian doesn't have any script that uses `c8`)

Not sure why the discrepancies between them - I do see historian has a c8 block in its package.json but no actual c8 dependency or script.  In any case, adding coverage for them would be a little bit more than just toggling the param .

alexvy86 (approved):

PR #27092, opened 2026-04-18
Make duration benchmarking API more consistant

CraigMacomber:
## Description

Currently there are a lot of ways to write duration benchmarks, and it's easy to miss tagging them properly.
This removes all the deprecated ways, and adds guidance to the remaining ones to funnel use into a consistent pattern.

## Reviewer Guidance


tools/benchmark/src/durationBenchmarking/configuration.ts
```diff
@@ -68,8 +83,70 @@ export interface BenchmarkTimer<T> {
...
+	 * 	let running: boolean;
+	 * 	do {
+	 * 		// Per-batch setup (not timed)
+	 * 		running = await state.timeBatchAsync(async () => { /* operation *\/ });
+	 * 		// Per-batch teardown (not timed)
+	 * 	} while (running);
+	 * },
+	 * ```
+	 */
+	timeBatchAsync(callback: () => Promise<unknown>): Promise<boolean>;
```
alexvy86, line 117: Is there a reason to use unknown here instead of void? We don't expect the callback to return anything, do we?
```suggestion
	timeBatchAsync(callback: () => Promise<void>): Promise<boolean>;
```

tools/benchmark/src/durationBenchmarking/getDuration.ts
```diff
@@ -100,11 +101,10 @@ export async function collectDurationData(args: DurationBenchmark): Promise<Coll
...
 	}
-	await options.after?.();
 	return data;
 }
 
-class BenchmarkState<T> implements BenchmarkTimer<T> {
+export class BenchmarkState<T> implements BenchmarkTimer<T> {
 	/**
 	 * Duration for each batch, in seconds.
 	 */
```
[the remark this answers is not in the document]
alexvy86: I don't recall if the fact that `timer` is a constructor parameter will make references to that name use that before trying to use the module-level one... worth double checking.

alexvy86 (commented):
Copilot comments seem worth addressing and left one small one myself

tools/benchmark/src/durationBenchmarking/configuration.ts
```diff
@@ -60,6 +60,21 @@ export interface BenchmarkTimer<T> {
 	 * Records the duration of a completed batch and advances internal state.
 	 * @param duration - The elapsed time for the batch in seconds. Compute this using {@link Timer.toSeconds}.
 	 * @returns `true` if another batch should be run, `false` if data collection is complete.
+	 * @example
```
Josmithr, line 63: Thanks for adding examples! I've noticed that Claude pays attention to these - I think we should leverage them more.

tools/benchmark/src/durationBenchmarking/configuration.ts
CraigMacomber: It is handy to be able to pass function to time like `() => computeFoo()` to measure how long it takes to compute Foo rather than forcing you to do `() => { computeFoo(); }`

Making this use void would ban that, and propagate back to a breaking API change on benchmarkFnAsync to make the same change there.

tools/benchmark/src/durationBenchmarking/getDuration.ts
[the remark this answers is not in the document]
CraigMacomber: AI was wrong, and I assume confused by shadowing. I have removing the shadowing to impropve clarity.

alexvy86 (approved):

alexvy86 (approved):

PR #27097, opened 2026-04-20
fix(routerlicious): enforce scope checks on PATCH /root endpoint (#27…

sonalideshpandemsft:
…073)

The `PATCH /:tenantId/:id/root` endpoint in Alfred's REST API called `verifyToken()` without passing the `requiredScopes` parameter. This meant any valid JWT — including tokens with only `doc:read` scope — could write arbitrary key-value data into a document's root map. The injected operations were permanently sequenced into the document's op log and replayed by all connected clients.

The fix threads a `requiredScopes` parameter through the `verifyRequest` → `verifyTokenWrapper` → `verifyToken` call chain and passes `[ScopeType.DocRead, ScopeType.DocWrite]` from the PATCH handler. This matches the scope enforcement already used by the adjacent `POST /broadcast-signal` endpoint via `verifyStorageToken`.

Two new tests confirm that:
- A `doc:read`-only token is rejected with 403
- A `doc:write`-only token (missing `doc:read`) is also rejected with 403

The existing "process normally" test with a full-scope token continues to pass, confirming no regression for legitimate callers.

The review process is outlined on [this wiki
page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

- This is a security fix (MSRC 113235). The changes are intentionally minimal and surgical — 4 hunks in the production file, all adding a single `requiredScopes` parameter through the existing function chain.
- `verifyRequest` and `verifyTokenWrapper` are module-private (not exported). The `create()` export signature is unchanged. Zero consumer impact.

### DIFF
--- a/server/routerlicious/packages/services-utils/src/auth.ts
+++ b/server/routerlicious/packages/services-utils/src/auth.ts
@@ -235,6 +235,7 @@ export async function verifyToken(
 	token: string,
 	tenantManager: ITenantManager,
 	options: IVerifyTokenOptions,
+	requiredScopes?: string[],
 ): Promise<void> {
 	if (options.requireDocumentId && !documentId) {
 		throw new NetworkError(403, "Missing documentId.");
@@ -265,6 +266,18 @@ export async function verifyToken(
 			);
 			if (isTokenRevoked) {
 				throw new NetworkError(403, "Permission denied. Access token has been revoked.");
+			}
+		}
+
+		if (requiredScopes) {
+			const hasAllRequiredScopes = requiredScopes.every((scope) =>
+				claims.scopes.includes(scope),
+			);
+			if (!hasAllRequiredScopes) {
+				throw new NetworkError(
+					403,
+					`Permission denied. Insufficient scopes. Required scopes: ${requiredScopes}`,
+				);
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


server/routerlicious/packages/services-utils/src/auth.ts
```diff
@@ -268,6 +269,18 @@ export async function verifyToken(
 			}
 		}
 
+		if (requiredScopes) {
```
sonalideshpandemsft, line 272: The PATCH `/:tenantId/:id/root` endpoint in Alfred's REST API had no scope enforcement — any valid JWT (even `doc:read`-only) could write to a document's root map.

PR #27073 added `requiredScopes` plumbing through `verifyRequest` → `verifyTokenWrapper` → `verifyToken` and passed `[ScopeType.DocRead, ScopeType.DocWrite]` from the PATCH handler. However, on the `release/server/6.0` branch, `verifyToken()` in `@fluidframework/server-services-utils` didn't have a `requiredScopes` parameter — that was added on `main` by PR #24853. So the scopes were silently ignored and tests expecting 403 got 200.

**The additional fix** (in auth.ts) backports the minimal piece from #24853:

- Adds `requiredScopes?: string[]` as an optional parameter to `verifyToken()`
- Adds scope validation logic: checks that every required scope is present in the token's `claims.scopes`, throwing a `403 NetworkError` if not

noencke (approved):

frankmueller-msft (approved):

PR #27100, opened 2026-04-20
feat(container-loader): add captureFullContainerState free function

anthony-murphy:
## Description

Adds a new `@legacy @alpha` free function, `captureFullContainerState`, that captures the full referenced state of an attached container using only driver-level services — an `IDocumentServiceFactory`, an `IUrlResolver`, and a request — with no runtime, codeLoader, or live `Container` instance. The output is a stringified `IPendingContainerState` in the same wire format produced by a live container's pending-state serialization, and can be handed to `loadExistingContainer` or `loadFrozenContainerFromPendingState` as `pendingLocalState`.

The captured state is a self-contained view of the container's referenced graph:

- Base snapshot with inlined contents for every blob reachable through referenced subtrees. `ISnapshotTree.unreferenced` (set by the summarizer from GC state) is honoured — dead subtrees are not read.
- Every loading-group snapshot declared by datastores, pre-fetched via `IDocumentStorageService.getSnapshot({ versionId, loadingGroupIds })` and walked with the same filter. Populates `IPendingContainerState.loadedGroupIdSnapshots`.
- Attachment blob contents keyed by storage id. Blobs that GC has explicitly marked `unreferencedTimestampMs`, tombstoned, or deleted are skipped; blobs absent from the GC graph are kept (GC lag tolerance). If the snapshot has no GC tree, every attachment blob is included.
- All ops with sequence numbers after the base snapshot's sequence number (as read from the snapshot's `.attributes` blob).

Attachment blob inlining piggybacks on the existing `ContainerStorageAdapter` cache: it consults `snapshotBlobs` by id before falling through to storage, so a frozen loader can serve the full referenced graph without a live storage service. No wire-format change.

`pendingRuntimeState` is `undefined` — no runtime is instantiated — so the output cannot carry DDS-level in-flight changes. Intended for state relay, inspection, and durable-state snapshot use cases.

Implementation notes:
- New file `captureReferencedContents.ts` holds the tree walker, GC parser, attachment-blob capture, and groupId fetcher. A handful of blob-manager / GC constants are duplicated locally with a comment pointing to the canonical definitions in `container-runtime` / `runtime-definitions`, to avoid a loader → runtime dependency.
- 15 unit tests against direct `ISnapshotTree` fixtures + mocked storage cover the GC filtering and groupId paths; 4 end-to-end scenarios against `LocalDeltaConnectionServer` in `local-server-tests` cover the round-trip through `loadFrozenContainerFromPendingState`.

## Reviewer Guidance

- Layering: blob-manager / GC constants and GC-parse logic are duplicated in `container-loader` rather than taking a dependency on `container-runtime` / `runtime-definitions`. Comment in the file points to the canonical sources. If you'd prefer an extracted shared package, flag it.
- GC semantics: blobs absent from `gcState.gcNodes` are *kept*, not skipped — intentional, to tolerate GC lag for recently-attached blobs. Only `unreferencedTimestampMs`, tombstones, and deletedNodes filter content out.
- Consistency: a new snapshot can land between snapshot fetch and ops fetch. The returned state remains internally consistent (ops are anchored to the captured snapshot) but may not reflect the very latest snapshot. Documented in the function's jsdoc; no retry logic added.
- Tests: integration tests cover the happy path and attachment-blob round-trip. Unit tests cover paths the local suite can't easily reach (no summarizer runs locally, so `ISnapshotTree.unreferenced` is never set end-to-end; `TestFluidObjectFactory` doesn't produce loading-group datastores).


packages/loader/container-loader/src/captureReferencedContents.ts
```diff
@@ -0,0 +1,327 @@
...
+	ISnapshot,
+	ISnapshotTree,
+} from "@fluidframework/driver-definitions/internal";
+import { readAndParse } from "@fluidframework/driver-utils/internal";
+
+import type { ISerializableBlobContents } from "./containerStorageAdapter.js";
+import type { SerializedSnapshotInfo } from "./serializedStateManager.js";
+import { getDocumentAttributes } from "./utils.js";
+
+// The following names are defined authoritatively in container-runtime and
```
anthony-murphy, line 18: we probably want to move all this stuff to somewhere else in our common-definitions. it would couple the code a bit more, but these can't be easily changed today give its all data format. also thought about a potential runtime-protocol package.

packages/loader/container-loader/src/captureReferencedContents.ts
```diff
@@ -0,0 +1,360 @@
...
+ */
+export async function captureGroupIdSnapshots(
+	baseSnapshot: ISnapshotTree,
+	storage: Pick<IDocumentStorageService, "readBlob"> & {
+		getSnapshot?: IDocumentStorageService["getSnapshot"];
+	},
+	versionId: string | undefined,
+	scenarioName: string,
+): Promise<Record<string, SerializedSnapshotInfo>> {
+	const getSnapshot = storage.getSnapshot;
```
anthony-murphy: Deep Review: `storage.getSnapshot` is extracted into a local variable here, which strips the `this` binding. When real driver implementations call this (e.g., `LocalDocumentStorageService.getSnapshot` which references `this.id`, `this.getVersions(...)`, etc.), `this` will be `undefined` in strict mode, causing a `TypeError`.

This is confirmed by the existing `.bind()` pattern elsewhere in the codebase — see `protocolTreeDocumentStorageService.ts:31` which binds the same method.

Unit tests don't catch this because `mockStorage` is a plain object with arrow-function-like properties that don't depend on `this`. Integration tests don't hit this path because test containers lack loading-group subtrees.

**Fix:** Either bind when extracting:
```ts
const getSnapshot = storage.getSnapshot?.bind(storage);
```
Or call the method directly on the object:
```ts
await storage.getSnapshot?.({ cacheSnapshot: false, versionId, loadingGroupIds: [groupId] });
```

anthony-murphy: Deep Review: Retracting this finding — the code already handles `this` binding correctly at line 371 with `storage.getSnapshot?.bind(storage)` and an explanatory comment. The original analysis missed the `.bind()` call. Apologies for the false positive.

anthony-murphy:
## Deep Review

*Reviewed commit `796e664` on 2026-04-28.*

**Readiness: 5/10 — 🔨 MAKING PROGRESS**

`captureFullContainerState` is a well-designed `@legacy @alpha` addition. The overall architecture, GC handling, concurrency controls, and `dispose()` placement all hold up on re-review. Two correctness/parity gaps surfaced this round and should be resolved before the `ICaptureFullContainerStateProps` interface freezes: a lossy UTF-8 round-trip on captured attachment-blob bytes (now in scope — the new `captureReferencedAttachmentBlobs` is the code path that introduces the encoding decision), and the absence of the `mixinMonitoringContext` / `createChildMonitoringContext` / `configProvider` wiring that PR #25394 established as the convention for the sibling function in this same file. A layering question (duplicated GC/blob-manager constants) is held as a question for the author since you've already engaged on it in [comment 3113713802](https://github.com/microsoft/FluidFramework/pull/27100#discussion_r3113713802).

The prior 7/10 summary classified the binary-attachment-blob concern as out-of-scope ("pre-existing pattern"). On re-review that classification was wrong: the new `captureReferencedAttachmentBlobs` is what creates the wire-format obligation here, and the runtime's own pending-blob serializer uses `base64` for exactly this reason. Re-tiering it in-scope is what moves the score from 7 → 5.

### Findings

- **Binary attachment blob bytes are corrupted by lossy UTF-8 round-trip during capture** — `code/packages/loader/container-loader/src/captureReferencedContents.ts:231-235` — (inline thread)
- **New free function omits monitoring-context wiring required by reviewer convention for its sibling in the same file** — `code/packages/loader/container-loader/src/createAndLoadContainerUtils.ts:308-340` — (inline thread)

### Questions for Author

1. **Binary attachment-blob support.** Is `captureFullContainerState` intended to support arbitrary binary attachment payloads (images, encrypted blobs), or only UTF-8-safe attachments? If the former, the encoding fix should land before the props interface freezes. If the latter, please document the restriction in the function's jsdoc and `ICaptureFullContainerStateProps`.
2. **Monitoring-context wiring.** Is omitting `mixinMonitoringContext` / `createChildMonitoringContext` intentional because no Loader/runtime is instantiated here? If yes, a one-line docblock note avoids relitigating PR #25394's review point on the next pass. The `configProvider?: IConfigProviderBase` field decision is the time-sensitive piece because of the `@alpha` props freeze.
3. **Layering / duplicated constants** (held until the above resolve). Per your [comment 3113713802](https://github.com/microsoft/FluidFramework/pull/27100#discussion_r3113713802): is the plan (a) land the duplication now and file a follow-up to extract to `common-definitions` / a runtime-protocol package, or (b) extract as a precondition for merging? ChumpChief is the established reviewer for the alpha-surface scaling question on this file (per PR #25513) and should weigh in on whether `captureFullContainerState` stays a top-level free function or becomes an overload.

### Path to Ready

To move from 5 → 7-8 (ALMOST READY):

- [ ] **Resolve the binary attachment-blob encoding** — either base64-encode attachment-blob bytes at the new encode site (and decode site in `serializedStateManager`), or extend the props/wire format to carry binary entries explicitly. Add a non-UTF-8 attachment-blob round-trip test in `captureFullContainerState.spec.ts` (e.g. `Uint8Array([0xff, 0xfe, 0x00])`) asserting byte-exact equality.
- [ ] **Resolve the monitoring-context wiring** — either wire `mixinMonitoringContext` + `createChildMonitoringContext` and add `configProvider?: IConfigProviderBase` to `ICaptureFullContainerStateProps`, or add a docblock comment explaining the deliberate divergence from the PR #25394 sibling pattern.
- [ ] **Confirm `@legacy @alpha` props shape is final** — `ICaptureFullContainerStateProps` is about to freeze; any additions (configProvider, binary-blob-mode flag) need to land before merge or become an overload-only path.

To move from 7-8 → 9-10 (READY):

- [ ] **ChumpChief sign-off on alpha-surface scaling** — free function vs. overload of `asLegacyAlpha`, per the PR #25513 thread.
- [ ] **jatgarg or dannimad sign-off** — jatgarg for the telemetry/configProvider decision (PR #25394 enforcement on this file), dannimad for the frozen-container API surface (co-author of #25653 which this output feeds).
- [ ] **Decide layering** — file a follow-up to extract `gcTreeKey` / `__gc*` / `blobsTreeName` / `IGarbageCollectionState` into `runtime-utils` (or add a contract test in `container-runtime` that fails on drift). Either choice is fine; the decision needs to be on record.

### Context for Reviewers

- This PR is the fourth API addition on the `container-loader.legacy.alpha` surface, completing a series: PR #25513 (`asLegacyAlpha`), #25538, #25590 (pending local state without blobs — leveraged here), #25742, and #25653 (`createFrozenDocumentServiceFactory`). The output of `captureFullContainerState` is designed to feed `loadFrozenContainerFromPendingState`. Suggested reviewers: dannimad, ChumpChief, jatgarg.
- The consistency-window caveat between snapshot fetch and ops fetch ("if a new snapshot lands between the snapshot fetch and the ops fetch, the returned state may not reflect the very latest snapshot, but remains internally consistent") is documented in jsdoc and continues the author's pattern from PR #25590 of being explicit about partial-offline limits.
- GC reachability via `ISnapshotTree.unreferenced` (rather than re-parsing `gc/__gc_root` to build a `referencedNodeIds` set) is the minimal correct surface — re-parsing would create a second source of truth that can drift from the summarizer. The GC tree is parsed only for the per-localId attachment-blob filters (`__tombstones`, `__deletedNodes`, `unreferencedTimestampMs`) that the `unreferenced` flag does not encode.
- Test split: 15 unit tests cover GC-filtering and groupId paths (the local test server doesn't run a summarizer, so `ISnapshotTree.unreferenced` is never set end-to-end, and `TestFluidObjectFactory` doesn't produce loading-group datastores); 4 integration tests cover the round-trip happy path through `loadFrozenContainerFromPendingState`. Author has been transparent about this in the reviewer guidance.

- 2026-04-21 · **7/10** — false-positive `this`-binding finding retracted; loading-group silent-drop concern raised; binary-blob UTF-8 marked out-of-scope (now reclassified in-scope)

</details>

PR #27105, opened 2026-04-20
tree: Add integration-style tests for loading old summary versions

jzaffiro:
Prevent any future bugs for loading old summaries in newer versions of the code by adding tests for that case. 

[AB#59669](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/59669)


packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,119 @@
...
+	label: sf.string,
+	child: sf.array("nodes", sf.object("child", { count: sf.number })),
+}) {}
+
+async function regenerateOldSummaries(): Promise<void> {
+	if (existsSync(outputDir)) {
+		console.log(`removing snapshot directory: ${outputDir}`);
+		rmSync(outputDir, { recursive: true, force: true });
+	}
+	mkdirSync(outputDir, { recursive: true });
```
CraigMacomber: I don't think we ever want to remove or update these files. We should only ever add additional variants, since we need to support loading old summaries, even if we don't generate them like that anymore.

So instead of "regenerateOldSummaries" I think we want something like "checkForMissingSummaries(addIfMissing: boolean)"

Rather than deleting all the summaries, it can load all of them (the string contents of the files) into a `Set<string>`.

Then when we loop through all the summaries we could generate, generate the string we would save to a file.

Then check if that string is in the set: if its not, push it into an array of missing summaries.

Then after the loop, if the array of missing summaries is not empty:
If `addIfMissing` is false: report an error listing all the ones which are missing.
If `addIfMissing` is true: write them all to disk with non-colliding names.

To help with the non-colliding names we can include a `-1` at the end of each file name, and increment that number if the file already exists.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,119 @@
...
+
+	for (const treeEncodeType of [
+		TreeCompressionStrategy.Compressed,
+		TreeCompressionStrategy.Uncompressed,
+	]) {
+		const treeEncodeKey = TreeCompressionStrategy[treeEncodeType];
+		for (const versionKey of Object.keys(FluidClientVersion)) {
+			const dir = path.join(outputDir, treeEncodeKey, versionKey);
```
CraigMacomber: I don't think we need to put treeEncodeKey and versionKey in both the folder structure and file name. I think just the file name should be enough.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,119 @@
...
+		rmSync(outputDir, { recursive: true, force: true });
+	}
+	mkdirSync(outputDir, { recursive: true });
+
+	for (const treeEncodeType of [
+		TreeCompressionStrategy.Compressed,
+		TreeCompressionStrategy.Uncompressed,
+	]) {
+		const treeEncodeKey = TreeCompressionStrategy[treeEncodeType];
+		for (const versionKey of Object.keys(FluidClientVersion)) {
```
CraigMacomber: It would be good to also include the cleanedPackageVersion in this just in case there is some change which someone didn't make a FluidClientVersion entry for.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
CraigMacomber: Note that once we do this, it should run as a test, and not the before hook.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,119 @@
+/*!
```
CraigMacomber, line 1: Putting a short README.md in the summary-load-snapshots folder noting what the files are linking to here for more information would be nice.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,110 @@
...
+			const { summary } = await tree.summarize(true);
+			writeFileSync(path.join(dir, "singleTree.json"), JSON.stringify(summary, undefined, 2));
+		}
+	}
+}
+
+describe("Summary load regression tests", () => {
+	before(async () => {
+		await generateOldSummaries();
+	});
```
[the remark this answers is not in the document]
jzaffiro: Addressed by committing snapshot files and adding a regenerate mode separate from the normal test flow

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,110 @@
...
+const sf = new SchemaFactory("test schema");
+class TestSchema extends sf.object("parent", {
+	label: sf.string,
+	child: sf.array("nodes", sf.object("child", { count: sf.number })),
+}) {}
+
+async function generateOldSummaries(): Promise<void> {
+	// Clean and recreate the output directory each run
+	rmSync(outputDir, { recursive: true, force: true });
+
```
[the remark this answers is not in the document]
jzaffiro: Addressed by committing snapshot files and adding a regenerate mode separate from the normal test flow

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
jzaffiro: Updated such that the file name contains the version and compression information.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
jzaffiro: Committed summary files and runs a "checkForMissingSummaries" test in addition to the load tests for each summary file in the directory.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,110 @@
...
+					const dataStoreRuntime = new MockFluidDataStoreRuntime();
+					const stFactory = configuredSharedTree(options).getFactory();
+
+					const tree = await stFactory.load(
+						dataStoreRuntime,
+						"test",
+						MockSharedObjectServices.createFromSummary(JSON.parse(summaryJson)),
+						stFactory.attributes,
+					);
+					assert(tree !== undefined, "Loaded tree should not be undefined");
```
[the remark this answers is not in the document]
jzaffiro: Updated and added a comment to fix assertions if summary content generation is updated.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,182 @@
...
+		jsonValidator: FormatValidatorBasic,
+		treeEncodeType,
+		minVersionForCollab:
+			versionKey === cleanedPackageVersion
+				? cleanedPackageVersion
+				: FluidClientVersion[versionKey as keyof typeof FluidClientVersion],
+	};
+
+	const dataStoreRuntime = new MockFluidDataStoreRuntime();
+	const stFactory = configuredSharedTree(options).getFactory();
```
CraigMacomber: https://github.com/microsoft/FluidFramework/wiki/Coding-Guidelines#-avoid-using-abbreviations

jzaffiro: Done - updated all abbreviations in the file.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,182 @@
...
+const sf = new SchemaFactory("test schema");
+class TestSchema extends sf.object("parent", {
+	label: sf.string,
+	child: sf.array("nodes", sf.object("child", { count: sf.number })),
+}) {}
+
+function listSnapshotFiles(dir: string): string[] {
+	if (!existsSync(dir)) {
+		return [];
+	}
```
CraigMacomber: If someone breaks this so we don't find any snapshots, I think it would be better to just fail the test, rather than act list we have no snapshots to test.

Thus I think an 
```suggestion
	assert(existsSync(dir));
```

would be better.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,182 @@
...
+		await checkForMissingSummaries(regenerateSnapshots);
+	});
+
+	for (const treeEncodeType of [
+		TreeCompressionStrategy.Compressed,
+		TreeCompressionStrategy.Uncompressed,
+	]) {
+		const treeEncodeKey = TreeCompressionStrategy[treeEncodeType];
+		describe(`Load singleTree summary with current minVersionForCollab and TreeCompressionStrategy.${treeEncodeKey}`, () => {
+			const files = listSnapshotFiles(outputDir).filter((f) =>
```
CraigMacomber: I think we want to just process all the json files, rather than further restricting the set of files.

In my opinion, if someone puts a json file in the snapshots folder, either its a snapshot and we should test it, or they screwed up and we should error. Silently acting as if their files does not exist is an easy way to let bugs slip through were we arn't testing what we thought we were.

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,182 @@
...
+describe("Summary load regression tests", () => {
+	it("has summary snapshots for all supported variants", async () => {
+		await checkForMissingSummaries(regenerateSnapshots);
+	});
+
+	for (const treeEncodeType of [
+		TreeCompressionStrategy.Compressed,
+		TreeCompressionStrategy.Uncompressed,
+	]) {
+		const treeEncodeKey = TreeCompressionStrategy[treeEncodeType];
```
CraigMacomber: There is no need to include this outer loop.

TreeCompressionStrategy has no impact on loading summaries, only on saving them

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
```diff
@@ -0,0 +1,182 @@
...
+	view.initialize({ label: "root", child: [] });
+	view.root.label = "foo";
+	view.root.child.push({ count: 1 });
+	view.root.child.push({ count: 2 });
+
+	const { summary } = await tree.summarize(true);
+	return `${JSON.stringify(summary, undefined, "\t")}\n`;
+}
+
+async function checkForMissingSummaries(addIfMissing: boolean): Promise<void> {
```
CraigMacomber, line 105: the functions new like this one could benefit from a doc comment.

This one in particular could use both a clarification of what it does, and why that is a thing we want to do (mainly so that we have an easy wat to keep around old summaries to ensure that similarly old documents will always load, even if we make changes to how we would encode them now).

packages/dds/tree/src/test/shared-tree/summaryLoad.integration.ts
jzaffiro: Makes sense - added checks in the listSnapshotFiles function as well as in the test itself (for non-snapshot json files).

CraigMacomber (approved):

PR #24166, opened 2025-03-27
(compat) Added support for layer compat validation at the Runtime / DataStore boundary

agarwal-navin:
## Description

Added logic to support layer compatibility validation at the Runtime / DataStore layer boundary.

For context, please look at #22877 which added the same support to the Runtime / Loader boundary. This logic added by this PR is the same as in #22877.

### Supported features
- Both the Runtime and DataStore layers have a `supportedFeatures` property that includes a set of features that a layer supports. This is advertised to the other layer at the layer boundary.
- Both the layers have a `requiredFeatures` array that includes a set of features that it requires the other layer to support in order to be compatible. Note that this is internal to that layer and is not advertised.
- At runtime, each layer validates that all the entries in `requiredFeatures` is present in the `supportedFeatures` of the other layer. If not, it throws a usage error that says layers are not compatible.
- Any change that adds new behavior or changes existing behavior across a layer boundary must add an entry to the supported features set. For example, changes such as "change the format of the the summary returned by the DataStore to Runtime during summarize call" should add an entry to the supported features set.

### Generation
- In addition to `supported features`, a `generation` is also added to both the layers. This is advertised to the other layer at the layer boundary.
- `Generation` is an integer which will be incremented every month. It can be used to validate compatibility more strictly (time-based) than `supported features` where layers are incompatible only if there are unsupported features.
_Note: The logic to update the generation will be added in a later PR. See [AB#27054](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/27054) for a proposed solution._
- One key reason for adding generation is to have a clear cut off point where we officially stop supporting layer combinations. Our compat tests will test combinations up to this cut off point. The idea is that we test what we support and whatever we don't test should not be supported. This will help us achieve that.
- Both the layers have a `minSupportedGeneration` that it needs the other layer to be at. Note that this is internal to that layer and is not advertised.
- At runtime, each layer validates that the `generation` of the other layer is `>= minSupportedGeneration`.
- For example, say that the Runtime is at generation 10 and has a minSupportedGeneration of 7 for the DataStore layer (3 month support window). If it sees that DataStore's generation is lower than 7, the layers are incompatible.

## Backwards compatibility
[...]

PR #27112, opened 2026-04-21
Move staging mode api to beta

sonalideshpandemsft:
This PR promotes the staging mode API from alpha to beta. Key changes:

- `enterStagingMode()` and `inStagingMode` are moved on `IContainerRuntimeBase` interface. A new `StageControls` interface (marked beta) is exported alongside them.
- `inStagingMode` and `isDirty` are moved from `IFluidDataStoreRuntimeAlpha` to  `IFluidDataStoreRuntime` interface.
- Alpha interfaces are not removed as future alpha features can reuse them.


packages/runtime/datastore/src/dataStoreRuntime.ts
```diff
@@ -542,16 +540,16 @@ export class FluidDataStoreRuntime
...
+	public get inStagingMode(): IFluidDataStoreRuntime["inStagingMode"] {
+		return this.dataStoreContext.containerRuntime?.inStagingMode;
 	}
 
 	/**
-	 * Implementation of IFluidDataStoreRuntimeAlpha.isDirty
+	 * Implementation of IFluidDataStoreRuntime.isDirty
 	 */
-	private get isDirty(): IFluidDataStoreRuntimeAlpha["isDirty"] {
+	public get isDirty(): IFluidDataStoreRuntime["isDirty"] {
```
anthony-murphy:
```suggestion
	public get isDirty(): boolean {
```

packages/runtime/runtime-utils/src/dataStoreHelpers.ts
```diff
@@ -139,23 +131,3 @@ export function createResponseError(
 		headers,
 	};
 }
-
-/**
- * Converts types to their alpha counterparts to expose alpha functionality.
- * @legacy @alpha
- */
-export function asLegacyAlpha(runtime: IContainerRuntimeBase): ContainerRuntimeBaseAlpha;
```
anthony-murphy, line 147: i thought we might keep some of the generic alpha things around, so future alpha features could reuse the layers

anthony-murphy: it might be annoying to do, as it will require empty interfaces for the alpha types which i know eslint complains about, so not required

packages/runtime/test-runtime-utils/src/mocks.ts
```diff
@@ -925,6 +925,8 @@ export class MockFluidDataStoreRuntime
 		return this;
 	}
 
+	public readonly inStagingMode: boolean = false;
+	public readonly isDirty: boolean = false;
```
anthony-murphy: Deep Review: `inStagingMode` and `isDirty` are hardcoded to `false` here. The real implementation delegates `inStagingMode` to the container runtime and computes `isDirty` from pending state. This follows the mock's existing pattern of minimal defaults, but could mask dirty-state bugs in tests that exercise staging mode. Consider making these configurable or allowing them to delegate to the mock container runtime. Not blocking.

anthony-murphy:
## Deep Review

This is a clean alpha-to-beta API promotion with no correctness or runtime defects identified. All higher-severity concerns were investigated and confirmed to be non-issues on factual grounds. Six polish items remain.

### Findings

1. **Misleading `IContainerRuntimeBaseInternal` documentation** — The JSDoc says "Experimental extension of IContainerRuntimeBase to support staging mode" but staging mode is now on `IContainerRuntimeBase` itself at `@beta`. The internal interface exists for `StageControlsInternal` (with the `squash` option), not for staging mode. *(see inline comment on stagingMode.ts:51)*

2. **Redundant optional chaining in `inStagingMode` getter** *(nit)* — `this.dataStoreContext.containerRuntime?.inStagingMode` uses `?.` but `containerRuntime` is typed as non-optional. TypeScript correctly infers `boolean` here so it's not a type-safety hole, but removing the `?.` would be clearer. *(see inline suggestion on dataStoreRuntime.ts)*

3. **Beta promotion lacks explicit readiness rationale** — The original staging mode PR (#23783) described the feature as "experimental" and "not fully implemented across all DDSes." A brief "what changed since then" note in the PR description would help reviewers understand why beta is appropriate now. This is a communication suggestion, not a code defect.

4. **`loadContainerRuntimeAlpha` name is vestigially alpha** *(nit)* — After merging alpha staging APIs into `IContainerRuntimeBase`, `loadContainerRuntimeAlpha` returns a type functionally identical to the non-alpha load path. The "Alpha" suffix no longer signals any alpha-only capability. Consider deprecating it in a follow-up PR or adding a comment explaining why the name persists.

5. **`MockFluidDataStoreRuntime` hardcodes `inStagingMode` and `isDirty` to `false`** — Follows the mock's existing pattern of minimal defaults, but could mask dirty-state bugs in tests that exercise staging mode. `MockContainerRuntime` also lacks `enterStagingMode()` / `inStagingMode` — no type violation since it doesn't declare `implements IContainerRuntimeBase`, but a follow-up to make these configurable would improve test fidelity. *(see inline comment on mocks.ts)*

6. **Layer-compat signaling for staging mode** — PR #24166 established `supportedFeatures`/`requiredFeatures` for cross-layer compatibility. This PR adds required members to `IContainerRuntimeBase` without a corresponding `supportedFeatures` entry. On investigation, generation-based layer compatibility validation is the primary safety mechanism and would reject incompatible pairings — `supportedFeatures` is used sparingly for specific feature-gated differences. If staging-mode capability should ever be optional across the Runtime↔DataStore boundary, a `supportedFeatures` entry would be appropriate. Currently a design consideration, not a defect.

### Path to Ready

To move from 8/10 → 9-10/10:

- [ ] Add a "why now" sentence to the PR description explaining what changed since the experimental phase to justify beta promotion (Finding 3)
- [ ] Update `IContainerRuntimeBaseInternal` JSDoc to clarify it exists for internal commit options, not for staging mode itself (Finding 1)
- [ ] Address anthony-murphy's resolved inline suggestions: `isDirty` getter visibility (thread on dataStoreRuntime.ts)
- [ ] Get markfields sign-off — as the original staging mode author and requested reviewer, his approval resolves the "is this beta-ready?" question
- [ ] Optionally: remove redundant `?.` in `inStagingMode` getter for clarity (Finding 2)

### Context for Reviewers

- **Staging mode history**: Introduced by markfields (#23783) as explicitly experimental with known gaps in DDS rollback support. Subsequently expanded (#24506 — readonlyInStagingMode policies, #24524 — squash control). This PR is the formal graduation to beta.
- **Follows `isReadOnly` precedent**: anthony-murphy (PR #24410) previously promoted `isReadOnly` from a delegated/optional pattern to a direct property on `IFluidDataStoreRuntime` using the same approach. markfields reviewed and approved that pattern.
- **`@sealed` limits blast radius**: `IContainerRuntimeBase` and `IFluidDataStoreRuntime` are `@sealed`, meaning only framework-provided implementations exist. Adding required members does not break external consumers.
- **`asLegacyAlpha` removal is expected**: The `@alpha` tier carries no stability guarantees in FluidFramework's API model. The removed cast helpers were a deliberate access gate for experimental features — their removal signals confidence the API surface is stable.

### Suggested Description Enhancement

Consider adding to the PR description:

> **Why promote to beta now?** *[Brief explanation of what changed since the experimental phase — e.g., DDS coverage improvements, stabilized squash semantics, e2e test coverage.]*
>
> **Note:** The `squash` commit option remains internal-only at beta, available through `StageControlsInternal` for framework use but not yet part of the public beta API.

<details><summary>Out of scope findings (2)</summary>

- **`isDirty` naming overlap** — `datastore-definitions/src/dataStoreRuntime.ts` & `container-runtime-definitions/src/containerRuntime.ts` — Same property name tracks different semantics at different layers (uncommitted local changes vs pending unacknowledged changes). Pre-existing; ensure documentation distinguishes the two.
- **`StageControls` type split from co-located staging types** — `runtime-definitions/src/dataStoreContext.ts` & `runtime-definitions/src/stagingMode.ts` — `StageControls` moved to `dataStoreContext.ts` while `StageControlsInternal` stays in `stagingMode.ts`. Reasonable since `StageControls` is a leaf type; document if it causes confusion.

</details>

sonalideshpandemsft:
Deep review: Update IContainerRuntimeBaseInternal JSDoc to clarify it exists for internal commit options, not for staging mode itself 

Doesn't commit options exists only for staging mode?

packages/runtime/runtime-definitions/src/stagingMode.ts
```diff
@@ -44,61 +44,14 @@ export interface StageControlsInternal extends StageControlsAlpha {
...
+ * returns {@link StageControlsInternal} (which exposes internal commit options such as squash)
  * @internal
  */
-export interface IContainerRuntimeBaseInternal extends ContainerRuntimeBaseAlpha {
+export interface IContainerRuntimeBaseInternal extends IContainerRuntimeBase {
 	/**
 	 * Enters staging mode, allowing changes to be staged before being committed or discarded.
 	 * @returns Controls for committing or discarding staged changes.
 	 */
 	enterStagingMode(): StageControlsInternal;
```
markfields, line 56: I was still hoping we only expose this on `IRuntime` to emphasize that this is a container-wide concept, not something to be accessed easily from within the Container Runtime.  Following up on the other PR where this will be used about this.

anthony-murphy: IRuntime is the loader/container/host facing interface. I think you mean IContainerRuntime, which is the inward facing

anthony-murphy: That said, it might make it too hard to access, as most consumers only get IContainerRuntimeBase, and i'd rather have more exposure, than force the need for casting

packages/runtime/datastore-definitions/src/dataStoreRuntime.ts
```diff
@@ -190,14 +190,15 @@ export interface IFluidDataStoreRuntime
...
 	readonly activeLocalOperationActivity?: "applyStashed" | "rollback" | undefined;
-}
 
-/**
- * @legacy @alpha
- * @sealed
- */
-export interface IFluidDataStoreRuntimeAlpha extends IFluidDataStoreRuntime {
+	/**
+	 * Indicates whether the container is currently in staging mode.
```
Josmithr, line 195: Nit: do we have any other documentation we can point to that explains what "staging mode" is?

packages/runtime/datastore/src/dataStoreRuntime.ts
```diff
@@ -545,16 +546,16 @@ export class FluidDataStoreRuntime
...
-	private get inStagingMode(): IFluidDataStoreRuntimeAlpha["inStagingMode"] {
-		return asLegacyAlpha(this.dataStoreContext.containerRuntime)?.inStagingMode;
+	public get inStagingMode(): boolean {
+		return this.dataStoreContext.containerRuntime.inStagingMode;
 	}
 
 	/**
-	 * Implementation of IFluidDataStoreRuntimeAlpha.isDirty
+	 * Implementation of IFluidDataStoreRuntime.isDirty
 	 */
```
Josmithr, line 557: We should probably remove these comments. They mask whatever documentation already exists on the interface (which would otherwise get inherited automatically in intellisense and such).

packages/runtime/runtime-definitions/src/dataStoreContext.ts
```diff
@@ -180,6 +180,27 @@ export interface IDataStore {
 	readonly entryPoint: IFluidHandleInternal<FluidObject>;
 }
 
+/**
+ * Controls for managing staged changes in staging mode.
```
Josmithr, line 184: Nit: do we have other documentation somewhere that explains what "staging mode" is that we could reference here (and from other relevant API members)?

sonalideshpandemsft: we have docs for enterStagingMode on IContainerRuntimeBase. should i reference that here?

packages/runtime/datastore-definitions/src/dataStoreRuntime.ts
sonalideshpandemsft: There is documentation on IContainerRuntimeBase. Will that be ok?

packages/runtime/runtime-definitions/src/dataStoreContext.ts
anthony-murphy: Deep Review: Building on Josmithr's point — beyond `@see` cross-references, the beta-promoted JSDoc should carry explicit caveats about known limitations. When staging mode was introduced in PR #23783, it was explicitly experimental with documented gaps: DDS rollback needs proper implementation per DDS (meaning `discard()` may throw for some DDS types), and squash-rebase semantics during commit are incomplete. Additionally, `readonlyInStagingMode` makes data stores *appear* readonly but does not enforce strict write prevention (per ChumpChief's feedback on #24506).

Recommended additions to the JSDoc on `enterStagingMode()`, `StageControls`, `inStagingMode`, `isDirty`, and `readonlyInStagingMode`:
1. DDS rollback support may be incomplete — `discard()` may throw for some DDS implementations
2. Squash-rebase semantics during commit are not yet fully specified
3. `readonlyInStagingMode` provides best-effort readonly appearance, not strict enforcement

The `@see` references the author has agreed to add are a great start — adding these limitation notes alongside would give consumers a complete picture of what `@beta` means for this API surface.

packages/runtime/datastore/src/dataStoreRuntime.ts
```diff
@@ -544,17 +547,11 @@ export class FluidDataStoreRuntime
 		this.minVersionForCollab = this.dataStoreContext.minVersionForCollab;
 	}
```
anthony-murphy, line 546: Deep Review: This public getter accesses `this.dataStoreContext.containerRuntime.inStagingMode` without a null guard, but the PR's own telemetry fix (lines 427–430) adds optional chaining with a comment acknowledging that `containerRuntime` can be `undefined` when `MonitoringContext` accesses `inStagingMode`. The previous alpha getter used `asLegacyAlpha(this.dataStoreContext.containerRuntime)?.inStagingMode` with optional chaining, so this is a behavioral regression.

Additionally, `MockFluidDataStoreContext` passes `undefined as any` for `containerRuntime`, confirming the gap exists in practice.

Suggested fix (consistent with the defensive pattern already in the telemetry lambda):
```suggestion
	return this.dataStoreContext.containerRuntime?.inStagingMode ?? false;
```

Alternatively, if the intent is to fail fast on contract violations, document the lifecycle invariant that `containerRuntime` is always initialized before this getter is accessible, and update `MockFluidDataStoreContext` to provide a real mock.

packages/runtime/runtime-utils/src/dataStoreHelpers.ts
```diff
@@ -145,6 +145,7 @@ export function createResponseError(
  * @legacy @alpha
  */
 export function asLegacyAlpha(runtime: IContainerRuntimeBase): ContainerRuntimeBaseAlpha;
+
```
markfields: Revert this file?  Or shouldn't we just delete `asLegacyAlpha` here since it's not exported anymore (or change its tags)

packages/runtime/runtime-definitions/src/dataStoreContext.ts
```diff
@@ -290,6 +313,21 @@ export interface IContainerRuntimeBase extends IEventProvider<IContainerRuntimeB
...
 	): Promise<{ snapshotTree: ISnapshotTree; sequenceNumber: number }>;
+
+	/**
+	 * Enter Staging Mode, such that ops submitted to the ContainerRuntime will not be sent to the ordering service.
+	 * To exit Staging Mode, call either discardChanges or commitChanges on the Stage Controls returned from this method.
+	 *
+	 * @returns Controls for committing or discarding staged changes.
+	 */
+	enterStagingMode(): StageControls;
+	/**
```
markfields, line 336: nit
```suggestion

	/**
```

packages/runtime/runtime-definitions/api-report/runtime-definitions.legacy.alpha.api.md
```diff
@@ -421,12 +421,16 @@ export interface OpAttributionKey {
...
-// @alpha @sealed @legacy
-export interface StageControlsAlpha {
+// @beta @sealed @legacy
+export interface StageControls {
     readonly commitChanges: () => void;
     readonly discardChanges: () => void;
 }
 
+// @alpha @sealed @deprecated @legacy
+export interface StageControlsAlpha extends StageControls {
```
markfields, line 431: Why are we marking as deprecated instead of just removing it immediately?

anthony-murphy: we have a bunch of prototyping that is easier to maintain if we keep it working, we'll remove after the next release so we have time to port the prototypes

packages/runtime/datastore/src/dataStoreRuntime.ts
```diff
@@ -425,7 +423,12 @@ export class FluidDataStoreRuntime
 			namespace: "FluidDataStoreRuntime",
 			properties: {
 				all: { dataStoreId: uuid(), dataStoreVersion: pkgVersion },
-				error: { inStagingMode: () => this.inStagingMode, isDirty: () => this.isDirty },
+				error: {
+					// MonitoringContext accesses inStagingMode when containerRuntime is undefined,
```
markfields: This is not ideal, the variable is typed without `undefined`. And I'm confused how it even happens because looking at the implementation, `FluidDataStoreContext._containerRuntime` is set in the first line of the constructor.

Not blocking the PR, but worth taking a few minutes to look into with Claude IMO, even after you merge this.

packages/runtime/datastore/src/dataStoreRuntime.ts
markfields: Ok if the only failure is from MockFluidDataStoreContext, then better to update that to pass `{ inStagingMode: false } as any` or something, and let this production code stay true to the types.

packages/runtime/datastore-definitions/src/dataStoreRuntime.ts
```diff
@@ -190,16 +190,24 @@ export interface IFluidDataStoreRuntime
...
 
 /**
  * @legacy @alpha
  * @sealed
  */
-export interface IFluidDataStoreRuntimeAlpha extends IFluidDataStoreRuntime {
-	readonly inStagingMode: boolean;
-	readonly isDirty: boolean;
-}
+export interface IFluidDataStoreRuntimeAlpha extends IFluidDataStoreRuntime {}
```
markfields, line 212: Can't we just remove this?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -840,7 +847,7 @@ export async function loadContainerRuntime(
  * @legacy @alpha
  */
 export async function loadContainerRuntimeAlpha(params: LoadContainerRuntimeParams): Promise<{
-	runtime: IContainerRuntime & ContainerRuntimeBaseAlpha & IRuntime;
+	runtime: IContainerRuntime & IContainerRuntimeBase & IRuntime;
```
markfields:
```suggestion
	runtime: IContainerRuntime & IRuntime;
```

markfields: (and remove the import above)

packages/runtime/runtime-definitions/src/stagingMode.ts
markfields: (discussed offline, it's impractical for consumers to use it off `IContainerRuntime`, sadly)

markfields (approved):

packages/runtime/datastore-definitions/src/dataStoreRuntime.ts
anthony-murphy: i wanted to keep it. otherwise every alpha feature will just need to re-add this type of layering, which isn't a huge deal, but i think its easier to do with a bit of structure like this in place.

anthony-murphy (approved):

packages/runtime/datastore/src/dataStoreRuntime.ts
anthony-murphy: Deep Review: Agreed — markfields' suggestion is the better approach. Since `containerRuntime` is typed as non-optional on `IFluidDataStoreContext`, the production code should stay true to the types. Updating `MockFluidDataStoreContext` to pass `{ inStagingMode: false } as any` instead of `undefined as any` fixes the test gap without adding unnecessary defensive coding.

sonalideshpandemsft:
`Add @deprecated tags (or remove, per team decision) on emptied ContainerRuntimeBaseAlpha, IFluidDataStoreRuntimeAlpha, and asLegacyAlpha overloads to resolve the within-PR inconsistency → moves to 8/10`: Deprecated tags for alpha interfaces is not required as future alpha features can reuse them

sonalideshpandemsft:
Get final sign-off from markfields on unresolved threads - @anthony-murphy @markfields looking for re-approval

packages/runtime/datastore/src/dataStoreRuntime.ts
sonalideshpandemsft: done

anthony-murphy:
## Deep Review

This PR cleanly promotes staging mode APIs from `@alpha` to `@beta` by merging alpha interfaces into their base counterparts, creating a new `StageControls` beta interface, and removing `asLegacyAlpha` cast-site usage. All significant findings from prior review rounds have been addressed — the null-guard direction was resolved per markfields' feedback, and the team has confirmed the design intent for keeping empty alpha interfaces as future extension points.

### Findings

- **Mock `inStagingMode` and `isDirty` hardcoded to `false`** — `packages/runtime/test-runtime-utils/src/mocks.ts` — (inline thread, resolved) Non-blocking; follows existing mock patterns but could mask staging-mode bugs in tests.
- **`inStagingMode` accessed without cross-layer feature gating** — `packages/runtime/datastore/src/dataStoreRuntime.ts` / `packages/runtime/container-runtime/src/dataStore.ts` — Unlike `isReadOnly` which uses `contextSupportsFeature()`, `inStagingMode` is accessed directly. This is fine: the required-interface contract on `IContainerRuntimeBase` is the correct enforcement mechanism for promoted beta APIs, and the failure mode in mixed-version deployments is benign (`undefined` → falsy → treated as not in staging mode). No action needed.
- **Inconsistent deprecation of emptied alpha interfaces** — Resolved by team decision. `StageControlsAlpha` is deprecated because it has a direct beta replacement, while `ContainerRuntimeBaseAlpha` and `IFluidDataStoreRuntimeAlpha` are intentionally kept as extension points for future alpha features. Thanks for the clarification, @sonalideshpandemsft.

### Path to Ready

- [x] **Confirm JSDoc updates are complete** — Josmithr's threads about `@see` cross-references to `enterStagingMode` and `{@inheritDoc}` removal appear addressed; verify in latest commit → moves to 9/10
- [x] **Update `ContainerRuntimeBaseAlpha` JSDoc** — Its doc comment still references staging mode specifically; a one-line update to reflect its new role as a general alpha extension point would prevent consumer confusion → moves to 9/10
- [x] **Get re-approval from markfields** — All threads are resolved; final sign-off confirms alignment → moves to 9/10
- [x] **CI green** — Confirm all pipeline checks pass on latest commit → moves to 10/10

### Context

Staging mode was introduced as explicitly experimental with known gaps (DDS rollback, squash-rebase semantics). This promotion is an API-accessibility change for Loop consumption, not a feature-completeness signal. The `asLegacyAlpha` helpers are being kept for future alpha features per anthony-murphy's direction.

<details><summary>Out of scope findings (3)</summary>

- **Beta promotion vs experimental positioning** — `packages/runtime/runtime-definitions/src/dataStoreContext.ts` — Product decision; ensure beta docs call out known limitations (DDS rollback may throw, squash-rebase incomplete).
- **`readonlyInStagingMode` strict-mode design question** — `packages/runtime/runtime-definitions/src/dataStoreContext.ts` — Pre-existing question from PR #24506; no action needed for this PR.
- **`asLegacyAlpha` not deprecated despite all internal usage removed** — `packages/runtime/runtime-utils/src/dataStoreHelpers.ts` — Team decision to keep for prototyping; consider removing after next release.

</details>

packages/runtime/datastore/src/dataStoreRuntime.ts
anthony-murphy: Deep Review: Thanks for the fix! Updating `MockFluidDataStoreContext` to provide `{ inStagingMode: false } as any` is the right approach — keeps production code true to the types as markfields suggested.

anthony-murphy (approved):

PR #27113, opened 2026-04-21
fix(ci): switch review fleet to JSON output and remove fix truncation

tylerbutler:
## Description

The review fleet's consolidated report was silently truncating the `fix` field of each finding at 200 characters with no indication that content was cut.

This removes that limit and replaces the underlying output format — a plaintext one-liner format parsed by regex — with structured JSON from each reviewer agent:

- Each reviewer prompt now instructs the agent to write `{ "findings": [...] }` to its output file instead of `[SEVERITY] file:line — desc — fix` lines.
- `consolidate_reviews.py` parses JSON instead of regex. Malformed agent output now fails loudly per file (warning printed, reviewer skipped) rather than silently dropping lines that don't match the pattern.
- `{ "findings": [] }` replaces the `<\!-- NO_ISSUES_FOUND -->` sentinel.

## Reviewer Guidance

### DIFF
--- a/.github/prompts/reviewers/api-compatibility.md
+++ b/.github/prompts/reviewers/api-compatibility.md
@@ -67,17 +67,30 @@ API Quality findings are **promoted +1 level**:
 
 ## Output Format
 
-Write your findings to `review-api-compatibility.md`. Use this exact format for each finding:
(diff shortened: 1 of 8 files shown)


jumyhre (approved):

tylerbutler:
Thanks for the review, @jumyhre!

PR #27119, opened 2026-04-21
build: install @fluid-tools/build-cli in AI-enabled codespace

tylerbutler:
## Description

Installs the `flub` CLI (`@fluid-tools/build-cli`) globally from npm into the stable AI-enabled codespace so agents have it available from first shell without any manual setup.

- New `scripts/codespace-setup/install-build-cli.sh` runs `pnpm add -g @fluid-tools/build-cli` during `onCreateCommand`.
- `Dockerfile` sets `PNPM_HOME=/home/node/.local/share/pnpm` and prepends it to `PATH` so pnpm-global shims are invokable from every shell.
- `.devcontainer/ai-agent/devcontainer.json` chains the new script after `on-create.sh` and bumps `prebuild-version` to `1.0.4` to force a prebuild rebuild.

The unstable `ai-agent-insiders` profile is unchanged and continues to build `flub` from the local `build-tools` workspace via `install-flub.sh`.

## Reviewer Guidance

### DIFF
--- a/scripts/codespace-setup/install-build-cli.sh
+++ b/scripts/codespace-setup/install-build-cli.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+# Installs @fluid-tools/build-cli (the `flub` CLI) globally from npm using pnpm.
+#
+# Prerequisites: node + corepack must already be set up (on-create.sh), and
+# PNPM_HOME must be on PATH (set in .devcontainer/Dockerfile).
+
+set -euo pipefail
+
+echo "Installing @fluid-tools/build-cli globally..."
+pnpm add -g @fluid-tools/build-cli
+
+echo "flub installed: $(which flub)"
+flub --version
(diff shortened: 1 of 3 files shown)


scripts/codespace-setup/install-build-cli.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# Installs @fluid-tools/build-cli (the `flub` CLI) globally from npm using pnpm.
+#
+# Prerequisites: node + corepack must already be set up (on-create.sh), and
```
noencke, line 4: Should we just move this step inside of on-create, so that this dependency is more localized and we don't need this extra file?

noencke (approved):

scripts/codespace-setup/install-build-cli.sh
tylerbutler: Using the individual files makes things a bit more composable.

PR #27125, opened 2026-04-22
feat(react): Support scoped undo/redo via transaction labels

Josmithr:
Adds support for scoped undo/redo stack management via labels.

## Undo/Redo Manager
createUndoRedo(branch) returns an UndoRedo manager that tracks commits on the given tree branch. One manager per branch is enforced — creating a second throws immediately.

undo, redo, canUndo, and canRedo each accept an optional symbol label:

With a label: targets the most recent commit whose label set contains that symbol, skipping non-matching commits.
Without a label: targets the most recent commit regardless of labels.
Labels are assigned per-transaction via runTransaction's { label: symbol } option. This lets multiple editors share a single manager while keeping independent undo/redo histories: each editor declares its own symbol, stamps its transactions with it, and passes it to undo/redo.

Redo invalidation is label-selective: when a new commit arrives, only redo entries whose label sets overlap with the new commit's labels are discarded. An anonymous (unlabeled) commit clears only anonymous redo entries; labeled redo entries are left intact. This means one editor making a new change will not clear another editor's redo stack.

## Text editor component updates

The `PlainTextView`, `QuillMainView` and `FormattedMainView` components have been updated to accept an optional pair of parameters: an undo/redo manager, and an optional label to apply to edits. The editors will scope their undo/redo stacks using this label (if supplied).

For the plain text views, a toolbar is now included if-and-only if an undo/redo manager is supplied. This toolbar includes undo/redo controls.

The formatted view now displays undo/redo controls in its toolbar if-and-only-if an undo/redo manager is supplied.

## `text-editor` application updates

Adds editor-level undo/redo controls, as well as top-level application undo/redo controls.

image


packages/framework/react/src/undoRedo.ts
```diff
@@ -3,115 +3,315 @@
...
+		const entry = this.#redoStack[index] ?? oob();
+		this.#pendingOperation = { kind: "redo", labels: entry.labels };
+		try {
+			entry.revertible.revert();
+		} finally {
+			this.#pendingOperation = undefined;
+		}
+		// Only remove from the stack after a successful revert.
+		// If revert() throws, the entry stays so the user can retry.
+		this.#redoStack.splice(index, 1);
```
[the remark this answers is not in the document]
Josmithr: Fixed — both `undo()` and `redo()` now call `revert(false)` so the revertible stays valid if it throws, then `dispose()` + splice only after success.

packages/framework/react/src/undoRedo.ts
```diff
@@ -3,115 +3,315 @@
...
+		const entry = this.#undoStack[index] ?? oob();
+		this.#pendingOperation = { kind: "undo", labels: entry.labels };
+		try {
+			entry.revertible.revert();
+		} finally {
+			this.#pendingOperation = undefined;
+		}
+		// Only remove from the stack after a successful revert.
+		// If revert() throws, the entry stays so the user can retry.
+		this.#undoStack.splice(index, 1);
```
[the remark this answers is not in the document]
Josmithr: Fixed — both `undo()` and `redo()` now call `revert(false)` so the revertible stays valid if it throws, then `dispose()` + splice only after success.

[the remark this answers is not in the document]
Josmithr: This was a documentation bug rather than a missing implementation — the claim that `undo(label)` discards overlapping redo entries doesn't have clear correct semantics (an undo moves an entry to redo; it doesn't diverge history the way a new user commit does). The erroneous sentence has been removed from the interface docs.

packages/framework/quill-react/src/plain/quillView.tsx
```diff
@@ -7,155 +7,212 @@ import {
...
+					<>
+						<button
+							type="button"
+							className="ql-undo"
+							disabled={undoRedo?.manager.canUndo(undoRedo.transactionLabel) !== true}
+							onClick={() => undoRedo?.manager.undo(undoRedo.transactionLabel)}
+						/>
+						<button
+							type="button"
+							className="ql-redo"
```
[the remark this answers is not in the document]
Josmithr: Fixed — `aria-label="Undo"` and `aria-label="Redo"` added to both buttons.

packages/framework/react/src/undoRedo.ts
```diff
@@ -3,115 +3,315 @@
...
+				}
+
+				const overlaps =
+					symbolLabels.size === 0
+						? entry.labels.size === 0
+						: [...symbolLabels].some((l) => entry.labels.has(l));
+				if (overlaps) {
+					entry.revertible.dispose();
+					this.#redoStack.splice(i, 1);
+				}
```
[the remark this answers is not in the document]
Josmithr: Fixed — replaced the `[...symbolLabels].some(...)` spread with a `labelsOverlap()` helper that iterates with `for...of`, avoiding the per-entry allocation.

packages/framework/react/src/text/plain/plainTextView.tsx
```diff
@@ -88,6 +135,49 @@ const PlainTextEditorView = withMemoizedTreeObservations(
...
+							onClick={() => undoRedo.manager.undo(undoRedo.transactionLabel)}
+							title="Undo"
+						/>
+						<button
+							type="button"
+							className="pt-redo"
+							disabled={!undoRedo.manager.canRedo(undoRedo.transactionLabel)}
+							onClick={() => undoRedo.manager.redo(undoRedo.transactionLabel)}
+							title="Redo"
+						/>
```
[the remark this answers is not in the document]
Josmithr: Fixed — `aria-label="Undo"` and `aria-label="Redo"` added to both buttons (alongside the existing `title` attributes).

packages/framework/quill-react/src/plain/quillView.tsx
```diff
@@ -7,155 +7,212 @@ import {
...
+			return () => {
+				quill.off("text-change", handleTextChange);
+				quillRef.current = undefined;
+				editor.innerHTML = "";
+				editor.className = "";
+				setUndoRedoContainer(undefined);
+				toolbar.remove();
+			};
+			// In React strict mode, effects run twice. The `!quillRef.current` check above
+			// makes the second call a no-op, preventing double-initialization of Quill.
```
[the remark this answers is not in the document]
Josmithr: Fixed — the comment now correctly describes React StrictMode's mount/unmount/remount cycle ("initialize/cleanup/initialize during development") rather than the previous incorrect description.

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -604,56 +601,42 @@ const FormattedTextEditorView = forwardRef<
...
+					<>
+						<button
+							type="button"
+							className="ql-undo"
+							disabled={undoRedo?.manager.canUndo(undoRedo.transactionLabel) !== true}
+							onClick={() => undoRedo?.manager.undo(undoRedo.transactionLabel)}
+						/>
+						<button
+							type="button"
+							className="ql-redo"
```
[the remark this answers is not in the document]
Josmithr: Fixed — `aria-label="Undo"` and `aria-label="Redo"` added to both buttons in the formatted editor (alongside the existing `title` attributes).

packages/framework/react/package.json
```diff
@@ -103,6 +103,7 @@
 		"@fluidframework/runtime-definitions": "workspace:~",
 		"@fluidframework/shared-object-base": "workspace:~",
 		"@fluidframework/tree": "workspace:~",
+		"@fluidframework/telemetry-utils": "workspace:~",
```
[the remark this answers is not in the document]
Josmithr: Fixed — `@fluidframework/telemetry-utils` is now sorted before `@fluidframework/tree` in `dependencies`.

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -604,56 +601,42 @@ const FormattedTextEditorView = forwardRef<
...
 
-			// Only update if there are actual differences
 			if (diff.ops.length > 0) {
 				isUpdating.current = true;
-
-				// Apply only the diff for surgical updates (better cursor preservation)
 				quillRef.current.updateContents(diff.ops);
-
 				isUpdating.current = false;
 			}
```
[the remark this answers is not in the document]
Josmithr: This is a pre-existing pattern in the codebase (not introduced by this PR) — Quill's content is synced during the render phase when an external tree change is detected. Addressing this would require a larger refactor (e.g. moving the sync into a `useLayoutEffect`) and is out of scope for this PR.

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -289,7 +310,62 @@ const UserPanel: FC<{
...
+						}}
+					>
+						↶
+					</button>
+					<button
+						type="button"
+						disabled={!manager.canRedo()}
+						onClick={() => manager.redo()}
+						title="Redo"
+						style={{
```
CraigMacomber, line 344: I don't really know the best conventions for styling our examples, but this seems like the kind of stuff that would be in CSS. Maybe you can use some existing styles from our quill CSS file, or make a new CSS file?

Also seems like the common parts should be deduplicated with the button above.

Josmithr: The CSS suggestion is probably a good one. But the existing code uses inline formatting like this, so I'm inclined to leave this alone for this PR. I could have Claude migrate the package to tailwind (which is what our Fluid Examples repo uses) or something in a separate PR.

Deduping I can do now though.

CraigMacomber: We do already have a css file (from quill): if you are trying to match it, you might want to use the styles it defines (so if we change theme things still look right among other things)

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -220,48 +223,66 @@ async function initFluid(): Promise<DualUserViews> {
 	};
 }
 
+const plainTextareaEditLabel = Symbol("plain-textarea");
+const plainQuillEditLabel = Symbol("plain-quill");
+const formattedEditLabel = Symbol("formatted");
+
 const viewLabels = {
```
CraigMacomber, line 226: I feel like this could use a doc comment.

Before this change, this was a declarative lists of kinds of views we can make. Now it is a concreate list of views we can have.

Before if you made two instances of the same kind of view that was valid, and no it is now since the editLabel baked into this list.

It also might make sense to just use a different pattern here, as this is likely not the approach a lot of applications would want to take, as having multiple of the same kind of editor view with separate undo/redo stack is likely common.

Maybe using the identity of the text node being edited as the label would be a better pattern?

CraigMacomber: I guess if you have multiple editors for the same node, we do want to separate those however. Maybe some combination of the node identity and the portion of the app UI doing the editing? If only we had some way a node to map multiple different derived identities. Perhaps two labels: one with the node, and one with the logical editor involved?

I guess maybe this actually works that way (I haven't look at the details too much: some docs on how these identities work would be nice)

packages/framework/react/src/test/undoRedo.test.ts
```diff
@@ -5,90 +5,515 @@
...
+	independentView,
+	type TreeBranchAlpha,
+	type TreeViewAlpha,
+	SchemaFactory,
+	TreeViewConfiguration,
+} from "@fluidframework/tree/alpha";
+
+import { createUndoRedo } from "../undoRedo.js";
+
+// ---------------------------------------------------------------------------
```
Josmithr: TODO: clean this up

packages/framework/react/src/text/plain/plainTextView.tsx
```diff
@@ -3,34 +3,74 @@
...
- * @remarks
- * Uses {@link @fluidframework/tree#TextAsTree.Tree} for the data-model and an HTML textarea for the UI.
+ * The `undoRedo` prop shape shared by all editor components that integrate with {@link UndoRedo}.
+ *
+ * @example
+ * ```tsx
+ * const titleLabel = Symbol("title-editor");
+ * const prop: UndoRedoProp = { manager, editLabel: titleLabel };
+ * <PlainTextMainView root={root} undoRedo={prop} />
+ * ```
```
Josmithr: TODO: this isn't useful. Remove it.

packages/framework/react/src/text/plain/plainTextView.tsx
```diff
@@ -3,34 +3,74 @@
...
+ *
+ * @internal
+ */
+export interface UndoRedoProp {
```
Josmithr: TODO: maybe "ScopedEditorProperties" or something? This isn't specifically about undo/redo, it's about scoping edits more generally.

Josmithr: Also update docs

examples/data-objects/text-editor/src/app.tsx
Josmithr: > Before this change, this was a declarative lists of kinds of views we can make. Now it is a concreate list of views we can have.
>
> Before if you made two instances of the same kind of view that was valid, and no it is now since the editLabel baked into this list.

I can restore this, if we think it was a valuable degree of freedom. But I don't know how important it is. We could also restore it if/when we want the example to have that sort of user experience.

> I guess if you have multiple editors for the same node, we do want to separate those however. Maybe some combination of the node identity and the portion of the app UI doing the editing? If only we had some way a node to map multiple different derived identities. Perhaps two labels: one with the node, and one with the logical editor involved?

This is already the case. Both our plain text editor and the quill plain text editor operate on the same node in the tree (but get separate undo/redo stacks). But including both the node ID and the "editor ID" in the label is probably a reasonable pattern. I can update it to do that.

packages/framework/react/src/undoRedo.ts
```diff
@@ -3,115 +3,333 @@
...
-			(commit: CommitMetadata, getRevertible?: RevertibleFactory) => {
-				if (getRevertible === undefined) return;
+const attachedBranches = new WeakSet<TreeBranchAlpha>();
+
+class UndoRedoManager implements UndoRedo {
+	readonly #undoStack: StackEntry[] = [];
+	readonly #redoStack: StackEntry[] = [];
+	readonly #unsubscribe: () => void;
+	readonly #branch: TreeBranchAlpha;
+	// Set synchronously around revert() calls so the changed event handler can attribute the
```
Josmithr: TODO: document why this is needed.

examples/data-objects/text-editor/src/app.tsx
Josmithr: This has been updated. We are no longer specifying static labels, so the original contract here is restored.

packages/framework/react/src/text/plain/plainTextView.tsx
Josmithr: Obsolete.

packages/framework/react/src/text/plain/plainTextView.tsx
Josmithr: Obsolete

examples/data-objects/text-editor/src/test/app.test.tsx
```diff
@@ -15,7 +15,7 @@ import { App, TextEditorRoot, treeConfig } from "../app.js";
 /**
  * Creates a TreeView for formatted text, initialized with the provided initial value.
  */
-function createFormattedTreeView(initialValue = ""): TreeView<typeof TextEditorRoot> {
+function createFormattedTreeView(initialValue = ""): TreeViewAlpha<typeof TextEditorRoot> {
```
Josmithr, line 18: Note for reviewers: transaction labeling is currently an `@alpha` feature, available through `TreeViewAlpha`. Usages of `TreeView` in the app have been replaced with `TreeViewAlpha` so we can leverage transaction labels for scoped undo/redo behaviors.

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -377,223 +379,219 @@ export function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp
...
-		undo: () => undoRedo?.undo(),
-		redo: () => undoRedo?.redo(),
-	}));
-
-	// Initialize Quill editor with formatting toolbar using Quill provided CSS
-	useEffect(() => {
-		if (!editorRef.current || quillRef.current) {
-			return;
-		}
+const FormattedTextEditorView = withMemoizedTreeObservations(
```
Josmithr, line 380: There are some unrelated changes made here to make the formatted view component mirror the plain text component more closely. I've extracted these changes into a separate PR here: https://github.com/microsoft/FluidFramework/pull/27206

packages/framework/react/src/undoRedo.ts
```diff
@@ -297,25 +289,68 @@ class UndoRedoManager implements UndoRedo {
...
-		assert(!this.#disposed, "Undo/redo manager is disposed.");
+	/**
+	 * Walks `stack` from the top and returns the index of the first entry matching `predicate`,
+	 * or `undefined` if none match.
+	 *
+	 * @param stack - The undo or redo stack to search.
+	 * @param predicate - Called for each entry from the top; the first entry for which it returns
+	 * `true` is selected.
+	 */
+	static #findLast(
```
CraigMacomber: This looks like an implementation of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLastIndex (except for the -1 instead of undefined): can we just use that instead?

If not, I'd recommend making this not private so that it can have tests. Might also make more sense as a free generic function since it isn't inherently tied to undo/redo and that would make testing easier.

packages/framework/react/src/undoRedo.ts
```diff
@@ -297,25 +289,68 @@ class UndoRedoManager implements UndoRedo {
...
+		stack: StackEntry[],
+		kind: "undo" | "redo",
+		predicate: (entry: StackEntry) => boolean,
+	): void {
+		const index = UndoRedoManager.#findLast(stack, predicate);
+		if (index === undefined) {
+			return;
+		}
+		const entry = stack[index] ?? oob();
+		this.#pendingOperation = { kind, labels: entry.labels };
```
CraigMacomber, line 338: I feel like this should be preceded by asserting no already pending operation

packages/framework/react/src/undoRedo.ts
```diff
@@ -297,25 +289,68 @@ class UndoRedoManager implements UndoRedo {
...
+		try {
+			// revert(false) reverts without auto-disposing, so the entry remains retryable if it throws.
+			entry.revertible.revert(false);
+		} finally {
+			this.#pendingOperation = undefined;
+		}
+		// Only remove and dispose after a successful revert.
+		// If revert() throws, the entry stays so the user can retry.
```
CraigMacomber: When does revert throw? Is there ever a way that someone can tell it left the tree in a valid state and thus know it's safe to retry?

packages/framework/react/src/undoRedo.ts
```diff
@@ -117,9 +120,9 @@ export interface UndoRedo {
 }
 
 /**
- * Determines if sets `a` and `b` share at least one symbol.
+ * Determines if sets `a` and `b` share at least one element.
  */
-function doLabelSetsOverlap(a: ReadonlySet<symbol>, b: ReadonlySet<symbol>): boolean {
+function doLabelSetsOverlap(a: ReadonlySet<unknown>, b: ReadonlySet<unknown>): boolean {
```
CraigMacomber: This is just the negation of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/isDisjointFrom

Can we refactor this to adopt the signature and name from `isDisjointFrom` and refer to it noting we should use it when possible (or just use that if we can).

Personally I'd be ok with this package targeting newer ES versions (like how its ESM only) as its currently alpha only and thus can't have any stable users. That API is available in node 22 (the oldest non EOL version) and all up to date browsers.

packages/framework/react/src/undoRedo.ts
```diff
@@ -130,16 +133,27 @@ function doLabelSetsOverlap(a: ReadonlySet<symbol>, b: ReadonlySet<symbol>): boo
 
 /**
  * One entry on the undo or redo stack, pairing a revertible with its commit's label set.
+ * @sealed
```
CraigMacomber: sealed doesn't really many anything on an interface that's not package exported.

packages/framework/react/src/undoRedo.ts
Josmithr: Yeah, that should be reasonable. Will update.

packages/framework/react/src/undoRedo.ts
Josmithr: Yeah, that's fair. I suppose we control that, so presumably it shouldn't. And violated constraints (currently) just silently no-op. Will update.

packages/framework/react/src/undoRedo.ts
Josmithr: It looks like support for `isDisjointFrom` was added in `ES2024`. The earliest version of TypeScript that supports that is `5.7`, and we're still on `5.4`. But I'll adopt that signature and leave a note.

packages/framework/react/src/undoRedo.ts
Josmithr: Actually, that API was added in ES2023. Since our documented client requirements still note that we support clients at ES2022 (and this documentation appears in the README for this package), it's probably best not to take a dependency on ES2023.

I don't disagree with your comment in the other thread about targeting newer versions in new/experimental libraries. But we should probably make sure our documentation matches.

For now, I've extracted this API and updated its signature to match, and left a note about migrating in the future.

packages/framework/react/src/utilities.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @privateRemarks
+ * This should be replaced with {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLastIndex | Array.prototype.findLastIndex}
+ * once we are able to target ES2023.
+ */
+export function findLastIndex<T>(
+	array: readonly T[],
+	predicate: (item: T) => boolean,
+): number {
+	for (let i = array.length - 1; i >= 0; i--) {
+		if (predicate(array[i] as T)) {
```
CraigMacomber: if you loop over array.enumerate() you get better type safety and can remove the cast.

packages/framework/react/src/test/undoRedo.test.ts
```diff
@@ -5,90 +5,513 @@
...
+			const view = createTree();
+			const manager = createUndoRedo(view);
+
+			view.runTransaction(() => {
+				view.root.value = 1;
+			});
+			assert.equal(manager.canUndo(), true);
+
+			manager.dispose();
+			assert.equal(manager.canUndo(), false);
```
CraigMacomber: Personally I think asserts testing Booleans should just do it in this style (see also many other places in these tests).
Not super important though (and note that if it returns something that might not be a Boolean, the tyle you have is better)

```suggestion
			assert(!manager.canUndo());
```

packages/framework/react/src/undoRedo.ts
```diff
@@ -3,115 +3,355 @@
...
+ * Returns a predicate that matches any stack entry whose label set contains `label`.
+ * When `label` is `undefined`, the predicate matches every entry (global operation).
+ *
+ * @param label - The value to match against, or `undefined` for a global match-all predicate.
+ */
+function labelPredicate(label: unknown): (entry: StackEntry) => boolean {
+	return label === undefined ? () => true : (entry) => entry.labels.has(label);
+}
+
+const attachedBranches = new WeakSet<TreeBranchAlpha>();
```
CraigMacomber, line 176: this could use a comment

packages/framework/react/src/undoRedo.ts
```diff
@@ -3,115 +3,355 @@
...
+			this.#pendingOperation = undefined;
 		}
 	}
 }
+
+/**
+ * Creates a {@link UndoRedo} manager that tracks commits on the given tree branch.
+ *
+ * @remarks
+ * A single instance must be created per tree branch. Passing a branch that already has an
```
CraigMacomber: this sounds like you must create one, not that you can't create more than one.

Josmithr: Fair.
```suggestion
 * At most, a single instance may be created per tree branch. Passing a branch that already has an
```

packages/framework/react/src/utilities.ts
Josmithr: Actually, in this case it looks like Claude was working around the fact that we have `noUncheckedIndexAccess` enabled. This just needed an `?? oob()`

CraigMacomber: I'd still prefer to remove the lack of type safty. A runtime check is better than nothing though, so you did make it better. array.enumerate can fix this if it were a forward iteration: I missed this was a backward iteration.

CraigMacomber (approved):

PR #27127, opened 2026-04-22
Add staging mode event

sonalideshpandemsft:
This change introduces a new event, `stagingModeChanged`, which notifies listeners whenever the staging mode state changes. The listener received a `StagingModeChangedEvent` which is a union.

- `{ inStagingMode: true }` - the runtime has entered staging mode.
 - `{ inStagingMode: false, commit: boolean }` - the runtime has exited staging mode. `commit` is `true` when changes were sent to the service, or `false` when they were discarded.

`orderSequentially` is updated to suppress `stagingModeChanged` events. Before this change, a single `enterStagingMode` / `exitStagingMode` call pair existed. Now:
1. `enterStagingMode` is split into a public one-liner and a private `enterStagingModeCore(silent: boolean)`.
2. `orderSequentially` passes `silent: true` -- both the enter and exit emit calls are gated by `if (!silent)`, so no `"stagingModeChanged"` event fires during `orderSequentially`.
3. The public `enterStagingMode` passes `silent: false` -- events fire normally for user initiated staging mode.


packages/runtime/runtime-definitions/api-report/runtime-definitions.legacy.alpha.api.md
```diff
@@ -108,6 +108,7 @@ export interface IContainerRuntimeBaseEvents extends IEvent {
     (event: "signal", listener: (message: IInboundSignalMessage, local: boolean) => void): any;
     // (undocumented)
     (event: "dispose", listener: () => void): any;
+    (event: "stagingModeChanged", listener: (canEnterStagingMode: boolean) => void): any;
```
sonalideshpandemsft: object for commit/rollback?

packages/runtime/runtime-definitions/api-report/runtime-definitions.legacy.alpha.api.md
```diff
@@ -108,6 +108,7 @@ export interface IContainerRuntimeBaseEvents extends IEvent {
     (event: "signal", listener: (message: IInboundSignalMessage, local: boolean) => void): any;
     // (undocumented)
     (event: "dispose", listener: () => void): any;
+    (event: "stagingModeChanged", listener: (inStagingMode: boolean) => void): any;
```
sonalideshpandemsft: I'm not sure if `inStagingMode` should be passed since it's already available on runtime

sonalideshpandemsft:
`On the enter path, if a listener throws during enterStagingMode(), the runtime is left in staging mode with no recovery. The exit path wraps its emit in a try block that closes the container on error. Was the asymmetry intentional?`

The exit path already wraps the listener in a try/catch block but enter does not. I can add the enter path in a try/catch as well but looking at how other listener errors are handled, I see [EventEmitterWithErrorHandling](https://github.com/microsoft/FluidFramework/blob/main/packages/utils/telemetry-utils/src/eventEmitterWithErrorHandling.ts#L20) used which catched the listener errors at `emit()` and routes to a single error handler. With this try/catch can be eliminated. Is there a reason why `ContainerRuntime` does not extend `EventEmitterWithErrorHandling<IContainerRuntimeEvents>`?

anthony-murphy:
Deep Review: Great question about `EventEmitterWithErrorHandling`. The current per-emit try/catch around the enter path is functional but ad-hoc — using `EventEmitterWithErrorHandling` would be a cleaner, centralized approach that's consistent with how other Fluid Framework components handle listener errors. That said, it would be a broader change to `ContainerRuntime`'s base class, so it's likely better addressed as a follow-up rather than in this PR. For now, adding try/catch symmetry on the enter path (matching the exit path) would be a reasonable incremental step if you'd like to keep the scope contained.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3745,6 +3749,13 @@ export class ContainerRuntime
```
anthony-murphy, line 3780: Deep Review: When `EnableRollback` is true and the runtime is not already in staging mode, `orderSequentially()` internally calls `enterStagingMode()` → `commitChanges()`/`discardChanges()`, all of which now emit `stagingModeChanged`. External listeners cannot distinguish user-initiated staging from this internal rollback bookkeeping.

Two concrete concerns:
1. **UI flicker** — consumers rendering staging indicators would see unexpected enter/exit flashes during `orderSequentially` calls.
2. **Future brittleness** — if `orderSequentially` is later reimplemented without staging mode, consumers calibrated to these implicit events break silently.

**Suggested options** (needs human judgment from reviewers with rollback domain knowledge):
- (a) Suppress the emit when staging mode is entered internally by `orderSequentially`
- (b) Add an optional `source: "explicit" | "orderSequentially"` field to `StagingModeChangedEvent` so consumers can filter
- (c) At minimum, add TSDoc guidance that the event may fire during `orderSequentially`

@markfields — you have the most context on the `orderSequentially` + rollback intersection from #10067. What's the intended contract here?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3712,6 +3712,10 @@ export class ContainerRuntime
 						this.channelCollection.notifyStagingMode(false);
 					},
 				);
+				this.emit("stagingModeChanged", {
+					inStagingMode: false,
+					commit: exitMethod === "commit",
+				});
```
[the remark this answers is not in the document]
anthony-murphy: Deep Review: The happy-path tests have been added — nice work! The enter, commit-exit, and discard-exit scenarios are now well covered.

One remaining gap: there are no tests exercising `stagingModeChanged` emission during `orderSequentially` when `EnableRollback` is true. That code path internally enters and exits staging mode, producing events that external listeners would receive. Once the design question about whether those events are intentional is resolved (see new inline thread), tests asserting the decided behavior for that path would close the coverage gap.

packages/runtime/container-runtime/src/containerRuntime.ts
sonalideshpandemsft: Suppressing `stagingModeChanged` event when staging mode is entered internally by orderSequentially

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3711,6 +3735,12 @@ export class ContainerRuntime
 						this.channelCollection.notifyStagingMode(false);
```
anthony-murphy, line 3740: Deep Review: The `silent` parameter passed into `enterStagingModeCore` is captured by the `exitStagingMode` closure returned here, controlling both enter and exit event suppression. The `orderSequentially` call site has a detailed block comment explaining *why* events are suppressed, but there's no corresponding comment at this closure capture point explaining that `silent` is intentionally captured from the enclosing scope.

Since multiple developers have refactored this area, a brief inline comment would help future readers understand the coupling — e.g.:
```suggestion
// `silent` is captured from the enclosing enterStagingModeCore call
// to suppress stagingModeChanged events on both enter and exit.
```

anthony-murphy (approved):

sonalideshpandemsft:
@markfields - I’m removing you as a required reviewer so I can get the PR merged, but feel free to take a look when you’re back :)

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: Thanks for implementing the suppression — that's exactly the right approach. The `silent` flag cleanly separates internal rollback bookkeeping from user-observable events. This concern is resolved. ✅

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3711,6 +3737,12 @@ export class ContainerRuntime
...
 					},
 				);
+				if (!silent) {
```
anthony-murphy: Deep Review: The `exitStagingMode` closure (created in `enterStagingModeCore`) does not validate that the captured `StageControlsInternal` instance is still the *current* one before performing irreversible actions.

**Scenario**: (1) User calls `enterStagingMode()`, gets controls A. (2) User calls `controlsA.commitChanges()` — staging exits normally. (3) User calls `enterStagingMode()` again, gets controls B. (4) User calls `controlsA.commitChanges()` again — no guard prevents this from running, which would flush session B's outbox, clear `this.stageControls`, and emit a spurious `stagingModeChanged` exit event.

This is a pre-existing vulnerability in the closure design, but this PR's new event emissions make the consequences externally observable — UI consumers listening for `stagingModeChanged` would receive incorrect notifications.

**Suggested fix** — add a session-identity guard at the top of `exitStagingMode`:
```typescript
if (this.stageControls !== stageControls) {
    throw new UsageError("Not in staging mode");
}
```

This is a single-line addition, orthogonal to the PR's goals, and consistent with the existing `UsageError` pattern used elsewhere in the staging API.

anthony-murphy (approved):

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3745,6 +3780,15 @@ export class ContainerRuntime
 
 		this.stageControls = stageControls;
 		this.channelCollection.notifyStagingMode(true);
+		if (!silent) {
+			try {
+				this.emit("stagingModeChanged", { inStagingMode: true });
+			} catch (error) {
+				const normalizedError = normalizeError(error);
+				this.closeFn(normalizedError);
```
anthony-murphy: Deep Review: The try/catch here calls `this.closeFn(normalizedError)` when a listener throws during `emit("stagingModeChanged")`, which closes the container. This is inconsistent with how all other runtime events handle listener errors.

For comparison, `this.emit("op", ...)`, `this.emit("signal", ...)`, `this.emit("dirty")`, `this.emit("saved")`, `this.emit("dispose")`, `this.emit("batchBegin", ...)`, and `this.emit("connectedToService", ...)` — none of these call `closeFn` on listener exceptions.

The same issue exists on the exit path (~line 3740-3753), where the `emit("stagingModeChanged", { inStagingMode: false, ... })` sits inside a try-catch originally designed for the staging transition logic (flush, rollback). A listener exception there also triggers `closeFn` even though the transition has already completed successfully.

**Suggested fix** (either approach):
1. **Move exit-path emit outside the try-catch** (after the catch block) and **remove `closeFn` from this enter-path catch** — so listener exceptions propagate but don't close the container.
2. **Or**, if container-fatal behavior is intentional, add TSDoc to the `stagingModeChanged` event documenting that listener exceptions will close the container (similar to `orderSequentially`'s documented callback-fatal contract).

The first approach is more consistent with the rest of the runtime's event emission patterns.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: The stale-controls guard is implemented in the current diff — the `if (this.stageControls !== stageControls)` check is present. This thread can be resolved. ✅

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3745,6 +3780,15 @@ export class ContainerRuntime
...
+				throw normalizedError;
+			}
+		}
 
 		return this.stageControls;
```
anthony-murphy: Deep Review: This returns `this.stageControls` (the instance field) rather than the local `stageControls` variable from line 3756. Since the `emit("stagingModeChanged")` on line 3785 runs arbitrary listener code synchronously before this return, a listener could theoretically cause `this.stageControls` to be cleared (e.g., by triggering `exitStagingMode` on a stale reference).

While the specific failure path isn't currently reachable through public API (listeners don't have the controls reference yet at emit time), this is a zero-cost defensive hardening:

```suggestion
return stageControls;
```

Using the local variable eliminates a class of re-entrancy hazard with no downside.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3681,10 +3705,15 @@ export class ContainerRuntime
 		// since we mark whole batches as "staged" or not to indicate whether to submit them.
 		this.flush();
 
+		// Note: `silent` is captured from the enclosing `enterStagingModeCore` call.
+		// When `true`, both enter and exit events are suppressed (see orderSequentially).
 		const exitStagingMode = (
 			discardOrCommit: () => IPendingMessage["batchInfo"][],
 			exitMethod: "commit" | "discard",
 		): void => {
+			if (this.stageControls !== stageControls) {
```
anthony-murphy, line 3714: Deep Review: This stale-controls guard (`if (this.stageControls !== stageControls)`) is new observable behavior — previously, calling `commitChanges()`/`discardChanges()` on stale controls would not throw this specific `UsageError`. The five new tests cover event emission but none exercise this guard.

Suggested additions to the test suite:
1. Enter staging → `discardChanges()` → call `commitChanges()` on the same controls → assert `UsageError("Not in staging mode")`
2. Enter staging → `discardChanges()` → enter staging again (new controls) → call `discardChanges()` on the *old* controls → assert `UsageError`

These would pin the guard's behavior against future refactors.

packages/runtime/runtime-definitions/src/dataStoreContext.ts
```diff
@@ -132,6 +132,17 @@ export interface IContainerRuntimeBaseEvents extends IEvent {
...
 	(event: "signal", listener: (message: IInboundSignalMessage, local: boolean) => void);
 	(event: "dispose", listener: () => void);
+	/**
+	 * Fires when the container runtime enters or exits staging mode.
+	 * @param stagingModeInfo - An object describing the staging mode state.
+	 * If `inStagingMode` is true, the runtime has entered staging mode.
+	 * If false, it has exited staging mode, and `commit` indicates whether changes were committed or discarded.
+	 *
+	 * @remarks
+	 * This event is not emitted when the container is disposed while in staging mode.
```
anthony-murphy, line 142: Deep Review: This documentation states the event is not emitted when the container is disposed while in staging mode. Currently this holds structurally (`dispose()` doesn't call `exitStagingMode` or emit `stagingModeChanged`), but there's no test pinning this documented semantic.

A simple pinning test would prevent regressions if `dispose()` is later refactored:
```
enter staging mode → dispose container → assert no stagingModeChanged exit event fires
```

packages/runtime/runtime-definitions/src/dataStoreContext.ts
```diff
@@ -132,6 +132,17 @@ export interface IContainerRuntimeBaseEvents extends IEvent {
...
+	 * If the container is disposed, staged changes are silently dropped.
+	 */
+	(event: "stagingModeChanged", listener: (stagingModeInfo: StagingModeChangedEvent) => void);
```
anthony-murphy, line 145: Deep Review: The `stagingModeChanged` event is added to `IContainerRuntimeBaseEvents`, which means every `IContainerRuntimeBase` consumer — including data store contexts — sees this event. This is consistent with the existing API surface (`enterStagingMode` and `inStagingMode` already live on `IContainerRuntimeBase`), but worth confirming this breadth is intentional.

The `@beta @legacy` tag limits the blast radius. A brief note in the PR description confirming the placement rationale would help future readers understand why this event lives on the base interface rather than a narrower one.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3745,6 +3780,13 @@ export class ContainerRuntime
 
 		this.stageControls = stageControls;
 		this.channelCollection.notifyStagingMode(true);
+		if (!silent) {
+			try {
+				this.emit("stagingModeChanged", { inStagingMode: true });
+			} catch (error) {
+				throw normalizeError(error);
```
sonalideshpandemsft: if the listener throws, is it expected to discard changes and exit staging mode? this will causes the caller to enter staging mode but never receive stage controls

anthony-murphy: no. the listener and the requestor could be totally separate, so we should still return the stage controls to the requestor

anthony-murphy: we should probably just log, and not actually throw

anthony-murphy (commented):
review

anthony-murphy (approved):

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: If a listener throws during `emit("stagingModeChanged")` on line 3785, the catch block here re-throws but does not roll back the state mutations already committed on lines 3781–3782 (`this.stageControls = stageControls` and `this.channelCollection.notifyStagingMode(true)`). The caller never receives the `StageControlsInternal` object and cannot call `commitChanges()` or `discardChanges()`. Subsequent calls to `enterStagingMode` throw `"Already in staging mode"` (line 3697), leaving the runtime permanently stuck.

You identified this exact scenario [in your comment](https://github.com/microsoft/FluidFramework/pull/27127#issuecomment-4308458947) — "the runtime is left in staging mode with no recovery."

**Suggested fix** — add rollback before the re-throw:
```typescript
this.stageControls = undefined;
this.channelCollection.notifyStagingMode(false);
throw normalizeError(error);
```

This preserves the documented behavior that `inStagingMode` is `true` when the enter event fires, while ensuring the runtime recovers if a listener throws. A surgical rollback here keeps the scope contained for this PR; switching to `EventEmitterWithErrorHandling` could be a good follow-up for a broader, centralized solution.

A companion test would be valuable: enter staging with a listener that throws → assert the runtime is *not* stuck in staging mode afterward.

sonalideshpandemsft: The enter path does not throw anymore. It logs the error and returns the stage controls. I'll open a follow up for swtiching to `EventEmitterWithErrorHandling`

anthony-murphy:
## Deep Review

*Reviewed commit `f5c05b9` on 2026-04-24.*

Well-designed addition of the `stagingModeChanged` event with a discriminated-union payload, correct internal suppression via the `silent` flag, and comprehensive test coverage (8 new tests). Two minor polish items remain around error-handling consistency and test coverage.

### Findings

- **Asymmetric error handling between enter and exit event emission** — `containerRuntime.ts:3748-3753` — (inline thread)
- **No test coverage for listener error handling on `stagingModeChanged`** — `containerRuntime.spec.ts:4973-5182` — (inline thread)

### Path to Ready

- [ ] **Wrap exit-path emit in try/catch** — Add try/catch around the exit-path `emit("stagingModeChanged")` matching the enter-event pattern, logging `StagingModeChangedError`. This aligns with anthony-murphy's guidance to "just log, not throw." → Moves to 9/10.
- [ ] **Add listener-error regression test** — One test: listener that throws on enter event → assert `enterStagingMode()` still returns valid controls and logs `StagingModeChangedError`. Optional: symmetric test for exit if try/catch is added. → Moves to 9-10/10.
- [ ] **Get sign-off from markfields** — Given this is a high-churn area where markfields has historically reviewed error handling patterns, their review would strengthen confidence.

### Context for Reviewers

- The `silent` flag correctly separates internal `orderSequentially` rollback bookkeeping from user-observable events. `containerRuntime.ts` has high churn (36 changes in 6 months) — anthony-murphy and markfields are the key area experts.
- The stale-controls guard (`if (this.stageControls !== stageControls)`) is a beneficial new correctness check closing a pre-existing gap.
- The PR's discriminated union payload (`{ inStagingMode: true } | { inStagingMode: false; commit: boolean }`) is superior to a simpler boolean — it carries strictly more information with no added complexity.
- The event is intentionally not emitted on dispose or during internal `orderSequentially` usage — both cases are documented in the code.

<details><summary>Suggested PR description</summary>

## Add staging mode event

Introduces a new `stagingModeChanged` event on `IContainerRuntimeBaseEvents` that notifies listeners when the container runtime enters or exits staging mode.

### Event payload (`StagingModeChangedEvent`)
- `{ inStagingMode: true }` — the runtime has entered staging mode
- `{ inStagingMode: false, commit: boolean }` — the runtime has exited staging mode. `commit` is `true` when staged changes were committed (not discarded), `false` when they were discarded.

### Key design decisions
- **Suppressed for internal use**: The event is NOT emitted when staging mode is used internally by `orderSequentially` for rollback support. `enterStagingMode` is split into a public one-liner and a private `enterStagingModeCore(silent)`, where `orderSequentially` passes `silent: true`.
- **Discriminated union type**: `StagingModeChangedEvent` is exported from `runtime-definitions` as a named type for downstream use in type annotations and exhaustiveness checks.
- **Stale-controls guard**: `exitStagingMode` validates the captured `StageControlsInternal` is still current before acting, preventing double-exit bugs.
- **Dispose behavior**: No exit event fires if the container is disposed while in staging mode.

### API surface
- New type: `StagingModeChangedEvent` (`@legacy @beta`)
- New event: `"stagingModeChanged"` on `IContainerRuntimeBaseEvents`
- **Additive, non-breaking change.**

</details>

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: Updating this assessment — the enter-path now correctly catches listener errors and logs via `sendErrorEvent` rather than closing the container, which addresses the primary concern raised here.

The remaining item is a consistency polish: the exit-path `emit("stagingModeChanged")` at lines 3748–3753 has no matching try/catch, so a throwing listener still propagates to the caller of `commitChanges()`/`discardChanges()`. Since the exit operation is already complete at that point (staging state cleaned up, batches committed/discarded, `notifyStagingMode(false)` called), wrapping in the same try/catch + `sendErrorEvent` pattern would be safe and consistent with anthony-murphy's guidance to "just log, not throw."

This is now a minor consistency suggestion, not a blocking issue.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: Updating this assessment — the enter-path try/catch with `sendErrorEvent` now correctly prevents the state-corruption scenario described above (the caller receives valid controls even if a listener throws, and the error is logged to telemetry rather than propagated).

The remaining gap is test coverage: none of the 8 new tests exercise the listener-throw code path. A regression test confirming that a throwing listener on the enter event doesn't prevent `enterStagingMode()` from returning valid controls — and that `StagingModeChangedError` is logged — would pin this behavior against future refactors.

This is now a minor suggestion for test completeness, not a blocking issue.

sonalideshpandemsft: I'll open a follow up for this

packages/runtime/container-runtime/src/containerRuntime.ts
sonalideshpandemsft: will follow up on this

PR #27130, opened 2026-04-22
fix(container-runtime): cancel GC timers on container close, not just dispose

CraigMacomber:
## Description

When a container closes, GC timers (session expiry and unreferenced-node
timers) were kept alive until dispose() was called. This caused timers to
fire unnecessarily after a container was functionally done — and could
cause memory leaks or, in Node.js environments, keep the event loop alive
and prevent clean process exit until dispose() eventually ran.

Adds an optional `close` method to `IRuntime` (backward-compatible)
which `Container` calls on close. `ContainerRuntime` implements it by
delegating to a new `GarbageCollector.dispose()` method that cancels all
timers.

See changeset for implications to users.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

## Reviewer Guidance

I do not know this code well: feedback on if this is conceptually a good idea, and if the implementation is good would both be great.


packages/runtime/container-runtime/src/gc/garbageCollection.ts
```diff
@@ -1174,14 +1174,17 @@ export class GarbageCollector implements IGarbageCollector {
...
 		// Clear all unreferenced node timers to prevent process hanging
 		for (const tracker of this.unreferencedNodesState.values()) {
-			tracker.stopTracking();
+			tracker.cancelTimers();
 		}
+	}
+
+	public dispose(): void {
+		this.close();
 		this.unreferencedNodesState.clear();
```
CraigMacomber, line 1187: If it is ok to clear `unreferencedNodesState` on close instead of dispose, we could avoid having both as part of the GC interface, and just dispose the GC during close leaving this code alone.

packages/common/container-definitions/src/runtime.ts
```diff
@@ -223,6 +223,18 @@ export interface IRuntime extends IDisposable {
...
+	 * Notifies the runtime that the container has closed. The runtime should cancel any timers
+	 * or other resources that are no longer needed once the container is closed, while preserving
+	 * state that may still be inspected by the caller before disposal.
+	 *
+	 * @remarks
+	 * This is optional for backwards compatibility with older runtime implementations. Runtimes
+	 * that do not implement this will have resources cleaned up only when {@link @fluidframework/core-interfaces#IDisposable.dispose}
+	 * is called.
+	 */
+	notifyClosed?(): void;
```
markfields: I'm surprised we need this, but will have to look closer to see why this isn't possible today.

packages/runtime/container-runtime/src/gc/gcDefinitions.ts
```diff
@@ -453,6 +453,12 @@ export interface IGarbageCollector {
 	 */
 	isNodeDeleted(nodePath: string): boolean;
 	setConnectionState(canSendOps: boolean, clientId?: string): void;
+	/**
+	 * Cancels all GC timers without destroying tracked state.
+	 * @remarks
+	 * Called when the container closes so timers do not keep the event loop alive or leak memory.
+	 */
+	close(): void;
```
markfields: I would think disposing GC is fine in a closed container

CraigMacomber: That will call unreferencedNodesState.clear() sooner which was my concern about that. So, I take it that unreferencedNodesState is not part of what can be inspected for a closed container?

CraigMacomber: After some investigation, I think its prabably fine as you suggest, and I have made the simplification.

packages/common/container-definitions/src/runtime.ts
CraigMacomber: I couldn't find another way, neither could Claude. It noted:

> There's no event subscription mechanism on IContainerContext and no existing hook from container → runtime for the "closed but not disposed" transition. The only lifecycle signals go the other direction (closeFn/disposeFn are runtime → container).

packages/loader/container-loader/src/container.ts
```diff
@@ -1074,6 +1074,8 @@ export class Container
 				this.mc.logger.sendErrorEvent({ eventName: "ContainerCloseException" }, newError);
 			}
 
+			this._runtime?.notifyClosed?.();
```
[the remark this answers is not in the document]
CraigMacomber: I'm not sure if this makes sense. I don't have a good understanding of our policy around exceptions in this part of the codebase, and would like advice on if this change would be an improvement.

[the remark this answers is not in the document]
markfields: Yeah I'd move it up by all the other subsystem close/dispose calls.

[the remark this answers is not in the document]
markfields: Maybe also add a comment trying to explain why it's "notify close" instead of "close".

Or maybe we do call it "close" and add a call to `this.closeFn()` in there and update internal (on CR) callers to call `close`.  What do you think?

markfields (approved):

packages/loader/container-loader/src/container.ts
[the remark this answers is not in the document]
CraigMacomber: > Yeah I'd move it up by all the other subsystem close/dispose calls.

I have now palced it at the end of that try block. This risks it not runing if one of the other error: if that becomes an issue we can move this to a location that gets hit more robustly (like where it was, but maybe in a another try as Copilot suggested)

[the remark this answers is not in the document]
CraigMacomber: I have also renamed it to "close", and adjusted the docs on runtime.close accordingly, detailing the semantics of this new state better.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -2391,6 +2391,10 @@ export class ContainerRuntime
 		}
 	}
 
+	public close(): void {
```
markfields, line 2400: This will likely cause confusion to devs in this class since it's unclear whether to call close() or closefn() when initiating the close from within (in case of error).

If you call this `close`, I suggest making `closeFn` private and calling it on the first line in here.  Otherwise I fear people would not know what to do when wanting to close the Container from within the ContainerRuntime layer (and will likely default to wanting to call `close` over `closeFn`, at least that's my guess)

Now we need to make sure `closeFn` is idempotent even when reentrant.  It just calls `Container.close` which is idempotent but doesn't handle reentrancy well from my reading.  I think you'd need to add a `this._lifecycleState` check at the top and return if it's closing, closed, or disposed.  Otherwise in the closing case it'll fail the `verifyClosedAfter` check in the reentrant case.  You could also update `DeltaManager.close` to return a bool indicating the early-return and only do `verifyClosedAfter` if that's false?

After considering this, feel free to switch back to `notifyClosed` too.  The bidirectional nature flow the close operation makes this pretty complicated (the fact that the host can initiate close _or_ the container can close itself from within)

CraigMacomber: closeFn implements IGarbageCollectionRuntime.closeFn and ISummarizerRuntime.closeFn, so we can't make the implementation of it private.

ISummarizerRuntime.closeFn is undocumented, and IGarbageCollectionRuntime.closeFn is documented to require an error, but its optional, and that property is reassignable despote the implemention being readonly.

This is already really confusing, so I agree I shouldn't make it worse.

I'll see if I can clean it up a bit.

.changeset/salty-pans-worry.md
```diff
@@ -0,0 +1,18 @@
+---
+"@fluidframework/container-definitions": minor
+"@fluidframework/container-runtime": minor
+"__section": fix
+---
+
```
Josmithr: Nit (pretty sure we don't want the leading line break?)
```suggestion
```

.changeset/salty-pans-worry.md
```diff
@@ -0,0 +1,18 @@
...
+GC timers are now cancelled when a container closes, not just when it is disposed
+
+Adds an optional `notifyClosed()` hook to `IRuntime` that `Container` calls on close.
+`ContainerRuntime` implements it by cancelling all GC timers (session expiry and unreferenced-node timers)
+without clearing tracked state.
+
+This prevents the timers from causing memory leaks after a container is closed but not disposed.
```
Josmithr: Nit: I think we generall capitalize "Container" when referring to a Fluid Container.

packages/common/container-definitions/src/runtime.ts
```diff
@@ -223,6 +223,21 @@ export interface IRuntime extends IDisposable {
...
+	 * The entryPoint is preserved, allowing content to still be read/inspected.
+	 *
+	 * @remarks
+	 * This enters an intermediate lifecycle stage between connected operation and full disposal.
+	 * The container may call this when it closes, before eventually calling
+	 * {@link @fluidframework/core-interfaces#IDisposable.dispose}.
+	 * Any resources this cleans up should also be cleaned up by `dispose`
+	 * in the case when `close` is not called.
+	 *
+	 * This is optional for backwards compatibility with older runtime implementations.
```
Josmithr, line 238: Should we add this to the list of breaking API changes we're planning for 3.0? We have a GitHub issue tracking those. Seems like we would want to make this required at that time.

packages/runtime/container-runtime/src/gc/gcDefinitions.ts
```diff
@@ -453,6 +453,12 @@ export interface IGarbageCollector {
 	 */
 	isNodeDeleted(nodePath: string): boolean;
 	setConnectionState(canSendOps: boolean, clientId?: string): void;
+	/**
+	 * Cancels all GC timers and clears tracked state so timers do not keep the event loop alive
+	 * or leak memory.
+	 * @remarks
+	 * Idempotent: Safe to call multiple times.
```
Josmithr: Nit (complete sentences)
```suggestion
	 * This is idempotent - it is safe to call multiple times.
```

Josmithr (approved):

packages/runtime/container-runtime/src/containerRuntime.ts
CraigMacomber: We have disposeFn and dispose. This is symetric with how close works now.

I have added documentation to point out how these "Fn" versions relate to the non-fn versions in several palces, and also managed to actually remove closeFn from both IGarbageCollectionRuntime and ISummarizerRuntime, allowing it to be made private. Should be much better now.

packages/common/container-definitions/src/runtime.ts
CraigMacomber: I don't know what our plans related to layer compat are for 3.0, but we might be supporting runtimes older than this? Regardless this is a beta API, and/or a layer compat thing, so it doesn't need to be coupled to 3.0: it seems like a seperte issue.

CraigMacomber: Specificallyf or this case, like many of our existing beta interfaces, we likley want to make a beta breaking change to seal them at somepoint (and also ensure customers arn't implementing them first). If we do that, and address the layer compat implications (possibly by waiting lojg enough after its sealed), then I think this could be required. None of this needs a major version.

Josmithr: Should we be tracking this as a potential *beta* breaking change for the next (or subsequent) appropriate version? Maybe with notes about layer compat?

Also thinking we should probably leave a note here in the comment that this will likely become required in a future version (probably without being too specific about which version).

packages/common/container-definitions/src/runtime.ts
```diff
@@ -236,6 +236,9 @@ export interface IRuntime extends IDisposable {
 	 * in the case when `close` is not called.
 	 *
 	 * This is optional for backwards compatibility with older runtime implementations.
+	 *
+	 * Do not confuse this with with the various `closeFn` callbacks (such as {@link IContainerContext.closeFn} or `IGarbageCollectorCreateParams.closeFn`:
+	 * these expose a way to intitate container close, not a way to get notified of container close (which might come from another source).
```
Josmithr: This is backwards, right? This should say "this exposes a way..."

packages/runtime/container-runtime/src/containerRuntime.ts
Josmithr: Not for this PR, but "Fn" name distinction here is quite gross. We should consider renaming the notification methods IMO.

packages/common/container-definitions/src/runtime.ts
CraigMacomber: The way I see it most of the beta+legacy API surface should, when we have time, get cleaned up. Most likely this interface should be made sealed, then use a mostly opaque type, and not expose this method, or most of the rest of this interface (maybe none of it) as beta+legacy APIs. This is true of most of the legacy+beta API surface. I don't think adding this method really imapcts that.

Since not implementing this is generally low impact, I don't see this new method really being a motivating factor to start this process, so I don't think we need to add a tracking items because of the new close method.

I have now added some docs explaining how closed relates to this, so if somone does get around to sealing `IRuntime`, its clear what the intentions for `close` are at that point.

packages/runtime/container-runtime/src/containerRuntime.ts
CraigMacomber: > Not for this PR, but "Fn" name distinction here is quite gross. We should consider renaming the notification methods IMO.

I agree, which is why I did what I could to remove and clarify them (we no longer expose closeFn from the runtime's internal APIs, and the rest now have docs).

As this crosses layer compat via container context, I didn't want to try and rename them: I suspect doing so would be a pain, but maybe someone wants to take up that project someday (I suspect it's not worth doing any time soon).

packages/common/container-definitions/src/runtime.ts
CraigMacomber: No, I think you got it backwards. Regardless I have made the docs more clear.

packages/runtime/container-runtime/src/containerRuntime.ts
markfields: Let's wait until we can have long-lived agentic workflows where a bot can carry the change all the way through over the course of however many years our compat policies require (now that they're explicit and not bound by checking telemetry for saturation in 1P partners this is doable)

markfields (approved):

Josmithr (approved):

PR #27131, opened 2026-04-22
ci(client): scope PR tests to changed packages via opt-in flag

tylerbutler:
## Description

> Draft — verifies the end-to-end CI-scoping pipeline. Contains two independent commits, structured so the `flub pnpm-run` commit can be cherry-picked into its own PR first (merge + release `@fluid-tools/build-cli`), after which this branch is rebased onto main and the root `package.json` scripts are migrated from `scripts/pnpm-run-scoped.cjs` to `flub pnpm-run` in a follow-up commit.

### Commit 1 — `feat(build-cli): add pnpm-run command`

New `flub pnpm-run` subcommand: thin wrapper around `pnpm` that reads `FLUID_PNPM_FILTER` from the environment and prepends it to the pnpm argv. When the variable is unset or empty, it falls back to `-r`, preserving today's "run across every workspace package" behavior. All argv after the command name is forwarded to pnpm verbatim. Self-contained — cleanly cherry-pickable to its own PR.

### Commit 2 — `ci(client): scope PR tests to changed packages via opt-in flag`

Adds an opt-in capability to the client build pipeline to run tests only for packages affected by a PR's diff (plus their transitive dependents), instead of every workspace package. Gated behind a new pipeline parameter `enableChangedPackageTestScoping` (default `false`) — when off, pipeline behavior is byte-identical to today.

**New:**
- `scripts/pnpm-run-scoped.cjs` — local shim with the same semantics as `flub pnpm-run`. Used for now because the `@fluid-tools/build-cli` version consumed by the client workspace is pinned in the catalog and doesn't yet include the new command. Will be deleted once the catalog version bumps.
- `tools/pipelines/templates/include-detect-changed-packages.yml` — ADO template that installs flub (via the existing `include-install-build-tools.yml`, registry path for speed), runs `flub list client --changed --branch <target> --json`, applies full-run trigger globs for root/tooling edits, and publishes `shouldRunTests` + `changedPackagesPnpmFilter` output variables. Every error path degrades safely to a full test run.

**Wired up in:**
- `build-npm-client-package.yml` — optional `detect_changes` job at the top of the build stage; `Coverage_tests` and each `Test_*` job conditionally `dependsOn` it, consume its outputs, short-circuit the "Start Test" step when `shouldRunTests == 'false'`, and forward the filter through `include-test-task.yml`'s new `pnpmFilter` parameter.
- `include-test-task.yml` — `pnpmFilter` parameter maps onto a `FLUID_PNPM_FILTER` env var on the npm task.
- `build-client.yml` — surfaces `enableChangedPackageTestScoping` as a top-level pipeline parameter so it can be flipped at queue time.

**Root `package.json` scripts updated** to delegate through `scripts/pnpm-run-scoped.cjs`: `ci:test:mocha`, `ci:test:jest`, `ci:test:realsvc:{local,tinylicious}`, `ci:test:stress:tinylicious`, `ci:check:are-the-types-wrong`, their `:coverage` variants, and the shared `test:mocha` / `test:jest:report`. Default behavior (no env var) matches existing `pnpm run -r ...`.

## Reviewer Guidance

See https://github.com/microsoft/FluidFramework/wiki/PR-Guidance — the detect template's bash script and the conditional YAML inside `build-npm-client-package.yml` are the load-bearing bits. The rollout flag is intentionally defaulted to `false` so this PR changes no pipeline behavior until a reviewer or pipeline operator opts in.


tylerbutler:
Superseded by a reworked branch at `test/filtered-ci/scoping`. The new implementation drops the custom `scripts/pnpm-run-scoped.cjs` wrapper entirely and uses pnpm's native `--filter "...[<sha>]"` selector via `npm_config_filter`. The branch-name pattern (`test/filtered-ci/*`) auto-activates the scoping path on PR push so the feature can be verified end-to-end without flipping a parameter at queue time.

PR #27132, opened 2026-04-22
feat(build-cli): add pnpm-run command

tylerbutler:
## Description

Adds `flub pnpm-run`, a thin wrapper that invokes `pnpm` with a workspace filter read from the `FLUID_PNPM_FILTER` environment variable. When the variable is unset or empty, the command falls back to `-r` (every workspace package), preserving today's recursive semantics. When set, its contents are tokenized (preserving quoted patterns) and prepended to the pnpm argv.

Every argument after the command name is forwarded to pnpm verbatim — callers use it as a drop-in replacement for `pnpm` in scripts that need to support optional per-call scoping. For example:

```json
"test:mocha": "flub pnpm-run run --no-sort --stream --no-bail test:mocha --color"
```

Running with no env var is byte-equivalent to the original `pnpm run -r ...` invocation. Running with `FLUID_PNPM_FILTER='--filter "@fluidframework/map..."'` scopes execution to that package and its transitive dependents.

Unit tests cover `tokenizeFilter` for empty/whitespace input, single/multiple double-quoted patterns, unquoted values, and single-quoted values.

## Motivation

Pipeline work in #27131 needs a way to scope test execution to packages affected by a PR. The current approach there uses a local `scripts/pnpm-run-scoped.cjs` shim because the client workspace pins `@fluid-tools/build-cli` to a catalog version that doesn't yet include this command. Once this PR merges and a new build-cli release ships, the sibling PR can delete the shim and migrate the root `package.json` scripts to `flub pnpm-run`.

## Reviewer Guidance

See https://github.com/microsoft/FluidFramework/wiki/PR-Guidance — the command deliberately bypasses oclif's flag parsing (`static strict = false`, no `this.parse()` call) so that pnpm-facing flags like `--no-bail` aren't intercepted. The generated `docs/pnpm-run.md` is not included here; it's produced by `oclif readme --multi` at build time.

### DIFF
--- a/build-tools/packages/build-cli/src/commands/pnpm-run.ts
+++ b/build-tools/packages/build-cli/src/commands/pnpm-run.ts
@@ -1,0 +1,101 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { spawn } from "node:child_process";
+import { Command } from "@oclif/core";
+
+/**
+ * Environment variable that controls which workspace packages the wrapped
+ * `pnpm` invocation targets. Pipeline jobs set this to one or more
+ * `--filter "<pattern>"` fragments (space-separated) to scope command
+ * execution to packages affected by a PR. When unset or empty, the command
+ * falls back to `-r` — the pre-existing "run across every workspace package"
+ * behavior — so local developers get unchanged semantics out of the box.
+ */
+const FILTER_ENV_VAR = "FLUID_PNPM_FILTER";
+
+/**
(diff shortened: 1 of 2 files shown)


tylerbutler:
Closing — investigation showed pnpm's native `--filter "...[ref]"` selector (pnpm.io/filtering) covers this use case via the npm-style `npm_config_filter` env var, so the sibling PR (#27131) can use pnpm directly and doesn't need a custom flub command after all. One caveat is that pnpm's `[ref]` uses a two-dot diff rather than merge-base (pnpm/pnpm#9907), which we already work around by computing merge-base ourselves in the detect job.

PR #27133, opened 2026-04-22
ci(client): scope PR tests to changed packages via pnpm filter

tylerbutler:
## Description

Adds an opt-in capability to the client build pipeline to run tests only for packages changed in a PR (plus transitive dependents), instead of every workspace package. Scoping is applied natively by pnpm via `--filter "...[<merge-base-sha>]"` — no wrapper scripts, no changes to existing test-script structure beyond one narrow fix. Non-opt-in PRs are byte-identical to the prior pipeline behavior.

**Activation:** opt-in via branch-name prefix — `detect_changes` runs only when the PR source branch name contains `test/filtered-ci/`. Once the feature has soaked, a follow-up can broaden this (label-based, opt-out, default-on).

## Changes

- **`tools/pipelines/templates/include-detect-changed-packages.yml`**: new `detect_changes` job. Computes the merge-base SHA vs the target branch and emits `shouldRunTests` and `scopedPnpmFilter` output variables. Merge-base rather than `origin/<branch>` directly because pnpm's `--filter "[ref]"` uses a two-dot diff ([pnpm/pnpm#9907](https://github.com/pnpm/pnpm/issues/9907)) and would otherwise pick up unrelated commits from the target branch.
- **`scripts/detect_changed_packages.py`**: Python detection logic — pure helpers plus a `main()` that writes ADO variables. Stdlib only (no install step on the agent — `python3` already ships on ADO Linux images, keeping the gate close to "git + python" fast). Unions the merge-base tree (`git ls-tree`) with the current tracked tree (`git ls-files`) when resolving package directories, so a package deleted on this branch still maps correctly. Diff is computed as `git diff <merge-base>..HEAD` (commit-only, immune to working-tree mutations from any future pre-step).
- **`scripts/test/test_detect_changed_packages.py`**: stdlib `unittest` coverage for the detection helpers. Wired into the client build job immediately after install (`python3 -m unittest discover`) so a regression in the change-detection logic fails fast, before any heavy build work.
- **`tools/pipelines/templates/build-npm-client-package.yml`**: `Coverage_tests` and `Test_*` jobs now `dependsOn: detect_changes`. When `shouldRunTests == 'false'` the job skips entirely (saves agent allocation, checkout, install). Otherwise `scopedPnpmFilter` is passed through as `npm_config_filter` so `pnpm -r` scopes the recursive run natively.
- **`tools/pipelines/templates/include-test-task.yml` + `package.json`**: `npm_config_filter` propagates to every `pnpm` invocation a script makes, not just `pnpm -r`. Non-recursive calls like `pnpm puppeteer …` fail under an active filter with `ERR_PNPM_RECURSIVE_EXEC_NO_PACKAGE`. The `pnpm puppeteer` setup calls in `test:jest` / `test:jest:bail` / `test:jest:report` are now prefixed with `cross-env npm_config_filter=` so they run unfiltered, and the contract is documented on the `pnpmFilter` parameter in `include-test-task.yml`.
- **Full-run safety net**: `FULL_RUN_PATTERNS` in the detection script forces a full run when the diff touches cross-cutting root files (`pnpm-lock.yaml`, `.pnpmfile.cjs`, `.npmrc`, `.nvmrc`, `fluidBuild.config.cjs`, `tools/`, `common/`, `scripts/`, etc.).
- Activation uses `in(dependencies.build.result, 'Succeeded', 'SucceededWithIssues')` rather than `succeeded()` because `succeeded()` treats a `Skipped` `detect_changes` dependency as non-success and would false-skip test jobs on non-opt-in PR builds.
- `detect_changes` uses a shallow clone with on-demand deepen: `fetchDepth: 200` covers most PRs; the script uses `git rev-parse --is-shallow-repository` and incrementally deepens (`fetch --deepen 1000`) once if the merge-base isn't reachable, falling through to a safe full-run outcome if it still misses. Never `--unshallow` — pulling full history is expensive and rarely needed.
- **Safe-fallback policy**: any unexpected error (missing merge-base, git failure, unparseable ref) emits a pipeline warning and falls back to a full run — never a silent skip. The aggressive "no packages affected" skip path is also surfaced as a `##vso[task.logissue type=warning]` with a full file dump so accidental silent suppression is auditable from the pipeline summary.

## Reviewer Guidance

- Is the `test/filtered-ci/` branch-name convention acceptable as the long-term opt-in mechanism, or should a follow-up switch to a label / opt-out / default-on?
- `FULL_RUN_PATTERNS` partially overlaps with `pr: paths: include:` in `build-client.yml`; kept in sync by convention, not programmatically. Comment-linked in both directions.
- Re-enabling coverage on scoped runs was raised in review and deferred to a follow-up — the filter plumbing is already compatible; the scoped-coverage baseline-comparison semantics are worth thinking through separately.


tools/pipelines/templates/include-detect-changed-packages.yml
```diff
@@ -0,0 +1,183 @@
...
+          '^pnpm-lock\.yaml$'
+          '^pnpm-workspace\.yaml$'
+          '^fluidBuild\.config\.cjs$'
+          '^tsconfig[^/]*\.json$'
+          '^biome\.'
+          '^tools/'
+          '^common/'
+          '^scripts/'
+          '^\.changeset/config\.json$'
+        )
```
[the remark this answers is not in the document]
alexvy86: The three examples mentioned here seem worth adding.

alexvy86 (approved):
I'm in for a trial run of this. The logic for the new template is pretty complex though, so before making it default behavior I'd strongly argue for moving it to a script that has unit tests which run automatically. A bug here that results in us not running any tests could easily go unnoticed.

alexvy86:
Also... `pnpm --filter` is such a weird flag 🤔 .

tools/pipelines/templates/include-detect-changed-packages.yml
```diff
@@ -0,0 +1,183 @@
...
+    # within a few hundred commits, so this is cheap in the common case.
+    fetchDepth: 200
+
+  - task: Bash@3
+    name: setChangedPackages
+    displayName: Detect changed packages
+    inputs:
+      targetType: inline
+      workingDirectory: '$(Pipeline.Workspace)/${{ parameters.buildDirectory }}'
+      script: |
```
jason-ha, line 32: inline scripts are problematic to investigate outside of pipeline run. This should be viable as a standalone bash script.

tools/pipelines/templates/build-npm-client-package.yml
```diff
@@ -561,7 +582,18 @@ extends:
...
+              - detect_changes
+            # Skip the job entirely when detect_changes explicitly concluded no
+            # packages were affected — saves agent allocation, checkout, and
+            # install on scoped PRs. Empty (detect_changes was skipped) still
+            # runs. See include-detect-changed-packages.yml for the invariants.
+            #
+            # We check `build.result` directly instead of using `succeeded()`
+            # because the latter treats a Skipped `detect_changes` dependency
+            # as non-success and false-skips this job on non-opt-in PR builds.
+            condition: and(in(dependencies.build.result, 'Succeeded', 'SucceededWithIssues'), ne(dependencies.detect_changes.outputs['setChangedPackages.shouldRunTests'], 'false'))
```
jason-ha, line 669: This job is called Coverage _tests.
Recently coverage was disabled until performance improves.
1. Perhaps it would be okay to have coverage enabled when testing is selective
2. The coverage checks were never fully implemented as far as I understand; assuming we want coverage to come back, how does this work with that?

tylerbutler: Good points — replying to each.

**1. Coverage when testing is selective:** Agreed in principle; a scoped run is cheap enough to absorb the coverage overhead. I'd prefer to leave that change for a follow-up though. The knob (`testCoverage` parameter in `build-npm-client-package.yml`) is evaluated at template compile time, while `scopedPnpmFilter` is a runtime output from `detect_changes`. Flipping coverage on conditionally would mean either duplicating each test step with a runtime `condition:` guard, or moving the coverage-vs-not choice into the test task itself (script reads the filter env, picks `test:mocha:coverage` vs `test:mocha`). Neither is hard but both want to land in a PR that's explicitly about coverage, not scoping, so the right owner can weigh in on perf tradeoffs and the comparison-baseline question below.

**2. How it works with coverage re-enablement:** The filter plumbing is orthogonal to whether coverage runs. `npm_config_filter` is set on the `Npm@1` task regardless of which script runs (`test:mocha` or `test:mocha:coverage`), and pnpm applies it identically to both `pnpm -r run test:mocha:coverage` and `pnpm -r run test:mocha`. So when coverage comes back, no new wiring is needed on the scoping side — the same filter will scope the coverage runs.

The one thing worth flagging is **baseline comparison semantics**: `Report Code Coverage Comparison` today compares the PR's coverage artifact against a baseline build, and on a scoped PR the artifact would only contain coverage data for the filtered packages. We'd want the comparison step to either (a) only compare the overlap with the baseline, or (b) be skipped entirely when scoping is active, to avoid a "coverage dropped from 80% to 5%!" false alarm because most packages weren't measured. Not something this PR has to solve, but worth noting for whoever picks up coverage re-enablement.

tools/pipelines/templates/include-detect-changed-packages.yml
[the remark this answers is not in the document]
tylerbutler: Added all three (`.pnpmfile.cjs`, `.npmrc`, `.nvmrc`) to `FULL_RUN_PATTERNS` and covered each with a unit test. See 89524e0 — `scripts/detect-changed-packages.ts` and `scripts/test/detect-changed-packages.test.ts`.

tools/pipelines/templates/include-detect-changed-packages.yml
```diff
@@ -0,0 +1,183 @@
...
+          while [ "${d}" != "." ] && [ "${d}" != "/" ]; do
+            if [ -f "${d}/package.json" ]; then
+              HAS_PACKAGE_CHANGE="true"
+              break 2
+            fi
+            d="$(dirname "${d}")"
+          done
+        done <<< "${CHANGED_FILES}"
+
+        if [ "${HAS_PACKAGE_CHANGE}" = "false" ]; then
```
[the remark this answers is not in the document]
tylerbutler: Fixed. `buildPackageDirSet` now unions the merge-base tree (`git ls-tree -r <mergeBase>`) with the working tree when building the package-dir set, so a deleted-package diff still maps to a known package dir. Regression test added in `scripts/test/detect-changed-packages.test.ts` — "detects a deleted package's file". See 89524e0.

tools/pipelines/templates/include-test-task.yml
```diff
@@ -25,6 +33,11 @@ steps:
       customCommand: 'run ${{ parameters.taskTestStep }}:coverage'
     condition: and(succeededOrFailed(), eq(variables['startTest'], 'true'))
     env:
+      # Only emit npm_config_filter when a filter is actually set. pnpm 10
+      # happens to treat an empty value the same as unset, but the conditional
+      # keeps the contract explicit and avoids relying on that behavior.
+      ${{ if ne(parameters.pnpmFilter, '') }}:
+        npm_config_filter: ${{ parameters.pnpmFilter }}
       # Tests can use this environment variable to behave differently when running from a test branch
```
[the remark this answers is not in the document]
tylerbutler: Confirmed empirically — `npm_config_filter` DOES propagate to non-recursive pnpm invocations and breaks `pnpm exec` with `ERR_PNPM_RECURSIVE_EXEC_NO_PACKAGE`. Fixed by prefixing the `pnpm puppeteer ...` call in `test:jest` / `test:jest:bail` / `test:jest:report` with `cross-env npm_config_filter=`, which clears the env var only for that command. Added a caveat comment in `include-test-task.yml` so future scripts chaining non-recursive pnpm before a recursive one don't reintroduce the bug. See 89524e0.

tools/pipelines/templates/include-detect-changed-packages.yml
tylerbutler: Extracted to `scripts/detect-changed-packages.ts` — a typed module with pure helpers (`normalizeTargetBranch`, `checkFullRunPatterns`, `buildPackageDirSet`, `findChangedPackages`) plus a thin `main()` that handles the pipeline I/O. The YAML template is now a 5-line wrapper that calls `npx --yes tsx@4.19.4 scripts/detect-changed-packages.ts`.

Tests live in `scripts/test/detect-changed-packages.test.ts` using Node's built-in `node:test` — no new test framework, no extra deps. They run in the build-client `build` job (early, before pnpm install, so they fail fast) via the same `npx --yes tsx@4.19.4 --test ...` invocation. 25 tests currently, covering each helper + the regression cases the other review threads flagged. See 89524e0.

tylerbutler: Will switch from npx --yes to pnpm exec with a proper devdep in the lockfile.

tools/pipelines/templates/build-npm-client-package.yml
jason-ha: It is that last part that I was wanting to make sure we think about. It also seemed doable, but plenty of things go wrong in the coverage space, in my experience, or at least require significant effort.
I am very in favor of directed testing that is the intent here - would hate to not have it if we can't be compatible with code coverage (that I expect/hope will come back). Some notes in the files might be good.

Re 1. I was thinking it would be simple turn on coverage if we are detecting (the conditions for running detect job are met). But if there is really a dependency on the results of the detect job, then I am expecting that the testCoverage logic could be adjusted to use runtime conditions. I'm trying to avoid looking at the details. :)

tools/pipelines/templates/include-detect-changed-packages.yml
jason-ha: We sure about bringing in tsx+esbuild and those new deps for this? (We can use tooling, but there is extra burden to justify extended deps these days to reduce exposure surface.)
Alternatives I can quickly think of are:
1. going back to bash
2. maybe using type-checked javascript
3. since there are early steps anyway as you note, we just transpile with typescript

Or do we have other uses for tsx looming?

scripts/test/detect-changed-packages.test.ts
```diff
@@ -0,0 +1,188 @@
...
+		const match = checkFullRunPatterns([".pnpmfile.cjs"]);
+		strictEqual(match?.source, "^\\.pnpmfile\\.cjs$");
+	});
+
+	it("matches .npmrc (added in review)", () => {
+		const match = checkFullRunPatterns([".npmrc"]);
+		strictEqual(match?.source, "^\\.npmrc$");
+	});
+
+	it("matches .nvmrc (added in review)", () => {
```
alexvy86:
```suggestion
	it("matches .pnpmfile.cjs", () => {
		const match = checkFullRunPatterns([".pnpmfile.cjs"]);
		strictEqual(match?.source, "^\\.pnpmfile\\.cjs$");
	});

	it("matches .npmrc", () => {
		const match = checkFullRunPatterns([".npmrc"]);
		strictEqual(match?.source, "^\\.npmrc$");
	});

	it("matches .nvmrc", () => {
```

scripts/test/detect-changed-packages.test.ts
```diff
@@ -0,0 +1,188 @@
...
+
+	it("exposes the pattern list for external audits", () => {
+		ok(FULL_RUN_PATTERNS.length > 0);
+		// Ensure each of the three review-added patterns made it into the
+		// exported list (not just the checker).
+		const sources = FULL_RUN_PATTERNS.map((r) => r.source);
+		ok(sources.includes("^\\.pnpmfile\\.cjs$"));
+		ok(sources.includes("^\\.npmrc$"));
+		ok(sources.includes("^\\.nvmrc$"));
+	});
```
alexvy86: Not sure I see a point to this test (particularly the 3 explicit checks, but in general too).
```suggestion
```

scripts/detect-changed-packages.ts
```diff
@@ -0,0 +1,279 @@
...
+	return false;
+}
+
+/** Thin wrapper for `git` calls. Returns stdout or `undefined` on failure. */
+function git(args: string[]): string | undefined {
+	try {
+		return execFileSync("git", args, { encoding: "utf8", stdio: ["ignore", "pipe", "pipe"] });
+	} catch {
+		return undefined;
+	}
```
alexvy86: At least let's log a warning with the error.

scripts/test/detect-changed-packages.test.ts
```diff
@@ -0,0 +1,188 @@
...
+		strictEqual(findChangedPackages(["packages/doomed/package.json"], pkgDirs), true);
+	});
+
+	it("detects a NEW package (added on this branch)", () => {
+		const withNew = new Set(["packages/new"]);
+		strictEqual(
+			findChangedPackages(["packages/new/package.json", "packages/new/src.ts"], withNew),
+			true,
+		);
+	});
```
alexvy86: There are redundant, no? One of these is enough to make sure `findChangedPackages` itself behaves as it should when passed a specific list of dirs. The interesting bit to address Jason's concern is to make sure that deleted packages actually make it into the list of pkgDirs at runtime.

scripts/detect-changed-packages.ts
```diff
@@ -0,0 +1,279 @@
...
+/**
+ * Return `true` if any changed file lives under a known package directory.
+ * A file at `packages/foo/src/x.ts` matches if `packages/foo` (or any
+ * ancestor above it, stopping at the root) is in `packageDirs`.
+ *
+ * The root pseudo-dir `"."` is deliberately ignored here: root-level package
+ * changes are already caught by `FULL_RUN_PATTERNS` and should not double-
+ * count as a per-package signal.
+ */
+export function findChangedPackages(
```
alexvy86: Nit. To me the current name suggests that it'll return a list of changed packages.
```suggestion
export function isChangedFileInOneOfThesePackages(
```

tools/pipelines/templates/build-npm-client-package.yml
ChumpChief: Just remember that the coverage disablement was not due to total runtime (as this change would address) but rather per-test timeouts being triggered when the timeout had a sufficiently slim margin compared to the normal test run time.  So the problem remains that an audit of individual tests to expand their timeout window is required before coverage tests can be reliably reenabled.

package.json
```diff
@@ -105,9 +105,9 @@
 		"test:copyresults": "copyfiles --exclude \"**/node_modules/**\" \"**/nyc/**\" nyc",
 		"test:coverage": "c8 npm test",
 		"test:fromroot": "mocha \"packages/**/dist/test/**/*.spec.*js\" --exit",
-		"test:jest": "assign-test-ports && pnpm puppeteer browsers install chrome-headless-shell && pnpm -r --no-sort --stream --no-bail test:jest --color",
-		"test:jest:bail": "assign-test-ports && pnpm puppeteer browsers install chrome-headless-shell && pnpm -r --no-sort --stream test:jest",
-		"test:jest:report": "assign-test-ports && pnpm puppeteer browsers install chrome-headless-shell && pnpm -r --no-sort --stream --no-bail --workspace-concurrency=4 test:jest",
+		"test:jest": "assign-test-ports && cross-env npm_config_filter= pnpm puppeteer browsers install chrome-headless-shell && pnpm -r --no-sort --stream --no-bail test:jest --color",
```
ChumpChief, line 108: This functionality would be potentially useful in a dev's inner loop.  Can this feature be crafted in some way that a dev can run a simple command locally and get the same scoped run?

scripts/detect_changed_packages.py
```diff
@@ -0,0 +1,295 @@
...
+# that could plausibly invalidate assumptions across the entire workspace.
+#
+# This list partially overlaps with `pr: paths: include:` in
+# tools/pipelines/build-client.yml (which decides whether the pipeline runs
+# at all). The concepts differ — one gates the pipeline, the other gates
+# scoping within a pipeline that's already running — but adding a new
+# cross-cutting root-level file generally warrants updating both. There's no
+# programmatic link, so keep them in sync by convention.
+FULL_RUN_PATTERNS: tuple[re.Pattern[str], ...] = (
+    re.compile(r"^package\.json$"),
```
CraigMacomber: When applying the same function to a bunch of things in a list, I think it would make more sense to not duplicate the call for each one, and instead use some sort of looping construct. I suspect a list comprehension would work well here, but its been so many years since I wrote python I'm not a great reviewer.

scripts/detect_changed_packages.py
```diff
@@ -0,0 +1,295 @@
...
+    re.compile(r"^common/"),
+    re.compile(r"^scripts/"),
+    re.compile(r"^\.changeset/config\.json$"),
+)
+
+
+def normalize_target_branch(branch: str) -> str:
+    """Azure Repos emits ``refs/heads/main``; GitHub emits just ``main``. Normalize."""
+    if branch.startswith("refs/heads/"):
+        return branch[len("refs/heads/") :]
```
CraigMacomber: I have no idea if/how we control what version of python is used with this, but https://docs.python.org/3.9/library/stdtypes.html#str.removeprefix seems like it should be used here unless we need to support python too old for that.

scripts/test/test_detect_changed_packages.mjs
```diff
@@ -0,0 +1,150 @@
...
+	const match = checkFullRunPatterns([".npmrc"]);
+	assert.ok(match);
+	assert.equal(match.source, "^\\.npmrc$");
+});
+
+test("checkFullRunPatterns matches .nvmrc", () => {
+	const match = checkFullRunPatterns([".nvmrc"]);
+	assert.ok(match);
+	assert.equal(match.source, "^\\.nvmrc$");
+});
```
tylerbutler: These seem duplicative and don't add much value.

tylerbutler:
Marking as a draft since I have rewritten this as a flub command. Assuming that lands, this will be closed.

PR #27135, opened 2026-04-22
fix(odsp-driver): Add telemetry so we can distinguish hangs at different points during the snapshot fetching process

lindsnguyen:
## Description

This is a repair item for: [Incident-720043682 Details - IcM](https://portal.microsofticm.com/imp/v5/incidents/details/720043682/summary)

Teams thread: [Brent Weatherall: Sev 2 Incident 720043682 : Increase in lower ring timed out load... | Loop > LiveSite ⚠️ | Microsoft Teams](https://teams.microsoft.com/l/message/19:9034865f2a4f49b5ab8df4569dbb0e73@thread.skype/1765240880928?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=422665a0-7ad3-4d0d-9171-e8881d0397d9&parentMessageId=1765240880928&teamName=Loop&channelName=LiveSite%20%E2%9A%A0%EF%B8%8F&createdTime=1765240880928)

The ask here is to improve the telemetry around fetching snapshot ("ObtainSnapshot") to help us identify where things are getting stuck prior to cancellations/failures.

We are just adding two events after the potental methods that are hanging so that in the future we'll know specifically what part hangs if this happens again.

## Testing

 Added a unit test that validates both telemetry events are logged when a snapshot is fetched.

fixes [AB#55435](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/55435)

### DIFF
--- a/packages/drivers/odsp-driver/src/fetchSnapshot.ts
+++ b/packages/drivers/odsp-driver/src/fetchSnapshot.ts
@@ -4,6 +4,7 @@
  */
 
 import { fromUtf8ToBase64 } from "@fluid-internal/client-utils";
+import { LogLevel } from "@fluidframework/core-interfaces";
 import { assert } from "@fluidframework/core-utils/internal";
 import { getW3CData } from "@fluidframework/driver-base/internal";
 import {
@@ -731,6 +732,7 @@ export const downloadSnapshot = mockify(
 		tokenFetchOptions: TokenFetchOptionsEx,
 		loadingGroupIds: string[] | undefined,
 		snapshotOptions: ISnapshotOptions | undefined,
+		logger?: TelemetryLoggerExt,
 		snapshotFormatFetchType?: SnapshotFormatSupportType,
 		controller?: AbortController,
 		epochTracker?: EpochTracker,
@@ -786,6 +788,7 @@ export const downloadSnapshot = mockify(
 			{ ...tokenFetchOptions, request: { url, method } },
 			"downloadSnapshot",
 		);
+		logger?.send({ category: "generic", eventName: "SnapshotAuthHeaderObtained" }, LogLevel.info);
 		assert(authHeader !== null, 0x1e5 /* "Storage token should not be null" */);
 		const { body, headers } = getFormBodyAndHeaders(odspResolvedUrl, authHeader, header);
 		const fetchOptions = {
@@ -813,6 +816,7 @@ export const downloadSnapshot = mockify(
 			true,
 			scenarioName,
 		) ?? fetchHelper(url, fetchOptions));
+		logger?.send({ category: "generic", eventName: "SnapshotFetchResponseReceived" }, LogLevel.info);
 
 		return {
 			odspResponse,
--- a/packages/drivers/odsp-driver/src/fetchSnapshot.ts
+++ b/packages/drivers/odsp-driver/src/fetchSnapshot.ts
@@ -788,8 +788,8 @@ export const downloadSnapshot = mockify(
 			{ ...tokenFetchOptions, request: { url, method } },
 			"downloadSnapshot",
 		);
-		logger?.send({ category: "generic", eventName: "SnapshotAuthHeaderObtained" }, LogLevel.info);
 		assert(authHeader !== null, 0x1e5 /* "Storage token should not be null" */);
+		logger?.send({ category: "generic", eventName: "SnapshotAuthHeaderObtained" }, LogLevel.info);
 		const { body, headers } = getFormBodyAndHeaders(odspResolvedUrl, authHeader, header);
 		const fetchOptions = {
 			body,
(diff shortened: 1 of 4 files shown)


packages/drivers/odsp-driver/src/fetchSnapshot.ts
```diff
@@ -813,6 +816,7 @@ export const downloadSnapshot = mockify(
 			true,
 			scenarioName,
 		) ?? fetchHelper(url, fetchOptions));
+		logger?.send({ category: "generic", eventName: "SnapshotFetchResponseReceived" }, LogLevel.info);
```
shlevari: Do we want LogLevel.info for these? The will get sampled down to 2% of sessions, if the main motivation for this telemetry is help with livesite that might not be the best. But, we also don't know how often this will be called so we don't know the impact to Loop COGS. I think Info is right for now, but, we should create a follow up item to reassess once we see how often this actually gets called.

packages/drivers/odsp-driver/src/test/fetchSnapshot.spec.ts
```diff
@@ -723,6 +724,86 @@ describe("Tests1 for snapshot fetch", () => {
...
+		assert(
+			mockLogger.matchEvents([
+				{ eventName: "SnapshotAuthHeaderObtained", category: "generic" },
+				{ eventName: "SnapshotFetchResponseReceived", category: "generic" },
+			]),
+			"SnapshotAuthHeaderObtained and SnapshotFetchResponseReceived events should be emitted at info log level",
+		);
+	});
+
+	it("SnapshotAuthHeaderObtained and SnapshotFetchResponseReceived events are filtered at essential log level", async () => {
```
shlevari: Do we really need this test? This feels like its just testing whether the info/sampling feature works, which feels like it should be tested elsewhere more strictly. From a testing perspective, I would more just want to make sure we have a test where if the scenario is hit it logs the event.

packages/drivers/odsp-driver/src/fetchSnapshot.ts
```diff
@@ -731,6 +732,7 @@ export const downloadSnapshot = mockify(
 		tokenFetchOptions: TokenFetchOptionsEx,
 		loadingGroupIds: string[] | undefined,
 		snapshotOptions: ISnapshotOptions | undefined,
+		logger?: TelemetryLoggerExt,
```
shlevari, line 737: I may be overthinking here, but, is it safe to add an optional parameter at the front of the optional queue? I worry that would be a potentially risky change. Does it make sense to put it at the end instead?

packages/drivers/odsp-driver/src/fetchSnapshot.ts
lindsnguyen: Are you suggesting we change it to essential? My undrestanding is that osdp doesn't get called a ton so maybe that's the argument to change it but also this type of log definately feels more info then essential. I can file something.

packages/drivers/odsp-driver/src/fetchSnapshot.ts
lindsnguyen: I originally had it at the end but there were just a bunch of undefined vars I had to pass in to all the spots where the other optional params were so actaully putting it earlier made the code cleaner.

Can you explain why it wouldn't be safe?

I did notice that the params declared in the doc above was out of date though, so updating that accordingly.

packages/drivers/odsp-driver/src/fetchSnapshot.ts
lindsnguyen: followup filed: https://dev.azure.com/fluidframework/internal/_workitems/edit/71402

shlevari (approved):

packages/drivers/odsp-driver/src/test/fetchSnapshot.spec.ts
```diff
@@ -723,6 +723,35 @@ describe("Tests1 for snapshot fetch", () => {
 		);
 	});
 
+	it("SnapshotAuthHeaderObtained and SnapshotFetchResponseReceived events are logged on snapshot fetch", async () => {
+		const snapshot: ISnapshot = {
+			blobContents,
+			snapshotTree: snapshotTreeWithGroupId,
+			ops: [],
+			latestSequenceNumber: 0,
```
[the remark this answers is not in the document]
lindsnguyen: dropping, description is out of date.

packages/drivers/odsp-driver/src/fetchSnapshot.ts
```diff
@@ -787,6 +792,10 @@ export const downloadSnapshot = mockify(
 			"downloadSnapshot",
 		);
 		assert(authHeader !== null, 0x1e5 /* "Storage token should not be null" */);
+		logger?.send(
+			{ category: "generic", eventName: "SnapshotAuthHeaderObtained" },
```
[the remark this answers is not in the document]
lindsnguyen: Changing it to this because we're in an internal file and therefore using sendTelemetryEvent makes sense.

MarioJGMsoft (approved):
The changes look good to me. I don't think there should be any issues with adding the logger.

PR #27141, opened 2026-04-23
Misc. improvements to collaborative text controls and `text-editor` example application

Josmithr:
- Removes the hardcoded headers from above exported text controls in `react` and `quill-react` - this isn't something we should be exporting as a part of our library. Users can add their own headings as needed.
- Updates exported text controls to ensure their entire editor space is clickable and causes the editor to be selected (previously, you could only accomplish this by clicking on a line that contained text, or the first line only if empty).
- Updates `quill`-based editor components to perform necessary cleanup on unmount.

Also updates the `text-editor` app to display all 3 editors at once, rather than only one at a time. Each is displayed as a collapsable card.

image

### DIFF
--- a/examples/data-objects/text-editor/src/app.tsx
+++ b/examples/data-objects/text-editor/src/app.tsx
@@ -331,7 +331,7 @@ const UserPanel: FC<{
 						</button>
 						{/*
 						  * Note: we are intentionally forcing the editor components to be unmounted when their respective cards are collapsed.
-						  * We are doing this to make it possible to do use this app to do performance analysis on individual editor components in isolation.
+						  * We are doing this to make it possible to use this app to do performance analysis on individual editor components in isolation.
 						  */}
 						{isExpanded && (
 							<div id={`${viewType}-panel`} style={{ padding: "12px" }}>
--- a/examples/data-objects/text-editor/src/app.tsx
+++ b/examples/data-objects/text-editor/src/app.tsx
@@ -330,9 +330,9 @@ const UserPanel: FC<{
 							</span>
 						</button>
 						{/*
-						  * Note: we are intentionally forcing the editor components to be unmounted when their respective cards are collapsed.
-						  * We are doing this to make it possible to use this app to do performance analysis on individual editor components in isolation.
-						  */}
+						 * Note: we are intentionally forcing the editor components to be unmounted when their respective cards are collapsed.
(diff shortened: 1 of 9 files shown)


examples/data-objects/text-editor/src/app.tsx
```diff
@@ -323,12 +325,12 @@ const UserPanel: FC<{
...
 						>
 							<span>{viewLabels[viewType].description}</span>
-							<span style={{ fontSize: "11px", color: "#666" }}>
+							<span aria-hidden="true" style={{ fontSize: "11px", color: "#666" }}>
 								{isExpanded ? "▲" : "▼"}
 							</span>
 						</button>
 						{isExpanded && (
-							<div style={{ padding: "12px" }}>
+							<div id={`${viewType}-panel`} style={{ padding: "12px" }}>
```
brrichards, line 337: I think this will still destroy the quill instance on collapse. I think `display: isExpanded ? undefined: "none"` would allow quill to persist, but if the teardown is the intended behavior then disregard this comment.

Josmithr: This was intentional, yes. Craig had noted that we may want to use this app to do some performance measurements. To make it easier to isolate the different editors, we wanted to ensure that collapsing actually cleans up anything that may pollute measurements.

But this would be a good thing to document in the code, because it's definitely non-standard / non-obvious. I'll add some notes.

brrichards: Ah that makes sense. Thanks for the explanation!

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -277,25 +286,66 @@ const UserPanel: FC<{
...
+							}}
+						>
+							<span>{viewLabels[viewType].description}</span>
+							<span aria-hidden="true" style={{ fontSize: "11px", color: "#666" }}>
+								{isExpanded ? "▲" : "▼"}
+							</span>
+						</button>
+						{/*
+						  * Note: we are intentionally forcing the editor components to be unmounted when their respective cards are collapsed.
+						  * We are doing this to make it possible to do use this app to do performance analysis on individual editor components in isolation.
```
brrichards: nit:
```suggestion
						  * We are doing this to make it possible to use this app to do performance analysis on individual editor components in isolation.
```

brrichards (approved):
Left a small nit, but changes look good to me!

brrichards (approved):

PR #27150, opened 2026-04-23
feat(container-definitions): expose resolvedUrl on IContainerContext

jatgarg:
## Description

 Partner teams building on top of the container runtime need access to document identity (`siteUrl` / `driveId` / `itemId` on an ODSP resolved URL, or equivalent fields on other drivers' resolved URL subtypes) from inside the runtime layer. Today the only available path is `runtime.getAbsoluteUrl('/')` → `urlResolver.getAbsoluteUrl(...)`, which is indirect and — depending on the driver — can pull in sharing-link generation concerns that are irrelevant when the caller just wants document identity.

  `IContainer.resolvedUrl` already carries this information on the host side, but `IContainerContext` did not surface it to the runtime. Hosts have been working around this by capturing `container.resolvedUrl` host-side and smuggling it down through the dependency synthesizer, which leaks Fluid-specific concerns into the host layer.                                                                                                         
                                                         
  This PR adds an optional `resolvedUrl?: IResolvedUrl | undefined` property to `IContainerContext`, mirroring `IContainer.resolvedUrl`. `ContainerContext` is wired via a `getResolvedUrl` callback so the value tracks the container's detached-to-attached lifecycle. Runtime code can now read document identity directly off the context, no resolver round-trip required.                                                                    
                                                         
  The property is optional for Loader/Runtime layer-compat — runtimes running against older Loaders will see `undefined` and must tolerate its absence.

## Reviewer Guidance

### DIFF
--- a/packages/common/container-definitions/src/runtime.ts
+++ b/packages/common/container-definitions/src/runtime.ts
@@ -16,6 +16,7 @@ import type {
 import type {
 	ISnapshot,
 	IDocumentMessage,
+	IResolvedUrl,
 	ISnapshotTree,
 	ISummaryContent,
 	IVersion,
@@ -405,6 +406,18 @@ export interface IContainerContext {
 	 * TODO: Optional for backwards compatibility. Make non-optional in version 0.19
 	 */
 	getAbsoluteUrl?(relativeUrl: string): Promise<string | undefined>;
+
+	/**
+	 * Represents the resolved url to the Container.
+	 * Will be undefined only when the container is in the {@link AttachState.Detached | detached} state.
+	 *
+	 * @remarks
+	 * Exposes the same value as `IContainer.resolvedUrl` to the runtime layer so
+	 * that runtime code can read document identity (including driver-specific
+	 * fields carried on the driver's resolved URL subtype) without routing
+	 * through {@link IContainerContext.getAbsoluteUrl} and the URL resolver.
+	 */
+	readonly resolvedUrl?: IResolvedUrl | undefined;
 
 	/**
 	 * Indicates the attachment state of the container to a host service.
--- a/.changeset/resolved-url-on-context.md
+++ b/.changeset/resolved-url-on-context.md
@@ -1,0 +1,15 @@
+---
+"@fluidframework/container-definitions": minor
+"@fluidframework/container-loader": minor
+"__section": legacy
+---
+Expose `resolvedUrl` on `IContainerContext`
+
+`IContainerContext` now carries an optional
+`resolvedUrl?: IResolvedUrl | undefined` property, mirroring
+`IContainer.resolvedUrl` on the runtime side. Runtime code can read document
+identity (including driver-specific fields carried on the driver's resolved URL
+subtype) directly from the context, instead of routing through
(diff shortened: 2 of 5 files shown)


packages/common/container-definitions/src/runtime.ts
```diff
@@ -406,6 +407,18 @@ export interface IContainerContext {
...
+	 * Represents the resolved url to the Container.
+	 * Will be undefined only when the container is in the {@link AttachState.Detached | detached} state.
+	 *
+	 * @remarks
+	 * Exposes the same value as `IContainer.resolvedUrl` to the runtime layer so
+	 * that runtime code can read document identity (including driver-specific
+	 * fields carried on the driver's resolved URL subtype) without routing
+	 * through {@link IContainerContext.getAbsoluteUrl} and the URL resolver.
+	 */
+	readonly resolvedUrl?: IResolvedUrl | undefined;
```
jatgarg, line 420: @anthony-murphy Do we need to make it optional? Since loader and container context are in same layer, should it just be non optional?

ChumpChief (changes requested):
The premise of this PR seems to be that customers plumbing resolvedUrl info down to the runtime via dependencySynthesizer or other mechanism they control is a problem.  However, IMO it is preferable as it allows for separation of concerns and keeps the customer in control of whether the runtime has access to this info.  Same reasoning why we removed the container id from the context back in #16180.

Providing this ambiently also "spaghettis" the resolvedUrl contract through more layers which is a negative for agility.

I'd like to see more context justifying why customers shouldn't make the explicit choice whether to expose this info to the container under a contract they control.

jatgarg:
> The premise of this PR seems to be that customers plumbing resolvedUrl info down to the runtime via dependencySynthesizer or other mechanism they control is a problem. However, IMO it is preferable as it allows for separation of concerns and keeps the customer in control of whether the runtime has access to this info. Same reasoning why we removed the container id from the context back in #16180.
> 
> Providing this ambiently also "spaghettis" the resolvedUrl contract through more layers which is a negative for agility.
> 
> I'd like to see more context justifying why customers shouldn't make the explicit choice whether to expose this info to the container under a contract they control.

@ChumpChief From Alice from Loop:
"Our plumbing (the PR I attempted) would leak FF specific internals to host layer (which doesn't necessarily use fluid, as we also have non-fluid components). Today, we use getAbsoluteUrl to get a sharing link, which is already on IContainerContext. Our ask is we get access to a non-sharing link (resolvedUrl) when we don't need a sharing link, which is exactly the same kind of contract."

Also, what do you exactly mean by this:
"Providing this ambiently also "spaghettis" the resolvedUrl contract through more layers which is a negative for agility."

They are just reading it using containerContext without it being available on the runtime layer. ContainerContext is still part of loader layer.

ChumpChief:
> Also, what do you exactly mean by this: "Providing this ambiently also "spaghettis" the resolvedUrl contract through more layers which is a negative for agility."
> 
> They are just reading it using containerContext without it being available on the runtime layer. ContainerContext is still part of loader layer.

IContainerContext is the primary loader<->runtime interface; anything exposed on it becomes subject to cross-layer compatibility contracts.

Will follow up offline on the ask.

PR #27151, opened 2026-04-23
feat(build-cli): bundle size collection and comparison tooling

TommyBrosman:
Adds a local dev workflow for collecting webpack bundle stats from a base
git ref and comparing them against the current ref - per-asset parsed/gzip
deltas, per-entrypoint totals, and per-package size breakdown. Checks out an
isolated shallow clone at the base revision so the outer repo is never
touched. Intended for inner-loop bundle investigations, not CI.

## Description

Core implementation (src/library/bundleSize):
- collectBundle: builds + runs webpack for one revision (local working tree or
  an inner enlistment at a given ref) and saves its analyzer.json.
- compareBundles: diffs two analyzer.json reports into per-asset, per-entrypoint,
  and per-package tables (text + JSON), tolerating differing asset/package sets.
- collectAndCompareBundles: orchestrates both sides then compares; base ref is
  the merge-base of HEAD by default, or used as-is with exactBase. Caches the
  base report by SHA to skip rebuilds.
- Supporting: compareJsonReports, types, analyzer.json readers, index.

flub commands (src/commands/generate, src/commands/check + docs):
- `generate bundleAnalysisRepo`, `check bundleAnalysisReposComparison`, and
  `generate bundleAnalysisReposWithComparison` wrap the above. Regenerated
  docs/generate.md and docs/check.md, with the companion deep-dive in
  docs/bundleAnalysisRepoDetails.md.

Client code (non-shipping):
- Stub script in bundle-size-tests that invokes generate bundleAnalysisReposWithComparison.
- bundle-size-tests now depends on the catalog-pinned buildTools package.
- pnpm-workspace exclusion so the inner enlistment isn't scanned by fluid-build.

## Todo

- [x] Use oclif
- [x] Determine where the scripts should live: bundle-analysis-tools (client) or build-tools.
- [x] README entry.
- [x] Need a better home for the README (currently I'm using the example project, which isn't the right place)
  - Moved to bundleDetails.md alongside the bundle.md auto-generated command documentation.
- [x] Thoughts on the bundle topic for the oclif CLI? Should we keep it, standardize on check, or something else?
  - [x] Standardize on existing verbs generate/check.
- [x] Make a pass over the comments, prefer `@remarks` for verbose "why" comments.
- [x] Determine whether collectAndCompare can be eliminated from the initial version of the scripts.
  - Keeping it for now, though most of the original functionality has been pushed into collect.
- [x] Are there methods that can be reused from the existing bundle size tools?
  - A more in-depth pass will be saved for future PRs.


pnpm-workspace.yaml
```diff
@@ -27,7 +27,9 @@ minimumReleaseAgeExclude: [
 ]
 resolutionMode: highest
 blockExoticSubdeps: true
-trustPolicy: no-downgrade
+# TODO: Do not check this in. Re-enable `trustPolicy: no-downgrade` after the temporary install unblock is no longer needed.
```
TommyBrosman: TODO: Revert and fix the lockfile. This is a workaround for a pnpm-related issue.

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
```diff
@@ -0,0 +1,196 @@
...
+const packageRoot = resolve(scriptDirectory, "..");
+
+/**
+ * Path to the tsx CLI entrypoint.
+ *
+ * `tsx`'s package.json `exports` map does not expose `./dist/cli.mjs` directly,
+ * but its `bin` field points to it. We resolve the package's own package.json
+ * (always exported) and then resolve the bin path relative to that.
+ */
+const tsxCliPath = (() => {
```
TommyBrosman: This might go away if I switch the scripts over to jiti.

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
```diff
@@ -0,0 +1,196 @@
...
+                          incremental detection.
+  --restore-only          Do not collect or compare. Use the state file left by a previous
+                          aborted run to check out the original branch, reinstall
+                          dependencies, and pop the matching stash (if any).
+
+Examples:
+  tsx ./scripts/collectAndCompareBundles.ts
+  tsx ./scripts/collectAndCompareBundles.ts --base-branch main --current-branch feature/my-changes
+  tsx ./scripts/collectAndCompareBundles.ts --clean-analysis-dir --skip-compare
+  tsx ./scripts/collectAndCompareBundles.ts --force-clean-build
```
TommyBrosman: This should be default behavior. I will add an opt-out flag.

pnpm-lock.yaml
```diff
@@ -172,7 +172,7 @@ importers:
         version: 0.64.0(@types/node@22.19.17)
       '@fluidframework/eslint-config-fluid':
         specifier: catalog:eslint
-        version: 9.0.0(@typescript-eslint/utils@8.56.1(eslint@9.39.1(jiti@2.6.1))(typescript@5.4.5))(eslint@9.39.1(jiti@2.6.1))(typescript@5.4.5)
+        version: 9.0.0(@typescript-eslint/utils@8.56.1(eslint@9.39.1(jiti@2.6.1))(typescript@5.9.3))(eslint@9.39.1(jiti@2.6.1))(typescript@5.9.3)
       eslint:
```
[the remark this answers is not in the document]
TommyBrosman: Good catch. My pnpm-lock includes some changes that shouldn't have come along.

ChumpChief (commented):
A few high-level thoughts
1. This script is doing a lot and appears to offer a lot of flexibility, but it might be better to pare it down to a very focused script with few configuration options.  E.g. maybe the goal is to tell the team "run this script regularly in your inner dev loop", in which case simplicity might be the priority.
2. I would treat repo-level clean, build, and webpack as expensive and try to avoid them at all costs.  E.g. probably prefer to only clean/build/webpack from the bundle-size-tests package.
3. Might want to use `git merge-base HEAD baseBranch` to find the comparison (rather than just `main`) - Folks with worktree-based setups like myself will often have `main` in an inconvenient place.
4. Probably avoid destructive actions in the repo (e.g. clean, stash to some degree) as this risks interrupting/slowing inner dev loop and reduces the tool's value as a result.  One option might be to create a fresh shallow clone at the reference commit in a temp directory, to leave the current repo untouched at the HEAD.  If you find a nice home for it, you might be able to even treat this as a cached output for faster repeat comparisons.
5. bundle-size-tests does seem like a good home for this

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
```diff
@@ -0,0 +1,191 @@
...
+	}
+	const optionArg = argv[index];
+	if (optionArg === undefined) {
+		return undefined;
+	}
+	if (optionArg.startsWith(optionPrefix)) {
+		return optionArg.slice(optionPrefix.length);
+	}
+	return argv[index + 1];
+}
```
tylerbutler: NONONONONONONONONONO please don't write argument parsing code. We have flub!

TommyBrosman: Yep, that was essentially the 2nd bullet of "things I want feedback on" in the PR description. I think you're saying it would be better to extend BaseCommand and use oclif for arguments?

TommyBrosman:
@ChumpChief:

> A few high-level thoughts
> 
> 1. This script is doing a lot and appears to offer a lot of flexibility, but it might be better to pare it down to a very focused script with few configuration options.  E.g. maybe the goal is to tell the team "run this script regularly in your inner dev loop", in which case simplicity might be the priority.

I was able to rip out a lot of arguments. One of the simplifications was changing collectBundles to only collect a single bundle. Some current usage examples:

```
# Collect a bundle from an existing revision, cloning a new repo if it does not exist
jiti ./scripts/collectBundle.ts --mode revision --revision v2.20.0

# Compare the bundles in bundleAnalysis/main and bundleAnalysis/current and write the comparison files (one is a plain text table, the other is JSON)
jiti ./scripts/compareBundles.ts

# Collect a bundle from the current branch and another from the revision with branch/label name v2.20.0, then generate a comparison
jiti ./scripts/collectAndCompareBundles.ts --base-revision v2.20.0
```

> 2. I would treat repo-level clean, build, and webpack as expensive and try to avoid them at all costs.  E.g. probably prefer to only clean/build/webpack from the bundle-size-tests package.

I ripped out the cleans-the-repo-by-default behavior. It was sketchy. I do have a  --force-clean-build in collectBundle still.

> 3. Might want to use `git merge-base HEAD baseBranch` to find the comparison (rather than just `main`) - Folks with worktree-based setups like myself will often have `main` in an inconvenient place.

See collectAndCompareBundles.ts#resolveMergeBase. I haven't used the worktree feature before, so let me know if this looks right.

> 4. Probably avoid destructive actions in the repo (e.g. clean, stash to some degree) as this risks interrupting/slowing inner dev loop and reduces the tool's value as a result.  One option might be to create a fresh shallow clone at the reference commit in a temp directory, to leave the current repo untouched at the HEAD.  If you find a nice home for it, you might be able to even treat this as a cached output for faster repeat comparisons.

collectBundle now checks out a git enlistment inside the bundleAnalysis directory. Note: I had to add an exclusion to the base workspace to get it to build correctly. Let me know if this seems reasonable.

> 5. bundle-size-tests does seem like a good home for this

Cool, I'll leave them there.

examples/utils/bundle-size-tests/scripts/collectBundles.ts
```diff
@@ -0,0 +1,586 @@
...
+ * @returns True if a stash entry was created, false if there was nothing to stash.
+ */
+function stashChanges(marker: string): boolean {
+	const output = execSync(`git stash push -u -m "${marker}"`, {
+		cwd: repoRoot,
+		encoding: "utf-8",
+	}).trim();
+	if (output.includes("No local changes")) {
+		return false;
+	}
```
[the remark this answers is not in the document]
TommyBrosman: No longer applies.

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
TommyBrosman: The scripts now use oclif. I had to build a helper module for showing help strings because I'm not running under a true oclif CLI, but other than that it was a pretty easy port.

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
```diff
@@ -0,0 +1,174 @@
...
+	if (result.error !== undefined) {
+		throw new Error(
+			`Failed to launch script ${scriptName}: ${result.error.message}`,
+		);
+	}
+	if (result.status !== 0) {
+		throw new Error(
+			`Script ${scriptName} exited with code ${result.status ?? "null (signal)"}.`,
+		);
+	}
```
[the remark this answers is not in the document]
TommyBrosman: It prints a message for this condition in the ternary-if.

pnpm-workspace.yaml
```diff
@@ -19,6 +19,9 @@ packages:
   - "!**/dist/**"
   - "!**/lib/**"
   - "!**/src/**"
+  # exclude inner FluidFramework enlistment used by examples/utils/bundle-size-tests
+  # for cross-revision bundle comparison (see examples/utils/bundle-size-tests/scripts/collectBundle.ts).
+  - "!**/bundleAnalysis/**"
```
ChumpChief: Alphabetize list

examples/utils/bundle-size-tests/scripts/compareBundles.ts
```diff
@@ -0,0 +1,452 @@
...
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import { gzipSync } from "node:zlib";
+
+import { decompressStatsFile } from "@fluidframework/bundle-size-tools";
```
ChumpChief: In my next PR I'm going to be adding output from the "JSON mode" from the bundle analyzer plugin (which also includes gzip sizes by default), you might prefer working off of that output instead.  It's smaller and won't require decompression, is a little more straightforward to parse, and you won't have to do your own gzipping.

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
```diff
@@ -0,0 +1,192 @@
...
+			console.log(`Collecting local bundle (label: ${currentLabel})...`);
+			console.log("=".repeat(80));
+			runScript("collectBundle.ts", ["--mode", "local", ...sharedCollectArgs]);
+
+			console.log(`\n${"=".repeat(80)}`);
+			console.log(
+				`Collecting base bundle (revision: ${baseRevision}, label: ${baseLabel})...`,
+			);
+			console.log("=".repeat(80));
+			runScript("collectBundle.ts", [
```
ChumpChief: To make this a speedier inner loop, consider:
1. Check if you've output the stats for a given baseRevision already and skip this step if so (i.e. write them to subdirectories or files named for the commit hash).
2. Consider writing the output of the merge-base revision to a global location (e.g. `/tmp/fluidBundleAnalysis/<commit hash>`) to permit opportunistically sharing this "cache" between clones or worktrees, and to avoid accidental cleanup.

examples/utils/bundle-size-tests/tsconfig.scripts.json
ChumpChief, line 1: Should this be included in some build script for validation?  I see it's referenced in lint but seems like there's some risk that compile breaks could sneak in if type checking doesn't run in CI.

examples/utils/bundle-size-tests/package.json
```diff
@@ -21,8 +21,11 @@
 		"check:biome": "biome check .",
 		"check:format": "npm run check:biome",
 		"clean": "rimraf --glob build dist lib bundleAnalysis \"**/*.tsbuildinfo\" \"**/*.build.log\" nyc",
-		"eslint": "eslint --quiet --format stylish src",
-		"eslint:fix": "eslint --quiet --format stylish src --fix --fix-type problem,suggestion,layout",
+		"collect:bundle": "jiti ./scripts/collectBundle.ts",
+		"collect:compare:bundles": "jiti ./scripts/collectAndCompareBundles.ts",
+		"compare:bundles": "node --max-old-space-size=8192 --import jiti/register ./scripts/compareBundles.ts",
```
ChumpChief: Maybe suggest narrowing this to just one easy command that we'd publicize to devs as part of their inner loop (i.e. just `compare`).  I think this is collectAndCompareBundles.ts.

Also probably add dev guidance to the README.md for what folks on the team should be doing with this tool.

examples/utils/bundle-size-tests/scripts/collectBundle.ts
```diff
@@ -0,0 +1,309 @@
...
+ * `compareBundles.ts` reads from `<bundleAnalysisDirectory>/<label>/bundleStats.msp.gz`.
+ *
+ * Lives under this package's `bundleAnalysis/` directory, which is matched by
+ * the repo-wide `.gitignore` entry for `bundleAnalysis`.
+ *
+ * Note: `npm run clean` in this package rimrafs `bundleAnalysis/`, so any state
+ * here (including the inner repo clone) is wiped on clean. The inner repo is
+ * re-cloned automatically on next use.
+ */
+const bundleAnalysisDirectory = resolve(outerPackageRoot, "bundleAnalysis");
```
ChumpChief: The bundleAnalysis directory gets uploaded wholesale as a pipeline artifact

* https://github.com/microsoft/FluidFramework/blob/main/tools/pipelines/templates/build-npm-client-package.yml#L411
* https://github.com/microsoft/FluidFramework/blob/main/package.json#L40
* https://github.com/microsoft/FluidFramework/blob/main/build-tools/packages/build-cli/src/commands/generate/bundleStats.ts#L46

This isn't a problem if your script isn't run in the pipeline, but might be good to preemptively pick a distinct location to avoid the risk of uploading the entire repo clone as an artifact.

examples/utils/bundle-size-tests/scripts/compareBundles.ts
```diff
@@ -0,0 +1,452 @@
...
+		...new Set([...Object.keys(baseAssets), ...Object.keys(currentAssets)]),
+	]
+		.sort()
+		.map((name) => ({
+			name,
+			baseStatSize: baseAssets[name]?.size ?? 0,
+			currentStatSize: currentAssets[name]?.size ?? 0,
+		}));
+
+	reporter.section(`=== Bundle Size Comparison: ${baseLabel} -> ${currentLabel} ===`);
```
ChumpChief: This probably factors a lot neater if you have `runCompare()` produce a comparison object (I think this might be the jsonObject that you pass to writeOutputFiles at the end), and then split out a dedicated `renderAsText()` that takes that comparison in.  Just so the comparison logic isn't mixed with rendering.

examples/utils/bundle-size-tests/scripts/compareBundles.ts
```diff
@@ -0,0 +1,452 @@
...
+ *
+ * @param outputDirectory - The directory where output files will be written
+ * @param textContent - The formatted text comparison report
+ * @param jsonObject - The structured comparison data as a JSON-serializable object
+ */
+function writeOutputFiles(
+	outputDirectory: string,
+	baseLabel: string,
+	textContent: string,
+	jsonObject: object,
```
ChumpChief: This would be nice to strongly type

examples/utils/bundle-size-tests/scripts/compareBundles.ts
```diff
@@ -0,0 +1,452 @@
...
+			entrypointName,
+			baseSize,
+			currentSize,
+			diff: currentSize - baseSize,
+		});
+	}
+
+	writeOutputFiles(outputDirectory, baseLabel, reporter.toText(), {
+		comparedAt: new Date().toISOString(),
+		baseLabel,
```
ChumpChief: Probably want a commit hash here rather than (or in addition to) the branch name.

Same for currentLabel, maybe with some flag for uncommitted changes?

examples/utils/bundle-size-tests/scripts/collectBundle.ts
```diff
@@ -0,0 +1,309 @@
...
+		"force-clean-build": Flags.boolean({
+			description:
+				"Run the full workspace clean ('npm run clean' at the repo root) before " +
+				"building. Off by default; opt in when stale incremental build state from a " +
+				"previous revision may interfere with the current one.",
+			default: false,
+		}),
+	};
+
+	public async run(): Promise<void> {
```
ChumpChief: Maybe consider showing a warning if this is run with uncommitted local changes (since this will make it difficult to reproduce the results as more changes are added)?

examples/utils/bundle-size-tests/scripts/collectBundle.ts
```diff
@@ -0,0 +1,309 @@
...
+async function ensureInnerRepo(): Promise<void> {
+	if (existsSync(resolve(innerRepoRoot, ".git"))) {
+		return;
+	}
+
+	const originUrl = await getOuterOriginUrl();
+	mkdirSync(dirname(innerRepoRoot), { recursive: true });
+	console.log(`\nCloning inner repo from ${originUrl} into ${innerRepoRoot}...`);
+	// Filter blobs to keep the clone fast; blobs are fetched lazily on checkout.
+	await simpleGit(dirname(innerRepoRoot)).clone(originUrl, innerRepoRoot, [
```
ChumpChief: Maybe make this a shallow clone too?  Depending on what you want to do for caching results, if you make a separate depth=0 clone for each revision under test (rather than checking out revisions on a long-lived repo) might be preferable.

You can also probably delete the repo after collecting the output if the caching is trustworthy.

build-tools/packages/build-cli/src/library/bundleSize/compareBundles.ts
```diff
@@ -0,0 +1,452 @@
...
+	outputDirectory: string,
+	baseLabel: string,
+	textContent: string,
+	jsonObject: object,
+): void {
+	mkdirSync(outputDirectory, { recursive: true });
+
+	const baseSuffix = sanitizeForFileName(baseLabel);
+	const currentSuffix = sanitizeForFileName(currentLabel);
+	const outputBaseName = `compare-${baseSuffix}-to-${currentSuffix}`;
```
ChumpChief, line 554: This will tend to overwrite prior analyses, which might be annoying if I'm iterating in an inner loop and comparing between runs.  Maybe there's a reasonable convention we could use to ensure unique filenames across runs?  timestamp, hash?

ChumpChief (commented):
Tool seems pretty cool, I like the direction :)

examples/utils/bundle-size-tests/tsconfig.scripts.json
TommyBrosman: Fixed. I used the `compile.dependsOn` feature in flub.

build-tools/packages/build-cli/src/library/bundleSize/compareBundles.ts
TommyBrosman: That's fair, I can change it. If I wanted to make it really clever I could also create a patch of the repo's working state to go along with each entry, but that seems like overkill. I'm (now) also trying to avoid running git commands on the outer repo in this script.

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
TommyBrosman: I implemented 1, still thinking about 2. In an early version of this script I wrote some files to a temp directory and others to the bundleAnalysis directory. Remembering which files were where was slightly annoying, though in theory I could just put a symlink to the temp cache of bundle analyses. I'll think about it more.

examples/utils/bundle-size-tests/scripts/collectBundle.ts
TommyBrosman: I implemented shallow-clone. I think deleting the inner repo makes sense, though I'll add a flag for debug that retains it.

examples/utils/bundle-size-tests/scripts/collectBundle.ts
TommyBrosman: I think running with uncommitted local changes will be the norm for the current branch. The base branch is a fresh checkout. I could generate a git patch and save it alongside the analysis results.

build-tools/packages/build-cli/src/library/bundleSize/compareBundles.ts
TommyBrosman: I ended up changing it to dump a patch. Let me know if it seems like overkill.

examples/utils/bundle-size-tests/scripts/collectBundle.ts
TommyBrosman: I don't think my script will ever be run in a pipeline, but it might not be a bad idea to make a new directory given how different the output is.

examples/utils/bundle-size-tests/scripts/compareBundles.ts
TommyBrosman: This is handled now in the form of a patch + timestamp. It's not exactly idempotent, and probably overkill. It seems nice for reproducibility though.

examples/utils/bundle-size-tests/scripts/compareBundles.ts
TommyBrosman: I added a computeComparison helper function. It cleaned up the code a lot, and will be useful if I need to add JSON output.

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
```diff
@@ -0,0 +1,250 @@
...
+ * resolved (e.g. unknown branch, detached state with no shared history).
+ *
+ * Using merge-base instead of the raw branch tip means the comparison is taken
+ * against the actual fork point, which is what users typically want — and it
+ * works for worktree-based setups where `main` may not exist as a local branch
+ * at the location they expect.
+ */
+async function resolveMergeBase(rev: string): Promise<string | undefined> {
+	try {
+		const output = await simpleGit(packageRoot).raw(["merge-base", "HEAD", rev]);
```
ChumpChief: Might be nice to parameterize HEAD here?  But optional.

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
```diff
@@ -0,0 +1,250 @@
...
+ * routing through a stable, documented entry point exposed by jiti's package
+ * exports map.
+ *
+ * @param scriptName - Script file name under ./scripts/
+ * @param scriptArgs - Arguments to forward to the script
+ */
+function runScript(scriptName: string, scriptArgs: string[]): void {
+	const scriptPath = resolve(scriptDirectory, scriptName);
+	// Bump the V8 heap so compareBundles.ts can hold two decompressed webpack
+	// stats objects in memory without hitting the default ~4 GB OOM.
```
ChumpChief: Shouldn't be necessary if you switch to using the analyzer.json that we're creating now which are <1MB each.

ChumpChief (commented):
Changes since my last review look pretty good, just a couple other random comments

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
TommyBrosman: I thought about this more. This script typically won't compare against multiple baseRevisions, so I think adding a set of hash-named repos in a tempdir is likely overkill. I typically like to be able to find my temp repos as well considering the amount of troubleshooting needed if the script or build breaks. I'll keep the directory structure as-is for now.

examples/utils/bundle-size-tests/scripts/collectBundle.ts
TommyBrosman: I'm going to pick another directory in the same outer directory. That way I won't pollute bundleAnalysis and it will still be easy to find.

examples/utils/bundle-size-tests/package.json
TommyBrosman: Done. The remaining script entry points to collectAndCompareBundles. I also added a README in scripts going over when to use which.

examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts
TommyBrosman: Done.

examples/utils/bundle-size-tests/src/fluidFrameworkAll.ts
```diff
@@ -0,0 +1,55 @@
...
+ * throws at evaluation time and exposes no `apisToBundle` to invoke.
+ */
+export function apisToBundle(): void {
+	// Reference every entrypoint so none of them are tree-shaken out of this
+	// aggregate bundle. The return values are intentionally unused.
+	aqueduct();
+	azureClient();
+	connectionState();
+	containerRuntime();
+	experimentalSharedTree();
```
ChumpChief: I like the idea of a comprehensive bundle, but I'd suggest rather than an "everything" bundle try to make one or two bundles that are representative of normal customer use.  E.g. no user will use both azureClient and odspClient/odspDriver/etc., probably they will just pick one of sharedTree vs. experimentalSharedTree, etc.

That way it'll be a more useful proxy for what a real customer will see.

build-tools/packages/build-cli/package.json
```diff
@@ -225,6 +225,9 @@
 			"bump": {
 				"description": "Bump the version of packages, release groups, and their dependencies."
 			},
+			"bundle": {
```
ChumpChief: You mention in the reviewer guidance to consider whether to add `bundle` - I'd probably go with some combination of `check` + `generate` (we do this for the other bundle tools already and seems to fit well).

build-tools/packages/build-cli/src/library/bundleSize/collectBundle.ts
```diff
@@ -0,0 +1,278 @@
...
+}
+
+/**
+ * Captures the outer repo's currently-staged diff to a sibling file next to
+ * the bundle stats, so the analysis is reproducible even with uncommitted
+ * changes. Only staged changes are recorded; unstaged changes are intentionally
+ * excluded (they're often noisy / random) but a warning is printed if any are
+ * detected so the user can `git add` the relevant pieces and re-run.
+ *
+ * This is purely a record: the patch is never applied, and the outer repo's
```
ChumpChief: In general, the tsdoc in this PR is very verbose - consider how it can be made terser.  In particular I see a lot of places where it's just directly describing how the internals of the function work, prefer to let the code speak for itself in that case (or use inline comments where needed).

Also consider splitting a terse top-level tsdoc comment from an `@remarks` block if there is value in detailed commentary beyond just the contract of the function.

build-tools/packages/build-cli/src/commands/generate/bundleAnalysisReposWithComparison.ts
ChumpChief, line 1: Nit, we don't typically kebab-case .ts files (camel case instead).

build-tools/packages/build-cli/src/commands/bundle/collectAndCompare.ts
```diff
@@ -0,0 +1,109 @@
...
+	];
+
+	public static readonly flags = {
+		"base-revision": Flags.string({
+			description:
+				"Revision to use as the comparison baseline (branch, tag, or commit SHA). The " +
+				"actual base used is the merge-base of HEAD and this revision (the fork point), so " +
+				"worktree-based setups where 'main' is in an unusual location still produce the " +
+				"expected comparison. Pass --exact-base to use the revision as-is instead.",
+			default: "main",
```
ChumpChief: Rather than defaulting to local `main` (which for those of us using git worktrees tends to be stale compared to upstream/main) consider leveraging pickFreshestRemote and using the `main` from there.

This surprised me when trying the branch out locally (claimed that a bunch of packages were changing that actually weren't).

build-tools/packages/build-cli/src/library/bundleSize/collectAndCompareBundles.ts
```diff
@@ -0,0 +1,246 @@
...
+			});
+			// Record the SHA that produced this report so a subsequent run
+			// against the same merge-base can skip the rebuild.
+			mkdirSync(baseLabelDirectory, { recursive: true });
+			writeFileSync(baseRevisionMarkerPath, `${baseRevision}\n`);
+
+			// Delete the inner repo now that the report is saved. It's a
+			// shallow clone of the outer repo's `origin`, so re-creating it
+			// on the next run is cheap; keeping it around just consumes disk
+			// (hundreds of MB once dependencies are installed).
```
ChumpChief: Disk space probably isn't a major concern (also I don't think it should have a huge impact on disk space since pnpm should just be placing symlinks?), and the deletion is actually kind of slow (thanks node_modules).  Probably consider how to get this out of the critical path, maybe either
* let the rm run async
* defer it to the end after the report has been produced so the user can go ahead and start reviewing it
* clone to a unique location each time (e.g. keyed by SHA) and let the clones accumulate, user runs clean themselves if they feel they need to

examples/utils/bundle-size-tests/webpack.config.cjs
```diff
@@ -74,6 +74,7 @@ module.exports = {
 		directory: "./src/sharedDirectory",
 		experimentalSharedTree: "./src/experimentalSharedTree",
 		fluidFramework: "./src/fluidFramework",
+		fluidFrameworkAll: "./src/fluidFrameworkAll",
```
ChumpChief: I've just learned that webpack makes its chunking decisions globally based on all provided entry points.  Which makes sense, but also represents another deviation from what a realistic customer scenario would look like.

Result is that even though this PR doesn't actually change any of the input packages, your tool reports large deltas because everything gets re-chunked.

Probably doesn't need addressing in this PR, but maybe we should consider breaking each scenario out into individual webpack calls to avoid odd chunking choices.  Filed AB#75724 to track.

build-tools/packages/build-cli/src/library/bundleSize/compareBundles.ts
```diff
@@ -0,0 +1,640 @@
...
+	assets: ComparisonRow[];
+	/** Gzip-size rows, limited to assets whose gzip size changed. */
+	gzipChangedAssets: ComparisonRow[];
+	/**
+	 * Per-entrypoint total parsed-size rows. Each row is a real shipped bundle;
+	 * see {@link compareEntrypointTotals}. Rows overlap and must not be summed —
+	 * the aggregate `fluidFrameworkAll` entrypoint is the single deduplicated
+	 * bundle-wide total.
+	 */
+	entrypoints: ComparisonRow[];
```
ChumpChief: At least currently, it seems like the entrypoints results are just a strict subset of the "all assets" results.  It seems like they would only ever diverge if we (1) continued webpacking all entrypoints simultaneously (see other comment about AB#75724) such that there is sharing opportunity and (2) added extra webpack config to request chunk splitting, which wouldn't be very representative of the customer I think (again because they wouldn't be double-importing these entrypoints anyway).

Maybe consider if this extra table is needed and/or if it's measuring the right thing.  It might be interesting to sum in "all possible dynamically loaded bundles" to each entrypoint's bundle to get a "worst case total" number, but that's probably better deferred to a followup PR.

ChumpChief (commented):
We talked about a couple of these offline already but wanted to get you this set of comments.

I think eventually there's a lot of opportunity to share/reuse with the pipeline bundle comparison tools (i.e. merge for the union of their functionalities), but I think starting with a checkin of a standalone tool like you're doing here is the right move to retain agility as we iterate even if there's some duplicated functionality short-term.

build-tools/packages/build-cli/src/commands/bundle/collectAndCompare.ts
TommyBrosman: Fair, I also hit this with my origin vs. upstream main.

build-tools/packages/build-cli/src/library/bundleSize/collectBundle.ts
TommyBrosman: I made a pass and moved most of the verbose docs into @remarks along with some of the longer inline comments that were more like invariants/architecture.

examples/utils/bundle-size-tests/src/fluidFrameworkAll.ts
TommyBrosman: I planned on building more scoped bundles later (not in this PR) as separate configs. I think there are cases where we want a bundle to not just have a specific set of APIs, but also have different bundler settings. I've been using the "all" target because I don't have the more granular "scenario" approach wired in (that's a later change). I can add a comment or something calling out the fact that this is a synthetic endpoint and not representative of a real consumer. Thoughts?

build-tools/packages/build-cli/src/commands/generate/bundleAnalysisReposWithComparison.ts
TommyBrosman: Fixed.

build-tools/packages/build-cli/src/library/bundleSize/collectAndCompareBundles.ts
TommyBrosman: It's about a 4 GB hit on my machine, and there are times when I am comparing 3 revisions. I think deferring cleanup until the end is the option that works best in the current design.

build-tools/packages/build-cli/src/library/bundleSize/compareBundles.ts
TommyBrosman: That's fair: these entrypoints still correspond to assets. I'll get rid of the additional table. I think the scenario entrypoints (future feature) will need their own table, but that's outside the scope of this PR.

build-tools/packages/build-cli/src/commands/bundle/collectAndCompare.ts
TommyBrosman: Fixed. The collect script uses pickFreshestRemote if the revision is not defined.

build-tools/packages/build-cli/package.json
TommyBrosman: Here's what I ended up with:
- `bundle collect` -> `generate bundleAnalysisRepo`
- `bundle compare` -> `check bundleAnalysisReposComparison`
- `bundle collectAndCompare` -> `generate bundleAnalysisReposWithComparison`

examples/utils/bundle-size-tests/src/fluidFrameworkAll.ts
ChumpChief: I think that given the global chunking shenanigans I mentioned finding in the other comment (AB#75724) I'd probably recommend against changing collateral in this PR.  You can see from the bundle-bot's comment how that flaw causes disruption even to completely unrelated scenarios.

The reason being that this sort of disruption impacts our long-term telemetry tracking progress on bundle goals.  So I'd like to address that issue first, which would then allow us to subsequently add/remove/modify whatever scenarios we like without disrupting all our other telemetry every time.

examples/utils/bundle-size-tests/eslint.config.mts
```diff
@@ -8,6 +8,15 @@ import { recommended } from "@fluidframework/eslint-config-fluid/flat.mts";
...
 	...recommended,
+	{
+		files: ["**/*.ts", "**/*.tsx", "**/*.mts", "**/*.cts"],
+		languageOptions: {
+			parserOptions: {
+				projectService: false,
+				project: ["./tsconfig.json", "./src/test/tsconfig.json"],
+			},
+		},
+	},
```
ChumpChief: There seem to be some unrelated changes in this PR (this eslint change, the compile settings in package.json) - probably strip them out as they don't seem directly related?

build-tools/packages/build-cli/src/commands/generate/bundleAnalysisRepo.ts
```diff
@@ -0,0 +1,103 @@
...
+		"package-dir": Flags.string({
+			description:
+				"Package root whose webpack bundles are built and whose analyzer.json is collected.",
+			default: ".",
+		}),
+		"analysis-dir": Flags.string({
+			description:
+				"Directory under which per-label analyzer stats are saved. Defaults to " +
+				"<package-dir>/compareBundlesOutput/analysis.",
+		}),
```
ChumpChief: Recommend dropping a lot of these flags, defaults seem fine.  Can always add them later if there's a real problem but otherwise it's just more mental load to a potential user and more possible error flows.

ChumpChief (commented):
I think the tool seems fine to go in - I'm sure we'll iterate on it some over time but it seems helpful as-is.

I do think we should probably drop the fluidFrameworkAll bundle for now though, just to minimize disruption to our existing bundle telemetry.  I can try to prioritize fixing AB#75724 to unblock it, but I think that should happen first so we can have as few "telemetry reset" events as possible.

examples/utils/bundle-size-tests/eslint.config.mts
TommyBrosman: Ah, this is leftover from when I moved the scripts out of client. I'll make a pass over the changes.

examples/utils/bundle-size-tests/src/fluidFrameworkAll.ts
TommyBrosman: I've removed it. The scenario-based approach will replace it in a future PR. The fluidFrameworkAll version was not that vital since it doesn't represent a real bundle anyway.

examples/utils/bundle-size-tests/eslint.config.mts
TommyBrosman: Fixed, and I also cleaned up some old lockfile cruft.

build-tools/packages/build-cli/src/commands/generate/bundleAnalysisRepo.ts
TommyBrosman: I ripped out a lot of the parameters, though I'm keeping `--package-dir` as it allows us to get results for other webpack configs/entrypoints (the scenario-based bundle size measurement approach).

ChumpChief (approved):

PR #27153, opened 2026-04-24
fix(tree): support attach/detach edits on multi-node chunks

brrichards:
# Summary

Currently there are no multi-node chunks in chunked forest. This PR is part of the effort to support them before enabling the chunker to batch nodes into multi top-level node chunks, specifically `UniformChunks`. This PR fixes how `detachEdit` and `attachEdit` handle the `chunks` array at `nodeIndex` when in the middle of a multi-node chunk.

### What surfaces the bug
The bug that surfaced when enabling multi-node chunks is in`attachEdit` and `detachEdit` in `chunkedForest.ts`. For both methods they required `nodeIndex` to fall on a chunk boundary in the fields `chunks` array. Once multi-node chunks were enabled this allowed node indexing into the middle of a chunk(not allowed). Example:
<pre>
Field: "x" -> chunks: [ BasicChunk(A), UniformChunk(B, C, D)] 
Attach at nodeIndex = 2
Attempt to attach to node C which isn't on a border(invalid) <- current bug
</pre>

### splitFieldAtIndex and how it resolves the bug
A new helper in `chunkTree` takes a field's `chunks` array and splits the chunk containing `nodeIndex` so that `nodeIndex` lands on a chunk boundary, returning the index of the chunk that now begins at `nodeIndex` for attach/detach to use. If `nodeIndex` already falls on an existing chunk boundary (including a single-node chunk), `chunks` is not mutated.

`nodeIndex` must satisfy `0 <= nodeIndex <= totalNodes`, where `totalNodes` is the sum of `topLevelLength` across all chunks. `nodeIndex === totalNodes` is supported because it is the valid and necessary splice point for attach/detach at the end of the field. Each half of a split chunk is re-chunked through `chunkRange` using the provided policy so the resulting chunks follow the same shape rules as the rest of the field.

`detachEdit` calls this helper method twice, once at each end of the detach range, to isolate the nodes to splice out of `chunks`.

### Testing
These tests are written to allow policy changes to happen, but if detach/attach edit logic changes (don't need to be on a node boundary anymore), then these tests will need to change with that logic change
| File | Tests | 
| ---- | ---- |
| chunkedForest.ts | Two new tests were added that manually create a chunked forest with a field containing a uniform chunk with 3 nodes. One test preforms detach in the middle of the uniform, and the other attach
| chunkTree.ts | Four new tests were added testing splitFieldAtIndex. They test all 3 intended paths. 1. Single node chunk - `chunks` array should not be mutated. 2. At `nodeIndex === totalNodes` - `chunks` array should not be mutated. 3. multi-node chunk - split should occur as intended. Final test calls `splitFieldAtIndex` on an empty field |


brrichards:
> Hey! You look nice today! Want me to review this PR?
> 
> Based on the diff (274 lines, 4 files), I've queued these reviewers:
> 
> * [x]  **Correctness** — logic errors, race conditions, lifecycle issues
> * [x]  **Security** — vulnerabilities, secret exposure, injection
> * [x]  **API Compatibility** — breaking changes, release tags, type design
> * [x]  **Performance** — algorithmic regressions, memory leaks
> * [x]  **Testing** — coverage gaps, hollow tests
> 
> Toggle checkboxes to adjust, then reply **yes** to start — or ask me anything!

yes

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +556,51 @@ export function chunkRange(
...
+			return i;
+		}
+		const chunk = chunks[i] ?? oob();
+		if (remaining < chunk.topLevelLength) {
+			const total = chunk.topLevelLength;
+			const cursor = chunk.cursor();
+			cursor.firstNode();
+			const before = chunkRange(cursor, policy, remaining, false);
+			const after = chunkRange(cursor, policy, total - remaining, true);
+			// TODO: this could fail for really long chunks being split (due to argument count limits).
```
Josmithr: Should we address this now?

CraigMacomber: currently, and like for all time, we will avoid making chunks long enough this is would be an issue since such cases would be really inefficient.

This really is just a case that the splice API is badly design (given the implementation limits in real JS implementations). Might be time to write our own splice if we don't have one already and use that. Annoyingly I think it would be slower than the built in one for the cases we care about.

For now I'm fine having this as is: if it crashes at runtime, we have a bug making massively long chunks on accident and perf is goanna suck.

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -122,4 +134,89 @@ describe("ChunkedForest", () => {
 			});
 		});
 	}
+
+	describe("mutation of chunks array inside a multi-node chunkShape", () => {
+		// Manually install a root field of [uniform(5)] so we can exercise splitting a uniform
+		// chunk at a node index in the middle. The visitor splices the chunk array by node index
+		// as if it were a chunk index, so these tests fail against the unpatched implementation.
```
CraigMacomber: This comment is unclear what exactly it applies to.

If its a comment for people working in this describe block, put a blank line after it so its clear this isn't about `setup`

If its about setup, use a ts-doc style comment instead, and focus more on what is returned by setup.

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -122,4 +134,89 @@ describe("ChunkedForest", () => {
...
+
+			const numberType: TreeNodeSchemaIdentifier = brand(numberSchema.identifier);
+			const numberShape = new TreeShape(numberType, true, []);
+			const uniform = new UniformChunk(numberShape.withTopLevelLength(5), [0, 1, 2, 3, 4]);
+			forest.roots.fields.set(rootFieldKey, [uniform]);
+
+			const detachedKey: FieldKey = brand("detached");
+			const detachedId = { minor: 0 };
+
+			return { forest, numberShape, detachedKey, detachedId };
```
CraigMacomber: I don't get why this is returning anything here except the forest. The other three things are constants which we could declare outside this function (its fine to have immutable data like in the describe block if its cheap to create/allocate.: we just want to avoid things likely to error, are mutable, or slow to create).

Splitting them out would also make them easier to document what they are and/or what they are for.

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -122,4 +134,89 @@ describe("ChunkedForest", () => {
...
+		}
+
+		// Combining the topLevelLengths of all chunks in the `chunks` array yields the total node count.
+		// We don't check the chunk shapes themselves as doing so would cause these tests to be fragile
+		// against policy/chunkRange changes.
+		function nodeCount(chunks: readonly { readonly topLevelLength: number }[]): number {
```
CraigMacomber: If you use the actual ChunkShape type here this code becomes more self documenting (for example the topLevelLength field gets a proper doc comment).

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -122,4 +134,89 @@ describe("ChunkedForest", () => {
...
+			const uniform = new UniformChunk(numberShape.withTopLevelLength(5), [0, 1, 2, 3, 4]);
+			forest.roots.fields.set(rootFieldKey, [uniform]);
+
+			const detachedKey: FieldKey = brand("detached");
+			const detachedId = { minor: 0 };
+
+			return { forest, numberShape, detachedKey, detachedId };
+		}
+
+		// Combining the topLevelLengths of all chunks in the `chunks` array yields the total node count.
```
CraigMacomber: This line seems like it should be a ts-doc comment, the later two seem like they should be comments inside this function's implementation if kept (but I'm unclear on why they exist: I'm not sure how "checking" the shapes is relevant to computing "nodeCount")

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +556,51 @@ export function chunkRange(
...
+	nodeIndex: number,
+	policy: ChunkCompressor,
+): number {
+	assertNonNegativeSafeInteger(nodeIndex);
+	let remaining = nodeIndex;
+	for (let i = 0; i < chunks.length; i++) {
+		if (remaining === 0) {
+			return i;
+		}
+		const chunk = chunks[i] ?? oob();
```
CraigMacomber: using a counting for loop like this, then having to deal with the risk of out of bounds is generally not ideal.

Using a `for of` loop over `Array.entries` is typically a cleaner way if you need both the index and the value.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +556,51 @@ export function chunkRange(
...
+ * `nodeIndex === totalNodes`.
+ */
+export function splitFieldAtIndex(
+	chunks: TreeChunk[],
+	nodeIndex: number,
+	policy: ChunkCompressor,
+): number {
+	assertNonNegativeSafeInteger(nodeIndex);
+	let remaining = nodeIndex;
+	for (let i = 0; i < chunks.length; i++) {
```
CraigMacomber: Also avoid single letter variables and use something more descriptive. This whole PR is fixing a confusion of chunk indexes with node indexes, which is a great example of why calling `i` just `i` might lead to bugs when we could instead use a name like `chunkIndex`

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
brrichards: The comment is for the setup. I'll go ahead and change it so its more focused and in a ts-doc style.

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
brrichards: Yeah that makes sense. I'll change it so we return just forest

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
brrichards: changed to use ChunkShape

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
brrichards: Yeah those could probably be removed. I was being overly wordy for this test, but could probably exclude this. It was more of a note to myself of why we check for nodes and not uniform chunks specifically in the tests

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
brrichards: Will fix this

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
brrichards: My bad, old habit of for loops with i as index. Will change this to `chunkIndex` along with the for of loop change

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -149,26 +165,24 @@ describe("ChunkedForest", () => {
...
-		// against policy/chunkRange changes.
-		function nodeCount(chunks: readonly { readonly topLevelLength: number }[]): number {
+		/**
+		 * Sums the top-level lengths of the given chunks.
+		 *
+		 * @param chunks - The chunks to sum.
+		 * @returns The total number of top-level nodes across the given `chunks`.
+		 */
+		function nodeCount(chunks: readonly TreeChunk[]): number {
 			return chunks.reduce((n, c) => n + c.topLevelLength, 0);
```
CraigMacomber: More descriptive variable names are preferred. Something like this would be better:

```suggestion
			return chunks.reduce((runningTotal, chunk) => runningTotal + chunk.topLevelLength, 0);
```

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -136,9 +137,24 @@ describe("ChunkedForest", () => {
...
+
+		/** Detached field id paired with `detachedKey`. */
+		const detachedId = { minor: 0 };
+
+		/**
+		 * Builds a fresh chunked forest for use in a single test case.
+		 *
+		 * @returns A chunked forest with a root field containing a single uniform chunk of 5 numbers.
+		 */
 		function setup() {
```
CraigMacomber: This code is way easier to follow now, but a further improvement would be to use a more specific name for this. Like setupForest, or forestWithChunk

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -136,9 +137,24 @@ describe("ChunkedForest", () => {
...
 			const builder = new SchemaFactory("chunkedForest.multiNodeUniform");
 			const forestSchema = new TreeStoredSchemaRepository(toInitialSchema(builder.number));
```
CraigMacomber: Number is available as a static:
```suggestion
			const forestSchema = new TreeStoredSchemaRepository(toInitialSchema(SchemaFactory.number));
```

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
brrichards: Thats a good call. I'll change it to setupForest

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
brrichards: Makes sense, I'll refrain from using single letter variable names for things like this from now on

brrichards:
There are some serious memory regressions with this change and should not be merged until the cause is found and addressed. It looks like it's a consequence of having multi-node chunks now and the general cost of editing them. Specifically with sequences of edits and how deltaVisitor deals with them.  

Changes to the loop in splitFieldAtIndex has accounted for most regression, but follow up work could help close the gap with ObjectForest. 
1. combine similar chunks after an edit
2. native splitAt in uniformChunk to avoid cursor walks
3. changes to visitDelta.ts to allow for bulk edits instead of single node operations (Think pasting in 100 text characters. Attach gets called 100 times and generate 100 new single node uniform chunks)

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +556,73 @@ export function chunkRange(
...
 	return output;
 }
+
+/**
+ * {@link splitFieldAtIndex} bisects a chunk at its midpoint and
+ * descends into the half holding the target rather than splitting directly at the requested
+ * index. Bisecting shrinks the cost of repeated splits within a large chunk. A sequence of N
+ * splits inside an N-sized chunk does O(N log N) total work, at the price of producing a few
+ * extra intermediate chunks during descent. Smaller chunks are split exactly so we don't pay
+ * the bisection overhead when the linear walk is already cheap.
```
CraigMacomber: This description never actually mentions what this constant does. Starting with that, then putting the explanation of why in a `@remarks` block would be better.

Also its unclear if this threshold is about the top level length, or the total node count.

I suspect we will also want a similar threshold for the work on growing existing chunks instead of creating adjacent tiny ones.

I think adding another property to ChunkPolicy might make more sense here.

maybe something like `uniformChunkNodeCountDynamicTargetMax`: "When a uniformChunk is being edited, attempt to leave chunks with a total node count around this or lower. This means splitting chunks larger than this evenly (for example to avoid N*M behavior in cases like splitFieldAtIndex when making N small edits one at a time to an M sized chunk), and merging/extending chunks when doing so does not exceed this (to avoid many tiny chunks while avoiding M*N costs related to editing large chunks many times)."

Maybe not something about how `uniformChunkNodeCount` has a similar role, but for the first time something is chunked, where we can heuristically pick a larger size only risking a one time O(size) cost to split rather than N*size  like we have with `uniformChunkNodeCountDynamicTargetMax`.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +556,73 @@ export function chunkRange(
...
+		// For chunks above the bisect threshold, cut at the midpoint and let the loop descend
+		// into whichever half holds nodeIndex. The other half is left untouched.
+		const splitPoint = total > splitFieldBisectThreshold ? Math.floor(total / 2) : remaining;
+		const cursor = chunk.cursor();
+		cursor.firstNode();
+		const before = chunkRange(cursor, policy, splitPoint, false);
+		const after = chunkRange(cursor, policy, total - splitPoint, true);
+		// TODO: this could fail for really long chunks being split (due to argument count limits).
+		chunks.splice(chunkIndex, 1, ...before, ...after);
+		chunk.referenceRemoved();
```
CraigMacomber, line 637: its easy to screw up ref counting in TS, so its best to clearly document every case we remove a ref if there isn't a clearly paired add of a ref nearby. I think this is correct, but documenting why would be good and help prevent people from making future changes which cause this to be wrong and introduce a bug.

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
```diff
@@ -755,4 +759,94 @@ describe("chunkTree", () => {
 			assert(infoNonIncremental !== undefined);
 		});
 	});
+
+	describe("splitFieldAtIndex", () => {
```
CraigMacomber, line 767: I don't think you have any tests ensuring the ref counts are correct in the results. Adding some asserts to the existing tests, or dedicated tests for it seems like a good idea.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +556,73 @@ export function chunkRange(
...
+	policy: ChunkCompressor,
+): number {
+	assertNonNegativeSafeInteger(nodeIndex);
+	let remaining = nodeIndex;
+	let chunkIndex = 0;
+	while (chunkIndex < chunks.length) {
+		if (remaining === 0) {
+			return chunkIndex;
+		}
+		const chunk = chunks[chunkIndex] ?? oob();
```
CraigMacomber, line 614: I'm not sure, but I think we might be able to avoid the need for an oob case here by iterating over chunks.enumerate() instead. If that works, it would be cleaner and less error prone.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -193,9 +203,18 @@ export class ChunkedForest implements IEditableForest {
...
-				const newField = sourceField.splice(source.start, source.end - source.start);
+				assert(source.start <= source.end, "detach range start must not exceed end");
+
+				const policy: ChunkCompressor = {
+					policy: this.forest.chunker,
+					idCompressor: this.forest.idCompressor,
+				};
+				// Split start first: splitting end later only expands chunks at positions >= startChunkIndex,
+				// which leaves startChunkIndex valid. The reverse order would shift endChunkIndex when
+				// source.start and source.end land in different chunks.
```
CraigMacomber, line 214: When we know something could be faster (but its not worth the complexity to optimize it for now), its nice to leave a note for people in the future wondering if they can optimize it.
```suggestion
				// source.start and source.end land in different chunks.
				// Performance: It's practical to have a variant of splitFieldAtIndex which can split at multiple locations in a single pass if the performance of this becomes worth optimizing.
```

Alternatively you could make splitFieldAtIndex be splitFieldAtIndices and take in an array or indexes (which must be in increasing order), and return an array of chunk indexes. Annoyingly this incurs two array allocations (for input and output), but I suppose thats very cheap compared to the rest of what splitFieldAtIndex does.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
brrichards: I agree with using enumerate to make this less error prone, but I had something similar to that in a previous commit and was experiencing some serious problems with performance when we inserted a reasonably sized text blob. Since every insert right now is going to be a single UniformChunk that doesnt get merged into adjacent UniformChunks, if we insert N textNodes then enumerate will have to walk all previous chunks for the target position. Which ends up being N inserts across N newly inserted textNodes which becomes O(N^2). I believe that problem will go away if we allow remerge UniformChunks (This would be huge for memory and performance). 

I could leave a comment mentioning why this is done this way, or a TODO for changing this

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
brrichards: I'll go ahead and update the `ChunkPolicy` with these changes. I imagine this is going to require some `ChunkPolicy` changes in other test files with the new property being added. Do we want to make this new property optional? I don't know if we have any consumer reaching into our internals for a custom `ChunkPolicy` that this change would break.

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
brrichards: You're right I'll update the tests

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
brrichards: Updated to include a document

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
CraigMacomber: I missed that this sometimes iterates multiple times at the same index, making Array.entries (which Is what I meant) not actually work well here.

Now that I read this more carefully, I see its rather clever, but a bit hard to read. I'll make a few minor suggestions in other comments about how to make it more readable.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +579,66 @@ export function chunkRange(
...
+		const total = chunk.topLevelLength;
+		if (remaining >= total) {
+			remaining -= total;
```
CraigMacomber:
```suggestion
			// nodeIndex is not in this chunk, so move forward one chunk and continue.
			remaining -= total;
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +579,66 @@ export function chunkRange(
...
+			chunkIndex++;
+			continue;
+		}
+		// For chunks above the bisect threshold, cut at the midpoint and let the loop descend
```
CraigMacomber:
```suggestion

		// nodeIndex falls within this chunk, so split the chunk.
		// This does not move the chunkIndex forward: the next iteration might need to split again at the same index.
		//
		// For chunks above the bisect threshold, cut at the midpoint and let the loop descend
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +579,66 @@ export function chunkRange(
...
+		const before = chunkRange(cursor, policy, splitPoint, false);
+		const after = chunkRange(cursor, policy, total - splitPoint, true);
+		// TODO: this could fail for really long chunks being split (due to argument count limits).
+		chunks.splice(chunkIndex, 1, ...before, ...after);
+		// The spliced-out slot held a ref to the original chunk. The two new chunks come with
+		// their own refs from chunkRange, so the slot's ref to the original needs to be released.
+		chunk.referenceRemoved();
+		if (splitPoint === remaining) {
+			return chunkIndex + before.length;
+		}
```
CraigMacomber: If you deleted this code, it would still work correct? This is just an optimization that saves going through the loop again, typically once, and hitting the return above correct?

I suspect this doesn't save significant time, and might be better to remove, which would simplify this function and make it easier to read and maintain.

If this is necessary or worthwhile, put a comment on it explaining why.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +579,66 @@ export function chunkRange(
...
+ *
+ * @param chunks - The array of {@link TreeChunk}s for the field to split. Mutated in place.
+ * @param nodeIndex - The index of the node to split at. Must be in `[0, totalNodes]`, where
+ * `totalNodes` is the sum of {@link TreeChunk.topLevelLength} across all chunks. `nodeIndex === totalNodes`
+ * is allowed as a splice point (e.g. attaching at the end of a chunk array).
+ * @param policy - The {@link ChunkCompressor} to use when splitting chunks and re-chunking each side
+ * of the split via {@link chunkRange}.
+ *
+ * @returns The index of the chunk in `chunks` that begins at `nodeIndex`, or `chunks.length` when
+ * `nodeIndex === totalNodes`.
```
CraigMacomber: I think this is a good design, but I think we can tweak the phrasing to make this one case instead of two (which also helps motivate why the API works like this)

```suggestion
 * @returns The index in `chunks` (after medications made by this function) where if a chunk were inserted at that index its first top level node would have index `nodeIndex` when treating `chunks` as a field.
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +579,66 @@ export function chunkRange(
 
 	return output;
 }
+
+/**
+ * Walks the `chunks` array of a field and splits a chunk if needed so that the node at the given
+ * index sits on a chunk boundary. After the call, the node at `nodeIndex` is the first node of the
+ * chunk at the returned index.
```
CraigMacomber: This description doesn't cover than case where nodeIndex is after the last node. Phrasing like I used inhttps://github.com/microsoft/FluidFramework/pull/27153/files#r3238211912 might be better.

Same for the nodeIndex parameter comment.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -556,6 +579,66 @@ export function chunkRange(
...
+ *
+ * @remarks
+ * Chunks whose top level length exceeds {@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax}
+ * are bisected at the midpoint and the loop descends into whichever half contains `nodeIndex`.
```
CraigMacomber: the fact this this bisection is iterative in a loop and not recursive is an implementation detail: such things should only be documented inside the function as comments, not in this doc comment which is for callers.

This is what I think is important to callers:

```suggestion
 * When splitting chunks, large chunks are split evenly so that repeated calls to this method (or similar operations)
 * avoid poor worst-case behavior. See {@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax} for details.
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
brrichards: Yeah this will still work without this if. It doesn't save significant time (5% quicker with small edits to large docs) so I'll go ahead and remove it.

CraigMacomber (approved):

PR #27155, opened 2026-04-24
Add script to detect misconfigurations in local environment

alexvy86:
## Description

Adds a script that can be invoked with `pnpm check:dev-env` from the root of the repo, to detect misconfigurations which can affect developer experience and have side effects that are hard-to-troubleshoot.

Currently it checks for:
- Non-default npm registry (either in `.npmrc` or through `NPM_CONFIG_REGISTRY`), including for specific scopes
- No proxy-related variables set in npm configuration

Looks like this:

image

## Reviewer Guidance


package.json
```diff
@@ -41,6 +41,7 @@
 		"bundle-analysis:run": "flub run bundleStats --dangerfile build-tools/packages/build-cli/lib/library/dangerfile.cjs",
 		"changeset": "flub changeset add --releaseGroup client",
 		"check:are-the-types-wrong": "fluid-build --task check:are-the-types-wrong",
+		"check:dev-env": "node scripts/check-dev-env.mjs",
 		"check:assertTags": "flub generate assertTags --all --validate",
 		"check:changesets": "vale .changeset --glob=*-*-*.md",
```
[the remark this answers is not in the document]
alexvy86: Fixed in latest

scripts/check-dev-env.mjs
```diff
@@ -0,0 +1,188 @@
...
+		warn(
+			"proxy environment variables",
+			`Proxy settings detected in environment:\n` +
+				found.map((entry) => `        ${entry}`).join("\n") +
+				`\n        These may redirect or block package resolution. If you are not on a network\n` +
+				`        that requires a proxy, consider unsetting them in your shell profile.`,
+		);
+	} else {
+		pass("no proxy environment variables set");
+	}
```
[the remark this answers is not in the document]
alexvy86: Fixed in latest

scripts/check-dev-env.mjs
```diff
@@ -0,0 +1,188 @@
...
+		.filter((k) => proxyVarNames.includes(k.toLowerCase()))
+		.map((k) => `'${k}' = '${process.env[k]}'`);
+
+	if (found.length > 0) {
+		warn(
+			"proxy environment variables",
+			`Proxy settings detected in environment:\n` +
+				found.map((entry) => `        ${entry}`).join("\n") +
+				`\n        These may redirect or block package resolution. If you are not on a network\n` +
+				`        that requires a proxy, consider unsetting them in your shell profile.`,
```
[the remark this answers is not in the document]
alexvy86: Fixed in latest.

scripts/check-dev-env.mjs
```diff
@@ -0,0 +1,230 @@
...
+					`        that requires a proxy, remove them from your user-level .npmrc (~/.npmrc).`,
+			);
+		} else {
+			pass("no proxy settings in npm config");
+		}
+	}
+}
+
+// ---------------------------------------------------------------------------
+// Run checks
```
Josmithr, line 196: Nit: we should replace these label comments with `// #region` / `// #endregion` comment blocks.

scripts/check-dev-env.mjs
```diff
@@ -0,0 +1,230 @@
+/*!
```
Josmithr, line 1: Nit: any reason we can't port this to TypeScript and execute it using `jiti`? We do this elsewhere.

scripts/check-dev-env.mjs
```diff
@@ -0,0 +1,230 @@
...
+ */
+
+/**
+ * Checks the local developer environment for common configuration issues.
+ * Run with: node scripts/check-dev-env.mjs
+ * Or via: pnpm check:dev-env
+ */
+
+import { execSync } from "node:child_process";
+
```
tylerbutler, line 13: Any reason this can't be a flub command with tests?

scripts/check-dev-env.mjs
TommyBrosman: +1 for jiti.

PR #27162, opened 2026-04-24
[main] Component Governance update packages

kian-thompson:
`lodash`:
CVE-2026-2950
CVE-2026-4800

`path-to-regexp`:
CVE-2026-4867
CVE-2026-4923


common/lib/common-utils/pnpm-lock.yaml
steffenloesch, line 1: `path-to-regexep` shows in here still as 8.2.0, which is vulnerable and needs to be at least 8.4.0. Same for server/gitrest/pnpm-lock.yaml, server/historian/pnpm-lock.yaml, server/routerlicious/pnpm-lock.yaml. It's being pulled in by `nise@6.1.1`. I think we need to address this before completing the PR.

steffenloesch (changes requested):

steffenloesch (approved):
Thanks for the changes. Looks good to me now.

PR #27165, opened 2026-04-24
ci: replace yes-reply with "start review" checkbox in PR review confirm flow

tylerbutler:
## Description

The PR review confirm flow (`pr-review-confirm.yml`) previously listened to every issue comment on a PR, classifying the reply as affirm / question / other and spinning up Copilot CLI to answer questions. That meant the workflow ran on every comment reply, even unrelated ones.

This PR collapses the trigger to a single UI affordance: a **Start review** checkbox at the bottom of the auto-route proposal comment. The fleet dispatches when that box transitions from unchecked to checked on the bot's proposal comment. The Q&A branch is removed entirely — this flow is code-review only.

Changes:

- **`.github/scripts/pr_review_propose.py`** — `build-comment` renders `- [ ] **Start review**` at the bottom. Removed `build-qa-context` and `render-qa-prompt`. The `START_LABEL` constant is now the single source of truth for the checkbox label (the workflow matches it as a literal substring).
- **`.github/workflows/pr-review-confirm.yml`** — trigger is `issue_comment: [edited]` only. The job-level `if:` gates everything upfront: PR comment, human sender, bot-authored proposal comment, and the `Start review` box transitioning from unchecked to checked (`contains(body, '- [x] **Start review**') && !contains(changes.body.from, '- [x] **Start review**')`). Editor's write access is verified by [`actions-cool/check-user-permission@v2.3.0`](https://github.com/actions-cool/check-user-permission) (author_association on an edit is the bot's, so the built-in gate isn't usable).
- **`.github/prompts/pr-review-qa.md`** — deleted (Q&A flow is gone).

## Reviewer Guidance

- The job `if:` block matches the `Start review` checkbox line as a literal substring. That literal is coupled to `START_LABEL` in `pr_review_propose.py` — flagged in comments on both sides. If that coupling is too fragile, an alternative is to adopt [`wadackel/checkbox-workflow-action`](https://github.com/wadackel/checkbox-workflow-action), but that would mean replacing `marocchino/sticky-pull-request-comment` + our `build-comment` helper in `pr-review-auto-route.yml` with its Configuration Mode (the action owns its own comment + state markers).
- The transition detection relies on `github.event.changes.body.from` being populated on `edited` events. Worth a sanity check on the rare case of a checkbox toggle via the API vs. the UI.

Smoke-tested the Python script locally — `build-comment`, `parse-checkboxes`, and `format-names` all return expected values.


.github/workflows/pr-review-confirm.yml
```diff
@@ -21,42 +21,81 @@ permissions:
...
+          # meaningful. Check the acting user's repo permission directly.
+          PERMISSION=$(gh api "repos/${REPO}/collaborators/${EDITOR}/permission" \
+            --jq '.permission' 2>/dev/null || echo "none")
+          echo "Editor ${EDITOR} has ${PERMISSION} permission on ${REPO}"
+          case "$PERMISSION" in
+            admin|maintain|write)
+              echo "authorized=true" >> "$GITHUB_OUTPUT" ;;
+            *)
+              echo "authorized=false" >> "$GITHUB_OUTPUT" ;;
+          esac
```
tylerbutler: research actions we can reuse for this.

.github/workflows/pr-review-confirm.yml
```diff
@@ -21,42 +21,81 @@ permissions:
...
+          printf '%s' "$CURRENT_BODY"  > /tmp/current.md
+          printf '%s' "$PREVIOUS_BODY" > /tmp/previous.md
+          CURRENT=$(python3 .github/scripts/pr_review_propose.py parse-start-flag /tmp/current.md)
+          PREVIOUS=$(python3 .github/scripts/pr_review_propose.py parse-start-flag /tmp/previous.md)
+          echo "previous=${PREVIOUS} current=${CURRENT}"
+          if [ "$PREVIOUS" = "false" ] && [ "$CURRENT" = "true" ]; then
+            echo "fire=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "fire=false" >> "$GITHUB_OUTPUT"
+          fi
```
tylerbutler: surely there are reusable actions for this too.

.github/workflows/pr-review-confirm.yml
tylerbutler: Swapped to [`actions-cool/check-user-permission@v2.3.0`](https://github.com/actions-cool/check-user-permission) in a27523b — pinned by SHA per the ratchet convention. `require: write` accepts write/maintain/admin. Downstream `if:` gates now check `steps.authz.outputs.require-result == 'true'`.

Considered but skipped:
- [scherermichael-oss/action-has-permission](https://github.com/marketplace/actions/has-permission) — unmaintained for years.
- [lannonbr/repo-permission-check-action](https://github.com/marketplace/actions/repo-permission-check-action) — works but less expressive (no `require:` threshold, have to interpret the output yourself).

.github/workflows/pr-review-confirm.yml
tylerbutler: Researched this — there are reusable actions, but they come with a bigger rework than I think is warranted. Flagging the tradeoff before doing it:

**[`wadackel/checkbox-workflow-action`](https://github.com/wadackel/checkbox-workflow-action)** (latest: `v1.1.0`, Sept 2025) is the closest fit. It has two modes:
- **Configuration Mode** creates/updates the comment with embedded `<!-- checkbox-workflow-action:managed -->` state metadata.
- **Detection Mode** reads that metadata to detect transitions and outputs `changed`, `state`, and `changes` (list of keys whose state changed).

The catch: Detection Mode requires the comment to have been created by its own Configuration Mode (state is stored in embedded markers that it owns). So to adopt it we'd have to:
1. Replace `marocchino/sticky-pull-request-comment` + `pr_review_propose.py build-comment` in `pr-review-auto-route.yml` with the action's Configuration Mode.
2. Restructure our reviewer list (label + em-dash + description) into the action's JSON5 `config` schema, wrapping our prose around it via `{{body}}` in `message`.
3. Drop `parse-checkboxes` / `parse-start-flag` (the action exposes state as JSON).

That's a decent diff across two workflows and the Python helper for a net-neutral payoff — the current transition logic is ~10 lines of shell plus a 7-line Python function.

Alternatives I also looked at:
- [`marocchino/checkbox-action`](https://github.com/marketplace/actions/checkbox-action) — similar shape, same coupling.
- Plain `contains(github.event.comment.body, '- [x] **Start review**')` combined with `!contains(github.event.changes.body.from, '- [x] **Start review**')` — no action needed at all, just YAML expressions. **Would this be preferable?** It'd let me drop `parse-start-flag` entirely.

Happy to do whichever you prefer: (a) leave as-is, (b) collapse to `contains()` expressions, or (c) adopt `checkbox-workflow-action` and restructure the auto-route comment authoring.

tylerbutler: Went with option (b) in 40f035a — moved the transition check into the job-level `if:` as two `contains()` expressions and deleted `parse-start-flag`/`_START_RE`. The YAML substring is coupled to `START_LABEL` in `pr_review_propose.py`; flagged that with a comment on the constant and on the `if:` block.

.github/scripts/pr_review_propose.py
```diff
@@ -80,7 +80,9 @@ def cmd_build_comment(args: argparse.Namespace) -> None:
         lines.append(f"- [{check}] **{r.label}** — {r.description}")
     lines += [
         "",
-        "Toggle checkboxes to adjust, then reply **yes** to start — or ask me anything!",
+        "Toggle the reviewer checkboxes above to adjust, then tick the box below to start:",
+        "",
+        f"- [ ] **{START_LABEL}**",
```
noencke, line 85: Is the bot able to _edit_ its comment? It might make sense to:
1. User checks the start review box
2. Bot kicks off the review
3. Bot replaces the check box with a Check emoji (no longer clickable) and it says "review started" or something.

Not worth doing if modifying the comment is difficult. The nice thing about this is that it removes the weird state where a user clicks the checkbox, then unchecks it and leaves it (but a review will still be happening). On the whole though, not a big deal at all.

noencke (approved):

.github/scripts/pr_review_propose.py
tylerbutler: Yeah, good point. Modifying the comment isn't that hard, it's just brittle. I'll leave that for a future iteration.

PR #27166, opened 2026-04-24
Benchmark API improvements

CraigMacomber:
## Description

Better centralize the mocha specific code, and make timeout, skips and the mode a bit easier to work with.

## Reviewer Guidance

### DIFF
--- a/tools/benchmark/src/mocha/benchmarkIt.ts
+++ b/tools/benchmark/src/mocha/benchmarkIt.ts
@@ -4,7 +4,13 @@
  */
 
 import type { Test } from "mocha";
-import { isChildProcess, qualifiedTitle, type BenchmarkOptions } from "../Configuration.js";
+import {
+	BenchmarkMode,
+	currentBenchmarkMode,
+	isChildProcess,
+	qualifiedTitle,
+	type BenchmarkOptions,
+} from "../Configuration.js";
 import { timer } from "../timer.js";
 import { supportParentProcess } from "./runner.js";
 import { emitResultsMocha } from "./runnerUtilities.js";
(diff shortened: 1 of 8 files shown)


tools/benchmark/src/mocha/benchmarkIt.ts
```diff
@@ -14,6 +20,48 @@ import { emitResultsMocha } from "./runnerUtilities.js";
...
+ * To skip or restrict to `.only`, use the `skip` and `only` options on this interface —
+ * Mocha does not support calling those methods on an already-registered test.
+ * @public
+ * @input
+ */
+export interface MochaBenchmarkOptions extends BenchmarkOptions {
+	/**
+	 * When true, `mocha`-provided functions use their `.only` counterparts to restrict the run to this test.
+	 */
+	readonly only?: boolean;
```
alexvy86, line 39: Would it be feasible to have `benchmarkIt` expose functions `.only()`, `.skip()`, `.skipOnCorrectness()`, `.skipOnPerformance()` or something along those lines? So muscle-memory on how to do those things works. I thought of the extra skip ones to keep the arguments the same, but make them easily discoverable with IntelliSense.

CraigMacomber: Personally I rather dislike how mocha deals with per test config. You can .skip on `it`, and on the returned test from it and they do very different things. You have to do .only and skip at before the test, but timeout is after on the Test object. The whole skip and only thing require a function with methods on it, which is a mess in ES6 and in typescript. Not to mention the abuse of globals.

Also, I don't really think that would make them more discoverable. Having the options be part of the options object you have to provide to benchmarkIt makes the easy to discover and autocomplete when setting other options, like the required title. To discover functions on the method like mocha does it, you have to invoke intelisense to look for methods on the function, which is a very uncommon pattern not wel supported by typeScript (can only be implemented in non TypeSafe ways) or ES6 (you can't do it with an ES6 class unless you subclass function, which basically forces eval of the function body.). And if we did that pattern, would we want to do it for the correctnessTimeoutMs? We can't return a mocha Test object and have that work like it does for mocha, so we would need some non-mocha like way to do that, which would then be inconsistent with the other APIs.

Also the mocha style approach is really annoying to write wrappers for, something people occasionally do for our benchmark APIs (I don't think we should be doing that much anymore, but I still dislike the pattern as things that are really hard to wrap/indirect are just annoying). Its also really hard for users to make conditional (like only skip for specific end to end test configs, or only set a timeout sometimes: the new timeout APIs is being added mostly since mocha's approach of using function calls and not optional parameters was so messy for timeouts I wanted a new non-mocha style API)

While I see a benefit to matching mocha closely for consistency with mocha, I think its a bad pattern, its mess to implement, we can't do it for everything, and I don't want to support multiple ways to do these things so for now I'm favoring not supporting the mocha style .only and .skip approach. ALl that said, I still almost did it since consistency with mocha really seems like a good idea, but my annoyance with and dislike of the whole pattern won out.

tools/benchmark/src/test/mocha/integration.spec.ts
```diff
@@ -143,6 +147,35 @@ describe("mocha integration", () => {
 			},
 		}).timeout(1);
 	}
+
+	it("only option restricts run to only that test", () => {
+		// Run a child mocha process with --integrationOnly, which registers:
+		// 1. benchmarkIt({ only: true, title: "only-integration-inner" }) — should run
+		// 2. benchmarkIt({ title: "only-integration-excluded" }) — should be skipped
+		// Mocha's exclusive mode (triggered by only: true) skips all non-only tests.
+		const { stdout } = integrationTest(["--integrationOnly"], false);
```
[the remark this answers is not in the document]
CraigMacomber: Non issue due to .only

alexvy86 (approved):

PR #27169, opened 2026-04-24
fix(build-cli): avoid full repo package scan in generate buildVersion

alexvy86:
## Description

`flub generate buildVersion` was calling `getContext()` to get access to a git client for reading tags. `getContext()` constructs a full `FluidRepo` object, which eagerly calls `readdirSync` on every release group directory defined in `fluidBuild.config.cjs` (including `build-tools/`, `server/routerlicious/`, etc.). This makes it impossible to run that flub command in sparse git checkouts, where any missing release group directory causes an immediate `ENOENT` crash, even though the command has no need for any of those packages.

The fix replaces the `getContext()` + `context.getGitRepository()` chain with a direct `Repository` instantiation using `getResolvedFluidRoot()`. This resolves the repo root by walking up the directory tree (no package scanning), and `Repository` wraps `simple-git` directly. The command's behavior is unchanged for full checkouts. This pattern is already used in `commands/check/policy.ts`.

## Reviewer Guidance

The change is small (4 lines added, 2 removed) and the logic is identical — only the path to a `SimpleGit` client changes. The existing `--tags` test flag continues to let the git call be bypassed entirely in unit tests.

### DIFF
--- a/build-tools/packages/build-cli/docs/vnext.md
+++ b/build-tools/packages/build-cli/docs/vnext.md
@@ -4,6 +4,7 @@
 Vnext commands are new implementations of standard flub commands using new infrastructure.
 
 * [`flub vnext check latestVersions`](#flub-vnext-check-latestversions)
+* [`flub vnext generate buildVersion`](#flub-vnext-generate-buildversion)
 * [`flub vnext generate changelog`](#flub-vnext-generate-changelog)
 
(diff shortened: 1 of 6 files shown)


tylerbutler:
I'd love it if we could make this a next command that uses the new command infra which has a different context strategy. At the very least we need to make sure the approach works for both command types.

PR #27175, opened 2026-04-27
build: add release notes and regenerate changelogs for 2.100.0

tylerbutler:
## Description

Release notes and per-package changelog generation for the 2.100.0 release.

- Adds `RELEASE_NOTES/2.100.0.md` (generated via `flub generate releaseNotes -g client -t minor`)
- Regenerates 166 `CHANGELOG.md` files across the client release group (via `flub generate changelog -g client`)
- Consumes 6 changesets from `.changeset/`

This PR is part of the 2.100.0 release prep series. **It must merge before the version-bump PR.**

## Reviewer Guidance

These files are generated artifacts. If wording changes are needed, edit the changesets in a separate PR (which must be merged first), then regenerate.


experimental/PropertyDDS/packages/property-common/platform-dependent/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # @fluid-internal/platform-dependent
 
+## 2.100.0
+
+Dependency updates only.
```
alexvy86: Probably we just missed listing it in the changeset, but this one should also have the note about Node22, no?

packages/tools/replay-tool/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # @fluid-internal/replay-tool
 
+## 2.100.0
+
+Dependency updates only.
```
alexvy86: I think I'd also add the note here.

RELEASE_NOTES/2.100.0.md
```diff
@@ -0,0 +1,257 @@
...
+- @fluidframework/tinylicious-client
+- @fluidframework/tinylicious-driver
+- @fluidframework/tool-utils
+- @fluidframework/tree
+- @fluidframework/tree-agent
+- @fluidframework/tree-agent-langchain
+- @fluidframework/tree-agent-ses
+- @fluidframework/type-factory
+- @fluidframework/undo-redo
+- fluid-framework
```
alexvy86, line 122: If we make the two changes above, probably want to add the packages here too.

packages/test/test-utils/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # @fluidframework/test-utils
 
+## 2.100.0
+
+Dependency updates only.
```
alexvy86: Probably here too because it's a `@fluidframework` (not `@fluid-internal` or `@fluid-private`).

alexvy86 (commented):
Left a couple of comments; nothing major. A couple of other packages (e.g. the ones under `packages/test/`) didn't get the Node22 note... but I can't convince myself that they absolutely need it. Many of them are `@fluid-private` and those I think are fine without it. The `@fluid-internal` ones... are not part of what we advertise as API surface for consumers, so I could see an argument they don't need it either (although we put it in a few of such packages outside `packages/test/`, which I don't mind). There's `@fluidframework/test-utils` though, which based on the scope I guess is technically part of our API surface so that one probably wants the note?

tylerbutler:
Force-pushed regenerated release notes and changelogs after #27178 merged.

Addresses the 4 review comments:
- `experimental/PropertyDDS/packages/property-common/platform-dependent/CHANGELOG.md` — `@fluid-internal/platform-dependent` now has the Node 22 entry.
- `packages/tools/replay-tool/CHANGELOG.md` — `@fluid-internal/replay-tool` now has the Node 22 entry.
- `packages/test/test-utils/CHANGELOG.md` — `@fluidframework/test-utils` now has the Node 22 entry.
- `RELEASE_NOTES/2.100.0.md` — affected-packages list expanded from 93 to 96 to include the three above.

Generated via the standard `flub generate releaseNotes -g client -t minor` + `flub generate changelog -g client` flow on top of the corrected changeset.

alexvy86 (approved):

PR #27181, opened 2026-04-27
fix(tree): batch uniform-shaped nodes into multi-node UniformChunks

brrichards:
## Summary 

This is the final PR to enable multi-node chunks in chunked Forest. This PR should not be merged until after [support attach/detach edits on multi-node chunks](https://github.com/microsoft/FluidFramework/pull/27153), [track chunkIndex on field entry so getField finds parent](https://github.com/microsoft/FluidFramework/pull/27084)

`chunkRange` was evaluating `maxTopLevelLength` as `ceil(nodesPerTopLevelNode / uniformChunkNodeCount)`, which is the inverse of the intended ratio. For a leaf shape with `nodesPerTopLevel = 1` (eg. plain text) with `uniformChunkNodeCount = 4`, this produced `1`, so each `UniformChunk` held a single node and batching never happened. 

The fix flips the division and clamps to `1`:

<pre> const maxTopLevelLength = Math.max(1, chunkCompressor.policy.uniformChunkNodeCount / nodesPerTopLevelNode);  </pre>
A 10-element number field with `uniformChunkNodeCount = 4` now produces chunks of length `[4, 4, 2]` instead of ten singletons. 

#### Tests
| File | Changes |
| ---- | ---- |
| chunkTree.spec.ts | Adds `uniformChunks` describe block. "batches uniform shaped nodes into chunks of uniformChunkNodeCount" - direct `chunkRange` over a 10-node field, asserts [4, 4, 2]. Uses a custom chunk policy. Added a single regression test for calling splitFieldAtIndex after chunkRange with a custom chunkPolicy that tests batching is done correctly, and that the bisection splits the uniformChunk with the intended behavior|

#### Notes
- ForestSummarizer has a single test that assumes each node gets it own chunk. Since this is no longer the case the test has been updated to assert 1 uniform chunk instead of 2
- Any operation that requires `enterNode` to be called on a uniformChunk with `topLevelLength = N`will cause the chunk to explode into N basicChunks. There is a TODO to not require this, but for now this is the intended behavior. 
## Reviewer Guidance                                                                               
                      

packages/dds/tree/src/test/feature-libraries/forest-summary/forestSummarizer.spec.ts
```diff
@@ -452,7 +452,7 @@ describe("ForestSummarizer", () => {
 					{
 						fooArray: new FooArray(["value1", "value2"]),
 					},
-					2 /* incrementalNodeCount */,
+					1 /* incrementalNodeCount */,
```
CraigMacomber, line 455: I'm not really clear on what this test is doing, or why a change is expected here (and its hard to tell from just looking at the review.

What does this change mean, and why is it expected?

brrichards: From what I was seeing this was an assertion that was to make sure that there were exactly the amount incrementalNodeCount as chunks. It just so happened that there were two because of the bug and that's why this was here. After we allow multi-node uniform chunks those two array nodes get correctly chunked into a single uniformChunk. So, it gets moved from 2 uniformChunks to 1 which will only have 1 incremental node for the single chunk. Since the other data types in the test don't chunk into uniformChunks they don't need new assertions.

CraigMacomber: Oh, that makes sense, and suggest this change improved the efficiency of incremental reuse here. Neat.

CraigMacomber (approved):

PR #27182, opened 2026-04-27
[fluid-runner] Expose createFluidRunnerContainerAndExecute and createFluidRunnerLogger

kian-thompson:
Our partner team requested some additional methods from the `fluid-runner` package.

## Api Update
For back-compatibility, the previous `@internal` names are retained as `@deprecated` aliases:

- `createContainerAndExecute` → use `createFluidRunnerContainerAndExecute`
- `createLogger` → use `createFluidRunnerLogger`
- `ITelemetryOptions` → use `IFileLoggerTelemetryOptions`


packages/tools/fluid-runner/src/logger/fileLogger.ts
```diff
@@ -27,7 +28,8 @@ export enum OutputFormat {
 
 /**
  * Options to provide upon creation of IFileLogger
- * @internal
+ * @legacy
+ * @beta
  */
 export interface ITelemetryOptions {
```
markfields: Rename to `IFileLoggerTelemetryOptions`?

packages/tools/fluid-runner/src/exportFile.ts
```diff
@@ -107,18 +113,38 @@ export async function exportFile(
...
+export async function createFluidRunnerContainerAndExecute(
 	localOdspSnapshot: string | Uint8Array,
 	fluidFileConverter: IFluidFileConverter,
-	logger: ITelemetryLoggerExt,
+	logger: ITelemetryBaseLogger,
 	options?: string,
 	timeout?: number,
 	disableNetworkFetch: boolean = false,
 ): Promise<string> {
+	const extLogger = createChildLogger({ logger });
```
markfields: nit: I'd call them `baseLogger` and `logger` 🤷

markfields (approved):

markfields (approved):

jason-ha (commented):
Make sure to build a FF test/ branch build and confirm integration with Pages.
They are using /internal imports and mocking right now. While that isn't "supported" it will be a problem for us if those integration breaks.

jason-ha:
Since this didn't make 2.103, please wait to build new test build until after #27476 merges in 2.110 and merged to this branch.

kian-thompson:
> Make sure to build a FF test/ branch build and confirm integration with Pages. They are using /internal imports and mocking right now. While that isn't "supported" it will be a problem for us if those integration breaks.

Test branch passes integration pipeline with latest state: https://dev.azure.com/office/OC/_build/results?buildId=50111707&view=results

packages/tools/fluid-runner/src/logger/loggerUtils.ts
```diff
@@ -5,15 +5,49 @@
...
 	createChildLogger,
 } from "@fluidframework/telemetry-utils/internal";
 
 import { CSVFileLogger } from "./csvFileLogger.js";
-import { type IFileLogger, type ITelemetryOptions, OutputFormat } from "./fileLogger.js";
+import {
+	type IFileLogger,
+	type IFileLoggerTelemetryOptions,
+	// eslint-disable-next-line import-x/no-deprecated
+	type ITelemetryOptions,
```
jason-ha: Why use `ITelemetryOptions` at all? It is just a deprecated alias for `IFileLoggerTelemetryOptions` which should be usable in all of the same places.

packages/tools/fluid-runner/src/exportFile.ts
```diff
@@ -11,17 +11,23 @@ import {
...
 
 import type { IFluidFileConverter } from "./codeLoaderBundle.js";
 import { FakeUrlResolver } from "./fakeUrlResolver.js";
 /* eslint-disable import-x/no-internal-modules */
-import type { ITelemetryOptions } from "./logger/fileLogger.js";
-import { createLogger, getTelemetryFileValidationError } from "./logger/loggerUtils.js";
+import type { IFileLoggerTelemetryOptions } from "./logger/fileLogger.js";
+import {
+	// eslint-disable-next-line import-x/no-deprecated
+	createLogger,
```
jason-ha: This can't use `createFluidRunnerLogger` instead, because that only provides `ITelemetryBaseLogger` for `logger` and `PerformanceEvent.timedExecAsync` expects `TelemetryLoggerExt | ITelemetryLoggerExt`, right?
So, what does this look like in the end?
Are we just stripping the deprecated from `createLogger` and keeping use here?
All lint disables should be commented. That would help clarify why use of deprecated thing is okay now.

packages/tools/fluid-runner/src/parseBundleAndExportFile.ts
```diff
@@ -76,6 +85,7 @@ export async function parseBundleAndExportFile(
 
 				fs.writeFileSync(
 					outputFile,
+					// eslint-disable-next-line import-x/no-deprecated
 					await createContainerAndExecute(
```
jason-ha: Why shouldn't `createFluidRunnerContainerAndExecute` be used here? Because it creates another child logger?

jason-ha (commented):
Deprecated uses need comment. (All list suppressions should have comments.)

packages/tools/fluid-runner/src/parseBundleAndExportFile.ts
```diff
@@ -76,10 +86,10 @@ export async function parseBundleAndExportFile(
 
 				fs.writeFileSync(
 					outputFile,
-					await createContainerAndExecute(
+					await createFluidRunnerContainerAndExecute(
 						getSnapshotFileContent(inputFile),
 						fluidExport,
-						logger,
+						baseLogger,
```
jason-ha: Can't stick with `logger` here? If not, probably should comment.

jason-ha (approved):
one question / suggestion

.changeset/expose-fluid-runner-helpers.md
```diff
@@ -0,0 +1,16 @@
+---
+"@fluidframework/fluid-runner": minor
+"__section": other
+---
+Expose `createFluidRunnerContainerAndExecute` and `createFluidRunnerLogger` as `@legacy @beta`
```
Josmithr: Our guidelines call to avoid using `inlineCode` syntax in changeset titles. Everywhere else is fine, but we like to avoid them in titles to avoid making the final changelog look too noisy.
```suggestion
Expose createFluidRunnerContainerAndExecute and createFluidRunnerLogger as legacy/beta APIs
```

.changeset/expose-fluid-runner-helpers.md
```diff
@@ -0,0 +1,16 @@
...
+
+The lower-level helpers `createFluidRunnerContainerAndExecute` and `createFluidRunnerLogger` from `@fluidframework/fluid-runner` are now part of the legacy/beta public API surface. Use `createFluidRunnerContainerAndExecute` to load a container from an ODSP snapshot and run caller-provided code against it. Use `createFluidRunnerLogger` to obtain a file-backed telemetry logger that can be passed to `createFluidRunnerContainerAndExecute`.
+
+The `IFileLogger` and `IFileLoggerTelemetryOptions` types — already exported from the package — have likewise been promoted from `@internal` to `@legacy @beta` so they can be referenced by consumers of these APIs. The signatures of `createFluidRunnerLogger` and `createFluidRunnerContainerAndExecute` use the public `ITelemetryBaseLogger` type from `@fluidframework/core-interfaces`.
```
Josmithr: Nit: these changesets are intended to be end-user facing. From their perspective, APIs that we have tagged `@internal` don't exist. So I would recommend re-writing these notes in terms of new APIs being added, rather than existing APIs being promoted.

.changeset/expose-fluid-runner-helpers.md
```diff
@@ -0,0 +1,16 @@
...
+
+For back-compatibility, the previous `@internal` names are retained as `@deprecated` aliases:
+
+- `createContainerAndExecute` → use `createFluidRunnerContainerAndExecute`
+- `createLogger` → use `createFluidRunnerLogger`
+- `ITelemetryOptions` → use `IFileLoggerTelemetryOptions`
```
Josmithr: If these notes are strictly about `@internal` APIs, then I think we can omit them, as they don't impact end-users. These might be good details to call out in the PR description, but probably don't belong in the changeset.

packages/tools/fluid-runner/src/logger/fileLogger.ts
```diff
@@ -47,3 +49,10 @@ export interface ITelemetryOptions {
 	/** Number of telemetry events per flush to telemetry file */
 	eventsPerFlush?: number;
 }
+
+/**
```
Josmithr, line 53: Nit (feel free to close without action): Since this is internal only, can't we just update all of the existing references to use the new type names? We don't generally try to avoid breaking changes for internal-only types. But this could also be done in a separate PR.

packages/tools/fluid-runner/src/logger/loggerUtils.ts
```diff
@@ -5,15 +5,47 @@
...
+ * @remarks
+ * All telemetry events should be sent through the returned `logger`. The returned `fileLogger` is the
+ * underlying sink — its `close()` method must be called at the end of execution to flush any buffered
+ * events to disk.
+ *
+ * If `options.outputFormat` is not supplied, telemetry is written as JSON. Use {@link OutputFormat.CSV}
+ * to write CSV instead. See {@link IFileLoggerTelemetryOptions} for supported options including default
+ * properties applied to every event and flush batching.
+ *
+ * @param filePath - Path to the file telemetry will be written to. Must not already exist.
```
Josmithr: Nit: what happens if it does exist? Do we throw an error? Would be good to be explicit about that.

packages/tools/fluid-runner/src/exportFile.ts
```diff
@@ -106,9 +116,53 @@ export async function exportFile(
...
+ *
+ * @remarks
+ * The container is loaded with `opsBeforeReturn: "cached"` and {@link @fluidframework/container-loader#waitContainerToCatchUp}
+ * is invoked before {@link IFluidFileConverter.execute} runs. The container is disposed once `execute` resolves
+ * (or rejects).
+ *
+ * @param localOdspSnapshot - The ODSP snapshot to load the container from. May be either the JSON snapshot
+ * as a string or the binary snapshot as a `Uint8Array`.
+ * @param fluidFileConverter - Caller-provided code loader and execution logic. See {@link IFluidFileConverter}.
+ * @param baseLogger - Telemetry logger that will receive events emitted during load and execution. Typically
```
Josmithr, line 131: Nit: any reason we can't just call this "logger"? I don't see any other logger-related properties here, so it doesn't seem necessary to qualify.

Josmithr (commented):
I left a few docs-related comments. Otherwise looks good to me. Also, there is some great documentation in this PR - thank you!

packages/tools/fluid-runner/src/exportFile.ts
MarioJGMsoft: It was called baseLogger to clarify the transition from `TelemetryLoggerExt` to `ITelemetryBaseLogger`. but we can change to logger if that makes it clearer

packages/tools/fluid-runner/src/logger/fileLogger.ts
MarioJGMsoft: I think it would be better if this is done in a separate PR. For now we just want to introduce the new types.

packages/tools/fluid-runner/src/exportFile.ts
MarioJGMsoft: Actually, in the function itself we have another variable called `logger`, so renaming it to `logger` would cause issues there.

packages/tools/fluid-runner/src/logger/loggerUtils.ts
MarioJGMsoft: Specified in the comment

.changeset/expose-fluid-runner-helpers.md
MarioJGMsoft: Removed from changeset and moved to PR description

.changeset/expose-fluid-runner-helpers.md
MarioJGMsoft: Applied change

.changeset/expose-fluid-runner-helpers.md
MarioJGMsoft: Applied change

packages/tools/fluid-runner/src/parseBundleAndExportFile.ts
MarioJGMsoft: We can't because `logger` in this case is of `TelemetryLoggerExt` type, so since we need to use `ITelemetryBaseLogger` type, we pass the baseLogger.

packages/tools/fluid-runner/src/logger/loggerUtils.ts
```diff
@@ -31,7 +31,8 @@ import { JSONFileLogger } from "./jsonFileLogger.js";
  * to write CSV instead. See {@link IFileLoggerTelemetryOptions} for supported options including default
  * properties applied to every event and flush batching.
  *
- * @param filePath - Path to the file telemetry will be written to. Must not already exist.
+ * @param filePath - Path to the file telemetry will be written to. If the file already exists its
```
Josmithr, line 34: Thanks for the updates! This is really useful!

packages/tools/fluid-runner/src/exportFile.ts
Josmithr: I think it's generally preferred to prioritize the user-facing naming over internal naming. So it would probably be better (in my opinion) to name this param "logger" and rename the function-local variable to something else. But this is a nitpick.

Josmithr (approved):

PR #21767, opened 2024-07-03
Offline: Add batchId to batch metadata on resubmit

markfields:
## Description

Batch ID is the key piece of information needed to detect/prevent duplication of changes due to a forked container.

We compute Batch ID as a tuple of original client ID and original Client Sequence number (of the batch start). We only stamp the batchId on batch metadata (of first message in the batch) on resubmit - if a batch is missing batchId we will be able to compute it from the client ID and Client Sequence Number.

The next PR will be more interesting - comparing batch IDs in PendingStateManager as incoming messages are processed. This just sets it up by adding batch ID to batch metadata on resubmit.

## Breaking Changes

This PR breaks the format of the stashed pending local state. Back-compat support for reading old-format stashes is included. New format is backwards compatible (additive). This is slightly superfluous anyway since there is no production usage of the stashed schema.

## Reviewer Guidance

For the end-to-end flow, see prototype PR #21266

Here are some tests I considered in my test plan but I am not planning to write at this point (i.e. scenarios without direct test coverage).  Many of these will be covered in the next PR where we have end-to-end coverage of the forked container scenarios.

* PendingStateManager
  * Load from old stashed format. `batchIdContext !== undefined`
  * Load twice from old stashed format. Batches are not correlated and duplication occurs
  * Flushing two batches when disconnected should yield unique batch IDs
  * Resubmitting twice preserves existing batchId
  * Resubmitting after rehydrate preserves stashed batchIdContext or batchId
* end-to-end
  * batchId is present on sequenced messages being processed when expected

PR #14239, opened 2023-02-17
Don't allow `FlushModeExperimental.Async` if the loader does not support reference sequence numbers

andre4i:
## Description

With `FlushMode==FlushModeExperimental.Async`, batches will be flushed across JS turns, drastically increasing the chances of ending up in states with mismatched reference sequence numbers in the same batch. This change https://github.com/microsoft/FluidFramework/pull/14150 alleviates the risk, by flushing batches earlier. However, as old loader do not receive the reference sequence numbers from the runtime, there is considerable risk of mixed batches in that case (new runtime + FlushModeExperimental.Async - old loader). Therefore, we need to explicitly block this combination. 

Loaders from versions older than the present will not work with `FlushModeExperimental.Async` as they don't specify it's supported features in the ContainerContext implementation at all.

PR #10930, opened 2022-06-30
Make flushMode immutable and remove all tracking of it

kian-thompson:
# [AB#393](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/393) and [AB#394](https://dev.azure.com/fluidframework/internal/_workitems/edit/394)

## Description

The `setFlushMode` API is not currently used and enables a set of scenarios which should not be supported. Maintaining this API is costly, and it adds a non-trivial amount of complexity to the adjacent logic.

Because the `flushMode` should no longer change, the `PendingStateManager` should not track the switches in the previous system. Instead, the `PendingStateManager` will track and replay batches explicitly rather than through the setting of flush modes.

See https://github.com/microsoft/FluidFramework/issues/9480 and https://github.com/microsoft/FluidFramework/issues/9481.

## Does this introduce a breaking change?

The `setFlushMode` has been removed from `IContainerRuntimeBase`. Please remove all usage of this method as FlushMode is now an immutable property for the container runtime, optionally provided at creation time via the `IContainerRuntimeOptions` interface.

The `PendingStateManager` no longer tracks flush mode as it is now an immutable property on the `IContainerRuntimeBase`. Instead, tracking of batches is done explicitly and they are replayed directly through usage of the `orderSequentially` method on `IContainerRuntimeBase`.

## Other information or known dependencies

TODO after this PR:
- Add tests to improve batch coverage in `containerRuntime.spec.ts`. See [comment](https://github.com/microsoft/FluidFramework/pull/10503#discussion_r892645598).
    - [AB#1060](https://dev.azure.com/fluidframework/internal/_workitems/edit/1060)
- Deprecate `IContainerRuntime.flush`. This method should not be exposed publicly.
    - [AB#1059](https://dev.azure.com/fluidframework/internal/_workitems/edit/1059)

PR #27184, opened 2026-04-28
Auto-Enable Duplicate Batch Detection on Pending-State Capture

dannimad:
# Auto-Enable Duplicate Batch Detection on Pending-State Capture

## Context

Duplicate batch detection in the container runtime catches "forked container" scenarios — two containers loaded from the same serialized pending state submitting batches with the same batchId. Today it is gated by two opt-in config flags (`Fluid.Container.enableOfflineFull` and `Fluid.ContainerRuntime.enableBatchIdTracking`) that callers must set explicitly.

This change removes the explicit flag gating and makes detection auto-activate whenever the runtime is involved in pending-state lifecycle: when it loads with pending state, when it loads a snapshot that already carries `recentBatchInfo`, or the moment `getPendingLocalState()` is called. The intent is to make safety the default — any runtime exposed to the offline-load fork risk gets detection automatically.

A short-lived kill-switch flag (`Fluid.ContainerRuntime.DisableDuplicateBatchDetection`) is added so the change can be rolled back via configuration if it regresses, then removed in a follow-up.

## Design Summary

- Activate the `DuplicateBatchDetector` when ANY of:
  1. The runtime is constructed with `context.pendingLocalState !== undefined` (rehydrated peer of a fork).
  2. The runtime is constructed with a non-empty `recentBatchInfo` array (snapshot already has tracked batches).
  3. `getPendingLocalState()` is called (this runtime captures pending state for offline rehydration).
- Once activated, the detector lives for the runtime's lifetime.
- Resubmits stamp `batchId` whenever the detector is active.
- The `FlushMode === TurnBased` precondition moves from the constructor to `getPendingLocalState()`.
- Add kill-switch `Fluid.ContainerRuntime.DisableDuplicateBatchDetection` (default false). Delete reads of the two old flags from `containerRuntime.ts`. Loader-layer reads of `enableOfflineFull` (in `container.ts` and `snapshotRefresher.ts`) are NOT touched — they gate unrelated SerializedStateManager and snapshot-refresh behavior.

## Critical Files

- [packages/runtime/container-runtime/src/containerRuntime.ts](packages/runtime/container-runtime/src/containerRuntime.ts) — primary changes
- [packages/runtime/container-runtime/src/opLifecycle/duplicateBatchDetector.ts](packages/runtime/container-runtime/src/opLifecycle/duplicateBatchDetector.ts) — no code changes; reused as-is
- [packages/runtime/container-runtime/src/test/containerRuntime.spec.ts](packages/runtime/container-runtime/src/test/containerRuntime.spec.ts) — update existing tests, add new ones
- [packages/test/test-end-to-end-tests/src/test/batching.spec.ts](packages/test/test-end-to-end-tests/src/test/batching.spec.ts) — drop now-defunct setting
- [packages/test/test-end-to-end-tests/src/test/offline/stashedOps.spec.ts](packages/test/test-end-to-end-tests/src/test/offline/stashedOps.spec.ts) — verify still passes (loader still reads `enableOfflineFull`)
- [packages/test/local-server-stress-tests/src/test/localServerStress.spec.ts](packages/test/local-server-stress-tests/src/test/localServerStress.spec.ts) — verify still passes
- New changeset under `.changeset/`

### DIFF
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -1517,17 +1517,34 @@ export class ContainerRuntime
 	 * serialized pending state and submit batches with the same batchId.
 	 *
 	 * @remarks
-	 * Activates automatically (sticky once on) when this runtime is involved in the
-	 * pending-state lifecycle:
+	 * Activates automatically (sticky once on) once we have any signal that fork
+	 * detection is in play in this session:
 	 *
 	 * - rehydrated from a captured pending state at construction (`pendingLocalState`),
 	 * - loaded from a snapshot whose `recentBatchInfo` blob is non-empty,
-	 * - or `getPendingLocalState()` is called (the moment we capture pending state).
+	 * - `getPendingLocalState()` is called (the moment we capture pending state),
+	 * - or an inbound batch arrives carrying an explicit `batchId` in its metadata,
+	 * which means some peer has activated stamping, so this runtime — even as a
+	 * pure observer — needs to be tracking too in order to stay consistent.
 	 *
 	 * Can be force-disabled via the `Fluid.ContainerRuntime.DisableDuplicateBatchDetection`
 	 * config (see {@link ContainerRuntime.duplicateBatchDetectionDisabled}).
 	 */
 	private duplicateBatchDetector: DuplicateBatchDetector | undefined;
+
+	/**
+	 * Whether this runtime should stamp `batchId` on resubmitted batches so peers
+	 * can detect forked-container collisions.
+	 *
+	 * @remarks
+	 * Conceptually orthogonal to {@link ContainerRuntime.duplicateBatchDetector}
+	 * (one is about inbound observation, the other about outbound stamping), but
+	 * we currently use a single all-or-nothing switch: stamping is on iff the
+	 * detector is active.
+	 */
+	private get enableBatchIdTracking(): boolean {
+		return this.duplicateBatchDetector !== undefined;
+	}
 	private readonly outbox: Outbox;
 	private readonly garbageCollector: IGarbageCollector;
 
@@ -3133,6 +3150,12 @@ export class ContainerRuntime
 
 			if ("batchStart" in inboundResult) {
 				const batchStart: BatchStartInfo = inboundResult.batchStart;
+				// An explicit batchId means some peer has activated stamping, so we
+				// must track too — even as a pure observer — to stay consistent with
+				// peers that will throw on the duplicate.
+				if (batchStart.batchId !== undefined) {
+					this.ensureDuplicateBatchDetector();
+				}
 				const result = this.duplicateBatchDetector?.processInboundBatch(batchStart);
 				if (result?.duplicate === true) {
 					const error = new DataCorruptionError(
@@ -4979,11 +5002,11 @@ export class ContainerRuntime
 		);
 
 		const resubmitInfo = {
-			// Stamp the batchId on resubmits whenever the duplicate batch detector is
-			// active (this runtime is part of the pending-state lifecycle and a forked
-			// peer might resubmit the same batches). Otherwise omit it to avoid
-			// inflating message metadata.
-			batchId: this.duplicateBatchDetector === undefined ? undefined : batchId,
+			// Stamp batchId on resubmits whenever fork-detection is in play in this
+			// session. Detector-on and stamping-on are conceptually distinct concerns
+			// (inbound observation vs. outbound cooperation), but currently share a
+			// single all-or-nothing switch (see {@link enableBatchIdTracking}).
+			batchId: this.enableBatchIdTracking ? batchId : undefined,
 			staged,
 		};
 
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -1449,13 +1449,10 @@ export class ContainerRuntime
 	private readonly _flushMode: FlushMode;
 	private readonly stagingModeAutoFlushThreshold: number;
 	/**
-	 * BatchId tracking is needed whenever there's a possibility of a "forked Container",
-	 * where the same local state is pending in two different running Containers, each of
-	 * which is trying to ensure it's persisted.
-	 * "Offline Load" from serialized pending state is one such scenario since two Containers
-	 * could load from the same serialized pending state.
+	 * Kill switch for {@link ContainerRuntime.duplicateBatchDetector} auto-activation.
+	 * When true, the detector is never constructed and resubmits omit batchIds.
 	 */
-	private readonly batchIdTrackingEnabled: boolean;
+	private readonly duplicateBatchDetectionDisabled: boolean;
 	private flushScheduled = false;
 
 	private canSendOps: boolean;
@@ -1515,7 +1512,22 @@ export class ContainerRuntime
 	private readonly inboundBatchAggregator: InboundBatchAggregator;
 	private readonly blobManager: BlobManager;
 	private readonly pendingStateManager: PendingStateManager;
-	private readonly duplicateBatchDetector: DuplicateBatchDetector | undefined;
+	/**
+	 * Detects "forked container" scenarios where two containers load from the same
+	 * serialized pending state and submit batches with the same batchId.
+	 *
+	 * @remarks
+	 * Activates automatically (sticky once on) when this runtime is involved in the
+	 * pending-state lifecycle:
+	 *
+	 * - rehydrated from a captured pending state at construction (`pendingLocalState`),
+	 * - loaded from a snapshot whose `recentBatchInfo` blob is non-empty,
+	 * - or `getPendingLocalState()` is called (the moment we capture pending state).
+	 *
+	 * Can be force-disabled via the `Fluid.ContainerRuntime.DisableDuplicateBatchDetection`
+	 * config (see {@link ContainerRuntime.duplicateBatchDetectionDisabled}).
+	 */
+	private duplicateBatchDetector: DuplicateBatchDetector | undefined;
 	private readonly outbox: Outbox;
 	private readonly garbageCollector: IGarbageCollector;
 
@@ -1884,21 +1896,21 @@ export class ContainerRuntime
 			this.mc.config.getNumber("Fluid.ContainerRuntime.StagingModeAutoFlushThreshold") ??
 			runtimeOptions.stagingModeAutoFlushThreshold ??
 			defaultStagingModeAutoFlushThreshold;
-		this.batchIdTrackingEnabled =
-			this.mc.config.getBoolean("Fluid.Container.enableOfflineFull") ??
-			this.mc.config.getBoolean("Fluid.ContainerRuntime.enableBatchIdTracking") ??
-			false;
+		this.duplicateBatchDetectionDisabled =
+			this.mc.config.getBoolean("Fluid.ContainerRuntime.DisableDuplicateBatchDetection") === true;
 
-		if (this.batchIdTrackingEnabled && this._flushMode !== FlushMode.TurnBased) {
-			const error = new UsageError("Offline mode is only supported in turn-based mode");
-			this.closeFn(error);
-			throw error;
-		}
-
-		// DuplicateBatchDetection is only enabled if Offline Load is enabled
-		// It maintains a cache of all batchIds/sequenceNumbers within the collab window.
-		// Don't waste resources doing so if not needed.
-		if (this.batchIdTrackingEnabled) {
+		// Activate the duplicate batch detector when this runtime is involved in the
+		// pending-state lifecycle: rehydrated from a captured pending state, or loading
+		// a snapshot that already tracks recent batches. Otherwise, lazily activate when
+		// getPendingLocalState() is called (see ensureDuplicateBatchDetector). Activation
+		// is sticky once on for the runtime's lifetime.
+		const loadedWithPendingState = pendingLocalState !== undefined;
+		const loadedWithRecentBatchInfo =
+			recentBatchInfo !== undefined && recentBatchInfo.length > 0;
+		if (
+			!this.duplicateBatchDetectionDisabled &&
+			(loadedWithPendingState || loadedWithRecentBatchInfo)
+		) {
 			this.duplicateBatchDetector = new DuplicateBatchDetector(recentBatchInfo);
 		}
 
@@ -4953,8 +4965,8 @@ export class ContainerRuntime
 	 * checks in the ConnectionStateHandler (Loader layer)
 	 *
 	 * The only exception to this would be if the Container "forks" due to misuse of the "Offline Load" feature.
-	 * If the "Offline Load" feature is enabled, the batchId is included in the resubmitted messages,
-	 * for correlation to detect container forking.
+	 * When the duplicate batch detector is active, the batchId is included in the resubmitted messages
+	 * so a forked peer can detect duplicates against its own tracked batch IDs.
 	 */
 	private reSubmitBatch(
 		batch: PendingMessageResubmitData[],
@@ -4966,9 +4978,11 @@ export class ContainerRuntime
 		);
 
 		const resubmitInfo = {
-			// Only include Batch ID if "Offline Load" feature is enabled
-			// It's only needed to identify batches across container forks arising from misuse of offline load.
-			batchId: this.batchIdTrackingEnabled ? batchId : undefined,
+			// Stamp the batchId on resubmits whenever the duplicate batch detector is
+			// active (this runtime is part of the pending-state lifecycle and a forked
+			// peer might resubmit the same batches). Otherwise omit it to avoid
+			// inflating message metadata.
+			batchId: this.duplicateBatchDetector === undefined ? undefined : batchId,
 			staged,
 		};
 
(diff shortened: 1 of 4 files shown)


dannimad:
Here's the exhaustive review of the current changes — auto-activating DuplicateBatchDetector based on pending-state lifecycle, with kill switch and FlushMode check moved to getPendingLocalState().

Critical / High priority
1. Asymmetric activation between forked containers and passive readers
With the old opt-in flag, every client in the session that wanted detection had to set the flag explicitly. With auto-activation, only clients that are themselves part of the pending-state lifecycle activate:

- A fresh third-party reader that never captured pending state and loads a snapshot without recentBatchInfo will never activate. If a forked pair (A, A′) both submit the same batch, that reader silently applies both. The fork itself is detected on A or A′, but other observers will see corruption.
- The chain depends on recentBatchInfo being persisted in summaries. Verify containerRuntime.ts:2712-2716 writes the blob whenever the detector is active so downstream loaders activate. Currently it only writes if getRecentBatchInfoForSummary returns non-undefined; if the writer was active but had no batches yet, the blob will be missing and a non-pending-state reader of a future summary will not activate. Worth confirming this is acceptable (probably yes — they have nothing to detect).
2. Stale ops between activation and serialization in getPendingLocalState()
The lazy activation order is: throw on non-TurnBased → ensureDuplicateBatchDetector() → flush() → snapshot pending state. Resubmits of any batches that were already submitted prior to the call were sent without batchId because the detector was undefined. If the host then forks via the captured pending state, the fork will resubmit those exact batches with a stamped batchId (its detector is active at construction). The original's already-acked unstamped batch and the fork's stamped resubmit cannot be correlated — the duplicate is undetectable. This was a property of the old opt-in too (you had to enable beforehand), but the auto-model implies it is "always on once you opt into pending-state," which can be misleading. Worth being explicit in docs/changeset that detection is only effective for ops submitted while the detector is active.

3. Removal of constructor-time guard for Immediate + pendingLocalState
Previously the runtime called closeFn(error); throw error if Offline Load was enabled in non-TurnBased mode. Now the constructor silently accepts pendingLocalState in Immediate mode and only throws at getPendingLocalState() time. Two consequences:

- Constructing with bad config no longer fails fast — code paths that re-load from pending state in Immediate mode now succeed and may then misbehave at op-replay time. Verify nothing else along the rehydration codepath assumes TurnBased.
- The e2e test "Can't capture pending state in Immediate Mode" no longer exercises the rehydrate path; it only exercises getPendingLocalState. Consider adding a load-with-pendingLocalState-in-Immediate-mode test that exercises applyStashedOp/resubmit and confirms behavior is sane (or that we reject it somewhere).
4. Test fixture pendingLocalState: {} is not a realistic shape
Tests pass a literal {}. The runtime casts to IPendingRuntimeState | undefined and reads pendingRuntimeState?.pending. With {}, pending is undefined — fine for activation, but it bypasses the rest of the rehydration codepath (containerRuntime.ts:1196, containerRuntime.ts:1101). It also flips the Summary metadata mismatch check off. The test still validates the activation surface, but it is not exercising real rehydration. Either document it as "construct-time activation flag" or use a more realistic fixture (e.g., the value returned by a real getPendingLocalState()).

5. Snapshot refresh path is not re-checked
A runtime that loads with recentBatchInfo === undefined (detector inactive) and is later refreshed via SerializedStateManager snapshot refresh won't reconsider activation — the detector is sticky-off until getPendingLocalState() is called. If a refreshed snapshot brings a non-empty recentBatchInfo blob, that information is dropped. Check whether the snapshot refresh codepath in fact re-runs runtime construction (it might) or whether this is a real gap.

6. Loader-layer flag still reads Fluid.Container.enableOfflineFull
The changeset notes this is intentional. Cross-check:

- container.ts:967 and snapshotRefresher.ts:101 still gate behavior on the loader-side flag.
- Two e2e tests and one stress test still set Fluid.Container.enableOfflineFull: true — those are the loader-side tests, not runtime-side, so this is consistent. Still worth a sanity check that the loader-side scenarios still get duplicate-batch protection at the runtime layer (they do, because they will go through getPendingLocalState() and rehydration).

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -5275,6 +5301,16 @@ export class ContainerRuntime
 			throw new UsageError("can't get state while manually accumulating a batch");
```
anthony-murphy, line 5324: Deep Review: Relocating the FlushMode check from the constructor to `getPendingLocalState()` silently drops a host-observable telemetry contract.

The old constructor block was:

```ts
const error = new UsageError("Offline mode is only supported in turn-based mode");
this.closeFn(error);
throw error;
```

It both threw **and** closed the container, producing a `fluid:telemetry:Container:ContainerClose` event. The deleted e2e test in `batching.spec.ts` (~lines 535–550) was an `itExpects(..., [{ eventName: "fluid:telemetry:Container:ContainerClose", error: "Offline mode is only supported in turn-based mode" }], ...)` block — direct evidence that the telemetry contract was being asserted.

The new code at `getPendingLocalState()` throws `UsageError("getPendingLocalState requires FlushMode.TurnBased")` but does **not** call `closeFn`, and the replacement test was downgraded to a plain `it(...)` doing only `assert.rejects(...)`. The changeset documents the relocation but doesn't mention the dropped close-and-`ContainerClose` behavior. Your own self-review (item 3) flags the same concern.

Two reasonable resolutions:

1. **Restore the close-on-throw contract**: call `this.closeFn(error)` before the throw in `getPendingLocalState()`, and re-promote the test to `itExpects` asserting the `ContainerClose` event. Lowest risk; preserves the host-observable contract verbatim.
2. **Accept the new contract explicitly**: call out in the changeset that the constructor no longer closes the container and that `getPendingLocalState()` now rejects without emitting `ContainerClose`, and add an `itExpects` test asserting whatever telemetry (if any) does fire on the new path so the contract remains tested rather than just asserted-as-rejection.

Either way, the asymmetric downgrade from `itExpects` → `assert.rejects` should not stand — the previous assertion was the only mechanical guard against this contract drifting.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1515,7 +1512,22 @@ export class ContainerRuntime
 	private readonly inboundBatchAggregator: InboundBatchAggregator;
 	private readonly blobManager: BlobManager;
 	private readonly pendingStateManager: PendingStateManager;
-	private readonly duplicateBatchDetector: DuplicateBatchDetector | undefined;
+	/**
+	 * Detects "forked container" scenarios where two containers load from the same
+	 * serialized pending state and submit batches with the same batchId.
+	 *
+	 * @remarks
+	 * Activates automatically (sticky once on) when this runtime is involved in the
```
markfields: I think we also need to activate it as soon as we process a batch with a batchId.

Reason being, there could be a third session that loaded without any pending state that then observes the parallel forks case.  It will process the two duplicate sequenced batches and we want to catch that immediately.

dannimad: Doesnt that session must start with a `recentBatchInfo`?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4966,9 +4979,11 @@ export class ContainerRuntime
...
 
 		const resubmitInfo = {
-			// Only include Batch ID if "Offline Load" feature is enabled
-			// It's only needed to identify batches across container forks arising from misuse of offline load.
-			batchId: this.batchIdTrackingEnabled ? batchId : undefined,
+			// Stamp the batchId on resubmits whenever the duplicate batch detector is
+			// active (this runtime is part of the pending-state lifecycle and a forked
+			// peer might resubmit the same batches). Otherwise omit it to avoid
+			// inflating message metadata.
+			batchId: this.duplicateBatchDetector === undefined ? undefined : batchId,
```
markfields: I think we should tease apart two related but different aspects of batch ID tracking:

1. Stamping batchId during resubmit makes sure all clients can detect any manner of container forking that's possible in the presence of pending state.
2. `DuplicateBatchDetector` (maybe should be renamed to `DuplicateSequencedBatchDetector`) covers parallel forks in particular, and needs to run on any client observing batch IDs (see other comment about this)

In particular, I'd suggest this:

- batchId stamping on resubmit:  Enabled as soon as pending state is in play (loaded from or captured). This is the thing previously called `batchIdTrackingEnabled`
  - Note: Even if `recentBatchInfo` is in the snapshot, we don't necessarily need to stamp batchIds.  We only need to do that if _that batch_ could be held locally by two forked containers, which required pending state interaction.
- DuplicateSequencedBatchDetector:  Enabled as soon as it ever sees an explicit batchId* (and naturally if `recentBatchInfo` is in the snapshot)


\* Footnote: there's a theoretical race condition where one of the duplicate sequenced batches is the _original_ submission (not resubmit), so it doesn't have an explicit batchId. An observing container could get caught with DSBD disabled.  I don't think it's worth covering this case.  It basically requires always having DSBD on.  We can track this as a follow-up IMO.

dannimad: If I understand your suggestion correctly, you're thinking more of an optimization rather than a simplification. Current PR approach looks simpler to me given that it's like an all or nothing switch (both DSBD and batch stamping). I'd be a bit concerned about some scenarios/combinations of one thing being enabled and another one disabled where we can't reason clearly why it's ok or not to be in that state. 
 
If you think that optimization is worth making I can start reason through it 
 
And also I'm considering this an optimization since seems like the motivation behind the distinction is not enabling batch detection in scenarios where it's not strictly needed

packages/runtime/container-runtime/src/containerRuntime.ts
markfields: nope.  It could have been running for days before the other clients did anything with pending state

packages/runtime/container-runtime/src/containerRuntime.ts
markfields: There's been confusion before between these two facets of batch tracking (you and Tony were concerned that there was some bug a few months back, it came down to this I believe).  Seeing `batchId: this.duplicateBatchDetector === undefined ? undefined : batchId` is what actually doesn't look right to me. Whether we're looking for incoming duplicate sequenced batches is orthogonal to whether we're liable to see this container fork going forward.
 
Perhaps the all-or-nothing approach is fine (you're right that my suggestion is also an optimization), but then my request would be to use different language.
 
Maybe just wrap `this.duplicateBatchDetector === undefined` in a property getter called `enableBatchIdTracking` or whatever it was called.

anthony-murphy:
## Deep Review

*Reviewed commit `84283f0` on 2026-04-29.*

**Readiness: 3/10 — GETTING STARTED**

Auto-activating `DuplicateBatchDetector` from runtime signals is the right direction, but a closer reading of the detector's own design comment surfaces a new design gap: a *pure observer* that processes an unstamped original batch before any stamped resubmit will not feed that batch to `processInboundBatch` (the inbound gate only fires when `batchStart.batchId !== undefined`), so a later stamped duplicate is recorded as a first sighting rather than throwing — defeating the detector's documented original-vs-resubmit comparison. This is adjacent to but mechanically distinct from the open pure-observer thread on this file (it's about arrival ordering, not late arrival). Previously raised concerns (FlushMode/`closeFn` contract change; outbound-stamping coupled to inbound-detect activation) are still active in their threads and held until this design question is resolved.

### Findings

- **Pure observers miss an unstamped original batch that arrives before any stamped resubmit** — packages/runtime/container-runtime/src/containerRuntime.ts:120-137 — (inline thread)

### Path to Ready

To reach 4–6 (MAKING PROGRESS) — resolve the new design question:
- [ ] Decide disposition of the pure-observer earliest-arrival gap: (a) preserve enough pre-activation tracking (e.g., a small bounded recent-batch ring) so a later activation can reconcile against earlier unstamped originals; (b) activate the detector unconditionally for fork-exposed sessions (likely all non-summarizer TurnBased runtimes); or (c) explicitly accept the gap and document it in the changeset as a known limitation. Coordinate with markfields on the open thread above since the conversations overlap.
- [ ] Add a regression test: a non-pending observer processes an unstamped original batch, then a stamped resubmit with the same effective batch ID; assert the throw (or the documented non-throw if option (c) is taken).

To reach 7–8 (ALMOST READY) — previously-raised work, held since the prior review:
- [ ] Resolve the FlushMode/`closeFn` contract change in the existing inline thread: either restore `this.closeFn(error)` before the new `getPendingLocalState()` throw and re-promote the e2e test from `assert.rejects` back to `itExpects(... ContainerClose ...)`, or document the new contract in the changeset and pin it with an `itExpects` test asserting whatever telemetry actually fires on the new path.
- [ ] Resolve markfields' open thread on inbound-detect vs outbound-stamping conflation: either split the two (introduce a separate getter so the inbound-`batchId` activation trigger doesn't also turn on outbound stamping) or accept the unified switch and rename `enableBatchIdTracking`-style language to make the coupling visible at every callsite (markfields' explicit suggestion in that thread).
- [ ] Audit `opLifecycle/outbox.ts`'s empty-batch resubmit path against the new `duplicateBatchDetector !== undefined` predicate. A quick grep for `enableOfflineFull` and `enableBatchIdTracking` across `container-runtime` should confirm nothing still keys off the removed flags.

To reach 9–10 (READY) — polish, held until the above is resolved:
- [ ] Add a unit test that sets `Fluid.ContainerRuntime.DisableDuplicateBatchDetection=true`, processes an inbound batch with an explicit `batchId`, and asserts `duplicateBatchDetector` remains `undefined` (covers the kill switch against the inbound-`batchId` trigger, not just construction-time triggers).
- [ ] Add an assertion that after `processBatchWithId(...)` triggers activation, `reSubmitBatch` then produces a stamped `batchId` — locks in the inbound→stamping coupling as currently shipped.
- [ ] Either pass loaded `recentBatchInfo` through summaries unconditionally when the kill switch is engaged, or document the asymmetric-rollback hazard in the kill-switch comment.
- [ ] Optional: add a summarizer short-circuit in `ensureDuplicateBatchDetector` so the inbound-`batchId` trigger doesn't fire on summarizer clients (efficiency only).
- [ ] Emit a one-shot activation telemetry event with `reason ∈ {ctorPendingState, ctorRecentBatchInfo, getPendingLocalState, inboundBatchId}` and a parallel suppressed-by-kill-switch event so rollout is observable. Confirm `stashedOps.spec.ts` and `localServerStress.spec.ts` still trip the detector via the new auto-triggers now that the historical `enableOfflineFull` co-gating is gone.
- [ ] Add an e2e variant rehydrating two containers from the same captured pending state with **no** offline-related flags set, asserting `DataCorruptionError` / `DuplicateBatch` telemetry on the duplicate side — directly validates the headline "safety by default" claim.
- [ ] One-line comment near the new `getPendingLocalState()` throw clarifying that this is operation-time validation of an immutable construction-time property (FlushMode is fixed once set), to pre-empt the constructor-vs-operation-time-throw harmonization question.

### Context for Reviewers

- **Suggested reviewers**: required — markfields (forked-container detection design and the original opt-in choice this PR overturns; already engaged on two threads), anthony-murphy (already engaged via the FlushMode thread). Optional — kian-thompson (PR #10930 made FlushMode immutable, which is load-bearing for the relocated TurnBased gate), andre4i / vladsud (precedent in PR #14239 on throw-vs-fallback for flushMode/feature mismatch), dannimad as author.
- **Historical context**: PR #21767 (markfields) introduced batch IDs as deliberately flag-gated, opt-in plumbing for forked-container detection — "the key piece of information needed to detect/prevent duplication of changes due to a forked container" — with explicit feature-flag gating and an acknowledged "reading this flag from two different places" comment in that review. The current PR collapses that gate into runtime-internal signals; the description lists three triggers, but a fourth (inbound-`batchId` gossip) is acknowledged in the in-code comment as "conceptually orthogonal" yet shipped coupled to outbound stamping — worth treating as a substantive design choice rather than implementation detail.
- **FlushMode immutability**: PR #10930 made `flushMode` an immutable property of the runtime. Because flushMode never changes after construction, throwing at `getPendingLocalState()` is control-flow-equivalent to throwing at construction — which makes the failure-mode change (close-and-throw → throw-only) the *only* observable difference, and the one the existing FlushMode inline thread targets.
- **Reviewer hot spots**: (1) the conflation of "detector active" and "stamping enabled" (flagged orthogonal in code, shipped coupled); (2) four (not three) activation triggers — the inbound-`batchId` trigger is missing from the PR description list; (3) failure-mode change for `getPendingLocalState` in non-TurnBased runtimes, invisible from the diff alone unless the reader knows flushMode is immutable; (4) kill-switch test coverage against every activation trigger, not just construction-time ones; (5) outbox empty-batch resubmit path against the new `duplicateBatchDetector !== undefined` predicate.
- **Out-of-PR cross-checks the review couldn't make**: whether the loader-layer `enableOfflineFull` reads in `container.ts` / `snapshotRefresher.ts` are genuinely independent of duplicate-batch detection (PR claims yes; not exercised in the diff); whether `stashedOps.spec.ts` and `localServerStress.spec.ts` actually trip the detector via the new auto-activation triggers now that the historical `enableOfflineFull` co-gating is gone; whether the outbox empty-batch resubmit path silently keys off the removed flags (`outbox.ts` is not in the PR diff).

### Suggested description

The current description is adequate but understates two design choices and one behavior change. Consider appending to the **Design Summary** section:

> **Fourth activation trigger:** in addition to the three listed, the runtime also activates the detector lazily when it processes an inbound batch carrying an explicit `batchId`. This is intentional — it ensures pure-observer clients in a session where peers have stamped will throw on duplicates rather than silently apply both. Note that activation also unifies outbound `batchId` stamping on resubmit (the same internal accessor gates both), so an observer that activates via this trigger will subsequently stamp its own resubmits.
>
> **Pure-observer earliest-arrival caveat:** the inbound trigger fires only on batches that already carry an explicit `batchId`. If an unstamped *original* batch is processed before any stamped resubmit arrives, that original is not recorded by the detector; a later stamped duplicate will be seen as a first sighting. This is the exact arrival ordering `DuplicateBatchDetector`'s design comment (`opLifecycle/duplicateBatchDetector.ts`) calls out, and needs explicit handling — either pre-activation tracking or unconditional activation for fork-exposed sessions.
>
> **Behavior change on FlushMode misuse:** the constructor previously called `this.closeFn(error)` and threw when offline-load was enabled in non-`TurnBased` mode, producing `fluid:telemetry:Container:ContainerClose`. The relocated check in `getPendingLocalState()` now rejects with `UsageError("getPendingLocalState requires FlushMode.TurnBased")` but does **not** close the container or emit `ContainerClose`. Hosts keying off that telemetry must migrate to observing the rejection at the call site.
>
> **Cost of the new default:** the per-runtime `DuplicateBatchDetector` cache is bounded by the collab window (entries pruned by minimum sequence number — see `duplicateBatchDetector.ts:54-99`) and observable via `fluid_DuplicateBatchDetector_recentBatchCount` in summary telemetry. The kill switch `Fluid.ContainerRuntime.DisableDuplicateBatchDetection` is the supported escape hatch.

- `c062bc6` 2026-04-28 · **5/10** — FlushMode/`closeFn` contract change posted as inline thread; activation surface accepted

</details>

PR #22454, opened 2024-09-09
Offline: Add recent batch info from DuplicateBatchDetector to summary

markfields:
## Description

Fixes [AB#13963](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/13963)

We've added a new data structure for tracking "recent" (based on MSN) batchIds which are at risk for seeing a duplicate sequence batchID come in.  This data structure is consulted and modified on each incoming batch.

Whenever we summarize, we need to include this state as of the given sequence number, and when we load from a snapshot we need to repopulate it as well.

Why? Because summaries can happen at any time, and a future container loading from that point will need to know the recent batch IDs that may possibly be followed by a sequenced duplicate.

PR #22310, opened 2024-08-24
Offline: Detect duplicate sequenced batches and close the container

markfields:
## Description

Fixes [AB#8884](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/8884)

If the same batch is submitted twice in parallel by two rehydrated forks of the same container, the duplicate will get sequenced, and all subsequent clients will need to notice this to avoid applying both copies.

For now, we will simply detect this case and close the container with a DataCorruptionError.

PR #21854, opened 2024-07-11
Send empty group batch on resubmit

dannimad:
Fixes [empty group batch on empty resubmit](https://github.com/microsoft/FluidFramework/commit/49aaffd262b8d607cc5c700c5cc1d323d167411c)
TODO:
- [x] Add tests to exercise this case explicitly
- [x] Fix Immediate test failures
- [x] Change batchStartCsn assert to throw exception
- [x] What about when reference seq num === undefined ? 
- [x] Should we do the same for blobAttachBatch ?

PR #27185, opened 2026-04-28
refactor(container-runtime): extract feature/manager classes from ContainerRuntime

anthony-murphy:
## Description

Drives `containerRuntime.ts` from **5582 → 4900 lines (-682, ~12%)** by introducing an `IRuntimeFeature` framework and extracting purpose-built classes. Motivation: the `ContainerRuntime` class is a monolith — too many concerns are tangled inside it, message dispatch is split across multiple residual switches, and inline closures plumb runtime state into nearly every subsystem constructor. This PR establishes a feature pattern that subsystems opt into, then migrates twelve subsystems / concerns onto it.

**Feature framework** (`runtimeFeature.ts`, `runtimeFeatureCollection.ts`): `IRuntimeFeature<TOps>` interface + composite `RuntimeFeatureCollection` that fans out lifecycle calls and routes ops O(1) via a `Map<ContainerMessageType, IRuntimeFeature>` built from each feature's `supportedOps`. Hooks: `onLoadFromSnapshot`, `setConnectionState`, `notifyStagingMode`, `dispose`, `contributeSummary`, `handleOp`, `applyStashedOp`, `reSubmitOp`, `rollbackStagedOp`. The op-routing hooks are parameterized on `TOps` so each feature's `message`/`messagesContent` parameters narrow to the variants it claims (helpers: `InboundRuntimeMessageFor`, `LocalRuntimeMessageFor`, `RuntimeMessagesContentFor`).

**Subsystems migrated to features** (own their own routing instead of being switched-on inside the runtime):
- `ChannelCollection`, `BlobManager`, `GarbageCollector`, `DocumentsSchemaController` — op routing + summary contribution
- `PendingStateManager`, `RemoteMessageProcessor`, `DuplicateBatchDetector`, `InboundBatchAggregator`, `DeltaScheduler`
- `SignalTelemetryManager` — self-resets via `setConnectionState`

**New feature classes**:
- `IdCompressorFeature` — owns the compressor instance, pending-range queue, on-boot/delayed load, IdAllocation message routing, summary blob, and outbound `LocalBatchMessage` shape
- `RuntimeOpsFeature` — Rejoin / ChunkedOp routing
- `SummarizerSubsystem` — election summary blob

**New helper classes** (not features, just monolith reduction):
- `ReconnectTracker` — consecutive-reconnects-with-no-progress detector
- `LoadingGroupSnapshotFetcher` — per-loadingGroupId snapshot cache + catch-up wait
- `ExtensionsManager` — ContainerExtension store, layer-compat plumbing, lazy event emitter wiring

**Op-routing collapse**: `validateAndProcessRuntimeMessages`, `applyStashedOp`, `reSubmit`, `rollbackStagedChange` all converged on the same shape — `features.<verb>Op` decides ownership; runtime only handles the unknown-type fail. The IdAllocation / Rejoin / ChunkedOp residual switches are gone.

**Inline closure cleanup**: outbox `generateIdAllocationOp` (17-line closure) moved into `IdCompressorFeature.generateAllocationOp`; runtime's `applyStashedOp` + `parseLocalOpContent` removed (PSM dispatches directly through `features.applyStashedOp`).

## Reviewer Guidance

This is a draft for early feedback. Specific things I'd appreciate eyes on:

- **Feature framework shape** — `IRuntimeFeature` is intentionally optional-everything; the runtime calls `features.<hook>(...)` and each member implements only what applies. Is the surface right? Anything missing for future migrations?
- **Op-routing dispatch ownership** — each `ContainerMessageType` has exactly one owning feature; duplicate claims throw at construction time (`registerOpClaims`). Registration order matters only for fan-out hooks (`onLoadFromSnapshot`, `contributeSummary`, `setConnectionState`, etc.), which iterate features in order.
- **`SignalTelemetryManager.setConnectionState` reset** moved from explicit call inside `setConnectionStateCore` to a self-tracked transition (was→is canSendOps). Behaviorally equivalent, but worth a check.
- **Skipped after analysis** (and noted at extraction time):
  - Folding `notifyReadOnlyState` into `setConnectionState` — separate signal path from the legacy `_deltaManager.on("readonly")` listener; risks regressions.
  - `IRuntimeContext` shared object — would touch every subsystem constructor; deserves its own branch.
  - `submitSummary` (375 lines) and `processInboundMessageOrBatch` (143 lines) — too tightly coupled to runtime state for clean extraction without the shared-context refactor.

**Test results:**
- `@fluidframework/container-runtime` mocha — 1966 passing, 2 pending, 0 failing
- `@fluid-internal/local-server-stress-tests` — 199 passing, 1 pending, 0 failing
- `@fluid-internal/local-server-tests` — 65 passing, 4 pending; 8 unrelated `captureFullContainerState` failures (function lives on a different branch)


packages/runtime/container-runtime/src/gc/garbageCollection.ts
```diff
@@ -1185,6 +1194,69 @@ export class GarbageCollector implements IGarbageCollector {
...
 	}
 
+	// === IRuntimeFeature ===
+
+	public async onLoadFromSnapshot(): Promise<void> {
+		await this.initializeBaseState();
+	}
+
+	public onConnectionStateChange(canSendOps: boolean, clientId: string | undefined): void {
+		this.setConnectionState(canSendOps, clientId);
```
anthony-murphy: just rename the method rather than adding another

packages/runtime/container-runtime/src/gc/garbageCollection.ts
```diff
@@ -1185,6 +1194,69 @@ export class GarbageCollector implements IGarbageCollector {
 		this.unreferencedNodesState.clear();
 	}
 
+	// === IRuntimeFeature ===
+
+	public async onLoadFromSnapshot(): Promise<void> {
+		await this.initializeBaseState();
```
anthony-murphy, line 1213: rename rather than wrap

packages/runtime/container-runtime/src/channelCollection.ts
```diff
@@ -924,6 +933,60 @@ export class ChannelCollection
 		}
 	}
 
+	public handleOp(message: unknown, messagesContent: unknown[], local: boolean): boolean {
```
anthony-murphy: i thought we were going to fix this so the input matched the original function to reduce casting

packages/runtime/container-runtime/src/channelCollection.ts
```diff
@@ -924,6 +933,60 @@ export class ChannelCollection
...
+		}
+		this.processMessages({
+			envelope: m,
+			messagesContent: messagesContent as IRuntimeMessagesContent[],
+			local,
+		});
+		return true;
+	}
+
+	public onConnectionStateChange(canSendOps: boolean, clientId: string | undefined): void {
```
anthony-murphy: should we just call this setConnectionState, as that is the exisitng pattern, so it would reduce churn, and wrappers

packages/runtime/container-runtime/src/channelCollection.ts
```diff
@@ -924,6 +933,60 @@ export class ChannelCollection
...
+	}
+
+	public contributeSummary(summaryTree: ISummaryTreeWithStats): void {
+		if (this.aliases.size > 0) {
+			addBlobToSummary(summaryTree, aliasBlobName, JSON.stringify([...this.aliases]));
+		}
+	}
+
+	public reSubmitOp(
+		message: unknown,
```
anthony-murphy: similar to handle op, can we make the types better to reduce casting?

packages/runtime/container-runtime/src/channelCollection.ts
```diff
@@ -924,6 +933,60 @@ export class ChannelCollection
...
+			m.type !== ContainerMessageType.Alias
+		) {
+			return false;
+		}
+		this.reSubmitContainerMessage(m, localOpMetadata, squash);
+		return true;
+	}
+
+	public rollbackStagedOp(message: unknown, localOpMetadata: unknown): boolean {
+		const m = message as LocalContainerRuntimeMessage;
```
anthony-murphy: again, more casting

packages/runtime/container-runtime/src/channelCollection.ts
```diff
@@ -1251,7 +1314,7 @@ export class ChannelCollection
 	 *
 	 * @param staging - A boolean indicating whether the container is in staging mode.
 	 */
-	public notifyStagingMode(staging: boolean): void {
+	public onStagingModeChange(staging: boolean): void {
```
anthony-murphy: i'd rather not rename as it just adds churn

packages/runtime/container-runtime/src/runtimeFeatureCollection.ts
```diff
@@ -0,0 +1,156 @@
...
+		return replacement;
+	}
+
+	public async onLoadFromSnapshot(): Promise<void> {
+		for (const f of this.features) {
+			await f.onLoadFromSnapshot?.();
+		}
+	}
+
+	public async onApplyStashedOps(seqNum: number): Promise<void> {
```
anthony-murphy: this doesn't look used. we should remove it

packages/runtime/container-runtime/src/runtimeFeatureCollection.ts
```diff
@@ -0,0 +1,188 @@
...
+	/**
+	 * Dispatch a stashed op to the feature that claims its type, returning
+	 * the result wrapper, or `undefined` if no feature claims the type.
+	 */
+	public async applyStashedOp(
+		opContents: LocalContainerRuntimeMessage,
+	): Promise<{ result: unknown } | undefined> {
+		const feature = this.opOwners.get(opContents.type as ContainerMessageType);
+		if (feature?.applyStashedOp === undefined) {
+			return undefined;
```
anthony-murphy, line 147: Deep Review: `RuntimeFeatureCollection.applyStashedOp` returns `undefined` for two distinct conditions: (a) no feature owns the op type, and (b) the owning feature has no `applyStashedOp`:

```ts
const feature = this.opOwners.get(opContents.type as ContainerMessageType);
if (feature?.applyStashedOp === undefined) { return undefined; }
return feature.applyStashedOp(opContents);
```

`PendingStateManager.dispatchStashedOp` then treats `undefined` as "unknown op type" and calls `this.stateHandler.closeFn(error); throw error;`. `chunkedOpsGuardFeature` (in `runtimeOpsFeature.ts`) claims `ContainerMessageType.ChunkedOp` via `supportedOps: CHUNKED_OPS` but provides only `handleOp` — so its claim is silently downgraded to "no handler" inside this collapse and the container closes. The sibling `rejoinFeature` in the same file does the right thing: it provides an explicit-throw `applyStashedOp`. The two guards in the same file diverge on this contract.

Well-formed local replay can't carry `ChunkedOp` (it's excluded from `LocalContainerRuntimeMessage` in `messageTypes.ts`), so the trigger requires corrupt or forward-incompatible pending JSON — but the asymmetry is real, and it is also present symmetrically in `handleOp` / `reSubmitOp` / `rollbackStagedOp`.

Suggestions, in order of effort:

1. **Smallest** — give `chunkedOpsGuardFeature.applyStashedOp` an explicit throw matching `rejoinFeature`. Closes the same-file divergence at minimum cost.
2. **More thorough** — distinguish the two return cases at the collection level, e.g. add a `hasOwner(type: ContainerMessageType): boolean` helper and let PSM check it before treating `undefined` as unknown, or switch to a tagged result `{ kind: "no-owner" | "no-apply" | "applied", result?: unknown }`. Apply uniformly across `handleOp` / `applyStashedOp` / `reSubmitOp` / `rollbackStagedOp` so the contract is consistent.

packages/runtime/container-runtime/src/runtimeFeatureCollection.ts
```diff
@@ -0,0 +1,188 @@
...
+		} else {
+			this.features.push(replacement);
+		}
+		this.registerOpClaims(replacement);
+		return replacement;
+	}
+
+	private registerOpClaims(feature: IRuntimeFeature): void {
+		if (feature.supportedOps === undefined) {
+			return;
```
anthony-murphy, line 74: Deep Review: The PR description says *"Op-routing dispatch order — feature registration order determines who claims each message type first… The channelCollection claim is intentionally before idCompressorFeature so FluidDataStoreOps short-circuit fastest."* That's not how this code works. `registerOpClaims` builds a `Map<ContainerMessageType, IRuntimeFeature>` and ``throw new Error(`RuntimeFeatureCollection: multiple features claim ${type}`)`` on a second claim — there is at most one owner per op type, and registration order is irrelevant for op routing. `channelCollection.supportedOps = [FluidDataStoreOp, Attach, Alias]` and `idCompressorFeature.supportedOps = [IdAllocation]` don't even overlap; even if they did the second `add()` would throw at construction time.

The in-source comment at the top of this file (the `Order matters for fan-out hooks` / `Op-routing dispatch is type-keyed and order-independent` block) already documents the correct invariant. Only the PR description disagrees.

Since this PR is `deep-review`-labelled, reviewers will read the metadata first and look for an ordering invariant that doesn't exist. Suggest editing the description to:

> **Op-routing dispatch order** — each `ContainerMessageType` has exactly one owning feature; duplicate claims throw at construction time (`registerOpClaims`). Registration order matters only for fan-out hooks (`onLoadFromSnapshot`, `contributeSummary`, `setConnectionState`, etc.), which iterate features in order.

packages/runtime/container-runtime/src/idCompressorFeature.ts
```diff
@@ -0,0 +1,221 @@
...
+				});
+			},
+		);
+	}
+
+	/**
+	 * Lazy load for "delayed" mode — finalizes any ranges that piled up in
+	 * {@link pendingOps} while the compressor was off.
+	 */
+	public loadDelayed(): void {
```
anthony-murphy, line 120: Deep Review: PR #20174 (vladsud, signed off by andre4i) deliberately authored the rule on schema transitions for the id compressor: *"this will only work for 'off' -> 'delayed' transitions. Anything else is too risky, and requires ability to initialize ID compressor synchronously!"* The warning still lives at `containerRuntime.ts` above the schema-transition decision point (`onSchemaChange`), but the actual loading site is now this `loadDelayed` method, which only documents *"Lazy load for 'delayed' mode — finalizes any ranges that piled up"* with no reference to the constraint.

It's true that `loadDelayed()` has multiple non-transition callers (load-time and connection-time) — but that is exactly why a future second transition caller could miss the invariant now that the implementation moved files.

Suggest copying a one-line warning into the docstring, e.g.:

> NOTE: only safe for `off → delayed` schema transitions; any other transition requires synchronous ID-compressor initialization (see `containerRuntime.onSchemaChange`).

That way the constraint travels with the implementation.

anthony-murphy:

- 🔄 **Mechanical Review**
  - ✅ mechanical (gpt-5.5)
  - 🔄 mechanical (claude-opus-4.7)
  - ✅ summarize (claude-opus-4.7)
  - 🔄 context (claude-opus-4.7)
  - 🔄 context (gpt-5.5)
  - ✅ summarize (claude-opus-4.7)
  - ✅ extract (claude-opus-4.7)
  - ✅ actual (claude-opus-4.7)
  - ✅ propose (claude-opus-4.7)
  - ✅ propose (gpt-5.5)
  - ✅ summarize (claude-opus-4.7)
- ✅ Adversarial Filter
- ✅ Synthesis
- ✅ Comments

packages/runtime/container-runtime/src/signalTelemetryProcessing.ts
```diff
@@ -79,6 +81,19 @@ export class SignalTelemetryManager {
 	 */
 	private broadcastSignalSequenceNumber: number = 0;
 
+	private wasCanSendOps = false;
```
anthony-murphy, line 84: Deep Review: The pre-PR runtime did `if (canSendOpsChanged) { this.signalTelemetryManager.resetTracking(); }` — reset on **both** `false→true` and `true→false` transitions. The new self-tracked behavior is `if (canSendOps && !this.wasCanSendOps) { this.resetTracking(); }` — only resets on `false→true`. The disconnect path no longer triggers a reset; any tracking state polluted while `canSendOps === false` is carried forward until the next connect.

The PR description flags this as *"behaviorally equivalent, but worth a check"* — it isn't strictly equivalent. Practical impact is narrow because telemetry is scoped via `message.clientId === this.clientId` and the next-send window is wiped before reuse, but the truth-table changed and nothing pins it: no test in `runtimeFeatureCollection.spec.ts` or `containerRuntime.spec.ts` asserts which `canSendOps` transitions trigger reset (read-only ↔ read-write while still connected, forced read-only toggle, reconnect during read-only).

Suggestions:

1. **Preferred** — restore byte-equivalence with `if (canSendOps !== this.wasCanSendOps) { this.resetTracking(); }`.
2. **Alternative** — confirm with telemetry owners (vladsud / ChumpChief, who debated `canSendOps` semantics in PR #16104) that "reset on reconnect only" is intentional.

In either case, add a focused unit test on `SignalTelemetryManager.setConnectionState` that asserts reset fires exactly on the chosen transition set and not on the others — covering the four `(wasCanSendOps, canSendOps)` combinations.

packages/runtime/container-runtime/src/runtimeFeatureCollection.ts
```diff
@@ -0,0 +1,177 @@
...
+
+	/**
+	 * Replace `oldFeature` (by reference) with `replacement`, preserving
+	 * registration order. Returns `replacement`. If `oldFeature` isn't
+	 * present, appends `replacement` instead.
+	 *
+	 * Primarily a test-fixture seam — production code should rarely need this.
+	 */
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	public replace<T extends IRuntimeFeature<any>>(oldFeature: T, replacement: T): T {
```
anthony-murphy, line 56: Deep Review: `replace<T>` swaps `this.features[index] = replacement`, deletes the old feature's op-claims, then `registerOpClaims(replacement)` and returns. `oldFeature.dispose?.()` is never called, even though `dispose` is part of `IRuntimeFeature` and the collection's own `dispose()` (lines 103-105) calls `f.dispose?.()` on each feature — so `replace` is the one lifecycle path that leaks.

Migrated subsystems implementing `dispose` include `BlobManager` (with internal emitters), `SummarizerSubsystem` (SummaryCollection / SummaryManager listeners), and `ChannelCollection`. The JSDoc says *"Primarily a test-fixture seam"* / *"production code should rarely need this"* — language that invites non-test use without forbidding it.

Suggestions:

1. **Preferred** — call `oldFeature.dispose?.()` inside `replace` to bring the contract in line with the rest of the collection's lifecycle treatment. One line, removes the footgun entirely.
2. **Alternative** — tighten the JSDoc to *"test-only — do not use in production"*, optionally rename to `replaceForTesting`, and assert/guard against production callers.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1931,6 +1877,7 @@ export class ContainerRuntime
 			submitMessage: (message: ContainerRuntimeGCMessage) => this.submit(message),
 			sessionExpiryTimerStarted: pendingRuntimeState?.sessionExpiryTimerStarted,
 		});
+		this.features.add(this.garbageCollector);
```
anthony-murphy, line 1913: Deep Review: The pre-PR `initializeBaseState` ran `await this.initializeSummarizer(loader); await this.garbageCollector.initializeBaseState();` — summarizer first, GC second. The new `features.onLoadFromSnapshot()` iterates in registration order, and `garbageCollector` is registered before `summarizerSubsystem`, so GC initializes first now. The PR description claims behavioral equivalence; on this hook that claim is false.

No concrete bug is triggered today: `SummarizerSubsystem.onLoadFromSnapshot` doesn't currently read GC base state — it builds election machinery from injected summary, delta-manager, quorum, loader, and runtime dependencies (see `summarizerSubsystem.ts:137-269`). But the ordering is observably inverted, and no test pins the registration order, so a future change could silently re-invert it.

This combines with a more general concern: the registration sequence in `ContainerRuntime` has no comment explaining why each feature is in its current slot, even though several fan-out hooks are order-sensitive (`SignalTelemetryManager`'s transition reset, `contributeSummary` summary-tree key ordering, GC vs. ChannelCollection on connect, GC vs. SummarizerSubsystem on load).

Suggestions:

1. Either register `summarizerSubsystem` before `garbageCollector` to preserve the prior load-order, or document explicitly that `onLoadFromSnapshot` ordering between these two is independent.
2. Add a comment at the registration site listing why order matters for each fan-out hook, plus a test that pins the registration order so refactors don't silently change `setConnectionState` / `contributeSummary` / `onLoadFromSnapshot` semantics.

packages/runtime/container-runtime/src/pendingStateManager.ts
```diff
@@ -133,13 +139,13 @@ export interface PendingBatchResubmitMetadata extends BatchResubmitInfo {
 export interface IRuntimeStateHandler {
```
anthony-murphy, line 139: Deep Review: `IRuntimeStateHandler` gains `closeFn`; `dispatchStashedOp` now calls `this.stateHandler.closeFn(error); throw error;` when no feature claims the op type. Pre-PR this lived in `ContainerRuntime.applyStashedOp`'s `default` branch — PSM is now both dispatcher and closer for an entire class of corruption-grade errors.

No demonstrated defect: the coupling is narrow (`Pick<RuntimeFeatureCollection, "applyStashedOp">`), PSM already owned corruption-detection semantics for forked-container / mismatched pending-message conditions, and existing tests cover stashed-op application, empty-batch handling, replay batching, batchId persistence, staged batches, and local-op metadata preservation. The concern is coverage proportional to the historical pattern: PSM has been a dense locus of forked-container / batch-id correctness work (#21714, #21767, #22310, #19802), and large mechanical moves through `containerRuntime.ts` have repeatedly required follow-up PRs.

Suggestion: add a unit/integration test tracing one local/stashed batch through `PendingStateManager.dispatchStashedOp → RuntimeFeatureCollection.applyStashedOp → features.reSubmitOp → duplicate-batch summary contribution`. Confirm batch boundaries, opMetadata/localOpMetadata separation, and batchId persistence are all preserved across the four-hook re-routing, and that `closeFn` here doesn't race with the existing `disposeOnce` path. Worth tagging markfields on the `dispatchStashedOp` change.

anthony-murphy:
## Deep Review

*Reviewed commit `c329522` on 2026-04-28.*

The `IRuntimeFeature` / `RuntimeFeatureCollection` extraction is structurally sound: an independent design pass converged on this PR's shape, no Tier 1 or Tier 2 issue survives deeper review, and in-package mocha is reported green. What remains is a cluster of Tier-3 framework-hardening, ordering-brittleness, and test-coverage items. Holding at 7/10 (same band) rather than tracking up to 8 because the actionable list grew rather than shrank — three new code-anchored polish items surfaced this round (summary-tree key reorder, duplicate-instance `add`, `InboundBatchAggregator.currentClientId` lag) on top of the seven previously-posted threads, and the previously-posted `idCompressorFeature.ts` thread is partially overtaken by evidence (the actual code already cross-references the `off → delayed` constraint).

### Findings

- **`chunkedOpsGuardFeature` claims `ChunkedOp` but omits `applyStashedOp`/`reSubmitOp`/`rollbackStagedOp`; collection collapses "no owner" and "owner without handler" so PSM closes the container with a misleading diagnostic in both** — `runtimeFeatureCollection.ts:147` — (inline thread)
- **PR description's "registration order determines who claims each message type first" contradicts the registry invariant** — `runtimeFeatureCollection.ts:74` — (inline thread)
- **Lifecycle ordering inversions vs. legacy: `dispose` (summarizer last) and `onLoadFromSnapshot` (GC before summarizer)** — `containerRuntime.ts:1881` — (inline thread; dispose-half addendum posted as reply)
- **Top-level summary-tree blob/tree key ordering shifted vs. legacy `addContainerStateToSummary`** — `runtimeFeatureCollection.ts` (contributeSummary) — (inline thread)
- **No completeness assertion and no runtime-level ordering pin — every fan-out hook's order is implicit in `add()` call sequence** — `runtimeFeatureCollection.ts` (registerOpClaims) — (inline thread)
- **`RuntimeFeatureCollection.add` silently accepts duplicate registration of the same feature instance** — `runtimeFeatureCollection.ts` (add) — (inline thread)
- **`RuntimeFeatureCollection.replace` does not dispose the replaced feature** — `runtimeFeatureCollection.ts:56` — (inline thread)
- **`SignalTelemetryManager.setConnectionState` resets on a different transition set than the legacy explicit call, and no test pins the truth-table** — `signalTelemetryProcessing.ts:84` — (inline thread)
- **PSM gains direct `Pick<RuntimeFeatureCollection, "applyStashedOp">` coupling and owns the `closeFn` close-on-unknown path; no end-to-end coverage tracing one stashed batch through the new dispatch** — `pendingStateManager.ts:139` — (inline thread)
- **`InboundBatchAggregator.currentClientId` lags runtime `clientId` until first `setConnectionState` callback** — `inboundBatchAggregator.ts` — (inline thread)
- **`isContainerMessageDirtyable` and `canStageMessageOfType` remain as parallel `switch` statements outside the feature contract** — `containerRuntime.ts` / `messageTypes.ts` — Adding any new `ContainerMessageType` still requires touching the dirtyable switch plus the feature claim, partially undercutting the "single owning feature per op type" property the framework sells. `canStageMessageOfType` is a staging-eligibility policy (legitimately a subset of owned types), but `isContainerMessageDirtyable` is per-type and arguably belongs on the owning feature as an optional `isDirtyable?(msg)` predicate. Follow-up scope, not blocking.
- **`IdCompressorFeature.loadDelayed` `off → delayed` constraint — previously-posted concern partially overtaken by evidence** — `idCompressorFeature.ts:120` — Re-reading the actual code: `loadDelayed` already documents that lazy load is only safe for `off → delayed`, that other transitions must not call it, and names `containerRuntime.onSchemaChange` plus PR #20174 as the decision point. The constraint cross-reference is already present. The thread can be marked resolved once the author confirms this reading; optionally tighten the `containerRuntime.onSchemaChange` cross-link to name `loadDelayed` explicitly.

### Path to Ready

To move from 7/10 → 8/10 (resolve the existing thread cluster):

- [ ] **Resolve the `chunkedOpsGuardFeature` asymmetry** — explicit-throw stubs matching `rejoinFeature` (ideally `assert(false, 0x93d)` + `closeFn`), or distinguish "no owner" from "owner without handler" at the collection level uniformly across all four hooks. Existing thread.
- [ ] **Fix the PR-description ordering claim** per the suggested edit below. Existing thread.
- [ ] **Decide and pin the `SignalTelemetryManager` reset transition** — either restore `if (canSendOps !== this.wasCanSendOps)` (symmetric reset) or confirm with telemetry owners that "reset on reconnect only" is intentional. Either way, add a focused unit test asserting reset fires exactly on the chosen transition set. Existing thread.
- [ ] **Decide on `RuntimeFeatureCollection.replace` lifecycle** — call `oldFeature.dispose?.()` (preferred — one line) or rename to `replaceForTesting` and tighten the JSDoc to test-only. Existing thread.
- [ ] **Fix or document the GC-vs-Summarizer load-order inversion AND the new dispose-order inversion (summarizer last)** — re-register `summarizerSubsystem` to preserve the prior sequence, or add a comment at the registration site listing why each feature is in its current slot. Existing thread (with new dispose-half addendum).
- [ ] **Add an end-to-end test tracing one stashed batch** through `PendingStateManager.dispatchStashedOp → features.applyStashedOp → features.reSubmitOp → duplicate-batch summary contribution`. Existing thread.

To move from 8/10 → 9-10/10 (close the systemic gaps surfaced this round):

- [ ] **Add a completeness-and-ordering test** — iterate `Object.values(ContainerMessageType)` to assert every type has an owner, plus a registration-order snapshot pinning fan-out call sequence. Closes the systemic gap behind the load-order, summary-tree, and duplicate-instance findings.
- [ ] **Decide the top-level summary-tree key order** — either restore the legacy sequence by reordering `add()` calls, or document the new canonical order in `RuntimeFeatureCollection.contributeSummary` and add a snapshot test pinning the top-level key sequence. A quick grep across storage drivers / snapshot-diff tests would settle whether anything downstream depends on the historical order.
- [ ] **One-line guard against duplicate same-instance `add`** — drop the `existing !== feature` short-circuit or throw if `this.features.includes(feature)` before registering op claims.
- [ ] **Initialize `InboundBatchAggregator.currentClientId` from runtime at construction** or restore the live `getClientId` callback so the value is always live.
- [ ] **File follow-up issues** — for routing `isContainerMessageDirtyable` through the feature contract, and for the deferred `IRuntimeContext` / runtime-subsystem-host idea referenced in "Alternatives considered".
- [ ] **Get sign-off from the domain experts whose prior decisions this PR touches** — required: vladsud (schema/telemetry/canSendOps), markfields (PSM/offline), kian-thompson (stashed-op dispatch); optional: noencke (typing), andre4i (channelCollection), ChumpChief (canSendOps), dannimad (resubmit).

### Suggested description edit

Replace:

> Op-routing dispatch order — feature registration order determines who claims each message type first. Currently: pendingStateManager → garbageCollector → channelCollection → … The channelCollection claim is intentionally before idCompressorFeature so FluidDataStoreOps short-circuit fastest.

with:

> **Op-routing dispatch order** — each `ContainerMessageType` has exactly one owning feature; duplicate claims throw at construction time (`registerOpClaims`). Registration order matters only for fan-out hooks (`onLoadFromSnapshot`, `contributeSummary`, `setConnectionState`, `dispose`), which iterate features in order.

Also consider adding a brief note that `dispose` and `onLoadFromSnapshot` ordering shifted vs. the prior explicit sequence — either with the rationale, or with a pointer to a forthcoming registration-order pin test.

### Context for Reviewers

- **Suggested reviewers — domain experts whose prior decisions this PR touches**: **vladsud** (PR #20174 DocumentSchemaController + id-compressor schema-transition rule; PR #16104 `canSendOps` semantics relevant to the SignalTelemetryManager finding); **markfields** (PRs #21714, #21767, #22310 Offline / batchId / CSN / DuplicateBatchDetector — three subsystems migrated to features here, plus reviewer of #21729); **noencke** (PR #17285 strict message types — the `as ContainerMessageType` casts inside `runtimeFeatureCollection.ts` quietly weaken the typing #17285 established); **kian-thompson** (PRs #14462, #15944 stashed-op format / batch boundaries on resubmit — directly relevant to the chunkedOpsGuard and PSM-coupling findings); **ChumpChief** (co-owns the `canSendOps` semantics discussion in PR #16104); **andre4i** (PRs #19863, #19968 DataStores / channelCollection; reviewer of #20174's `DocumentSchemaChange` resubmit semantics); **dannimad** (PR #21854 empty group batch on resubmit — touches the resubmit path now reshaped). Required: vladsud, markfields, kian-thompson. Optional: noencke, andre4i, ChumpChief, dannimad.
- **Independent design pass converged on this PR's shape** — a separate review-time design exploration arrived at essentially the same typed runtime-subsystem contract: declared owned message types, single-owner enforcement at registration, generic-narrowed `message`/`contents`, optional lifecycle hooks, and table-driven dispatch replacing the four switches. Beyond the feature framework, the PR also extracts `ReconnectTracker`, `LoadingGroupSnapshotFetcher`, and `ExtensionsManager` as non-feature helpers — sensible monolith-reduction wins not anticipated by the design pass.
- **`containerRuntime.ts` is the single highest-churn file in the area** — 33 commits in 6 months across 17 PRs, all *targeted* feature changes; this PR is the first structural extraction at this scale (net −682 lines on `containerRuntime.ts`, ~+1700 across new feature files). The `IRuntimeFeature` / `RuntimeFeatureCollection` pattern has no precedent in the historical record. Bugs introduced here will be hard to localize via blame after merge because most lines are *moved*, not authored. Historical pattern: large mechanical moves in this file have repeatedly required follow-up PRs (#19802 → #21266 → #21714 → #21767 → #22310). Suggest keeping the draft open through at least one full rotation of the offline/forked-container e2e tests.
- **Generic typing of `IRuntimeFeature<TOps>` extends PR #17285's direction; new package-surface exports.** PR #17285 introduced `ContainerRuntimeMessage<TType, TContents>`; this PR layers `IRuntimeFeature<TOps extends AnyRuntimeOpType>` plus `InboundRuntimeMessageFor<TOps>` / `LocalRuntimeMessageFor<TOps>` / `RuntimeMessagesContentFor<TOps>` on top. Worth a glance from noencke / jason-ha — and review the `index.ts` additions explicitly as public/package surface (keep only exports needed by shipping consumers or existing tests).
- **`DocumentSchemaController` and `IdCompressorFeature` resubmit rationales.** PR #20174 confirmed schema-change ops are not directly resubmitted (CAS semantics); the new `DocumentsSchemaController.reSubmitOp` calls `pendingOpNotAcked()` and `applyStashedOp` claims-and-drops, preserving that contract. `IdCompressorFeature.reSubmitOp` is a silent no-op — the 10-line worked-example comment from the old code (out-of-order risk; runtime submits a fresh range covering all pending IDs before replay) was trimmed to 2 lines. The invariant is preserved but the rationale is gone — optional polish: keep one sentence pointing at the runtime-side counterpart.
- **`canSendOps` semantics history** (pairs with the SignalTelemetryManager finding): PR #16104 review thread (vladsud / ChumpChief, ~6 round-trips) established that `canSendOps` is true on read connections too. The new self-tracked transition `wasCanSendOps → canSendOps` will reset on every `false → true` of `canSendOps`, including read → write upgrade within a single session, where the old explicit call inside `setConnectionStateCore` may have been called on different boundaries.

- `6a7679f` 2026-04-28 · **7/10** — seven threads still apply; three new code-anchored polish items surfaced.
- `343e6db` 2026-04-28 · **7/10** — refactor structurally sound; five Tier-3 polish items, no Tier 1/2.

</details>

packages/runtime/container-runtime/src/runtimeFeatureCollection.ts
```diff
@@ -0,0 +1,177 @@
...
+		for (const f of this.features) {
+			f.notifyStagingMode?.(active);
+		}
+	}
+
+	public dispose(): void {
+		for (const f of this.features) {
+			f.dispose?.();
+		}
+	}
```
anthony-murphy, line 107: Deep Review: The top-level summary-tree blob/tree key ordering shifted vs. the deleted `addContainerStateToSummary`.

Old (deleted at pr.diff lines ~1147-1190) added in fixed order: `idCompressor` → `chunks` → `recentBatchInfo` → `aliases` → `electedSummarizer` → `blobs` tree → `gc` tree.

New `features.contributeSummary` iterates registration order: `chunks` (RemoteMessageProcessor) → `recentBatchInfo` (DuplicateBatchDetector) → `gc` tree (GarbageCollector) → `aliases` (ChannelCollection) → `blobs` tree (BlobManager) → `idCompressor` (IdCompressorFeature) → `electedSummarizer` (SummarizerSubsystem).

Looked into whether this is a wire-format break: PR #17285's *"set the properties in the order they appear here … stringified values can be compared"* invariant applies to `TypedContainerRuntimeMessage` and PSM pending-message stringification (`messageTypes.ts:64-82`, `pendingStateManager.ts:156-162`), not to summary-tree key order; the upload site (`containerRuntime.ts:4057-4060`) sends `summaryTree` as a named-entry object. So no demonstrated break — but the ordering observably changed without being called out, and no test pins it.

Suggestions:

1. **Preferred** — reorder feature registration in `containerRuntime.ts` to match the previous summary-tree key sequence; the registration-order pin test (separate thread) then guards against future drift.
2. **Alternative** — document the new canonical order in `RuntimeFeatureCollection.contributeSummary` and add a snapshot test pinning the top-level summary-tree key sequence.

A quick grep across storage drivers / snapshot-diff tests outside `container-runtime` would settle whether anything downstream relies on the historical order.

packages/runtime/container-runtime/src/runtimeFeatureCollection.ts
```diff
@@ -0,0 +1,177 @@
...
+		if (feature.supportedOps === undefined) {
+			return;
+		}
+		for (const type of feature.supportedOps) {
+			const existing = this.opOwners.get(type);
+			if (existing !== undefined && existing !== feature) {
+				throw new Error(`RuntimeFeatureCollection: multiple features claim ${type}`);
+			}
+			this.opOwners.set(type, feature);
+		}
```
anthony-murphy, line 82: Deep Review: Two cheap test-only additions would close a class of risk that this round's findings keep surfacing — the `dispose` and `onLoadFromSnapshot` order inversions, the summary-tree key reorder, the duplicate-instance `add` gap, and the `chunkedOpsGuardFeature` completeness gap are all instances where `add()` order or registration completeness silently shifted observable behavior.

1. **Completeness assertion** — iterate `Object.values(ContainerMessageType)` in a freshly-constructed `ContainerRuntime`'s feature collection and assert each has an owner. Forces the registry to be the single source of truth for op-type ownership. Could also tighten `registerOpClaims` to require each owning feature to implement at least one op-routing hook for each claimed type (i.e. surface the chunkedOpsGuard "claims but no `applyStashedOp`" case at construction).
2. **Ordering pin** — assert the registration array (or each fan-out hook's call order) matches a hand-curated expected sequence. So that any future re-register that silently reorders `dispose` / `contributeSummary` / `onLoadFromSnapshot` / `setConnectionState` surfaces in CI.

The existing `runtimeFeatureCollection.spec.ts` covers fan-out order on the abstract collection (e.g. `:13-33`, `:90-105`, `:166-184`) but no test pins the **`ContainerRuntime`** registration array. That's the gap.

packages/runtime/container-runtime/src/runtimeFeatureCollection.ts
```diff
@@ -0,0 +1,177 @@
...
+ * rollbackStagedOp) dispatch via a single
+ * `Map<ContainerMessageType, IRuntimeFeature>` built at registration time.
+ * Each op type has at most one owning feature.
+ *
+ * @internal
+ */
+export class RuntimeFeatureCollection {
+	private readonly features: IRuntimeFeature[] = [];
+
+	private readonly opOwners = new Map<AnyRuntimeOpType, IRuntimeFeature>();
```
anthony-murphy, line 30: Deep Review: `add()` calls `registerOpClaims(feature)`, and `registerOpClaims` only throws when `existing !== undefined && existing !== feature`. The `existing !== feature` short-circuit silently accepts a re-add of the **same instance** — the feature ends up in the `this.features` array twice, so all fan-out hooks (lifecycle, `contributeSummary`, `setConnectionState`, `dispose`) fire twice for that subsystem.

PR #17285 (noencke) and PR #20174 (vladsud) established exactly-once ownership of message types and exactly-once schema/blob writes; a duplicated feature would silently double-write summary blobs and double-fire connection-state callbacks. No current caller exercises this, but the contract gap is real.

Suggestion — one-line guard in `add()`:

```ts
if (this.features.includes(feature)) {
    throw new Error(`RuntimeFeatureCollection: feature already registered`);
}
```

or drop the `existing !== feature` short-circuit in `registerOpClaims` and require callers to use `replace()` for the legitimate re-registration case.

packages/runtime/container-runtime/src/inboundBatchAggregator.ts
```diff
@@ -28,16 +29,16 @@ type IRuntimeMessageMetadata =
  * This class ensures that we aggregate a complete batch of incoming ops before processing them. It basically ensures
  * that we never start processing ops in a batch IF we do not have all ops in the batch.
  */
-export class InboundBatchAggregator {
+export class InboundBatchAggregator implements IRuntimeFeature<never> {
```
anthony-murphy, line 32: Deep Review: The diff replaces a live callback (`getClientId: () => this.clientId`) with a cached field `currentClientId` that's updated only by `setConnectionState`. Previously the `localBatch` flag was computed against the live `clientId` on every comparison; now any inbound op processed before the first `setConnectionState` callback compares against `undefined`.

Narrowing impact: `currentClientId` is read only in error-detail objects (`localBatch` for the system-message-during-batch error, `localBatch`/`localMessage` for interleaved-client batch corruption errors), so it's telemetry-on-error-paths only. But it's an unnecessary semantic change in a "behavior-preserving" refactor — and on the rare path where one of these errors fires before the first `setConnectionState` callback, the telemetry will misreport `localBatch`/`localMessage`.

Suggestions:

1. **Smallest** — initialize `currentClientId` from the runtime at construction so the first read is never `undefined`.
2. **Cleanest** — retain the original `getClientId` callback (or expose `clientId` through a runtime-context object) so the value is always live and there's no caching-sync hazard.

anthony-murphy:
## Deep Review

*Reviewed commit `6a013ab` on 2026-05-05.*

A disciplined refactor: 12 subsystems extracted from `ContainerRuntime` onto a new `IRuntimeFeature` framework, in-package mocha green, and an independent design pass converged on essentially the same shape. No Tier 1/2 issues survive analysis. Remaining items are doc/test polish on top of the existing inline thread cluster, plus two behavior diffs the PR description already calls out (`SignalTelemetryManager` reset and summary-tree blob iteration order) that are worth pinning explicitly before promotion out of draft.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add a *Known behavior diffs (telemetry-only)* section to the PR description covering (a) `SignalTelemetryManager` reset now firing only on `false → true` `canSendOps` (was: any `canSendOps` change while attached; detached-reset precluded by the existing `setConnectionStateCore` assert) and (b) summary-tree blob iteration order moving from fixed (`metadata → idCompressor → chunks → recentBatchInfo → aliases → electedSummarizer → blobs → gc`) to feature-registration order (`metadata → chunks → recentBatchInfo → gc → aliases → blobs → idCompressor → electedSummarizer`) — `ISummaryTreeWithStats` is dictionary-keyed so tree hashes are unaffected, but ordered-iteration consumers (snapshot diff harnesses, stringify-based telemetry) would observe a delta.
- [ ] Add a comment block at the feature-registration site naming the cross-feature ordering constraints (specifically `DocumentsSchemaController` precedes `IdCompressorFeature` for `onLoadFromSnapshot`/`contributeSummary`, since schema must be applied before id-compressor observes `off → delayed`).
- [ ] Add a one- or two-line comment beside `IRuntimeStateHandler.closeFn` (or beside `dispatchStashedOp`'s `closeFn(error); throw error;` site) noting alignment with the existing close-then-throw pattern at `containerRuntime.ts:2956-2968` / `:4511-4522`. PR #19802's detached invariant is preserved by `applyStashedOpsAt`'s `isAttached()` gate; the layering rationale is currently undocumented.
- [ ] Add a construction-time exhaustiveness assertion verifying every `ContainerMessageType` enum value is present in `opOwners` (mirrors the existing duplicate-claim throw in `registerOpClaims`).
- [ ] Add (or link from a comment in `IdCompressorFeature`) a stash → rehydrate → resubmit test driving a mixed `FluidDataStoreOp` + `IdAllocation` batch — assert exactly one fresh range is generated by `generateAllocationOp` and no `IdAllocation` op is double-submitted. The no-op `reSubmitOp`/`applyStashedOp` contract is load-bearing for offline/forked-container scenarios (PR #21767) and currently unpinned at the new feature-collection boundary.

### Context for Reviewers

- **Highest-churn region in the runtime — bugs introduced here will be hard to localize via blame.** `containerRuntime.ts` shows 30 commits in 6mo across 18 PRs; the `containerRuntime.ts` hunk (+235/-921) touches code introduced or rewritten by nearly all of the offline/forked-container work merged in 2024 (#21714, #21767, #21854, #22310, #22454). Most lines here are *moved*, not authored.
- **Domain experts whose prior decisions this PR re-homes** — request reviews from **markfields** (PSM / RemoteMessageProcessor / DuplicateBatchDetector / Outbox-resubmit; #21714, #21767, #22310, #22454), **vladsud** (DocumentsSchemaController feature + schema-change resubmit semantics; #20174, and `canSendOps` semantics in #16104), **noencke** (op-type generics on `IRuntimeFeature`; #17285); courtesy ping **tyler-cai-microsoft** on offline back-compat for #21767's `enableOfflineLoad` gate now reading through `RuntimeFeatureCollection.reSubmitOp`, and **andre4i** on schema-change documentation location.
- **Loader/runtime back-compat surface to verify** — `enableOfflineLoad` is the single gate for all offline scenarios (per #21767 review thread). PSM, RemoteMessageProcessor, DuplicateBatchDetector, and resubmit logic all moved onto features; `IRuntimeStateHandler` adds `closeFn` and removes `applyStashedOp`. Specifically check (1) old stashed-state formats still parse through the new `dispatchStashedOp` path; (2) `enableOfflineLoad`-gated behavior in resubmit is preserved through `RuntimeFeatureCollection.reSubmitOp`; (3) the `IRuntimeStateHandler` interface change is internal-only (no external implementers).
- **`IRuntimeFeature` is a new exported surface** (`src/index.ts` adds `IRuntimeFeature` and helper types). #10067's rollback-infrastructure review pushed back on exposing new runtime extension shape before it was fully supported. Confirm the export is intentionally `@internal` and downstream consumers are not being offered a premature extension API. The added `RuntimeFeatureCollection` unit tests align with the historical request for test seams.

- **markfields sign-off** on `PendingStateManager.dispatchStashedOp` calling `closeFn` — author of #19802, owner of offline/forked-container infrastructure. Is the close-then-throw pattern at the PSM boundary acceptable given the #19802 detached invariant is preserved by `applyStashedOpsAt`'s `isAttached()` gate?
- **vladsud sign-off** on `DocumentsSchemaController implements IRuntimeFeature<ContainerMessageType.DocumentSchemaChange>` and the resubmit/applyStashed semantics (preserved verbatim, but now living in `documentSchema.ts` rather than the runtime resubmit switch). Is the per-feature documentation location sufficient for the PR #20174 commitment?
- **vladsud sign-off** on `IdCompressorFeature` ownership of the `off → delayed` transition, the `onLoadFromSnapshot` ordering relative to `DocumentsSchemaController`, and the no-op `reSubmitOp` contract.
- **noencke sanity-check** on `InboundRuntimeMessageFor` / `LocalRuntimeMessageFor` / `RuntimeMessagesContentFor` and the unavoidable widening to `IRuntimeMessagesContent[]` at the collection-dispatch boundary — does the new helper layer regress the typed-boundary work from PR #17285?
- **Design taste call** — single combined `IRuntimeFeature` vs. split message+lifecycle contracts, and direct subsystem implementation vs. adapter classes. Both independent design proposals preferred the alternatives; the PR's choices are defensible. Reviewer judgment, not a code-fix item.
- **Load-latency check** — the lifecycle fan-out (`for (const f of this.features) { await f.onLoadFromSnapshot?.(); }`) is sequential; verify nothing in the prior implementation ran in parallel and would regress load latency.

</details>

anthony-murphy:
## Deep Review

*Reviewed commit `bb86e39` on 2026-06-02.*

Not ready for sign-off. A disciplined 12-subsystem extraction onto the new `IRuntimeFeature` framework, but a silent 5× cache-TTL bump (60s → 5min) landed in `LoadingGroupSnapshotFetcher` during what the PR description bills as pure monolith reduction — flagged inline. Remaining items are the doc/test/polish backlog from the prior review, several still un-actioned.

### Path to Ready

- [ ] Resolve inline threads (covers the `LoadingGroupSnapshotFetcher` TTL regression and the prior author self-review threads from 2026-04-28).
- [ ] Add a *Known behavior diffs* section to the PR description covering (a) the `LoadingGroupSnapshotFetcher` TTL change — if intentional, cite recorded-usage data; otherwise revert; (b) `SignalTelemetryManager` reset now firing only on `false → true` `canSendOps`; (c) summary-tree blob iteration order shifting from fixed sequence to feature-registration order (dict-keyed, so tree hashes unaffected, but stringify-based consumers will see a delta); (d) lifecycle fan-out ordering inversions on `setConnectionState` (GC↔ChannelCollection), `initializeBaseState`, and `dispose` (Summarizer last rather than first); (e) the op-routing exhaustiveness gap — `registerOpClaims` enforces duplicate-owner failure but not coverage.
- [ ] Replace the plain `throw new Error(...)` in `runtimeOpsFeature.ts` `chunkedOpsGuardFeature` with the tagged `assert(false, 0x93d /* ChunkedOp should not reach the feature dispatch path */)` so telemetry classification is preserved.
- [ ] Add a construction-time exhaustiveness assertion in `RuntimeFeatureCollection.registerOpClaims` verifying every `ContainerMessageType` enum value is covered (mirrors the existing duplicate-claim throw). *Carried from prior Path to Ready — still open.*
- [ ] Add a runtime check in `registerOpClaims` requiring `handleOp` to be present whenever `supportedOps` is non-empty (closes the optional-hook footgun the PR's own risks section calls out).
- [ ] Add a one- or two-line comment beside `IRuntimeStateHandler.closeFn` (or `dispatchStashedOp`'s `closeFn(error); throw error;` site) noting alignment with the existing close-then-throw pattern and the PSM layering rationale. *Carried from prior Path to Ready — still open.*
- [ ] Add a comment block at the feature-registration site naming cross-feature ordering constraints (`DocumentsSchemaController` precedes `IdCompressorFeature` for `onLoadFromSnapshot`/`contributeSummary`) and documenting the inverted orderings on `setConnectionState`/`initializeBaseState`/`dispose`. *Carried from prior Path to Ready — still open.*
- [ ] Add a targeted spec for `SignalTelemetryManager` covering `false → true`, `true → true`, and `true → false → true` `canSendOps` transitions to pin the author-flagged reset semantics.
- [ ] Add (or link from) a stash → rehydrate → resubmit test driving a mixed `FluidDataStoreOp` + `IdAllocation` batch — assert exactly one fresh range is generated by `generateAllocationOp` and no `IdAllocation` op is double-submitted. *Carried from prior Path to Ready — still open.*
- [ ] Add a test exercising `GarbageCollector.rollbackStagedOp` through the new `RuntimeFeatureCollection` dispatch path so the `TombstoneLoaded` rollback is pinned end-to-end (`runtimeFeatureCollection.spec.ts` currently covers `handleOp` routing and duplicate claims but not rollback).

### Context for Reviewers

- **Highest-churn region in the runtime.** `containerRuntime.ts` shows churn=34 in 6mo across 17+ PRs from 9 distinct authors; this PR alone is +235/-921 there. Offline/reconnect, resubmit ordering, batch-metadata preservation, and stashed-op replay are the load-bearing validation. Rebase conflicts are virtually guaranteed.
- **`IRuntimeFeature` generics extend PR #17285's discriminated-union work.** New helpers `InboundRuntimeMessageFor<TOps>`, `LocalRuntimeMessageFor<TOps>`, `RuntimeMessagesContentFor<TOps>` narrow over `ContainerRuntimeMessage<TType, TContents>`. `noencke` and `jason-ha` are best placed to confirm each migrated handler preserves prior unknown-message and compatibility behavior.
- **`DocumentSchema` CAS-on-resubmit + `IdCompressor` activation rules preserved verbatim.** `DocumentsSchemaController.reSubmitOp()` still calls `pendingOpNotAcked()` instead of resubmitting (PR #20174 contract). `vladsud` should specifically verify the resubmit path still routes through `pendingOpNotAcked` for every prior code path (including staging/rollback), and that `IdCompressor` extraction hasn't widened the historically narrow activation rules.
- **Summary contribution is now centralized.** `RuntimeFeatureCollection.contributeSummary` fans out across `BlobManager`, `DuplicateBatchDetector`, `RemoteMessageProcessor` partial chunks, `ChannelCollection` aliases, GC, ID compressor, document schema, and summarizer election state. Tree-blob iteration order shifts from fixed sequence to feature-registration order — `ISummaryTreeWithStats` is dict-keyed so hashes unaffected, but ordered-iteration consumers (snapshot diff harnesses, stringify-based telemetry) would observe a delta.
- **This author's "pure refactor" PRs have a history of subtle path/type semantics changes** (PR #21729 drew substantive `markfields` review comments on hidden behavior changes). The `LoadingGroupSnapshotFetcher` TTL bump flagged inline is exactly that pattern recurring; reviewers (especially `markfields`) should diff-read each extracted feature for restriction/typing tightening beyond what the description claims.
- **Suggested required reviewers:** `markfields` (offline/batchId/`DuplicateBatchDetector`; #21714, #21767, #22310, #22454; heaviest reviewer of `containerRuntime.ts` and of the author's prior #21729 refactor), `vladsud` (owner of `summary/documentSchema.ts` #20174 and runtime feature/compat + ID-compressor activation rules), `noencke` (discriminated-union container-runtime message-type system #17285). *Strongly recommended:* `kian-thompson` (last touched `applyStashedOp` contract, #14462), `agarwal-navin` (GC op author, #18632), `jason-ha` (pushed for tighter generic constraints during #17285 review).

- **`markfields` sign-off** on `PendingStateManager.dispatchStashedOp` calling `closeFn` directly — owner of offline/forked-container infrastructure (#19802); is the close-then-throw pattern at the PSM boundary acceptable given the detached invariant is preserved by `applyStashedOpsAt`'s `isAttached()` gate?
- **`vladsud` sign-off** on `DocumentsSchemaController implements IRuntimeFeature<DocumentSchemaChange>` and the resubmit/applyStashed semantics — is the per-feature documentation location sufficient for the PR #20174 commitment?
- **`vladsud` sign-off** on `IdCompressorFeature` ownership of the `off → delayed` transition, its `onLoadFromSnapshot` ordering relative to `DocumentsSchemaController`, and the no-op `reSubmitOp` contract.
- **`noencke` sanity-check** on `InboundRuntimeMessageFor` / `LocalRuntimeMessageFor` / `RuntimeMessagesContentFor` and the unavoidable widening to `IRuntimeMessagesContent[]` at the collection-dispatch boundary — does the new helper layer regress the typed-boundary work from PR #17285?
- **Design taste call** — single combined `IRuntimeFeature` vs split message+lifecycle contracts, and direct subsystem implementation vs adapter classes. Both independent design proposals preferred the split alternatives; the PR's "single all-optional interface" choice is defensible but worth a reviewer call.
- **Load-latency check** — the lifecycle fan-out (`for (const f of this.features) { await f.onLoadFromSnapshot?.(); }`) is sequential; verify nothing in the prior implementation ran in parallel and would regress load latency.
- **`@internal` confirmation** — `IRuntimeFeature` and helper types are newly exported from `src/index.ts`; confirm they are intentionally `@internal` and not being offered as a premature extension API.

</details>

- `6a013ab` 2026-05-05 · **7/10** — A disciplined refactor: 12 subsystems extracted from `ContainerRuntime` onto a new `IRuntimeFeature` framework

</details>

packages/runtime/container-runtime/src/loadingGroupSnapshotFetcher.ts
```diff
@@ -0,0 +1,154 @@
...
+ * Caches per-loading-group snapshots and waits for the delta manager to
+ * catch up if the fetched snapshot is ahead of the runtime's current
+ * sequence number. Extracted from ContainerRuntime so the runtime stays
+ * focused on the live runtime pipeline.
+ *
+ * @internal
+ */
+export class LoadingGroupSnapshotFetcher {
+	private readonly cache = new PromiseCache<string, ISnapshot>({
+		expiry: { policy: "absolute", durationMs: 5 * 60 * 1000 },
```
anthony-murphy, line 42: Deep Review: This silently extends the per-loadingGroupId snapshot cache TTL from **60s to 5 minutes** (5×). The pre-PR `ContainerRuntime.snapshotCacheForLoadingGroupIds` (deleted in this PR) was hard-coded to `durationMs: 60000` with the explicit comment *"Add expiry policy of 1 minute. Starting with 1 min and based on recorded usage we can tweak it later on."* The PR description frames `LoadingGroupSnapshotFetcher` as a pure monolith-reduction helper and does not call out any cache-policy change. Consumers requesting the same `loadingGroupIds` within the 1–5 minute window now get a stale cached `ISnapshot` instead of refetching, lagging the server by up to 5× the prior bound.

Restore `durationMs: 60_000` to preserve existing behavior. If the 5-minute value is intentional, split the TTL bump into a separate PR with cited recorded-usage data (per the comment the original author left inviting future adjustment) and call it out in the description.

```suggestion
            expiry: { policy: "absolute", durationMs: 60_000 },
```

PR #27186, opened 2026-04-28
feat: Add lockfiles for compat test dependencies

Abe27342:
Updates test-version-utils so that necessary dependencies are installed with a lockfile.

This is done by using a nested pnpm workspace with all necessary dependencies. Installing in this nested workspace is enforced via a postinstall script in test-version-utils. Each prior FF version that's used in some compat test has its own package.json as a package within this workspace. These package.json files as well as a versions.cjs file are managed via an update script that will need to be run with each minor release.

Some tests that we have also install explicit versions outside of the compat matrix. Authors of such tests must now add these dependencies to a file adjacent to versions.cjs.

The motivating factor for these changes is to remove all remaining places where we perform package installations without lock files in CI flows.

See #27186 for an earlier approach that kept installing packages lazily at test run time. The approach here was picked over that one as CI time for installing dependencies early does not increase much, and enforcing installation happens at build time opens us up to more possibilities around tightening the typechecking of compat tests.

See [here](https://dev.azure.com/fluidframework/internal/_build/results?buildId=397518) for a test run of the e2e test pipeline (only local stage ran, others are not important). Cache keys to be fixed in a follow-up PR.


packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,419 @@
...
+// ---------------------------------------------------------------------------
+
+function resolveRangeViaRegistry(range: string): string {
+	if (semver.valid(range)) return range;
+	let result: string;
+	try {
+		result = execSync(
+			`${pnpmCmd} view "@fluidframework/container-loader@${range}" version --json`,
+			{ encoding: "utf8" },
+		).trim();
```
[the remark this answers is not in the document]
Abe27342: This is generally pre-existing and actual risk here is extremely low: we're not sticking random string interpolations in `pnpm view`, just specific ranges we calculate with well-defined behaviors. Also this PR makes the situation generally better by removing `pnpm view` from being a thing that ever happens during test execution. The only time this script is getting run now, the developer that's running it is about to do a `pnpm install --no-frozen-lockfile` anyway to install the new deps.

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -3,466 +3,276 @@
...
- * Dynamically loads a package from the specified module path.
+ * Dynamically loads a package from the specified module directory.
  *
+ * Uses Node's standard module resolution algorithm via `createRequire`, which naturally handles
+ * both version-specific `node_modules/` and the hoisted workspace-root `node_modules/` without
+ * manual path construction. The resolved file is then loaded with `import()`.
+ *
+ * @param modulePath - Path to the version directory (e.g. `compat-workspaces/full/2.83.0`).
+ * The resolver starts here and walks up to find hoisted packages.
+ * @param pkg - Package name to load (e.g. `@fluidframework/container-loader`).
```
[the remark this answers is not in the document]
Abe27342: resolved

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,419 @@
...
+	if (existsSync(pkgJsonPath)) {
+		const existing = readFileSync(pkgJsonPath, "utf8");
+		if (existing === newContent) return false; // no change
+	}
+
+	mkdirSync(versionDir, { recursive: true });
+	writeFileSync(pkgJsonPath, newContent, "utf8");
+	console.log(
+		`  ${existsSync(pkgJsonPath) ? "Updated" : "Created"} ${path.relative(pkgRoot, pkgJsonPath)}`,
+	);
```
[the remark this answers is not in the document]
Abe27342: fixed

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -3,466 +3,276 @@
...
+		const matching = allVersions
+			.filter((v) => semver.valid(v) && semver.satisfies(v, requested))
+			.sort(semver.rcompare);
+		if (matching.length > 0) {
+			resolutionCache.set(requested, matching[0]);
+			return matching[0];
 		}
-	} finally {
-		release();
 	}
```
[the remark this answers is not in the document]
Abe27342: going to claim this is out of scope. If reviewer feels strongly, go ahead and push back. I think it'd be better to figure out what is actually likely to break here first by running through the process for a few releases.

packages/test/test-version-utils/compat-workspaces/full/.npmrc
```diff
@@ -0,0 +1,2 @@
+registry=https://registry.npmjs.org/
```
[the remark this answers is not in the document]
Abe27342: Good call. Gone.

packages/test/test-version-utils/src/compatPackageList.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Checks if the given version has the SparseMatrix moved to sequence-deprecated.
+ * @internal
+ */
+export function versionHasMovedSparsedMatrix(version: string): boolean {
+	// SparseMatrix was moved to "@fluid-experimental/sequence-deprecated" in "2.0.0-internal.2.0.0"
+	return (
+		version >= "2.0.0-internal.2.0.0" || (!version.includes("internal") && version >= "2.0.0")
+	);
```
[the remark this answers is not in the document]
Abe27342: pre-existing, not going to fix here.

I did update the PR so this doesn't show up in the diff anymore though.

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -184,285 +133,97 @@ export function resolveVersion(requested: string, installed: boolean): string {
...
-		if (importPath !== ".") {
-			console.warn(
-				`Package ${pkg} main used despite request for ${importPath} entry (no "exports" property found).`,
-			);
-		}
-		primaryExport = pkgJson.main;
+	// createRequire anchored to the version directory. Node's resolution algorithm walks up
+	// through that directory's node_modules, then the workspace-root node_modules (hoisted), so
+	// we do not need to pass the workspace root separately.
+	const requirePath =
```
Abe27342: we could put the old package loading logic back if we want.

It has the benefit of grabbing our ESM entrypoints on loaded packages.

It has the negative of reimplementing a ton of node's module loading logic.

A third alternative would be to put some boilerplate files in each of the generated directories that allow calling `import` from the context of that file (this is not something node lets you do with import unlike what it lets you do with createRequire).

Open to reviewer's opinion here :)

packages/test/test-version-utils/src/compatConfig.ts
```diff
@@ -505,19 +504,21 @@ export const configList = new Lazy<readonly CompatConfig[]>(() => {
  * @internal
  */
 export async function mochaGlobalSetup(): Promise<void> {
-	const versions = new Set(configList.value.map((value) => value.compatVersion));
-	if (versions.size === 0) {
+	const configs = configList.value;
```
Abe27342, line 512: The logic in this file is pretty similar to what we need in `updateCompatVersions.ts`. But not quite with how manifests are structured today. It's probably we want to refactor those two to go through the same codepaths though. All of the super complex decisions are only happening in one place with the current state of the PR, but there are still some areas I notice with duplicated code.

I'm open to trying to do more work in that realm, but I think for ease of review it might be better as a follow-up. Putting this comment here to get reviewer's thoughts though.

packages/test/test-version-utils/compat-workspaces/full/.npmrc
```diff
@@ -0,0 +1 @@
+node-linker=hoisted
```
alexvy86: In the other PR (or somewhere) there was a comment about why hoisting, right? Would be good to document here IMO.

packages/test/test-version-utils/src/testApi.ts
```diff
@@ -131,31 +69,33 @@ export interface InstalledPackage {
...
 ): Promise<InstalledPackage | undefined> => {
-	const pkg = await ensureInstalled(
-		getRequestedVersion(baseVersion, version),
-		packageListToInstall,
-		force,
-	);
+	const requestedStr = getRequestedVersion(baseVersion, version);
+	if (semver.satisfies(pkgVersion, requestedStr)) {
+		return undefined;
+	}
```
alexvy86, line 89: Checking my understanding: this is saying "the current version (i.e. the one running this code, aka "me") satisfies the requested one, and since I'm running I'm obviously installed; nothing to do"? But it doesn't need to return something about its own modulePath?

packages/test/test-version-utils/package.json
```diff
@@ -38,17 +38,19 @@
 		"check:exports:index": "api-extractor run --config api-extractor/api-extractor-lint-index.json",
 		"check:format": "npm run check:biome",
 		"ci:build:docs": "api-extractor run",
-		"clean": "rimraf --glob dist lib \"**/*.tsbuildinfo\" \"**/*.build.log\" _api-extractor-temp node_modules/.legacy nyc",
+		"clean": "rimraf --glob dist lib \"**/*.tsbuildinfo\" \"**/*.build.log\" _api-extractor-temp nyc",
 		"eslint": "eslint --quiet --format stylish src",
 		"eslint:fix": "eslint --quiet --format stylish src --fix --fix-type problem,suggestion,layout",
 		"format": "npm run format:biome",
 		"format:biome": "biome check . --write",
+		"postinstall": "pnpm --dir compat-workspaces/full install --frozen-lockfile",
```
alexvy86, line 46: Do we need to do anything so the new postinstall script actually runs when the e2e tests package is installed in the e2e tests pipeline?

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,321 @@
...
+const versionsCjsPath = path.join(compatWorkspacesDir, "versions.cjs");
+
+/** Oldest compatible version — HUMAN-MAINTAINED. Update when the OCV policy changes. */
+const OCV = "2.0.0-internal.5.4.2";
+
+/**
+ * Number of "majors" before the public 2.0.0 release: 8 internal + 5 RC.
+ * This is the same constant used in genFullBackCompatConfig in compatConfig.ts.
+ */
+const NUM_INTERNAL_MAJORS_BEFORE_PUBLIC_2 = 8 + 5;
```
alexvy86: Nit: export it from that file and import here, since we can import from src/?

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,321 @@
...
+function resolveRangeViaRegistry(range: string): string {
+	if (semver.valid(range)) return range;
+	let result: string;
+	try {
+		result = execSync(
+			`${pnpmCmd} view "@fluidframework/container-loader@${range}" version --json`,
+			{ encoding: "utf8" },
+		).trim();
+	} catch (e) {
+		throw new Error(`pnpm view failed for range "${range}": ${e}`);
```
alexvy86: The only two callers of `resolveRangeViaRegistry` swallow the error. Should this write to stderr instead since the callers don't seem to care when it throws?

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,321 @@
...
+
+	// Resolve standard versions
+	console.log("\nResolving standard versions...");
+	let nMinus1 = resolveVersionForDelta(pkgVer, -1);
+	const nMinus2 = resolveVersionForDelta(pkgVer, -2);
+
+	if (nMinus1 === undefined) {
+		console.warn("  N-1 not found, falling back to N-2");
+		nMinus1 = nMinus2;
+	}
```
alexvy86: Why can this happen?

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -3,173 +3,99 @@
...
+// ---------------------------------------------------------------------------
 
-let cachedInstalledJson: InstalledJson | undefined;
-function writeAndUpdateInstalledJson(data: InstalledJson): void {
-	cachedInstalledJson = data;
-	writeFileSync(installedJsonPath, JSON.stringify(data, undefined, 2), { encoding: "utf8" });
+/**
+ * Schema for the committed `compat-workspaces/versions.cjs` file.
+ * MACHINE-MAINTAINED by `scripts/updateCompatVersions.ts` — do not edit by hand.
+ * @internal
```
alexvy86: New things in this file aren't package-exported so don't need `@internal` tags.

packages/test/test-version-utils/src/versionUtils.ts
alexvy86: Don't feel confident enough to comment on this but I suspect @jason-ha might have thoughts.

alexvy86 (commented):
I did what I could reviewing the new script and the changes to `versionUtils.ts` but can't get my brain to wrap around all the potential implications. I'd mostly be trusting the validation we can do through PR and test/ branch runs of e2e tests.

packages/test/test-version-utils/compat-workspaces/full/.npmrc
Abe27342: There's actually no strong reason AFAIK beyond "it used to be this way" (when we were installing with npm). I'm open to changing here, or if we don't want to, at least leaving a comment saying "it's not clear this is necessary, we picked hoisting only because we didn't want to investigate if changing it off that had any problems"

packages/test/test-version-utils/package.json
Abe27342: This approach should just work since postinstall gets executed when the package is itself installed. But worth testing, there are definitely various things I can imagine going wrong (first of all: are we publishing enough of the package's files?).

I'll add a link to the PR description at that point and resolve this comment.

packages/test/test-version-utils/compat-workspaces/full/pnpm-lock.yaml
```diff
@@ -0,0 +1,15590 @@
+lockfileVersion: '9.0'
```
Abe27342, line 1: pnpm cache key in pipeline should include this file

packages/test/test-version-utils/compat-workspaces/full/2.5.0/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "compat-2.5.0",
```
jason-ha, line 2: What is special about 2.5.0?
(Interestingly, independently in telemetry query I found that 2.5.0 was most recent version used by "Loop Mobile" host. Wasn't expecting to see it here though.)

packages/test/test-version-utils/compat-workspaces/full/.npmrc
jason-ha: With version mixing we should avoid any hoisting anywhere if possible.

packages/test/test-version-utils/compat-workspaces/full/pnpm-workspace.yaml
```diff
@@ -0,0 +1,19 @@
...
+# The security settings below are intentional — do not relax them.
+minimumReleaseAge: 1440
+
+# Exclude our own FF scopes so compat tests can install versions we just shipped
+# without hitting the age gate (new releases are < 1440 min old for a few hours).
+minimumReleaseAgeExclude:
+  - "@fluidframework/*"
+  - "@fluid-experimental/*"
+  - "@fluid-internal/*"
+  - "@fluid-private/*"
```
jason-ha, line 11: pre-existing question: why do we have `@fluid-private` packages to install?
I know that `@fluid-private/test-dds-utils` and `@fluid-private/test-loader-utils` aren't marked as private for some reason. But why would we ever need them to be installed?

packages/test/test-version-utils/compat-workspaces/explicit-versions.mjs
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * After editing this file, re-run `pnpm run update-compat-versions` from
+ * packages/test/test-version-utils to regenerate workspace package.json files and lockfiles,
+ * then commit all changes.
+ */
+export const explicitVersions = [
+	"0.56.0", // legacyChunking.spec.ts -- pre-2.x chunking behavior
+	"2.0.0-internal.1.4.6", // compression.spec.ts -- loader before compression field was introduced
+	"2.0.0-internal.7.0.0", // entryPointCompat.spec.ts -- loader before entryPoint API change
+];
```
jason-ha, line 21: nit
```suggestion
] as const;

```
so the full exact contents can be observed where used.

packages/test/test-version-utils/compat-workspaces/full/pnpm-workspace.yaml
Abe27342: I think the only reason (but not toally sure) is due to how include-test-real-service.yml is set up. I think if we adopt the design I proposed [here](https://dev.azure.com/fluidframework/internal/_workitems/edit/71403) instead we can stop publishing those packages completely.

Fortunately, we are already publishing them to an ADO feed that we do not have partners upstream, so there's little risk of anything other than our own engineering systems breaking if we try.

packages/test/test-version-utils/compat-workspaces/full/2.5.0/package.json
Abe27342: In what sense? as in why it's not 2.3.0 similar to other versions?

The compat matrix we support includes latest "legacy minors" from each "legacy major" (actual minor multiple of 10). Looks like we released several "legacy minors" before establishing the breaking policy back around the 2.0.0 days. So seems correct to me that this is here.

If you look at an existing compat test run, you'll see we also pick 2.5.0 there, see [here](https://dev.azure.com/fluidframework/internal/_build/results?buildId=396865&view=logs&j=26483432-40f5-552c-5792-70aed8cde738&t=0ec67500-082c-56c4-f7ff-020724ca34e2).

However as per my note on teams I think there are pre-existing things that are wrong with the set of versions we test, so feel free to point out anything else you notice that's wrong (will punt to follow-up if pre-existing)

packages/test/test-version-utils/compat-workspaces/full/.npmrc
Abe27342: Yeah I'm inclined to agree. I'll see if there are any problems with switching to default pnpm module structure as part of this PR.

packages/test/test-version-utils/src/versionUtils.ts
Abe27342: I've changed my mind.

One very large plus of *not* using createRequire / import is that it gives us very explicit control over which version we load.

Current iteration of the PR actually can silently load not compat versions of these packages, since createRequire will happily walk up the module tree and happily load from test-version-utils' dependencies!

I'm going to restore the old approach here. If we want to do anything else let's take as follow-up.

packages/test/test-version-utils/compat-workspaces/versions.cjs
jason-ha, line 1: Should we name this versionsGenerated or generated-versions and have symmetry?

packages/test/test-version-utils/compat-workspaces/versions.cjs
```diff
@@ -0,0 +1,28 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+// MACHINE-GENERATED by scripts/updateCompatVersions.ts — do not edit by hand.
+// Run `pnpm run update-compat-versions` from packages/test/test-version-utils to regenerate.
+module.exports = {
```
jason-ha, line 10: Why not .mjs like explicit-versions?

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,321 @@
+#!/usr/bin/env node
```
jason-ha: But we can't runt this. It is a .ts file and needs transpiled.
Is this to shortcut some other setup?

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,321 @@
...
+const scriptDir = fileURLToPath(new URL(".", import.meta.url));
+const pkgRoot = path.resolve(scriptDir, "..");
+const gitRoot = execSync("git rev-parse --show-toplevel", { encoding: "utf8", cwd: pkgRoot })
+	.trim()
+	.replace(/\\/g, "/");
+const compatWorkspacesDir = path.join(pkgRoot, "compat-workspaces");
+const versionsCjsPath = path.join(compatWorkspacesDir, "versions.cjs");
+
+/** Oldest compatible version — HUMAN-MAINTAINED. Update when the OCV policy changes. */
+const OCV = "2.0.0-internal.5.4.2";
```
jason-ha: Could this be in explicit versions file? Then there is only one place to touch excluding maintenance?

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,321 @@
...
+
+/**
+ * Number of "majors" before the public 2.0.0 release: 8 internal + 5 RC.
+ * This is the same constant used in genFullBackCompatConfig in compatConfig.ts.
+ */
+const NUM_INTERNAL_MAJORS_BEFORE_PUBLIC_2 = 8 + 5;
+
+const pnpmCmd = process.platform === "win32" ? "pnpm.cmd" : "pnpm";
```
jason-ha: Is there a note for this. `pnpm` should work for win32, I would think.

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,321 @@
...
+
+// ---------------------------------------------------------------------------
+// Version resolution
+// ---------------------------------------------------------------------------
+
+function resolveRangeViaRegistry(range: string): string {
+	if (semver.valid(range)) return range;
```
jason-ha: This function needs notes.
The first line seems to say, if you gave me a range, okay were done. It wants not a range? A specific version?

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,321 @@
...
+	return version;
+}
+
+function resolveVersionForDelta(pkgVer: string, delta: number): string | undefined {
+	const range = calculateRequestedRange(pkgVer, delta);
+	try {
+		return resolveRangeViaRegistry(range);
+	} catch {
+		return undefined;
+	}
```
jason-ha, line 99: Feels like there should be some logging verbosity here

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,321 @@
...
+	const fullDir = path.join(compatWorkspacesDir, "full");
+	removeStaleVersionDirs(fullDir, allVersions);
+	for (const version of allVersions) {
+		const changed = syncVersionDirectory(fullDir, version);
+		if (changed) anyChanged = true;
+	}
+
+	if (!anyChanged) {
+		console.log("\nNo package.json changes — versions are up to date.");
+	}
```
jason-ha: Doesn't capture stale removals

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,321 @@
...
+
+	// Regenerate lockfile
+	pnpmInstallWorkspace(fullDir);
```
jason-ha: Why if there are no changes (see note that stale removals aren't tracked by anyChanges)?

jason-ha (commented):
Partial review - mostly of the new content - flushing what I have and will continue

packages/test/test-version-utils/compat-workspaces/full/2.5.0/package.json
jason-ha: Right. 2.5 is the latest before 2.10. It just looked different than the rest. This is all good.
I think there are too many of these and we only need to check with some, but that can be after the restructuring.

packages/test/test-version-utils/compat-workspaces/full/pnpm-workspace.yaml
jason-ha: Maybe if that pipeline is doing dev installs. No need to rock the boat here.

packages/test/test-version-utils/compat-workspaces/full/.npmrc
Abe27342: ok, tried it. There are indeed issues with moving to isolated related to how we load local-server. I filed a work item here for this: https://dev.azure.com/fluidframework/internal/_workitems/edit/72053.

Since those issues are pre-existing and not critical for the point of this PR, not going to investigate any further for now.

packages/test/test-version-utils/src/testApi.ts
Abe27342: That's right.

We don't actually use the return value anymore (unrelated to this PR). So the `undefined` doesn't matter. I'll refactor this in a small followup to just be `Promise<void>` since it touches some more files that I'm not otherwise changing in this PR and it's mostly pre-existing.

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: This is pretty clearly an attempt to mimic this pre-existing logic in `getRequestedVersion`:

```
export function getRequestedVersion(
	baseVersion: string,
	requested?: number | string,
	adjustPublicMajor: boolean = false,
): string {
	const calculatedRange = calculateRequestedRange(baseVersion, requested, adjustPublicMajor);
	try {
		// Returns the exact version that was requested (i.e. 2.0.0-rc.2.0.2).
		// Will throw if the requested version range is not valid.
		return resolveVersion(calculatedRange, false);
	} catch (err: any) {
		// If we tried fetching N-1 and it failed, try N-2. It is possible that we are trying to bump the current branch
		// to a new version. If that is the case, then N-1 may not be published yet, and we should try to use N-2 in it's place.
		if (requested === -1) {
			const resolvedVersion = getRequestedVersion(baseVersion, -2, adjustPublicMajor);
			// Here we cache the result so we don't have to enter the try/catch flow again.
			// Note: This will cache the resolved version range (i.e. >=2.0.0-rc.4.0.0 <2.0.0-rc.5.0.0). Because of this,
			// it will not cause any conflicts when trying to fetch the current version
			// i.e. `getRequestedVersion("2.0.0-rc.5.0.0", 0, false)` will still return "2.0.0-rc.5.0.0".
			resolutionCache.set(calculatedRange, resolvedVersion);
			return resolvedVersion;
		} else {
			throw new Error(`Error trying to getRequestedVersion: ${err}`);
		}
	}
}
```

but one big advantage of this new approach is we shouldn't ever be in such scenarios, since set of versions we compat test against is not determined dynamically but at well-defined points in branch lifetime. So this shouldn't be necessary, good catch making me rethink that assumption :)

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: I think the throws as-is are convenient for control flow but I updated the outer-more logic to use a helper that takes care of converting logs to undefined.

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: Gone :)

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: Yup. Same for OCV

packages/test/test-version-utils/src/testApi.ts
Abe27342: https://dev.azure.com/fluidframework/internal/_workitems/edit/72063

packages/test/test-version-utils/package.json
Abe27342: https://dev.azure.com/fluidframework/internal/_build/results?buildId=397364&view=results is an *almost* clean run. Only issues with it are related to our pnpm caching logic.

I think any changes I make there have the potential to break other areas, so I'll prefer to push those to a follow-up PR which only touches cache paths. Only impact of doing that should be some slightly slower installs until I get that follow-up in.

packages/test/test-version-utils/compat-workspaces/full/pnpm-workspace.yaml
Abe27342: Yeah. I'm somewhat motivated to pick up that item after this one, since it's the last place AFAIK we still have a lockless install. But that's for another day :)

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: TBH this was agent generated and I figured it was necessary to let [jiti](https://www.npmjs.com/package/jiti) work, but don't see any mention of that in their docs. I'll double-check it's required.

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: already fixed in latest to pull value from the previous place we defined this constant :)

packages/test/test-version-utils/compat-workspaces/full/.npmrc
Abe27342: ...and it turns out there are also issues with moving to hoisted! (because the old setup involved multiple pnpm workspaces, this one only involves one, and our module loading logic made some bad assumptions about the dep tree). It turns out that annoying bit with server-local-server is the only problem with isolated node_modules, and there's a pretty isolated patch we can use to smooth over that and use isolated node_modules (specifically see changes in testApi.ts and associated comments). This seems better than the bad assumptions we'd have to add to handle the hoisted case, so I went with it. PR now implements isolated.

packages/test/test-version-utils/compat-workspaces/explicit-versions.mjs
Abe27342: this file doesn't have a build step currently, and I'd prefer not to add one for simplicity. So no `as` (no typescript :))

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: Hmm there's a rabbit hole here I don't want to go down again. This was copied from `versionUtils.ts` that has this note:

```
// See https://github.com/nodejs/node-v0.x-archive/issues/2318.
```

but seems like it is indeed not necessary to specify .cmd in this script. Probably due to behavior differences between execFileSync and execSync

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: It is indeed not necessary. Removed.

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: Added doc comments here and to other functions describing intent. They all seem relatively reasonable--lmk if you want any more changes.

("range" is really "rangeSpec" in the sense of "valid things you could put into [semver calculator](https://semver.npmjs.com/)". one such thing is simply an already valid version)

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: Good catch. Does now.

packages/test/test-version-utils/compat-workspaces/versions.cjs
Abe27342: .mjs is more troublesome to load in a .cjs environment (needs async). `test-version-utils` has some sync functionality to check if a given package is installed which I don't want to change here.

We could probably untangle the dep web here if need be given we don't dual-build test-version-utils anyway, but I'd rather not take that on in this PR.

packages/test/test-version-utils/.npmignore
```diff
@@ -5,3 +5,6 @@ src/test
 dist/test
 lib/test
 **/_api-extractor-temp/**
+# Default .npmignore rules only include the root node_modules.
+# Adding an explicit rule here ensures we don't publish the compat_workspace's node_modules.
+**/node_modules
```
Abe27342, line 10: (this rule is in fact necessary, a prior version of this PR produced a test package that was 100MB :))

packages/test/test-version-utils/src/testApi.ts
```diff
@@ -424,11 +369,19 @@ async function loadDriver(baseVersion: string, requested?: number | string): Pro
...
 		];
 
-		// Load the "@fluidframework/server-local-server" package directly without checking for version compatibility.
-		// This is because the server packages have different versions that client packages and the versions requested
-		// do not apply to server packages.
+		// Load the "@fluidframework/server-local-server" package directly without checking for
+		// version compatibility. Server packages have different versioning from client packages.
+		// @fluidframework/server-local-server is not a direct dependency of the compat workspace,
+		// but it is a known dependency of @fluidframework/local-driver, so use that as the base path.
+		// The extra work here is done to handle the structure of pnpm's isolated node_module tree.
```
alexvy86:
```suggestion
		// The extra work here is done to handle the structure of pnpm's isolated node_module tree
		// (as configured in `compat-workspaces/full/.npmrc` in this package).
```

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -3,173 +3,97 @@
...
-	}
-}
-const readInstalledJsonLazy = new LazyPromise(async () => readInstalledJson());
-async function getInstalledJson(): Promise<InstalledJson> {
-	return cachedInstalledJson ?? (await readInstalledJsonLazy);
+/**
+ * Returns all exact versions recorded in the manifest.
+ */
+export function getAllManifestVersions(manifest: CompatVersionsManifest): string[] {
+	return manifest.versions;
```
alexvy86, line 82: Nit: a function feels unnecessary if this is all it does.

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -184,27 +108,24 @@ export function resolveVersion(requested: string, installed: boolean): string {
...
 			}
-		});
-		if (found) {
-			return found;
 		}
-		throw new Error(
-			`No matching version found in ${baseModulePath} (requested: ${requested})`,
-		);
+
+		throw new Error(`No version in manifest satisfies range: "${requested}"`);
```
alexvy86, line 125: This error should be "could not read manifest" or something along those lines, and the current error should go at the end of the preceding `if` block, no?

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -242,163 +163,46 @@ export function resolveVersion(requested: string, installed: boolean): string {
...
 
+// ---------------------------------------------------------------------------
+// Package loading
+// ---------------------------------------------------------------------------
+
 /**
  * Dynamically loads a package from the specified module path.
  *
+ * @param modulePath - Path to the version directory (e.g. `compat-workspaces/full/2.83.0`).
+ * The resolver starts here and walks up to find hoisted packages.
```
alexvy86: Is the "hoisted packages" bit still accurate after the switch to isolated node_modules?

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -458,11 +262,17 @@ export const loadPackage = async (
...
+// Version arithmetic
+// ---------------------------------------------------------------------------
+
 /**
- * Helper function for `getRequestedVersion()`. This function calculates the requested version **range**
- * based on the base version and the requested value, **without** validating it via npm.
+ * Computes the semver range corresponding to a delta from a base version, without resolving it
+ * to an exact version. Used by both the test runtime and the `update-compat-versions` script.
+ *
+ * @internal
```
alexvy86:
```suggestion
```

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,345 @@
...
+
+// ---------------------------------------------------------------------------
+// Version resolution
+// ---------------------------------------------------------------------------
+
+/**
+ * Resolves a semver dependency spec to the single highest version matching that spec which is published in the npm registry.
+ * @param rangeSpec - A valid (as per [semver](https://www.npmjs.com/package/semver)) range specification
+ */
+function resolveRangeViaRegistry(rangeSpec: string): string {
```
alexvy86, line 72: Nit: nearly identical to some logic in `resolveVersion()` in `versionUtils.ts`, maybe make it reusable there? IMO the caching bit in there could be done after a helper returns the actual version.

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,345 @@
...
+	mkdirSync(versionDir, { recursive: true });
+	writeFileSync(pkgJsonPath, newContent, "utf8");
+	console.log(`  ${existed ? "Updated" : "Created"} ${path.relative(pkgRoot, pkgJsonPath)}`);
+	return true;
+}
+
+/**
+ * Deletes any subdirectory of `workspaceDir` whose name is not in `keepVersions`
+ * (excluding `node_modules`). Returns `true` iff at least one directory was removed.
+ */
```
alexvy86: A bit less ambiguous.
```suggestion
 * Deletes any subdirectory (excluding `node_modules`) of `workspaceDir` whose name is not in `keepVersions`.
 * Returns `true` iff at least one directory was removed.
 */
```

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: There's some weirdness here I don't want to untangle in this PR. I'll deduplicate in the followup that also refactors `ensureInstalled` and co

alexvy86 (approved):
LGTM as far as I can tell. The one remaining nit (`getAllManifestVersions` in `versionUtils.ts` seems unnecessary, inline-able) can be addressed in the upcoming PR(s).

packages/test/test-version-utils/compat-workspaces/explicit-versions.mjs
jason-ha: Of course. No `as` for JS.

packages/test/test-version-utils/README.md
```diff
@@ -146,13 +146,37 @@ resolve the latest version that matches it.
 OPEN ISSUE: while these API can be used directly, currently the default global mocha hook will still run and install the
 default set of legacy versions whether it is necessary or not.
 
-## Implementation notes
+## Updating compat versions
+
+After a Fluid Framework version bump, run from this package's directory:
+
+```
+pnpm run update-compat-versions
```
scottn12, line 154: Are we planning to update the release docs to include this as a step after bumping? (sorry I can't check, eng.ms is down ☹️ )

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,345 @@
...
+	mkdirSync,
+	readdirSync,
+	readFileSync,
+	rmSync,
+	writeFileSync,
+} from "node:fs";
+import * as path from "node:path";
+import { fileURLToPath, pathToFileURL } from "node:url";
+
+import * as semver from "semver";
```
scottn12, line 41: nit: We prefer `semver-ts` when possible

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -3,173 +3,98 @@
...
+ *
+ * ## Updating compat versions
+ *
+ * After a version bump, run:
+ * `pnpm run update-compat-versions`
+ *
+ * This regenerates `generated-versions.cjs` and all per-version `package.json` files, then runs
+ * `pnpm install --no-frozen-lockfile` in the workspace to update the committed lockfile.
+ * Commit all changes produced by the script.
+ */
```
scottn12, line 28: nit: We probably don't need these comments here IMO

scottn12 (approved):
This looks like a really nice improvement, thanks for doing this Abram!

packages/test/test-version-utils/README.md
Abe27342: Yes, that's the plan.

packages/test/test-version-utils/scripts/updateCompatVersions.ts
Abe27342: test-version-utils doesn't have a semver-ts dependency. I wouldn't mind switching it for consistency, but we do meaningfully use `semver.satisfies` which they do not support. Seems less important than in production code as well (I see that's tree-shakeable, that's nice)

packages/test/test-version-utils/compat-workspaces/versions.cjs
jason-ha: There is more support for ESM under CJS now with Node 22. https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require (We might be able to specify "module-sync" in a bunch of our packages and get rid of CommonJS. :) )
There is something I don't understand yet about the organization of things. When I search for uses of those scripts they are referenced in .ts files. That sounds like they must be modified after the test-version-utils package is built.
(That doesn't seem like a requirement for manually managed explicit-version and it could be TypeScript, but let's assume not .ts for simplicity / symmetry...)

So, I am assuming test-version-utils is built (jiti evidence says otherwise), updateCompatVersion is run (looks like via jiti) that potentially updates generated-versions, and finally [without build] something is using versionUtils.ts that is dependent on generated-versions.
- It might be helpful if the versionUtils helpers that updateCompatVersion uses were split out. Maybe it doesn't really matter because there is something else that is reading current generated-versions via versionUtils.ts.
- What is the without build criteria? (If we eliminate no build requirement, then .ts will transpile to CJS + ESM.)

So, then I wonder how do we get away with .mjs for explicit-versions.? The answer seems to be that explicit-versions is not a real sibling to generated-versions. Instead, it is baked into generated-versions.
- The comment in explicit-versions should note that its entries are incorporated. That was not my reading.
- Also it seems like explicit-versions could be .ts since the only use is via jiti.

packages/test/test-version-utils/scripts/updateCompatVersions.ts
jason-ha: There are missing notes.
Or more likely semver.valid should be removed.
1. Usage never expects a single specific version to be specified here. It will currently always return `null`.
2. If given a specific version, there is no check against registry that such a version exists. This is what needs noted if it is kept.
(The docs for `valid` are limited. I wanted to first expected that a range was also going to return non-null.)
(I don't know what it means for a spec that has metadata "+foo" suffix to be deemed okay if use is kept.)

packages/test/test-version-utils/src/testApi.ts
```diff
@@ -424,11 +369,20 @@ async function loadDriver(baseVersion: string, requested?: number | string): Pro
...
+		// @fluidframework/server-local-server is not a direct dependency of the compat workspace,
+		// but it is a known dependency of @fluidframework/local-driver, so use that as the base path.
+		// The extra work here is done to handle the structure of pnpm's isolated node_module tree
+		// (as configured in `compat-workspaces/full/.npmrc` in this package).
+		// This pnpm blog post is a good illustration of that structure: https://pnpm.io/blog/2020/05/27/flat-node-modules-is-not-the-only-way
+		const localDriverDependenciesPath = await fs.realpath(
+			path.join(modulePath, "node_modules", "@fluidframework/local-driver"),
+		);
+		// Strip the trailing /node_modules/@fluidframework/local-driver to get to the path where server-local-server will also be available.
+		const localServerModulePath = path.join(localDriverDependenciesPath, "..", "..", "..");
```
jason-ha, line 383: This looks odd and I think could benefit from other comments or maybe rename `localServerModulePath`
- localDDP = realpath(modulePath + "n_m/@ff/local-driver")
- localServerModulePath = localDDP - "n_m/@ff/local-driver"
There is a material difference to doing `localServerModulePath = realpath(modulePath)`.
I think `localServerModulePath` is more accurately described as `localServerModuleHostPath` as it is the path to the folder that host all of local-driver's deps including server-local-server.

I'm not requesting any immediate change though. I think this really is funky and could be reworked.
1) This logic should not attempt to build `LocalDriverApi` assuming `LocalDeltaConnectionServer` from local-driver. Technically, local-driver doesn't need a full dep on server-local-server. The one thing it needs for TypeScript consumers is `ILocalDeltaConnectionServer` interface. All of the imports local-driver has are type imports. Preference would be for local-driver to depend on a definitions package - just types. This will break if cleaned up.
2) [known] `loadPackages` is fraught and needs replaced.

The local driver API `LocalDeltaConnectionServer` property should be its own thing. Perhaps some other metadata should be in this stuff to pick a server-local-server version and `LocalDeltaConnectionServer`.
OR local-driver should embrace a server-local-server version at least for testing.

cc: @scottn12, @agarwal-navin

Abe27342: It's relatively wrong that we're grabbing this dependency in the first place, yeah. The assumptions here are probably a little less brittle than before, but certainly look very weird.

> OR local-driver should embrace a [new?] server-local-server version at least for testing.

I think moving in this direction is the best choice. I don't think any of our tests really care that the version of server-local-server being used is some particularly old or not-old version, and it's certainly the case that in the wild while we can expect relatively old client bits, we should always have a somewhat modern server.

That would simplify this to a simple static import (or we could remove it from the test-version-utils API and push importing it to the test author or \@fluid-private/test-utils

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -242,163 +160,45 @@ export function resolveVersion(requested: string, installed: boolean): string {
...
 /**
  * Dynamically loads a package from the specified module path.
  *
+ * @param modulePath - Path to the version directory (e.g. `compat-workspaces/full/2.83.0`).
+ * @param pkg - Package name to load (e.g. `@fluidframework/container-loader`).
+ * @remarks
+ * This function reimplements part of Node's module resolution logic. It would be possible to use `createRequire` / `import` alternatively,
+ * but if this approach is taken, the compat workspace where prior versions of FF get installed *must* be moved outside of a context that might
+ * have other versions of FF installed in parent folders. Otherwise, Node's module resolution might happily load incorrect versions of FF packages
+ * from parent folders instead of the intended versions, which would silently break the tests.
```
jason-ha, line 201: Would compat-workspaces need to be moved out?
Doesn't pnpm's isolation help with protection there?
Any holes might be mitigated by not having the current (local) versions be special. They ideally are just one of the sets in the full compat matrix. (Bridging the two workspaces would likely need to be done using `link:` specs instead of `workspace:`.)

packages/test/test-version-utils/src/testApi.ts
Abe27342: Also re: 2, let's be careful when getting rid of it. I almost accidentally regressed all the tests where they were accidentally loading modern versions and thinking they were compat ones in this PR, which is what prompted me to put it back and leave a warning :)

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -242,163 +160,45 @@ export function resolveVersion(requested: string, installed: boolean): string {
...
+	const versionDir = path.join(fullWorkspaceDir, version);
+
+	if (existsSync(versionDir)) {
+		return { version, modulePath: versionDir };
 	}
+
 	throw new Error(
-		`Requested version ${requested} resolved to ${version} is not installed at ${modulePath}`,
+		`Version ${version} is not installed in compat-workspaces/full/.\n` +
+			`Run \`pnpm run update-compat-versions\` to regenerate the workspace, then re-run \`pnpm install\`.`,
```
jason-ha, line 184: Looks like `update-compat-versions` does install. What is this one supposed to do?

packages/test/test-version-utils/src/versionUtils.ts
Abe27342: Assuming you mean moved to outside of test-version-utils (e.g. place them in packages/test/compat-workspace)?

That's the simplest way I can imagine mitigating this concern, yeah. It does require some other work to happen first, since it ruins the ability of test-version-utils to work outside the context of the FF repo.

packages/test/test-version-utils/src/versionUtils.ts
Abe27342: No need for the instruction. Updating in #27218

packages/test/test-version-utils/compat-workspaces/full/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "homepage": "https://fluidframework.com",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/FluidFramework.git",
+    "directory": "packages/test/test-version-utils/compat-workspaces/full"
+  },
+  "license": "MIT",
+  "author": "Microsoft and contributors",
+  "scripts": {
+    "preinstall": "node ../scripts/only-pnpm.cjs"
```
jason-ha, line 14: Seems like the only-pnpm script should be local to this and not reach up. I don't see any other uses of the script.

packages/test/test-version-utils/compat-workspaces/scripts/only-pnpm.cjs
jason-ha, line 1: Would be nice to not have .cjs files unless we need them. This could be .js right?

packages/test/test-version-utils/compat-workspaces/full/.npmrc
```diff
@@ -0,0 +1 @@
+node-linker=isolated
```
jason-ha, line 1: Given that this gets a standalone install AND that can be after a package download, it should probably have some of the other root workspace settings.
```text
frozen-lockfile=true
strict-peer-dependencies=true
```
seem minimal good to have.
Not sure about `engine-strict=true`.

packages/test/test-version-utils/package.json
jason-ha, line 46: This uses `pnpm install` and compat-worspaces/full has preinstall only-pnpm check.
Unless someone is doing something manually funny, the compat-workspaces/full preinstall seems moot. This is the place that needs pnpm check. I am assuming that if pnpm is not available at post-install, then there is a nice error and no change needed here.

jason-ha, line 46: Maybe we prefer to have the frozen-lockfile element enforced via .npmrc in compat-workspaces/full.
It could be nice to have a --prod argument here and in compat update just for extra, good measure. If anything dev shows up in those places, it would be nice to have an explicit decision.

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -0,0 +1,345 @@
...
+	const fullDir = path.join(compatWorkspacesDir, "full");
+	if (removeStaleVersionDirs(fullDir, allVersions)) anyChanged = true;
+	for (const version of allVersions) {
+		const changed = syncVersionDirectory(fullDir, version);
+		if (changed) anyChanged = true;
+	}
+
+	if (anyChanged) {
+		// Regenerate lockfile
+		pnpmInstallWorkspace(fullDir);
```
jason-ha, line 334: Might be nice to have a debug option (script argument), to only update the lockfile and not actually install (`--lockfile-only`). Maybe script args are passed to pnpm.
I could see use in CI wanting to specify `--store-dir`.

packages/test/test-version-utils/package.json
```diff
@@ -38,17 +38,19 @@
 		"check:exports:index": "api-extractor run --config api-extractor/api-extractor-lint-index.json",
 		"check:format": "npm run check:biome",
 		"ci:build:docs": "api-extractor run",
-		"clean": "rimraf --glob dist lib \"**/*.tsbuildinfo\" \"**/*.build.log\" _api-extractor-temp node_modules/.legacy nyc",
+		"clean": "rimraf --glob dist lib \"**/*.tsbuildinfo\" \"**/*.build.log\" _api-extractor-temp nyc",
```
jason-ha, line 41: Removal of these always feel like they should be done a few weeks later.
It doesn't hurt to let people know that if they previously have done X, then post this change they can cleanup Y.

.prettierignore
jason-ha, line 1: We have to maintain this apart from biome.jsonc? They haven't diverged?

packages/test/test-version-utils/README.md
jason-ha: I recommend https://dev.azure.com/fluidframework/internal/_wiki/wikis/FF%20Internal%20Wiki?wikiVersion=GBmain - at least when eng.ms is down

.prettierignore
alexvy86: I remember at some point we were in a position where we had to use both because biome didn't handle all the files we wanted (I want to say `package.json` but I'm not 100% sure). @tylerbutler might remember, and/or know if we can completely get rid of prettier now.

packages/test/test-version-utils/package.json
Abe27342: Yeah, in theory the only-pnpm check should be unnecessary. But it doesn't particularly hurt. It also *is* true if someone is going to do some manual work in that workspace, which they might want to e.g. to manage dependencies for security reasons, they should be using pnpm. Also if we want to not have such a check, we'd need to ignore the nested workspace from policy-check.

Planning on not taking action on this one.

packages/test/test-version-utils/compat-workspaces/scripts/only-pnpm.cjs
Abe27342: personally I don't think the explicit extension is very problematic for this context, so not particularly motivated to change. If we want to change them, we should be consistent across the repo. All other instances of the script are .cjs files (and this package needs its own instance at least for now so that the postinstall script works for pipelines that consume the package from a context other than FF repo)

.prettierignore
Abe27342: No additional context here from me but yeah, would be great to only have one formatter.

packages/test/test-version-utils/package.json
Abe27342: I don't think we get much value from trying to distinguish in these workspaces. The dependencies themselves are all autogenerated by the update script

PR #27188, opened 2026-04-28
test(driver-web-cache): Migrate from `jest` to `mocha`

Josmithr:
Updates the `driver-web-cache` package to use `mocha` for testing instead of `jest`. Part of a larger effort to move away from `jest` where we can.

[AB#71299](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/71299)


packages/drivers/driver-web-cache/package.json
```diff
@@ -82,8 +82,9 @@
 		"format:biome": "biome check . --write",
 		"lint": "fluid-build . --task lint",
 		"lint:fix": "fluid-build . --task eslint:fix --task format",
-		"test": "npm run test:jest",
-		"test:jest": "jest --ci",
+		"test": "npm run test:mocha",
+		"test:mocha": "cross-env FLUID_TEST_MODULE_SYSTEM=CJS mocha",
```
alexvy86, line 86: I don't know enough about this part of the codebase to say if we should be using FLUID_TEST_MODULE_SYSTEM here or not, but calling it out to make sure it's intentional. Based on `"Tests use fake-indexeddb v3 (CJS-only)"` in the mocha config, seems like it is?

packages/drivers/driver-web-cache/src/test/FluidCacheTimer.test.ts
```diff
@@ -73,52 +72,64 @@ function getMockCacheEntry(itemKey: string, options?: { docId: string }): ICache
...
+		(window.indexedDB as unknown) = new FDBFactory();
+	});
+
+	afterEach(() => {
+		if (fluidCache !== undefined) {
+			clearTimeout(fluidCache["dbCloseTimer"]);
+			fluidCache["db"]?.close();
+		}
+		extraDb?.close();
+		extraDb = undefined;
```
alexvy86, line 91: With jest were we getting this behavior by default?

packages/drivers/driver-web-cache/src/test/FluidCache.test.ts
```diff
@@ -222,78 +236,81 @@ for (const immediateClose of [true, false]) {
...
-			const cacheEntry = getMockCacheEntry("someKey");
-			const cachedItem = { dateToStore: "foo" };
-			await fluidCache.put(cacheEntry, cachedItem);
-
-			const result = await fluidCache.get(cacheEntry);
-			expect(result).toEqual(undefined);
+			// put() should return gracefully even when the DB is blocked by an older client.
+			// We intentionally do not call get() after put() here — a second blocked open
+			// request creates a second leaked fake-indexeddb connection that never closes,
+			// deadlocking the waitForOthersClosed loop in fake-indexeddb v3.
```
alexvy86, line 289: Interesting; the jest test did call get afterwards. Do we know why the change in behavior with mocha?

packages/drivers/driver-web-cache/package.json
Josmithr: Yes, intentional for that reason.

packages/drivers/driver-web-cache/src/test/FluidCacheTimer.test.ts
Josmithr: Ah, good question. No, I don't think so. This change was made to allow us to not specify `exit: true` to mocha. I'm not sure if jest behaves differently in that regard, but Claude identified these cleanup changes as being necessary when I prompted it to remove the exit setting.

packages/drivers/driver-web-cache/src/test/FluidCache.test.ts
Josmithr: This was needed in order to not specify `exit: true` in the mocha config.

alexvy86 (approved):

PR #27191, opened 2026-04-28
feat(build-cli): add `flub check trustPolicy` - audit lockfile against pnpm's no-downgrade trust policy

TommyBrosman:
pnpm 10's trustPolicy: no-downgrade rejects installs where a package's publisher or signing key has changed since it was first trusted. Running it against our existing workspace only surfaces one violation at a time and isn't usable as a CI signal because we can't pre-populate the trusted-publishers store from a single run. This new command audits every `name@version` and reports the full violation set in one shot by generating a temporary workspace with a single large package.json containing every single { package, version } entry from the repo.

Tracking bug in the pnpm repo: https://github.com/pnpm/pnpm/issues/10622

## Description

- **New oclif command** `flub check trustPolicy` in the build-cli package, with `--json`, `--keep`, `--tempDir`, and `--verbose` flags.
- **Lockfile enumeration** parses the pnpm lockfile and collects every unique `name@version` from both the `packages` and `snapshots` sections, skipping URL, tarball, and git refs (which trust policy doesn't apply to).
- **Install loop** runs `pnpm install --trust-policy no-downgrade --reporter ndjson`, parses pnpm's NDJSON stdout for `ERR_PNPM_TRUST_DOWNGRADE` events, adds each offender to `--trust-policy-exclude`, and re-runs until pnpm exits cleanly or stops surfacing new violations.
- **Strict NDJSON parsing** throws with the offending line on JSON-parse failure, unrecognized event codes, or missing `package.{name,version}` so any pnpm contract change surfaces loudly instead of silently skipping violations.
- **Structured output** prints a human-readable report by default or a stable JSON document with `tempDir`, `exitCode`, `iterations`, `elapsedSec`, `totalUniqueDependencies`, and the sorted `violations` list under `--json`.

Sample run on this branch: 8 iterations / 2606 unique dependencies / ~82s, surfacing 7 violations (`@langchain/core@0.3.80`, `@octokit/endpoint@9.0.6`, `axios@0.30.3`, `chokidar@4.0.3`, `detect-port@1.6.1`, `semver@5.7.2`, `semver@6.3.1`).


build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,429 @@
...
+	} as const;
+
+	public async run(): Promise<void> {
+		const context = await this.getContext();
+		const repoRoot = context.root;
+		const tempDir = path.resolve(
+			this.flags.tempDir ?? path.join(repoRoot, ".trust-audit-temp"),
+		);
+
+		const lockfilePath = path.join(repoRoot, "pnpm-lock.yaml");
```
CraigMacomber: Don't parse the lock file. Pnpm might change that format in the future, and I'd rather not have another tool we have to update when that happens.

Use [`pnpm list -r --json --depth Infinity`](https://pnpm.io/cli/list) or something like that (that format might also change, but I suspect is more stable)

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,429 @@
...
+ * 2. Materialize a scratch workspace at `<repoRoot>/.trust-audit-temp/`
+ * with one leaf project per `(name, version)`. Each leaf depends on
+ * the *real* registry name (no `npm:` aliases) because pnpm 10's
+ * `--trust-policy-exclude` only matches by registry name.
+ * 3. Run `pnpm install` against the scratch workspace with NDJSON
+ * reporting. pnpm aborts at the first violation; we add the
+ * offender to the exclude list and re-run, repeating until pnpm
+ * either succeeds or stops surfacing new violations.
+ */
+export default class CheckTrustPolicyCommand extends BaseCommand<
```
CraigMacomber: The docs for this should link a upstream tracking issue for this missing feature of pnpm (if its not implemented yet) or the release of pnpm it will be in so we can know when to delete it and replace it with built in support.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,429 @@
...
+		let iteration = 0;
+		const start = Date.now();
+		// eslint-disable-next-line no-constant-condition
+		while (true) {
+			iteration++;
+			const excludeFlags: string[] = [];
+			for (const v of [...violationSet].sort()) {
+				excludeFlags.push("--trust-policy-exclude", v);
+			}
+			const installArgs = [
```
CraigMacomber: Include --lockfile-only to avoid most of the cost of the install (actually putting the package in node _modules)

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
TommyBrosman: That worked.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
TommyBrosman: It saves 9s locally! (67.8s vs. 77s before)

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
TommyBrosman: Done. I also added a comment on the tracking bug describing the "get all the trust violations" workflow and how it could be used as part of onboarding onto using trustPolicy: https://github.com/pnpm/pnpm/issues/10622

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,429 @@
...
+	const safeVersion = version.replace(/[^\dA-Za-z]+/g, "-").replace(/^-+|-+$/g, "");
+	return `${safeName}-${safeVersion}`.toLowerCase();
+}
+
+/**
+ * Creates `tempDir` containing:
+ *
+ * - `pnpm-workspace.yaml` declaring the leaf glob and `trustPolicy: no-downgrade`.
+ * - One leaf project per `(name, version)` under `projects/<slug>/`,
+ * each pulling in exactly one real (non-aliased) dependency.
```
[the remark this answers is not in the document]
TommyBrosman: Copilot, that is literally in the PR description.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { spawn } from "node:child_process";
+import { existsSync, mkdirSync, rmSync, writeFileSync } from "node:fs";
```
tylerbutler: Prefer the `/promises` imports and async for file i/o.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+ * `--trust-policy-exclude` only matches by registry name.
+ * 3. Run `pnpm install` against the scratch workspace with NDJSON
+ * reporting. pnpm aborts at the first violation; we add the
+ * offender to the exclude list and re-run, repeating until pnpm
+ * either succeeds or stops surfacing new violations.
+ *
+ * See https://github.com/pnpm/pnpm/issues/10622 for the bug that motivated
+ * this command.
+ */
+export default class CheckTrustPolicyCommand extends BaseCommand<
```
tylerbutler: Since this is a new command, please inherit from `BaseCommandWithBuildProject`, and use `getBuildProject` instead of getContext.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+		}),
+		...BaseCommand.flags,
+	} as const;
+
+	public async run(): Promise<void> {
+		const context = await this.getContext();
+		const repoRoot = context.root;
+		const tempDir = path.resolve(
+			this.flags.tempDir ?? path.join(repoRoot, ".trust-audit-temp"),
+		);
```
tylerbutler, line 105: IMO this should work for any release group or standalone package in the repo. Maybe accept a path to a workspace (which you can get from the buildProject if needed) as a flag so it can be executed on arbitrary workspaces.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+
+		const lockfilePath = path.join(repoRoot, "pnpm-lock.yaml");
+		if (!existsSync(lockfilePath)) {
+			this.error(`No pnpm-lock.yaml found in ${repoRoot}`);
+		}
```
tylerbutler: Is this a proxy for "workspace root?" If so then the BuildProject will help.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+
+		this.verbose("Enumerating installed dependencies via `pnpm list -r --json`...");
```
tylerbutler: The backticks are rendered as-is in the terminal, so not sure it's worth it.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+ * to this process so progress is visible during long operations.
+ */
+function runPnpm(args: string[], cwd: string, streamLive: boolean): Promise<PnpmRunResult> {
+	return new Promise((resolveRun, rejectRun) => {
+		const child = spawn("pnpm", args, {
+			cwd,
+			shell: true,
+			stdio: ["ignore", "pipe", "pipe"],
+			env: { ...process.env, CI: "1" },
+		});
```
tylerbutler: We use a wrapper called execa that I would recommend using here. If you're especially forward thinking you could adopt tiny-exec now, which is what I would like to replace execa with eventually. This code works, but it's really low level and a higher level API would be easier to read and maintain (IMO).

.gitignore
```diff
@@ -47,6 +47,9 @@ docs/resources/_gen/
 # PNPM store (when mounting host file system in docker container)
 .pnpm-store/
 
+# Scratch dir created by `flub check trustPolicy`.
+.trust-audit-temp/
```
tylerbutler: Will this ignore any folder in the tree with that name? I think that's what you want.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+		"Audits the repo's lockfile against pnpm's `no-downgrade` trust policy.";
+
+	static readonly description =
+		"Materializes a scratch workspace under `.trust-audit-temp/` containing one leaf project per pinned dependency, then runs `pnpm install --trust-policy no-downgrade` and iteratively excludes each violation until pnpm either succeeds or stops surfacing new violations. Reports the full list of trust-downgrade violations.";
+
+	static readonly flags = {
+		json: Flags.boolean({
+			description: "Emit JSON instead of a text report.",
+			default: false,
+		}),
```
tylerbutler: OCLIF has a built-in JSON flag that you can opt into, and I suggest doing that. It works by chaging run to return an object. When --json is passed, the object will be JSON serialized. This is nicer than rolling your own because the json is pretty printed and is pipeable to other downstream processes.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+		keep: Flags.boolean({
+			description: "Do not delete the scratch workspace after running.",
+			default: false,
+		}),
+		tempDir: Flags.directory({
+			description: "Scratch workspace directory (default: <repo-root>/.trust-audit-temp).",
```
tylerbutler: Is there value to putting this in the repo boundary vs. a system-wide temp directory? E.g. /tmp on linux?

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+
+		const violationSet = new Set<string>();
+		let lastResult: PnpmRunResult | undefined;
+		let iteration = 0;
+		const start = Date.now();
+		// eslint-disable-next-line no-constant-condition
+		while (true) {
+			iteration++;
+			const excludeFlags: string[] = [];
+			for (const v of [...violationSet].sort()) {
```
tylerbutler: nit: does this sort on every iteration?

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+					undefined,
+					2,
+				),
+			);
+		} else {
+			this.log(`Audited via pnpm install in: ${tempDir}`);
+			this.log(`  Final pnpm exit code: ${exitCode}`);
+			this.log(`  Iterations: ${iteration}`);
+			this.log(`  Unique pinned versions: ${pinned.length}`);
+			this.log(`  Elapsed: ${elapsedSec}s`);
```
tylerbutler: Perfer this.info because it respects --quiet.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+						iterations: iteration,
+						elapsedSec,
+						totalUniqueDependencies: pinned.length,
+						violations,
+					},
+					undefined,
+					2,
+				),
+			);
+		} else {
```
tylerbutler: This gets a lot simpler with the built-in JSON flag - then you can just return the object.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+				}
+			}
+		}
+
+		if (!this.flags.keep) {
+			this.verbose(`Cleaning up temp dir: ${tempDir}`);
+			rmSync(tempDir, { recursive: true, force: true });
+		} else {
+			this.verbose(`Leaving temp dir in place: ${tempDir}`);
+		}
```
tylerbutler: You probably want to try/catch/finally some of these.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+ *
+ * @returns A lowercase string with runs of non-alphanumeric characters replaced
+ * by `-`, with leading/trailing hyphens trimmed. For example,
+ * `\@fluidframework/tree` + `1.0.0` → `fluidframework-tree-1-0-0`.
+ */
+function slugify(name: string, version: string): string {
+	const safeName = name.replace(/[^\dA-Za-z]+/g, "-").replace(/^-+|-+$/g, "");
+	const safeVersion = version.replace(/[^\dA-Za-z]+/g, "-").replace(/^-+|-+$/g, "");
+	return `${safeName}-${safeVersion}`.toLowerCase();
+}
```
tylerbutler: Honestly might be worth a tiny dep, otherwise just make sure it's got unit tests.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,442 @@
...
+ * We avoid putting `trustPolicyExclude` entries in the YAML because
+ * pnpm 10's YAML form silently drops double-quoted scalars and rejects
+ * bare scoped names; CLI flags are easier to control across iterations.
+ *
+ * @returns The number of leaf projects written (one per entry in `pinned`).
+ */
+function writeAuditWorkspace(tempDir: string, pinned: readonly PinnedVersion[]): number {
+	if (existsSync(tempDir)) {
+		rmSync(tempDir, { recursive: true, force: true });
+	}
```
tylerbutler: TOC/TOU; just try to remove and handle the failure.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
TommyBrosman: I've been going back and forth on the OS temp dir approach. On one hand, it doesn't clutter up the repo. On the other hand, it's harder to find. I think using the OS temp dir makes sense if you never have to inspect the output. The script exposes a "keep" flag for the case where you might want to examine the output. I think I want to keep it in the repo unless it is significantly simpler not to.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
TommyBrosman: It doesn't, fortunately. The expression is evaluated once and returns an Iterable.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
TommyBrosman: Fixed, much simpler.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
TommyBrosman: I changed it, though the new approach might be overkill. It searches the path it sees to find which known workspaces matches.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,429 @@
...
+				break;
+			}
+
+			const newViolations = found.filter((v) => !violationSet.has(v));
+			if (newViolations.length === 0) {
+				this.verbose(
+					`pnpm exited with code ${lastResult.code} but no new trust-policy violations were detected. Stopping.`,
+				);
+				break;
+			}
```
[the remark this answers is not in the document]
TommyBrosman: Fixed.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
TommyBrosman: I have no idea why Claude rolled his own. Fixed.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,534 @@
...
+			exists: true,
+		}),
+		tempDir: Flags.directory({
+			description: "Scratch workspace directory (default: <workspace>/.trust-audit-temp).",
+		}),
+		...BaseCommandWithBuildProject.flags,
+	} as const;
+
+	public async run(): Promise<TrustPolicyAuditResult> {
+		const searchPath = path.resolve(this.flags.path ?? process.cwd());
```
tylerbutler, line 84: nit: you can give the flag a default value of process.cwd() in the flag declaration above - then this defensive code can go away.

build-tools/packages/build-cli/src/commands/check/trustPolicy.ts
```diff
@@ -0,0 +1,534 @@
...
+			if (
+				(searchPath === dir || searchPath.startsWith(`${dir}${path.sep}`)) &&
+				(workspaceDir === undefined || dir.length > workspaceDir.length)
+			) {
+				workspaceDir = dir;
+			}
+		}
+		if (workspaceDir === undefined) {
+			this.error(`No workspace in build project ${buildProject.root} contains ${searchPath}.`);
+		}
```
tylerbutler, line 101: Hmmm, I see what you mean. You could consider taking a workspace name or release group name as a flag, but I think this is fine. Easy enough to refactor later if needed, and the comment is clear.

build-tools/packages/build-cli/docs/build-perf.md
```diff
@@ -75,7 +75,7 @@ DESCRIPTION
 EXAMPLES
   Collect public (PR) build data.
 
-    $ flub build-perf collect --mode public --project public --buildDefId 11 --outputDir ./output --adoApiToken \
+    $ flub build-perf collect --mode public --project public --buildDefId 11 --outputDir ./output --adoApiToken `
```
TommyBrosman: I'm not sure why this is showing up.

tylerbutler (approved):

tylerbutler:
Do a full build-tools local build and commit the results - that should address the build failures.

PR #27196, opened 2026-04-28
docs(client-changeset): `@` handling guidance

jason-ha:
`@foo` becomes a GitHub user mention in release notes and often should be avoided.


tylerbutler (approved):
LGTM - your call on the copilot suggestions. First one seems wrong, second one more interesting.

alexvy86 (approved):

alexvy86 (approved):

tylerbutler (approved):

PR #27200, opened 2026-04-29
feat(tree): internals for field batch format with specialized node shapes

justus-camp-microsoft:
## Description

Adds internals for a new experimental chunked-forest codec format (`vTextExperimental`, version key `"text"`) targeted at compressing runs of nodes that differ only in a few properties — for example, styled text runs where many nodes share most of their structural skeleton.

The core addition is `EncodedSpecializedNodeShape` (`f` shape variant): a node shape that derives from another node shape by overlaying property-level overrides. A base node shape defines the structural skeleton; the specialization specifies overrides for those properties.

Decode-side support lives in `SpecializedNodeDecoder`, which walks the specialized `f` shapes back to a given `c` shape via `resolveToNodeShape`, then merges via `applySpecialization`: base field order is preserved, overridden fields are replaced in place, and new fields are appended. Cycles and missing bases are caught with asserts.

Existing types broadened for the new union:
- `EncodedChunkShapeV1OrV2` → `EncodedChunkShape` (now `V1 | V2 | VTextExperimental`).
- `shapesV2` factored out of `EncodedChunkShapeV2` so the new format can extend it.

This PR is decoder + format scaffolding only. Encoders are not yet emitting `f` shapes; that will follow in a separate change. The codec will be registered as part of introducing the encoder.

## Reviewer Guidance


packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,109 @@ export function decode(
 	);
 }
 
+/**
+ * Walks a chain of specialized node shapes (`f`) back to the concrete `c` node shape they are
```
CraigMacomber: From this description and the function name, it sounds like this returns the base shape.

From the implementation, this looks like it looks up shape, and normalizes out any specialized node shapes. I think the docs and likely function name could use adjustment.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,109 @@ export function decode(
...
+ * ultimately based on.
+ */
+function resolveToNodeShape(
```
CraigMacomber: thus function, as well as applySpecialization would be great candidates to export specifically for testing, and to unit test. That would make simpler more targeted testing practical for cases like multiple specialized shapes chaining back to a base shape

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,109 @@ export function decode(
...
+	shapeIndex: number,
+	context: DecoderContext<EncodedChunkShape>,
+	visited: Set<number> = new Set(),
+): EncodedNodeShape {
+	assert(!visited.has(shapeIndex), "cyclic specialized node shape chain");
+	visited.add(shapeIndex);
+	const encoded = context.shapes[shapeIndex];
+	assert(encoded !== undefined, "base shape index out of bounds");
+	if (encoded.c !== undefined) {
```
CraigMacomber: due to the annoyances of these short and semantically unclear discriminators in the persisted format, a comment here would be nice for people who don't recognize this pattern or know which cases c and f are

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,109 @@ export function decode(
...
+		return encoded.c;
+	}
+	assert(
+		"f" in encoded && encoded.f !== undefined,
+		"specialized node shape base must resolve to a node shape",
```
CraigMacomber: I think this message isn't right, as this isn't testing the base, and the base can be either specialized or a regular node shape.

CraigMacomber:
Note about the PR title: this is a new format for an existing codec, not a new codec (our code doesn't separate these well yet, but I've been trying to make it clearer. I'd describe this as "add internals for experimental field batch codec format with specialized node shapes" (or maybe shortened to "internals for field batch format with specialized node shapes" for the description)

It might make sense to include the addition of the new format to fieldBatchCodecBuilder as part of this (lack of that is why I called this "internals" above), and thus support actually encoding and decoding the new format (just not using `f` in the encoder yet. That would opt the new code into per format test coverage for things like round tripping and json schema snapshotting. Since the format version is unstable (a string) that should be safe. I'm fine with it either way though.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
justus-camp-microsoft: I exported these functions and added some tests for each of them.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
justus-camp-microsoft: Added a comment here about what a `c` and `f` shape denote

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
justus-camp-microsoft: You're right, updated this one

justus-camp-microsoft:
> Note about the PR title: this is a new format for an existing codec, not a new codec (our code doesn't separate these well yet, but I've been trying to make it clearer. I'd describe this as "add internals for experimental field batch codec format with specialized node shapes" (or maybe shortened to "internals for field batch format with specialized node shapes" for the description)
> 
> It might make sense to include the addition of the new format to fieldBatchCodecBuilder as part of this (lack of that is why I called this "internals" above), and thus support actually encoding and decoding the new format (just not using `f` in the encoder yet. That would opt the new code into per format test coverage for things like round tripping and json schema snapshotting. Since the format version is unstable (a string) that should be safe. I'm fine with it either way though.

I updated the title/description to denote that the codec isn't actually registered and that this is internals. I think it's easier to register the codec once we have both the encoder and decoder so I'll wait for the next change to do that if that's alright.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,120 @@ export function decode(
...
+/**
+ * Resolves `shapeIndex` to a fully-merged {@link EncodedNodeShape}, normalizing away any
+ * specialized node shapes (`f`) along the way by applying their overlays via
+ * {@link applySpecialization} until a concrete node shape is reached.
+ *
+ * @remarks
+ * Exported for testing.
+ */
+export function normalizeToNodeShape(
+	shapeIndex: number,
```
CraigMacomber: For clarity, use our ShapeIndex type alias for number here and in the set below.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,120 @@ export function decode(
...
+	context: DecoderContext<EncodedChunkShape>,
+	visited: Set<number> = new Set(),
```
CraigMacomber: Extending the API docs to cover parameters, and noting why this exists and how most callers don't need to provide it would be good.

Maybe a more specific name for it would help too: our codec does have logic to traverse and visit all shape which this might be confused with.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,120 @@ export function decode(
 	);
 }
 
+/**
+ * Resolves `shapeIndex` to a fully-merged {@link EncodedNodeShape}, normalizing away any
```
CraigMacomber: I think this should either:
1. document that it only is allowed to be used on EncodedNodeShape | SpecalizedNodeShape
OR
2. work on any node shape, and return a union of all the types except SpecalizedNodeShape
OR
3. Take in the input as type EncodedNodeShape | SpecalizedNodeShape instead of a shape index.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,120 @@ export function decode(
...
+ * value wins.
+ *
+ * For `fields`, override entries whose key matches a key in `base.fields` replace that entry's
+ * shape index in place; override entries with keys not present in `base` are appended at the end
+ * in `spec.fields` order. Field order from `base` is preserved.
+ *
+ * @remarks
+ * Exported for testing.
+ */
+export function applySpecialization(
```
CraigMacomber, line 151: EncodedSpecializedNodeShape documents the merge rules: this should refer to/link that and that should likely link here as well.

Also I noticed an issue with those documented rules:

> `extraFields` are inherited unless the specialization sets them as own properties — to
>  * inherit, the property must be omitted; setting it explicitly (even to `false` or
>  * `undefined`) is treated as an override.

These types are supposed to be JSON compatible since they get json serialized. JSON does not preserve undefined properties, so we should not have them be significant or it won't work with real encoded data (I'm not sure if this is a docs bug or a real format issue)

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,120 @@ export function decode(
...
+		const overrideShape = overrides.get(key);
+		if (overrideShape === undefined) {
+			mergedFields.push([keyEncoded, shapeIndex]);
+		} else {
+			overriddenKeys.add(key);
+			mergedFields.push([keyEncoded, overrideShape]);
+		}
+	}
+	for (const [keyEncoded, shapeIndex] of spec.fields ?? []) {
+		if (!overriddenKeys.has(context.identifier<FieldKey>(keyEncoded))) {
```
CraigMacomber: This could just use overrides instead of overriddenKeys, then overriddenKeys can be entirely removed.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,120 @@ export function decode(
...
+		baseKeys.add(key);
+		const overrideShape = overrides.get(key);
+		if (overrideShape === undefined) {
+			mergedFields.push([keyEncoded, shapeIndex]);
+		} else {
+			overriddenKeys.add(key);
+			mergedFields.push([keyEncoded, overrideShape]);
+		}
+	}
+	for (const [keyEncoded, shapeIndex] of spec.fields ?? []) {
```
CraigMacomber: A code comment on these loops would be helpful. This one could be something like:

```suggestion
	// Add all fields for all overrides from spec that are new fields, in the order they are specified
	for (const [keyEncoded, shapeIndex] of spec.fields ?? []) {
```

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,120 @@ export function decode(
...
+ * For `fields`, override entries whose key matches a key in `base.fields` replace that entry's
+ * shape index in place; override entries with keys not present in `base` are appended at the end
+ * in `spec.fields` order. Field order from `base` is preserved.
+ *
+ * @remarks
+ * Exported for testing.
+ */
+export function applySpecialization(
+	base: EncodedNodeShape,
+	spec: EncodedSpecializedNodeShape,
```
CraigMacomber: "spec" is an ambiguous abbreviation (and we try to avoid abbreviations in the first place). Most people would assume this means specification, but here it might mean specialization.

Renaming this to "overrides" (and "overrides" below to "fieldOverrides") would be more clear.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,120 @@ export function decode(
...
+	context: DecoderContext<EncodedChunkShape>,
+): EncodedNodeShape {
+	const overrides = new Map<FieldKey, number>();
+	for (const [keyEncoded, shapeIndex] of spec.fields ?? []) {
```
CraigMacomber: This algorithm with its three loops seems like it could be simplified, and made a bit mor intuitive if you refactored it.

Copy the base fields array: `const fields = [...base.fields]`;
Make a lookup for where to override: "const indexFromKey = new Map(base.fields.map(...));
Make a pass over spec.fields, either updating items in fields using indexFromKey, or appending to the array.

I feel like that approach is more obviously a correct implementation of the state merge policy.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
CraigMacomber: Also that doc calls these rules "Merge rules": given we have a bunch of logic about merges in this codebase which is about a different thing, calling them specialization rules or override rules would be more clear I think.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,120 @@ export function decode(
...
+		if (!overriddenKeys.has(context.identifier<FieldKey>(keyEncoded))) {
+			mergedFields.push([keyEncoded, shapeIndex]);
+		}
+	}
+
+	return {
+		type: base.type,
+		value: "value" in spec ? spec.value : base.value,
+		fields: mergedFields.length > 0 ? mergedFields : undefined,
+		extraFields: "extraFields" in spec ? spec.extraFields : base.extraFields,
```
CraigMacomber: This is going to break due to JSON not persisting undefined fields if trying to remove optional fields in an override. I think you might need to store some other value (like 0, false or null: 0 is the shortest of those so maybe pick that) to encode that case.

The fact you missed this implies that there might be a gap in test coverage, either for this case and/or not testing with the JSON round tripping.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +79,120 @@ export function decode(
...
+ * to a {@link NodeDecoder} built from the merged shape.
+ */
+export class SpecializedNodeDecoder implements ChunkDecoder {
+	private readonly inner: NodeDecoder;
+	public constructor(
+		shape: EncodedSpecializedNodeShape,
+		context: DecoderContext<EncodedChunkShape>,
+	) {
+		this.inner = new NodeDecoder(
+			applySpecialization(normalizeToNodeShape(shape.base, context), shape, context),
```
CraigMacomber: can't you replace this with a single call to normalizeToNodeShape on shape? (Might be work tweaking the signature for normalizeToNodeShape to enable that)

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -76,9 +78,129 @@ export function decode(
...
+		type: base.type,
+		value: resolveOverride(overrides.value, base.value),
+		fields: fields.length > 0 ? fields : undefined,
+		extraFields: resolveOverride(overrides.extraFields, base.extraFields),
+	};
+}
+
+// `undefined` means the override is absent (inherit from base); `null` is the explicit-clear
+// sentinel needed because JSON.stringify drops `undefined`-valued properties, making
+// property-presence indistinguishable from absent on the wire.
```
CraigMacomber: Use a tsdoc comment, and put these details in the `@param override` section.

CraigMacomber (approved):

PR #27202, opened 2026-04-29
Transaction labels for undo redo bug fix

daesunp:
## Description

Fixes a bug when you have pass a label into a transaction, once you perform an undo, the resulting revertible no longer has the label.


packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
```diff
@@ -1454,6 +1454,174 @@ describe("SchematizingSimpleTreeView", () => {
...
+			assert(revertible !== undefined);
+			revertible.revert();
+
+			assert.equal(events.length, 2);
+			assert.equal(events[1]?.kind, CommitKind.Undo);
+			assert.equal(events[1]?.label, testLabel);
+			assert.deepEqual(events[1]?.labels.tree, { label: testLabel, sublabels: [] });
+		});
+
+		it("redo of revert reuses the original label tree", () => {
```
Josmithr: Nit: I think "revert of revert" is more accurate

Josmithr (approved):
Looks great! Thanks for the fix!

Definitely look at Copilot's suggestions before merging though :)

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1364,12 +1380,20 @@ export class TreeCheckout implements ITreeCheckout {
...
+		try {
+			this.#transaction.activeBranch.apply(
+				change,
+				kind === CommitKind.Default || kind === CommitKind.Redo
+					? CommitKind.Undo
+					: CommitKind.Redo,
+			);
+		} finally {
+			this.labelTreeNode = undefined;
+		}
```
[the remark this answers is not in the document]
daesunp: Updated to keep track of `previousLabelTreeNode` to restore instead of setting to `undefined`

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -726,11 +726,14 @@ export class TreeCheckout implements ITreeCheckout {
...
 							}
+							// Capture the label tree active when the commit was produced so that the
+							// resulting revert commit can inherit the same transaction labels.
 							const revertible = this.createRevertible(
 								revision,
 								kind,
 								this,
 								onRevertibleDisposed,
+								this.labelTreeNode,
 							);
```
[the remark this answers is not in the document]
daesunp: added `cloneLabelTree` function to pass in deep clone instead of the labelTree directly

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
daesunp: updated with suggestion

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -713,6 +726,10 @@ export class TreeCheckout implements ITreeCheckout {
 				const kind = event.type === "append" ? event.kind : CommitKind.Default;
 				const { change, revision } = commit;
 
+				// Snapshot the label tree for this commit before any listener runs.
+				const commitLabelTree =
```
Josmithr, line 733: Should we do this inside `getRevertible` to avoid the clone cost when `getRevertible` isn't called?

Josmithr: `getRevertible` can't be called more than once, so it shouldn't result in any duplicated computation (I think)

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -134,6 +134,19 @@ function* collectTreeLabels(node: LabelTree): IterableIterator<unknown> {
 	}
 }
 
+/**
+ * Deep-clones a {@link LabelTree} so the result is independent of the source.
+ * Used when capturing the label tree on a revertible so that subsequent mutations
```
Josmithr, line 140: Nit (since everything after the first line describes usage rather than being part of the semantic description)
```suggestion
 * @remarks
 * Used when capturing the label tree on a revertible so that subsequent mutations
```

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Updated with suggestion

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated with suggestion

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: I ended up moving this back to outside the factory, as I realized it doesn't cover one of the issues copilot brought up if the labelTree was mutated during a `changed` emit. I added a unit test to cover this scenario as well.

I also realized that if we have a separately started transaction inside the changed event with a label, that label would be nested into the just closed labelTree as we do not clear it before the changed event is emitted. I don't think this is what we want (I think two separate labelTree roots would make more sense). I added a TODO comment to follow up and fix this as well.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -726,7 +727,10 @@ export class TreeCheckout implements ITreeCheckout {
 				const kind = event.type === "append" ? event.kind : CommitKind.Default;
 				const { change, revision } = commit;
 
-				// Snapshot the label tree for this commit before any listener runs.
+				// Snapshot the label tree for this commit before any listener runs, so the captured
+				// value is stable against in-place mutations to `this.labelTreeNode` (e.g. by a
```
Josmithr: Nit: I think we can remove the note about re-entrant edits, since we don't permit those.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1008,13 +1030,16 @@ export class TreeCheckout implements ITreeCheckout {
 	 * @param kind - The {@link CommitKind} that produced this revertible (e.g., Default, Undo, Redo).
 	 * @param checkout - The {@link TreeCheckout} instance this revertible belongs to.
 	 * @param onRevertibleDisposed - Callback function that will be called when the revertible is disposed.
+	 * @param labelTree - The {@link LabelTree} (if any) active when the original commit was produced.
+	 * The revert commit inherits these labels so that hosts can scope undo/redo by transaction label.
```
Josmithr: Nit: I might generalize this a bit to say something like "The revert commit inherits these labels so that commits and their reverts can be associated." (or something)

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1364,12 +1399,23 @@ export class TreeCheckout implements ITreeCheckout {
...
 		}
 
-		this.#transaction.activeBranch.apply(
-			change,
-			kind === CommitKind.Default || kind === CommitKind.Redo
-				? CommitKind.Undo
-				: CommitKind.Redo,
-		);
+		// Install the original commit's label tree so the revert commit's metadata inherits
+		// the same labels — letting hosts scope undo/redo by label. Reusing the captured tree
```
Josmithr: Nit: Same feedback as https://github.com/microsoft/FluidFramework/pull/27202/changes#r3229048088

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1364,12 +1399,23 @@ export class TreeCheckout implements ITreeCheckout {
...
+			this.#transaction.activeBranch.apply(
+				change,
+				kind === CommitKind.Default || kind === CommitKind.Redo
+					? CommitKind.Undo
+					: CommitKind.Redo,
+			);
+		} finally {
+			// Restore rather than clear: a revert triggered from inside another labeled
+			// transaction's changed event must not wipe that outer commit's label state.
+			this.labelTreeNode = previousLabelTreeNode;
```
Josmithr, line 1416: Nit: if possible and we don't already have one, a unit test that verifies this is restored correctly when transaction application fails would be good to add.

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
```diff
@@ -1454,6 +1454,229 @@ describe("SchematizingSimpleTreeView", () => {
...
+				}
+			});
+
+			assert(sourceRevertible !== undefined);
+			sourceRevertible.clone(targetView).revert();
+
+			assert.deepEqual(undoLabels?.tree, { label: testLabel, sublabels: [] });
+		});
+
+		it("revertible's captured labels survive a re-entrant runTransaction from a changed listener", () => {
```
Josmithr: I think you said you were going to revert this, since you're about to publish a separate PR that prevents this case?

Josmithr (approved):
Left a couple more nitpicks, but otherwise looks good to me! The UndoRedo code looks much cleaner now! Thanks!!

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Updated with suggestion

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Updated with suggestion

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
daesunp: reverted

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Added unit test

Josmithr (approved):

PR #26994, opened 2026-04-10
refactor(client-presence): split into -definitions and -runtime

jason-ha:
# Client SDK

refactor most of `presence` into two sub, internal packages:
 - presence-definitions - common interfaces for API and between internal layers
 - presence-runtime - implementations

## presence-runtime source organization

```
        @fluidframework/presence
                  |
        +---------^---------+-------+
        |                   |       |
    ./states           ./runtime    |
        |                   |       |
        ^---------v---------+       |
        |         |         |       |
        |   ./workspace     |       |
        |         |         |       |
        ^-------v-+---------^       |
        |       |           |       |
        |   ./utils         |       |
        |       |           |       |
        ^-------^---v-------^-------^
                    |
    @fluid-internal/presence-definitions
```

### utils

**Purpose:** Utility functions and shared implementations.

**Files:**
- `internalUtils.ts` → `objectEntries`, `objectKeys`, `getOrCreateRecord`, JSON helpers
- `timerManager.ts` → `TimerManager` class
- `broadcastControls.ts` → `OptionalBroadcastControl`, `RequiredBroadcastControl` classes
- `valueManager.ts` → `brandIVM`, `unbrandIVM`

### workspace

**Purpose:** Workspace abstractions and state datastore contracts.

**Files:**
- `stateDatastore.ts` → `StateDatastore` interface, `handleFromDatastore`, `datastoreFromHandle`
   - Split from packages/framework/presence/src/stateDatastore.ts
- `presenceStates.ts` → `PresenceStatesImpl`, `createPresenceStates`, workspace implementation

### runtime

**Purpose:** Core runtime - presence manager, datastore manager, system workspace.

**Exposes:** extension elements `ContainerPresenceFactory` and `extensionId`

**Files:**
- `extension/containerPresence.ts` → `ContainerPresenceFactory`, `extensionId`
    - Split from packages/framework/presence/src/getPresence.ts
- `presenceDatastoreManager.ts` → `PresenceDatastoreManagerImpl`
- `presenceManager.ts` → `PresenceManager` class, `createPresenceManager`
- `protocol.ts` → Message type definitions, `joinMessageType`, `datastoreUpdateMessageType`, `acknowledgementMessageType`
- `runtimeTypes.ts` → `IEphemeralRuntime`, `ExtensionHost`
- `systemWorkspace.ts` → `SystemWorkspace`, `createSystemWorkspace`

### states

**Purpose:** State manager implementations.

**Exposes:** `StateFactory` and `Notifications` (factory)

[...]

PR #27203, opened 2026-04-29
build: update type test baselines to 2.100.0 on main

tylerbutler:
## Description

Post-release type test baseline update: reset and regenerate type test baselines on `main` against the 2.100.0 client release.

Type tests were disabled (`typeValidation.disabled: true`) for two packages because neither has a viable previous-version entrypoint to generate tests against this cycle:

- **`fluid-framework`** — the default `legacy/alpha` entrypoint is not exported by `fluid-framework@2.100.0` (or by the current package).
- **`@fluid-internal/presence-definitions`** — `@fluid-internal/presence-definitions@2.100.0` does not export `beta`. The package was split out from `@fluidframework/presence` post-2.92 (#26994), and the new `-definitions` package only exposes `.` and `./internal/*`.

## Reviewer Guidance


packages/dds/merge-tree/src/test/types/validateMergeTreePrevious.generated.ts
```diff
@@ -7,7 +7,7 @@
  * THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
  * Generated by "flub generate typetests" from @fluid-tools/build-cli.
  *
- * Baseline (previous) version: 2.92.0
+ * Baseline (previous) version: 2.100.0
```
anthony-murphy, line 10: since when did we inline this?

anthony-murphy (approved):

PR #27206, opened 2026-04-29
refactor(quill-react): Update the formatted text editor component to better mirror the plain text editor component

Josmithr:
Leverages `withMemoizedTreeObservations`.

### DIFF
--- a/packages/framework/quill-react/src/formatted/quillFormattedView.tsx
+++ b/packages/framework/quill-react/src/formatted/quillFormattedView.tsx
@@ -433,9 +433,8 @@ const FormattedTextEditorView = withMemoizedTreeObservations(
 				isUpdating.current = true;
 
 				// Wrap all tree mutations in a transaction so they undo/redo as one atomic unit.
-				// If the node is not part of a branch (e.g. unhydrated), apply edits directly.
 				const context = TreeAlpha.context(root);
-				const applyDelta = (): void => {
+				context.runTransaction(() => {
 					// Helper to count Unicode codepoints in a string
 					const codepointCount = (s: string): number => [...s].length;
 
@@ -541,9 +540,7 @@ const FormattedTextEditorView = withMemoizedTreeObservations(
 							cpPos += codepointCount(op.insert);
 						}
 					}
-				};
-
-				context.runTransaction(applyDelta);
+				});
 				isUpdating.current = false;
 			};
 
--- a/packages/framework/quill-react/src/formatted/quillFormattedView.tsx
+++ b/packages/framework/quill-react/src/formatted/quillFormattedView.tsx
@@ -6,21 +6,14 @@
 import { assert } from "@fluidframework/core-utils/internal";
 import {
 	type PropTreeNode,
-	unwrapPropTreeNode,
+	withMemoizedTreeObservations,
 	type UndoRedo,
 } from "@fluidframework/react/internal";
-import { Tree, TreeAlpha, FormattedTextAsTree } from "@fluidframework/tree/internal";
+import { TreeAlpha, FormattedTextAsTree } from "@fluidframework/tree/internal";
 export { FormattedTextAsTree } from "@fluidframework/tree/internal";
 import Quill, { type EmitterSource } from "quill";
 import DeltaPackage from "quill-delta";
-import {
-	forwardRef,
-	useEffect,
-	useImperativeHandle,
-	useReducer,
-	useRef,
-	useState,
-} from "react";
+import { type FC, useEffect, useRef, useState } from "react";
 import * as ReactDOM from "react-dom";
 
 // Workaround for quill-delta's export style not working well with node16 module resolution.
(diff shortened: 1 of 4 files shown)


packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -371,229 +371,215 @@ export function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp
...
-		undo: () => undoRedo?.undo(),
-		redo: () => undoRedo?.redo(),
-	}));
-
-	// Initialize Quill editor with formatting toolbar using Quill provided CSS
-	useEffect(() => {
-		if (!editorRef.current || quillRef.current) {
-			return;
-		}
+const FormattedTextEditorView = withMemoizedTreeObservations(
```
Josmithr, line 378: Note for reviewers: I recommend enabling the "Hide whitespace" option while reviewing the changes below. Much of this is indentation changes.

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -371,229 +371,215 @@ export function buildDeltaFromTree(root: FormattedTextAsTree.Tree): QuillDeltaOp
...
-						// Advance by inserted content length
-						utf16Pos += op.insert.length;
-						cpPos += codepointCount(op.insert);
 					}
+				};
+				if (context.isBranch()) {
+					context.runTransaction(applyDelta);
+				} else {
+					applyDelta();
 				}
```
CraigMacomber: As long as you don't need rollback /abort (and we don't), you can just do:
```suggestion
				context.runTransaction(applyDelta);
```

Since this is possible now, you could also just inline the applyDelta as well. Comment on `context` above also needs an update.

brrichards (commented):
I think copilot made a couple good suggestions of the handful it provided, but these changes overall look good!

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
Josmithr: Nice! I'll update all the relevant places.

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -33,28 +26,26 @@ const Delta = DeltaPackage.default;
...
 }
 
-/**
- * Ref handle exposing undo/redo methods for the formatted editor.
- * @input @internal
- */
-export type FormattedEditorHandle = Pick<UndoRedo, "undo" | "redo">;
+type FormattedMainViewPropsInner = Omit<FormattedMainViewProps, "root"> & {
+	readonly root: FormattedTextAsTree.Tree;
+};
```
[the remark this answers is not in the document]
Josmithr: This is the correct typing for the inner lambda passed to `withMemoizedTreeObservations`.

packages/framework/quill-react/src/index.ts
```diff
@@ -8,5 +8,4 @@ export {
 	FormattedTextAsTree,
 	FormattedMainView,
```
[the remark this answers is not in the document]
Josmithr: We're fine removing internal APIs

packages/framework/quill-react/src/formatted/quillFormattedView.tsx
```diff
@@ -33,28 +26,26 @@ const Delta = DeltaPackage.default;
...
  * @internal
  */
-export const FormattedMainView = forwardRef<FormattedEditorHandle, FormattedMainViewProps>(
-	({ root, undoRedo }, ref) => {
-		return <FormattedTextEditorView root={root} undoRedo={undoRedo} ref={ref} />;
-	},
-);
+export const FormattedMainView: FC<FormattedMainViewProps> = ({ root, undoRedo }) => {
+	return <FormattedTextEditorView root={root} undoRedo={undoRedo} />;
+};
```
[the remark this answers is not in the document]
Josmithr: This is the correct typing for the inner lambda passed to `withMemoizedTreeObservations`.

packages/framework/quill-react/src/test/textEditor.test.tsx
```diff
@@ -1064,5 +1041,49 @@ describe("textEditor", () => {
...
+					it("undo and redo buttons are disabled when undoRedo is not provided", () => {
+						const { tree } = createFormattedTreeView();
+						const rendered = render(<FormattedMainView root={toPropTreeNode(tree)} />, {
+							reactStrictMode,
+						});
+
+						const undoButton = rendered.container.querySelector<HTMLButtonElement>(".ql-undo");
+						const redoButton = rendered.container.querySelector<HTMLButtonElement>(".ql-redo");
+						assert.ok(undoButton !== null, "Undo button should exist");
+						assert.ok(redoButton !== null, "Redo button should exist");
```
CraigMacomber: Personally I prefer:

```suggestion
						assert(undoButton !== null, "Undo button should exist");
						assert(redoButton !== null, "Redo button should exist");
```

packages/framework/quill-react/src/test/textEditor.test.tsx
```diff
@@ -1064,5 +1041,49 @@ describe("textEditor", () => {
...
+						assert.ok(undoButton.disabled === true, "Undo button should be disabled");
+						assert.ok(redoButton.disabled === true, "Redo button should be disabled");
```
CraigMacomber: Personally I prefer letting the assert util generate messages for equality as it includes what the value was if not expected,

```suggestion
						assert.equal(undoButton.disabled, true);
						assert.equal(redoButton.disabled, true);
```

CraigMacomber: same below

packages/framework/quill-react/src/test/textEditor.test.tsx
CraigMacomber: same below

CraigMacomber (approved):

packages/framework/quill-react/src/test/textEditor.test.tsx
Josmithr: Claude seems to have copied this from the other tests in this file. We should probably clean those up too (separately). These tests are now updated.

packages/framework/quill-react/src/test/textEditor.test.tsx
Josmithr: In this case, `disabled` isn't optional, so just asserting `!undoButton.disabled` is simpler. I went with that. And for booleans like this, I think the error semantic error message is more useful than "Expected false but was true`, so I'd prefer to keep it.

brrichards (approved):

PR #27207, opened 2026-04-29
feat(telemetry-utils): deprecate LogLevel.default/error and tag unspecified events as essential

### LINKED ISSUE #26969: Remove deprecated LogLevel.default and LogLevel.error
## Summary

`LogLevel.default` and `LogLevel.error` in `@fluidframework/core-interfaces` are deprecated in favor of the semantically clearer `LogLevel.info` and `LogLevel.essential` respectively.

This issue tracks their removal in the v3.0 major release.

## Details

- `LogLevel.default` (= 20) → replaced by `LogLevel.info`
- `LogLevel.error` (= 30) → replaced by `LogLevel.essential`

The new values were added in PR #26947 alongside deprecation of the old values.

MarioJGMsoft:
## Description

Tracks issue #26969 (planned removal in v3.0).

This PR has two parts:

### 1. Deprecate `LogLevel.default` and `LogLevel.error`

Adds `@deprecated` tags on `LogLevel.default` and `LogLevel.error` and migrates
all in-repo callsites:

 Adds `@deprecated` tags on `LogLevel.default` and `LogLevel.error` in  `@fluidframework/core-interfaces` and migrates all in-repo callsites:
                                                                                                                                                                                                                                                                                                  
  - `LogLevel.default` → context-dependent (see changeset / #26969):                                                                                                                                                                                                                              
    - As an event's `logLevel` argument: `LogLevel.essential`                                                                                                                                                                                                                                     
    - As a logger's `minLogLevel` threshold: `LogLevel.info`                                                                                                                                                                                                                                      
  - `LogLevel.error`   → `LogLevel.essential`   

Affected files include `telemetry-utils` (`logger.ts`, `mockLogger.ts`, `telemetryTypes.ts`, and the related tests) and the corresponding API reports.

### 2. Treat events with no explicit `logLevel` as `essential` in `TelemetryLogger`

Two related changes in `packages/utils/telemetry-utils/src/logger.ts`:

- `ChildLogger.shouldFilterOutEvent` now falls back to `LogLevel.essential` (`30`) when an event arrives without a `logLevel` (previously `LogLevel.default` = `20`).
- `ChildLogger.send` now forwards `LogLevel.essential` to the base logger when no `logLevel` was supplied, instead of forwarding `undefined`.

Additionally, the parameter defaults on `TelemetryLogger.sendTelemetryEvent` and `sendPerformanceEvent` were dropped (they were `= LogLevel.default`). Callers that don't pass a `logLevel` now propagate `undefined` through the chain, where `ChildLogger` applies the new `essential` fallback above.

### 3. `createSampledLogger` and `MultiSinkLogger.send` now forward `error` and `logLevel`

Two pre-existing bugs surfaced while wiring up the `essential` fallback above:

- **`createSampledLogger`** (`packages/utils/telemetry-utils/src/utils.ts`) silently dropped `error` and `logLevel` arguments before delegating to the wrapped logger — the wrapper's method signatures didn't even accept those arguments. The wrapper now accepts and forwards them:
  - `send` accepts `logLevel?`
  - `sendTelemetryEvent` accepts `error?` and `logLevel?` (`verbose | info`)
  - `sendErrorEvent` accepts `error?`
  - `sendPerformanceEvent` accepts `error?` and `logLevel?` (`verbose | info`)

- **`MultiSinkLogger.send`** (`packages/utils/telemetry-utils/src/logger.ts`) had the same bug — `logLevel` was dropped before fanning out to sinks. It now accepts `logLevel?` and forwards it to every registered sink (defaulting to `LogLevel.essential` when omitted, consistent with the part 2 behavior).

### New test coverage:
 - `telemetryLogger.spec.ts` — new `describe("logLevel forwarding")` covering `sendTelemetryEvent`, `sendPerformanceEvent`, and `sendErrorEvent`: default fallback to `essential`, explicit `info` forwarding, and `category: "error"` upgrading to `essential` even when `info` was requested. (These tests live on the parent class because `sendTelemetryEvent` / `sendPerformanceEvent`  are implemented there, not on `ChildLogger`.)                                                                                                                                                                                                                                                 
  - `childLogger.spec.ts` — `describe("logLevel forwarding")` verifies `ChildLogger.send` propagates the level through to the underlying sink (`essential` default, explicit `info`, and `verbose` once the sink opts in via `minLogLevel`).                                                                                                                                                                                                                                                                           
  - `multiSinkLogger.spec.ts` — `describe("logLevel propagation")` verifies the `essential` fallback, explicit-level forwarding for both  `sendTelemetryEvent` and `sendPerformanceEvent`, and includes a regression  test for the `MultiSinkLogger.send` drop bug.                                                                                                                                                                                                                                                 
  - `utils.spec.ts` — `describe("logLevel forwarding")` verifies the sampled  logger forwards `logLevel` through `send`, `sendTelemetryEvent`, and `sendPerformanceEvent`, and propagates `undefined` when the caller omits it.                                                                                                                                                                                                                  
                                                                                         
### Misc

- Fixed an incorrect `type` import ordering in `container.spec.ts`.

## Behavior change for downstream consumers

Although no public API is removed in this PR, downstream consumers that filter telemetry by numeric level should be aware:

Internal Fluid events that were previously emitted to the base logger with
`logLevel = undefined` (effectively treated as `20`) will now arrive tagged as
`30` (`essential`). If a host application starts dropping `logLevel = 20` events
as "non-essential", it must ensure **all** running Fluid versions include this
change — older Fluid versions still send those same critical events at level
`20`, and dropping them would lose telemetry.

## Breaking Changes

No public API surface is removed. See #26969 for the planned v3.0 removal of `LogLevel.default` and `LogLevel.error`. See the "Behavior change for downstream consumers" section above for a runtime behavior note.

## Reviewer Guidance

Specific asks:
- Confirm the `LogLevel.essential` fallback in `ChildLogger.shouldFilterOutEvent` and `ChildLogger.send` is the intended semantic (vs. leaving `undefined` to flow to the base logger).
- Confirm dropping the `= LogLevel.default` parameter defaults on `sendTelemetryEvent` / `sendPerformanceEvent` is acceptable for consumers who may have relied on the implicit default.
- Confirm the `createSampledLogger` and `MultiSinkLogger.send` argument forwarding fix is acceptable as part of this PR (it changes observable behavior for hosts wrapping with these helpers).


packages/loader/container-loader/src/test/container.spec.ts
```diff
@@ -9,7 +9,8 @@ import {
 	TypedEventEmitter,
 	type IProvideLayerCompatDetails,
 } from "@fluid-internal/client-utils";
-import { AttachState, type IAudience } from "@fluidframework/container-definitions";
```
markfields, line 12: nit: can revert this file?

packages/common/core-interfaces/src/logger.ts
```diff
@@ -120,13 +122,13 @@ export interface ITelemetryBaseLogger {
 	/**
 	 * Log a telemetry event, if it meets the appropriate log-level threshold (see {@link ITelemetryBaseLogger.minLogLevel}).
 	 * @param event - The event to log.
-	 * @param logLevel - The log level of the event. Default: {@link LogLevelConst.default | LogLevel.default}.
+	 * @param logLevel - The log level of the event. Default: {@link LogLevelConst.info | LogLevel.info}.
```
markfields: Isn't the default supposed to be essential?  Maybe it's clearer if we say it like "suggested interpretation of undefined: LogLevel.essential"?  Since this is an interface so claiming a "default value" is kind of in vain, it could be implemented any way.

packages/loader/container-loader/src/test/container.spec.ts
MarioJGMsoft: It didn't feel big enough to deserve it's own PR, so I decided to add it here, but I can definitely split it off for it's own thing.

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -230,12 +230,15 @@ export abstract class TelemetryLogger implements TelemetryLoggerExt {
 	public sendTelemetryEvent(
 		event: ITelemetryGenericEventExt,
 		error?: unknown,
-		logLevel: typeof LogLevel.verbose | typeof LogLevel.default = LogLevel.default,
+		logLevel:
+			| typeof LogLevel.verbose
+			| typeof LogLevel.info
+			| typeof LogLevel.essential = LogLevel.essential,
```
jason-ha: These cannot be changes to be different from the definition in `TelemetryLoggerExt` and `ITelemetryLoggerExt`.
TypeScript's silly function variance allowance strikes again. :(

MarioJGMsoft: Applied change, should work properly now

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/logger.ts
jason-ha: `TelemetryLoggerExt` versions should probably have the error note. The text should also be updated to say "essential" instead of "an error".

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -807,7 +810,7 @@ export class PerformanceEvent {
 		event: ITelemetryGenericEventExt,
 		private readonly markers: IPerformanceEventMarkers = { end: true, cancel: "generic" },
 		private readonly emitLogs: boolean = true,
-		private readonly logLevel?: typeof LogLevel.verbose | typeof LogLevel.default,
+		private readonly logLevel?: typeof LogLevel.verbose | typeof LogLevel.info,
```
jason-ha: This is a protected ctor.
So, we should be able to make `logLevel` required and the type can be `typeof LogLevel.verbose | typeof LogLevel.info | undefined`. (We can't quite make the type `LogLevel` yet - need to be able to change TelemetryLoggerExt first.)
I think constructor can also be made private. (Adds clarity.)

packages/utils/telemetry-utils/src/telemetryTypes.ts
```diff
@@ -183,12 +183,12 @@ export interface TelemetryLoggerExt extends ITelemetryBaseLogger {
 	 * Send an information telemetry event.
 	 * @param event - Event to send.
 	 * @param error - Optional error object to log.
-	 * @param logLevel - Optional level of the log. Default: {@link @fluidframework/core-interfaces#LogLevel.default}.
+	 * @param logLevel - Optional level of the log. If undefined, the logLevel should be treated as {@link @fluidframework/core-interfaces#LogLevelConst.essential | LogLevel.essential}.
```
jason-ha: In this version of interface, we are the only implementors. So, we can document this for the callers. Can say "will be treated as". Or the prior "Default:" text was fine.

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
jason-ha, line 41: What is the logic for which `send` calls get LogLevel and which don't? Why change at all?

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
```diff
@@ -269,7 +273,40 @@ describe("ChildLogger", () => {
...
+		});
+
+		it("Forwards explicit LogLevel.info to the sink via `sendTelemetryEvent`", () => {
+			const { sink, recorded } = createRecordingSink();
+			const child = createChildLogger({ logger: sink });
+
+			child.sendTelemetryEvent({ eventName: "chainExplicit" }, undefined, LogLevel.info);
+
+			assert.strictEqual(recorded[0]?.logLevel, LogLevel.info);
+		});
```
jason-ha: Looking over the code, `ChildLogger` doesn't directly implement `sendTelemetryEvent`. That is all handled by the parent class `TelemetryLogger` that is modified in here, but none of its tests were.
`TelemetryLogger` tests should be updated and test `sendPerformanceEvent` in addition to `sendTelemetryEvent`.

.changeset/salty-colts-appear.md
```diff
@@ -0,0 +1,15 @@
...
+"__section": deprecation
+---
+
+Deprecated LogLevel.default and LogLevel.error
+
+`LogLevel.default` and `LogLevel.error` in `@fluidframework/core-interfaces` are deprecated in favor of the semantically clearer `LogLevel.info` and `LogLevel.essential` respectively.
+
+#### Migration
+
+- `LogLevel.default` (= `20`) → use `LogLevel.info`
```
jason-ha: I keep thinking this one should be nuanced. In the context of use are an event's `logLevel` the recommendation is to use `essential`. In the context of `minLogLevel`, `info` is right.
Since we can change guidance and this is not super simple, I would direct readers to the tracking issue.

jason-ha (commented):
A few docs and test suggestions. Thanks!

.changeset/salty-colts-appear.md
MarioJGMsoft: That makes sense, we want to communicate how if they're changing the `logLevel` of an event they should use `essentail` and if they change the `logLevel` of `minLogLevel` they should keep it as `info`

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
MarioJGMsoft: This is a mistake on my part, it's part of the change where the send in childLogger had to contain a required logLevel, I missed changing it back. Thanks for the catch

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
MarioJGMsoft: You're right, I decided to completely remove those events and add more events to the `telemetryLogger.spec.ts`

packages/utils/telemetry-utils/src/telemetryTypes.ts
MarioJGMsoft: Applied change to docs

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: Applied change to code

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: Applied change to docs

.changeset/salty-colts-appear.md
```diff
@@ -5,11 +5,13 @@
...
 Deprecated LogLevel.default and LogLevel.error
 
-`LogLevel.default` and `LogLevel.error` in `@fluidframework/core-interfaces` are deprecated in favor of the semantically clearer `LogLevel.info` and `LogLevel.essential` respectively.
+`LogLevel.default` and `LogLevel.error` in `@fluidframework/core-interfaces` are deprecated in favor of the semantically clearer `LogLevel.info` and `LogLevel.essential`.
 
 #### Migration
 
-- `LogLevel.default` (= `20`) → use `LogLevel.info`
-- `LogLevel.error` (= `30`) → use `LogLevel.essential`
+The recommended replacement depends on how the value is used:
```
jason-ha: The `default` replacement follows this guidance. `error` should always got to `essential`.

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
```diff
@@ -276,37 +276,4 @@ describe("ChildLogger", () => {
...
-
-		it("Forwards explicit LogLevel.info to the sink via `sendTelemetryEvent`", () => {
-			const { sink, recorded } = createRecordingSink();
-			const child = createChildLogger({ logger: sink });
-
-			child.sendTelemetryEvent({ eventName: "chainExplicit" }, undefined, LogLevel.info);
-
-			assert.strictEqual(recorded[0]?.logLevel, LogLevel.info);
-		});
-	});
```
jason-ha: Some test that level comes through via ChildLogger is still appropriate. If it didn't send that along in its `send` implementation, then we'd want to catch it. Just like the case for MultiSink.

jason-ha:
Please also remember to update your PR description

.changeset/salty-colts-appear.md
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/test/childLogger.spec.ts
MarioJGMsoft: Applied change

.changeset/salty-colts-appear.md
```diff
@@ -9,9 +9,11 @@ Deprecated LogLevel.default and LogLevel.error
 
 #### Migration
 
-The recommended replacement depends on how the value is used:
+The recommended replacement for `LogLevel.default` depends on how the value is used:
 
-- For an **event's `logLevel`** (e.g. the `logLevel` argument to `sendTelemetryEvent` / `sendPerformanceEvent` / `ITelemetryBaseLogger.send`), the recommendation is `LogLevel.essential`.
-- For a logger's **`minLogLevel`** (the threshold that filters events), `LogLevel.info` is the recommendation.
+- For an **event's default `logLevel`** (e.g. the `logLevel` argument to `sendTelemetryEvent` / `sendPerformanceEvent` / `ITelemetryBaseLogger.send`), the recommendation is `LogLevel.essential`.
```
jason-ha: This is a little funny because we don't want anyone implementing `sendTelemetryEvent` / `sendPerformanceEvent`. Just `ITelemetryBaseLogger.send` will do.

.changeset/salty-colts-appear.md
```diff
@@ -9,9 +9,11 @@ Deprecated LogLevel.default and LogLevel.error
...
+- For a logger's **default `minLogLevel`** (the threshold that filters events), `LogLevel.info` is the recommendation.
+
+The replacement for `LogLevel.error` should always be essential.
```
jason-ha: "essential" -> "`LogLevel.essential`" so the replacement is symmetric with the source.

jason-ha (approved):
I'm going to approve - you'll need docs sign-off anyway

.changeset/salty-colts-appear.md
```diff
@@ -0,0 +1,19 @@
+---
+"@fluidframework/core-interfaces": minor
+"__section": deprecation
+---
+
+Deprecated LogLevel.default and LogLevel.error
```
jzaffiro:
```suggestion
Deprecate LogLevel.default and LogLevel.error
```

jzaffiro (approved):
Minor nit, but approved for docs!

.changeset/salty-colts-appear.md
MarioJGMsoft: Applied change

.changeset/salty-colts-appear.md
MarioJGMsoft: Applied change

.changeset/salty-colts-appear.md
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -533,7 +533,7 @@ export class ChildLogger extends TelemetryLogger {
 		if (this.shouldFilterOutEvent(event, logLevel)) {
 			return;
 		}
-		this.baseLogger.send(this.prepareEvent(event), logLevel);
+		this.baseLogger.send(this.prepareEvent(event), logLevel ?? LogLevel.essential);
```
markfields, line 536: What happened to making logLevel required on send for the sub-classes?  Now we have two places in the same flow filling in defaults

packages/utils/telemetry-utils/src/telemetryTypes.ts
```diff
@@ -136,14 +136,15 @@ export interface ITelemetryLoggerExt extends ITelemetryBaseLogger {
 	 * Send an information telemetry event.
 	 * @param event - Event to send.
 	 * @param error - Optional error object to log.
-	 * @param logLevel - Optional level of the log. Default: {@link @fluidframework/core-interfaces#LogLevel.essential}.
+	 * @param logLevel - Optional level of the log. If undefined, the logLevel should be treated as {@link @fluidframework/core-interfaces#LogLevel.essential}.
+	 * If the event's category is `error`, the logLevel will be upgraded to {@link @fluidframework/core-interfaces#LogLevel.essential}.
```
markfields, line 140: super nit:

```suggestion
	 * If the event's category is `error`, the logLevel should be upgraded to {@link @fluidframework/core-interfaces#LogLevel.essential}.
```

packages/utils/telemetry-utils/src/telemetryTypes.ts
```diff
@@ -183,12 +185,13 @@ export interface TelemetryLoggerExt extends ITelemetryBaseLogger {
...
+	 * If the event's category is `error`, the logLevel will be upgraded to {@link @fluidframework/core-interfaces#LogLevelConst.essential | LogLevel.essential}.
 	 */
 	sendTelemetryEvent(
 		event: ITelemetryGenericEventExt,
 		error?: unknown,
-		logLevel?: typeof LogLevel.verbose | typeof LogLevel.default,
+		logLevel?: typeof LogLevel.verbose | typeof LogLevel.info,
 	): void;
 
 	/**
```
markfields, line 197: Add a note here that it's going to use logLevel essential?

markfields (approved):
LGTM overall

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: It couldn't be made there might still be chances that ChildLogger and Multisink receive undefined from old containerRuntime (At least that's what I understood). @jason-ha to confirm

packages/utils/telemetry-utils/src/telemetryTypes.ts
MarioJGMsoft: If you're referring to `sendTelemetryEvent`, that note already exists on line 188.  If you're referring to `sendErrorEvent`, since the change itself doesn't affect the interface, I don't think it's necessary to comment that there.

packages/utils/telemetry-utils/src/telemetryTypes.ts
MarioJGMsoft: So this is referring to the flow in `sendTelemetryEvent` and `sendPerformanceEvent` , where they do this: `event.category === "error" ? LogLevel.essential : (logLevel ?? LogLevel.essential)`, so if it's an error, we will upgrade it to essential no matter what.

ChumpChief (approved):

packages/utils/telemetry-utils/src/logger.ts
jason-ha: @markfields, I think we don't have good guardrails for interfaces that span layers. Right now, these can; so, we should respect the interface spec.
Remember also that TypeScript has a whole with function variance and lets you implement an interface with a more restrictive parameter set and says nothing; not even a warning.

PR #27210, opened 2026-04-30
Update benchmark tool

CraigMacomber:
## Description

Update benchmark tool, adopting new APIs.

## Reviewer Guidance


packages/test/test-end-to-end-tests/src/test/benchmark/DocumentCreator.ts
```diff
@@ -151,21 +151,12 @@ export function benchmarkAll<T extends IBenchmarkParameters>(
...
 					await obj.before?.();
-					let duration: number;
-					do {
-						const before = state.timer.now();
-						await obj.run();
-						const after = state.timer.now();
-						duration = state.timer.toSeconds(before, after);
-						// Collect data
-					} while (state.recordBatch(duration));
+					await state.timeAllBatchesAsync(async () => obj.run());
```
alexvy86, line 156: Shouldn't this one use the new batchless helper? Since it had `minBatchDurationSeconds: 0,` before.

CraigMacomber: You should only use the batchless helper if you can't use the more accurate timing of the batched one. As this test does not actually have any non-timed code between the measurements, its opting out of batching was simply a regression in timing precision.

In practice, since this skips straight to the collect data phase, this will end up with batches of size 1 anyway, but this measurement function has no reason to depend on that.

How it was before it would silently produce incorrect data if ever run with a batch size larger than one, making this test needlessly fragile. Moving to a batchless setup would have fixed it as well, but doing so violates the explicit documentation about when to use the batchless setup, and would force this code to keep its own inner timing loop (since the batchless setup is explicitly for the case where timeAllBatches is impossible, it does not provide it).

Now, if someone decided this code actually isn't super slow and wanted to improve data quality and allow warmup, they will get automatically sized batches and the correspondingly higher data quality.

alexvy86 (approved):

Abe27342 (approved):

PR #27064, opened 2026-04-16
Update Cross-Client Compat Policy

scottn12:
## Description

This PR updates the cross-client compat policy to be time-based. For more details see [Proposal Doc](https://microsoft.sharepoint-df.com/:fl:/g/contentstorage/CSP_35278ae7-2935-4dcc-a64a-054383a5d114/IQDI4Ea7FaAjTqu9MFTUYRBqAbsX7-BOw_Aoeme-Necc3b8?e=tjs8f7&nav=cz0lMkZjb250ZW50c3RvcmFnZSUyRkNTUF8zNTI3OGFlNy0yOTM1LTRkY2MtYTY0YS0wNTQzODNhNWQxMTQmZD1iJTIxNTRvbk5UVXB6RTJtU2dWRGc2WFJGTHRnNlU1NktKbEF1dlFQUlR5R0M3WFdWcHZ3ZndielFyc1RQRDFTNW15NCZmPTAxNEFXUVRRV0k0QkRMV0ZOQUVOSEtYUEpRS1RLR0NFREsmYz0lMkYmYT1Mb29wQXBwJnA9JTQwZmx1aWR4JTJGbG9vcC1wYWdlLWNvbnRhaW5lciZ4PSU3QiUyMnclMjIlM0ElMjJUMFJUVUh4dGFXTnliM052Wm5RdWMyaGhjbVZ3YjJsdWRDMWtaaTVqYjIxOFlpRTFORzl1VGxSVmNIcEZNbTFUWjFaRVp6WllVa1pNZEdjMlZUVTJTMHBzUVhWMlVWQlNWSGxIUXpkWVYxWndkbmRtZDJKNlVYSnpWRkJFTVZNMWJYazBmREF4TkVGWFVWUlJValkwTTBRMU5GbEpWRlpLUlV4TlUxbzFNMGxLV1ZBMVNETSUzRCUyMiUyQyUyMmklMjIlM0ElMjIwYWRmZWU5NC00ZjQwLTRjM2QtOWZmMi0wOWI2MzYwNmQyMmYlMjIlN0Q%3D).

The goal of this PR is to make the in-repo documentation the source of truth for the cross-client compat policy going forward. Follow-up PRs (see below) will align tooling, tests, and release process with the new policy.

## Follow Ups
- Update declarative model scenarios/documentation as necessary
- Update e2e test infrastructure to use checkpoint-based version combinations
- Establish security patch process for checkpoint releases
- Add mechanism to automatically designate checkpoints
[...]

Issue #23289, opened 2024-12-10
Remove `CompatibilityMode` type


markfields:
~~(Someone correct me if I'm wrong)~~

~~We don't intend to support collaboration between 1.x and 3.0 clients.  Both are compatible with 2.x, so users can be migrated over the 2.x timeframe.~~

~~So in 3.0 we can remove "1" from the union type `CompatibilityMode` (maybe should change the structure of that type to make it easier to deprecate/remove in the future), and remove  `compatibilityModeRuntimeOptions["1"]` from https://github.com/microsoft/FluidFramework/blob/main/packages/framework/fluid-static/src/compatibilityConfiguration.ts~~

~~Then we'll be able to clean up the code supporting those old configurations which aren't otherwise used.~~

Update: 

The `CompatibilityMode` type is deprecated and will be removed in 3.0.0. 

Customers should directly pass in a `minVersionForCollaboration` or `minVersionForCollab` going forward.

markfields:
@vladsud - On second thought, I don't think this has to happen at the 3.0.0 moment, what do you think? Here are the options it overrides:

```typescript
"1": {
	// 1.x clients are compatible with TurnBased flushing, but here we elect to remain on Immediate flush mode
	// as a work-around for inability to send batches larger than 1Mb. Immediate flushing keeps batches smaller as
	// fewer messages will be included per flush.
	flushMode: FlushMode.Immediate,
	// Op compression is on by default but introduces a new type of op which is not compatible with 1.x clients.
	compressionOptions: {
		minimumBatchSizeInBytes: Number.POSITIVE_INFINITY, // disabled
		compressionAlgorithm: CompressionAlgorithms.lz4,
	},
	// Grouped batching is on by default but introduces a new type of op which is not compatible with 1.x clients.
	enableGroupedBatching: false,
	// TODO: Include explicit disables for things that are currently off-by-default?

	// Explicitly disable running Sweep in compat mode "1". Sweep is supported only in 2.x. So, when 1.x and 2.x
	// clients are running in parallel, running sweep will fail 1.x clients.
	gcOptions: { enableGCSweep: undefined },
},
```

`flushMode` and `enableGroupedBatching` are being removed from the Legacy-Alpha options list in 2.20 (tracked in sub-issues under #22835). `compressionOptions` and `gcOptions` aren't changing, so it's not like this is blocking some other breaking change.

If you agree, I'll close this and track it in ADO.  cc @jason-ha @ChumpChief

[...]

Issue #23271, opened 2024-12-10
Client 3.0 Breaking Changes


jason-ha:
The attached sub-issues chronicle all of the major (breaking) changes to take place in client 3.0 release.

| | |
| -------------------------- | ------------ |
| List status | Maximal set finalized |
| Estimated Release Date | 2026-08-24 |

Issues removed from set:
#27461 - `createBlobPayloadPending` default behavior will not be changed.
#26500 - There were no beta only breaking changes.
#27180 - Requiring an explicit `oldestSupportedClient` was deferred to the Client 3.10 beta/legacy breaking-change window tracked by #27471.

PR #27212, opened 2026-04-30
Compat: Add minVersionForCollab option to declarative model and deprecate CompatibilityMode

scottn12:
## Description

This PR updates the declarative model to accept any (valid) `minVersionForCollab`. This is intended to replace the existing `CompatibilityMode` type (deprecated in this PR), which accepts either `"1"` or `"2"`. This was done to ensure the declarative model is aligned to support the latest changes to the cross-client compat policy (see https://github.com/microsoft/FluidFramework/pull/27064).

The following are the major changes in this PR:
- `MinimumVersionForCollab` promoted from a beta to public API. This was done so it could be used in public declarative model APIs.
- `CompatibilityMode` param in `createDOProviderContainerRuntimeFactory()` now also accepts`MinimumVersionForCollab` (previously only the `CompatibilityMode` type).
- Deprecated `CompatibilityMode` type (will be removed in 3.0).
- Deprecated `minVersionForCollabOverride` (will be removed in 3.0). 

## Reviewer Guidance
- Promoting `MinimumVersionForCollab` will require API council review. If it's easier, this can be split into a separate PR.

## Misc
[AB#72040](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/72040)


jason-ha:
Rather than keeping these two types separate, where practical in API surface have them as a single input.
- Internally, at the outermost bottleneck filter for `CompatibilityMode` and convert to the appropriate min version.
- In API be sure to note what values are deprecated since most callers are using a `CompatibilityMode` enumeration.
   - Be sure to update in repo use to the non-deprecated pattern. (Do a test build where `CompatibilityMode` doesn't exist. That change can be kept in a branch that stages the breaking change.)

scottn12:
> Rather than keeping these two types separate, where practical in API surface have them as a single input.
> 
> * Internally, at the outermost bottleneck filter for `CompatibilityMode` and convert to the appropriate min version.
> * In API be sure to note what values are deprecated since most callers are using a `CompatibilityMode` enumeration.
>   
>   * Be sure to update in repo use to the non-deprecated pattern. (Do a test build where `CompatibilityMode` doesn't exist. That change can be kept in a branch that stages the breaking change.)

@jason-ha, updated in latest to do this

packages/framework/fluid-static/src/utils.ts
```diff
@@ -147,12 +148,24 @@ export function makeFluidObject<
 }
 
 /**
- * Maps CompatibilityMode to a semver valid string that can be passed to the container runtime.
+ * Resolves the `compatibilityMode` input — either a {@link MinimumVersionForCollab}
+ * semver string or a legacy {@link CompatibilityMode} value — into a precise
+ * {@link MinimumVersionForCollab}.
+ *
+ * TODO: This can be removed when the deprecated CompatibilityMode is removed.
```
jason-ha: Good to have work items associated with TODOs. I don't think this is needed as compile will break when `CompatibilityMode` is removed. Visiting is forced.

packages/framework/fluid-static/src/rootDataObject.ts
```diff
@@ -214,26 +219,30 @@ export function createDOProviderContainerRuntimeFactory(props: {
 	 * If not provided, the default for the given compatibilityMode will be used.
 	 * @remarks
 	 * This is useful when runtime options are overridden and change the minimum version for collab.
+	 *
+	 * @deprecated Pass a {@link @fluidframework/runtime-definitions#MinimumVersionForCollab}
+	 * semver string directly via `compatibilityMode` instead.
 	 */
 	minVersionForCollabOverride?: MinimumVersionForCollab;
```
jason-ha: This function is noted as `@internal`. This can't be cleaned up (removed) now?

packages/framework/fluid-static/src/treeRootDataObject.ts
```diff
@@ -229,26 +231,30 @@ export function createTreeContainerRuntimeFactory(props: {
...
 	const [registryEntries, sharedObjects] = parseDataObjectsFromSharedObjects(schema);
 	const registry = rootDataStoreRegistry ?? new FluidDataStoreRegistry(registryEntries);
 
 	return new TreeContainerRuntimeFactory(
-		compatibilityMode,
 		new TreeRootDataObjectFactory(sharedObjects, registry),
 		{
 			runtimeOptions: runtimeOptionOverrides,
-			minVersionForCollab: minVersionForCollabOverride,
+			minVersionForCollab: minVersionForCollabOverride ?? minVersionForCollab,
```
jason-ha: I think the preferred logic is:
 - if compatibilityMode is semver (not "1" | "2"), then use it independent of minVersionForCollabOverride
 - else minVersionForCollabOverride if defined
 - else compatibilityMode promoted to semantic version

The subtle difference is that minVersionForCollabOverride really is ignored when new pattern is used.

packages/framework/fluid-static/src/compatibilityConfiguration.ts
```diff
@@ -27,3 +27,16 @@ export const compatibilityModeRuntimeOptions: Record<
...
+/**
+ * Determines default runtime options for the given minVersionForCollab.
+ *
+ * @internal
+ */
+export function defaultRuntimeOptionsForMinVersion(
+	minVersionForCollab: MinimumVersionForCollab,
+): IContainerRuntimeOptionsInternal {
+	return minVersionForCollabToDefaultRuntimeOptions[
+		gte(minVersionForCollab, "2.0.0") ? "2" : "1"
```
jason-ha, line 45: feels a little like going backwards.
If I understand correctly, then these are basically for additional runtime options above those that minVersionForCollab gives us from the container-runtime `getMinVersionForCollabDefaults`, yeah?
Maybe we should note that other options are selected by container-runtime.

packages/service-clients/azure-client/src/interfaces.ts
```diff
@@ -53,7 +55,8 @@ export interface AzureClientPropsInternal extends AzureClientProps {
 		compatibilityMode,
 	}: {
 		schema: ContainerSchema;
-		compatibilityMode: CompatibilityMode;
+		// eslint-disable-next-line import-x/no-deprecated
+		compatibilityMode: MinimumVersionForCollab | CompatibilityMode;
```
jason-ha: Not useful to resolve before here? Interface is `@internal`.

packages/service-clients/end-to-end-tests/azure-client/src/test/multiprocess/childClient.tool.ts
jason-ha, line 1: Can all of the tests in here share a common constant. And can that be above 2.0.0?
I don't think any of them are doing compat testing really and could use the current version even.

packages/service-clients/odsp-client/src/odspClient.ts
```diff
@@ -166,7 +166,7 @@ export class OdspClient {
 	private getLoaderProps(schema: ContainerSchema): ILoaderProps {
 		const runtimeFactory = createDOProviderContainerRuntimeFactory({
 			schema,
-			compatibilityMode: "2",
+			compatibilityMode: "2.0.0",
```
jason-ha: Is there planned work to fixup minVersion support for OdspClient?

packages/service-clients/azure-client/src/AzureClient.ts
```diff
@@ -138,12 +141,16 @@ export class AzureClient {
...
+	 * either a `MinimumVersionForCollab` semver string (e.g. `"1.0.0"`, `"2.0.0"`) or a legacy
+	 * {@link @fluidframework/fluid-static#CompatibilityMode} value. The legacy values `"1"` and
+	 * `"2"` are **deprecated** equivalents of `"1.0.0"` and `"2.0.0"`.
 	 * @returns New detached container instance along with associated services.
 	 */
 	public async createContainer<const TContainerSchema extends ContainerSchema>(
 		containerSchema: TContainerSchema,
-		compatibilityMode: CompatibilityMode,
+		// eslint-disable-next-line import-x/no-deprecated
+		compatibilityMode: MinimumVersionForCollab | CompatibilityMode,
```
jason-ha, line 176: I don't see any examples clean up. I think that mostly means `CompatibilityMode` deprecation is mostly only impacting us.
I think to have proper deprecation you want have an override for each of these methods and you deprecate the one that takes `CompatibilityMode`.

jason-ha (commented):
Would be nice to clean up more of the internal APIs but can be done later.

packages/framework/fluid-static/src/rootDataObject.ts
scottn12: Good call - updated in latest. Since it's just internal we don't need to wait until a legacy/beta breaking release, correct?

packages/framework/fluid-static/src/compatibilityConfiguration.ts
scottn12: That is correct. This logic is mostly pre-existing but we needed to add a helper function since we accept more than just "1" and "2" now. I added comments in latest.

packages/framework/fluid-static/src/treeRootDataObject.ts
scottn12: I removed `minVersionForCollabOverride` in latest so this should be simpler now

packages/framework/fluid-static/src/utils.ts
scottn12: Added in latest

packages/service-clients/azure-client/src/AzureClient.ts
scottn12: Should be properly deprecated in latest. Updated examples as well

packages/service-clients/azure-client/src/interfaces.ts
scottn12: Updated in latest

packages/service-clients/end-to-end-tests/azure-client/src/test/multiprocess/childClient.tool.ts
scottn12: Good idea - updated in latest to use `pkgVersion` for tests. Should be nice to catch regressions in the future :)

packages/service-clients/odsp-client/src/odspClient.ts
scottn12: I was originally thinking of doing a separate PR but it probably makes sense to just do it in one PR. Added in latest

packages/framework/fluid-static/src/rootDataObject.ts
jason-ha: Correct. Change internal at any time (unless it cross layer boundaries and this does not).

packages/service-clients/end-to-end-tests/azure-client/src/test/tsconfig.json
```diff
@@ -1,11 +1,11 @@
 {
 	"extends": "../../../../../../common/build/build-common/tsconfig.test.node16.json",
 	"compilerOptions": {
-		"rootDir": "./",
-		"outDir": "../../lib/test",
+		"rootDir": "../",
+		"outDir": "../../lib",
```
jason-ha, line 8: Ha! I just cleaned up to test packages that weren't following the standard pattern ( #27322 ).

Put `gen-version` is quite inflexible. So, this pattern makes sense.
Please just add a comment that this deviates to accommodate `gen-version`.

packages/service-clients/end-to-end-tests/azure-client/src/test/utils.ts
```diff
@@ -8,6 +8,8 @@ import type { ConfigTypes, IConfigProviderBase } from "@fluidframework/core-inte
 import type { IMember } from "@fluidframework/fluid-static";
 import type { ISharedMap, IValueChanged } from "@fluidframework/map/legacy";
 
+export { pkgVersion } from "../packageVersion.js";
```
jason-ha: I would love to have some explanation here.
`gen-version` doesn't make a lot of sense on its own for a test-only package. I think we want to note that it is used to get the current client version.
(A more [logically] ideal value would be something that is shared from FF package; though I don't think we would want to actually try to support that.)
It would also be awesome to rename the constant, which we can do from the export statement with `pkgVersion as minVersionForCollab` or `pkgVersion as currentVersion` or `pkgVersion as latestVersion` or ???

packages/service-clients/end-to-end-tests/azure-client/package.json
jason-ha: This needs to have `build:genver` added.
Will conflict with #27265 but I expect this will merge first as that one is waiting on another PR to merge.

packages/service-clients/odsp-client/src/odspClient.ts
```diff
@@ -111,13 +112,41 @@ export class OdspClient {
...
+	 */
+	public async createContainer<T extends ContainerSchema>(
+		containerSchema: T,
+	): Promise<{
+		container: IOdspFluidContainer<T>;
+		services: IOdspContainerServices;
+	}>;
+	public async createContainer<T extends ContainerSchema>(
+		containerSchema: T,
+		minVersionForCollab: MinimumVersionForCollab = "2.0.0",
```
jason-ha: Can OdspClient work with 1.x?
Instead of `MinimumVersionForCollab` might want to use `2.${bigint}.${bigint}` or
```ts
MinimumVersionForCollab & `2.${string}`  // which requires visit in 3.0
  or
Exclude<MinimumVersionForCollab, `1.${string}`>  // which allow MinV4C to float up to 3
```

packages/service-clients/tinylicious-client/src/interfaces.ts
```diff
@@ -64,7 +64,7 @@ export interface TinyliciousConnectionConfig {
  * Any functionality regarding how the data is handled within the FluidContainer itself (e.g., which data objects or
  * DDSes to use) will not be included here but rather on the FluidContainer class itself.
  *
- * Returned by {@link TinyliciousClient.createContainer} and {@link TinyliciousClient.getContainer} alongside the FluidContainer.
+ * Returned by `TinyliciousClient.createContainer` and `TinyliciousClient.getContainer` alongside the FluidContainer.
```
jason-ha: I think you want to use qualifiers to retain the link.
See https://deepwiki.com/microsoft/tsdoc/2.3-declaration-references#2-syntax-and-grammar
I think adding `:(1)` to what was there before should work.

examples/benchmarks/tablebench/src/azure.ts
jason-ha: I don't trust our linting to catch all cases - here is one that was missed.
Highly recommend a test build with deprecated methods removed. That change can be kept as stash for when the break is made.

.changeset/zesty-otters-collab.md
```diff
@@ -0,0 +1,29 @@
+---
+"@fluidframework/fluid-static": minor
+"fluid-framework": minor
+"__section": legacy
+---
+`createTreeContainerRuntimeFactory` no longer accepts `minVersionForCollabOverride`
+
+The `minVersionForCollabOverride` property on the `props` argument of `createTreeContainerRuntimeFactory` has been removed.
```
jason-ha: That is a beta breaking change. It will need to be kept until 2.110 - [file deprecation issue](https://github.com/microsoft/FluidFramework/wiki/API-Deprecation) under #26499.

packages/service-clients/odsp-client/README.md
```diff
@@ -104,7 +104,7 @@ const containerSchema = {
 	],
 };
 const odspClient = new OdspClient(clientProps);
-const { container, services } = await odspClient.createContainer(containerSchema);
+const { container, services } = await odspClient.createContainer(containerSchema, "2.0.0");
```
jason-ha: Can we use a higher number? We don't want anyone to set 2.0.0 if they could be setting 2.100.0 - especially if onboarding.

packages/framework/fluid-static/src/utils.ts
```diff
@@ -147,12 +148,24 @@ export function makeFluidObject<
 }
 
 /**
- * Maps CompatibilityMode to a semver valid string that can be passed to the container runtime.
+ * Resolves the `compatibilityMode` input — either a `MinimumVersionForCollab`
+ * semver string or a legacy `CompatibilityMode` value — into a precise
+ * `MinimumVersionForCollab`.
+ *
+ * TODO: This can be removed when the deprecated CompatibilityMode is removed - AB#73679
```
jason-ha: nit: I think a preferred format is 
```suggestion
 * TODO: AB#73679: This can be removed when the deprecated CompatibilityMode is removed
```

packages/framework/fluid-static/src/types.ts
```diff
@@ -22,6 +22,12 @@ import type { ITree } from "@fluidframework/tree";
  * In "1" mode we support full interop between 2.x clients and 1.x clients,
  * while in "2" mode we only support interop between 2.x clients.
  *
+ * @deprecated Specify the minimum Fluid Framework version directly via the
```
jason-ha, line 25: [No need to change] I think it would be okay to NOT deprecate `CompatiblityMode` because it isn't a type that customers use - it mostly adds noise our codebase making it look funny when we use it because we have to.
The deprecation is done via the method overrides.

jason-ha (commented):
Looking good
You will need a second change set and 3.0 breaking issue under #23271 to deprecate the *Client methods.
There is a 3.0 issue for [removing 1 from `CompatibilityMode`](https://github.com/microsoft/FluidFramework/issues/23289) and I think that can transition to straight up `CompatibilityMode` removal.

packages/framework/fluid-static/src/compatibilityConfiguration.ts
```diff
@@ -4,19 +4,19 @@
...
 import type { IContainerRuntimeOptionsInternal } from "@fluidframework/container-runtime/internal";
-
-import type { CompatibilityMode } from "./types.js";
+import type { MinimumVersionForCollab } from "@fluidframework/runtime-definitions/internal";
+import { gte } from "semver-ts";
 
 /**
- * The CompatibilityMode selected determines the set of runtime options to use. In "1" mode we support
- * full interop with true 1.x clients, while in "2" mode we only support interop with 2.x clients.
+ * The `minVersionForCollab` determines the set of runtime options to use.
```
Josmithr: Nit: I don't think this description is very accurate. Probably worth reworking this.

packages/framework/fluid-static/src/compatibilityConfiguration.ts
```diff
@@ -4,19 +4,19 @@
...
+ * For a 1.x `minVersionForCollab` we support full interop with true 1.x clients.
```
Josmithr: Nit: these specifics probably belong in a `@remarks` block. The summary block should be reserved for a brief semantic description.

packages/framework/fluid-static/src/rootDataObject.ts
```diff
@@ -196,9 +197,13 @@ export function createDOProviderContainerRuntimeFactory(props: {
 	 */
 	schema: ContainerSchema;
 	/**
-	 * See {@link CompatibilityMode} and compatibilityModeRuntimeOptions for more details.
+	 * Minimum Fluid Framework version required for collaboration. Accepts a
+	 * {@link @fluidframework/runtime-definitions#MinimumVersionForCollab} semver string;
```
Josmithr: Nit
```suggestion
	 * {@link @fluidframework/runtime-definitions#MinimumVersionForCollab} SemVer string;
```

packages/framework/fluid-static/src/rootDataObject.ts
```diff
@@ -263,31 +257,28 @@ class DOProviderContainerRuntimeFactory extends BaseContainerRuntimeFactory {
 	 * since it can take care of constructing the root data object factory based on the schema.
 	 *
 	 * @param schema - The schema for the container
-	 * @param compatibilityMode - Compatibility mode
 	 * @param rootDataObjectFactory - A factory that can construct the root data object.
+	 * @param config - Resolved minimum version for collab (required) and optional runtime option overrides.
```
Josmithr: Nit: It was a miss that we permitted the abbreviation "collab" in the type name for "MinVersionForCollab". Our guidelines generally call to avoid using abbreviations that aren't truly ubiquitous. "Min", for example, is well understood. "Collab" less so.

At the very least, I would request that we expand "collab" to "collaboration" in documentation and any contexts where we aren't referring to the "MinVersionForCollab" type directly.

packages/framework/fluid-static/src/treeRootDataObject.ts
```diff
@@ -211,9 +209,13 @@ export function createTreeContainerRuntimeFactory(props: {
 	readonly schema: TreeContainerSchema;
 
 	/**
-	 * See {@link CompatibilityMode} and compatibilityModeRuntimeOptions for more details.
+	 * Minimum Fluid Framework version required for collaboration. Accepts a
+	 * {@link @fluidframework/runtime-definitions#MinimumVersionForCollab} semver string;
```
Josmithr: Nit
```suggestion
	 * {@link @fluidframework/runtime-definitions#MinimumVersionForCollab} SemVer string;
```

packages/framework/fluid-static/src/types.ts
```diff
@@ -22,6 +22,12 @@ import type { ITree } from "@fluidframework/tree";
...
+ * `minVersionForCollab` parameter, which accepts a
+ * {@link @fluidframework/runtime-definitions#MinimumVersionForCollab} semver string. The
```
Josmithr: Nit
```suggestion
 * {@link @fluidframework/runtime-definitions#MinimumVersionForCollab} SemVer string. The
```

packages/framework/fluid-static/src/utils.ts
```diff
@@ -147,12 +148,24 @@ export function makeFluidObject<
 }
 
 /**
- * Maps CompatibilityMode to a semver valid string that can be passed to the container runtime.
+ * Resolves the `compatibilityMode` input — either a `MinimumVersionForCollab`
+ * semver string or a legacy `CompatibilityMode` value — into a precise
```
Josmithr: Nit
```suggestion
 * SemVer string or a legacy `CompatibilityMode` value — into a precise
```

packages/service-clients/azure-client/src/AzureClient.ts
```diff
@@ -138,15 +141,42 @@ export class AzureClient {
 	 * @typeparam TContainerSchema - Used to infer the the type of 'initialObjects' in the returned container.
 	 * (normally not explicitly specified.)
 	 * @param containerSchema - Container schema for the new container.
-	 * @param compatibilityMode - Compatibility mode the container should run in.
+	 * @param minVersionForCollab - Minimum Fluid Framework version required for collaboration, as a
+	 * `MinimumVersionForCollab` semver string (e.g. `"1.0.0"`, `"2.0.0"`).
```
Josmithr: Nit (PascalCase "SemVer", and omit the typing information, since that's already captured by the type-system)
```suggestion
	 * SemVer string (e.g. `"1.0.0"`, `"2.0.0"`).
```

packages/service-clients/azure-client/src/AzureClient.ts
```diff
@@ -138,15 +141,42 @@ export class AzureClient {
...
 		containerSchema: TContainerSchema,
+		minVersionForCollab: MinimumVersionForCollab,
+	): Promise<{
+		container: IFluidContainer<TContainerSchema>;
+		services: AzureContainerServices;
+	}>;
+	/**
+	 * Creates a new detached container instance in the Azure Fluid Relay.
+	 * @typeparam TContainerSchema - Used to infer the the type of 'initialObjects' in the returned container.
+	 * (normally not explicitly specified.)
```
Josmithr: Nit: 
```suggestion
	 * This does not normally need to be specified explicitly.
```

packages/service-clients/azure-client/src/AzureClient.ts
```diff
@@ -138,15 +141,42 @@ export class AzureClient {
...
+	 * @param containerSchema - Container schema for the new container.
+	 * @param compatibilityMode - Legacy {@link @fluidframework/fluid-static#CompatibilityMode} value.
+	 * @returns New detached container instance along with associated services.
+	 * @deprecated Pass a `MinimumVersionForCollab` semver string (e.g. `"2.0.0"`) instead. The legacy
```
Josmithr: Nit
```suggestion
	 * @deprecated Pass a `MinimumVersionForCollab` SemVer string (e.g. `"2.0.0"`) instead. The legacy
```

packages/service-clients/azure-client/src/AzureClient.ts
```diff
@@ -172,16 +202,46 @@ export class AzureClient {
 	 * (normally not explicitly specified.)
 	 * @param id - Unique ID of the container in Azure Fluid Relay.
 	 * @param containerSchema - Container schema used to access data objects in the container.
-	 * @param compatibilityMode - Compatibility mode the container should run in.
+	 * @param minVersionForCollab - Minimum framework version required for collaboration, as a
+	 * `MinimumVersionForCollab` semver string (e.g. `"1.0.0"`, `"2.0.0"`).
```
Josmithr: Nit
```suggestion
	 * SemVer string (e.g. `"1.0.0"`, `"2.0.0"`).
```

packages/service-clients/azure-client/src/AzureClient.ts
Josmithr: (Same feedback for instances below)

packages/service-clients/azure-client/src/AzureClient.ts
```diff
@@ -210,16 +270,47 @@ export class AzureClient {
 	 * @param id - Unique ID of the source container in Azure Fluid Relay.
 	 * @param containerSchema - Container schema used to access data objects in the container.
 	 * @param version - Unique version of the source container in Azure Fluid Relay.
-	 * @param compatibilityMode - Compatibility mode the container should run in.
+	 * @param minVersionForCollab - Minimum framework version required for collaboration, as a
+	 * `MinimumVersionForCollab` semver string (e.g. `"1.0.0"`, `"2.0.0"`).
```
Josmithr: Nit
```suggestion
	 * SemVer string (e.g. `"1.0.0"`, `"2.0.0"`).
```

packages/service-clients/azure-client/src/AzureClient.ts
```diff
@@ -210,16 +270,47 @@ export class AzureClient {
...
+	/**
+	 * Load a specific version of a container for viewing only.
+	 * @typeparam TContainerSchema - Used to infer the the type of 'initialObjects' in the returned container.
+	 * (normally not explicitly specified.)
+	 * @param id - Unique ID of the source container in Azure Fluid Relay.
+	 * @param containerSchema - Container schema used to access data objects in the container.
+	 * @param version - Unique version of the source container in Azure Fluid Relay.
+	 * @param compatibilityMode - Legacy {@link @fluidframework/fluid-static#CompatibilityMode} value.
 	 * @returns Loaded container instance at the specified version.
+	 * @deprecated Pass a `MinimumVersionForCollab` semver string (e.g. `"2.0.0"`) instead. The legacy
```
Josmithr: Nit
```suggestion
	 * @deprecated Pass a `MinimumVersionForCollab` SemVer string (e.g. `"2.0.0"`) instead. The legacy
```

.changeset/zesty-otters-collab.md
scottn12: Ah I thought that since `createDOProviderContainerRuntimeFactory` was internal we could skip this but I see `createTreeContainerRuntimeFactory` is not internal. Filed an issue - https://github.com/microsoft/FluidFramework/issues/27356

examples/benchmarks/tablebench/src/azure.ts
scottn12: Good idea. Did that and updated the reamining tests/examples in latest

packages/framework/fluid-static/src/compatibilityConfiguration.ts
scottn12: Updated in latest

packages/framework/fluid-static/src/rootDataObject.ts
scottn12: Ah yeah that is unfortunate, wish it was easy to fix :(

This was the only use of "collab" in this PR so it's fixed in latest

packages/framework/fluid-static/src/types.ts
scottn12: Yeah that's true. At least it (hopefully) won't be around as long since this should be removed in 3.0.0.

packages/service-clients/end-to-end-tests/azure-client/src/test/tsconfig.json
scottn12: Haha funny timing. Added a comment in latest

packages/service-clients/end-to-end-tests/azure-client/src/test/utils.ts
scottn12: Yeah makes sense, I added `pkgVersion as currentVersion` here and in the other test file changes

packages/service-clients/end-to-end-tests/azure-client/package.json
scottn12: Sounds good, added in latest

packages/service-clients/odsp-client/src/odspClient.ts
scottn12: Ah good catch - updated in latest to use `Exclude` so we don't have to update it again.

packages/service-clients/odsp-client/README.md
scottn12: Updated in latest

packages/service-clients/tinylicious-client/src/interfaces.ts
scottn12: Fixed in latest

.changeset/zesty-otters-collab.md
scottn12: I ended up pulling out this change into a separate PR so we can merge in the rest of the changes sooner. I'll open a follow up PR to remove it in 2.110.0

packages/framework/fluid-static/src/treeRootDataObject.ts
scottn12: > I think the preferred logic is:
> 
> * if compatibilityMode is semver (not "1" | "2"), then use it independent of minVersionForCollabOverride
> * else minVersionForCollabOverride if defined
> * else compatibilityMode promoted to semantic version
> 
> The subtle difference is that minVersionForCollabOverride really is ignored when new pattern is used.

Updated in latest to use this logic now that `minVersionForCollabOverride` is added back in for now.

.changeset/soft-beds-beam.md
```diff
@@ -0,0 +1,18 @@
+---
+"@fluidframework/azure-client": minor
+"@fluidframework/odsp-client": minor
+"@fluidframework/tinylicious-client": minor
+"__section": deprecation
+---
+Service client createContainer/getContainer overloads taking CompatibilityMode are deprecated
+
+The `createContainer` and `getContainer` overloads on `AzureClient`, `OdspClient`, and `TinyliciousClient` (plus `AzureClient.viewContainerVersion`) that accept a [`CompatibilityMode`](https://fluidframework.com/docs/api/fluid-static/compatibilitymode-typealias) (`"1"` / `"2"`) argument are now deprecated.
+Pass a [`MinimumVersionForCollab`](https://fluidframework.com/docs/api/runtime-definitions/minimumversionforcollab-typealias) semver string instead — it specifies the minimum collaborating client version directly.
```
jason-ha: "semver" -> "SemVer"

.changeset/zesty-otters-collab.md
```diff
@@ -3,27 +3,22 @@
...
 ---
-`createTreeContainerRuntimeFactory` no longer accepts `minVersionForCollabOverride`
+createTreeContainerRuntimeFactory accepts a MinimumVersionForCollab semver and deprecates minVersionForCollabOverride
 
-The `minVersionForCollabOverride` property on the `props` argument of `createTreeContainerRuntimeFactory` has been removed.
-
-Pass a [`MinimumVersionForCollab`](https://fluidframework.com/docs/api/runtime-definitions/minimumversionforcollab-typealias) semver string (for example `"2.10.0"`) directly via the `compatibilityMode` property instead. `compatibilityMode` now accepts either a `MinimumVersionForCollab` semver string or the existing (deprecated) `CompatibilityMode` values `"1"` / `"2"`.
-
-Before:
+The `compatibilityMode` property on `createTreeContainerRuntimeFactory` now accepts a [`MinimumVersionForCollab`](https://fluidframework.com/docs/api/runtime-definitions/minimumversionforcollab-typealias) semver string in addition to the legacy [`CompatibilityMode`](https://fluidframework.com/docs/api/fluid-static/compatibilitymode-typealias) values `"1"` and `"2"`.
```
jason-ha: semver -> SemVer

packages/service-clients/azure-client/src/AzureClient.ts
```diff
@@ -172,16 +202,46 @@ export class AzureClient {
...
+	}>;
+	/**
+	 * Accesses the existing container given its unique ID in the Azure Fluid Relay.
+	 * @typeparam TContainerSchema - Used to infer the the type of 'initialObjects' in the returned container.
+	 * (normally not explicitly specified.)
+	 * @param id - Unique ID of the container in Azure Fluid Relay.
+	 * @param containerSchema - Container schema used to access data objects in the container.
+	 * @param compatibilityMode - Legacy {@link @fluidframework/fluid-static#CompatibilityMode} value.
 	 * @returns Existing container instance along with associated services.
+	 * @deprecated Pass a `MinimumVersionForCollab` semver string (e.g. `"2.0.0"`) instead. The legacy
```
jason-ha: semver -> SemVer

packages/service-clients/odsp-client/src/odspClient.ts
jason-ha, line 1: 4x semver -> SemVer

packages/service-clients/tinylicious-client/src/TinyliciousClient.ts
jason-ha, line 1: 4x semver -> SemVer

.changeset/zesty-otters-collab.md
```diff
@@ -0,0 +1,24 @@
+---
+"@fluidframework/fluid-static": minor
+"fluid-framework": minor
+"__section": legacy
+---
+createTreeContainerRuntimeFactory accepts a MinimumVersionForCollab semver and deprecates minVersionForCollabOverride
+
+The `compatibilityMode` property on `createTreeContainerRuntimeFactory` now accepts a [`MinimumVersionForCollab`](https://fluidframework.com/docs/api/runtime-definitions/minimumversionforcollab-typealias) semver string in addition to the legacy [`CompatibilityMode`](https://fluidframework.com/docs/api/fluid-static/compatibilitymode-typealias) values `"1"` and `"2"`.
```
jason-ha: This should refer to the breaking change issue - https://github.com/microsoft/FluidFramework/issues/27356
That issue should have all the migration details like the before after below.
You don't need to have the details here. (I recommend not having them here (or in any such deprecation) in case we might want to change the guidance.) It is enough to say there is a deprecation and look to the issue for details.

.changeset/soft-beds-beam.md
```diff
@@ -0,0 +1,18 @@
...
+
```
jason-ha, line 11: Similar to createTreeContainerRuntimeFactory break, there should be a reference to tracking issue.
I think https://github.com/microsoft/FluidFramework/issues/23289 (removal of `CompatiblityMode`) can cover this and document the migrations.

jason-ha (commented):
The build PR, that I thought this one would beat, was merged.
I will post recommended resolution in a moment.

packages/service-clients/end-to-end-tests/azure-client/package.json
jason-ha: The resolve the conflict I think we want:
```json
			"build:test:esm": [
				"^api-extractor:esnext",
				"^build:esnext",
				"build:genver"
			]
```

.changeset/zesty-otters-collab.md
scottn12: Updated in latest

.changeset/soft-beds-beam.md
scottn12: Updated in latest

jason-ha (approved):
I can approve this. However, looking at createTree... and Josh's notes on name, I propose https://github.com/scottn12/FluidFramework/compare/updateDecModelForPolicy...jason-ha:FluidFramework:fluid-static/createTreeContainerRuntimeFactory-cleanup.

scottn12:
> I can approve this. However, looking at createTree... and Josh's notes on name, I propose [scottn12/FluidFramework@updateDecModelForPolicy...jason-ha:FluidFramework:fluid-static/createTreeContainerRuntimeFactory-cleanup](https://github.com/scottn12/FluidFramework/compare/updateDecModelForPolicy...jason-ha:FluidFramework:fluid-static/createTreeContainerRuntimeFactory-cleanup).

@jason-ha, LGTM - applied in latest :)

packages/service-clients/azure-client/README.md
```diff
@@ -129,7 +129,7 @@ const schema = {
 	],
 };
 const azureClient = new AzureClient(props);
-const { container, services } = await azureClient.createContainer(schema, "2" /* compatibilityMode */);
+const { container, services } = await azureClient.createContainer(schema, "2.100.0" /* minVersionForCollab */);
```
jason-ha, line 133: Can happen later as parameter names are not externally meaningful. We want `minVersionForCollaboration`.

jason-ha (approved):

packages/framework/fluid-static/src/treeRootDataObject.ts
```diff
@@ -211,8 +209,45 @@ export function createTreeContainerRuntimeFactory(props: {
...
+}): IRuntimeFactory;
+
+/**
+ * Creates an {@link @fluidframework/aqueduct#IRuntimeFactory} which constructs containers
+ * with an entry point containing single tree-based root data object.
+ *
+ * @remarks
+ * The entry point is opaque to caller.
+ * The root data object's registry and shared objects are configured based on the provided
+ * SharedTree and optionally data store registry.
```
jzaffiro:
```suggestion
 * SharedTree and optional data store registry.
```
nit

jzaffiro (approved):
Looks good pending minor comment!

.changeset/zesty-otters-collab.md
```diff
@@ -0,0 +1,10 @@
+---
+"@fluidframework/fluid-static": minor
+"fluid-framework": minor
```
Josmithr: Nit: I don't believe we export `createTreeContainerRuntimeFactory` via `fluid-framework`, so I think we should probably omit that here.

Josmithr (approved):
API changes look good. I left a comment about one of the changeset files.

My only other feedback is that it seems like we should have more / new test coverage exercising the new degrees of freedom for specifying minVersionForCollab in Container creation - I didn't see any.

Josmithr (approved):
Approving for API and docs

PR #24779, opened 2025-06-06
fix(Offline): Applying stashed ops with empty batches

markfields:
Fixes [AB#39270](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/39270)

## Description

Recent work for Staging Mode broke this.  Fixing it here.

I accidentally got rid of the empty grouped batch object in the PSM placeholder. So when we roundtripped though stashed local state, there was no way to recognize the empty batch (it actually would crash trying to JSON.parse undefined)

There's also another small fix in PendingStateManager.  Since ID Allocation ops aren't resubmitted, we don't care to track their batch IDs.  This means we shouldn't include batchId when replaying them (they're going to get dropped anyway) - without this fix we were triggering the empty batch logic (since they're not actually submitted on replay) when we shouldn't be resubmitting an empty batch.

PR #27213, opened 2026-04-30
feat(container-runtime): add groupedOpCount to grouped batch metadata

anthony-murphy:
## Description

Adds a `groupedOpCount` field to the metadata of grouped batch envelope messages, exposing the number of inner ops in the batch on the wire. This lets wire-level consumers (Word telemetry in particular) record batch sizes without having to parse the grouped batch contents.

Fixes [AB#71749](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/71749).

The field is set on:

- The single envelope produced by grouping a batch (`OpGroupingManager.groupBatch`).
- The empty-grouped-batch placeholder used when a resubmitted batch becomes empty (`OpGroupingManager.createEmptyGroupedBatch`, value `0`).
- The final chunk of a chunked grouped batch (`OpSplitter.splitSingletonBatchMessage`), so the count survives chunking. Intermediate chunks remain unchanged, mirroring how the existing `batch` flag is preserved only on the final chunk.

Compression preserves message metadata as-is, so compressed (non-chunked) grouped batches carry the field through unchanged with no additional code.

`IBatchMetadata` is internal to `container-runtime` and not in any API report, so this is not a public API change.

## Reviewer Guidance

- The wire-format addition is a single optional field on the outer envelope's metadata. It is only read by external telemetry consumers — the runtime itself does not act on it, so behavior is unchanged.
- The `OpGroupingManager` unit tests assert the new field on both the grouped envelope and the empty-grouped-batch placeholder. `opSplitter.spec.ts` has a new `Propagates groupedOpCount onto the last chunk only` test asserting intermediate chunks carry no metadata, the last chunk's envelope carries `{ batch, groupedOpCount }`, and reassembly via `processChunk` restores the field on the reconstructed message.


anthony-murphy:

- 🔄 **Mechanical Review**
  - ✅ mechanical (gpt-5.5)
  - ✅ context (gpt-5.5)
  - ✅ context (claude-opus-4.7)
  - ✅ extract (claude-opus-4.7)
  - ✅ propose (gpt-5.5)
  - ✅ propose (claude-opus-4.7)
  - ✅ actual (claude-opus-4.7)

anthony-murphy:
## Deep Review

*Reviewed commit `c8be39d` on 2026-04-30.*

Small, additive wire-metadata change (a single optional `groupedOpCount` field on `IBatchMetadata`) populated at three sites: `groupBatch`, `createEmptyGroupedBatch`, and the last chunk in `splitSingletonBatchMessage`. No correctness issues found, and the design matches what an independent blind proposal would have produced. The one gating concern is missing test coverage for the new chunked-grouped-batch propagation logic in `opSplitter.ts` — which you've already offered to add in the PR description.

### Findings

- **Chunked grouped-batch metadata propagation is untested** — packages/runtime/container-runtime/src/opLifecycle/opSplitter.ts:172-185 — (inline thread)

### Path to Ready

To move from 6/10 → 7–8 (ALMOST READY):
- [ ] Add the splitter-path unit test you already offered — assertions on the last chunk carrying `groupedOpCount`, intermediate chunks not carrying it, and post-reassembly preservation.

Once the Tier 2 item lands, the next pass would surface these polish items toward 9–10 (READY):
- [ ] Resolve the `typeof rawGroupedOpCount === "number"` guard in `opSplitter.ts:178-180` — either justify it (e.g. call-site type is `Record<string, unknown>`) or inline `firstMessage.metadata?.groupedOpCount` to match how `metadata?.batch` is read two lines later.
- [ ] Extend the JSDoc on `IBatchMetadata.groupedOpCount` (`metadata.ts:43-47`) to enumerate all three observable cases (absent on length-1 fast path / `0` on empty-resubmit placeholder / `N` on grouped envelope) so external telemetry consumers don't need to reverse-engineer the contract.
- [ ] Consider an end-to-end `groupBatch → compress` test asserting `groupedOpCount` survives compression (would also retroactively cover `batchId`/`batch`).
- [ ] Consider a one-line comment in `splitSingletonBatchMessage` locking in the deliberate "last chunk only" propagation choice (mirroring the `batch` flag) vs per-chunk.
- [ ] Tag `markfields` (gatekeeper for `IBatchMetadata` / `opGroupingManager`) and `andre4i` (original author of `splitSingletonBatchMessage` and the "preserve batch metadata on the last chunk" pattern this PR extends) for review.

### Questions for Author

1. Will you add the chunking-path test you already offered? That resolves the only gating item.
2. Was the `typeof rawGroupedOpCount === "number"` guard added to defend against a call site where `metadata` is typed as `Record<string, unknown>`, or is it defensive on a structurally-typed `IBatchMetadata | undefined`? The answer determines whether to inline the read or keep the guard.
3. Should the JSDoc on `IBatchMetadata.groupedOpCount` enumerate the three observable cases (absent / `0` / `N`)?

### Context for Reviewers

- **Compression and chunk-metadata invariants verified.** `OpCompressor` preserves the `metadata` reference verbatim (`opCompressor.ts:56`), so `groupedOpCount` survives compression as the PR description claims. The pre-existing `splitSingletonBatchMessage` pattern of preserving select metadata only on the last chunk (currently `batch`) is mirrored here intentionally. Setting `groupedOpCount: 0` on the empty-batch placeholder is consistent with the empty `contents` array in `emptyGroupedBatch`. `asBatchMetadata` in `metadata.ts` has been a non-validating cast since #21767, so no current code path rejects unknown metadata keys — additive wire-schema change is safe.
- **Area ownership.** `IBatchMetadata` / `opGroupingManager.ts` is markfields' gatekept area (related PRs #21738, #21767, #21854, #24287, #24364, #24779). markfields also established the no-runtime-typeguard convention for `IBatchMetadata` callers in #21767 (*"didn't think the runtime checks were that interesting … let's not waste time with the runtime checks"*) — relevant to the `typeof` guard question above. `opSplitter.ts` was designed by andre4i (#13329, #14150, #14279, #18171); #13329 created `splitSingletonBatchMessage` and the "preserve batch metadata on the last chunk" pattern this PR extends.
- **Reviewer-guidance checklist (passes).** (a) Wire-schema additivity — purely additive optional field, no runtime reader, so the saturation concern from #13329 doesn't apply. (b) `metadata` vs `localOpMetadata` — #21854 established that only fields needing to round-trip on the wire belong on `metadata`; `groupedOpCount` is for wire telemetry, so `metadata` is correct. (c) Grouping-disabled asserts — #21854 added `assert(this.groupedBatchingEnabled)` in placeholder/grouping paths; this PR preserves them (`opGroupingManager.ts:91-94` and line 126).

packages/runtime/container-runtime/src/opLifecycle/opSplitter.ts
```diff
@@ -172,11 +172,16 @@ export class OpSplitter {
...
+		// groupedOpCount is surfaced here so wire-level telemetry can read it once
+		// per chunked grouped batch (mirroring where the batch flag is preserved).
+		const rawGroupedOpCount = firstMessage.metadata?.groupedOpCount;
+		const groupedOpCount =
+			typeof rawGroupedOpCount === "number" ? rawGroupedOpCount : undefined;
 		const lastChunk = chunkToBatchMessage(
 			chunks[chunks.length - 1],
 			batch.referenceSequenceNumber,
-			{ batch: firstMessage.metadata?.batch },
+			{ batch: firstMessage.metadata?.batch, groupedOpCount },
```
anthony-murphy: Deep Review: The new `groupedOpCount` propagation onto the last chunk has no test coverage. `OpGroupingManager.spec.ts` was updated for the grouping side, but `opSplitter.spec.ts` (lines 333–399) only validates chunk counts and content reconstruction — it never asserts metadata propagation. The PR description acknowledges this gap (*"No new chunking test was added — happy to add one if you'd like coverage for the splitter path."*).

This matters because the splitter is exactly where wire/chunk schema bugs have historically appeared (PRs #13329, #14150, #14279, #18171). Since `groupedOpCount` has no runtime reader, a future refactor of `chunkToBatchMessage` (whose signature this PR widens) could silently drop the field with no test or production signal — the canonical failure mode for telemetry-only fields.

**Suggested test** in `opSplitter.spec.ts` (or `remoteMessageProcessor.spec.ts`): split a grouped-batch envelope whose first message has `metadata: { groupedOpCount: N }` and assert:
- (a) intermediate chunks do not carry `metadata.groupedOpCount`;
- (b) the returned last chunk has `metadata.groupedOpCount === N` and the `batch` flag preserved;
- (c) reassembly yields `originalMetadata.groupedOpCount === N`.

This is the only gating item — happy to take you up on the offer.

anthony-murphy: Added in a27c08c — new `Propagates groupedOpCount onto the last chunk only` test in `opSplitter.spec.ts` covers all three assertions (intermediate chunks empty, last chunk carries `{ batch, groupedOpCount }`, reassembly preserves the field via `originalMetadata`).

anthony-murphy:
## Deep Review

*Reviewed commit `a27c08c` on 2026-04-30.*

The gating Tier 2 item from the prior pass — missing chunked-grouped-batch metadata propagation test — was addressed in `a27c08c` with the new `Propagates groupedOpCount onto the last chunk only` test in `opSplitter.spec.ts`, covering intermediate-chunk absence, last-chunk presence of `{ batch, groupedOpCount }`, and reassembly via `processChunk`. The polish concern about the `typeof rawGroupedOpCount === "number"` guard is also resolved — `opSplitter.ts` now reads `firstMessage.metadata?.groupedOpCount` directly. Review converges on no remaining correctness, API, or coverage gaps; only a stale line in the PR description remains.

### Findings

- **PR description's "Reviewer Guidance" claims no chunking test was added, but one is in the diff** — The Reviewer Guidance bullet says *"No new chunking test was added — happy to add one if you'd like coverage for the splitter path,"* but `opSplitter.spec.ts` now contains `Propagates groupedOpCount onto the last chunk only`. Worth a quick edit so future reviewers don't request coverage that already exists (see Suggested description below).

### Path to Ready

To move from 8/10 → 9–10 (READY):
- [ ] Update the PR description's Reviewer Guidance line to point to the new chunking test instead of saying none was added (one-line edit).
- [ ] Get sign-off from `markfields` (recent owner of `IBatchMetadata` and `opGroupingManager.ts`) on the deliberate wire-metadata-vs-`localOpMetadata` placement and the `groupedOpCount: 0` value on the empty-batch placeholder.
- [ ] Get sign-off from `andre4i` (owner of `opSplitter.ts` and the last-chunk-only convention this PR extends) on the splitter change.
- [ ] Take this PR out of draft once reviewers have signed off.

### Context for Reviewers

- **Suggested reviewer set.** `markfields` (most recent modifier of `metadata.ts` and `opGroupingManager.ts`) and `andre4i` (original author of `opSplitter.ts` and the "preserve only `batch` on the last chunk" pattern) as required; `dannimad`, `kian-thompson`, `vladsud` optional.
- **`metadata` vs `localOpMetadata` is deliberate here.** A prior convention (#21854) preferred `localOpMetadata` for fields that don't need round-tripping. This PR puts `groupedOpCount` on wire `metadata` on purpose — the whole point is wire-level telemetry consumption (Word) without parsing batch contents. The JSDoc and PR description make the wire-level intent explicit; `emptyBatch` remains in `localOpMetadata` as before.
- **Empty-batch placeholder is a historically fragile area.** The placeholder has been reworked across #21854, #24287, and #24779 (the last fixed a regression where it was accidentally dropped and replay would crash on `JSON.parse(undefined)`). This PR adds `groupedOpCount: 0` on its wire metadata; review confirmed receive-side ignores it and `ungroupOp` propagates inner `subMessage.metadata` (not envelope metadata), so PSM/stashed-state replay should be inert. Worth a confirming look from someone with #24779 context.
- **Splitter pattern is preserved.** #13329 established that the last chunk's outer envelope carries the minimum needed (`batch`) while the full original metadata rides inside `originalMetadata`. This PR extends the outer-envelope set by exactly one optional field with a comment tying the choice to the `batch` precedent; `processChunk` already restores full original metadata at reassembly, so no splitter/processChunk signature change was required.
- **Optional-field design fits prior direction.** #21767 deliberately removed runtime `isBatchMetadata` checks because all `IBatchMetadata` props are optional and callers handle `undefined`. Adding another optional field is consistent; the JSDoc enumerates absent / `0` / `N` semantics.
- **Inner-metadata assertion preserved.** `assert(Object.keys(rest).length === 0, /* cannot group ops with metadata */)` in `opGroupingManager.ts` is unchanged, so this PR adds one explicit outer-envelope telemetry field rather than reopening arbitrary metadata passthrough — consistent with the original concern in #14512 about grouping hiding service-inspectable metadata.
- **Design space was independently reproduced.** A blind design pass independently arrived at the same shape (extend `IBatchMetadata` with an optional inner-op count, set in the grouping manager, forwarded by the splitter). Three minor divergences (field name, per-chunk vs last-chunk-only propagation, tagging the singleton early-return path with `1`) were each examined and the PR's choices were judged at-least-as-good: the name `groupedOpCount` disambiguates ops vs bytes; last-chunk-only mirrors the existing `batch` flag; omitting `1` on the singleton early-return preserves absent/0/N tri-state semantics.

### Suggested description

Replace the second bullet under **Reviewer Guidance** (currently *"No new chunking test was added — happy to add one if you'd like coverage for the splitter path."*) with:

> - The `OpGroupingManager` unit tests assert the new field on both the grouped envelope and the empty-grouped-batch placeholder. `opSplitter.spec.ts` has a new `Propagates groupedOpCount onto the last chunk only` test asserting intermediate chunks carry no metadata, the last chunk's envelope carries `{ batch, groupedOpCount }`, and reassembly via `processChunk` restores the field on the reconstructed message.

- `c8be39d` 2026-04-30 · **6/10** — chunked-grouped-batch metadata propagation untested; polish notes on a `typeof` guard and JSDoc enumeration

</details>

packages/runtime/container-runtime/src/metadata.ts
```diff
@@ -40,6 +40,16 @@ export interface IBatchMetadata {
...
 	 */
 	batchId?: BatchId;
+	/**
+	 * Set on the envelope of a grouped batch op to the number of inner ops it contains.
+	 * Exposed on the wire so consumers can record batch sizes in telemetry without parsing the grouped batch contents.
+	 *
+	 * Observable values:
+	 * - Absent: this is not a grouped batch envelope (e.g. a singleton batch that bypassed grouping).
+	 * - `0`: empty-grouped-batch placeholder produced when a resubmitted batch becomes empty.
+	 * - `N` (N \> 0): grouped batch with N inner ops. For a chunked grouped batch this appears only on the last chunk's envelope (intermediate chunks carry no metadata).
```
anthony-murphy, line 50: Deep Review: This adds an optional `groupedOpCount` field to `IBatchMetadata`, which is consumed by `PendingStateManager` / stashed local state alongside its wire role (the empty-grouped-batch placeholder participates in stashed-state roundtrips per #24779). PR #21767 (markfields, *Offline: Add batchId to batch metadata on resubmit*) established the precedent of accompanying additive `IBatchMetadata` fields with an explicit stashed-state back-compat note. The current changeset and PR description say nothing about old runtimes loading a new pending-state stash containing `groupedOpCount`, or new runtimes loading an old stash without it.

**Suggested wording** (changeset `code/.changeset/grouped-batch-op-count.md` or PR description):

> The new field is optional and additive; old runtimes loading a new pending-state stash containing `groupedOpCount` will safely ignore it, and new runtimes loading an old stash without `groupedOpCount` are unaffected (the field is consumed only off-runtime by external telemetry).

Worth a sign-off from `@markfields` given his ownership of #21767 / #24287 / #24779. The empty-grouped-batch placeholder (where `groupedOpCount: 0` is now stamped on `placeholderMessage.metadata`) has been the locus of stashed-state regressions before — adding any metadata key to that surface deserves an explicit back-compat note even when the change is purely additive.

packages/runtime/container-runtime/src/opLifecycle/opSplitter.ts
```diff
@@ -171,12 +171,18 @@ export class OpSplitter {
...
 
-		// The last chunk will be part of the new batch and needs to
-		// preserve the batch metadata of the original batch
+		// The last chunk will be part of the new batch and needs to preserve the
+		// batch metadata of the original batch. groupedOpCount is surfaced here
+		// (and only here, not on intermediate chunks) so wire-level telemetry can
+		// read it once per chunked grouped batch — mirroring how the batch flag
+		// is preserved only on the last chunk.
 		const lastChunk = chunkToBatchMessage(
 			chunks[chunks.length - 1],
```
anthony-murphy, line 184: Deep Review: This places `groupedOpCount` only on the last chunk's envelope, and the new `Propagates groupedOpCount onto the last chunk only` test in `opSplitter.spec.ts` actively asserts intermediate chunks have `metadata === undefined`, locking in the convention. This mirrors the existing last-chunk-only `batch` flag (PR #13329, andre4i) — defensible and forward-compatible since stamping additional chunks later is a purely additive change.

For context: two independent blind design passes both recommended stamping the count on **every** chunk so a wire-level observer reading a single chunk in isolation can recover the count without correlating across the chunk sequence. That's a low-confidence concern (the Word telemetry consumer in AB#71749 likely observes reassembled envelopes), but the tradeoff isn't called out anywhere in the PR.

**Suggested:**
1. Confirm with the AB#71749 consumer that reassembled-envelope visibility is sufficient (no per-chunk observation point).
2. Optionally extend the comment near this block noting the rule could be relaxed additively if a future consumer needs per-chunk metadata.
3. Get a quick sign-off from `@andre4i` since this extends his last-chunk-only convention from #13329.

Keeping the last-chunk-only design is fine — just want the choice surfaced and acknowledged rather than implicit.

anthony-murphy: the problem with stamping every message is it leads to double counting of mesages by an observer. chuck message doesn't really contain ops, the contain parts of a payload, which will eventually become ops. only when the last message is processed do all the ops content get reconstructed, so that's why we only stamp the last message

anthony-murphy: Deep Review: Thanks — the double-counting rationale (chunks carry payload fragments, not ops; only the last chunk yields reconstructed ops) is the right argument and matches the last-chunk-only convention from PR #13329 for the `batch` flag. Consider folding that one-liner into the comment block at `opSplitter.ts:174-178` so the rationale lives next to the code; the existing comment already says intermediate chunks would let observers double-count, but the *why* (chunks aren't ops) is the part worth surfacing. Treating this thread as resolved on the design question — pending optional `@andre4i` sign-off on extending his convention.

packages/runtime/container-runtime/src/metadata.ts
anthony-murphy: Went stronger than the suggested back-compat note in `ceefa6ac70`. The JSDoc on `groupedOpCount` now states the actual invariant: the field is always recomputed at outbound time from `batch.messages.length` (or `0` for the empty placeholder), never propagated from stashed pending state. On resubmit, ops re-enter `virtualizeBatch` → `groupBatch`, so a resubmitted batch with squashed/dropped/added ops produces a fresh count matching the actual outbound size. There's no stash→wire round-trip to back-compat in the first place.

packages/runtime/container-runtime/src/opLifecycle/opSplitter.ts
anthony-murphy: Promoted the rationale from this thread into the code comment in `ceefa6ac70` (`opSplitter.ts:174-179`): intermediate chunks don't carry ops — they carry payload parts that only become ops once the last chunk is processed and the payload is reassembled, so stamping every chunk would let an observer double-count. Keeps last-chunk-only as the documented design.

packages/runtime/container-runtime/src/metadata.ts
```diff
@@ -40,6 +40,20 @@ export interface IBatchMetadata {
...
```
anthony-murphy, line 50: Deep Review: `groupedOpCount` is the first deliberately advisory / non-service-validated member of `IBatchMetadata`. Pre-PR, the only fields here were `batch` and `batchId` — both runtime/service-meaningful — and the convention established in PR #14512 (vladsud) was that envelope metadata must be inspectable by the service. The corollary is still in place at `opGroupingManager.ts:146-154`: `assert(Object.keys(rest).length === 0, 0x5dd /* cannot group ops with metadata */)` after stripping `batch`/`batchId`. The new field's own JSDoc states *"the runtime does not validate that an inbound value matches the batch's actual inner op count. It is consumed exclusively by off-runtime telemetry."*

The practical risk is narrow — the inner-op assert still protects against `groupedOpCount` accidentally entering inner ops, and the count is recomputed on every outbound flush — but the precedent of an unvalidated, telemetry-only optional on the outer envelope is new and isn't flagged anywhere in source. Future contributors looking at `IBatchMetadata` won't know that adding another advisory field needs explicit area-owner sign-off.

**Suggested:** add a one-line comment next to `groupedOpCount` explicitly noting that it is the first advisory / non-service-validated member of `IBatchMetadata` and a deliberate departure from the #14512-era "service-inspectable only" rule, so the next contributor doesn't generalise it. Acceptable substitute: a short paragraph in the PR description acknowledging the precedent plus an explicit ack from `@vladsud` (and/or `@markfields`) on the wire-format addition.

This is the only remaining gating item; happy to take a brief PR-description note + area-owner ack as the resolution if you'd rather not add the in-source comment.

packages/runtime/container-runtime/src/metadata.ts
```diff
@@ -40,6 +40,20 @@ export interface IBatchMetadata {
...
+	 *
+	 * The field is intentionally advisory-only: the runtime does not validate that an inbound value matches the batch's actual inner op count. It is consumed exclusively by off-runtime telemetry.
```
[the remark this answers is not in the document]
anthony-murphy: Good catch — fixed in `6e2d057e42`. Updated the "Absent" bullet to call out that absence is ambiguous during rollout (also covers grouped envelopes from older runtimes that predate the field), and noted that telemetry consumers should fall back to parsing envelope contents if a precise count is required while the rollout is in flight.

packages/runtime/container-runtime/src/opLifecycle/opSplitter.ts
```diff
@@ -171,12 +171,19 @@ export class OpSplitter {
...
-		// The last chunk will be part of the new batch and needs to
-		// preserve the batch metadata of the original batch
+		// The last chunk will be part of the new batch and needs to preserve the
+		// batch metadata of the original batch. groupedOpCount is surfaced here
+		// (and only here, not on intermediate chunks) because intermediate chunks
+		// don't carry ops — they carry parts of a payload that only become ops
+		// once the last chunk is processed and the payload is reassembled.
+		// Stamping every chunk would let an observer double-count messages.
 		const lastChunk = chunkToBatchMessage(
 			chunks[chunks.length - 1],
```
anthony-murphy, line 184: Deep Review: The final-chunk envelope now selectively forwards `groupedOpCount` alongside `batch`, but still drops `batchId` — and this PR is the moment that selective-forwarding choice becomes load-bearing.

Before this PR the splitter forwarded only `{ batch: firstMessage.metadata?.batch }`, so the missing `batchId` was arguably a pre-existing gap. With this change the envelope is now `{ batch, groupedOpCount }`, while `OpGroupingManager.groupBatch` upstream stamps the full `{ batchId: groupedBatchId, groupedOpCount: batch.messages.length }`. PR #21767 introduced `batchId` as *"the key piece of information needed to detect/prevent duplication of changes due to a forked container,"* and PR #21854's review thread explicitly required `resubmittingBatchId` to be threaded through the empty-grouped-batch metadata for the same dedup purpose.

Runtime correctness is preserved — `splitOp` populates `chunk.originalMetadata` from the source message's full `op.metadata` on the final chunk (`opSplitter.ts:306-311`), and `processChunk` restores `completeMessage.metadata = chunkedContent.originalMetadata` (`opSplitter.ts:228-239`) before `remoteMessageProcessor.ts:161-175` reads `asBatchMetadata(message.metadata)?.batchId`. So the gap is wire-visible only: any service-/observer-side consumer reading the chunked envelope directly sees `groupedOpCount` but not `batchId`, while the equivalent non-chunked grouped envelope carries both.

**Suggested:** either also forward `batchId: firstMessage.metadata?.batchId` onto the final-chunk envelope and extend the `Propagates groupedOpCount onto the last chunk only` test to assert it round-trips, **or** add an inline comment + negative test pinning the omission as intentional and documenting that no wire-level dedup consumer needs `batchId` on chunked envelopes. Worth a quick sign-off from `@markfields` (owner of the `batchId` design from #21767/#21854) either way.

packages/runtime/container-runtime/src/metadata.ts
anthony-murphy: Not taking either option. The existing JSDoc on `groupedOpCount` already says "intentionally advisory-only" and "consumed exclusively by off-runtime telemetry" — that's the precedent signal a future contributor needs at the call site. A meta-note about this being "the first such field" is documentation about the documentation rather than about the API, so it would add noise without adding information. Likewise not opening a separate area-owner ack process for what is a runtime-internal, non-public field with no service-side consumer. Treating this as resolved.

anthony-murphy:

- ✅ Mechanical Review
  - ✅ context (claude-opus-4.7)
  - ✅ context (gpt-5.5)
  - ✅ extract (claude-opus-4.7)
  - ✅ actual (claude-opus-4.7)
  - ✅ propose (claude-opus-4.7)
  - ✅ propose (gpt-5.5)

anthony-murphy:
## Deep Review

*Reviewed commit `8573fb6` on 2026-05-05.*

Small, well-scoped additive change exposing `groupedOpCount?: number` on the runtime-internal `IBatchMetadata`, stamped at three production sites (grouped envelope, empty-grouped-batch placeholder, last chunk of a chunked grouped batch) and preserved through compression and chunking. No correctness, API, security, concurrency, or performance concerns. The advisory-only-precedent thread on `metadata.ts:50` is closed by author decision; one Tier 3 polish remains — the chunked last-chunk envelope now selectively forwards `groupedOpCount` but still drops `batchId`, a wire-only consistency call worth a `@markfields` sign-off (open thread on `opSplitter.ts:181`).

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- Envelope-level metadata exposure on grouped messages was contested when grouped batching landed (#14512, vladsud). Adding `groupedOpCount` on the outer envelope deliberately re-uses the inspectability surface grouping originally collapsed; the JSDoc frames it as advisory-only and re-stamped at outbound time, with no runtime trust of inbound values.
- Last-chunk-only metadata propagation extends andre4i's #13329 convention. Chunks carry payload fragments, not ops, so stamping every chunk would let a wire observer double-count grouped operations. The new `Propagates groupedOpCount onto the last chunk only` test and the comment block at `opSplitter.ts:174-179` lock the rule in.
- `createEmptyGroupedBatch` is a historically high-friction surface (#21854, #24287, #24364, #24779). The PR's "always recompute on outbound" invariant holds because resubmit re-runs `createEmptyGroupedBatch`, but the regression history warrants a second eye.
- `IBatchMetadata` is internal to `container-runtime` and not in any API report — no public API surface touched.
- Suggested reviewers: **markfields** (primary — `IBatchMetadata` additive pattern, empty-grouped-batch placeholder, `batchId` wire-format design from #21767/#21854); **dannimad** (`createEmptyGroupedBatch` author, #21854); **vladsud** (envelope-metadata policy / advisory-field precedent); **andre4i** (last-chunk-only convention extension).

- **`@markfields` sign-off** on (a) the empty-grouped-batch placeholder change at `opGroupingManager.ts:103` (he authored #24779 fixing the same path's stashed-state regression) and (b) the chunked-path `batchId` decision in the open thread on `opSplitter.ts:181` (he owns the #21767/#21854 `batchId` design).
- **`@vladsud` sign-off** on adding `groupedOpCount` to envelope metadata — first deliberately advisory, non-service-validated field on `IBatchMetadata`, departing from the #14512-era "service-inspectable only" rule. Author has declined an in-source "first advisory field" comment (closed thread on `metadata.ts:50`), so this judgment lives entirely on the human reviewer.
- **`@andre4i` optional ack** on extending the last-chunk-only metadata convention from #13329 to a third metadata field; rationale (chunks aren't ops; double-counting hazard) is now in source per the resolved thread on `opSplitter.ts:181` (comment 3174877769).
- **Confirm with the AB#71749 / Word telemetry consumer** that observing `groupedOpCount` only on the **last** chunk envelope (not every chunk) is acceptable for their ingestion point.
- **Cannot be assessed by the pipeline** — whether host-side telemetry consumers read from the wire envelope vs. the reassembled message at a point where last-chunk-only placement is observable; whether the structural "always recompute on outbound" invariant continues to hold under future Outbox / staging-mode refactors.

</details>

- `81b6231` 2026-05-05 · **8/10** — advisory-only-field gate plus new chunked-envelope `batchId`-consistency thread
- `07f401b` 2026-05-05 · **8/10** — advisory-only-field precedent remained as sole gate
- `6e2d057` 2026-05-04 · **8/10** — JSDoc "Absent" wording tightened; advisory-only-field polish remained as sole gate
- `17edd87` 2026-05-04 · **8/10** — resubmit/recompute concern resolved via JSDoc invariant; advisory-only-field polish remained
- `ceefa6a` 2026-05-04 · **8/10** — splitter-test gap resolved; two missing-test polish items remained, both mitigated by structural arguments
- `371a96c` 2026-05-04 · **8/10** — initial review; advisory-only field aligns with existing precedents, only low-confidence polish items remained

</details>

packages/runtime/container-runtime/src/metadata.ts
anthony-murphy: Deep Review: Acknowledged — treating this resolved. The JSDoc on `groupedOpCount` ("intentionally advisory-only", "consumed exclusively by off-runtime telemetry") does carry the precedent signal at the call site, and an in-source meta-note about it being the first such field would be documentation-about-documentation. Dropping the advisory-field precedent from the readiness gate; the area-owner ack on envelope-metadata policy stays in the For-human-reviewer agenda for `@vladsud`/`@markfields` to weigh as part of normal sign-off.

packages/runtime/container-runtime/src/opLifecycle/opSplitter.ts
anthony-murphy: @markfields @dannimad is this missing something? or do we not set batch id for grouped ops?

anthony-murphy: oh. i think i see. the reconstructed batch will have it, but we don't have it on the outer grouped batch ops.

anthony-murphy: Took option 2 in `6577208cea`: added a 3-line comment at `opSplitter.ts:180-182` pinning the omission as intentional — `batchId` is a runtime dedup field consumed only after `processChunk` restores `originalMetadata`, not by wire observers, so the chunked envelope keeping `{ batch, groupedOpCount }` is the deliberate behavior. No negative test added — the existing `Propagates groupedOpCount onto the last chunk only` test already deep-equals the last chunk's metadata to `{ batch: true, groupedOpCount }`, which would catch a future accidental `batchId` addition.

anthony-murphy:
## Deep Review

*Reviewed commit `6577208` on 2026-05-06.*

**Readiness: 10/10 — READY**

Ready for sign-off. Small, well-scoped additive change exposing `groupedOpCount?: number` on the runtime-internal `IBatchMetadata`, stamped at the grouped envelope, the empty-grouped-batch placeholder, and the last chunk of a chunked grouped batch, and preserved through compression. The last gating thread on `opSplitter.ts:184` (chunked envelope forwards `groupedOpCount` but drops `batchId`) closed in `6577208cea` with a 3-line comment at `opSplitter.ts:180-182` pinning the omission as intentional — `batchId` is a runtime dedup field consumed only after `processChunk` restores `originalMetadata`, not by wire observers — and the existing `Propagates groupedOpCount onto the last chunk only` test deep-equals the last chunk's metadata to `{ batch: true, groupedOpCount }`, which would catch a future accidental `batchId` addition. No author actions remain; outstanding items are area-owner sign-offs and one external-consumer confirmation listed below.

### Context for Reviewers

- Envelope-level metadata exposure on grouped messages was contested when grouped batching landed (#14512, vladsud). `groupedOpCount` deliberately re-uses the inspectability surface grouping originally collapsed; the JSDoc frames it as advisory-only and re-stamped at outbound time, with no runtime trust of inbound values.
- Last-chunk-only metadata propagation extends andre4i's #13329 convention. Chunks carry payload fragments, not ops, so stamping every chunk would let a wire observer double-count grouped operations. The comment block at `opSplitter.ts:174-179` plus the new `opSplitter.ts:180-182` block document both the last-chunk-only rule and the deliberate `batchId`-omission asymmetry.
- `createEmptyGroupedBatch` is a historically high-friction surface (#21854, #24287, #24364, #24779). The PR's "always recompute on outbound" invariant holds because resubmit re-runs `createEmptyGroupedBatch` rather than reusing stashed metadata, and the JSDoc on `groupedOpCount` now states this invariant explicitly.
- `IBatchMetadata` is internal to `container-runtime` and not in any API report — no public API surface touched.
- Suggested reviewers: **markfields** (primary — `IBatchMetadata` additive pattern, empty-grouped-batch placeholder, `batchId` design from #21767/#21854); **vladsud** (envelope-metadata policy / advisory-field precedent); **dannimad** (`createEmptyGroupedBatch` author, #21854); **kian-thompson** (grouped-batching rollout posture); **andre4i** (last-chunk-only convention extension).

- **`@markfields` sign-off** on (a) the empty-grouped-batch placeholder change at `opGroupingManager.ts:103` (he authored #24779 fixing the same path's stashed-state regression and #21767/#21854's `batchId` design that the new `opSplitter.ts:180-182` comment defers to) and (b) confirming the chunked-envelope `{ batch, groupedOpCount }`-but-no-`batchId` shape is the right wire contract for service/observer consumers (resolved thread on `opSplitter.ts:184`, comments 3196431025/3196434785/3196435159).
- **`@vladsud` sign-off** on adding `groupedOpCount` to envelope metadata as the first deliberately advisory, non-service-validated field on `IBatchMetadata` — a departure from the #14512-era "service-inspectable only" rule. Author has declined an in-source "first advisory field" comment (closed thread on `metadata.ts:50`, comment 3193249441), so this judgment lives entirely on the human reviewer.
- **`@vladsud` / `@markfields` design call** on whether the singleton grouped fast path should also stamp `groupedOpCount: 1` so that "absent" eventually becomes unambiguous post-rollout. Author's documented position is that the field is scoped to grouped envelopes by name and the JSDoc adequately documents the singleton-bypass case.
- **`@andre4i` optional ack** on extending the last-chunk-only metadata convention from #13329 to a third metadata field (`groupedOpCount`) plus the `batchId`-yes-on-non-chunked / no-on-chunked-envelope asymmetry now documented at `opSplitter.ts:180-182`.
- **Confirm with the AB#71749 / Word telemetry consumer** that observing `groupedOpCount` only on the **last** chunk envelope (not every chunk, and absent on singleton-bypass envelopes) is acceptable for their ingestion point.
- **Cannot be assessed by the pipeline** — whether host-side telemetry consumers read from the wire envelope vs. the reassembled message at a point where last-chunk-only placement is observable; whether the structural "always recompute on outbound" invariant for the empty-grouped-batch placeholder continues to hold under future Outbox / staging-mode refactors.

</details>

- `8573fb6` 2026-05-05 · **9/10** — chunked-envelope `batchId`-omission concern was the sole gate
- `81b6231` 2026-05-05 · **8/10** — advisory-only-field gate plus new chunked-envelope `batchId`-consistency thread
- `07f401b` 2026-05-05 · **8/10** — advisory-only-field precedent remained as sole gate
- `6e2d057` 2026-05-04 · **8/10** — JSDoc "Absent" wording tightened; advisory-only-field polish remained as sole gate
- `17edd87` 2026-05-04 · **8/10** — resubmit/recompute concern resolved via JSDoc invariant; advisory-only-field polish remained
- `ceefa6a` 2026-05-04 · **8/10** — splitter-test gap resolved; two missing-test polish items remained, both mitigated by structural arguments
- `371a96c` 2026-05-04 · **8/10** — initial review; advisory-only field aligns with existing precedents, only low-confidence polish items remained

</details>

kian-thompson (approved):

PR #13569, opened 2023-01-10
Fix for refreshSummaryFromAck not always downloading latest summary as expected

agarwal-navin:
## Bug
When refereshSummaryFromAck is called, if the ack is newer that the currently tracked latest summary, it attempts to download the latest snapshot from the server and update state from it. However, its possible that the snapshot that is downloaded is older than the one corresponding to the received ack because the its downloaded from the cache. The latest summary state will be incorrectly updated from an older summary leading to inconsistent state.

## Fix
When fetching the latest version via getVersions API, pass the `FetchSource.NoCache` parameter which will ensure that the snapshot is always downloaded via a network call and never from cache. If the downloaded snapshot is still older that the received ack, throw an error because this will lead to inconsistent state.

[AB#3139](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/3139)

PR #12419, opened 2022-10-12
Add TestOverride.SweepTimeout

markfields:
## Background

As-is, the Sweep Timeout will always be at least 6 days due to hardcoded Snapshot Cache limit (5 days) and Buffer (1 day).  This means it's impossible to test the Sweep Timer and the code that runs after something is SweepReady!

There have been multiple attempts on my part to address this:  #11084, #12044, #12331

Much of the difficulty has arisen from the challenge of aligning driver policy with GC config.  I'm finally punting on that and tracking it as a separate item ([AB#2238](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/2238)).

## This change

It's pretty simple:

* Persist SweepTimeoutMs in the summary
* For new containers, that can be overridden via the `"Fluid.GarbageCollection.TestOverride.SweepTimeoutMs"` config setting
* Default behavior for new containers (and for backfilling existing containers with no value persisted) is SweepTimeout + 6 days (same as it has been)

PR #27216, opened 2026-05-01
feat(container-runtime): enable batchId tracking by default with kill-switch and perf telemetry

dannimad:
Enables batchId tracking and the `DuplicateBatchDetector` by default in `FlushMode.TurnBased` so that "parallel fork" duplicate batches (the same local state sequenced twice from two containers rehydrated from the same serialized pending state) are caught for all consumers, not only those who opted into Offline Load.

Changes in `container-runtime`:

- **Default-on in TurnBased.** `batchIdTrackingEnabled` is now `true` whenever the runtime is in `FlushMode.TurnBased` and the kill-switch is not set. In `FlushMode.Immediate` the detector is silently skipped (it has nothing meaningful to do without batches).
- **Kill-switch.** New config flag `Fluid.ContainerRuntime.DisableBatchIdTracking` disables the feature without a code change if a regression is observed in the field.
- **Removed the legacy `Fluid.ContainerRuntime.enableBatchIdTracking` opt-in flag.** It is no longer needed — the feature is unconditionally on. `Fluid.Container.enableOfflineFull` remains, since Offline Load still requires the same `TurnBased`-only constraint and continues to throw `UsageError` if combined with `FlushMode.Immediate` (back-compat preserved).


packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1890,20 +1890,34 @@ export class ContainerRuntime
...
+		const offlineLoadRequested =
+			this.mc.config.getBoolean("Fluid.Container.enableOfflineFull") === true;
+		const disableBatchIdTracking =
+			this.mc.config.getBoolean("Fluid.ContainerRuntime.DisableBatchIdTracking") === true;
+
+		if (offlineLoadRequested && this._flushMode !== FlushMode.TurnBased) {
 			const error = new UsageError("Offline mode is only supported in turn-based mode");
 			this.closeFn(error);
 			throw error;
 		}
```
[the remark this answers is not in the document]
dannimad: offline load flag will also by removed in a follow up

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1890,20 +1890,34 @@ export class ContainerRuntime
...
 
-		// DuplicateBatchDetection is only enabled if Offline Load is enabled
-		// It maintains a cache of all batchIds/sequenceNumbers within the collab window.
-		// Don't waste resources doing so if not needed.
+		this.batchIdTrackingEnabled =
+			!disableBatchIdTracking && this._flushMode === FlushMode.TurnBased;
+
+		this.mc.logger.sendTelemetryEvent({
```
anthony-murphy: there is quite a bit of work being done to reduce logs. i wonder if we really need this? from logs we can know the version of the runtime being used and it seems we could infer this from the version itself. i also think we have generalized logs about config values, so might not even need to know if kill switch is on

packages/test/snapshots/src/validateSnapshots.ts
```diff
@@ -26,6 +27,79 @@ import {
...
+// `\"packageVersion\":\"0.19.5\"`. Mirrors the replacement done by
+// `compareWithReferenceSnapshot` so back-compat snapshots from older runtimes don't fail
+// solely on the embedded version string.
+const packageVersionRegex = /\\"packageversion\\":\\"[^"]+\\"/gi;
+const packageVersionPlaceholder = '\\"packageVersion\\":\\"X\\"';
+
+function normalizePackageVersions(snapshot: IFileSnapshot): IFileSnapshot {
+	return JSON.parse(
+		JSON.stringify(snapshot).replace(packageVersionRegex, packageVersionPlaceholder),
+	) as IFileSnapshot;
```
anthony-murphy: i'd prefer if this only existed in one place. can we move to replay tool use it there, and re-export for use here

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy, line 1908: Deep Review: Retiring `Fluid.ContainerRuntime.enableBatchIdTracking` is fine in principle (it was opt-in, so a host setting it to `false` was equivalent to "not opting in," not a durable off-override) — but the migration path isn't surfaced anywhere a host operator would naturally see it. The diff deletes the `getBoolean("Fluid.ContainerRuntime.enableBatchIdTracking")` read and replaces it with the inverse `DisableBatchIdTracking` key; any host that previously set the legacy flag in either direction now has its value ignored without a deprecation signal in the field.

Suggestions, in increasing order of effort:

1. **Cheapest:** Add one bullet to the PR description / release notes naming the retired key and pointing hosts at `Fluid.ContainerRuntime.DisableBatchIdTracking` — explicitly calling out the **inverted polarity** so a host that previously had `enableBatchIdTracking=false` (intending "off") doesn't carry that intent forward by setting `DisableBatchIdTracking=false` (which means "on").
2. **Optional, data-driven:** Read the legacy key for one release and emit a one-time `DeprecatedBatchIdTrackingFlagObserved` telemetry event when seen, so retirement is data-driven rather than assumption-driven. The new `BatchIdTrackingEnablement` event already provides ongoing rollout observability and would pair naturally with a deprecation counterpart.

If the team already knows no caller sets the legacy key, the bullet alone is enough.

anthony-murphy: we know no caller sets the legacy key.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: Looking at this from the description-vs-code angle: the PR description's Telemetry section explicitly promises *"One-time `BatchIdTrackingEnablement` event on container construction (fields: `enabled`, `disableBatchIdTracking`, `flushMode`) — lets dashboards attribute regressions to enablement state."* A repo-wide grep for `BatchIdTrackingEnablement` returns zero matches, and the changed constructor block here only computes `offlineLoadRequested` / `disableBatchIdTracking`, assigns `this.batchIdTrackingEnabled`, and conditionally constructs `DuplicateBatchDetector` — there is no `this.mc.logger.sendTelemetryEvent(...)` call. The only telemetry the PR ships lives in `DuplicateBatchDetector.getRecentBatchInfoForSummary`, which only fires when a summary is produced — idle/short-lived sessions surface no per-session activation signal during rollout.

Your log-volume concern is reasonable and may well be the right answer here — runtime version plus the generalized config-key telemetry can plausibly substitute for a dedicated event. The remaining issue is just that the description and code currently disagree. Either path resolves it:

1. **Remove the bullet from the PR description** (and accept that runtime version + generalized config-key logs are sufficient for rollout attribution).
2. **Emit the event** with the documented fields right after `this.batchIdTrackingEnabled` is assigned, with focused unit tests for default-enabled / kill-switch / Immediate-mode shapes.

No strong opinion on which — option 1 is cheaper and matches the log-reduction direction you describe.

packages/runtime/container-runtime/src/containerRuntime.ts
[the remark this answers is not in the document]
anthony-murphy: Deep Review: Following up on this thread now that batchId tracking is default-on. The asymmetry the original comment flagged is still live in the current diff: the `offlineLoadRequested` `UsageError` validates only flush mode, not the kill-switch, and the constructor comment immediately above still describes batchId tracking as the prerequisite for Offline Load. Pre-PR, `enableOfflineFull=true` unconditionally enabled tracking; post-PR, setting `enableOfflineFull=true` together with `Fluid.ContainerRuntime.DisableBatchIdTracking=true` in TurnBased silently strips the prerequisite — Offline Load proceeds with fork detection disabled, no error, no telemetry. The PR description says Offline Load is "back-compat preserved" but only the FlushMode portion currently is.

@dannimad you mentioned `enableOfflineFull` will be removed in a follow-up — that resolves this cleanly if it's imminent. A couple of cheap options either way:

- **(a)** Link the follow-up issue here and add a one-line comment in the constructor noting the asymmetry is intentionally transient, **or**
- **(b)** Add the symmetric guard now — extend the existing `UsageError` to fire when `offlineLoadRequested && disableBatchIdTracking`, or emit a `BatchIdTrackingDisabledForOfflineLoad` warning telemetry so the misconfig is at least observable while the follow-up lands.

Either closes the loop on this thread.

anthony-murphy: Deep Review: Thanks — that confirmation (no caller sets the legacy key) is exactly the data point needed to retire it without a deprecation log. The original "option 1: bullet in the description / release notes pointing at the inverted-polarity replacement" still has value for any external host that might pick this up later, but it's no longer load-bearing. Happy to consider this resolved on the deprecation-signal angle.

anthony-murphy:

- 🔄 **Mechanical Review**
  - ✅ mechanical (gpt-5.5)
  - 🔄 context (gpt-5.5)
  - ✅ context (claude-opus-4.7)
  - ✅ extract (claude-opus-4.7)
  - ✅ propose (claude-opus-4.7)
  - 🔄 propose (gpt-5.5)
  - 🔄 actual (claude-opus-4.7)

anthony-murphy:
## Deep Review

*Reviewed commit `90edb00` on 2026-05-05.*

**Readiness: 6/10 — 🔨 MAKING PROGRESS**

Design continues to converge — independent solution-space proposals matched the PR's shape (default-on `DuplicateBatchDetector` in `TurnBased`, dedicated `DisableBatchIdTracking` kill-switch, per-summary peak/processed counters). One Tier 2 issue survives this pass: the PR description still promises a `BatchIdTrackingEnablement` rollout-attribution event that is not emitted in the diff. The Offline-Load + kill-switch asymmetry from the prior review remains tracked on its existing inline thread (`3173973950`, still unresolved). Both are bounded one-touch fixes; both have author engagement. Holding at 6/10 — no Tier 1, one Tier 2, three Tier 3 polish items held until Tier 2 closes.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- **Default-off → default-on reverses a deliberate prior scoping decision.** The pre-PR code carried an explicit author rationale comment (removed by this PR): *"DuplicateBatchDetection is only enabled if Offline Load is enabled … Don't waste resources doing so if not needed."* The new `peakRecentBatchCount` / `processedBatchCount` counters are precisely the instrumentation needed to validate that prior cost concern empirically. Note `processedBatchCount++` is incremented before the duplicate check, so it counts duplicates as "processed" — confirm that matches dashboard intent.
- **Silent fallback in `FlushMode.Immediate` matches `vladsud`'s prior guidance (PR #14239).** This PR adopts that style for the default-on path while the *explicit* `enableOfflineFull` opt-in still throws `UsageError` — preserving the counter-point that explicit opt-in conflicts should remain observable. The asymmetry is intentional.
- **`containerRuntime.ts` is the highest-churn file in this diff** (30 commits / 6mo). Recent related PRs: #12419 (markfields), #13569 (agarwal-navin), #14239 (andre4i — FlushMode/UsageError pattern, with vladsud review). `duplicateBatchDetector.ts` has no related PRs in the index — low historical reviewer base on the file itself.
- **Suggested reviewers:** **vladsud** (telemetry naming + UsageError-vs-silent ergonomics), **andre4i** (FlushMode invariants in ctor, #14239), **anthony-murphy** (offline-load batching test changes, #12498), **markfields/agarwal-navin** (recent `containerRuntime.ts` churn).

- **Needs human judgment:**
  - **`anthony-murphy`'s open log-volume question on the enablement event** ("do we really need this? we have generalized logs about config values") is yours to call. If "no", the Tier 2 collapses to a description amendment.
  - **`vladsud`** — set the silent-fallback-over-throw precedent in #14239; right person to validate the deliberate Offline-Load asymmetry (silent for the new default-on path, loud `UsageError` retained for the legacy explicit `enableOfflineFull` opt-in).
  - **`andre4i` / `markfields`** — confirm no production runtime path (resubmit during reconnect, `applyStashedOp` replay, transient ordering) can produce a `(sequenceNumber, batchId)` shape that trips the duplicate-batch invariant now that detection is default-on. The `fewerBatches.spec.ts` comment ("trip its invariants and short-circuit") indicates the invariant is an assertion, not graceful handling.
- **Cannot be assessed by the pipeline:**
  - Runtime/perf cost of the new `peakRecentBatchCount` / `processedBatchCount` counters and per-summary reset on real workloads.
  - Whether the follow-up that removes `enableOfflineFull` is close enough that the asymmetry can land as-is.
  - Whether any external host outside this repo sets the retired `Fluid.ContainerRuntime.enableBatchIdTracking` key (in-tree confirmed clear on thread `3184554880`).

</details>

- `031056a` 2026-05-05 · **6/10** — enablement event missing; offline-load asymmetry re-pinged
- `b29d0ff` 2026-05-04 · **8/10** — three minor polish items; enablement event referenced as if it existed (later confirmed missing)

</details>

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1890,20 +1890,27 @@ export class ContainerRuntime
...
+		// and is also a prerequisite for the Offline Load feature. It is enabled by default in
+		// TurnBased mode; the kill-switch below allows disabling it without a code change if a
+		// regression is observed. Offline Load still requires TurnBased, so consumers that
+		// explicitly opt into it with FlushMode.Immediate continue to get a UsageError.
+		const offlineLoadRequested =
+			this.mc.config.getBoolean("Fluid.Container.enableOfflineFull") === true;
+		const disableBatchIdTracking =
+			this.mc.config.getBoolean("Fluid.ContainerRuntime.DisableBatchIdTracking") === true;
+
+		if (offlineLoadRequested && this._flushMode !== FlushMode.TurnBased) {
```
markfields, line 1903: I think we should probably set `disableBatchIdTracking = true` if we're not TurnBased.  I don't remember the exact sticking point why we don't think Immediate Mode works, but if that's the case then let's continue to avoid that combo of batchId + Immediate.

Note: In practice this case won't be hit, but it makes the code safer IMO.

markfields: Oh haha that's exactly what you do, I should have kept reading

anthony-murphy (approved):

PR #27217, opened 2026-05-01
feat(tree): add BranchCheckout for viewless, branch-bound checkouts

jzaffiro:
## Description

Adds `BranchCheckout`, a `TreeCheckout` subclass that is permanently bound to the `SharedTreeBranch` it was created over. Companion helpers `forkAsBranchCheckout(parent)` (fork any checkout into a viewless branch checkout) and `getBranchCheckout(branch)` (look up the canonical `BranchCheckout` for a given branch via a module-level `WeakMap`) round out the surface. All three are package-internal (`@internal`).

To make this work cleanly, `TreeCheckout.fork` is refactored into a generic `forkWith<T>(ctor)`, which subclasses (and `forkAsBranchCheckout`) call directly. The `@throwIfBroken` guard now sits on `forkWith` rather than on `fork`, so every fork entry point — including subclass overrides — runs the broken-state check.

`BranchCheckout` overrides `[disposeSymbol]` (not `dispose`) so that both explicit disposal and the merge auto-dispose path (`TreeCheckout.merge` calls `checkout[disposeSymbol]()` directly) clear the registry entry. `switchBranch` is overridden to throw a `UsageError`, with the base parameter preserved to keep the override signature-compatible with `TreeCheckout`.

## Reviewer Guidance

- The new symbols are intentionally `@internal` with the exception of the `getBranch` and `getViewOfBranch` functions to allow public access to the new branch objects. 
- A point worth gut-checking: should the `WeakMap` registry live on the SharedTree instance instead of the module? Module-global is simpler and works because branch identities are unique across trees, but happy to refactor if you prefer per-tree isolation.
- There's a pre-existing TODO at `treeCheckout.ts:322-323` (`createTreeCheckout` always sets `isSharedBranch=true`) that affects how `forkAsBranchCheckout` interacts with views from `independentView` — out of scope here, but worth a follow-up.
- 18 tests in `branchCheckout.spec.ts` cover construction, viewlessness, permanence, edits/merges, dispose semantics (including the merge auto-dispose registry cleanup), broken-state propagation, chained forking, and `disposeForksAfterTransaction` propagation.


packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -502,12 +502,18 @@ export type FormatVersion = number | string | undefined;
 // @alpha
 export function generateSchemaFromSimpleSchema(simple: SimpleTreeSchema): TreeSchema;
 
+// @alpha
+export function getBranch<TSchema extends ImplicitFieldSchema | UnsafeUnknownSchema>(view: TreeViewAlpha<TSchema>): TreeBranchAlpha;
```
CraigMacomber: The generics here are not helpful, see https://github.com/microsoft/FluidFramework/wiki/Coding-Guidelines#-avoid-unnecessary-use-of-generics

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,252 @@
...
+ * Package-internal helper backing the public {@link getBranch} entry point. Not part of the
+ * public alpha surface — consumers should call {@link getBranch} instead.
+ */
+export function forkAsBranchCheckout(parent: TreeCheckout): BranchCheckout {
+	return parent.forkWith(BranchCheckout);
+}
+
+/**
+ * Returns the branch bound to the given view, forking one from the view's checkout
+ * if no live branch is currently registered for it.
```
CraigMacomber: "checkout" is an implementation detail, not a documented user facing concept, so it should not be referenced in user facing documentation like this.

I also don't think we want to say this forks anything for similar reasons: the user of this isn't supposed to know this lazily creates a branch object to work around some of our internal limitations (we should be able to change that without impacting this APi surface)

I think we want something more like:

```suggestion
 * Returns the branch currently bound to the given view.
```

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * @remarks
```
CraigMacomber, line 316: I think most of this section belongs in either privateRemarks or comments in the function.

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,252 @@
...
+ * 3. Otherwise, a fresh `BranchCheckout` is forked via `forkAsBranchCheckout` and recorded
+ * against the view's `mainBranch` for future calls.
+ *
+ * @typeParam TSchema - The schema type of the tree view.
+ * @param view - A {@link TreeViewAlpha} returned by the Fluid Framework. External implementations
+ * are not supported and will cause a `UsageError` to be thrown.
+ *
+ * @alpha
+ */
+export function getBranch<TSchema extends ImplicitFieldSchema | UnsafeUnknownSchema>(
```
CraigMacomber: There is an interesting design question here: when to we dispose of the branch? Having a branch exist has cost (keeps a forest up to date over time.

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,252 @@
...
+export function forkAsBranchCheckout(parent: TreeCheckout): BranchCheckout {
+	return parent.forkWith(BranchCheckout);
+}
+
+/**
+ * Returns the branch bound to the given view, forking one from the view's checkout
+ * if no live branch is currently registered for it.
+ *
+ * @remarks
+ * Idempotent per view: successive calls on the same view return the same {@link TreeBranchAlpha}
```
CraigMacomber: I don't think this is the best way to put this:

This is true if you call getBranch a bunch in a row, but not always, since when a view is in a transaction (or if its moved to another branch some other way in the future) this will not return the same object.

packages/dds/tree/src/shared-tree/branchCheckout.ts
CraigMacomber: I think we should note (likely in privateRemarks) that when we stabilize this API, it should likely be as a property of TreeView rather than a free function, and that this approach is only done for now to keep coupling low as we work to improve it.

packages/dds/tree/src/shared-tree/branchCheckout.ts
jzaffiro: Updated the wording.

packages/dds/tree/src/shared-tree/branchCheckout.ts
jzaffiro: Added a comment noting this open question - how do we decide when views should be disposed currently? Is there an existing time period after which we do not guarantee that views (or any other objects) are reachable?

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,264 @@
...
+ * before this graduates from `@alpha`.
+ *
+ * @alpha
+ */
+export function getBranch(view: TreeBranchAlpha): TreeBranchAlpha {
+	if (!(view instanceof SchematizingSimpleTreeView)) {
+		throw new UsageError(
+			"The `view` argument to `getBranch` must be a view returned by the Fluid Framework — external implementations are not supported.",
+		);
+	}
```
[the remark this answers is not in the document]
jzaffiro: This is intentional to avoid strange generic types.

packages/dds/tree/src/test/shared-tree/branchCheckout.spec.ts
```diff
@@ -0,0 +1,576 @@
...
+			assert.strictEqual(branch.disposed, false);
+			// And a second view sees the prior edits.
+			const v2 = getViewOfBranch(branch, config);
+			assert.strictEqual(v2.root.x, 3);
+		});
+	});
+
+	// Verifies that the WeakMaps in branchCheckout.ts (notably `branchCheckoutMap`, which is
+	// keyed by `SharedTreeBranch`) do not pin disposed branches in memory.
+	// Uses the two-async-major-GC pattern from packages/framework/react/src/test/useObservation.spec.tsx.
```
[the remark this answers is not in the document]
jzaffiro: fixed in previous commit

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -502,12 +502,18 @@ export type FormatVersion = number | string | undefined;
 // @alpha
 export function generateSchemaFromSimpleSchema(simple: SimpleTreeSchema): TreeSchema;
 
+// @alpha
+export function getBranch(view: TreeBranchAlpha): TreeBranchAlpha;
```
Josmithr, line 506: Would it be worth adding a changeset for these?

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,379 @@
...
+const lazyBranchForViewMap = new WeakMap<
+	SharedTreeBranch<SharedTreeEditBuilder, SharedTreeChange>,
+	WeakRef<BranchCheckout>
+>();
+
+/**
+ * Per-{@link BranchCheckout} state passed to the {@link FinalizationRegistry} callback.
+ *
+ * @remarks
+ * **Heap-snapshot tip:** named class so leaks of finalization state show up clearly.
```
Josmithr: Nit
```suggestion
 * **Heap-snapshot tip:** this is a named class so leaks of finalization state show up clearly.
```

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,379 @@
...
+ *
+ * **Critical invariant:** instances must not transitively reference the `BranchCheckout` they were
+ * registered for. If they did, the registry's strong hold on this object would pin the checkout
+ * and the finalizer would never fire — the same trap solved by `SubscriptionsWrapper` in
+ * `packages/framework/react/src/useObservation.ts`.
+ *
+ * The state intentionally holds **no** references to the checkout's resources (transaction, forest,
+ * events, views, revertibles, ...). All those resources are part of a self-cycle with the
+ * `BranchCheckout` itself; once nothing external retains the checkout, the JS GC collects the whole
```
Josmithr: Nit: our guidelines call to avoid abbreviations wherever reasonable.
```suggestion
 * `BranchCheckout` itself; once nothing external retains the checkout, the JavaScript garbage collector collects the whole
```

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,379 @@
...
+	 *
+	 * @remarks
+	 * The callback must not close over the `BranchCheckout` or anything that reaches it — doing so
+	 * would prevent the finalizer from ever firing. Intended for testing and (future) telemetry.
+	 */
+	public onFinalized: (() => void) | undefined;
+}
+
+/**
+ * Side-table mapping each `BranchCheckout` to its finalization state, so external code (tests,
```
Josmithr: Nit: "Side-table"? Not sure what this means in this context.

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,379 @@
...
+ * Unlike {@link TreeCheckout}, a `BranchCheckout` cannot be retargeted to a different branch via `switchBranch` —
+ * calling it always throws a `UsageError`.
+ *
+ * It is "viewless" in the sense that no {@link SchematizingSimpleTreeView} is attached at construction time —
+ * a view can still be materialized on demand via the inherited `viewWith`.
+ *
+ * Lifecycle: a `BranchCheckout` is independent from any other checkout that observes the same data.
+ * Disposing the parent of a {@link forkAsBranchCheckout} does not dispose the child, and vice versa;
+ * merging is explicit via {@link TreeCheckout.merge}.
+ *
```
Josmithr: Nit
```suggestion
```

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,379 @@
...
+ * against the view's `mainBranch` for future calls.
+ *
+ * Open design question: when should the lazily-forked branch be disposed? A live `BranchCheckout`
+ * has ongoing cost (it keeps a forest up to date), so the lifetime semantics need to be revisited
+ * before this graduates from `@alpha`.
+ *
+ * @alpha
+ */
+export function getBranch(view: TreeBranchAlpha): TreeBranchAlpha {
+	if (!(view instanceof SchematizingSimpleTreeView)) {
```
Josmithr, line 351: Nit: it would probably be good to add a `@throws` block to the docs above noting the conditions in which this will error.

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,379 @@
...
+ *
+ * @alpha
+ */
+export function getViewOfBranch<TSchema extends ImplicitFieldSchema>(
+	branch: TreeBranchAlpha,
+	config: TreeViewConfiguration<TSchema>,
+): TreeViewAlpha<TSchema> {
+	if (!(branch instanceof BranchCheckout)) {
+		throw new UsageError(
+			"The `branch` argument to `getViewOfBranch` must be a `TreeBranchAlpha` returned by `getBranch`.",
```
Josmithr, line 390: Nit: it would probably be good to add a `@throws` block to the docs above noting the conditions in which this will error.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1269,6 +1310,23 @@ export class TreeCheckout implements ITreeCheckout {
...
 			"The branch has already been disposed and cannot be disposed again.",
 		);
+		this.disposeOnce();
+	}
+
+	/**
+	 * Silent-idempotent disposal: performs the actual cleanup work without throwing on repeat
+	 * calls. Intended for contexts (such as `FinalizationRegistry` callbacks) where throwing is
+	 * undesirable — currently reserved for that purpose; no in-tree caller invokes it from a
+	 * finalizer today. Subsequent calls are no-ops.
```
Josmithr: Nit: might be better to omit the " no in-tree caller invokes it from a finalizer today" note, since that could become easily out-of-date.

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,379 @@
...
+ * a view can still be materialized on demand via the inherited `viewWith`.
+ *
+ * Lifecycle: a `BranchCheckout` is independent from any other checkout that observes the same data.
+ * Disposing the parent of a {@link forkAsBranchCheckout} does not dispose the child, and vice versa;
+ * merging is explicit via {@link TreeCheckout.merge}.
+ *
+ */
+export class BranchCheckout extends TreeCheckout {
+	public constructor(
+		branch: SharedTreeBranch<SharedTreeEditBuilder, SharedTreeChange>,
```
Josmithr, line 186: Probably not for this PR, since I suspect it would introduce a lot of scope creep. But this has **a lot** of parameters. Probably worth converting this to a property bag so call-sites get nicely labeled arguments.

Josmithr:
Fantastic documentation here. Thank you for being so thorough!

packages/dds/tree/src/shared-tree/branchCheckout.ts
jzaffiro: [AB#74748](https://dev.azure.com/fluidframework/internal/_workitems/edit/74748) for tracking

packages/dds/tree/api-report/tree.alpha.api.md
jzaffiro: done :)

packages/dds/tree/src/shared-tree/branchCheckout.ts
jzaffiro: Updated the wording to be more clear.

yann-achard-MS:
What scenarios/application workflows are these new capabilities meant to support?

packages/dds/tree/src/shared-tree/branchCheckout.ts
```diff
@@ -0,0 +1,386 @@
...
+/**
+ * Returns the branch currently bound to the given view.
+ *
+ * @remarks
+ * Repeated calls with the same view typically return the same {@link TreeBranchAlpha} instance,
+ * but this is not guaranteed: for example, while the view is participating in a transaction
+ * its underlying branch may differ from the one observed outside the transaction, and a future
+ * change that retargets a view to another branch would likewise cause a different instance to be
+ * returned.
+ *
```
yann-achard-MS, line 322: I find this confusing:
- Isn't the view bound to the branch as opposed to the branch to the view?
- "Returns the branch currently bound to the given view" sounds like it will never generate a new `TreeBranchAlpha` instance but path3 does that.
- It seems confusing to have an API that sometimes forks and sometimes doesn't: if I merge new changes in the returned `TreeBranchAlpha`, it's hard to tell whether that will affect the view I passed in.

Some notes about the remarks here:

It seems to me that there are only two cases where a different branch is returned:
- The previously returned one was disposed.
- The view has been switched to a different branch with `switchBranch` (which, if I understand correctly, is the second example you give). However, I don't think we expose this capability to the user, so maybe that should not be mentioned.

I don't think the first example (the view is participating in a transaction) applies because we don't allow forking a view that is in an open transaction so you should get an error as opposed to a different branch. The fact that a branch is involved in the implementation of transactions (the "underlying branch") should not be mentioned here since that's private implementation details.

jzaffiro: - With the current state of the code, I assumed "view bound to branch" == "branch bound to view" - if that's not correct, I can update the doc comment to something along the lines of "Returns the underlying branch of the given view."
- For the 2nd and 3rd bullet points, maybe it would be better then to have a separate function to create a new branchCheckout? My goal here was for this API to return a branch regardless of the situation it's called in. 
I can update the remarks to remove private implementation details.

PR #27218, opened 2026-05-01
Clean up test-version-utils method semantics

Abe27342:
## Description

Followup to #27186 which addresses some minor points of feedback I didn't want to bundle into that (or which were made after it merged):

- Simplify return type of `ensurePackageInstalled` (and rename to `ensureVersionLoaded` to better reflect its purpose)
- Split `resolveVersion` into two functions with more natural signatures: `resolveRangeViaRegistry` and `resolveRangeViaManifest`
- Deduplicate `resolveRangeViaRegistry` in update script by making it reference the one in `versionUtils.ts`
- Minor other bits of feedback on docs clarity

### DIFF
--- a/packages/test/test-version-utils/src/describeWithVersions.ts
+++ b/packages/test/test-version-utils/src/describeWithVersions.ts
@@ -34,7 +34,7 @@ export interface IRequestedFluidVersions {
 	requestAbsoluteVersions?: string[];
 }
 
-const installRequiredVersions = async (config: IRequestedFluidVersions): Promise<void> => {
+const loadVersions = async (config: IRequestedFluidVersions): Promise<void> => {
 	const loadPromises: Promise<void>[] = [];
 	if (config.requestAbsoluteVersions !== undefined) {
 		loadPromises.push(
@@ -77,7 +77,7 @@ function createTestSuiteWithInstalledVersion(
 		before("Create TestObjectProvider", async function () {
 			this.timeout(Math.max(defaultTimeoutMs, timeoutMs));
 
-			await installRequiredVersions(requiredVersions);
+			await loadVersions(requiredVersions);
 			provider = await getVersionedTestObjectProvider(
 				pkgVersion, // baseVersion
 				pkgVersion, // loaderVersion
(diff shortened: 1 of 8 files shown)


packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -3,30 +3,6 @@
  * Licensed under the MIT License.
  */
 
-/**
```
alexvy86, line 6: Double checking if this is an intentional deletion of the comments? I know we have some version of this in several places, so it could get out of sync, but some version of it (or reference to a single source of truth) seems worth keeping.

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -94,70 +63,81 @@ const resolutionCache = new Map<string, string>();
...
+});
+
+/**
+ * Resolves a semver dependency spec to the most recent installed version under compat-workspaces which
+ * satisfies that range.
+ *
+ * Throws if the currently installed compat workspace does not include any version that matches the spec.
+ * @param rangeSpec - A valid (as per [semver](https://www.npmjs.com/package/semver)) range specification
+ */
+function resolveRangeViaManifest(rangeSpec: string): string {
```
alexvy86, line 131: `resolveRangeViaManifest` does not use `cached` like `resolveRangeViaRegistry`, whereas both paths shared a cache before. Since it's only doing local stuff (vs checks on npmjs) and I wouldn't expect this to be on a hot path, I'm not super worried about performance, but confirming the change was intentional/expected?

packages/test/test-version-utils/src/describeWithVersions.ts
```diff
@@ -35,21 +35,21 @@ export interface IRequestedFluidVersions {
 }
 
 const installRequiredVersions = async (config: IRequestedFluidVersions): Promise<void> => {
```
alexvy86: Nit, for extra clarity on what it actually does.
```suggestion
const loadVersions = async (config: IRequestedFluidVersions): Promise<void> => {
```

packages/test/test-version-utils/src/versionUtils.ts
Abe27342: Yes this is intentional deletion and feedback from Scott's review. There are descriptions of this at the top of the update script + in the package README which both are better places for this kind of information.

packages/test/test-version-utils/src/versionUtils.ts
Abe27342: Yeah it seemed unnecessary to put a cache on this layer for the reasons you called out

alexvy86 (approved):

PR #27219, opened 2026-05-01
fix(api-markdown-documenter): Fix some misc. bugs

Josmithr:
- Fixes an issue where a model with no packages (or an empty set of packages after applying configured filtering) would fail to generate a document for the model itself.
- Fixed an issue where incorrect heading IDs would be generated for items configured to generate their own documents.


alexvy86 (approved):
Seems good as far as I can tell :)

alexvy86 (approved):

PR #27220, opened 2026-05-01
feat(container-loader): add captureFullContainerState free function

markfields:
Adds `captureFullContainerState`, a `@legacy @alpha` free function in `container-loader` that
produces an `IPendingContainerState` JSON for an attached document **without instantiating a
Loader, Container, or runtime**. It drives only `IUrlResolver` + `IDocumentServiceFactory`,
fetches the latest snapshot via `getSnapshot`/`getSnapshotTree`, drains ops via
`fetchMessages(seq+1, …)`, and inlines blob contents and loading-group snapshots so the
artifact is fully portable.

This is the missing piece in the frozen-container series:

- #25513 (`asLegacyAlpha`/`ContainerAlpha` surface)
- #25538 (`loadFrozenContainerFromPendingState`)
- #25590 (frozen-container blob reads — explicitly called out the offline-blob gap this PR closes)
- #25653 (`createFrozenDocumentServiceFactory`)
- #25742 (frozen-container hardening)

Together with `loadFrozenContainerFromPendingState`, `captureFullContainerState` enables
fully-offline frozen-container scenarios.

### Design notes

- GC respect is two-layered: snapshot tree walk skips `unreferenced: true` subtrees, and the
  attachment-blob filter parses GC tombstones / deleted-nodes with documented GC-lag
  tolerance (blobs absent from the GC graph are kept).
- Concurrency is bounded (`mapWithConcurrency`: 32 blobs, 4 group snapshots).
- Loading-group snapshots are intentionally NOT proactively captured — `captureFullContainerState`
  throws `UsageError` if any referenced subtree carries a `groupId`. This will be reintroduced when
  there's a known consumer and e2e harness.
- Some snapshot/GC/blob-tree constants are duplicated from `container-runtime`. A
  cross-package contract test fails CI on drift.
- Attachment-blob payloads are encoded via base64, which is different from existing inlined
  blobs since those are text-based payloads.

See also #27100, and [this markdown file](https://github.com/microsoft/FluidFramework/blob/f4621c2c1d61adde3f7d5d9d952f296ac278108d/full-container-state-review-notes.md) used during branch development,

### Follow-ups

See this [list of follow-ups](https://github.com/microsoft/FluidFramework/blob/f4621c2c1d61adde3f7d5d9d952f296ac278108d/full-container-state-review-notes.md#part-4--follow-up-tasks-ready-to-lift-into-ado-items) accrued during development.  These will be moved to ADO tasks once the PR is merged.

Additionally, these from Deep Review on a later iteration:

* https://github.com/microsoft/FluidFramework/pull/27220#discussion_r3204969959
* https://github.com/microsoft/FluidFramework/pull/27220#discussion_r3204970080

And some of my own thoughts on final review pass:
* Do we have roundtrip test coverage of both `loadExistingContainer` and `loadFrozenContainerFromPendingState`?
* Dedupe GC types (and code) or add contract tests to prevent drift


anthony-murphy:

- ✅ Mechanical Review
  - ✅ context (claude-opus-4.7)
  - ✅ context (gpt-5.5)
- 🔄 **Adversarial Filter**
  - ✅ prosecution (claude-opus-4.7)
  - 🔄 defense (gpt-5.5)

packages/loader/container-loader/src/captureReferencedContents.ts
```diff
@@ -0,0 +1,365 @@
...
+	if (tableBlobId !== undefined) {
+		const entries = await readAndParse<[string, string][]>(storage, tableBlobId);
+		for (const [localId, storageId] of entries) {
+			redirectTable.set(localId, storageId);
+		}
+	}
+	for (const [key, storageId] of Object.entries(blobsTree.blobs)) {
+		if (key !== redirectTableBlobName) {
+			// Identity mapping: storage ids referenced directly in handles (legacy).
+			redirectTable.set(storageId, storageId);
```
anthony-murphy, line 286: Deep Review: Identity redirect-table entries appear to bypass GC filtering for attachment blobs.

The filter at lines 224–228 keeps a storage id only when its local id is **not** in `unreferencedLocalIds`:

```ts
if (unreferencedLocalIds?.has(localId) !== true) {
    storageIdsToFetch.add(storageId);
}
```

But `readRedirectTable` (lines 255–258) unconditionally synthesizes an identity entry for every storage id present in `.blobs` that isn't `.redirectTable`:

```ts
redirectTable.set(storageId, storageId);
```

`BlobManager` intentionally **excludes** identity mappings from GC data (see `blobManager.ts` — `if (localId !== storageId) { gcData.gcNodes[getGCNodePathFromLocalId(localId)] = []; }`), while summaries always emit the storage id as an attachment. So when GC marks the real local id (e.g. `/_blobs/l1`) unreferenced, the `l1 → s1` entry is correctly filtered, but the synthesized `s1 → s1` identity entry is not represented in the GC graph and re-adds `s1` to `storageIdsToFetch`. Net effect: a GC-unreferenced blob is captured anyway whenever a redirect entry uses the storage id as its key.

**Suggested fix:** build storage-id reachability from non-identity redirect entries first, and only fall back to an identity mapping when no non-identity local id covers that storage id (or otherwise prove the legacy storage-id handle is itself live).

**Suggested test:** the existing tests at `test/captureReferencedContents.spec.ts:200-219` and `:262-276` cover ordinary GC filtering and absent-GC retention but not the mixed layout. Add a case with one storage id reachable via a live local id and a second reachable only via a GC-unreferenced local id, and assert the second is filtered.

packages/loader/container-loader/src/captureReferencedContents.ts
```diff
@@ -0,0 +1,328 @@
...
+
+	const storageIdsToFetch = new Set<string>();
+	for (const [localId, storageId] of localIdToStorageId) {
+		if (unreferencedLocalIds?.has(localId) !== true) {
+			storageIdsToFetch.add(storageId);
+		}
+	}
+
+	const contents: ISerializableBlobContents = {};
+	await mapWithConcurrency([...storageIdsToFetch], maxReadConcurrency, async (storageId) => {
```
anthony-murphy, line 260: Deep Review: Lossy UTF-8 round-trip silently corrupts non-UTF-8 attachment-blob bytes (Tier 1, wire-format).

```ts
const buffer = await storage.readBlob(storageId);
contents[storageId] = bufferToString(buffer, "utf8");
```

Both `captureReferencedAttachmentBlobs` (encode at ~:243-247, fetch site further down) and `readReferencedSnapshotBlobs` inline arbitrary blob bytes via `bufferToString(buffer, "utf8")`. Attachment blobs are arbitrary user payloads (images, encrypted data, binary). Invalid UTF-8 byte sequences are replaced with `U+FFFD` and the load-side `stringToBuffer(..., "utf8")` in `serializedStateManager.ts:268-273` cannot recover them.

The runtime's own pending-blob serializer uses **base64** for exactly this reason (`blobManager.ts:951-960`, `:331-343`); the attachment APIs accept/return `ArrayBufferLike` (`blobManager.ts:385`, `:453-456`). The historical `getBlobContentsFromTree` / `serializedStateManager.getPendingLocalState` path does **not** inline attachment-blob contents — only the redirect table — so the existing UTF-8 convention does not transfer; this PR introduces a *new* wire-format obligation.

The author's own `full-container-state-review-notes.md` (Part 2 §A, lines 132-143) flags this as "must address before re-review" — still in the diff. The only attachment-blob e2e test uploads `stringToBuffer("attachment-blob-payload", "utf8")` and verifies via `bufferToString(retrievedBlob, "utf8")`, so the failure mode is silent in CI. Reproduces with `Uint8Array([0xff, 0xfe, 0x00, 0x80, 0xc0])`.

**Suggested fix.** Pick one of the two paths the review notes already enumerate:

1. **Preferred**: base64-encode attachment-blob bytes at the encode site in `captureReferencedAttachmentBlobs`; decode in the load path. Extend the wire format with either a separate `binaryAttachmentBlobs` map keyed by storage id or a per-entry encoding flag so old captures still load.
2. **Minimum**: document a UTF-8-only restriction in JSDoc on `ICaptureFullContainerStateProps` and detect/throw on non-UTF-8 round-trip at the encode site (compare `bufferToString(buf,"utf8")` re-encoded back to bytes against the original).

Either way, **add a non-UTF-8 round-trip e2e test** with payload `Uint8Array([0xff, 0xfe, 0x00, 0x80, 0xc0])` asserting byte-exact equality after rehydration through `loadFrozenContainerFromPendingState`.

`readReferencedSnapshotBlobs` is in the same situation in principle — verify whether all callsites guarantee text payloads or whether the same fix applies there too.

**Questions for you:**
- Are there callers (current or planned) of `captureFullContainerState` that use attachment blobs with non-UTF-8 bytes? If "no, by contract," option (2) closes the gap. If "yes" or "unknown," only option (1) is durable.
- For the structural snapshot blobs in `readReferencedSnapshotBlobs` — are all current snapshot-tree blob payloads guaranteed text by container-runtime contract?

packages/loader/container-loader/src/captureReferencedContents.ts
anthony-murphy: Deep Review: Withdrawing this concern. On deeper analysis, `BlobManager`'s load path verifies `isDeleted`/inactive **before** falling through to storage reads (`blobManager.ts:385-392`, `:423-437`), so a synthesized identity redirect entry for a GC-unreferenced storage id does not actually re-surface the blob at runtime. The two-layered GC story (snapshot-tree `unreferenced: true` skipping plus the attachment-blob filter parsing `__tombstones` / `__deletedNodes`) is sufficient. Apologies for the noise — the higher-priority wire-format concern is captured in a new thread on this file.

packages/test/local-server-tests/src/test/wireFormatConstants.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * `container-loader` duplicates a handful of wire-format constants because the
+ * authoritative definitions live in `container-runtime` and
+ * `runtime-definitions`, which the loader cannot depend on. This contract test
+ * imports both copies and asserts they match
+ *
+ * Ideally these never change, if they do great care will be needed
+ * to preserve the correctness of the container-loader code that uses them.
+ */
+describe("wireFormatConstants contract", () => {
```
anthony-murphy, line 30: this seems ok. any strong reason not to push them to like driver definitions? i think that where the other snapshot format keys and interfaces live

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -245,6 +258,165 @@ export async function loadFrozenContainerFromPendingState(
...
+
+		const deltaStorage = await documentService.connectToDeltaStorage();
+		const opsStream = deltaStorage.fetchMessages(
+			attributes.sequenceNumber + 1,
+			undefined,
+			undefined,
+			false,
+			"captureFullContainerState",
+		);
+		const savedOps: ISequencedDocumentMessage[] = [];
```
anthony-murphy, line 398: Deep Review: Post-snapshot `blobAttach` blobs are not inlined into the captured artifact — offline load will fail to resolve those handles (Tier 2, correctness).

`captureFullContainerState` populates `attachmentBlobContents` only via `captureReferencedAttachmentBlobs(baseSnapshot, storage, gcData)` (`captureReferencedContents.ts:235-258`), which walks `baseSnapshot.trees[".blobs"]` and the in-snapshot redirect table. It then drains ops via `fetchMessages(attributes.sequenceNumber + 1, …)` into `savedOps`. A `blobAttach` op carries only `metadata.{ localId, blobId }` (`containerRuntime.ts:2014-2022`); replay only rebuilds the redirect table (`blobManager.ts:750-775`), it never fetches blob bytes. The load-side cache in `containerStorageAdapter.ts:241-247` resolves through `attachmentBlobContents` / `snapshotBlobs` before falling back to live storage.

Net: a blob uploaded *after* `baseSnapshot` but before capture appears in `savedOps` but its bytes never enter the portable artifact. In a frozen-load scenario without live storage (the artifact's stated purpose), the handle is unresolvable.

The new spec files do not cover this path — there is no test that uploads a blob *after* the base snapshot, captures, and round-trips through frozen load. The gap is unmonitored in CI.

**Suggested fix.** While draining `savedOps`, detect `blobAttach` messages and inline their `metadata.blobId` contents into `attachmentBlobContents` if not already present (respecting the same GC filter applied to base-snapshot blobs).

**Suggested test.** In `local-server-tests/src/test/captureFullContainerState.spec.ts`, add a case that (1) attaches a container, (2) takes a base snapshot, (3) uploads an attachment blob, (4) calls `captureFullContainerState`, (5) loads via `loadFrozenContainerFromPendingState` with **no live storage**, and (6) asserts the handle resolves to the original bytes.

packages/loader/container-loader/src/serializedStateManager.ts
```diff
@@ -83,6 +84,21 @@ export interface IPendingContainerState extends SnapshotWithBlobs {
...
+	 * Carried separately from {@link SnapshotWithBlobs.snapshotBlobs} because
+	 * attachment blobs may contain arbitrary binary payloads, and the
+	 * UTF-8 encoding used for `snapshotBlobs` (which holds JSON/text the
+	 * runtime authors) would corrupt non-UTF-8 byte sequences with
+	 * replacement characters. Populated by `captureFullContainerState`; the
+	 * live container's pending-state path leaves this `undefined` because
+	 * it does not inline attachment blob contents.
+	 *
+	 * On load, entries are decoded from base64 and merged into the same
+	 * blob cache that `snapshotBlobs` populates.
```
anthony-murphy, line 99: Deep Review: `IPendingContainerState` serialized wire-format extended without gatekeeper sign-off or forward-compat JSDoc (Tier 2, process + compat).

This PR adds optional `attachmentBlobContents?: IBase64BlobContents` to `IPendingContainerState`, reads it here (`:281-296`), writes it in `pendingLocalStateStore.ts:93-118`. No schema-version bump and no forward/backward-compat JSDoc note on the field itself.

anthony-murphy on PR #20504 (2024-04-08) established the standing rule: *"modifying our serialized format is a big deal, and generally should have a full design review, so finding other ways to do things is generally preferred."* Reinforced on PR #20198: *"this is serialized state, so we shouldn't change it"*; *"any existing usage of the serialized data will break."*

The blast radius is narrowed by the field being optional and the producing API being `@alpha @legacy`, but the core compat behavior is non-trivial: an old loader receiving new-producer state silently drops `attachmentBlobContents`, making attachment blobs unreachable in offline / frozen-load scenarios — the artifact's *only* purpose.

**Suggested fix.**

1. Tag **anthony-murphy** and **dannimad** explicitly on the format extension when leaving draft.
2. Add a JSDoc note on `IPendingContainerState.attachmentBlobContents` that an old loader silently ignores this field and will fail to read attachment blobs in offline / frozen-load scenarios — and that's why the producing API is `@alpha`.
3. Update the changeset (`wide-foxes-behave.md`) to call out the format extension itself, not only the encoding fix.

anthony-murphy:
## Deep Review

*Reviewed commit `1a4bcfa` on 2026-05-06.*

**Readiness: 4/10 — 🔨 MAKING PROGRESS**

Not ready for sign-off. The prior round's UTF-8 wire-format defect is resolved via the new `IBase64BlobContents` split, but two concerns from the last review remain live: (1) attachment blobs uploaded *after* the base snapshot are preserved as `blobAttach` ops in `savedOps` and never base64-inlined into `attachmentBlobContents`, so a truly offline rehydrate silently fails for those handles — flagged inline on `createAndLoadContainerUtils.ts:394`; (2) the `IPendingContainerState` extension that delivered the encoding fix is itself a serialized-format change subject to the documented gatekeeper rule and still lacks a forward/backward-compat JSDoc note — flagged inline on `serializedStateManager.ts:99`. Tier 3 polish (constants keep-in-sync doc, stale review-notes file, loading-groups limitation in changeset, copilot-reviewer wording/diagnostic asks) is held until the Tier 2 fix lands.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Update PR description — remove the resolved "Known open issue: UTF-8 lossy encoding" paragraph and add a one-line note that `IPendingContainerState` was extended with `attachmentBlobContents`
- [ ] Update the changeset (`wide-foxes-behave.md`) to call out the `IPendingContainerState` wire-format extension, not only the encoding fix
- [ ] Tag **anthony-murphy** and **dannimad** on the format extension when leaving draft (documented gatekeepers for `IPendingContainerState` per #20504, #20198)

### Context for Reviewers

- The two-encoding wire-format split (`IBase64BlobContents` for attachment blobs, `ISerializableBlobContents` for snapshot blobs) is a **deliberate fix** for the UTF-8 corruption flagged last round — do not suggest collapsing back onto a single `getBlobContentsFromTree`-style path (PR #21908) without recognizing the asymmetry is intentional.
- `serializedStateManager.ts` is the high-churn file in this surface (7 related PRs: #12498, #19697, #20191, #20198, #20504, #20507, #21026). Even the small ~28-line hunk here carries the wire-format extension that drives the open Tier 2 thread — give it disproportionate attention.
- Loading-group containers are deliberately refused via `UsageError` (no e2e harness, no known consumer). This rules out the offline+groupId customer from PR #20565 (`loadedGroupIdSnapshots` is left `undefined`); fail-fast is preferable to a silent partial artifact, but this is a parity gap to track. The in-tree `full-container-state-review-notes.md` still references a `captureGroupIdSnapshots` helper that was removed in favor of the throw — stale.
- GC handling is two-layered and intentional: snapshot tree walk skips `unreferenced: true` subtrees; attachment-blob filter parses `__tombstones` / `__deletedNodes` with documented GC-lag tolerance. Tombstoned attachment blobs (recoverable at runtime via `allowTombstone` per PR #13266) are foreclosed by the captured artifact — fine for read-only frozen-load, called out in the capture-boundary JSDoc. Anthony-murphy's earlier identity-redirect-table concern (#3184981464) was withdrawn after analysis.
- Seven runtime constants (`blobsTreeName`, `redirectTableBlobName`, `blobManagerBasePath`, `gcTreeKey`, `gcBlobPrefix`, `gcTombstoneBlobKey`, `gcDeletedBlobKey`) are hand-duplicated in the loader to avoid a loader→runtime layering dependency, defended by a 7-assertion cross-package contract test (`wireFormatConstants.spec.ts`). Original consolidation site: PR #21729. Anthony's open thread #3192306888 asks whether they should live in `driver-definitions` instead — worth answering when Tier 2 lands.
- Suggested reviewers when out of draft: **anthony-murphy** (`IPendingContainerState` gatekeeper #20504/#20198, blob-manager constants #21729), **dannimad** (`serializedStateManager` / `IPendingContainerState` evolution #19697/#20191/#20198/#20504; consumer-side via #25653), **tyler-cai-microsoft** (offline+groupId machinery #20565), **agarwal-navin** (attachment-blob tombstone semantics #13266), **jatgarg** (`containerStorageAdapter` blob extraction #21908; `mixinMonitoringContext` convention #25394), **ChumpChief** (alpha-surface scaling #25513).

- **anthony-murphy** — Confirm the wire-format direction for the format extension: an optional field on `IPendingContainerState` with a JSDoc compat note (current) vs. a versioned envelope vs. another shape. Also: should the `IPendingContainerState` literal be constructed in `captureFullContainerState` directly (current) or via a helper in `serializedStateManager` for shape ownership? And: should the duplicated runtime constants live in `driver-definitions` (per thread #3192306888) rather than be hand-duplicated?
- **dannimad** — Procedurally required as the consumer-side owner via #25653 `createFrozenDocumentServiceFactory`; confirm the captured artifact shape meets rehydrate-side expectations.
- **jatgarg** — Procedural sign-off on the `mixinMonitoringContext` / `configProvider` divergence in `createAndLoadContainerUtils.ts` (the convention enforced on the sibling `loadSummarizerContainerAndMakeSummary` in PR #25394). The author's docblock acknowledges the divergence; confirm or require an optional `configProvider?` prop now for forward-compat.
- **ChumpChief** — Alpha-surface scaling. The "how does this scale?" question from PR #25513 remains open across four alpha additions. Resolve the convention (overload of `asLegacyAlpha` vs. standalone free functions).
- **tyler-cai-microsoft** — Confirm the loading-groups `UsageError` is acceptable scope and follow-up tracking captures the parity gap with #20565.
- **agarwal-navin** — Confirm the GC-filter / tombstone-recovery tradeoff for the captured artifact (tombstoned blobs are foreclosed by the artifact, different from the live runtime's `allowTombstone` recoverability).
- **Cannot be assessed by the pipeline**:
  - Real-world memory/throughput on long-lived documents with millions of trailing ops.
  - Whether `mapWithConcurrency` limits (32 blobs / 4 group snapshots) are correct under production driver-throttling regimes (notably ODSP).
  - Whether the captured artifact loaded by `loadFrozenContainerFromPendingState` reproduces identical runtime behavior across all document shapes; the local-server test exercises one round-trip only.

</details>

- `cdda9ba` 2026-05-05 · **4/10** — UTF-8 wire-format defect resolved; post-snapshot `blobAttach` gap and format-extension JSDoc surface
- `1d5a51b` 2026-05-05 · **3/10** — UTF-8 attachment-blob round-trip flagged Tier 1
- `57a1a76` 2026-05-04 · **5/10** — GC-redirect-identity and unbounded-savedOps flagged (both later reclassified)

</details>

packages/test/local-server-tests/src/test/wireFormatConstants.spec.ts
markfields: Opened AB#72934

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
markfields: Fixed in branch

packages/loader/container-loader/src/captureReferencedContents.ts
```diff
@@ -0,0 +1,442 @@
...
+	}
+	for (const [key, subTree] of Object.entries(tree.trees)) {
+		if (isRoot && key === blobsTreeName) {
+			const tableBlobId = subTree.blobs[redirectTableBlobName];
+			if (tableBlobId !== undefined) {
+				ids.add(tableBlobId);
+			}
+		} else {
+			collectReferencedBlobIds(subTree, false, ids);
+		}
```
anthony-murphy, line 199: Deep Review: Base-snapshot `unreferenced` filter drops blobs for nodes that replayed `savedOps` can revive (Tier 2, correctness — self-containment).

`collectReferencedBlobIds` early-returns for any subtree with `tree.unreferenced === true` (`captureReferencedContents.ts:181-195`), so those snapshot blobs are not inlined into `snapshotBlobs`. `captureFullContainerState` then drains all post-snapshot ops via `deltaStorage.fetchMessages(seq+1, …)` into `savedOps` (`createAndLoadContainerUtils.ts:390-425`), and the loader replays those ops during pending-state load.

Fluid GC's `GarbageCollector.addedOutboundReference` records a new outbound reference as `"Revived"` and clears the unreferenced state (`packages/runtime/container-runtime/src/gc/garbageCollection.ts:1110-1168`). If a post-snapshot op stores a handle to a node/data store that was `unreferenced` in the base snapshot but not tombstoned/deleted, the captured artifact represents it as reachable after replay — but that node's base-snapshot blobs were deliberately omitted. A frozen/offline load (the artifact's stated purpose) then either fails or falls through to live storage when resolving the revived handle. The artifact's central self-contained guarantee is broken on a GC-supported transition.

`captureFullContainerState.spec.ts:430-511` covers post-snapshot **attachment** blobs with live `readBlob` disabled, but does not cover revived **unreferenced snapshot subtrees**.

**Suggested fix.** Pick one:

1. **Full fix** — inline `unreferenced` snapshot subtrees too when `savedOps.length > 0`, preserving artifact self-containment for any GC-revival path.
2. **Conservative fix** — fail fast in `captureFullContainerState` when the base snapshot contains any `unreferenced` subtree and `savedOps.length > 0`, rather than emitting a non-self-contained artifact.

The choice is itself an API-shape decision worth surfacing.

**Suggested test.** Take a base snapshot containing an `unreferenced` data-store subtree, perform a post-snapshot op that stores a handle to that subtree (triggering GC revival), `captureFullContainerState`, then rehydrate via `loadFrozenContainerFromPendingState` with live `readBlob` disabled — assert the revived handle resolves to the original blob bytes.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Confirmed resolved in this SHA. `extractBlobAttachReferences` + `inlineAttachmentBlobsByReference` scrape `(localId, storageId)` pairs from `savedOps` `BlobAttach` messages and backfill bytes through the same GC-filtered, base64-encoded path; the local-server integration test exercises a runtime-uploaded post-snapshot blob with live `readBlob` disabled. Note: a related self-containment hole on the same axis (revived `unreferenced` snapshot subtrees via replayed `savedOps`) is flagged in a new thread on `captureReferencedContents.ts`.

packages/loader/container-loader/src/serializedStateManager.ts
markfields: Based on the scenarios here, we don't need to worry much about old loaders trying to interpret this new property

packages/loader/container-loader/src/captureReferencedContents.ts
markfields: Will open a follow-up task for this (noted in PR's root MD file for now)

anthony-murphy:

- 🔄 **Mechanical Review**
  - ✅ mechanical (gpt-5.5)
  - ✅ context (gpt-5.5)
  - ✅ context (claude-opus-4.7)
  - ✅ extract (claude-opus-4.7)
  - 🔄 propose (claude-opus-4.7)
  - ✅ actual (claude-opus-4.7)
  - 🔄 propose (gpt-5.5)

packages/loader/container-loader/src/serializedStateManager.ts
anthony-murphy: Deep Review: Reasonable position for `@alpha @legacy`. To close the loop on the standing `IPendingContainerState`-shape standard (#20504, #20198), add **anthony-murphy** as a requested reviewer (or @-mention on this file and `wide-foxes-behave.md`) so the area gatekeeper's nod is captured in the PR record before merge. The changeset now calls out the extension explicitly, so this is purely a process ask — no code change expected.

packages/loader/container-loader/src/captureReferencedContents.ts
anthony-murphy: Deep Review: Thanks for the follow-up commitment. Two suggestions to keep the artifact's stated self-containment guarantee honest until the follow-up lands:

1. Add a JSDoc note on `captureFullContainerState` calling out revived `unreferenced` subtrees as a known gap (so consumers don't read the central self-containment guarantee at face value).
2. Consider a conservative interim — fail-fast with `UsageError` when the base snapshot contains any `unreferenced` subtree and `savedOps.length > 0`, mirroring the loading-groups `UsageError` posture and preventing emission of a non-self-contained artifact for the only document shape that exercises the gap.

Worth gating the follow-up to land before this API graduates from `@alpha @legacy`.

packages/loader/container-loader/src/captureReferencedContents.ts
```diff
@@ -0,0 +1,442 @@
...
+		const blobContents = snapshot.blobContents;
+		return {
+			tree: snapshot.snapshotTree,
+			read: async (id) => blobContents.get(id) ?? storage.readBlob(id),
+		};
+	}
+	return { tree: snapshot, read: async (id) => storage.readBlob(id) };
+}
+
+/**
```
anthony-murphy, line 217: Deep Review: UTF-8/base64 encoding split — process sign-off ask, no code change.

`captureReferencedAttachmentBlobs` writes base64 (`bufferToString(buffer, "base64")`) into a new `IBase64BlobContents`; `readReferencedSnapshotBlobs` writes UTF-8 into `ISerializableBlobContents` (`serializedStateManager.ts:284-299`, decode in `containerStorageAdapter.ts:515-527`). The split is technically correct and aligns with the runtime's existing pending-blob serializer (`blobManager.ts:340-343`, `:951-960`); the non-UTF-8 round-trip e2e test at `captureFullContainerState.spec.ts:351-427` covers the binary case.

Process concern: vladsud's *"we should stick to one format across layers for blobs"* position from PR #20507 (with **anthony-murphy** explicitly looped in) hasn't been revisited on the record. **vladsud** is already requested; please also tag **anthony-murphy** so the area's "one format" stance is consciously superseded rather than silently bypassed. The binary-safety argument for the split is sound — no code change expected.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -245,6 +262,185 @@ export async function loadFrozenContainerFromPendingState(
...
+	readonly documentServiceFactory: IDocumentServiceFactory;
+	/**
+	 * The request identifying the container whose state is to be captured.
+	 */
+	readonly request: IRequest;
+	/**
+	 * Optional logger for driver-side telemetry.
+	 */
+	readonly logger?: ITelemetryBaseLogger | undefined;
+}
```
anthony-murphy, line 287: Deep Review: Add `abortSignal?: AbortSignal` to `ICaptureFullContainerStateProps` before `@alpha` graduation.

`ICaptureFullContainerStateProps` exposes only `urlResolver`, `documentServiceFactory`, `request`, `logger`. The capture body concurrently reads snapshot/attachment blobs (bounded concurrency 32) and drains the op stream without a cancellation check. `IDocumentDeltaStorageService.fetchMessages` already accepts `abortSignal?: AbortSignal` (`packages/common/driver-definitions/src/storage.ts:103-109`), so the underlying primitives support it.

Suggested fix: add `abortSignal?: AbortSignal` to `ICaptureFullContainerStateProps`, propagate it into the `mapWithConcurrency` blob/group loops and the `fetchMessages` op-drain loop, and call `signal.throwIfAborted()` between bounded batches. Additive and cheaper to land before the API graduates from `@alpha`.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -245,6 +262,185 @@ export async function loadFrozenContainerFromPendingState(
...
+
+		if (postSnapshotBlobReferences.length > 0) {
+			const added = await inlineAttachmentBlobsByReference(
+				postSnapshotBlobReferences,
+				storage,
+				unreferencedAttachmentBlobLocalIds(gcData),
+				attachmentBlobContents,
+			);
+			Object.assign(attachmentBlobContents, added);
+		}
```
anthony-murphy, line 425: Deep Review: Op-stream drain trusts the driver silently — assert contiguity to fail closed.

`captureFullContainerState` drains `connectToDeltaStorage().fetchMessages(attributes.sequenceNumber + 1, …)` into `savedOps` and emits whatever the driver yields. Other parts of the same function fail-closed on suspicious states (`UsageError` on loading groups; `GenericError` on empty snapshot fetch); ops are the one place that trusts the driver silently. A buggy driver / network truncation would produce an artifact whose ops cannot be replayed, surfacing as a confusing rehydration-time failure pointing at the loader rather than at the capture.

Suggested fix: track a previous-sequence-number across the drain loop and assert each op's `sequenceNumber === prev + 1`; throw on a gap with the offending pair in the message. Optional: also assert the snapshot's `latestSequenceNumber >= attributes.sequenceNumber` at fetch time.

packages/loader/container-loader/src/captureReferencedContents.ts
```diff
@@ -0,0 +1,442 @@
...
+ */
+export const wireFormatConstants = {
+	blobsTreeName: ".blobs",
+	redirectTableBlobName: ".redirectTable",
+	blobManagerBasePath: "_blobs",
+	gcTreeKey: "gc",
+	gcBlobPrefix: "__gc",
+	gcTombstoneBlobKey: "__tombstones",
+	gcDeletedBlobKey: "__deletedNodes",
+} as const;
```
anthony-murphy, line 43: Deep Review: Loader↔runtime constant duplication — gatekeeper nod ask, no code change.

This PR introduces a `wireFormatConstants` block in the loader mirroring seven runtime-side constants (`.blobs`, `.redirectTable`, `_blobs`, `gc`, `__gc`, `__tombstones`, `__deletedNodes`) and a cross-package `deepStrictEqual` contract test (`wireFormatConstants.spec.ts`). PR #21729 (**anthony-murphy**, 2024-07-02) was the deliberate refactor that produced these constants in their current form, with markfields reviewing for path-format discipline. This PR is the first cross-layer consumer; the chosen pattern (duplicate + contract test) sets precedent for any future shared wire-format constant.

Tag **anthony-murphy** for an explicit nod on the duplicate+contract-test layering choice, or merge with a clear note that AB#72934 will revisit (per thread #3192306888). No code change expected.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
markfields: This is a good idea, adding it to follow-ups mentioned in PR description

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
markfields: Another good point. A downside to not using the DeltaManager code in the container loader, I guess.  Tracking as a follow-up.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -245,6 +262,185 @@ export async function loadFrozenContainerFromPendingState(
...
+ * no feature flags and instantiates no runtime, so there is nothing for a
+ * monitoring context to gate or attribute. If a future change introduces
+ * config-gated behavior or runtime-attributed telemetry, add the wiring
+ * back together with that change.
+ * @legacy @alpha
+ */
+export async function captureFullContainerState({
+	urlResolver,
+	documentServiceFactory,
+	request,
```
markfields, line 336: @anthony-murphy thinking about replacing these 3 params with just `documentService` and having caller piece that together (maybe in follow-up PR).  Does that work?  I don't totally understand the environment in which this will be called and where these params will come from

anthony-murphy: its tricky. overall i would love to move in a direction where fluid takes a document service, but it could be hard right now due lack of usage that way. there won't be any app examples, and i'm not sure the odsp driver supports it.  i would be 100% down to support both maybe via overload or union to start moving in that direction, as we need to start somewhere

anthony-murphy (approved):

packages/loader/container-loader/src/serializedStateManager.ts
```diff
@@ -83,6 +84,21 @@ export interface IPendingContainerState extends SnapshotWithBlobs {
...
+	/**
+	 * Attachment blob contents inlined by storage id, encoded as base64.
+	 *
+	 * Carried separately from {@link SnapshotWithBlobs.snapshotBlobs} because
+	 * attachment blobs may contain arbitrary binary payloads, and the
+	 * UTF-8 encoding used for `snapshotBlobs` (which holds JSON/text the
+	 * runtime authors) would corrupt non-UTF-8 byte sequences with
+	 * replacement characters. Populated by `captureFullContainerState`; the
+	 * live container's pending-state path leaves this `undefined` because
+	 * it does not inline attachment blob contents.
```
vladsud, line 96: It's probably much larger change (and should be considered on its own), but I wonder if something like directory / map should support arrays as payloads? In addition to strings. But that would likely require massive changes across whole stack.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
vladsud: I'd vote for this direction. I think we can expose both - the core function that takes instance, and wrapper that takes factory. Might need to make some changes to the service interface. For examlpe, it would be nice to get resolved URI from service (assuming it's not exposed today) and thus get rid of URI resolver dependency as well

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -245,6 +262,185 @@ export async function loadFrozenContainerFromPendingState(
...
+	}
+
+	const documentService = await documentServiceFactory.createDocumentService(
+		resolvedUrl,
+		logger,
+	);
+	try {
+		const storage = await documentService.connectToStorage();
+
+		const versions = await storage.getVersions(
```
vladsud, line 351: any opportunities to dedup some code with Container.load()?

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -245,6 +262,185 @@ export async function loadFrozenContainerFromPendingState(
...
+ * config-gated behavior or runtime-attributed telemetry, add the wiring
+ * back together with that change.
+ * @legacy @alpha
+ */
+export async function captureFullContainerState({
+	urlResolver,
+	documentServiceFactory,
+	request,
+	logger,
+}: ICaptureFullContainerStateProps): Promise<string> {
```
vladsud, line 338: If I got it right, this flow does not have websocket connection and fetching ops from websocket.
It might be enough for prototype, but we would need ops from socket (replace with REST call once/if we have such API from SPO in the future).

packages/loader/container-loader/src/test/captureReferencedContents.spec.ts
```diff
@@ -0,0 +1,388 @@
+/*!
```
vladsud, line 1: Do we need some kind of stress test system here? I.e. a collection of random snapshots that go through new code path vs. full container load (with container runtime instantiation) and check that they produce same result in terms of ops fetched, blobs captured?
That assumes that we have equvalent flow for full container load (which we do not), so maybe we can't imlement it directly. But maybe some parts of it can be implemented, like asking GC subsystem for referenced blobs and compare that with what container-loader layer computed.

I just worry that as we make changes in runtime, they may not reflect in this layer.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
markfields: L390 we connect to delta storage and fetch messages. I'll have to look at what that's doing... is it possible it's already that REST API we're talking about? I forget and on mobile at the moment.

markfields: Ohhh that's Storage's deltas, not PUSH, right? So would be missing the very latest messages?

packages/loader/container-loader/src/test/captureReferencedContents.spec.ts
anthony-murphy: this could be integrated into local-server-stress, which already does pending state capture from live containers

.changeset/wide-foxes-behave.md
```diff
@@ -0,0 +1,10 @@
+---
+"@fluidframework/container-loader": minor
+"@fluidframework/container-runtime": minor
+"__section": feature
+---
+Add a new free function `captureFullContainerState` for serializing a container without loading it, for rehydrating into a future session.
```
jzaffiro:
```suggestion
Serialize a container without loading it for rehydrating a future session with `captureFullContainerState`
```
I would mention that it's a free function somewhere in the longer description as opposed to the title.

markfields: I'm going to drop the changeset, we don't need to advertise this at the moment.  We can do so when promoting to beta.

anthony-murphy:
## Deep Review

*Reviewed commit `ad41359` on 2026-05-08.*

**Readiness: 8/10 — 🟢 ALMOST READY**

No Tier 1 or Tier 2 issues remain. The two newest concrete asks (`AbortSignal` plumbing on `ICaptureFullContainerStateProps`, op-stream contiguity assert in the `fetchMessages` drain) are author-accepted as follow-ups. What's left are governance and deferred-capability items — gatekeeper acks from **anthony-murphy** on the wire-format extension, UTF-8/base64 split, and loader-side constant duplication; the revived-`unreferenced`-subtree JSDoc caveat; an `attachmentBlobContents` forward-compat TSDoc; and confirming the loading-group fail-fast is tracked against an ADO item with a target consumer.

### Path to Ready

- [ ] Resolve inline threads — link the AbortSignal (#3204969959) and op-contiguity-assert (#3204970080) follow-up ADO items from each thread, then resolve
- [ ] Add **anthony-murphy** as a requested reviewer (or @-mention on `serializedStateManager.ts`, `captureReferencedContents.ts`, and the changeset file) so the wire-format extension, UTF-8/base64 encoding split, and loader-side constant duplication are explicitly acked by the area's gatekeeper (closes the process side of three threads — #3192334059, #3204969859, #3204970184 — as one batch)
- [ ] Add a JSDoc note on `captureFullContainerState` calling out revived-`unreferenced` subtrees as a known self-containment caveat (#3202581532); decide whether the conservative interim `UsageError` (when base snapshot has any `unreferenced` subtree AND `savedOps.length > 0`) lands before `@alpha` graduation or whether the follow-up ADO item lands first
- [ ] Add a TSDoc note on `IPendingContainerState.attachmentBlobContents` covering the forward-compat asymmetry (an older `container-loader` reading state written by a newer `captureFullContainerState` silently drops the field, making attachment blobs unreachable in offline / frozen-load scenarios)
- [ ] Confirm the loading-group `UsageError` deferral is tracked against an ADO item with a target consumer (the PR risk-section text alone — "no known consumer + no e2e harness" — does not auto-prioritize)

### Context for Reviewers

- Fifth in the `asLegacyAlpha` / frozen-container series (predecessors #25513, #25538, #25590, #25653, #25742) — together with `loadFrozenContainerFromPendingState`, this PR enables fully-offline frozen-container scenarios.
- The two-encoding wire-format split (`IBase64BlobContents` for attachment blobs, `ISerializableBlobContents` for snapshot blobs) is a **deliberate** binary-safety fix mirroring the runtime's own pending-blob serializer (`blobManager.ts:340-343`, `:951-960`). Do not suggest collapsing back to a single `getBlobContentsFromTree`-style path (PR #21908) without recognising the asymmetry is intentional. vladsud's "one format across layers for blobs" position from PR #20507 is being consciously superseded by the binary-safety rationale; thread #3204969859 is the place to capture that.
- Post-snapshot `BlobAttach` op-metadata backfill (`extractBlobAttachReferences` + `inlineAttachmentBlobsByReference`) closes the prior round's offline-blob gap; verified by the local-server test exercising a runtime-uploaded post-snapshot blob with live `readBlob` disabled (`captureFullContainerState.spec.ts:430-511`). Loader-side decode was originally introduced by **wes-carlson** in PR #10621.
- Loading-group containers are deliberately refused via `UsageError` (no e2e harness, no known consumer); `loadedGroupIdSnapshots` is left `undefined`. Fail-fast is preferable to a silent partial artifact, but it is a parity gap with PR #20565 (**tyler-cai-microsoft**) to track.
- GC handling is two-layered and intentional: snapshot-tree walk skips `unreferenced: true` subtrees; the attachment-blob filter parses `__tombstones` / `__deletedNodes` with documented GC-lag tolerance. Tombstoned attachment blobs (recoverable at runtime via `allowTombstone` per PR #13266 by **agarwal-navin**) are foreclosed by the captured artifact — fine for read-only frozen-load, called out in the capture-boundary JSDoc.
- Seven runtime constants are hand-duplicated in the loader to avoid a loader→runtime layering dependency, defended by a 7-assertion cross-package contract test (`wireFormatConstants.spec.ts`). Original consolidation site: PR #21729 (**anthony-murphy**). Open thread #3192306888 (driver-definitions home) tracked by author via AB#72934.

- **anthony-murphy** — Wire-format direction: optional field with JSDoc compat note (current) vs versioned envelope vs another shape; whether the `IPendingContainerState` literal should be constructed in `captureFullContainerState` directly or via a helper in `serializedStateManager` for shape ownership; whether the seven duplicated runtime constants belong in `driver-definitions` per AB#72934. Confirms or retires the "one format across layers" stance from PR #20507 in light of the binary-safety rationale.
- **dannimad** — Procedurally required as consumer-side owner via #25653 (`createFrozenDocumentServiceFactory`); confirm the captured artifact shape meets rehydrate-side expectations.
- **jatgarg** — Procedural sign-off on the `mixinMonitoringContext` / `configProvider` divergence in `createAndLoadContainerUtils.ts` (the convention enforced on the sibling `loadSummarizerContainerAndMakeSummary` in PR #25394). Author's docblock acknowledges the divergence; confirm or require an optional `configProvider?` prop now for forward-compat.
- **ChumpChief** — Alpha-surface scaling: the "how does this scale?" question from PR #25513 remains open across four alpha additions; resolve the convention (overload of `asLegacyAlpha` vs standalone free functions).
- **tyler-cai-microsoft** — Confirm the loading-groups `UsageError` is acceptable scope and follow-up tracking captures the parity gap with PR #20565.
- **agarwal-navin** — Confirm the GC-filter / tombstone-recovery tradeoff for the captured artifact (tombstoned blobs foreclosed by the artifact, different from the live runtime's `allowTombstone` recoverability), and weigh in on the revived-`unreferenced` deferral.
- **vladsud** — Already engaged on threads #3209182218 (binary payloads in directory/map), #3209209246 (dedup vs `Container.load()`), #3209227685 (websocket vs storage-deltas op fetch); these are forward-looking design discussions, not blockers for this PR.
- **Cannot be assessed by the pipeline**:
  - Real-world memory/throughput on long-lived documents with millions of trailing ops.
  - Whether `mapWithConcurrency` limits (32 blobs / 4 group snapshots) are correct under production driver-throttling regimes (notably ODSP).
  - Whether the captured artifact loaded by `loadFrozenContainerFromPendingState` reproduces identical runtime behavior across all document shapes; the local-server test exercises one round-trip only.
  - vladsud's stress-test ask in thread #3209258894 (random snapshots through capture vs full-container-load comparison) — anthony-murphy suggested integrating into local-server-stress; design judgment call on scope.

</details>

- `c21964d` 2026-05-07 · **7/10** — polish/process items remain — gatekeeper acks, `AbortSignal`, op-contiguity assert
- `6e66f7b` 2026-05-07 · **4/10** — post-snapshot `blobAttach` gap resolved; new revived-`unreferenced` self-containment hole and wire-format-extension JSDoc still open
- `1a4bcfa` 2026-05-06 · **4/10** — UTF-8 wire-format defect resolved; post-snapshot `blobAttach` gap and format-extension JSDoc surface
- `cdda9ba` 2026-05-05 · **4/10** — UTF-8 wire-format defect resolved; post-snapshot `blobAttach` gap and format-extension JSDoc surface
- `1d5a51b` 2026-05-05 · **3/10** — UTF-8 attachment-blob round-trip flagged Tier 1
- `57a1a76` 2026-05-04 · **5/10** — GC-redirect-identity and unbounded-`savedOps` flagged (both later reclassified)

</details>

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Thanks — accepting as a follow-up is reasonable given the API is still `@alpha @legacy`. One ask: gate the follow-up to land before graduation. The prop is additive, so it's cheaper now than once the surface is frozen; please link the ADO item here when filed and resolve the thread.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Thanks for tracking this. The contiguity assert keeps the artifact's replay invariants honest at capture time and matches the fail-closed posture used elsewhere in `captureFullContainerState` (`UsageError` on loading groups, `GenericError` on empty snapshot fetch). Please link the follow-up ADO item here when filed and resolve the thread; ideally land before `@alpha` graduation alongside #3204969959.

packages/loader/container-loader/src/captureReferencedContents.ts
```diff
@@ -0,0 +1,446 @@
...
+		return added;
+	}
+	await mapWithConcurrency([...storageIdsToFetch], maxReadConcurrency, async (storageId) => {
+		const buffer = await storage.readBlob(storageId);
+		added[storageId] = bufferToString(buffer, "base64");
+	});
+	return added;
+}
+
+/**
```
anthony-murphy, line 422: Deep Review: Loading-group fail-fast — confirm follow-up tracking against a target consumer.

`snapshotHasLoadingGroups(baseSnapshot)` throws `UsageError` whenever any referenced subtree carries a `groupId`, and `captureFullContainerState` leaves `loadedGroupIdSnapshots: undefined`. Fail-fast is the right interim for an `@alpha @legacy` surface — preferable to emitting an incomplete artifact — and consistent with PR #20565's pending-state model for loading groups (**tyler-cai-microsoft**).

No code change requested. Two asks:

1. Confirm the deferral is tracked against an open ADO/work item with a target consumer. The PR risk-section text alone ("no known consumer + no e2e harness") does not auto-prioritize; once a consumer surfaces, the per-`groupId` `storage.getSnapshot({ loadingGroupIds: [groupId], cacheSnapshot: false })` path needs to follow PR #20565's storage convention rather than inventing a new one.
2. Tag **tyler-cai-microsoft** to confirm the parity gap with PR #20565 is acceptable scope for v1.

anthony-murphy (approved):

PR #27221, opened 2026-05-01
Update pnpm store cache-keys to include compat-workspaces pnpm-lock.yaml

Abe27342:
## Description

Follow-up to #27186 which updates the cache keys we use for pnpm module stores to better reflect the new status of installing at the repo root where appropriate (i.e. compat testing dependencies now are always installed).

Test build links:

[e2e tests](https://dev.azure.com/fluidframework/internal/_build/results?buildId=397767&view=results)
[stress tests](https://dev.azure.com/fluidframework/internal/_build/results?buildId=397768&view=results)

### DIFF
--- a/tools/pipelines/templates/include-install-pnpm.yml
+++ b/tools/pipelines/templates/include-install-pnpm.yml
@@ -32,8 +32,12 @@ parameters:
   default: $(ado-feeds-primary-registry)
 
 # An optional extra segment appended to the pnpm store cache key (separated by |).
-# Pass a file path to have its contents hashed into the key — useful when installation
-# pulls in additional lockfiles (e.g. compat-workspaces/full/pnpm-lock.yaml).
+# Note that Azure pipelines interpret unquoted cache keys as files rather than literal strings. E.g. the following specification:
+#
+# additionalCacheKey: 'foo | "bar"'
+#
+# will use the hashed contents of the file "foo" (hashing nothing if that file does not exist) as well as the string "bar".
+# Use this parameter if the set of dependencies to be installed may be be better distinguished using files other than the root pnpm-lock.
 - name: additionalCacheKey
   type: string
   default: ''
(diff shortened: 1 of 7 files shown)


tools/pipelines/templates/include-test-real-service.yml
```diff
@@ -231,15 +251,15 @@ stages:
...
+            # whenever legacy package versions are added or removed. Including it in the pnpm store
+            # cache key ensures a cache miss whenever the set of compat packages changes.
+            additionalCacheKey: '$(Pipeline.Workspace)/$(FluidFrameworkDirectory)/packages/test/test-version-utils/compat-workspaces/full/pnpm-lock.yaml'
 
         - template: /tools/pipelines/templates/include-setup-npmrc-for-download.yml@self
 
         # Set up Docker environment if running against docker
         - ${{ if eq(parameters.runAgainstDocker, true) }}:
-          # Checks out FluidFramework repo
-          - checkout: self
```
Abe27342, line 240: the "self" -> "same version as pipeline artifact" is technically a behavior change. But pretty sure it's one we want (rather than possibly version skew the test workload and the docker environment)

alexvy86: Agreed.

tools/pipelines/templates/include-install-pnpm.yml
```diff
@@ -31,6 +31,13 @@ parameters:
   type: string
   default: $(ado-feeds-primary-registry)
 
+# An optional extra segment appended to the pnpm store cache key (separated by |).
+# Pass a file path to have its contents hashed into the key — useful when installation
+# pulls in additional lockfiles (e.g. compat-workspaces/full/pnpm-lock.yaml).
+- name: additionalCacheKey
```
alexvy86, line 41: Given what we learned recently about constant string vs path-like value, I'd suggest this to make it extra clear and prevent potential mistakes.
```suggestion
- name: additionalFilePathCacheKey
```

alexvy86 (approved):
One small thing but otherwise seems good.

tools/pipelines/templates/include-install-pnpm.yml
Abe27342: Given you can get a string literal by passing a quoted string, I'd rather document the gotcha. Done :)

alexvy86 (approved):

PR #27223, opened 2026-05-04
test(tree): add staged optional schema snapshots

noencke:
## Description

Adds staged optional field cases to the SharedTree stored schema snapshot coverage. These cases mirror the staged allowed-types coverage and capture before-update, after-update, root-field, and nested staged optional rollout states.

The tests verify that staged optional fields use the existing persisted schema shapes: restrictive schemas keep staged optional fields required, permissive schemas include them as optional, and the permissive staged-optional schema matches the fully optional after-update schema.

Also adds a new test case for staged allowed types that was lacking some coverage.

## Reviewer Guidance

The main thing to review is whether the new staged optional test documents are the right analogs to the existing staged allowed-types snapshot cases.

### DIFF
--- a/packages/dds/tree/src/test/simple-tree/toStoredSchema.spec.ts
+++ b/packages/dds/tree/src/test/simple-tree/toStoredSchema.spec.ts
@@ -7,7 +7,12 @@ import { strict as assert } from "node:assert";
 
 import { validateUsageError } from "@fluidframework/test-runtime-utils/internal";
 
-import { EmptyKey } from "../../core/index.js";
+import {
+	EmptyKey,
+	ObjectNodeStoredSchema,
+	type TreeNodeSchemaIdentifier,
+	type TreeStoredSchema,
+} from "../../core/index.js";
 import {
 	cursorForJsonableTreeField,
 	defaultSchemaPolicy,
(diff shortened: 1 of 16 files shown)


packages/dds/tree/src/test/simple-tree/toStoredSchema.spec.ts
```diff
@@ -202,6 +226,43 @@ describe("toStoredSchema", () => {
 					});
 				});
 			}
+
+			it("partially includes staged allowed types by upgrade identity", () => {
```
jenn-le: I don't think I understand the point of these tests very well. They don't seem to be testing toStoredSchema? What exactly are they mean to be testing? From the name of the test, I would expect it to be with a custom includeStaged filter passed in

noencke: They are testing `toStoredSchema` because they are loading the configs from the other file (testTrees.ts) which is where the actual call to `toStoredSchema` is. I added some explanatory comments that justify the test and make it easy to connect the dots. Admittedly it's probably not the most important test in the world - so I collapsed the two tests into one case - but I think I agree with the agent that it technically does cover an otherwise uncovered case so I think it is worth keeping.

jenn-le (approved):

PR #27230, opened 2026-05-05
docs: clarify API skill export reachability

noencke:
## Description

Updates the API changes skill to explicitly require new customer-facing exports to be reachable from the package's public entrypoint. This clarifies that agents should bubble intended public APIs through every relevant `index.ts` barrel up to the package root, rather than stopping at a module-adjacent export.

This helps prevent incomplete API changes where API Extractor may not report the intended API and consumers cannot import it from the package's top-level entrypoint.

## Reviewer Guidance


tylerbutler (approved):
LGTM

PR #27232, opened 2026-05-05
ci: harden PR review fleet workflow

tylerbutler:
## Description

Hardens the PR Review Fleet workflow against prompt-injected PR diffs by removing shell/Git tool access from the credentialed Copilot reviewer process.

The workflow now precomputes review context before `COPILOT_GITHUB_TOKEN` is in scope:
- `pr-diff.patch`
- `changed-files.txt`
- `api-report-files.txt`

The Copilot reviewer step keeps only `read` and `write` tools, checks out the exact captured `head_sha`, and disables persisted checkout credentials. Reviewer prompts now direct agents to use the prepared context files instead of requesting shell/Git commands.

Static workflow tests cover the security invariants and label-only dispatch behavior.

## Reviewer Guidance

Please focus on the GitHub Actions security boundary: Git context should be prepared before the Copilot token is in scope, and the credentialed Copilot step should not have shell/Git access.


.github/prompts/reviewers/api-compatibility.md
```diff
@@ -95,8 +95,10 @@ If you find NO high-confidence issues:
...
 
-1. Read the PR diff from `pr-diff.patch` in the current directory
-2. Check if any `*.api.md` files changed: `git diff --name-only origin/$BASE_REF...HEAD -- '*.api.md'` — if so, give those packages extra scrutiny
-3. For files that export public APIs, read the full file and any related `index.ts` barrel exports
-4. Apply the high-confidence gate to every finding before including it
-5. Write your review to `review-api-compatibility.json`
+1. Read the prepared PR diff from `pr-diff.patch` in the current directory
+2. Read `changed-files.txt` when you need the complete changed-file list
+3. Read `api-report-files.txt` to see whether any `*.api.md` files changed. If the file is non-empty, give those packages extra scrutiny.
+4. Do not request or run shell/Git commands; all review context available to you has been precomputed by the workflow
```
noencke: The other steps can (sort of, mostly) be read as instructions to do in sequence, whereas this new item is more of a general warning than it is a particular step. I'd suggest moving it above the list and making it its own prominent warning that the agent reads before starting the instruction steps.

Same for the other reviewer files.

.github/scripts/test_pr_review_dispatch_workflow.py
```diff
@@ -0,0 +1,27 @@
...
+from pathlib import Path
+
+
+WORKFLOW = Path(__file__).parents[1] / "workflows" / "pr-review-dispatch.yml"
+
+
+def test_dispatcher_runs_only_when_fleet_label_is_added() -> None:
+    text = WORKFLOW.read_text(encoding="utf-8")
+
+    assert "- labeled" in text
```
noencke: This file could use some documentation/comments. I have no idea why we don't want "synchronize" in our YAML but we do want "labelled". I think these were the tests you mentioned that have dubious value. I am definitely on team Dubious. Probably just delete them? IMO we should be doing thorough security reviews at PR time, rather than deterministic (and surely full of gaps) tests at runtime that give us a false sense of security.

noencke: These seems like the kind of tests that I might write as sanity checks during development and then delete them. Perhaps the agent is doing that same kind of process.

noencke (approved):

PR #27233, opened 2026-05-05
test: Update cross-client testing matrix based on compatibility checkpoints 

scottn12:
## Description

This PR aligns the e2e cross-client compat matrix with the time-based Compatibility Checkpoint policy from #27064. The changes in this PR ensure we only test the version combinations we explicitly support from the policy.

- Adds `src/checkpoints.ts` — designated checkpoint list, source of truth for `CompatibilityCheckpoints.md‎` table.
- `genCrossClientCompatConfig` now pairs the current build against each in-window prior checkpoint (CC-(N-1), CC-(N-2), CC-(N-3)) in both directions, resolved to the earliest minor of each checkpoint range (`~lowerBoundVersion`).
- `update-compat-versions` resolves the install set from checkpoint data instead of the legacy fast/slow-train deltas.
- Updated documentation to align with these changes

Updated example test matrix:
```
compat cross-client - create with 2.103.0 (N) + load with 2.40.0 (CC-3)
  ✔ Example test
compat cross-client - create with 2.103.0 (N) + load with 2.0.9 (CC-2)
  ✔ Example test
compat cross-client - create with 2.103.0 (N) + load with 1.4.0 (CC-1)
  ✔ Example test
compat cross-client - create with 2.40.0 (CC-3) + load with 2.103.0 (N)
  ✔ Example test
compat cross-client - create with 2.0.9 (CC-2) + load with 2.103.0 (N)
  ✔ Example test
compat cross-client - create with 1.4.0 (CC-1) + load with 2.103.0 (N)
  ✔ Example test
```

## Considerations
- With these changes, we may drop test support for a CC before advancing `lowestMinVersionForCollab`. This means that customers could still technically use version combinations that are not tested. However, this will only ever apply to version combinations that are explicitly unsupported by our compat policy.

## Validation

- **Unit tests:** `pnpm test` in `packages/test/test-version-utils`: new coverage for `checkpoints.ts`
- **Doc generation:** `pnpm run update-compat-versions` runs end-to-end and reports `CompatibilityCheckpoints.md table is already up to date`, confirming the embedded table matches the data in `checkpoints.ts`.
- **Cross-client matrix changed as intended:** new permutations pair the current build against each in-window prior CC (resolved via `~lowerBoundVersion`) in both directions - see the example test matrix above.
- **Layer-compat unchanged:** dumped `configList.value` on this branch and on `origin/main` and the versions combinations are unchanged.
## Misc

[AB#72039](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/72039)


CrossClientCompatibilityDevGuide.md
```diff
@@ -219,9 +224,10 @@ To tighten runtime enforcement:
    (e.g., `2.x` → `3.x`), also narrow the `MinimumVersionForCollab` type in
    [compatibilityDefinitions.ts](./packages/runtime/runtime-definitions/src/compatibilityDefinitions.ts)
    to drop the now-unsupported major from its definition.
-3. **Update the e2e test matrix:** The `FullCompat` version matrix is derived
-   from the currently supported checkpoints — update it so tests only run
-   against versions within the new window.
+3. **Update the e2e test matrix:** The `FullCompat` and cross-client matrices
```
agarwal-navin: If I understand correctly, there is no need to update the e2e test matrix right? If so, I would either remove this point or rename it and its description to clearly state that no action is required and this happens automatically.

packages/test/test-version-utils/README.md
```diff
@@ -147,7 +145,8 @@ default set of legacy versions whether it is necessary or not.
 
 ## Updating compat versions
 
-After a Fluid Framework version bump, run from this package's directory:
+After a Fluid Framework version bump or after a new compatibility checkpoint is designated, run from this package's
+directory:
 
 ```
 pnpm run update-compat-versions
```
agarwal-navin, line 152: Is this part of the release process documentation? Also, will it happen as part of running the release script?

packages/test/test-version-utils/package.json
```diff
@@ -40,6 +40,7 @@
 		"eslint:fix": "eslint --quiet --format stylish src --fix --fix-type problem,suggestion,layout",
 		"format": "npm run format:biome",
 		"format:biome": "biome check . --write",
+		"generate-checkpoints-doc": "jiti scripts/generateCheckpointsDoc.ts",
```
agarwal-navin: Could you explain the difference between this command and `update-compat-versions`. Especially, when they should be run in the process of designating a new checkpoint / making a release. It is getting confusing and hard to keep track of the various steps.
I would suggest re-thinking this process to make it as simple as possible including the commands to update the compat versions + docs.

agarwal-navin: Maybe there should be just one command that does both?

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -205,46 +202,50 @@ async function main(): Promise<void> {
...
+		checkpointVersions.push(resolved);
+		console.log(`  ${checkpoint.name} (${range}): ${resolved}`);
+	}
+	console.log(`  ocv: ${oldestCompatibleVersion}`);
+
+	// Resolve the layer-compat adjacent-minor version (N-1). Used by `defaultVersionsForLayerCompat`,
+	// which keeps adjacent-minor semantics because Fluid's intra-client layer-compat checks are
+	// stricter than the cross-client window.
+	console.log("\nResolving layer-compat (N-1)...");
+	const nMinus1 = resolveVersionForMinorDelta(pkgVer, -1);
```
agarwal-navin, line 243: Before this change, this logic was getting nMinus2 as well. Is that not needed anymore or is it calculated as part of the for loop later?

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -205,46 +202,50 @@ async function main(): Promise<void> {
...
-	console.log("\nResolving cross-client slow-train versions...");
-	const crossClientVersions: string[] = [];
-	for (const delta of [-1, -2]) {
-		const resolved = resolveVersionForMajorDelta(pkgVer, delta);
-		if (resolved === undefined || !semver.gte(resolved, "1.0.0")) {
-			console.log(`  Slow-train N${delta}: not found or below 1.0.0, skipping`);
-			continue;
-		}
+	const checkpointVersions: string[] = [];
+	const seenVersions = new Set<string>([oldestCompatibleVersion]);
```
agarwal-navin, line 221: A little confused by this because cross-client compat doesn't have anything to do with `oldestCompatibleVersion` right. Is this logic combined for both layer-compat and cross-client compat?
Can you add comments explaining this? If possible, separating out layer-compat and cross-client compat logic would make this easier to understand.

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -205,46 +202,50 @@ async function main(): Promise<void> {
...
-		throw new Error("Could not resolve N-1 or N-2. Check registry connectivity.");
-	}
-
-	console.log(`  n-1: ${nMinus1}`);
-	console.log(`  n-2: ${nMinus2}`);
-	console.log(`  ocv: ${oldestCompatibleVersion}`);
-
-	// seenVersions tracks all resolved versions to avoid duplicates across tiers.
-	const seenVersions = new Set([nMinus1, nMinus2, oldestCompatibleVersion]);
+	// Resolve the in-window prior Compatibility Checkpoints that drive the
```
agarwal-navin, line 210: Just from reading these changes, it's hard to tell whether they are doing the right thing or not - particularly for layer-compat because that should not change. Can you add make sure that has not regressed and add the point including validation to the PR description.

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,369 @@
...
+import * as semver from "semver";
+
+/**
+ * Schema for a compatibility checkpoint.
+ *
+ * @internal
+ */
+export interface Checkpoint {
+	readonly name: string;
+	readonly index: number;
```
agarwal-navin, line 38: Can you add a comment explaining what index is. Also, may as well add a comment for name.

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,369 @@
...
+ * and CompatibilityCheckpoints.md.
+ */
+
+import { existsSync } from "node:fs";
+import * as path from "node:path";
+
+import * as semver from "semver";
+
+/**
+ * Schema for a compatibility checkpoint.
```
agarwal-navin: Schema makes it sound like this is more than just an interfact / type - like it is something that cannot be changed without breaking document format or something.
Can you add more details about this inteface like what it's used for, etc.

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,369 @@
...
+ *
+ * @internal
+ */
+export interface Checkpoint {
+	readonly name: string;
+	readonly index: number;
+	/**
+	 * Lower bound of the CC range.
+	 */
+	readonly openingVersion: string;
```
agarwal-navin: `openingVersion` doesn't seem to convey what this about. The comment calls this lower bound so maybe name it that - `lowerBoundVersion`?

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,369 @@
...
+	/** ISO date (YYYY-MM-DD). */
+	readonly earliestDate: string;
```
agarwal-navin: What does earliest date signify? Is it the date when `openingVersion` was released? Or is it the earliest date for this particular checkpoint and opening version is >= that date?
Can you add more details to the comment.
Also, if my understanding of `earliestDate` is correct, maybe something like `startDate` might be better as it signifies what is the start date for this checkpoint. earliest make it sound like any date after that might be fine (or something like that - basically not very clear).

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,369 @@
...
+	/**
+	 * Extra semver range fragments outside the `>=opening <next.opening`
```
agarwal-navin: Not sure if I understand trhis. Is it just to accomodate versions that are not the current version format - <major.minor.patch>?

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,369 @@
...
+
+/**
+ * Width of the cross-client compatibility window in either direction. Two
+ * clients are guaranteed compatible iff their checkpoints are within
+ * `windowRadius` indexes of each other (currently 3, ~18 months given
+ * the 6-month cadence).
+ *
+ * @internal
+ */
+export const windowRadius = 3;
```
agarwal-navin: This is basically the current policy right? I would suggest naming this to something of that effect. This is maybe the most important thing to track and it should be more apparent.
Maybe `fullCompatibilityWindow`?

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,369 @@
...
+	// A blank line before the END marker keeps output stable under Prettier.
+	return `${before}\n${renderCheckpointsTable()}\n\n${after}`;
+}
+
+/**
+ * Walks up from `startDir` to locate the repository root (first ancestor directory
+ * containing a `.git` entry). Throws if none is found.
+ * @internal
+ */
+export function findRepoRoot(startDir: string): string {
```
agarwal-navin: This, and a few other functions are used only in tests. I would move them out of this file and keep this concise and only production code.

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,369 @@
...
+	 * last checkpoint overall.
+	 */
+	readonly closingVersion?: string;
+}
+
+/**
+ * Future / TBD checkpoints for documentation purposes only. Move an entry to
+ * {@link checkpoints} once it is officially designated.
+ */
+const futureCheckpoints: readonly DocumentedCheckpoint[] = [
```
agarwal-navin: I would keep this close the to the current checkpoints const.

agarwal-navin: I see that this is just for documentation. Not sure I understand why this is needed. Can you elaborate (and also update the comment with the description) - same for the interface above this.

packages/test/test-version-utils/src/checkpoints.ts
agarwal-navin: nvm, ignore this - I missed that it used in non-test code.

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,369 @@
...
+ */
+export const windowRadius = 3;
+
+/**
+ * Returns the highest checkpoint whose `openingVersion` is at or below
+ * `version`.
+ *
+ * @internal
+ */
+export function getCurrentCheckpoint(version: string): Checkpoint {
```
agarwal-navin, line 155: Can you ensure there are tests for all these free functions so they don't regress. It's a little hard to review them to make sure they do what they are supposed to do.

packages/test/test-version-utils/src/checkpoints.ts
agarwal-navin: Shouldn't there be a utility for this already? build-tools may have one - we may not be able to use that one but check if we can use something existing rather than manual.

CrossClientCompatibilityDevGuide.md
scottn12: Yes, removed in latest

packages/test/test-version-utils/scripts/updateCompatVersions.ts
scottn12: Yes, `oldestCompatibleVersion` is back-compat. `seenVersions` is used to dedupe for the different compat types (checkpoints, layer-compat N-1, back-compat) so each version only gets installed once. Added a comment in latest

packages/test/test-version-utils/scripts/updateCompatVersions.ts
scottn12: Yes - N-2 was for the old cross-client version generation. We now get its versions from the checkpoints block above, so the explicit nMinus2 is no longer needed. Updated the inline comment to call this out.

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Added in latest

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Added comments for each of these in latest

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Updated in latest and renamed to `startDate`

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Yes pretty much, updated in latest to clarify

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Updated comment and name to `fullCompatibilityWindowSize` in latest

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Since we now generate the checkpoints table from this code instead of manually updating I thought it would be nice to be able to add "future" rows to project what future checkpoints may look like (as we currently do). I updated the comments in latest to clarify

packages/test/test-version-utils/package.json
scottn12: Good call, I consolidated them into `update-compat-versions` in latest.

packages/test/test-version-utils/README.md
scottn12: Now that it's consolidated into `update-compat-versions` it will be ran during the release process

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Added additional tests in latest

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Updated to `lowerBoundVersion` in latest

packages/test/test-version-utils/scripts/updateCompatVersions.ts
scottn12: Yes, layer compat should be unchanged from these changes and I manually checked for regression. Updated the PR description with relevant info

packages/test/test-version-utils/src/checkpoints.ts
scottn12: I replaced `findRepoRoot` with an existing tool from `@fluid-tools/build-infrastructure` in latest. It does add a new dependency, but it's only a test package so I don't think we care too much about bundle size

agarwal-navin (approved):

packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -6,24 +6,31 @@
 /**
  * Update script for the compat test workspaces.
  *
- * Run this script after a Fluid Framework version bump as part of the release process:
+ * Run this script after:
+ * - A Fluid Framework version bump as part of the release process
+ * - A new compatibility checkpoint is designated
```
jason-ha, line 11: +Any time checkpoints.ts is updated, right?

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,371 @@
...
+		index: 6,
+		lowerBoundVersion: "4.0.0",
+		startDate: "2027-01-06",
+		status: "tbd",
+	},
+	{
+		name: "CC-7",
+		index: 7,
+		lowerBoundVersion: "5.0.0",
+		startDate: "2027-07-06",
```
jason-ha, line 133: Dates are off. (have been off) Easy to adjust though. Love the centralization :)

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -0,0 +1,371 @@
...
+
+/**
+ * Size of the cross-client compatibility window in both directions (older and newer).
+ * Two clients are guaranteed compatible if their checkpoints are within
+ * `fullCompatibilityWindowSize` indexes of each other (currently 3, ~18 months given
+ * the 6-month cadence).
+ *
+ * @internal
+ */
+export const fullCompatibilityWindowSize = 3;
```
jason-ha, line 147: This is not what I understand the policy to be. It may so happen that this works for now, but I thought the policy was explicitly time-based. The CC coverage range depends on time.
Ideally this is a time to match the policy and the settings allow for manual adjustment.

CrossClientCompatibilityDevGuide.md
```diff
@@ -197,7 +197,12 @@ Designation is the act of officially marking a Fluid Framework release as a new
 
 A new checkpoint should be designated no less than 6 months after the previous one. It should also land on a new major or beta boundary (e.g., `3.0.0`, `2.100.0`), so the prior checkpoint's range can extend cleanly to the new boundary.
 
-**To designate a new checkpoint:** update the [Compatibility Checkpoints](./CompatibilityCheckpoints.md) document and include a changeset noting the new boundary so it appears in the release notes.
+**To designate a new checkpoint:**
+
+1. Add the new checkpoint to `checkpoints` in [`packages/test/test-version-utils/src/checkpoints.ts`](./packages/test/test-version-utils/src/checkpoints.ts), and remove the corresponding future (TBD) estimate from the `futureCheckpoints` array in the same file.
+2. From `packages/test/test-version-utils`, run `pnpm run update-compat-versions` to refresh the per-version compat workspaces and regenerate the table in [`CompatibilityCheckpoints.md`](./CompatibilityCheckpoints.md). Do **not** edit that table by hand.
+3. From `packages/test/test-version-utils`, run `pnpm run update-compat-versions` to update the installed versions used by e2e tests.
```
jason-ha, line 204: There are two entries that say to run `pnpm run update-compat-versions`. Can be made one, right?

CrossClientCompatibilityDevGuide.md
```diff
@@ -197,7 +197,12 @@ Designation is the act of officially marking a Fluid Framework release as a new
 
 A new checkpoint should be designated no less than 6 months after the previous one. It should also land on a new major or beta boundary (e.g., `3.0.0`, `2.100.0`), so the prior checkpoint's range can extend cleanly to the new boundary.
 
-**To designate a new checkpoint:** update the [Compatibility Checkpoints](./CompatibilityCheckpoints.md) document and include a changeset noting the new boundary so it appears in the release notes.
+**To designate a new checkpoint:**
+
+1. Add the new checkpoint to `checkpoints` in [`packages/test/test-version-utils/src/checkpoints.ts`](./packages/test/test-version-utils/src/checkpoints.ts), and remove the corresponding future (TBD) estimate from the `futureCheckpoints` array in the same file.
+2. From `packages/test/test-version-utils`, run `pnpm run update-compat-versions` to refresh the per-version compat workspaces and regenerate the table in [`CompatibilityCheckpoints.md`](./CompatibilityCheckpoints.md). Do **not** edit that table by hand.
```
jason-ha, line 203: Recommend saying "Run \`pnpm -r --filter @fluid-private/test-version-utils update-compat-versions\`" so that it can be run from any directory within client workspace (single command versus multiple including directory change).

packages/test/test-version-utils/compat-workspaces/generated-versions.cjs
```diff
@@ -15,10 +15,12 @@ module.exports = {
 		"2.63.0",
 		"2.53.1",
 		"2.43.0",
+		"2.40.0",
 		"2.33.2",
 		"2.23.0",
 		"2.13.0",
 		"2.5.0",
+		"2.0.9",
```
jason-ha, line 23: With even more versions in here, does testing distinguish need to check cross-client versus cross-layer?
Or are the combos exploding?

packages/test/test-version-utils/src/compatConfig.ts
jason-ha, line 341: This doesn't seem like it adds the current version against the lower-bound of the current CC. That definitely should be in the mix. (With exception for self-test.)

jason-ha (commented):
Finally got to review :D and have some questions.
Also, I am wondering if we should use "CC#" instead of "CC-#" to avoid confusion with "N-#" terms. Could also use alternate separator like "CC:#" or `#` ("CC#n").

packages/test/test-version-utils/scripts/updateCompatVersions.ts
scottn12: Yes it should re-run for every entry in `checkpoints.ts`

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Yes, `fullCompatibilityWindowSize` is just a proxy for test purposes, but maybe it would make more sense to just calculate the compatibility at runtime based on the dates.

packages/test/test-version-utils/compat-workspaces/generated-versions.cjs
scottn12: I don't think it would hurt to distinguish, but IMO we don't need to since the number of versions should be stabilizing.

packages/test/test-version-utils/src/compatConfig.ts
scottn12: Yes we missed that, should be added in a follow-up

packages/test/test-version-utils/compat-workspaces/generated-versions.cjs
jason-ha: Here is a handful of questions - if this is unsettled, then maybe we discuss in Teams.
1. Why do we have all of these particular versions?
2. I think you might be saying that from cross-client we'd just have a few. But what does cross-layer compat require for good coverage? Same?
   - If same (I am expecting that isn't the case), then why couldn't we reduce the set now?
   - If different, then how do we know when any particular version can go away?
3. Is there an algorithm to determine what cross-layer compat requires?

packages/test/test-version-utils/src/compatConfig.ts
jason-ha: At this point is further work considered handed over? Or do you still have time to follow up on some items in here?
Edit: I now see there is a follow-up PR

scottn12: I would say further work is mostly handed off but I'm happy to help! Especially with follow-ups to work I've recently done

packages/test/test-version-utils/scripts/updateCompatVersions.ts
jason-ha: Resolved in #27547

CrossClientCompatibilityDevGuide.md
jason-ha: Resolved in #27547

packages/test/test-version-utils/src/compatConfig.ts
jason-ha: Resolved in #27547

packages/test/test-version-utils/compat-workspaces/generated-versions.cjs
scottn12: Feel free to open a thread in Teams if any of these answers aren't satisfactory

> Why do we have all of these particular versions?

These are the versions that resolve from our in-window cross-client and layer-compat scenarios, plus the full back-compat versions and a small hardcoded list. Not sure if the hardcoded list is still needed.

> I think you might be saying that from cross-client we'd just have a few. But what does cross-layer compat require for good coverage? Same?

I believe cross-layer is just N (current version), N-1 (last major version), and OCV (oldest compatible version). Not sure if this needs to be updated as well.

The relative versions (N-1 and the in-window checkpoints) get recomputed whenever we run update-compat-versions, so stale ones are replaced as the current version moves forward. OCV and the explicit list are pinned until we change them manually.

> Is there an algorithm to determine what cross-layer compat requires?

I think it's just hardcoded to N, N-1, and OCV.

PR #27238, opened 2026-05-06
feat(tree): Allow event deferral during transactions

Josmithr:
Adds a `deferEvents` flag to `runTransaction`, which allows users to opt-into deffering events until after the transaction has completed.


packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -316,27 +334,36 @@ let bufferTreeEvents: boolean = false;
  * @remarks
  * Note: this should be used with caution. User application behaviors are implicitly coupled to event timing.
  * Disrupting this timing can lead to unexpected behavior.
+ * @param callback - Function to invoke while events are buffered.
+ * @param options - Optional configuration. See {@link WithBufferedTreeEventsOptions}.
  */
-export function withBufferedTreeEvents(callback: () => void): void {
+export function withBufferedTreeEvents<TResult>(
```
noencke: I think it might be cleaner to have the primary callback passed to `withBufferedTreeEvents` return true/false to determine discard behavior, rather than having this additional parameter. It would make sense since the whole point of this method is to buffer events, and the return value of the main callback would be controlling whether that happens or not.

The downside is that then you couldn't return the return value of the transaction directly, so you might have to add a few extra lines at the call site to capture the return value within the callback and have it outside. But the call site is already split up anyway to accommodate the current scheme:

```
const result = callback();
discard = options?.shouldDiscard?.(result) === true;
return result;
```

It's not like it's a super clean one liner, so I don't think it'd make it too much messier. And then this whole interface would get much simpler and self-describing. What do you think?

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -542,6 +582,22 @@ class KernelEventBuffer implements Listenable<KernelEvents> {
 		}
 	}
```
noencke, line 645: Can this method delegate to `discard()` now?

It clears childrenChangeBuffer, fieldMarksBuffer, and invalidatedFieldMarkKeys - sometimes.
It also clears subTreeChangedBuffer - sometimes.

But it appears to me that the "sometimes" is purely an optimization - i.e., the only time it _doesn't_ clear them is because they are already clear. If that is true, then I think the cleanliness of removing all those clears and simply calling `discard()` at the end of `flush()` is worth it. I imagine that `clear()` on an empty map/set in JS has negligible perf cost.

packages/framework/fluid-framework/api-report/fluid-framework.alpha.api.md
```diff
@@ -1485,12 +1485,21 @@ export interface RunTransaction {
     readonly rollback: typeof rollback;
 }
 
+// @alpha @input
+export interface RunTransactionAsyncParams extends RunTransactionParams {
```
noencke: IMO, no reason to add this until we need it. `runTransactionAsync` can just take the more general `RunTransactionParams` for now. We can always add a type like this later if necessary.

Josmithr: That was how I originally had it, but it seemed a bit weird to me for `runTransaction` to take `RunTransactionSyncParams`, while `runTransactionAsync` takes just `RunTransactionParams`.

But maybe it would make more sense to just do some type renames (these are alpha APIs after all).

Let me know what you think. I don't have particularly strong opinions here 🤷‍♂️

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
Josmithr: Let me play around with it and see how it works out. I think I like it.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
Josmithr: Yeah, that seems reasonable. Will update.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -316,27 +334,36 @@ let bufferTreeEvents: boolean = false;
...
-	} else {
-		bufferTreeEvents = true;
-		try {
-			callback();
-		} finally {
-			bufferTreeEvents = false;
-			flushEventsEmitter.emit("flush");
-		}
+		// Already buffering - just run the callback. The outermost call owns the flush/discard decision.
+		return callback();
```
Josmithr: Shame on me for not reviewing AI changes thoroughly enough. This isn't okay, and complicates the ability to let the user decide whether or not to emit or discard the buffered events. I'll explore this a bit and determine whether or not allowing event discard is worth the complexity (for now).

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
Josmithr: Updated :)

packages/framework/fluid-framework/api-report/fluid-framework.alpha.api.md
noencke: Since it's an input type, I think customers pay more attention to the "shape" that shows up in auto-complete than they do the name of the type. I don't think the lack of "Async" is that weird. Also don't feel very strongly about it either, I'd be fine with either. But in a vacuum, less API churn and less code probably tips the scales.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -351,6 +388,34 @@ class KernelEventBuffer implements Listenable<KernelEvents> {
...
+		this.popFrame.bind(this),
+	);
+
+	/**
+	 * Stack of saved buffer states, one entry per active nested {@link withBufferedTreeEvents}
+	 * call that this buffer was alive for. The current (live) state is always held in the
+	 * regular buffer fields; this stack holds previously-saved states that will be restored
+	 * (or merged into) when the corresponding nested call ends.
+	 */
+	readonly #frameStack: BufferFrame[] = [];
```
noencke: I think it would (probably) be more elegant if we used the first element in this stack to be the current (live) state, rather than having a top-level copy of each of the four buffer structures. It would make disposal easier too - you could just drop the buffer frame rather than clearing the buffers.

The upside to the way it's currently done is that we're not re-allocating the three buffer structures for top-level edits (though we still are for nested transactions). That may or may not be good. Is it actually a notice-able performance benefit? If it is, we should probably mark it as such - "We retain the top buffer frame outside of the stack as a performance optimization to reduce set/map allocation in the top-level edit (common) case".

In _either_ case - can we at least re-use the BufferFrame type for the top-level frame?

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -526,20 +591,134 @@ class KernelEventBuffer implements Listenable<KernelEvents> {
...
+
+	/**
+	 * Discards any events buffered due to {@link withBufferedTreeEvents} without emitting them.
+	 *
+	 * @remarks
+	 * Used by transaction code paths that know the tree is in the same state it started in
+	 * (e.g. a rolled-back synchronous transaction) so the buffered events represent net-zero
+	 * changes that should not be observed by listeners.
+	 */
+	public clearBuffers(): void {
```
noencke: This doesn't clear buffers in the stack - should it? Either way, it should document explicitly whether it does or doesn't.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -526,20 +591,134 @@ class KernelEventBuffer implements Listenable<KernelEvents> {
...
+	 * If the frame stack is empty, this buffer joined the scope after the surrounding call's
+	 * {@link KernelEventBuffer.pushFrame | pushFrame} (i.e., it was constructed mid-scope), so
+	 * its current state already represents only events from the popping scope. In that case
+	 * `discard` clears the current buffers and `!discard` keeps them so they bubble up.
+	 */
+	public popFrame(discard: boolean): void {
+		this.#assertNotDisposed();
+		const prev = this.#frameStack.pop();
+		if (prev === undefined) {
+			// Late-joining buffer: all of current belongs to the popping scope.
```
noencke: Er, wait - when does this happen? Can this actually happen? The doc comment on this method calls it out but I still don't understand - maybe it could use an actual example of how we would get into this state ("start a transaction, then do a thing, then start another transaction, then end the transaction" - whatever the repro is).

packages/framework/fluid-framework/api-report/fluid-framework.alpha.api.md
Josmithr: That's likely true for the intellisense experience, but I think much less true for the published API docs experience.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
Josmithr: Done in latest.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
Josmithr: Yeah, this was confusing. I've refactored things pretty significantly now - this is now effectively moot.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
Josmithr: I've refactored things significantly. This should be a lot more clear now via both the code and the docs.

packages/framework/fluid-framework/api-report/fluid-framework.alpha.api.md
Josmithr: But I don't think the original type names are likely to cause any active confusion. Reverted.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -513,32 +623,61 @@ class KernelEventBuffer implements Listenable<KernelEvents> {
...
+			this.#current.subTreeChangedBuffer = true;
 		} else {
 			this.#events.emit("subtreeChangedAfterBatch");
 		}
 	}
 
 	/**
-	 * Flushes any events buffered due to {@link withBufferedTreeEvents}.
+	 * Begin a buffering scope. A fresh empty frame is pushed onto the stack and becomes the live
+	 * (current) frame; any outer scope's frame remains beneath, untouched.
```
```suggestion
	 * (current) frame; any outer scope's frame remains beneath, untouched until this frame is popped.
```

noencke (approved):

PR #27240, opened 2026-05-06
feat: add legacy beta map compatibility interfaces

jenn-le:
## Description

Adds legacy beta map compatibility interfaces for APIs that need to align with JavaScript `Map` while still using Fluid's map abstraction.

This change introduces `FluidMapLegacy` in `@fluidframework/core-interfaces/legacy`. It extends `FluidMap` and only adds the legacy/built-in `Map`-style members needed for compatibility: `clear`, `delete` returning `boolean`, `set` returning `this`, and a `forEach` callback typed with `FluidMapLegacy`.

This also adds `IDirectoryBeta` and `ISharedMapBeta` in `@fluidframework/map/legacy`. The existing `IDirectory` and `ISharedMap` interfaces are unchanged and continue to extend the built-in `Map`; the new beta variants provide an opt-in shape that consumes `FluidMapLegacy`.


packages/dds/map/src/interfaces.ts
```diff
@@ -9,6 +9,8 @@ import type {
 	IEventProvider,
 	IEventThisPlaceHolder,
 } from "@fluidframework/core-interfaces";
+// eslint-disable-next-line import-x/no-internal-modules
```
jenn-le: Is there a proper way to use beta APIs here?

CraigMacomber: Import from internal

packages/dds/map/src/interfaces.ts
```diff
@@ -119,6 +121,36 @@ export interface IDirectory
...
+/**
+ * Beta version of {@link IDirectory} which uses {@link @fluidframework/core-interfaces#FluidMapLegacy} for its map-like API.
+ *
+ * @sealed
+ * @legacy @beta
+ */
+export interface IDirectoryBeta
+	// TODO: Use `unknown` instead (breaking change).
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	extends Omit<IDirectory, keyof Map<string, any>>,
```
CraigMacomber: I think it would be cleaner to exclude "get" and "set" from the keys of map you are filtering out rather than restating them below.

packages/dds/map/src/test/types/fluidMapLegacyTypes.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import type { FluidMapLegacy } from "@fluidframework/core-interfaces/legacy";
+
+import type { IDirectory, IDirectoryBeta, ISharedMap, ISharedMapBeta } from "../../index.js";
+
+type requireTrue<_X extends true> = true;
+type isAssignableTo<Source, Destination> = [Source] extends [Destination] ? true : false;
```
CraigMacomber: For the use cases in this file, I think requireAssignableTo (which you can import from build-tools) is a better fit. Does basically the same thing in one step, and gives better error messages when it fails than this approach.

packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -149,3 +149,44 @@ export interface FluidMap<K, V> extends FluidReadonlyMap<K, V> {
...
+/**
+ * Like TypeScript's built in `Map` type, while still extending {@link FluidMap}.
+ *
+ * @privateRemarks
+ * This interface exists for legacy APIs which were already exposed as built-in `Map`-like types.
+ * New APIs should prefer {@link FluidMap} unless they need to preserve compatibility with an existing `Map` contract.
+ *
+ * @sealed @legacy @beta
+ */
+export interface FluidMapLegacy<K, V> extends FluidMap<K, V> {
```
CraigMacomber: Do we plan to ever use this for anything other than shared map and shared directory? If not, it should probably be in that package.

CraigMacomber: If you do move this, you should be able to inline the questionable get and set APIs those maps use here, which would simplify things a bit.

packages/common/core-interfaces/src/fluidMap.ts
```diff
@@ -149,3 +149,44 @@ export interface FluidMap<K, V> extends FluidReadonlyMap<K, V> {
...
+	 */
+	delete(key: K): boolean;
+
+	/**
+	 * Adds a new element with a specified key and value to the map.
+	 * If an element with the same key already exists, the element will be updated.
+	 *
+	 * @returns The map itself.
+	 */
+	set(key: K, value: V): this;
```
CraigMacomber: Don't we override/omit this in all cases that use this interface? Maybe we don't need to state it here?

packages/dds/map/src/interfaces.ts
```diff
@@ -378,9 +430,11 @@ export interface ISharedMapEvents extends ISharedObjectEvents {
  * @sealed
  * @legacy @beta
  */
-// TODO: Use `unknown` instead (breaking change).
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export interface ISharedMap extends ISharedObject<ISharedMapEvents>, Map<string, any> {
+export interface ISharedMap
+	// TODO: Use `unknown` instead (breaking change).
```
Josmithr, line 434: Nit: this comment was for the eslint-disable comment that got shifted down below the extends clause

packages/dds/map/src/interfaces.ts
```diff
@@ -119,6 +120,57 @@ export interface IDirectory
...
+/**
+ * Beta version of {@link IDirectory} which uses {@link FluidMapLegacy} for its map-like API.
+ *
+ * @sealed
+ * @legacy @beta
+ */
+export interface IDirectoryBeta
+	extends Omit<IDirectory, Exclude<keyof Map<string, unknown>, "get" | "set">>,
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		FluidMapLegacy<string, any> {}
```
Josmithr, line 172: Not for this PR, but I wonder if we could reasonably get away with using `unknown` here instead of `any`?

Josmithr (approved):

packages/dds/map/src/interfaces.ts
```diff
@@ -119,6 +120,57 @@ export interface IDirectory
...
+
+	/**
+	 * Executes the provided function once per each key/value pair in the map.
+	 */
+	forEach(
+		callbackfn: (value: V, key: K, map: Map<K, V>) => void,
+		// Typing inherited from Map.
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		thisArg?: any,
+	): void;
```
jason-ha, line 153: @CraigMacomber & @Josmithr (Jenn is out), this doesn't seem to all add up to me. The caller can give a callback that expects FluidMap or Map. I would guess that all of the implementations then must return Map_s that are a superset of FluidMap. But we don't implement `FluidMapLegacy` anywhere. We tell customers that they can cast. That would mean FluidMap (should satisfy Map) to be a valid cast. But we know that isn't the case. So, I suspect this is a case where TypeScript is unsound.
Do we need these both? If so, can we make things sound or what am I missing?

jason-ha: And these forEach_es give access to non-legacy *Map_s where as the main interface does not. Because the implementations give access to cloned maps?

PR #27239, opened 2026-05-06
chore(bundle-size-tools): remove dead code orphaned by the analyzer.json switch

ChumpChief:
[AB#73000](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73000)

## Description

Companion cleanup to #27224. After bundle-size-tools' consumer switched to reading `analyzer.json`, the entire `WebpackStatsProcessor` pipeline (gunzip → walk → entry processor → total processor) and the bundle-buddy plumbing have zero remaining consumers. Delete it.

This is purely dead-code removal — no behavior changes, no renames, no restructures. Final shape: +8 / -1081 across 26 files.

**Files deleted (13):**
- `BundleBuddyConfigWebpackPlugin.ts` (was always orphaned — no producer used it)
- `ADO/DefaultStatsProcessors.ts`, `ADO/PrCommentsUtils.ts` (the latter has been orphaned since PR #27158 removed the danger flow), `ADO/getBundleBuddyConfigMap.ts`
- `statsProcessors/{bundleBuddyConfigProcessor,entryStatsProcessor,totalSizeStatsProcessor,index}.ts`
- `utilities/{decompressStatsFile,getChunkAndDependenciesSizes,getChunkParsedSize,getLastCommitHashFromPR,runProcessorOnStatsFile}.ts`

**Symbols removed from existing files:**
- `getStatsFileFromZip`, `getStatsFileFromFileSystem` (msp.gz readers)
- `getBundleBuddyConfigFileFromZip`, `getBundleBuddyConfigFromFileSystem`
- `getBundlePathsFromZipObject`, `getBundlePathsFromFileSystem`, `getBundleFilePathsFromFolder` (msp.gz path discovery)
- Original `getBundleSummaries` + `GetBundleSummariesArgs` (the stats-processor-driven entry point; `getBundleSummariesFromAnalyzer` remains as the live one)
- `totalSizeMetricName` (the synthetic "Total Size" metric is no longer produced; build-cli's `bundleSizeDiff` regression filter that referenced it is updated to drop the now-dead branch)
- Types: `BundleBuddyConfig`, `ChunkToAnalyze`, `WebpackStatsProcessor`

**Runtime deps no longer reachable**, dropped from `bundle-size-tools/package.json`: `fflate`, `msgpack-lite`, `@types/msgpack-lite`.

**Out of scope** (separate follow-up PRs):
[...]

PR #27224, opened 2026-05-04
feat(bundle-size-tools): emit and consume analyzer.json for bundle size diff

ChumpChief:
[AB#72999](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/72999)

## Description

webpack-bundle-analyzer's `analyzerMode: "json"` output (`analyzer.json`) is a much smaller, pre-summarized form of the per-asset data we use for bundle-size comparisons (~750KB vs ~3.7MB for `bundleStats.msp.gz`, the compressed raw stats we currently consume — and ~377MB raw). We use roughly 1% of what's in the full webpack stats. Switching to `analyzer.json` lets us drop the entire `WebpackStatsProcessor` pipeline (gunzip → msgpack-decode → walk → entry processor → total processor) and just read pre-summarized per-asset `parsedSize` directly.

Two commits, separable for review:

1. **Producer** (`examples/utils/bundle-size-tests/webpack.config.cjs`): adds a second `BundleAnalyzerPlugin` instance with `analyzerMode: "json"`, emitting `bundleAnalysis/analyzer.json` alongside the existing `report.html`/`report.json`/`bundleStats.msp.gz`. Existing outputs are unchanged — `flub generate bundleStats` sanity checks and the FF-internal telemetry handler keep working as-is.
2. **Consumer** (`bundle-size-tools`): adds `analyzer.json` reader functions (`getAnalyzerJsonFromZip`, `getAnalyzerJsonFromFileSystem`), path discovery (`getAnalyzerFilePathsFromFolder` and the `getAnalyzerPathsFrom*` wrappers), and `getBundleSummariesFromAnalyzer`. `ADOSizeComparator` switches to the analyzer.json path internally. `result.json` shape and `BundleMetric` (parsedSize-only) are unchanged, and per-entrypoint metric names/values match between the two paths. One delta: the msp.gz path's synthesized per-bundle "Total Size" metric is no longer emitted (it's derivable from the per-asset metrics if a downstream consumer needs it).

The new readers use `BundleAnalyzerPlugin.JsonReport` from `@types/webpack-bundle-analyzer` (added as a devDep) so the type comes from the producer authoritatively rather than a hand-rolled subset.

[...]

PR #27158, opened 2026-04-24
feat: re-enable PR bundle size comparison (ADO side, 1 of 3)

ChumpChief:
[AB#72998](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/72998)

## Description

Stands up the ADO half of the PR bundle size comparison flow that's been disabled since the backing GitHub PAT expired ([AB#56981](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/56981)). This PR produces the comparison data; a follow-up adds the consumer that posts it to PRs.

**What lands here**

- `ADOSizeComparator.getSizeComparison()` — discriminated-union return (`{ kind: "success" | "error" }`) of the existing baseline-discovery + compare logic, decoupled from any rendering.
- `flub generate bundleSizeDiff` — new command that calls the above and writes `artifacts/bundleSizeDiff/result.json`.
- New PR-only `bundleSizeDiff` pipeline artifact published from `build-npm-client-package.yml`, alongside the existing non-PR `bundleAnalysis` artifact.
- Full retirement of the Danger flow: `dangerfile.cts`, `flub publish bundleStats` / `flub run bundleStats`, the HTML renderer (`getCommentForBundleDiff` / `getSimpleComment`), the `(DISABLED)` pipeline step in `build-npm-client-package.yml`, the dead-code sibling step in `build-npm-package.yml`, the `bundle-analysis:run` root npm script, the stale `.vscode/launch.json` entry, and the `danger` dev-dep from both root and `build-cli` package.json files. Resolves [AB#3710](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/3710).

**What doesn't land here**

Because no consumer exists yet for the new artifact, PR builds won't post bundle-size comments or apply the `size regression` label. Artifacts accumulate unused until the consumer ships. No user-visible change on PRs at merge time.

## Staging plan (3 PRs)

This is the first of three related PRs:

1. **This PR** — ADO-side machinery + artifact producer, Danger retirement.
2. **Next PR** — iterate on \`result.json\`'s shape (e.g. drop \`Total Size\`, emit per-metric \`diffBytes\` / \`diffPercent\`, add provenance) and simplify \`@fluidframework/bundle-size-tools\`.
3. **Third PR** — GitHub Actions workflow that consumes the artifact to post the PR comment and reconcile the \`size regression\` label.

[...]

PR #27242, opened 2026-05-06
chore(bundle-size-tools): drop unused/flawed legacy behaviors

ChumpChief:
[AB#73001](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73001)

## Description

Follow-up to #27224 (analyzer.json swap) and #27239 (dead-code cleanup). Part of [AB#56981](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/56981) (re-enable PR bundle size comparison).

Three pre-existing behaviors are removed because they don't fit the `bundleSizeDiff` design that #27158 began (a producer that emits a structured artifact, consumed by a future GitHub Actions workflow). Each is independent; commits are split for review.

### 1. `sizeRegressionDetected` field on `bundleSizeDiff` output (commit 1)

**Background.** `flub generate bundleSizeDiff` runs a producer-side regression check: any non-total metric growing by more than a hardcoded 5KB threshold sets a `sizeRegressionDetected: boolean` field on `result.json`'s `"changes"` variant.

**Why remove.** That threshold is a policy decision, not raw data. The broader design has the producer emit unopinionated per-bundle deltas and the consumer (the future GH Actions workflow that will post PR comments / manage labels) apply its own threshold. Producer-side hardcoding pre-commits us to one definition of "regression" that the consumer can't override.

**Removed.**
- `detectSizeRegression` function
- `sizeRegressionThresholdBytes` constant
- `sizeRegressionDetected` field from `BundleSizeDiffResult`'s `"changes"` variant
- `BundleMetric` type import (only used by the removed function)

**Compatibility.** No consumer reads `result.json` yet — the GH workflow that will is a future PR — so this is a no-op at the seam and lets that future consumer pick its own threshold.

### 2. `tagWaiting` mechanism in `ADOSizeComparator` (commit 2)

**Background.** When the comparator runs against a baseline commit whose CI build hasn't completed yet, it tags the PR's own build with a marker (via `tagBuildAsWaiting`) so that a downstream process could re-run the comparison once the baseline finished.

**Why remove.** The downstream half (consuming the tag and retriggering the PR build) was never written, so the tagging is effectively no-op overhead today. The new `bundleSizeDiff` design — and PR 3's planned scheduled re-run — replaces this need entirely with workflow-side polling.

**Removed.**
- `bundle-size-tools/src/ADO/getBuildTagForCommit.ts` (whole file)
- `ADOSizeComparator.tagBuildAsWaiting` private method
- `ADOSizeComparator` constructor param `adoBuildId`
- `getSizeComparison`'s `tagWaiting` parameter
- `IADOConstants` fields `prBuildDefinitionId` and `projectRepoGuid` (only used by tagWaiting and the now-gone PR-comments machinery)
- Corresponding barrel exports

### 3. Naive fallback in `ADOSizeComparator` (commit 2)

**Background.** When no CI build is found for the baseline commit (or the build has no usable artifact), the comparator walks prior commits via `getPriorCommit` looking for one with a usable build, and compares the PR against that.

**Why remove.** The resulting comparison is **incorrect-by-design**: the PR is being compared against a commit on `main` that's N commits older than the merge-base, so main-side deltas between merge-base and merge-base−N get falsely attributed to the PR.

**Removed.**
- `ADOSizeComparator.naiveFallbackCommitGenerator` static method
- `ADOSizeComparator` constructor param `getFallbackCommit`
- `utilities/gitCommands.ts`'s `getPriorCommit` (only used by the fallback)
- Corresponding barrel exports

**Cleanup side-effect.** With both `tagWaiting` and the fallback gone, `getSizeComparison` flattens from a while-loop with continue/fallback branches into a linear sequence of early-return validations, and error messages become more specific — `"No CI build found for baseline commit X"` / `"Baseline build did not publish bundle artifacts"` instead of the generic `"Could not find a usable baseline build"`.

### 4. `BannedModulesPlugin` (commit 3)

**Background.** A custom Webpack plugin shipped by `bundle-size-tools` and used by `bundle-size-tests` to fail the build if any chunk imports a banned module. The only banned module configured today is Node-core `assert`, whose polyfill is large in browser builds.

**Why remove.** Webpack 5 supports the same effect natively via `resolve.fallback: { assert: false }`, which blocks the polyfill from being bundled at all. The plugin existed because earlier Webpack versions didn't have this option. The native config is a few lines; the plugin is a ~70-line file plus the burden of being a public API export of `bundle-size-tools`.

**Removed.**
- `bundle-size-tools/src/bannedModulesPlugin/bannedModulesPlugin.ts` (whole file)
- `BannedModule`, `BannedModulesPlugin`, `BannedModulesPluginOptions` exports

**Dependency drops** (consequences of the swap):
- `webpack` removed from `bundle-size-tools/package.json` (was only used by the plugin)
- `@fluidframework/bundle-size-tools` removed from `bundle-size-tests/package.json` devDependencies (was only used here for the plugin)

## Reviewer Guidance

Three independent removals, each in its own commit — easy to review commit-by-commit.


pnpm-lock.yaml
```diff
@@ -33588,7 +33582,7 @@ snapshots:
       fast-glob: 3.3.3
       is-glob: 4.0.3
       minimatch: 9.0.9
-      semver: 7.7.3
+      semver: 7.7.4
```
TommyBrosman, line 33585: I think this is just a side-effect of the package removal, but worth verifying.

build-tools/packages/bundle-size-tools/src/ADO/AdoSizeComparator.ts
```diff
@@ -64,140 +62,76 @@ export class ADOSizeComparator {
...
-				}
-
-				// Found usable baseline zip
-				break;
+
+			const baselineBuild = recentBuilds.find(
+				(build) => build.sourceVersion === baselineCommit,
+			);
+
+			if (baselineBuild === undefined) {
```
TommyBrosman, line 94: Does this happen in practice when there isn't a CI build for a particular revision? Are all merged commits guaranteed to have CI builds under normal circumstances?

pnpm-lock.yaml
ChumpChief: Yep, I could strip it out but I checked the release notes and it's pretty benign, might as well take it I think.

build-tools/packages/bundle-size-tools/src/ADO/AdoSizeComparator.ts
ChumpChief: Generally yes, but the mainstream case where it won't is when the PR is racing the CI build.  E.g. if you merge main to your PR immediately after a commit lands on main, your PR might run this step before the respective CI build finishes.  When adding the commenting logic I'll be adding handling for that potential ordering.

TommyBrosman (approved):

build-tools/packages/bundle-size-tools/src/ADO/AdoSizeComparator.ts
ChumpChief: Other case is if the PR's merge base is really old, since we use a fixed search window IADOConstants.buildsToSearch (I'll probably expand this a bit but it will still be finite)

PR #27243, opened 2026-05-06
feat(server-services): upgrade mongodb driver from v5 to v6

zhajie:
## Description

Upgrades the `mongodb` Node.js driver in `@fluidframework/server-services`
from `^5.9.2` to `^6.21.0` and fixes all v5→v6 breaking changes in the
MongoDB wrapper layer (`src/mongodb.ts`).

### Breaking changes in the driver addressed

**1. `findOneAndUpdate` return type change**
In v6, `findOneAndUpdate` / `findOneAndReplace` / `findOneAndDelete` return
the document directly (`T | null`) instead of a `ModifyResult` wrapper.
Fixed in `MongoCollection.findOrCreate` and `MongoCollection.findAndUpdate`
by accessing the result directly instead of via `.value`.

**2. `returnOriginal` option removed**
The `returnOriginal` option was removed in v6 in favor of
`returnDocument: "before" | "after"`. Fixed in `MongoCollection.findOrCreate`
default options:
- Before: `{ returnOriginal: true, upsert: true }`
- After: `{ returnDocument: "before" as const, upsert: true }`

**3. `keepAlive` removed from `MongoClientOptions`**
The `keepAlive` boolean was removed from `MongoClientOptions` in v6
(`keepAliveInitialDelay` remains valid). Removed `keepAlive: true` from
`MongoDbFactory.connect`.

### Transitive dependency changes
- `bson`: 5.5.1 → 6.10.4
- `mongodb-connection-string-url`: 2.6.0 → 3.0.2
- `@mongodb-js/saslprep`: moved from optional to required (1.1.1 → 1.4.11)
- `socks`: moved from required to optional peer dependency
- `whatwg-url`: 11.0.0 → 14.2.0 (via mongodb-connection-string-url)

## Reviewer Guidance

All MongoDB usage is centralized in `packages/services` behind the
`ICollection`/`IDb` abstraction interfaces, so downstream packages are
unaffected by the driver upgrade. No MapReduce, `Collection.count()`, or
callback-based `find()` calls exist in the codebase. All 89 unit tests pass.


server/routerlicious/packages/services/src/mongodb.ts
```diff
@@ -762,7 +758,6 @@ export class MongoDbFactory implements core.IDbFactory {
 		// Need to cast to any before MongoClientOptions due to missing properties in d.ts
 		const options: MongoClientOptions = {
 			directConnection: this.directConnection ?? false,
-			keepAlive: true,
 			keepAliveInitialDelay: this.keepAliveInitialDelay,
```
zhangxin511: keepAliveInitialDelay can be removed as well

server/routerlicious/packages/services/src/mongodb.ts
```diff
@@ -316,9 +316,7 @@ export class MongoCollection<T extends Document> implements core.ICollection<T>,
 					options,
 				);
 
-				return result.value
-					? { value: result.value, existing: true }
-					: { value, existing: false };
+				return result ? { value: result as T, existing: true } : { value, existing: false };
```
zhangxin511: Let's do ` value: result as unknown as T,` instead

zhajie: addressed

server/routerlicious/packages/services/src/mongodb.ts
zhajie: updated

zhangxin511 (approved):

zhangxin511 (approved):

PR #27245, opened 2026-05-06
Use releaseJob template in LTS deployment pipeline

jumyhre:
This change updates LTS' deployment pipeline to use the release job template. This change is loosely based on the preexisting template from main. The main changes between this pipeline and main are: 
1) It does not include the steps to use build-tools artifacts and adjust their lockfiles, since pnpm uses a version of build-tools from main.
2) We're using node 20 on LTS instead of 22


alexvy86 (approved):
LGTM

PR #27247, opened 2026-05-06
build(eslint-config-fluid): update dependencies

tylerbutler:
Updates eslint dependencies.

The `react-hooks/component-hook-factories` rule has been removed and is thus no longer in our config.


common/build/eslint-config-fluid/package.json
```diff
@@ -97,16 +97,16 @@
 			"chokidar": "^5",
 			"diff@>=5 <6": "^5.2.2",
 			"js-yaml": "^4.1.1",
+			"minimatch@>=10 <11": "^10.2.4",
```
ChumpChief: Bad sort?

common/build/eslint-config-fluid/pnpm-workspace.yaml
```diff
@@ -16,5 +16,5 @@ trustPolicy: no-downgrade
 # List of packages known to be safe but for whatever reason were published at a date after another version of the same
 # package (including later major versions) which had better provenance information.
 # ALWAYS REVIEW CAREFULLY BEFORE ADDING SOMETHING TO THIS LIST.
-trustPolicyExclude: []
+trustPolicyExclude: ["semver"]
```
ChumpChief, line 19: I know @TommyBrosman has been looking at the trust policy stuff lately (e.g. #27254) - might want to get an opinion from someone who knows about that stuff more than I do.

ChumpChief (commented):
Just a nit but you might want a second opinion on the trust policy change.

common/build/eslint-config-fluid/pnpm-workspace.yaml
tylerbutler: Good point - I won't commit this way.

common/build/eslint-config-fluid/package.json
tylerbutler: I ran an explicit sort and it produced this - not sure why it's producing so much churn but it's passing CI.

PR #27251, opened 2026-05-06
fix(docs): boost api title matches and fix search ranking in Typesense

WayneFerrao:
## Description

The Typesense plugin had no additionalSearchParameters configured, so a mention of "tree" buried in body text ranked the same as a page titled "Tree". This caused searches like "Tree" to return 175+ poorly-ordered results, and exact type names like "TreeNodeApi" to return 0 results.

Changes:

Added query_by_weights — page title (lvl1) is weighted 6× higher than body content, descending through hierarchy levels
Added prioritize_exact_match: true — exact type name searches find the right page first
Added sort_by: _text_match:desc — strongest overall match wins
Added num_typos — typo tolerance on hierarchy levels, strict on content to reduce noise
Refactored additionalSearchParameters to derive query_by, query_by_weights, and num_typos from a single ordered array, so adding/removing a search field can't silently misalign the three lists

A separate scraper config change (typesense.config.json) will be applied directly on the VM and the index rebuilt via ./run-scraper after this deploys.

### DIFF
--- a/docs/docusaurus.config.ts
+++ b/docs/docusaurus.config.ts
@@ -189,8 +189,15 @@ const config: Config = {
 					],
 					apiKey: TYPESENSE_API_KEY,
 				},
-				// Optional
 				contextualSearch: true,
+				additionalSearchParameters: {
+					query_by:
+						"hierarchy.lvl1,hierarchy.lvl2,hierarchy.lvl3,hierarchy.lvl4,hierarchy.lvl5,content",
+					query_by_weights: "6,5,4,3,2,1",
+					sort_by: "_text_match:desc",
+					prioritize_exact_match: true,
+					num_typos: "1,1,1,1,1,0",
+				},
 			},
 		}),
 	} satisfies Preset.ThemeConfig,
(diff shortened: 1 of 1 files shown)


docs/docusaurus.config.ts
```diff
@@ -18,6 +18,19 @@ const TYPESENSE_API_KEY = process.env.TYPESENSE_API_KEY;
 
 const isTypesenseConfigured = TYPESENSE_HOST !== undefined && TYPESENSE_API_KEY !== undefined;
 
+// Each entry is [field, weight, numTypos]. Order determines priority: first = highest ranked.
+const typesenseSearchFields: [field: string, weight: number, numTypos: number][] = [
```
Josmithr, line 22: Nit: We generally try to avoid using abbreviations. Also, "numTypos" or even "numberOfTypos" doesn't quite offer a clear explanation of what it *does*. Is this a number of *allowed* typos presumably? I think we should update the variable name and add a comment explaining what it is.

Josmithr: A link to relevant TypeSense docs here would also likely be helpful.

Josmithr (approved):
Left a nitpick but otherwise looks reasonable to me. Have you tested the impact? Or can we not reasonably do that without re-indexing? How will we determine whether or not this is a sufficient fix?

PR #27258, opened 2026-05-07
fix(eslint-config-fluid): disable import-x/order rule

tylerbutler:
## Description

Disables `import-x/order` rule in preparation for #26497. Also updates the changelog and version so we can release v11.

## Reviewer Guidance

The main thing to verify is that the final override placement in `library/configs/base.mts` is late enough to win over imported presets.


common/build/eslint-config-fluid/library/configs/base.mts
```diff
@@ -127,4 +127,11 @@ export const baseConfig: FlatConfigArray = [
 			],
 		},
 	},
+	// Keep import ordering delegated to Biome's organizeImports action. This must come after
+	// shared and formatter configs so import-x presets cannot re-enable the rule.
+	{
+		rules: {
```
ChumpChief: Should we instead just remove it from common/build/eslint-config-fluid/library/rules/base.mts?  I haven't read the followup PRs yet so maybe there's a reason but it looks like it's already off-by-default in the recommended config.

Approving because it's prob fine either way but maybe a simplification opportunity?

ChumpChief (approved):

common/build/eslint-config-fluid/library/configs/base.mts
tylerbutler: Done in latest.

ChumpChief (approved):

PR #27261, opened 2026-05-07
Remove setSensitivityLabelHeaderPostFix feature gate

steffenloesch:
## Description

The `Fluid.Driver.Odsp.setSensitivityLabelHeaderPostFix` config gate is fully rolled out (value `true`) per https://ecs.skype.com/Fluid/FluidExperiences/configurations/1787074, so this PR removes the gate and the now-dead conditional code that depended on it.

Concretely:
- `OdspDelayLoadedDeltaStream.joinSessionCore` no longer reads the gate or threads its value into `fetchJoinSession`.
- `fetchJoinSession` (in `vroom.ts`) drops the `setSensitivityLabelHeader` parameter; the `Prefer: Return-Sensitivity-Labels` line is now always added to the join-session request body, matching post-rollout production behavior.
- The `setSensitivityLabelHeader` property on the `JoinSession` telemetry event is removed since it would always be `true`.
- New unit test in `vroom.spec.ts` asserts the `Prefer: Return-Sensitivity-Labels` header is present on the outbound request body. The test stubs `EpochTracker.fetchAndParseAsJSON` (the network boundary) and inspects the body it receives — so it exercises the real `fetchJoinSession` implementation rather than a mock.

`fetchJoinSession` is not part of any `*.api.md`, so the signature change is internal-only.

## Reviewer Guidance

Telemetry consumers (Kusto, dashboards) that pivot on the `setSensitivityLabelHeader` property of the `JoinSession` event will see it disappear after this lands; the value would always be `true` post-rollout, so any such consumer was effectively a no-op filter. Heads-up to telemetry owners worth a quick scan.


shlevari (approved):

jatgarg (approved):

jzaffiro (approved):
Approved for docs!

PR #27265, opened 2026-05-08
build: scoped builds for speed

jason-ha:
Add two top-level scripts to build subset of full build:
1. `build:compile:esm:packages` - generates only ESM production code
2. `build:compile:esm` - production and test ESM

## Supporting changes to complete `fluid-build --task build:esnext` without CommonJS dependencies, except for special cases.

1. Remove extraneous `^tsc` dependencies in examples.
2. Reduce general `build:esnext` to not depend on `^tsc` and `^api`.
3. Remove extraneous presence build dependencies.
4. Adjust PropertyDDS dependencies to build CommonJS where it used.
5. mocha-test-setup's CommonJS is used in ESM testing; so, build CommonJS when ESM is requested.

## Supporting general build adjustments

1. Add `^api/cjs/esm` dependencies for `build:test:cjs/esm` to account for test-only packages without production build. 
2. Refine presence `eslint` dependencies to upstream ESM build.
3. Simplify build-tools scripts to just use `tsc` since none of the packages dual-emit.
   - Remove spurious `dist` specs in build-infrastructure that aren't built.
   - Simplify entrypoint configs as there are no specialized types.

Note general `tsc` dependency is left to depend on `^api` and `^build:esnext`.


README.md
```diff
@@ -175,12 +175,19 @@ This section contains common workflows and patterns to increase inner dev loop e
...
--   `pnpm install` from the root of the repository to install dependencies. This is necessary for new clones or after pulling changes from the main branch.
--   `pnpm run build:fast` from the root of the repository to perform an incremental build that matches the CI build process. Incremental builds tend to leave extra files laying around, so running a clean is sometimes needed to cleanup ghost tests
--   `pnpm run build:fast -- <path>` to build only a specific part of the repository.
--   `pnpm run build` within a package directory to build that package.
--   `pnpm run build:compile` for cross-package compilation.
--   `pnpm run format` to format the code.
+From the root of the repository:
+
+-   `pnpm install` to install dependencies. This is necessary for new clones or after pulling changes from the main branch.
+-   `pnpm build:fast` to perform an incremental build that matches the CI build process. Incremental builds tend to leave extra files laying around, so running a clean is sometimes needed to cleanup ghost tests.
```
ChumpChief, line 181: Does `build:fast` still provide value over `build`?  I feel like it predates a lot of improvements to the standard build and I'm not sure if it still holds any advantage.  At least on my machine it isn't really faster than standard `build`, both just over 11 minutes (and `build:fast` spams logs heavily about memory pressure).

Maybe consider removing `build:fast` as a trade for adding these new commands in a followup, to keep the matrix of supported build methodologies small and simple for devs to understand the "right" way to build Fluid.

fluidBuild.config.cjs
```diff
@@ -159,8 +184,11 @@ module.exports = {
 		"format:prettier": [],
 		"prettier": [],
 		"prettier:fix": [],
-		"webpack": ["^tsc", "^build:esnext"],
-		"webpack:profile": ["^tsc", "^build:esnext"],
+		// TODO: ensure webpack dependencies are correct and minimal, ideally just depending
+		// on "^api-extractor:esnext", "^build:esnext". Or distribute these setting to the
+		// individual packages that need them instead of having poor config in the root config.
+		"webpack": ["^api-extractor:esnext", "^build:esnext"],
+		"webpack:profile": ["^api-extractor:esnext", "^build:esnext"],
```
MarioJGMsoft, line 188: Why can this change not be done in this PR? What's the blocker here?

README.md
```diff
@@ -215,8 +222,11 @@ which could have other implications for your system, so proceed with caution.
 
 ## Testing
 
-You can run all of our tests from the root of the repo, or you can run a scoped set of tests by running the `test`
-command from the package you're interested in.
+You can run:
+
+1. all of our tests from the root of the repo using `pnpm test`
+2. a scoped set of tests by running the `pnpm test` command from the package you're interested in
+3. incrementally build and test using `pnpm build-and-test <optional name-regex>` (see specific [package.json](./package.json) scripts). Note that test output will only be shown if there is a test error.
```
ChumpChief, line 229: Similar comment here about reducing the noise/number of ways to run dev commands to increase confidence (also I see the root package.json uses the `--worker` flag).  Surely it's more convenient to just run the commands individually?

packages/test/test-end-to-end-tests/package.json
```diff
@@ -152,7 +152,7 @@
 			"build:test:esm": [
 				"^build:esnext",
 				"build:genver",
-				"^tsc"
+				"@fluid-internal/test-driver-definitions#tsc"
```
ChumpChief, line 156: This might benefit from a comment

package.json
```diff
@@ -23,6 +23,8 @@
 		"build-and-test:unit:esm": "fluid-build --task test:unit:esm --worker",
 		"build:api": "fluid-build --task build:api",
 		"build:compile": "fluid-build --task compile",
+		"build:compile:esm": "fluid-build --task compile:esm --worker",
```
ChumpChief, line 26: Does `--worker` give better results on your machine?  If not, I'd suggest dropping it here.

ChumpChief (approved):
Love the idea of only running esm builds for a faster loop.  I'm mainly interested whether `--worker` actually helps, and how we can start to trim back to a minimal set of build commands that are recommendable for regular use.

README.md
jason-ha: Comparing the ESM only builds on a slower system (Windows laptop) I had an 8-minute build with `--worker` and then a 12-minute build without. The invocation of `tsc` in worker can save a lot of time.
Using `--worker`:
```text
	[244/251] ✓ @fluid-example/smde: [worker] tsc --project ./tsconfig.json - 15.998s
	[245/251] ✓ @fluid-example/table-tree: [worker] tsc --project ./tsconfig.json - 17.033s
```
Without:
```text
	[244/251] ✓ @fluid-example/smde: tsc --project ./tsconfig.json - 26.336s
	[245/251] ✓ @fluid-example/table-tree: tsc --project ./tsconfig.json - 30.435s
```
It seems like `--worker` should be the default.

fluidBuild.config.cjs
jason-ha: The comment is stale. The TODO (from an early commit) was done (in a follow-up).

packages/test/test-end-to-end-tests/package.json
jason-ha: now commented

README.md
ChumpChief: Mkay sounds good.  Maybe solicit feedback from the team over time to see if results vary (not blocking).  Would probably still be good for us to aim to reduce the number of different ways to build over time for simplicity (and iron out any reasons that someone would be dissatisfied with the default).

ChumpChief (approved):

MarioJGMsoft (approved):

PR #27268, opened 2026-05-08
feat(container-runtime, container-loader): make offline capabilities default-on, consolidate flags into disableOfflineFull

dannimad:
## Description

Offline capabilities (batch-ID tracking, duplicate-batch detection, offline load, and snapshot refresh) are now **on by default** for interactive clients. The previous patchwork of opt-in flags has been replaced by a single inverted kill-switch.

**Removed flags:**
- `Fluid.Container.enableOfflineFull`
- `Fluid.Container.enableOfflineLoad` (config flag; deprecated `IContainerLoadOptions.enableOfflineLoad` option is unchanged)
- `Fluid.Container.enableOfflineSnapshotRefresh`
- `Fluid.ContainerRuntime.DisableBatchIdTracking`

**New flag:**
- `Fluid.Container.disableOfflineFull` — set to `true` to turn off all offline features

**Behavior change:** Runtime-level offline features (batch-ID tracking, duplicate-batch detection) now silently degrade when prerequisites (TurnBased flush mode + grouped batching) are not met, instead of throwing a `UsageError`. This matches the existing behavior for the FlushMode.Immediate and non-grouped-batching cases.

## Reviewer Guidance

Key things to verify:
- `container.ts`: three config reads collapsed into one negated gate; `options.enableOfflineLoad !== false` still honored for the deprecated public option
- `snapshotRefresher.ts`: `#snapshotRefreshEnabled` simplified to `= offlineLoadEnabled` — the field is kept but the separate config gate is gone
- `containerRuntime.ts`: two `UsageError` throw blocks removed; `batchIdTrackingEnabled` now gated by `!offlineDisabled` instead of `!disableBatchIdTracking`
- Two "Offline Load opt-in errors when..." tests deleted (tested the now-removed `UsageError` paths)
- `enableOfflineSnapshotRefresh()` test helpers removed from `snapshotRefresher.spec.ts` and `serializedStateManager.spec.ts`


packages/loader/container-loader/src/snapshotRefresher.ts
```diff
@@ -95,10 +95,7 @@ export class SnapshotRefresher implements IDisposable {
 
 		this.snapshotRefreshTimeoutMs = snapshotRefreshTimeoutMs ?? this.snapshotRefreshTimeoutMs;
 
-		this.#snapshotRefreshEnabled =
-			this.offlineLoadEnabled &&
-			(this.mc.config.getBoolean("Fluid.Container.enableOfflineSnapshotRefresh") ??
-				this.mc.config.getBoolean("Fluid.Container.enableOfflineFull")) === true;
+		this.#snapshotRefreshEnabled = this.offlineLoadEnabled;
```
anthony-murphy: Deep Review: Snapshot-refresh default-on reverses the #20840 stress kill-switch with no on-record AB#7810 evidence.

This line collapses

```ts
#snapshotRefreshEnabled = offlineLoadEnabled && (config("enableOfflineSnapshotRefresh") ?? config("enableOfflineFull")) === true
```

down to `#snapshotRefreshEnabled = this.offlineLoadEnabled`. The flag being removed is the exact mitigation anthony-murphy added in PR #20840 (Merged 2024-04-25):

> *"After #20427 stress is seeing a large increase in 'Ranges finalized out of order.' In order to reduce these errors i've put this functionality under a feature flag to temporarily mitigate AB#7810."*

Neither the PR description nor `.changeset/offline-full-default-on.md` cites AB#7810 status, the original stress regression, or a fresh stress run validating default-on. The global `Fluid.Container.disableOfflineFull` kill-switch caps blast radius if the regression returns, but that's a mitigation, not evidence the underlying bug is fixed.

**Suggested fix:** Add a paragraph to the PR description and a "Risk" line to `.changeset/offline-full-default-on.md` citing AB#7810 status and linking a recent service-load stress run showing the "Ranges finalized out of order" rate is clean with snapshot refresh tied to `offlineLoadEnabled`. Tag anthony-murphy explicitly — he authored the kill-switch and owns the snapshot-refresh stability story.

Related: the load-stress matrix (`packages/test/test-service-load/src/runner.ts`, `optionsMatrix.ts`) still gates stashing on `enableOfflineLoad === true` and varies the now-removed `enableOfflineSnapshotRefresh` option — default/undefined loader options exercise default-on product behavior but not the stash-and-resume stress path. Worth aligning at the same time.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1890,20 +1890,24 @@ export class ContainerRuntime
...
-			this.mc.config.getBoolean("Fluid.ContainerRuntime.enableBatchIdTracking") ??
-			false;
+		// BatchId tracking powers DuplicateBatchDetector (catching forked-container duplicates)
+		// and is also a prerequisite for the Offline Load feature. Offline is on by default;
+		// the disableOfflineFull kill-switch turns off all offline capabilities. Tracking still
+		// requires TurnBased flush mode and grouped batching, so it silently degrades when those
+		// prerequisites are missing. Grouped batching is required because resubmits can produce
+		// empty batches that must still be sent on the wire as a placeholder grouped batch to
+		// preserve their batchId (see OpGroupingManager.createEmptyGroupedBatch /
+		// outbox.flushEmptyBatch).
```
anthony-murphy, line 1900: Deep Review: Silent degradation of batch-id tracking emits no telemetry — overrides andre4i's prior pushback on #14239.

This block replaces two `UsageError`-and-`closeFn` throws with

```ts
this.batchIdTrackingEnabled = !offlineDisabled && this._flushMode === FlushMode.TurnBased && this.groupedBatchingEnabled;
```

No `mc.logger.sendTelemetryEvent` is emitted on the silent-skip path. New tests (`Default-on tracking is silently skipped in FlushMode.Immediate (no UsageError)` and the grouped-batching-off counterpart) confirm the silent path is intended.

PR #14239 (andre4i, 2023-02-21) raised exactly this tradeoff:

> *"I would at least post an error. The risk is to think that all is well and fail silently with the expectation that stuff is supposed to work differently while it doesn't."*

Now that the feature is on by default, partners can no longer rely on "I would have seen a throw."

**Suggested fix:** Emit a one-shot telemetry event from the constructor when `!offlineDisabled` is true but one of `_flushMode === FlushMode.TurnBased` / `groupedBatchingEnabled` is false. Payload should name which prereq is missing, plus both `flushMode` and `groupedBatchingEnabled` values. Single emit per container, low volume — restores the observability the deleted `UsageError` provided via container-close error reporting.

Worth tagging andre4i to confirm whether a telemetry event satisfies his concern.

packages/loader/container-loader/src/snapshotRefresher.ts
dannimad: Reinstated `Fluid.Container.disableOfflineSnapshotRefresh` as a finer-grained sub-switch — when set to `true`, snapshot refresh is disabled independently of the rest of offline load. The `#snapshotRefreshEnabled` field is now `offlineLoadEnabled && config.getBoolean("Fluid.Container.disableOfflineSnapshotRefresh") !== true`. Updated `.changeset/offline-full-default-on.md` to document the sub-switch. Coverage in `snapshotRefresher.spec.ts` (`disableOfflineSnapshotRefresh config` block). Preserves the #20840 mitigation path without depending on AB#7810 status. (commit 20b7ae3fea)

packages/runtime/container-runtime/src/containerRuntime.ts
dannimad: Added a one-shot `OfflineBatchIdTrackingDegraded` telemetry event in the constructor, gated by `!offlineDisabled && !this.batchIdTrackingEnabled`. Payload: `reason` (one of `NotTurnBased`, `GroupedBatchingDisabled`, or both comma-joined), `flushMode`, `groupedBatchingEnabled`. The event is suppressed when the user explicitly set `disableOfflineFull` — that case is intentional, not degradation. New tests in `containerRuntime.spec.ts` under `Duplicate Batch Detection > OfflineBatchIdTrackingDegraded telemetry` (4 tests covering both positive paths plus happy-path and kill-switch suppression). Tagging @andre4i to confirm this satisfies the observability concern from #14239. (commit 20b7ae3fea)

packages/loader/container-loader/src/snapshotRefresher.ts
anthony-murphy: Deep Review: Reinstating `Fluid.Container.disableOfflineSnapshotRefresh` as a sub-switch addresses the mechanical risk — thanks. The remaining piece of the original ask is the AB#7810 disposition: the changeset names the sub-switch as a "stability mitigation for server-side ordering regressions," but neither it nor the PR description states whether AB#7810 is resolved (link the fix), or — if still open — what the staged-rollout/monitoring plan is (which telemetry is watched, when `disableOfflineSnapshotRefresh` would be flipped in production). Adding one "Risk" line to `.changeset/offline-full-default-on.md` covering this would close the thread.

Also flagged separately as new threads on this commit: the changeset's removed-flags sentence omits `Fluid.Container.enableOfflineSnapshotRefresh`, and the service-load `optionsMatrix.ts` no longer exercises either path of the new sub-switch — both directly support the kill-switch-as-mitigation safety story.

Tagging anthony-murphy for the AB#7810 disposition call.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: The one-shot `OfflineBatchIdTrackingDegraded` event with `reason` / `flushMode` / `groupedBatchingEnabled` payload restores the observability the deleted `UsageError` provided, and suppressing it when the user explicitly set `disableOfflineFull` is the right call. From the pipeline's side this resolves the structural concern. Deferring final sign-off to andre4i — his original #14239 pushback was the basis for the ask, and his confirmation that telemetry (vs. a louder failure) is sufficient is what closes the thread.

One related polish item flagged separately as a new inline thread: the "Pre-fix" preamble on the new regression test in `containerRuntime.spec.ts:685-690` describes a fix this PR doesn't actually make — the `groupedBatchingEnabled` gate already existed; what changed is removal of the `UsageError` safety-net, which makes that test more important, not less.

packages/test/test-service-load/src/optionsMatrix.ts
```diff
@@ -35,7 +34,6 @@ const loaderOptionsMatrix: OptionsMatrix<ILoaderOptionsExperimental> = {
 	provideScopeLoader: booleanCases,
 	maxClientLeaveWaitTime: numberCases,
 	enableOfflineLoad: booleanCases,
-	enableOfflineSnapshotRefresh: booleanCases,
 	snapshotRefreshTimeoutMs: [undefined, 60 * 5 * 1000 /* 5min */],
```
anthony-murphy, line 37: Deep Review: Stress matrix loses coverage of the snapshot-refresh-off path.

This diff removes `enableOfflineSnapshotRefresh?: boolean` from `ILoaderOptionsExperimental` and `enableOfflineSnapshotRefresh: booleanCases` from `loaderOptionsMatrix`, but nothing replaces them. The new `Fluid.Container.disableOfflineSnapshotRefresh` config (referenced by `snapshotRefresher.ts:101-103` and called out in `.changeset/offline-full-default-on.md` as the stability-mitigation lever) is not wired into the service-load options matrix. Before this PR, stress exercised both refresh-on and refresh-off paths; after this PR, only refresh-on.

PR #21504 (dannimad, 2024-06-27) explicitly added `enableOfflineSnapshotRefresh: booleanCases` when periodic refresh shipped — the matrix coverage was intentional for the same stability reason the sub-switch exists.

**Suggested fix:** Add `disableOfflineSnapshotRefresh: booleanCases` to `loaderOptionsMatrix` (or wire the config-flag equivalent through the load-stress framework's config-override path) so stress continues to exercise both branches of the operator-facing emergency lever. A `disableOfflineFull: booleanCases` entry alongside would be cheap insurance for the global kill-switch.

.changeset/offline-full-default-on.md
```diff
@@ -0,0 +1,13 @@
...
+"@fluidframework/container-runtime": minor
+"@fluidframework/container-loader": minor
+"__section": other
+---
+
+Consolidate offline feature flags into a single `disableOfflineFull` kill-switch
+
+All offline capabilities (batch-ID tracking, duplicate-batch detection, offline load, snapshot refresh) are now **on by default**. The previous opt-in flags `Fluid.Container.enableOfflineFull`, `Fluid.Container.enableOfflineLoad`, and `Fluid.ContainerRuntime.DisableBatchIdTracking` have been removed.
+
+To disable all offline features, set the config flag `Fluid.Container.disableOfflineFull` to `true`. A finer-grained sub-switch `Fluid.Container.disableOfflineSnapshotRefresh` can independently disable just the snapshot-refresh path while leaving the rest of offline load enabled — retained as a stability mitigation for server-side ordering regressions.
```
anthony-murphy: Deep Review: Removed-flags sentence omits `Fluid.Container.enableOfflineSnapshotRefresh`.

The changeset lists `Fluid.Container.enableOfflineFull`, `Fluid.Container.enableOfflineLoad`, and `Fluid.ContainerRuntime.DisableBatchIdTracking` as removed, but not `Fluid.Container.enableOfflineSnapshotRefresh`. `snapshotRefresher.ts:101-103` only reads the new `disableOfflineSnapshotRefresh` flag, and the diff removes test helpers that set the old flag — the omission is mechanical.

**Suggested fix:** Add `Fluid.Container.enableOfflineSnapshotRefresh` to the removed-flags sentence, and point migrating hosts who relied on the old opt-in to `Fluid.Container.disableOfflineSnapshotRefresh=true` when they need the historical stability mitigation.

.changeset/offline-full-default-on.md
```diff
@@ -0,0 +1,13 @@
...
+
+Runtime-level offline features (batch-ID tracking, duplicate-batch detection) silently degrade instead of throwing a `UsageError` when prerequisites (TurnBased flush mode + grouped batching) are not met. When this happens, a one-shot `OfflineBatchIdTrackingDegraded` telemetry event is emitted naming the missing prerequisite.
```
anthony-murphy, line 13: Deep Review: `Fluid.ContainerRuntime.DisableBatchIdTracking` migration not mapped.

Diff lines 609-610 and 633 in `containerRuntime.ts` remove all reads of `Fluid.ContainerRuntime.DisableBatchIdTracking`, and the `UsageError` safety-net throws are also removed in this PR. The changeset lists the flag as removed but does not tell operators who set it (as a mitigation for a batch-tracking regression) how to preserve old behavior — they will silently get tracking re-enabled with no `UsageError` to surface the change.

Both blind design proposals during review independently recommended keeping the key for one release as a deprecation alias; the clean removal here is intentional, but the migration doc gap is not.

**Suggested fix:** Add an explicit migration sentence: `Fluid.ContainerRuntime.DisableBatchIdTracking=true` → `Fluid.Container.disableOfflineFull=true` to preserve the prior "tracking off" behavior.

packages/loader/container-loader/src/container.ts
```diff
@@ -963,9 +963,8 @@ export class Container
 
 		const offlineLoadEnabled =
 			this.isInteractiveClient &&
-			(this.mc.config.getBoolean("Fluid.Container.enableOfflineLoad") ??
-				this.mc.config.getBoolean("Fluid.Container.enableOfflineFull") ??
-				options.enableOfflineLoad !== false);
+			options.enableOfflineLoad !== false &&
+			this.mc.config.getBoolean("Fluid.Container.disableOfflineFull") !== true;
```
anthony-murphy: Deep Review: Two undocumented behavior flips in the `offlineLoadEnabled` collapse.

Pre-PR: `offlineLoadEnabled = getBoolean("Fluid.Container.enableOfflineLoad") ?? getBoolean("Fluid.Container.enableOfflineFull") ?? options.enableOfflineLoad !== false`.

Post-PR (this hunk): the legacy keys are ignored entirely, and `options.enableOfflineLoad !== false && disableOfflineFull !== true`.

Two silent semantic changes:

1. A host setting `Fluid.Container.enableOfflineLoad=false` or `enableOfflineFull=false` previously short-circuited to disable offline. Now those keys are ignored and offline silently re-enables. Partners holding these defensively as a kill-switch lose the kill.
2. `options.enableOfflineLoad` becomes a hard veto — previously config could override the deprecated public option; now it cannot.

The changeset documents neither flip. Both blind design proposals recommended honoring `=== false` on the old keys for one release.

**Suggested fix:** Either (a) honor `=== false` on the two legacy keys for one release as a deprecation alias (3 lines — closes the only realistic back-compat hole), or (b) add an explicit sentence to `.changeset/offline-full-default-on.md` calling out both flips and pointing hosts to `disableOfflineFull=true`.

packages/loader/container-loader/src/snapshotRefresher.ts
dannimad: Added a Risk paragraph to `.changeset/offline-full-default-on.md` citing AB#7810 resolution (fix tracked in https://github.com/microsoft/FluidFramework/pull/26898) and explaining the `disableOfflineSnapshotRefresh` sub-switch as a belt-and-suspenders emergency lever if the regression class re-emerges. Also added the missing `enableOfflineSnapshotRefresh` to the removed-flags list and re-added matrix coverage in `optionsMatrix.ts`. (commit d00ccbf919)

packages/runtime/container-runtime/src/containerRuntime.ts
dannimad: Rewrote the regression-test preamble (containerRuntime.spec.ts) to call out that the removal of the `UsageError` safety-net in this PR makes this regression test the primary guarantee that the empty-batch path is unreachable, not just an assertion of the `groupedBatchingEnabled` gate. (commit d00ccbf919)

Still tagging @andre4i for telemetry-vs-louder-failure sign-off on the original ask.

packages/test/test-service-load/src/optionsMatrix.ts
dannimad: Restored coverage. `loaderOptionsMatrix` in `optionsMatrix.ts` now includes `disableOfflineSnapshotRefresh: booleanCases` and (as cheap insurance) `disableOfflineFull: booleanCases`, so stress continues to exercise both branches of each operator-facing emergency lever. (commit d00ccbf919)

.changeset/offline-full-default-on.md
dannimad: Added `Fluid.Container.enableOfflineSnapshotRefresh` back to the removed-flags list in the changeset, plus a migration pointer in the new Migration section directing hosts who relied on the old opt-in to `Fluid.Container.disableOfflineSnapshotRefresh=true`. (commit d00ccbf919)

.changeset/offline-full-default-on.md
dannimad: Added an explicit migration line to the changeset: hosts setting `Fluid.ContainerRuntime.DisableBatchIdTracking=true` should now set `Fluid.Container.disableOfflineFull=true` instead. (commit d00ccbf919)

packages/loader/container-loader/src/container.ts
dannimad: Went with option (a) — the deprecation alias. Extracted the gate into `isOfflineLoadEnabled` in `container.ts` and now honor `Fluid.Container.enableOfflineLoad === false` and `Fluid.Container.enableOfflineFull === false` for one release as deprecation aliases. Partners holding either key defensively as a kill-switch keep their kill. Covered by 7 unit tests in `container.spec.ts` (`isOfflineLoadEnabled` describe block), including explicit cases for each legacy `=== false` path and a negative case verifying `=== true` is a no-op. Migration also documented in the changeset. (commit d00ccbf919)

packages/loader/container-loader/src/snapshotRefresher.ts
anthony-murphy: Deep Review: AB#7810 disposition note (citing #26898) and the `disableOfflineSnapshotRefresh` Risk paragraph close the structural piece of this thread. Final sign-off is anthony-murphy's call — flagged in the For human reviewer block of the summary.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: The rewritten regression-test preamble correctly frames the test as the primary guarantee that the empty-batch path is unreachable post-`UsageError`-removal. Structural piece closed; final telemetry-vs-louder-failure sign-off is andre4i's call — flagged in the For human reviewer block.

packages/test/test-service-load/src/optionsMatrix.ts
anthony-murphy: Deep Review: Re-adding the entries doesn't actually restore coverage — `disableOfflineFull` and `disableOfflineSnapshotRefresh` are declared on `ILoaderOptionsExperimental` and routed through `loaderOptionsMatrix`, but the product reads them via `IConfigProvider` (`containerRuntime.ts:1901-1907` reads `Fluid.Container.disableOfflineFull` from `this.mc.config`; `snapshotRefresher.ts:98-103` reads `Fluid.Container.disableOfflineSnapshotRefresh` from `this.mc.config`). Neither key is surfaced on `ILoaderOptions`. In `test-service-load/src/runner.ts:267-290`, generated values flow into `options: runConfig.loaderConfig`, while `configProvider` is constructed separately from `testConfiguration` (`utils.ts:167-170`). Net result: pairwise stress runs record the booleans in `loaderOptions` telemetry without toggling either kill-switch path, while the product runs only the default.

**Suggested fix:** Move `disableOfflineFull` and `disableOfflineSnapshotRefresh` out of `loaderOptionsMatrix` / `ILoaderOptionsExperimental` and into the configurations matrix surfaced via `generateConfigurations`, so they end up in the `IConfigProvider` settings constructed for each run. Verify via a quick local run that resulting telemetry shows the kill-switch path being exercised, not just the loader-options field being recorded.

.changeset/offline-full-default-on.md
anthony-murphy: Deep Review: Removed-flags list and migration pointer look correct.

.changeset/offline-full-default-on.md
anthony-murphy: Deep Review: Migration sentence is in. One residual: the changeset documents the migration but a host that has `Fluid.ContainerRuntime.DisableBatchIdTracking=true` deployed today gets `batchIdTrackingEnabled` silently flipping from `false` to `true` on upgrade with no `UsageError` and no warning. Either honor `=== true` for one release as a deprecation alias symmetric with the new loader-side `=== false` aliases on `enableOfflineLoad` / `enableOfflineFull` (thread 3222156327), or emit a one-shot deprecation telemetry when the legacy key is observed (mirroring the `OfflineBatchIdTrackingDegraded` pattern from thread 3211824403). Either option is cheap and matches the rest of the PR's migration story.

packages/loader/container-loader/src/container.ts
anthony-murphy: Deep Review: Option (a) deprecation aliases land cleanly in `isOfflineLoadEnabled` with the truth-table tests in `container.spec.ts` — thanks. One asymmetry exposed by the fix: the runtime gate at `containerRuntime.ts:1900-1907` reads only the new `Fluid.Container.disableOfflineFull`, so a host with `Fluid.Container.enableOfflineFull === false` deployed defensively now lands in a partially-degraded state — loader treats offline as disabled, runtime silently turns batch-id tracking and duplicate-batch detection ON. That's the exact migration-contract scenario the loader-side alias was added to prevent. Flagged as a new inline on `containerRuntime.ts:1900` with the suggested OR-in fix.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1891,39 +1891,41 @@ export class ContainerRuntime
...
-			this.closeFn(error);
-			throw error;
-		}
+		// and is also a prerequisite for the Offline Load feature. Offline is on by default;
+		// the disableOfflineFull kill-switch turns off all offline capabilities. Tracking still
+		// requires TurnBased flush mode and grouped batching, so it silently degrades when those
+		// prerequisites are missing. Grouped batching is required because resubmits can produce
+		// empty batches that must still be sent on the wire as a placeholder grouped batch to
+		// preserve their batchId (see OpGroupingManager.createEmptyGroupedBatch /
+		// outbox.flushEmptyBatch).
```
anthony-murphy, line 1900: Deep Review: Runtime gate does not honor the legacy `=== false` aliases the loader now honors — asymmetric one-release migration contract.

This block:

```ts
const offlineDisabled =
    this.mc.config.getBoolean("Fluid.Container.disableOfflineFull") === true;

this.batchIdTrackingEnabled =
    !offlineDisabled &&
    this._flushMode === FlushMode.TurnBased &&
    this.groupedBatchingEnabled;
```

consults only the new `Fluid.Container.disableOfflineFull` key. In this same commit (d00ccbf) the loader-side `isOfflineLoadEnabled` (`container.ts:313-340`, exercised in `container.spec.ts:243-255`) explicitly preserves `Fluid.Container.enableOfflineLoad === false` and `Fluid.Container.enableOfflineFull === false` as one-release deprecation aliases (per the resolution of thread 3222156327). The changeset (`.changeset/offline-full-default-on.md:17-19`) advertises this `=== false` aliasing as the migration path.

Two consequences:

1. A host that has `Fluid.Container.enableOfflineFull === false` deployed defensively gets the loader treating offline as disabled while the runtime silently turns batch-id tracking and duplicate-batch detection ON — the exact partially-degraded migration scenario the loader alias was added to prevent.
2. The previously-supported `Fluid.ContainerRuntime.DisableBatchIdTracking === true` kill-switch is removed with no alias and no warning telemetry (separately flagged on thread 3222156197); `batchIdTrackingEnabled` flips from `false` to `true` silently on upgrade.

**Suggested fix:** Mirror the loader by OR-ing the legacy `=== false` checks for the one-release window:

```ts
const offlineDisabled =
    this.mc.config.getBoolean("Fluid.Container.disableOfflineFull") === true ||
    this.mc.config.getBoolean("Fluid.Container.enableOfflineFull") === false ||
    this.mc.config.getBoolean("Fluid.Container.enableOfflineLoad") === false;
```

Add a truth-table-style unit test mirroring `container.spec.ts:243-255`.

packages/test/test-service-load/src/optionsMatrix.ts
dannimad: Fixed properly. Removed `disableOfflineFull` and `disableOfflineSnapshotRefresh` from `ILoaderOptionsExperimental` / `loaderOptionsMatrix` (where they were inert because the product reads them via `IConfigProvider`, not loader options), and introduced a default `configurationsMatrix` in `optionsMatrix.ts` containing both kill-switches as `booleanCases`. `generateConfigurations` now applies that matrix via `applyOverrides`, so stress runs actually toggle the kill-switch paths in the config provider instead of just recording the boolean fields in `RunConfigOptions` telemetry. (commit 8e8e64ae11)

.changeset/offline-full-default-on.md
dannimad: Went with option (a) — symmetric deprecation alias. The runtime gate at `containerRuntime.ts` now honors `Fluid.ContainerRuntime.DisableBatchIdTracking === true` for one release alongside the new `Fluid.Container.disableOfflineFull` kill-switch, so hosts with the legacy key deployed today do not silently flip on upgrade. Same commit also adds the symmetric `Fluid.Container.enableOfflineFull === false` honoring from the related thread. Covered by new tests in `containerRuntime.spec.ts` (`Legacy deprecation aliases for batch-id tracking` block) and called out in the changeset migration section. (commit 8e8e64ae11)

packages/runtime/container-runtime/src/containerRuntime.ts
dannimad: Fixed. `offlineDisabled` in `containerRuntime.ts` now ORs in the two legacy paths:

```ts
const offlineDisabled =
    this.mc.config.getBoolean("Fluid.Container.disableOfflineFull") === true ||
    this.mc.config.getBoolean("Fluid.Container.enableOfflineFull") === false ||
    this.mc.config.getBoolean("Fluid.ContainerRuntime.DisableBatchIdTracking") === true;
```

So a host with either legacy key deployed defensively turns batch-id tracking + duplicate-batch detection off symmetrically with the loader-side `isOfflineLoadEnabled` — no partial-degraded state. Note I intentionally did not add `Fluid.Container.enableOfflineLoad === false` here because pre-PR that key only affected offline load, not batch-id tracking; preserving historical scoping. Covered by 3 new tests in `containerRuntime.spec.ts` (`Legacy deprecation aliases for batch-id tracking`) and the changeset's migration section. (commit 8e8e64ae11)

anthony-murphy:
## Deep Review

*Reviewed commit `8e8e64a` on 2026-05-11.*

Not blocking. The 8e8e64a round resolved both Tier 2 concerns from the prior review: the runtime gate at `containerRuntime.ts:1900-1907` now ORs in `Fluid.Container.enableOfflineFull === false` and `Fluid.ContainerRuntime.DisableBatchIdTracking === true` (symmetric one-release deprecation aliases, covered by the new `Legacy deprecation aliases for batch-id tracking` block), and the stress kill-switches were re-routed from `loaderOptionsMatrix` into a new default `configurationsMatrix` so they actually flow through `IConfigProvider`. What remains is a small cluster of polish items flagged inline (stress-runner predicate, matrix `undefined` case, summarizer-side telemetry suppression) plus pending human sign-offs.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- This PR reverses #20840 — `Fluid.Container.enableOfflineSnapshotRefresh` was added there to mitigate AB#7810 ("Ranges finalized out of order") after #20427. The changeset asserts AB#7810 is resolved (fix in #26898); the retained `disableOfflineSnapshotRefresh` sub-switch is the belt-and-suspenders rollback lever.
- Snapshot refresh has tightly-coupled correctness constraints from the #20257 → #20427 → #20840 chain (background refresh during pending-state load, session-expiry / dirty / saved-op trimming conditions) and #20565 (groupId snapshot availability for stashed ops). Focus default-on review on those interactions, not the flag-renaming mechanics.
- Where to scrutinize hardest in the diff: (1) the runtime legacy-alias block in `containerRuntime.ts:1903-1912` — verify the deprecation telemetry/migration plan; (2) `OfflineBatchIdTrackingDegraded` is now the only post-merge signal that partner misconfig hit the silent-degrade path — verify it is wired into a category SREs notice; (3) the `snapshotRefresher.ts` default flip; (4) the `optionsMatrix.ts` shift from `if (overrides === undefined) return [{}]` to always merging into `configurationsMatrix` is a quiet stress-matrix expansion (see inline threads).
- Suggested reviewers: **anthony-murphy** (authored #20840, the snapshot-refresh kill-switch this PR's default removes); **andre4i** (authored the silent-fallback-without-telemetry pushback on #14239; tagged on thread 3211824403); **markfields** (designed batchId tracking in #21767; should weigh in on the runtime default-flip); **vladsud** (reviewed #14239 / #20427; standing fuzz-coverage ask in this area); **tyler-cai-microsoft** (authored #20565 groupId offline rework); **ChumpChief** (structural reviewer for `container.ts`).

- **Needs human judgment — anthony-murphy on AB#7810 disposition** (thread 3211824322): the changeset cites #26898 and retains `disableOfflineSnapshotRefresh` as the rollback lever; final acceptance that AB#7810 is genuinely closed and default-on snapshot refresh is safe in current production telemetry is anthony-murphy's call as the original kill-switch author.
- **Needs human judgment — andre4i on the `OfflineBatchIdTrackingDegraded` telemetry-vs-throw tradeoff** (thread 3211824403): the structural ask is met; final acceptance that telemetry suffices versus a louder failure is rooted in his #14239 pushback.
- **Needs human judgment — markfields on the runtime default-flip**: should sign off on flipping `batchIdTrackingEnabled` to default-on with silent-degradation-plus-telemetry semantics (he designed batchId tracking in #21767).
- **Needs human judgment — vladsud / dannimad on partner rollout shape**: whether the migration shape (loader honors `enableOfflineLoad === false` and `enableOfflineFull === false`; runtime honors `enableOfflineFull === false` and `DisableBatchIdTracking === true` but not `enableOfflineLoad === false`; snapshot-refresh legacy key intentionally not aliased) is acceptable for known partner config patterns.
- **Cannot be assessed by the pipeline — service-load stress signal**: whether default-on snapshot refresh and default-on batch-id tracking are safe in practice depends on a live stress run, particularly once the inline polish items land and the kill-switch axes are actually being varied through the config provider.
- **Cannot be assessed by the pipeline — fuzz-coverage follow-up**: vladsud's standing #20427 ask applies directly now that this PR enlarges the population running offline-load + snapshot-refresh by default. Best filed as a follow-up tracking issue rather than gating this PR.

</details>

- `d00ccbf` 2026-05-11 · **5/10** — asymmetric runtime gate and stress-matrix entries routed through the wrong config surface
- `9c9fa27` 2026-05-11 · **7/10** — both structural concerns resolved; only Tier 3 polish remained
- `20b7ae3` 2026-05-11 · **7/10** — both structural concerns resolved; only Tier 3 polish remains
- `675c4bc` 2026-05-11 · **4/10** — snapshot-refresh default-on still reverses #20840 stress kill-switch with no on-record AB#7810 evidence
- `7c75430` 2026-05-08 · **5/10** — snapshot-refresh and silent-degradation concerns flagged inline

</details>

packages/test/test-service-load/src/optionsMatrix.ts
anthony-murphy: Deep Review: Re-routing both kill-switches into `configurationsMatrix` so they flow through `applyOverrides` → `IConfigProvider` is the right fix — confirmed by reading `optionsMatrix.ts:35-50` against `runner.ts`'s config-provider construction. Stress now actually toggles the kill-switch paths instead of just recording the boolean fields.

Two small follow-ups surfaced by the new wiring:

1. **Stress-runner stashing predicate is now misaligned.** `runner.ts:553-559` decides whether to call `getPendingLocalState()` from `runConfig.loaderConfig?.enableOfflineLoad === true` only. With `disableOfflineFull: booleanCases` now actually toggling, seeds where `enableOfflineLoad: true` co-occurs with `disableOfflineFull: true` will hit `isOfflineLoadEnabled` → `false` (`container.ts:323-340` short-circuits on `disableOfflineFull === true`), and `SerializedStateManager.getPendingLocalState()` throws (`serializedStateManager.ts:421-424`). The runner catches it and emits `ScheduleOfflineFailed` — not a crash, but the seed loses its intended stash-and-resume coverage and adds noise. Either gate the runner's predicate on the effective offline-load decision (mirroring `isOfflineLoadEnabled`'s combination of option + `disableOfflineFull` + legacy `=== false` aliases), or exclude the `disableOfflineFull: true` × `enableOfflineLoad: true` combo from generated seeds. One-line fix in `runner.ts`.

2. **Matrix never exercises the literal product default.** `booleanCases = [true, false]` for both new keys means every seed sets one of the two values explicitly — no seed runs with the keys *unset*, which is the literal default the product ships and which `getBoolean` returns `undefined` for. Today the gates treat `=== false` and `undefined` equivalently, but that's exactly the future-divergence case the matrix should guard against. Use `[undefined, true, false]` for both keys, or document explicitly why the unset path is not exercised. Two-character change.

.changeset/offline-full-default-on.md
anthony-murphy: Deep Review: Symmetric `Fluid.ContainerRuntime.DisableBatchIdTracking === true` deprecation alias in the runtime gate, `=== false` aliases on the loader keys, plus the `Legacy deprecation aliases for batch-id tracking` test block — closes the migration-contract gap. Thread resolved.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: `offlineDisabled` now ORs in `Fluid.Container.enableOfflineFull === false` and `Fluid.ContainerRuntime.DisableBatchIdTracking === true`, mirroring the loader-side `isOfflineLoadEnabled`. The intentional omission of `Fluid.Container.enableOfflineLoad === false` (preserving its pre-PR loader-only scoping) is the right call. New `Legacy deprecation aliases for batch-id tracking` tests cover the truth table. Thread resolved.

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: One follow-up on the `OfflineBatchIdTrackingDegraded` emission (`containerRuntime.ts:1920-1936`) now that the structural piece is in: the new `else if (!offlineDisabled)` branch fires whenever `flushMode !== TurnBased || !groupedBatching`, but consults neither `this.clientDetails` nor `this.isSummarizerClient`. The loader-side `isOfflineLoadEnabled` (`container.ts:323-340`) correctly excludes non-interactive clients, so the two layers will be out of sync for summarizer sessions and routine summarizer noise will dominate the channel that's supposed to be the post-merge signal that partner misconfig hit the silent-degrade path. Suggest gating the emission on `this.clientDetails?.capabilities?.interactive` (or `!this.isSummarizerClient`), matching the loader's exclusion, and adding one test mirroring the existing `OfflineBatchIdTrackingDegraded telemetry` block in `containerRuntime.spec.ts`.

packages/runtime/container-runtime/src/containerRuntime.ts
dannimad: enableOfflineFull flag is currently unused in production, so no need for keep any kind of back compat for it.

PR #27272, opened 2026-05-10
build(build-tools): update ts-morph and typescript dependencies

jason-ha:
Update:
 - ts-morph to ^27.0.2
 - typescript to ~5.9.3 (same as api-extractor)

Replace `typescript` module types exposed through `TscUtils` with 5.4 and 5.9 union expecting the consumers use either.
  - no runtime check performed; just as none was previously, nor were peerDependencies specified
  - some helpers were added to `TscUtil` interface to help with using non-specific TypeScript version

Set skipLibCheck in build-cli per @github/copilot-sdk use of vscode-jsonrpc that does not have release supporting TS 5.6+.

### DIFF
--- a/build-tools/packages/build-cli/src/commands/publish/tarballs.ts
+++ b/build-tools/packages/build-cli/src/commands/publish/tarballs.ts
@@ -180,8 +180,9 @@ async function extractPackageJsonFromTarball(
 	let unzipped: Uint8Array;
 
 	{
-		// eslint-disable-next-line no-return-assign -- assigning the chunk to unzipped is intentional
-		const gunzip = new Gunzip((chunk: Uint8Array) => (unzipped = chunk));
+		const gunzip = new Gunzip((chunk: Uint8Array) => {
+			unzipped = chunk;
+		});
 		gunzip.push(tarball, /* final */ true);
 	}
 	// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
--- a/build-tools/packages/build-cli/src/commands/publish/tarballs.ts
+++ b/build-tools/packages/build-cli/src/commands/publish/tarballs.ts
@@ -185,7 +185,7 @@ async function extractPackageJsonFromTarball(
 		gunzip.push(tarball, /* final */ true);
 	}
 	// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-	const data = untar(unzipped!);
+	const data = untar(unzipped!.buffer as ArrayBuffer);
 	// eslint-disable-next-line unicorn/prefer-string-slice -- substring is clearer than slice in this case
 	const prefix = data[0].filename.substring(0, data[0].filename.indexOf("/") + 1);
 	const packageJsonText = data.find((f) => f.filename === `${prefix}package.json`)?.fileData;
--- a/build-tools/packages/build-cli/tsconfig.json
+++ b/build-tools/packages/build-cli/tsconfig.json
@@ -11,5 +11,10 @@
 		"target": "ES2022",
 		"noUncheckedIndexedAccess": false,
 		"exactOptionalPropertyTypes": false,
+		// vscode-jsonrpc (transitive dep via @github/copilot-sdk) uses IterableIterator in its
+		// LinkedMap types, which is incompatible with TS 5.6+'s MapIterator.
+		// See https://github.com/microsoft/vscode-languageserver-node/issues/1590 tracking the issue.
+		// Fix will be in vscode-jsonrpc 9.0.0 which has not been released as of 2026-05-10.
+		"skipLibCheck": true,
 	},
 }
--- a/build-tools/pnpm-lock.yaml
+++ b/build-tools/pnpm-lock.yaml
@@ -45,13 +45,13 @@ importers:
         version: 2.4.5
       '@commitlint/cli':
         specifier: ^20.1.0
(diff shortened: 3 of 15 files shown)


build-tools/packages/build-cli/tsconfig.json
```diff
@@ -11,5 +11,10 @@
 		"target": "ES2022",
 		"noUncheckedIndexedAccess": false,
 		"exactOptionalPropertyTypes": false,
+		// vscode-jsonrpc (transitive dep via @github/copilot-sdk) uses IterableIterator in its
+		// LinkedMap types, which is incompatible with TS 5.6+'s MapIterator.
+		// See https://github.com/microsoft/vscode-languageserver-node/issues/1590 tracking the issue.
+		// Fix will be in vscode-jsonrpc 9.0.0 which has not been released as of 2026-05-10.
```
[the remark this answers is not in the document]
jason-ha: This suggestion is odd. "Currently" or "time of writing" is nebulous and requires extra work to know the timing,

tylerbutler, line 17: nit: might be nice to include this in the DEV.md since there are notes on other deps and settings there.

tylerbutler (approved):

build-tools/pnpm-lock.yaml
```diff
@@ -1916,6 +1916,9 @@ packages:
   '@ts-morph/common@0.23.0':
```
alexvy86, line 1922: Nit: did we run dedupe? A couple of things in the lockfile feel like they might go away.

build-tools/packages/build-cli/src/commands/publish/tarballs.ts
```diff
@@ -185,7 +185,7 @@ async function extractPackageJsonFromTarball(
 		gunzip.push(tarball, /* final */ true);
 	}
 	// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-	const data = untar(unzipped!);
+	const data = untar(unzipped!.buffer as ArrayBuffer);
```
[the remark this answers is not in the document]
jason-ha: There is no need to use slice and check offset+length when the calling pattern is exactly one call to Gunzip.push as is the case here.

build-tools/pnpm-lock.yaml
jason-ha: dedupe made no changes - once build-tools updates to the release after this update is published I expect a lot of entries to go away.

jason-ha:
This needs updated to not assume the build-tools version of TypeScript (now 5.9) is what is always resolved in the context of use/users.

build-tools/pnpm-lock.yaml
jason-ha: In most recent update, 5.4.5 is kept as a production dependency in case any other package uses TscUtil support. So deps won't be cleaned up until we collectively move workspaces forward.

jason-ha:
> ## 🔭 PR Review Fleet Report
> Note
> 
> This report is generated by an experimental AI review fleet and is provided as a **beta feature**. Findings are a starting point for discussion, not a gate. Use your own judgement.
> 
> **Verdict:** ❌ Request Changes
> 
> 0 Spicy, 1 Pungent, 0 Smelly
> 
> ### Findings
> Sev	#	Area	File	What	Fix
> 🧄 Pungent	H1	Correctness	`build-tools/packages/build-cli/src/commands/publish/tarballs.ts:189`	`untar(unzipped!.buffer as ArrayBuffer)` is unsafe when the `Uint8Array` produced by fflate's streaming `Gunzip` callback has a non-zero `byteOffset`. `Uint8Array.prototype.buffer` always returns the _entire_ underlying `ArrayBuffer`, not just the slice the view represents. fflate v0.8's streaming decompressor may call the callback with a subarray view of its internal output buffer (i.e. `byteOffset > 0` and/or `byteLength < buffer.byteLength`). Passing the full `ArrayBuffer` to `untar` would cause it to parse extra leading bytes as tar headers, producing wrong entries or throwing on malformed data. The previous code passed the `Uint8Array` directly, which carried the correct view metadata and was safe.	Use `unzipped!.buffer.slice(unzipped!.byteOffset, unzipped!.byteOffset + unzipped!.byteLength)` instead of `unzipped!.buffer` so `untar` receives only the bytes that represent the decompressed content, regardless of the view's offset within the underlying buffer.
> _[View workflow run](https://github.com/microsoft/FluidFramework/actions/runs/25768005820)_

Similar comment was issued at https://github.com/microsoft/FluidFramework/pull/27272#discussion_r3214494946 but overlooks that more is known and offset will always be zero.

tylerbutler (commented):
:shipit:

alexvy86 (approved):
Stamping to unblock. I did my best to review within my bandwidth constraints and lack of expertise with this kind of programmatic use of TS. Nothing caught my eye in particular, but I'd suggest more eyes unless you feel confident enough about it 👍 .

jason-ha:
> Stamping to unblock. I did my best to review within my bandwidth constraints and lack of expertise with this kind of programmatic use of TS. Nothing caught my eye in particular, but I'd suggest more eyes unless you feel confident enough about it 👍 .

Thanks - I feel good with the TS (apart from wishing there was better support). Was able to pull from pattern used in copilot-board.

PR #27274, opened 2026-05-11
Revise Fluid Framework Compatibility Considerations

CraigMacomber:
## Description

Edit Fluid Framework Compatibility Considerations in an attempt to make it a bit more clear about a few things.

## Reviewer Guidance


FluidCompatibilityConsiderations.md
```diff
@@ -2,53 +2,55 @@
...
+    BuildTime[Build Time] --APIs--> TypeScript
+        TypeScript --TypeCheck--> PackageCompat[Package Compatibility]
+
+    InMemory[In Memory] --External APIs/Behaviors--> CustomerUse[Use by customer]
+        CustomerUse --Run Application--> PackageCompat
+    InMemory --Internal APIs/Behaviors--> LayeredArch[Layered architecture]
+        LayeredArch --Interactions between layers--> LayerCompat[Layer compatibility]
+
+    SerializedData[Serialized Data] --Ops--> MultiClient[Multi-client collaboration]
+        MultiClient --Collaboration via ops--> CrossClientCompat[Cross-client compatibility]
```
markfields: Snapshot format changes impact cross-client compat as well.  We cannot write in a new format until the ability to read that format has saturated (aka until it's present across the whole cross-client collab compat window)

FluidCompatibilityConsiderations.md
```diff
@@ -2,53 +2,55 @@
...
+- **Package compatibility** arises because applications depend on Fluid packages using version ranges which guarantee compatibility according to our [API support levels](./docs/docs/build/releases-and-apitags.mdx#api-support-levels) for both Type compatibility and runtime behavior.
 - **Layer compatibility** arises because Fluid's modular design consists of four distinct layers (Driver, Loader, Runtime, and Datastore), each of which can be versioned independently. These layers must interoperate at runtime even when they're at different versions. They interact by calling APIs (mostly internal) on other layers and the signatures and behavior of these APIs must be compatible.
-
-**From Data:**
-
 - **Data-at-rest compatibility** arises because documents (stored as summaries/snapshots) may be dormant for extended periods and then reopened by clients running newer versions of Fluid.
 - **Cross-client compatibility** arises because multiple clients collaborating on the same document in real-time by exchanging ops may be running different versions of Fluid during rolling upgrades or version transitions.
```
markfields, line 33: - nit: can we order these 4 bullets the same order as the mermaid diagram?
- Let's also mention snapshots here

FluidCompatibilityConsiderations.md
```diff
@@ -2,53 +2,55 @@
...
+Package compatibility implies that we cannot break existing APIs within the supported set of versions. For example, if we were to say we support compatibility of public APIs where the major version matches, we can only break them when releasing a major version (with reasonable documentation) but we cannot break them in minor or patch releases.
+See [API support levels](./docs/docs/build/releases-and-apitags.mdx#api-support-levels) for the actual guarantees we make.
 
 ### Motivation
 
-Application developers need a clear, predictable upgrade path to newer Fluid versions. When API changes occur, documented and well-communicated breaking changes allow teams to plan and execute upgrades with confidence.
+This allows users of our packages to use [dependency ranges](https://github.com/npm/node-semver) like "^2.100.0" providing predictable upgrade path to newer Fluid versions, ensuring bug fixes are easy to integrate, and it's clear when breaking changes need to be checked for.
```
markfields: nit
```suggestion
This allows users of our packages to use [dependency ranges](https://github.com/npm/node-semver) like "^2.100.0" providing predictable upgrade path to newer Fluid versions, ensuring bug fixes are easy to integrate, and it's clear when to check for breaking changes.
```

FluidCompatibilityConsiderations.md
```diff
@@ -2,53 +2,55 @@
...
+Layer compatibility implies that a single client can have different versions for different compatibility layers we support - Driver, Loader, Runtime and Datastore. For example, Driver v1.0, Loader v2.0, Runtime v3.0 and both Datastore v3.1 and v3.2 are used on the same client. Multiple Datastore versions can coexist within a single Runtime because each datastore type may come from a separately-versioned package loaded through the runtime's code-loading registry. The APIs at the boundaries of these layers have strict compatibility requirements at _runtime_ (distinct from package and type compatibility, which are about build-time dependencies), to support the full range of versions that may be calling them from another layer.
+
+The APIs between these layers are often internal, and their implementations involve downcasting, so special mechanisms beyond simple type checking are needed to ensure their compatibility.
```
markfields: Downcasting aside, I think the larger way that type checking falls short is that we only build the cross-layer-calling code with all layers at the same version.  We could choose to import and union the types from all supported older versions of the layer abstractions, but don't do that.

About downcasting, can you remind me of a couple examples? Ideally we could avoid this. In any case, that feels more of an in-house concern, not sure consumers of this doc need to think about that.

FluidCompatibilityConsiderations.md
```diff
@@ -2,53 +2,55 @@
...
-        D --Interactions between layers--> E[Layer compatibility]
-
-    F[Data] --Ops--> G[Multi-client collaboration]
-        G --Collaboration via ops--> J[Cross-client compatibility]
-    F --Ops/Snapshots--> I[Persistence]
-        I --Read saved files--> H[Data-at-rest compatibility]
+    BuildTime[Build Time] --APIs--> TypeScript
+        TypeScript --TypeCheck--> PackageCompat[Package Compatibility]
+
+    InMemory[In Memory] --External APIs/Behaviors--> CustomerUse[Use by customer]
```
agarwal-navin: The term "In Memory" seems incomplete compared to the other 2 items here which have a phase (Build, Serialized) and a form (Time, Data).
How about something like "In Memory State" or "Run Time" or "Runtime State".

FluidCompatibilityConsiderations.md
```diff
@@ -2,53 +2,55 @@
...
+        Persistence --Read saved files--> DataAtRestCompat[Data-at-rest compatibility]
 ```
 
-### How Code and Data create Compatibility Dimensions
+### How These Aspects Create Compatibility Dimensions
 
-**From Code:**
-
-- **API compatibility** arises because applications depend on public APIs that are released across versions (including alpha and beta APIs). Applications need a stable, predictable upgrade path as Fluid releases new package versions.
+- **Package compatibility** arises because applications depend on Fluid packages using version ranges which guarantee compatibility according to our [API support levels](./docs/docs/build/releases-and-apitags.mdx#api-support-levels) for both Type compatibility and runtime behavior.
```
agarwal-navin, line 31: Not sure about replacing API with Package - the term APIs and API compatibility is used in other places - even the linked doc talks in terms of APIs.
My recommendation would be to leave this as API - it does include type checks, etc. IMO as those are for APIs. If we want to change this to Package, I would change the other docs to reflect this terminology.

FluidCompatibilityConsiderations.md
CraigMacomber: I added a note about snapshots, though generally I think the requirements from "Data-at-rest compatibility" which included the ability to read data from any current or past version is a super set of the requirements for "Cross-client compatibility" which is just for concurrent versions, which makes calling out such cases unimportant.

FluidCompatibilityConsiderations.md
CraigMacomber: We have cases where for a particular version of a package, the API signature and documentation might be identical, but the private internals could have changed in an incompatible way (which could even simply be that the version of our own packages they use internally was increased). In such cases, I think communicated this in terms of package compatibility makes more sense than API compatibility, as most developers will only include the API signatures, and maybe its docs in what they consider to be the API, while we need to also include the package version since our APIs are not structural and depend on implementation details.

We also want to include compatibility related to transitive dependencies, peer deps etc, which is also a stretch to consider part of API compatibility.

For our customers, what matters is what package versions they can use. They pick versions of our packages to use, not which APIs to use. 

For example if I were to change implementation details of our internal handle types in some specific package version, that might not impact customer visible APIs at all, but the packages using the new implementation only works with specific versions of other packages, and ensuring this compatibility works is done via ensuring legal versions of packages are used together as part of our layering rules via things like package.json  dependency expressions.

FluidCompatibilityConsiderations.md
CraigMacomber: Any time we make assumptions about some value which are not captured by the type, thats effectively down casting. This is done intentionally any time we have a type which has implementation details we don't want to expose, but wish to expose values of to user code which they can pass back into our own code. To be robust, such cases should be `@sealed` to indicate that users can not provide their own implementation as the published API does not specific the entire assumed contract we have internally. This is a perfectly good pattern that we do in a ton of places, like ISharedObject, IChannel, fluid handles, TreeNode etc.

Downcasting is the act making an assumption not reflected in the type (usually by  converting to a more specific type via a cast). The fact that we do it with specific types is important for customers because it means its unsupported for them to implement those types themselves.

Since this document tries to cover why these kinds of compatibility issues exist, I think mentioning down casting makes sense rather than just saying some times have APIs which if you implement them things will not work sometimes.

CraigMacomber: Oh, and I forgot that its more complex than users not being able to implement them, since we actually need to constrain which internal implementations of such types can be mixed, so saying they are sealed and users can't implement them is actually not the relevant part here for layer compat.

CraigMacomber: I have added an example to the document to help clarify.

FluidCompatibilityConsiderations.md
CraigMacomber: Changed to Run Time.

FluidCompatibilityConsiderations.md
markfields: I like the new naming, for the reasons Craig explained

FluidCompatibilityConsiderations.md
markfields: Ok looks good.  I guess this point is more of an internal concern and not worth burdening a consumer with?

> the larger way that type checking falls short is that we only build the cross-layer-calling code with all layers at the same version. We could choose to import and union the types from all supported older versions of the layer abstractions, but don't do that

But wherever our internal-facing docs about properly implementing layer compat are - this point should be added there if it's missing.

markfields (approved):

PR #27277, opened 2026-05-11
Document "PAS Test" leak

CraigMacomber:
## Description

This test leaks memory between iterations when running lots of performance iterations. This documents that issue with it.

## Reviewer Guidance


packages/test/test-end-to-end-tests/src/test/benchmark/PasTable.all.spec.ts
```diff
@@ -45,6 +43,9 @@ describeCompat("PAS Test", "NoCompat", () => {
 			containerRuntimeFactory = new MockContainerRuntimeFactory();
 			matrix = createLocalMatrix("matrix1", dataStoreRuntime);
 
+			// Every iteration of this benchmark will allocate lots of new SharedString instances
+			// The old ones from the previous iteration will no longer be reachable from the matrix after they are replaced,
+			// but are leaked in the container as Fluid's GC won't collect them for a very long time (much longer than this test will run).
```
CraigMacomber, line 48:
```suggestion
			// but are leaked in the container as Fluid's GC won't collect them for a very long time (much longer than this test will run).
			// This behavior is undesired and should be fixed if practical.
```

alexvy86 (approved):

PR #27278, opened 2026-05-11
fix(tree): Optionally use UUIDs for non-finalized identifiers in tree attach summaries

Abe27342:


Abe27342:
~~I haven't fully reviewed this on my end yet, but if someone wants to take a stab before I have feel free. I'm at the point I want CI checks to start running so I can get it in a mergeable (pending approval) state.~~ Oops, wrong PR. Thought this was https://github.com/microsoft/FluidFramework/pull/27281

Abe27342:
Superceded by other PRs - closing

PR #15423, opened 2023-05-03
feat(sequence): interval stickiness

connorskees:
Builds on top of https://github.com/microsoft/FluidFramework/pull/15247 in order to implement interval stickiness

PR #27279, opened 2026-05-11
fix(sequence): preserve original add() call shape in applyStashedOp

anthony-murphy:
## Summary
- `IntervalCollection.applyStashedOp` was reconstructing add/change ops by always attaching the serialized `startSide`/`endSide` to numeric positions. Serialized ops include sides unconditionally for future-proofing, so this turned every pending interval add into a `{ pos, side: Side.Before }` shape — which `assertStickinessEnabled` treats as a stickiness request, throwing `attempted to set interval stickiness without enabling \`intervalStickinessEnabled\` feature flag` whenever a pending state was rehydrated with the flag off.
- The fix preserves the original `add()`/`change()` call shape on replay by dropping the redundant default `Side.Before` for numeric positions before calling `add`/`change` from `applyStashedOp`. `assertStickinessEnabled` (originally introduced in #15423) is unchanged — the gate is still strictly shape-based, the call shape is just now correct.
- Regresses behavior introduced in #19675, which wired `applyStashedOp` to call `add`/`change` directly with `toSequencePlace(value.start, value.startSide)`. `serialize()` has always emitted concrete sides, but no stashed-op test covered the `intervalStickinessEnabled: false` path until now.
- The `?? defaultSide` substitutions at `intervalCollection.ts:1430-1431,1674-1675` are pure literal-centralization — those paths (`ackChangeInterval`, `ackAdd`) don't go through `assertStickinessEnabled`, so they're behaviorally inert and separate from the `applyStashedOp` shape fix; bisect accordingly.
- Adds round-trip coverage that serializes an interval and applies it as a stashed op on a fresh client. Six cases total: add + change × {non-sticky/flag off, non-sticky/flag on, sticky/flag on}.

## Test plan
- [x] `pnpm -F @fluidframework/sequence test` — 1664 passing, including the six new round-trip cases and the previously-flaky 100+ fuzz seeds that broke when an earlier draft mutated frozen op values.


anthony-murphy:
## Deep Review

*Reviewed commit `d376500` on 2026-05-11.*

Diagnosis and fix are correct and well-scoped. The new round-trip suite covers six positive paths (add/change × {flag-off non-sticky, flag-on non-sticky, flag-on sticky}). The one remaining gap is the symmetric **negative**-guard case — a genuinely sticky serialized op replayed with `intervalStickinessEnabled: false` should still throw — flagged inline. ~5–10 lines, reuses the existing scaffolding.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- The `assertStickinessEnabled` gate from #15423 (connorskees) is deliberately strict and shape-based; in review of #15423, Abe27342 pushed for a more lenient value-based form and that suggestion was rejected. This PR preserves the gate's intent by fixing the call shape rather than relaxing the gate.
- The regression was introduced by #19675 (anthony-murphy, 2024-02-21), which wired `applyStashedOp` to call `add`/`change` directly with `toSequencePlace(value.start, value.startSide)`. #19675 itself taught `toSequencePlace` to accept `undefined` sides for exactly this case, but the new `applyStashedOp` call sites in the same PR still passed the serialized side unconditionally — this PR is the missing half.
- `intervalStashedOps.spec.ts` has existed since #18240 (jzaffiro) but every prior case built synthetic op values without `startSide`/`endSide`, exercising only the legacy `?? Side.Before` fallback branch. The new `round-trip via applyStashedOp` describe block is the first true serialize→apply round trip in this file.
- The `defaultSide` constant duplicates merge-tree's `endpointPosAndSide` plain-numeric default (`sequencePlace.ts:73-74`); the link is comment-only. Documented convention since #16310 (Abe27342), so the contract is harder to break silently than it looks, but a `PLAIN_NUMERIC_DEFAULT_SIDE` export from merge-tree or a one-line unit assertion would pin it at compile/test time.
- `intervalCollection.ts` is hot but stable — the recent #24283/#24306/#24829/#24884 series (anthony-murphy) is encapsulation/unification refactor, not bug-fix thrash. The `Side.Before → defaultSide` substitutions at lines 1431, 1435, 1678, 1679 are pure constant rename.
- Suggested reviewers: **jzaffiro** (owns the test file, best fit on the round-trip pattern and the negative-guard extension), **Abe27342** and/or **connorskees** (stickiness-semantics confirmation — designers of #15423 and #16310).

- **Stickiness-semantics sign-off** — confirm the PR does not weaken the `assertStickinessEnabled` contract from #15423. Best fit: Abe27342, connorskees.
- **Test convention call** — jzaffiro on whether the round-trip pattern should be the new convention for stashed-op tests in this file, including the shape of the negative-guard extension.
- **Scope-of-fix design call** — boundary-local canonicalization in `applyStashedOp` vs. centralizing the `Side.Before → drop side` rule inside `toSequencePlace`/`toOptionalSequencePlace`. Defensible either way; the local choice preserves explicit `Side.Before` as a meaningful API input on `add`/`change`.
- **Cross-package coupling polish** — owner judgment on whether to export a shared `PLAIN_NUMERIC_DEFAULT_SIDE` from merge-tree's `sequencePlace.ts` or add a unit assertion pinning `endpointPosAndSide(0, 0).startSide === defaultSide`. The PR is correct without it; this is a future-proofing call.
- **Frozen-op invariant** — the new comment says stashed-op `value` "can be deep-frozen by upstream layers (e.g. test mocks)," but no `Object.freeze`/`deepFreeze` was found in the runtime stashed-op forwarding path. A runtime owner can confirm whether this is a current invariant or forward-looking defensive coding.

</details>

- `96ef245` 2026-05-11 · **8/10** — change-arm test gap flagged; PR-description tweak (call out #19675) suggested

</details>

packages/dds/sequence/src/test/intervalStashedOps.spec.ts
```diff
@@ -167,4 +168,139 @@ describe("Interval Stashed Ops on client ", () => {
...
+
+			const targetCollection = target.getIntervalCollection(label);
+			target["applyStashedOp"]({
+				key: label,
+				type: "act",
+				value: { opName: IntervalOpType.ADD, value: sourceInterval.serialize() },
+			} satisfies IMapOperation);
+
+			const [replayed] = Array.from(targetCollection) as SequenceIntervalClass[];
+			assertIntervalMatches("add", target, replayed, source, sourceInterval);
```
anthony-murphy, line 237: Deep Review: The new `round-trip via applyStashedOp` block covers six positive paths (add/change × {flag-off non-sticky, flag-on non-sticky, flag-on sticky}) but the symmetric **negative**-guard case is missing: a genuinely sticky serialized op (e.g. `endSide: Side.After`) replayed with `intervalStickinessEnabled: false` on the target should still throw `assertStickinessEnabled`'s `UsageError`.

The implementation behaves correctly today — `Side.After !== defaultSide`, so the side passes through and `assertStickinessEnabled` (`intervalCollection.ts:1155-1164`) fires — but no test pins this contract. A future "simplification" of the collapse logic (e.g. "always strip sides on replay") would silently weaken the feature gate without any test failure, which is the same class of regression as the one this PR fixes, in the opposite direction. The PR description argues *"the gate is still strictly shape-based, the call shape is just now correct"* — that claim deserves a test, not just a comment.

Suggested ~5–10 lines, reusing the existing `createSharedString` / `assertAddRoundTrip` scaffolding: serialize a sticky interval (`{ start: "start", end: { pos: 5, side: Side.After } }`) on an `intervalStickinessEnabled: true` source, then `assert.throws` the `applyStashedOp` on an `intervalStickinessEnabled: false` target.

anthony-murphy:
## Deep Review

*Reviewed commit `73b715d` on 2026-05-11.*

Diagnosis and fix are correct, surgical, and well-scoped. The new round-trip suite covers six positive paths (add/change × {flag-off non-sticky, flag-on non-sticky, flag-on sticky}) and the previously regressed `intervalStickinessEnabled: false` rehydration path. Two polish items remain inline: the symmetric negative-guard test (sticky source → flag-off target should still throw) and the `defaultSide` cross-package coupling with merge-tree's `endpointPosAndSide` default.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- The `assertStickinessEnabled` gate from #15423 (connorskees) is deliberately strict and shape-based; in review of #15423, Abe27342 pushed for a more lenient value-based form and that suggestion was rejected. This PR preserves the gate's intent by fixing the call shape rather than relaxing the gate.
- The regression was introduced by #19675 (anthony-murphy, 2024-02-21), which wired `applyStashedOp` to call `add`/`change` directly with `toSequencePlace(value.start, value.startSide)`. #19675 itself taught `toSequencePlace` to accept `undefined` sides for exactly this case, but the new `applyStashedOp` call sites in the same PR still passed the serialized side unconditionally — this PR is the missing half.
- `intervalStashedOps.spec.ts` has existed since #18240 (jzaffiro) but every prior case built synthetic op values without `startSide`/`endSide`, exercising only the legacy `?? Side.Before` fallback branch. The new `round-trip via applyStashedOp` describe block is the first true serialize→apply round trip in this file.
- The `defaultSide` constant duplicates merge-tree's `endpointPosAndSide` plain-numeric default (`sequencePlace.ts:23-24`); the link is comment-only. Documented convention since #16310 (Abe27342), so the contract is harder to break silently than it looks, but a shared exported constant or a one-line unit assertion would pin it at compile/test time.
- `intervalCollection.ts` is hot but stable — the recent #24283/#24306/#24829/#24884 series (anthony-murphy) is encapsulation/unification refactor, not bug-fix thrash. The `Side.Before → defaultSide` substitutions at lines 1431, 1435, 1678, 1679 are pure constant rename. The locals-based shape (computing `startSide`/`endSide` into locals rather than mutating `value`) is load-bearing — it aligns with the deep-freeze constraint and Abe27342's prior #10612 feedback against in-place mutation of serialized intervals.
- Suggested reviewers: **jzaffiro** (owns the test file, best fit on the round-trip pattern and the negative-guard extension), **Abe27342** and/or **connorskees** (stickiness-semantics confirmation — designers of #15423 and #16310).

- **Stickiness-semantics sign-off** — confirm the PR does not weaken the `assertStickinessEnabled` contract from #15423. Best fit: Abe27342, connorskees.
- **Test convention call** — jzaffiro on whether the round-trip pattern should be the new convention for stashed-op tests in this file, including the shape of the negative-guard extension.
- **Scope-of-fix design call** — boundary-local canonicalization in `applyStashedOp` vs. centralizing the `Side.Before → drop side` rule inside `toSequencePlace`/`toOptionalSequencePlace`. Defensible either way; the local choice preserves explicit `Side.Before` as a meaningful API input on `add`/`change`.
- **Cross-package coupling polish** — owner judgment on whether to export a shared `PLAIN_NUMERIC_DEFAULT_SIDE` from merge-tree's `sequencePlace.ts` or add a unit assertion pinning `endpointPosAndSide({ pos: 0 }, { pos: 1 }).startSide === defaultSide`. The PR is correct without it; this is a future-proofing call.
- **Frozen-op invariant** — the new comment says stashed-op `value` "can be deep-frozen by upstream layers (e.g. test mocks)," but no `Object.freeze`/`deepFreeze` was found in the runtime stashed-op forwarding path. A runtime owner can confirm whether this is a current invariant or forward-looking defensive coding.

</details>

- `d376500` 2026-05-11 · **8/10** — negative-guard test gap flagged inline; six-case round-trip suite landed
- `96ef245` 2026-05-11 · **8/10** — change-arm test gap flagged; PR-description tweak (call out #19675) suggested

</details>

packages/dds/sequence/src/intervalCollection.ts
```diff
@@ -124,6 +124,12 @@ function compressInterval(interval: ISerializedInterval): CompressedSerializedIn
 	return base;
 }
 
+// The side that an endpoint resolves to when none is specified — see
+// endpointPosAndSide in merge-tree, which produces this for a plain numeric
+// position. Used to fill in serialized ops written by older clients (no
+// startSide/endSide field) and to detect that a serialized side is redundant.
+const defaultSide = Side.Before;
```
anthony-murphy: Deep Review: `defaultSide = Side.Before` here silently duplicates merge-tree's `endpointPosAndSide` plain-numeric default — the comment anchors the semantics to merge-tree ("the side that an endpoint resolves to when none is specified — see endpointPosAndSide in merge-tree, which produces this for a plain numeric position") but nothing enforces the link. The constant is then used as the equality threshold at lines ~943-965 (`value.startSide === defaultSide ? undefined : value.startSide`) and additionally replaces two pre-existing `?? Side.Before` literals at lines 1431/1435 and 1678/1679, broadening the blast radius beyond `applyStashedOp` to the `addInterval`/`ackChangeInterval` paths.

The merge-tree default has historical contention (#16310 review, Abe27342: *"I would've expected `Side.Before` here to match half-open range expectations by default/our previous behavior"*), so it is exactly the kind of value a future contributor might revisit. Documentation at `merge-tree/src/sequencePlace.ts:23-24` records the convention but does not fail builds — if merge-tree's default ever drifts, the regression returns silently with no compile error or test failure outside this file's round-trip block.

Two options, either resolves it:

- **(a) Shared symbol** — export the implicit-side default from merge-tree's `sequencePlace.ts` as a named constant (e.g. `PLAIN_NUMERIC_DEFAULT_SIDE`) and import it here so both packages share one symbol.
- **(b) Lighter — test pin** — add a one-line assertion in `intervalStashedOps.spec.ts` that `endpointPosAndSide({ pos: 0 }, { pos: 1 }).startSide === defaultSide`. Locks the two together with a test failure if they diverge, no API surface change.

Would you accept (b) as the lightweight resolution, or do you prefer (a)?

packages/dds/sequence/src/test/intervalStashedOps.spec.ts
anthony-murphy: Claude here. Added in 574c0ab — `stickiness gate › sticky serialized op replayed against intervalStickinessEnabled=false throws` in `intervalStashedOps.spec.ts`. Serializes a START-sticky interval (`{ start: "start", end: { pos: 5, side: Side.After } }`) on a flag-on source and asserts that `applyStashedOp` against a flag-off target throws via `/intervalStickinessEnabled/`. Resolving.

packages/dds/sequence/src/intervalCollection.ts
anthony-murphy: Claude here. Went with option (a) per the author. `defaultSide` now lives in `merge-tree/src/sequencePlace.ts` as `@internal`, exported alongside `Side`/`endpointPosAndSide`. Both `endpointPosAndSide` and `normalizePlace` use it in merge-tree; `intervalCollection.ts` imports the same symbol — coupling is now by construction, no drift possible, no separate test pin needed. See 574c0ab. Resolving.

packages/dds/sequence/src/test/intervalStashedOps.spec.ts
anthony-murphy: Deep Review: Resolved by the new "stickiness gate" describe block in this SHA (`intervalStashedOps.spec.ts:287-314`) — sticky serialized op (`endSide: Side.After`) replayed on an `intervalStickinessEnabled: false` target now asserts the `UsageError`, with the comment explicitly calling out the future-simplification risk this guards against. Thanks for adding it.

packages/dds/sequence/src/intervalCollection.ts
anthony-murphy: Deep Review: Withdrawing this concern — on re-reading, `defaultSide` isn't sequence-layer leakage. It's consumed inside merge-tree's own canonical place-resolution helpers (`sequencePlace.ts:79-84` `endpointPosAndSide`, `sequencePlace.ts:100-103` `normalizePlace`), so it's a property of `SequencePlace`'s default-resolution rule (already documented at `sequencePlace.ts:23-24`), not a sequence-only concept. Sequence importing it is a knock-on of already consuming `SequencePlace`/`endpointPosAndSide` from merge-tree. No change needed; please ignore the (a)/(b) question.

packages/dds/merge-tree/src/sequencePlace.ts
```diff
@@ -52,6 +52,16 @@ export enum Side {
...
 
+/**
+ * The {@link Side} that a `SequencePlace` resolves to when no side is
+ * explicitly provided — e.g. a plain numeric position, or `"start"`/`"end"`.
+ * Exported so callers that need to detect "the caller didn't specify a side"
+ * (such as serialized op replay) can compare against this without hardcoding
+ * `Side.Before`.
+ * @internal
+ */
+export const defaultSide = Side.Before;
```
anthony-murphy, line 67: Deep Review: The new JSDoc on `defaultSide` reads *"The Side that a SequencePlace resolves to when no side is explicitly provided — e.g. a plain numeric position, or `"start"`/`"end"`."* The `"start"`/`"end"` half is inaccurate: `normalizePlace` in this same file (unchanged by this PR) maps `"start"` → `{ pos: -1, side: Side.After }` and `"end"` → `{ pos: -1, side: Side.Before }`, so `"start"` does **not** default to `defaultSide` (= `Side.Before`) — it resolves to `Side.After`. `endpointPosAndSide` does coerce `"start"`/`"end"` to `defaultSide`, but that's a quirk of its `startIsPlainEndpoint`/`endIsPlainEndpoint` boolean, not a property of `SequencePlace` semantics.

This matters because `defaultSide` is the very `@internal` surface this PR newly exports — future callers grepping for it will read this doc.

Suggested tightening: drop the `"start"`/`"end"` example, or note explicitly that `"start"`/`"end"` literals are handled by `normalizePlace` with their own conventions (`"start"` → `Side.After`, `"end"` → `Side.Before`) and `defaultSide` only governs the bare-numeric case.

anthony-murphy: Claude here. Good catch — `endpointPosAndSide` coercing `"start"`/`"end"` to `Side.Before` was a coincidence of its plain-endpoint boolean, not a property of `SequencePlace` semantics, and `normalizePlace` is the authoritative source for those literals. Tightened the JSDoc in 642e199 to scope `defaultSide` to bare-numeric only and explicitly note `"start"` → `Side.After`, `"end"` → `Side.Before` via `normalizePlace`. Resolving.

anthony-murphy:
## Deep Review

*Reviewed commit `7aaf43d` on 2026-05-11.*

**Readiness: 10/10 — READY**

Ready for human sign-off. No change in posture from the prior review at `642e199`: all three inline threads remain resolved, the JSDoc on the newly-exported `defaultSide` is scoped to the bare-numeric case, and the PR description carries the literal-centralization bisect note for the `?? defaultSide` ack-path swaps. No actionable items remain for the author.

### Context for Reviewers

- PR completes the original intent of #19675 rather than reversing it. #19675 broadened `toSequencePlace` to accept `undefined` sides (anthony-murphy's own inline on that PR: *"these are reused for apply stashed op, so need to be exported, and support undefined sides"*) but #19675's new `applyStashedOp` call sites still forwarded `value.startSide`/`value.endSide` verbatim. `intervalCollection.ts:945-955` here is the missing half.
- The `assertStickinessEnabled` gate from #15423 (connorskees) is deliberately shape-based — `!(typeof start === "number" && typeof end === "number")`, not a value test. In #15423 review, Abe27342 pushed for a value-based form and it was rejected. The bug existed because `applyStashedOp` was injecting `{ pos, side }` objects on what were originally plain-number calls, flipping the shape. The new negative test (`intervalStashedOps.spec.ts:287-314`) pins that contract against a future "always strip sides on replay" simplification.
- Test gap was structural. `intervalStashedOps.spec.ts` (#18240, jzaffiro) and #19675/#20046 (anthony-murphy) never added an `intervalStickinessEnabled: false` round-trip case — even #19675, which introduced the regression. The new add × change × {flag off / flag on / sticky} matrix plus the negative-guard case is proportionate to that gap.
- `defaultSide` placement in merge-tree (not sequence) follows #21976 (jzaffiro, moved `SequencePlace`/`Side` into merge-tree) and the #22696 `@internal` tightening convention. The symbol is consumed inside merge-tree's own canonical place-resolution helpers (`endpointPosAndSide`, `normalizePlace`), not just by sequence.
- The four `?? defaultSide` substitutions at `intervalCollection.ts:1437-1441,1681-1685` (ack paths) are behaviorally inert — those paths don't go through `assertStickinessEnabled`. Separate from the `applyStashedOp` shape fix; bisect accordingly.
- PR exceeds the mechanical-fix scope in three benign ways: named `defaultSide` constant (vs. inline magic `Side.Before`), locals-based shape (anticipates deep-frozen op values from upstream — an earlier in-place-mutation draft broke 100+ fuzz seeds), and the full two-client `serialize()` → `applyStashedOp` round-trip harness with the negative-gate case.
- Suggested reviewers: **jzaffiro** (owns the test file and `SequencePlace`/`sequencePlace` in merge-tree — #18240, #21976, #22574) and **Abe27342** (stickiness-semantics — #15423, #16310). Optional: **connorskees** if any `defaultSide` JSDoc / `assertStickinessEnabled` question becomes contentious (originated `Side`/`endpointPosAndSide`/`assertStickinessEnabled` in #15423).

- **Stickiness-semantics sign-off** — confirm the PR does not weaken the `assertStickinessEnabled` contract from #15423. Best fit: Abe27342, connorskees.
- **Test convention call** — jzaffiro on whether the new round-trip pattern in `intervalStashedOps.spec.ts` (two `SharedString`s + `MockContainerRuntimeFactory`, real `serialize()` → `applyStashedOp` round-trip) should be the new convention for stashed-op tests in this file.
- **Scope-of-fix design call** — boundary-local canonicalization in `applyStashedOp` vs. a value-based `assertStickinessEnabled` gate. Defensible either way; the local choice preserves explicit `Side.Before` as a meaningful API input on `add`/`change`, and a value-based rewrite would have to reconcile the historical `IntervalStickiness.END` default for bare-numeric endpoints. If the team wants to revisit the gate's layer, file a follow-up issue rather than expand this PR.
- **Frozen-op invariant** — the new comment at `intervalCollection.ts:946-952` says stashed-op `value` *"can be deep-frozen by upstream layers (e.g. test mocks)"*. No `Object.freeze`/`deepFreeze` was found in the runtime stashed-op forwarding path. A runtime owner can confirm whether this is a current invariant or forward-looking defensive coding — either way the locals-based shape is safe.

</details>

- `642e199` 2026-05-11 · **10/10** — both Path-to-Ready items landed; all inline threads resolved
- `e2d91d6` 2026-05-11 · **9/10** — JSDoc accuracy on `defaultSide` flagged; optional PR-description bisect note
- `574c0ab` 2026-05-11 · **9/10** — both prior inline threads resolved; PR-description scope nit remained
- `73b715d` 2026-05-11 · **8/10** — negative-guard test and `defaultSide` cross-package coupling flagged inline
- `d376500` 2026-05-11 · **8/10** — negative-guard test gap flagged inline; six-case round-trip suite landed
- `96ef245` 2026-05-11 · **8/10** — change-arm test gap flagged; PR-description tweak (call out #19675) suggested

</details>

sonalideshpandemsft (approved):

jzaffiro (approved):

PR #27281, opened 2026-05-11
fix(tree): Add option to heal unresolvable identifier IDs at doc load time

Abe27342:
## Description

Behavior:
- `chunkDecoding.readValue` (the `SpecialField.Identifier` path) now wraps the `normalizeToSessionSpace` + `decompress` call in try/catch. On failure, when `healUnresolvableIdsOnDecode` is enabled and a `sharedObjectId` is available, it returns a deterministic stable UUID computed as `uuidv5("${sharedObjectId}|${opSpaceId}", healingNamespace)`. All clients loading the same blob agree on the resulting value, and the existing chunked-forest encode path (also from d43d50d7563) writes the string back out at the next summary so the healing does not need to be re-applied on subsequent loads.

Plumbing:
- `FieldBatchEncodingContext` and `IdDecodingContext` gain `healUnresolvableIdsOnDecode?: boolean` and `sharedObjectId?: string`.
- `ChangeEncodingContext` and `EditManagerEncodingContext` carry the same fields so the flag reaches chunked-forest decode for builds/refreshers inside trunk-commit changes (via `modularChangeCodecV1.decodeDetachedNodes`) in addition to `ForestSummarizer.load`.
- `EditManagerSummarizer` accepts the flag and shared-object id as constructor args.
- `SharedTreeOptionsBeta` exposes `healUnresolvableIdsOnDecode?: boolean` (default `false`) via `configuredSharedTreeBetaLegacy`. Off by default so it does not mask genuine decode failures in documents that aren't actually corrupt.

The `RevisionTagCodec` and `DetachedFieldIndex` v1/v2 codecs are left unchanged: v2 already emits long IDs when summarizing in detached state, so those codecs do not produce unresolvable persisted IDs to heal.

Adds an end-to-end test that reproduces the SharedTree-attaches-to-attached- container scenario and verifies healed loading succeeds when the option is enabled.

## Breaking Changes

N/A -- all behavioral changes are opt-in


packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -126,15 +144,35 @@ export function readValue(
...
-					)
-				: streamValue;
+			try {
+				return idCompressor.decompress(
+					idCompressor.normalizeToSessionSpace(
+						streamValue as OpSpaceCompressedId,
+						idDecodingContext.originatorId,
+					),
+				);
+			} catch (error) {
```
Abe27342: Check here must be more robust. We should only be doing this at summary load time, failures to decode in ops will likely be session-specific and this can make things worse (and definitely unrecoverable)

Abe27342: Plumbing that flag through context caused a bunch more files to be touched. I think it's worth doing so though. And I'd rather keep the parameter required.

Abe27342: Done, also dropped reliance on try/catch safety in id-compressor.

Abe27342:
~~Haven't finished reviewing this myself yet but it's in a state where I want agentic review + PR checks running on it.~~

Made a first pass. OK with the state.

packages/dds/tree/src/core/change-family/changeFamily.ts
```diff
@@ -25,6 +25,29 @@ export interface ChangeEncodingContext {
...
+	 * @remarks
+	 * Used to gate decode-time recovery behavior — for example, healing of
+	 * unresolvable identifier IDs — that should only run when loading a
+	 * (possibly broken) attach-summary blob, never when applying ops.
+	 */
+	readonly isSummary: boolean;
+	/**
+	 * If `true`, identifier values that the local id-compressor cannot resolve
+	 * during decode are healed into deterministic stable UUIDs instead of
+	 * throwing. See `FieldBatchEncodingContext.healUnresolvableIdsOnDecode`.
```
Josmithr: Nit: Can we use a `{@link}` here? Could also put this note in a `@see` block.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -59,13 +60,32 @@ export interface IdDecodingContext {
 	 * The creator of any local Ids to be decoded.
 	 */
 	originatorId: SessionId;
+	/**
+	 * See {@link FieldBatchEncodingContext.isSummary}.
```
Josmithr: Nit: `{@inheritDoc}` instead of `See {@link}` for these?

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -126,15 +146,40 @@ export function readValue(
...
+					idDecodingContext.sharedObjectId !== undefined
+				) {
+					// Documents written before the encode-side fix for non-finalized identifier
+					// values can persist negative op-space IDs that are no
+					// longer resolvable once the originating session's local state has been stripped.
+					// When loading such a summary with the heal-on-decode option on, synthesize a deterministic
+					// stable UUID so all readers of the same blob agree on the resulting value.
+					//
+					// The heal path is intentionally restricted to summary loads — an
+					// unresolvable id encountered while applying an op should still surface as
```
Josmithr: Nit
```suggestion
					// unresolvable ID encountered while applying an op should still surface as
```

packages/dds/tree/src/shared-tree/sharedTree.ts
```diff
@@ -570,7 +575,33 @@ export function getCodecTreeForSharedTreeFormat(
...
+	 * session id and the unresolvable op-space integer, so every reader of the same
+	 * blob (other than the originator) agrees on the resulting in-memory id.
+	 * Healed ids are written back out at the next summary in their stable UUID form,
+	 * so the inconsistency does not need to be re-healed on every load.
+	 *
+	 * Off by default because enabling it for documents that are not actually corrupt
+	 * would mask genuine bugs that otherwise surface as decode failures.
+	 *
+	 * This mitigation is also not perfect as the client that originated the non-finalized
+	 * short ID will not apply the same "healing" and will continue to use the original ID.
```
ChumpChief: What happens in that case?  How important is it for the originating client to no longer be joined?

.changeset/petite-pillows-wave.md
```diff
@@ -0,0 +1,17 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": fix
+---
+Add `SharedTreeOptionsBeta.healUnresolvableIdsOnDecode` to recover documents with corrupted identifier IDs
```
CraigMacomber:
```suggestion

Add SharedTreeOptionsBeta.healUnresolvableIdsOnDecode to recover documents with corrupted identifiers
```

https://github.com/microsoft/FluidFramework/blob/main/.changeset/README.md#formatting and identifier IDs is redundant.

.changeset/petite-pillows-wave.md
```diff
@@ -0,0 +1,17 @@
...
+
+A SharedTree bug resulted in some corrupted documents due to their attach summary persisting identifier-field values in an incorrect format.
```
CraigMacomber: This corruption can still happen if users continue to use older versions of the code, and even after this until we land a fix, so I think this is better:

```suggestion
A SharedTree bug can result in corrupted documents due to their attach summary compressing identifier-field values in way that cannot be uncompressed.
```

.changeset/petite-pillows-wave.md
```diff
@@ -0,0 +1,17 @@
...
+This bug manifested in an error with message "Unknown op space ID.".
```
CraigMacomber:
```suggestion
This bug manifested as remote clients processing the op throwing an error with the message "Unknown op space ID.".
This leaves the tree unable to be loaded or summarized by current and future clients without this workaround.
```

.changeset/petite-pillows-wave.md
```diff
@@ -0,0 +1,17 @@
...
+
+This change adds an option (`healUnresolvableIdsOnDecode`) to `configuredSharedTreeBetaLegacy` which will allow documents affected by this bug to load again when enabled.
+Enabling this option carries some risk, see documentation on the interface itself for more details.
+
+### Who is affected
```
CraigMacomber: https://github.com/microsoft/FluidFramework/blob/main/.changeset/README.md#formatting

```suggestion
#### Who is affected
```

packages/dds/tree/src/shared-tree/sharedTree.ts
Abe27342: Typical decoherence. Nothing meaningful on the Fluid side, so it's pretty much whatever the implications are to the calling application.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
Abe27342: Does inheritdoc work on some version of VSCode I'm not on or do I have the wrong syntax in screenshot below? If so, sure, sounds good. If not, I think leaving it as links is better with the dev tooling we actually have. Links at least render inline and are clickable, whereas inheritdocs do this:

image

I'd get it if this were exposed publicly (I'm sure API reports do a better job at doing reasonable things with that tag), but this is an internal type.

(still thanks for making me check that the link resolved; it did not previously but will with next update).

.changeset/petite-pillows-wave.md
```diff
@@ -0,0 +1,17 @@
...
+
+Only applications whose SharedTree schema uses `identifier` fields, where the SharedTree was attached to a container that was already attached, are at risk.
```
CraigMacomber: I've tweaked this to exclude more cases, and narrow by the more likely to eliminate use cases option first.

```suggestion
Only SharedTrees attached to a container that was already attached can be impacted.
Furthermore, this bug only occurs when the attached tree contains [`identifier`](https://fluidframework.com/docs/api/tree/schemafactory-class#identifier-property) fields which contain implicitly generated default values.
```

.changeset/petite-pillows-wave.md
```diff
@@ -0,0 +1,17 @@
...
+Documents that meet those conditions and now fail to load can be recovered by setting `healUnresolvableIdsOnDecode: true`.
```
CraigMacomber: I think "now" here is confusing (they failed to load before this change):
```suggestion
Documents that meet those conditions and thus fail to load can now be recovered by setting `healUnresolvableIdsOnDecode: true`.
```

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1247,7 +1247,9 @@ export interface SharedTreeOptions extends SharedTreeOptionsBeta, Partial<CodecW
 }
 
 // @beta @input
-export type SharedTreeOptionsBeta = ForestOptions & Partial<CodecWriteOptionsBeta>;
+export interface SharedTreeOptionsBeta extends ForestOptions, Partial<CodecWriteOptionsBeta> {
+    readonly healUnresolvableIdsOnDecode?: boolean;
```
CraigMacomber: We try to avoid abbreviations like id" in our APIs (it could mean identity, it could mean identifier). The identifier field creation APIs spell it out so this should as well. I suspect we failed at this consistency in a few places, but lets stick to it where we can.

packages/dds/tree/src/shared-tree/sharedTree.ts
```diff
@@ -570,7 +575,33 @@ export function getCodecTreeForSharedTreeFormat(
...
  */
-export type SharedTreeOptionsBeta = ForestOptions & Partial<CodecWriteOptionsBeta>;
+export interface SharedTreeOptionsBeta extends ForestOptions, Partial<CodecWriteOptionsBeta> {
+	/**
+	 * When `true`, an attempt to decode a non-finalized compressed ID that the local
+	 * id-compressor cannot resolve produces a deterministic stable UUID (a "healed"
+	 * ID) instead of throwing.
+	 *
+	 * @remarks
+	 * Defaults to `false`.
```
Josmithr: Nit: Rather than putting this in a remarks block, I'd recommend `@defaultValue false`.

.changeset/petite-pillows-wave.md
```diff
@@ -0,0 +1,17 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": fix
+---
+Add SharedTreeOptionsBeta.healUnresolvableIdsOnDecode to recover documents with corrupted identifiers
```
CraigMacomber:
```suggestion
Add SharedTreeOptionsBeta.healUnresolvableIdentifiersOnDecode to recover documents with corrupted identifiers
```

.changeset/petite-pillows-wave.md
```diff
@@ -0,0 +1,17 @@
...
+
+A SharedTree bug can result in corrupted documents due to their attach summary compressing identifier-field values in a way that cannot be uncompressed.
+This bug manifested as remote clients processing the op throwing an error with the message "Unknown op space ID.".
+
+This change adds an option (`healUnresolvableIdsOnDecode`) to `configuredSharedTreeBetaLegacy` which will allow documents affected by this bug to load again when enabled.
```
CraigMacomber:
```suggestion
This change adds an option (`healUnresolvableIdentifiersOnDecode`) to `configuredSharedTreeBetaLegacy` which will allow documents affected by this bug to load again when enabled.
```

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -59,13 +61,32 @@ export interface IdDecodingContext {
...
+	 * See {@link FieldBatchEncodingContext.sharedObjectId}.
+	 */
+	sharedObjectId?: string;
 }
+
+/**
+ * Arbitrary namespace for the "heal an unresolvable identifier into a stable UUID"
+ * path in {@link readValue}. This scheme requires consensus across all clients to function.
+ */
+const healingNamespace = "1b671a64-40d5-491e-99b0-da01ff1f3341";
```
CraigMacomber: Bad AI. Do not copy UUIDs from https://github.com/uuidjs/uuid#uuid-- Get yourself a real freshly generated v4 UUID, and document that you did so.

CraigMacomber: (I recommend googling every UUID you see in a code review to check for this)

packages/dds/tree/src/feature-libraries/chunked-forest/codec/codecs.ts
```diff
@@ -106,6 +106,30 @@ export interface FieldBatchEncodingContext {
...
+	 * Off by default. Used only to recover documents whose attach summary was
+	 * written with non-finalized op-space IDs before the encode-side fix
+	 * shipped. Only takes effect when `isSummary` is also `true`.
+	 * See {@link SharedTreeOptionsBeta.healUnresolvableIdentifiersOnDecode}.
+	 */
+	readonly healUnresolvableIdentifiersOnDecode?: boolean;
+	/**
+	 * The SharedTree's shared-object id, used as input to the deterministic
+	 * UUID derivation when `healUnresolvableIdentifiersOnDecode` triggers. Required
+	 * for that path; ignored otherwise.
```
CraigMacomber, line 130:
```suggestion
	 * for that path; ignored otherwise.
	 * @remarks
	 * This allows us to ensure that multiple attaches,
	 * in the same or different documents, with the same session offsets, get different UUIDs.
```

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
Abe27342: wow, great catch.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
Josmithr: VSCode intellisense doesn't support TSDoc `inheritDoc` syntax, unfortunately. So it ends up being a glorified link in that context. But the generated docs support it.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -126,15 +147,40 @@ export function readValue(
...
 			const idCompressor = idDecodingContext.idCompressor;
-			return typeof streamValue === "number"
-				? idCompressor.decompress(
-						idCompressor.normalizeToSessionSpace(
-							streamValue as OpSpaceCompressedId,
-							idDecodingContext.originatorId,
-						),
-					)
-				: streamValue;
+			if (idDecodingContext.isSummary === true && streamValue < 0) {
```
CraigMacomber, line 157: Looks like `isFinalId` is not package exported, so comparing to 0 is the simplest solution here for now:
```suggestion
			// OpSpaceCompressedIds are negative, and require a session-id to compute their value.
			// Due to a bug, we have some special casing for them (see below).
			// TODO: isFinalId should probably be exported from id-compressor and that could be used to do the narrowing here. 
			if (idDecodingContext.isSummary === true && streamValue < 0) {
```

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -126,15 +147,40 @@ export function readValue(
...
+					// The heal path is intentionally restricted to summary loads — an
+					// unresolvable ID encountered while applying an op should still surface as
+					// an error, since it indicates a real bug rather than a recoverable state.
+					return uuidV5(
+						`${idDecodingContext.sharedObjectId}|${streamValue}`,
+						healingNamespace,
+					);
+				}
+				// See `SharedTreeOptionsBeta.healUnresolvableIdentifiersOnDecode` for details on this error.
+				throw new Error(
```
CraigMacomber, line 177: Its rare to have a use case for an exception that's not a assert or a usage error, but I guess this is one of them. This seems correct, I guess.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -126,15 +147,40 @@ export function readValue(
...
+					"Encountered a non-finalized op-space identifier while loading a summary.",
+				);
+			}
+			return idCompressor.decompress(
+				idCompressor.normalizeToSessionSpace(
+					streamValue as OpSpaceCompressedId,
+					idDecodingContext.originatorId,
```
CraigMacomber, line 184:
```suggestion
					// TODO: due to a bug (See SharedTreeOptionsBeta.healUnresolvableIdentifiersOnDecode),
					// this this is actually the wrong identifier for summaries (current session not the one that encoded it):
					// the "if" above mitigates that case.
					idDecodingContext.originatorId,
```

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -126,15 +147,40 @@ export function readValue(
...
+					// unresolvable ID encountered while applying an op should still surface as
+					// an error, since it indicates a real bug rather than a recoverable state.
+					return uuidV5(
+						`${idDecodingContext.sharedObjectId}|${streamValue}`,
+						healingNamespace,
+					);
+				}
+				// See `SharedTreeOptionsBeta.healUnresolvableIdentifiersOnDecode` for details on this error.
+				throw new Error(
+					"Encountered a non-finalized op-space identifier while loading a summary.",
```
CraigMacomber: This is a user facing string, so we should be helpful, and use less internal specific terms.

```suggestion
					"Summary could not be loaded due incorrectly encoded identifier. See SharedTreeOptionsBeta.healUnresolvableIdentifiersOnDecode for mitigation.",
```

packages/dds/tree/src/shared-tree/sharedTree.ts
```diff
@@ -570,7 +575,32 @@ export function getCodecTreeForSharedTreeFormat(
  * Configuration options for SharedTree.
  * @beta @input
  */
-export type SharedTreeOptionsBeta = ForestOptions & Partial<CodecWriteOptionsBeta>;
+export interface SharedTreeOptionsBeta extends ForestOptions, Partial<CodecWriteOptionsBeta> {
+	/**
+	 * When `true`, an attempt to decode a non-finalized compressed ID that the local
+	 * id-compressor cannot resolve produces a deterministic stable UUID (a "healed"
+	 * ID) instead of throwing.
```
CraigMacomber: Finalization of identifiers, and id-compressors aren't really supposed to be a user facing concepts for most tree users, and these are using facing docs.

```suggestion
	 * When `true`, when an improperly encoded identifier is encountered in a summary, a new identifier will be generated instead of throwing an error.
```

packages/dds/tree/src/shared-tree/sharedTree.ts
```diff
@@ -570,7 +575,32 @@ export function getCodecTreeForSharedTreeFormat(
...
+	 *
+	 * @defaultValue `false`
+	 *
```
CraigMacomber, line 584:
```suggestion
	 * @remarks
```

packages/dds/tree/src/shared-tree/sharedTree.ts
```diff
@@ -570,7 +575,32 @@ export function getCodecTreeForSharedTreeFormat(
...
+	 * The intended use is recovering documents whose attach summary was persisted
```
CraigMacomber, line 586: A lot of this is a bit more internal facing terminology and details that we likely want.

I'd expect something more like we have in the changeset here (including the exact case impacted), and if we want to have details on the implementation involved, that could go in privateRemarks.

packages/dds/tree/src/test/feature-libraries/chunked-forest/codec/chunkDecoding.spec.ts
```diff
@@ -151,6 +155,147 @@ describe("chunkDecoding", () => {
...
+						originatorId,
+						isSummary: true,
+						healUnresolvableIdentifiersOnDecode: true,
+						sharedObjectId: "doc-a",
+					};
+					const result = readValue(makeStream(opSpaceId), SpecialField.Identifier, ctx);
+					assert.equal(typeof result, "string");
+					assert.match(
+						result as string,
+						/^[\da-f]{8}-[\da-f]{4}-[1-5][\da-f]{3}-[89ab][\da-f]{3}-[\da-f]{12}$/i,
```
CraigMacomber: I'd prefer the actual UUID be included here: test says "stable" and we need this ID to be the same across code changes, so it seems like a string constant here would catch more bugs about missing stability that this would.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -59,13 +61,32 @@ export interface IdDecodingContext {
...
+	 */
+	healUnresolvableIdentifiersOnDecode?: boolean;
+	/**
+	 * See {@link FieldBatchEncodingContext.sharedObjectId}.
+	 */
+	sharedObjectId?: string;
 }
+
+/**
+ * Arbitrary namespace for the "heal an unresolvable identifier into a stable UUID"
```
CraigMacomber:
```suggestion
 * Random v4 UUID generated as a namespace for the "heal an unresolvable identifier into a stable UUID"
```

packages/dds/tree/src/test/feature-libraries/chunked-forest/codec/chunkDecoding.spec.ts
Abe27342: Yeah fair, that gives coverage over if someone changes the healing namespace uuid or healing scheme that this doesn't. done.

packages/dds/tree/src/shared-tree/sharedTree.ts
Abe27342: Fair enough. I've moved mention of non-finalized to private remarks.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
Abe27342: ok, not taking this suggestion then since the generated docs shouldn't include this interface.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
Abe27342: This isn't really the right place to put that comment (or at least not the single right place). One spot is in this bit of ForestSummarizer:

image

but there might be some other places that need similar changes (EditManager summary skirts by right now for other reasons but I'd expect some of the plumbing there is wrong too).

Not going to include as part of this PR.

packages/dds/tree/src/test/feature-libraries/chunked-forest/codec/chunkDecoding.spec.ts
```diff
@@ -151,6 +155,144 @@ describe("chunkDecoding", () => {
...
+					const ctx: IdDecodingContext = {
+						idCompressor: testIdCompressor,
+						originatorId,
+						isSummary: true,
+						healUnresolvableIdentifiersOnDecode: true,
+						sharedObjectId: "doc-a",
+					};
+					const result = readValue(makeStream(opSpaceId), SpecialField.Identifier, ctx);
+					assert.equal(typeof result, "string");
+					assert.equal(result as string, "d5d534e7-5e2c-53c3-b26c-9fd81e6fbc37");
```
CraigMacomber: Not important, but I think this can be simplified:

```suggestion
					assert.equal(result, "d5d534e7-5e2c-53c3-b26c-9fd81e6fbc37");
```

Abe27342: I'll fix in follow-up if there's nothing else and things look good otherwise. Want to avoid blocking the release for too long.

packages/dds/tree/src/test/feature-libraries/chunked-forest/codec/chunkDecoding.spec.ts
```diff
@@ -151,6 +155,144 @@ describe("chunkDecoding", () => {
...
+				});
+
+				it("produces the same UUID for the same (sharedObjectId, opSpaceId) inputs", () => {
+					const { opSpaceId, originatorId } = makeUnresolvableOpSpaceId();
+					const heal = (): unknown =>
+						readValue(makeStream(opSpaceId), SpecialField.Identifier, {
+							idCompressor: testIdCompressor,
+							originatorId,
```
CraigMacomber, line 221: We should test that the originator id doesn't impact the results (use different values for it in the two calls) as well, since currently we know this value is wrong (from the current session, and thus different for different clients)

Abe27342: Same--would prefer to fix in follow-up.

Abe27342: Checks failed--fixed in latest.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
CraigMacomber: There are multiple places where we need to do follow-up work, and this is one of them since this code will have to change to tolerate not having a session id here.

Personally I think having a TODO here makes since this code can be wrong if it runs (that's why we need the if above) and will have to change in the future.

I think having more TODO's and comments about places where code has know to be wrong behaviors and thus doesn't perform as documented is better than having too few.

CraigMacomber (approved):

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
Abe27342: from what I can tell we don't actually necessarily need to touch this code or context object when we fix up our write format to include originator's session ID where we need it (I think at each summary blob suffices). We just need to supply the correct value at the call site.

I recognize it's probably dependent on exactly what your plans are in the area though.

ChumpChief (approved):

PR #27282, opened 2026-05-11
Promote onAssertionFailure to beta

brrichards:
# Summary
Promotes `onAssertionFailure` from `@alpha` to `@beta`

`onAssertionFailure` lets apps register a first chance callback that fires immediately before an assertion failure throws.

The API shape is unchanged only the release tag is updated. The comment in `fluid-framework/src/index.ts` is updated to match, and the affected API report files are regenerated.

## Reviewer Guidance


packages/common/core-utils/src/assert.ts
brrichards:
```suggestion
 * import { onAssertionFailure } from "fluid-framework/beta";
```

CraigMacomber (approved):
Docs look good to me, and code changes look correct, but still needs API review as this adds a beta API.

Josmithr (approved):

PR #27283, opened 2026-05-11
feat(build-infra): add merge-base-aware package change helpers

tylerbutler:
## Summary

Adds three new helpers to `@fluid-tools/build-infrastructure` for attributing changed files to workspace packages across two git refs, and extends `getMergeBaseRemote` to handle shallow clones.

### New helpers

- **`listPackageJsonPaths(git, ref?)`** — enumerates tracked `package.json` paths. Uses `git ls-tree -r --name-only <ref>` when `ref` is supplied, otherwise `git ls-files` against the current working tree.
- **`getPackageDirsAtRef(git, ref?)`** — returns the `Set<string>` of repo-relative directories that contain a `package.json` at the given ref (or HEAD).
- **`isFileInPackageDir(file, packageDirs)`** — tests whether a file lives inside any directory in `packageDirs` (or any nested subdirectory of one) by walking POSIX-style ancestors. Empty strings and the repo-root pseudo-dir (`.`) are handled correctly.

### `getMergeBaseRemote` updates

- When the initial `merge-base` lookup fails in a **shallow repository**, the clone is now deepened (`git fetch --deepen 1000 <remote> <branch>`) and the lookup retried once.
- New optional `onDeepen?: (message: string) => void` callback lets callers surface a status message when the retry path is taken.
- The returned ref is now trimmed (previously included a trailing newline). The one internal caller (`getChangedFilesSinceRef`) passes the result to `git.diff`, which tolerates either form, so this is a safe behavior tightening.

## Motivation

These helpers exist so callers (e.g. a CI changed-package scoping command) can correctly attribute changes to packages that were **added, removed, or moved** between a merge base and HEAD — something the existing `getChangedSinceRef` cannot do alone because it only consults the current working tree via `readPkgUp`.

The shallow-clone enhancement to `getMergeBaseRemote` makes the helper usable in Azure DevOps PR builds, which typically check out shallow clones.

## Tests

- Adds 6 unit tests covering `isFileInPackageDir` (known-dir hit, deep nested walk, root-only changes, unrelated sibling, empty input, root pseudo-dir non-match).
- All existing build-infrastructure tests continue to pass (58 total).


build-tools/packages/build-infrastructure/src/test/git.test.ts
```diff
@@ -161,3 +170,228 @@ describe("getFiles", () => {
...
+	});
+
+	it("rethrows the original error when the repo is not shallow", async () => {
+		const original = new Error("fatal: Not a valid object name");
+		const mock = makeGitMock({
+			rawResponses: [
+				() => {
+					throw original;
+				},
+				() => "false\n",
```
alexvy86: Nit: would love to always see comments with the calls that each response corresponds to, like in the test above and some other calls below, as well as `expect()` on the args (again, like above) so it's easier to validate if the expected git calls are the ones happening.

build-tools/packages/build-infrastructure/src/git.ts
```diff
@@ -59,25 +59,51 @@ export function findGitRootSync(cwd = process.cwd()): string {
...
+		const base = await git.raw("merge-base", compareRef, localRef);
+		return base.trim();
+	} catch (error) {
+		// Don't let a failure of the secondary `rev-parse` call mask the original `merge-base` error.
+		let isShallow: string;
+		try {
+			isShallow = (await git.raw("rev-parse", "--is-shallow-repository")).trim();
+		} catch {
+			throw error;
+		}
```
alexvy86: Redundant try-catch can be removed.

build-tools/packages/build-infrastructure/src/git.ts
```diff
@@ -59,25 +59,51 @@ export function findGitRootSync(cwd = process.cwd()): string {
...
+		if (isShallow !== "true" || remote === undefined) {
+			throw error;
+		}
```
alexvy86: Worth it to add logging if there's no remote, to ease troubleshooting?

build-tools/packages/build-infrastructure/src/git.ts
```diff
@@ -121,6 +147,77 @@ function filePathsToDirectories(files: string[]): string[] {
 	return [...dirs];
 }
 
+/**
+ * Matches paths that end in `package.json` (either at the root or under a directory).
+ */
```
alexvy86, line 163:
```suggestion
/**
 * Matches paths that end in `package.json` (either at the root or under a directory).
 * @remarks It assumes paths with POSIX separators (forward slashes), like the
 * ones returned by `git ls-files` and `git ls-tree` (even in Windows).
 * Be careful if using this to test paths not generated by git calls.
 */
```

build-tools/packages/build-infrastructure/src/git.ts
```diff
@@ -121,6 +147,77 @@ function filePathsToDirectories(files: string[]): string[] {
...
+ *
+ * Useful for attributing changed files to packages when the set of packages may have changed
+ * between two refs (e.g. packages added, removed, or moved between a merge base and HEAD).
+ *
+ * A `package.json` at the repo root produces a `"."` entry from `path.posix.dirname`, which is
+ * intentionally filtered out: a root-level package would otherwise match every file in the
+ * repo, defeating the purpose of per-package attribution. Callers that need to attribute
+ * changes to a root-level package should handle that case separately.
+ */
+export async function getPackageDirsAtRef(git: SimpleGit, ref?: string): Promise<Set<string>> {
```
alexvy86: Nit. If the parameter is optional, it shouldn't be part of the function name IMO.
```suggestion
export async function getPackageDirs(git: SimpleGit, ref?: string): Promise<Set<string>> {
```

build-tools/packages/build-infrastructure/src/git.ts
```diff
@@ -121,6 +147,77 @@ function filePathsToDirectories(files: string[]): string[] {
...
+}
+
+/**
+ * Returns the set of repo-relative directories that contain a `package.json` at the given ref
+ * (or in the current working tree when `ref` is omitted).
+ *
+ * Useful for attributing changed files to packages when the set of packages may have changed
+ * between two refs (e.g. packages added, removed, or moved between a merge base and HEAD).
+ *
+ * A `package.json` at the repo root produces a `"."` entry from `path.posix.dirname`, which is
```
alexvy86, line 210: Do we need similar handling for the `package.json` files at the root of the other release groups like server?

build-tools/packages/build-infrastructure/src/git.ts
```diff
@@ -121,6 +154,81 @@ function filePathsToDirectories(files: string[]): string[] {
...
+ * current working tree.
+ *
+ * @privateRemarks
+ * Both branches list every tracked file and then filter to `package.json` entries on the JS
+ * side. We deliberately do not pass a pathspec to `git ls-files`, because pathspec wildcards
+ * use `FNM_PATHNAME` semantics and would not match nested `package.json` files. Filtering
+ * here keeps both code paths symmetric and uses a single source of truth
+ * (`packageJsonPathPattern`) for what counts as a package manifest.
+ */
+export async function listPackageJsonPaths(git: SimpleGit, ref?: string): Promise<string[]> {
```
jason-ha, line 193: What is the intended behavior for package.json's in various deleted states when no ref is supplied?

build-tools/packages/build-infrastructure/src/test/git.test.ts
```diff
@@ -161,3 +170,250 @@ describe("getFiles", () => {
 		);
 	});
 });
+
+describe("isFileInPackageDir", () => {
```
jason-ha, line 174: Worth having a test where "file" is a directory? Where "file" is exactly in "packageDirs"?

build-tools/packages/build-infrastructure/src/test/git.test.ts
```diff
@@ -396,6 +396,47 @@ describe("listPackageJsonPaths", () => {
 	});
 });
 
+describe("listPackageJsonPaths: staged deletion (local)", () => {
+	const git = simpleGit(process.cwd());
+	// A package.json that exists in the test repo and can be safely removed/restored in a test.
+	const targetPkgAbs = path.join(testRepoRoot, "packages/group3/pkg-g/package.json");
+	// `git ls-files`/`ls-tree` invoked from a subdirectory scope paths to that subdirectory, so
+	// compute the expected path relative to the process cwd (where simpleGit was rooted).
+	const targetPkgRel = path.relative(process.cwd(), targetPkgAbs).split(path.sep).join("/");
```
tylerbutler: can't you just initialize simpleGit at testRepoRoot?

build-tools/packages/build-infrastructure/src/git.ts
tylerbutler: This is now documented and backed up by tests (assuming I wrote them correctly :smile:)

build-tools/packages/build-infrastructure/src/test/git.test.ts
tylerbutler: Done in latest.

jason-ha (approved):

PR #27284, opened 2026-05-11
refactor(build-cli): centralize Azure Pipelines logging commands

tylerbutler:
## Summary

Adds a small `library/azureDevops/pipelineCommands.ts` module with two helpers that build the `##vso[...]` strings recognized by Azure Pipelines as [logging commands](https://learn.microsoft.com/azure/devops/pipelines/scripts/logging-commands):

- **`formatSetVariable(name, value, { isOutput })`** — builds `##vso[task.setvariable variable=NAME[;isOutput=true]]VALUE`.
- **`formatLogIssue(level, message)`** — builds `##vso[task.logissue type=warning|error]MESSAGE`.

The helpers only format the strings; emitting them is still the caller's responsibility (typically `this.log(...)` from a command).

## Callers refactored

Three existing commands had inline `##vso[...]` template literals duplicated across multiple call sites:

- [`packages/build-cli/src/commands/check/latestVersions.ts`](packages/build-cli/src/commands/check/latestVersions.ts)
- [`packages/build-cli/src/commands/vnext/check/latestVersions.ts`](packages/build-cli/src/commands/vnext/check/latestVersions.ts)
- [`packages/build-cli/src/commands/generate/buildVersion.ts`](packages/build-cli/src/commands/generate/buildVersion.ts)

All three now use the helpers.

## Casing normalization

The codebase previously had a mix of `isOutput=true` and `isoutput=true` in different call sites. The helper emits `isOutput=true` consistently, which matches the casing in the [Azure DevOps docs](https://learn.microsoft.com/azure/devops/pipelines/scripts/logging-commands#setvariable). Azure DevOps is case-insensitive on this attribute, so **runtime behavior is unchanged**. The vnext `latestVersions` test assertions were updated accordingly.

## Motivation

A follow-up PR introduces a new CI scoping command that also needs to emit pipeline commands. Centralizing the format here keeps that PR small and avoids further duplication of the literal `##vso[...]` strings.

## Tests

All existing build-cli tests pass (495 total).

### DIFF
--- a/build-tools/packages/build-cli/src/commands/check/latestVersions.ts
+++ b/build-tools/packages/build-cli/src/commands/check/latestVersions.ts
@@ -4,7 +4,7 @@
  */
 
 import { findPackageOrReleaseGroup, packageOrReleaseGroupArg, semverArg } from "../../args.js";
-import { formatSetVariable } from "../../library/azureDevops/pipelineCommands.js";
+import { generateSetVariableString } from "../../library/azureDevops/pipelineCommands.js";
 import { BaseCommand } from "../../library/commands/base.js";
 import { isLatestInMajor } from "../../library/latestVersions.js";
 
@@ -53,9 +53,11 @@ export default class LatestVersionsCommand extends BaseCommand<typeof LatestVers
 			this.log(
 				`Version ${versionInput.version} is the latest version for major version ${result.majorVersion}`,
 			);
-			this.log(formatSetVariable("shouldDeploy", "true", { isOutput: true }));
+			this.log(generateSetVariableString("shouldDeploy", "true", { isOutput: true }));
 			this.log(
-				formatSetVariable("majorVersion", String(result.majorVersion), { isOutput: true }),
+				generateSetVariableString("majorVersion", String(result.majorVersion), {
+					isOutput: true,
+				}),
 			);
 			return;
 		}
@@ -64,9 +66,11 @@ export default class LatestVersionsCommand extends BaseCommand<typeof LatestVers
 			this.log(
 				`##[warning]skipping deployment stage. input version ${versionInput.version} does not match the latest version ${result.latestVersion}`,
 			);
-			this.log(formatSetVariable("shouldDeploy", "false", { isOutput: true }));
+			this.log(generateSetVariableString("shouldDeploy", "false", { isOutput: true }));
 			this.log(
-				formatSetVariable("majorVersion", String(result.majorVersion), { isOutput: true }),
+				generateSetVariableString("majorVersion", String(result.majorVersion), {
+					isOutput: true,
+				}),
 			);
 			return;
 		}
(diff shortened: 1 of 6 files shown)


build-tools/packages/build-cli/src/library/azureDevops/pipelineCommands.ts
```diff
@@ -0,0 +1,79 @@
...
+	isOutput?: boolean;
+}
+
+/**
+ * Formats a `##vso[task.setvariable ...]` pipeline command that sets a pipeline variable.
+ *
+ * Reserved characters in `name` and `value` are escaped per Azure Pipelines logging command
+ * rules so that values containing `;`, `]`, `%`, or newlines do not corrupt the command.
+ */
+export function formatSetVariable(
```
alexvy86: Nit: personally I'd call these `generateBlaString`. When I read `formatBla` my mind doesn't immediately think of what this is doing. But if others think it's fine, it's not a blocker.

build-tools/packages/build-cli/src/test/library/azureDevops/pipelineCommands.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+		it("includes isOutput=true when requested", () => {
+			assert.equal(
+				formatSetVariable("myVar", "hello", { isOutput: true }),
+				"##vso[task.setvariable variable=myVar;isOutput=true]hello",
+			);
+		});
+
+		it("escapes reserved characters in the value", () => {
+			// `;` and `]` and `%` and CR/LF must be escaped in the data portion.
```
alexvy86:
```suggestion
			// `%` and CR/LF (but not `;` and `]`) must be escaped in the data portion.
```

build-tools/packages/build-cli/src/commands/check/latestVersions.ts
```diff
@@ -52,9 +53,11 @@ export default class LatestVersionsCommand extends BaseCommand<typeof LatestVers
 			this.log(
 				`Version ${versionInput.version} is the latest version for major version ${result.majorVersion}`,
 			);
-			this.log(`##vso[task.setvariable variable=shouldDeploy;isoutput=true]true`);
+			this.log(generateSetVariableString("shouldDeploy", "true", { isOutput: true }));
 			this.log(
-				`##vso[task.setvariable variable=majorVersion;isoutput=true]${result.majorVersion}`,
+				generateSetVariableString("majorVersion", String(result.majorVersion), {
```
jason-ha: not blocking: I feel like we don't see a lot of `String` use. Don't know if there is a reason for that.
Certainly looks functional to me.

build-tools/packages/build-cli/src/commands/check/latestVersions.ts
```diff
@@ -63,19 +66,23 @@ export default class LatestVersionsCommand extends BaseCommand<typeof LatestVers
 			this.log(
 				`##[warning]skipping deployment stage. input version ${versionInput.version} does not match the latest version ${result.latestVersion}`,
```
jason-ha: not blocking: Any plans for `##[warning]`?

jason-ha (approved):
LGTM

build-tools/packages/build-cli/src/commands/check/latestVersions.ts
tylerbutler: Cleaned up by allowing additional types in the function args. Much cleaner now IMO.

build-tools/packages/build-cli/src/commands/check/latestVersions.ts
tylerbutler: No, I missed this. Good catch - added in latest.

jason-ha (approved):

PR #27285, opened 2026-05-12
edit lock for transactions in "changed" event callbacks

daesunp:
## Description

Edit lock for edits done in `changed` event callbacks.


.changeset/tough-lies-say.md
```diff
@@ -0,0 +1,43 @@
...
+"__section": tree
+---
+Starting a transaction inside a tree-change event listener now produces a usage error
+
+Calling `runTransaction` or `runTransactionAsync` from inside a `nodeChanged` or `treeChanged` event listener now throws a `UsageError`.
+This brings transactions into line with the existing rule that tree mutations from inside a change-event listener are forbidden.
+
+Before this change, starting a transaction from a listener would push a transaction frame onto the running outer transaction's bookkeeping before the inner edit was attempted,
+leaving the tree's transaction labels in a corrupted state even when the inner edit itself threw.
+The new check rejects the call before any state mutation, so the tree's transaction labels remain consistent.
```
CraigMacomber: Doesn't this leave the tree in a broken state, so it's impossible for it to matter if the transaction labels are consistent or not?

Might be better to rephase this to something like:

```suggestion
The updated code throws a `UsageError` before it can corrupt anything.
```

.changeset/tough-lies-say.md
```diff
@@ -0,0 +1,43 @@
...
+```typescript
+// Before — throws at runtime
+Tree.on(view.root, "nodeChanged", () => {
+	view.runTransaction(() => {
+		// ...
+	});
+});
+
+// After — schedule the transaction outside the listener
+Tree.on(view.root, "nodeChanged", () => {
```
CraigMacomber: I don't think we should recommend this approach to people. Its bad for a few reasons:
1. If you do edits in response to document changes, its very easy to end up with edits triggering edits in a loop (on one client or between multiple clients) resulting in infinite edits, consuming unbounded bandwidth, storage, battery life CPU etc.
2. Its easy to end up with multiple edits, one from each client connected, fighting over making the change (even if all but the first are no-ops due to merge resolution, its inefficient)
3. It can lead to screwy attribution as the document show up last changed by a user who did not personally make the change
4. It can screw up undo/redo, as it creates a change by the current user, that was not trigger by any of that user's actual actions.
5. Its async, so its possible the edit trying to be done is no longer valid (a change could have happened in-between), the view could have been disposed, etc. If you need to do a change asynchronously, one should use a branch instead of holding onto the view or nodes async (unless they setup invalidation for the values they are holding onto which this example does not do).

I think we can simply say applications should not make edits in response to edits.

.changeset/tough-lies-say.md
```diff
@@ -0,0 +1,43 @@
...
+The error message follows the existing format used by the other listener-time guards:
+
+```
+Running a transaction is forbidden during a nodeChanged or treeChanged event
+```
+
+After this error is thrown, the affected tree view enters a broken state and subsequent operations on it will throw an `Invalid use of ... after it was put into an invalid state` error.
+Discard the view and acquire a fresh one to recover.
+
+#### Migration
```
CraigMacomber: What actually worked before? Could you make transactions, but they would crash if there were any edits in them (in which case they could only be no-ops), or could you actually slip transactions with real edits inside of them thorough our validation before this? If only the no-op case worked, I don't think there is anything to migrate, simply don't make no-op transactions inside events.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -888,6 +888,11 @@ export class TreeCheckout implements ITreeCheckout {
 
 	private mountTransaction(params: RunTransactionParams | undefined, isAsync: boolean): void {
 		this.checkNotDisposed();
+		// Starting a transaction during a nodeChanged/treeChanged event is forbidden for the same
+		// reason direct edits are: it would push a label frame and (typically) commit edits while
+		// the tree's invariants are mid-flight. Without this guard, nested transactions started
+		// from a listener leak label frames into the running outer transaction's label tree
```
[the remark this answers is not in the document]
daesunp: updated with suggestion

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
```diff
@@ -1357,6 +1357,87 @@ describe("sharedTreeView", () => {
...
+			// Without the guard, a runTransaction call from inside a listener pushes a label
+			// frame and starts a nested transaction before the inner edit's lock check fires,
+			// corrupting the outer transaction's label tree. The guard rejects at mount time.
+			expectErrorDuringEdit({
+				duringEdit: (view) =>
+					view.runTransaction(() => {
+						view.root.number = 4;
+					}),
+				error: "Running a transaction is forbidden during a nodeChanged or treeChanged event",
+			});
```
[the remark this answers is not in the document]
daesunp: Updated to use helper that has validateUsageError

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
```diff
@@ -1357,6 +1357,87 @@ describe("sharedTreeView", () => {
...
+				"direct edit after listener throw should be blocked",
+			);
+			assert.throws(
+				() =>
+					view.runTransaction(() => {
+						view.root.number = 6;
+					}),
+				validateUsageError(/invalid state/),
+				"runTransaction after listener throw should be blocked",
+			);
```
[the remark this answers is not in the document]
daesunp: Updated wording to be `/Invalid use of .* after it was put into an invalid state/`

.changeset/tough-lies-say.md
```diff
@@ -0,0 +1,43 @@
+---
+"@fluidframework/tree": minor
+"__section": tree
+---
+Starting a transaction inside a tree-change event listener now produces a usage error
+
+Calling `runTransaction` or `runTransactionAsync` from inside a `nodeChanged` or `treeChanged` event listener now throws a `UsageError`.
```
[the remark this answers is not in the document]
daesunp: reworded to "Calling runTransaction from inside a nodeChanged or treeChanged event listener now throws a UsageError, and runTransactionAsync rejects with one."

.changeset/tough-lies-say.md
daesunp: Updated with suggestion

.changeset/tough-lies-say.md
daesunp: Removed recommendation in changeset

.changeset/tough-lies-say.md
```diff
@@ -0,0 +1,29 @@
+---
+"@fluidframework/tree": minor
```
CraigMacomber, line 2: most of the tree APIs are reexported via the fluid-framework package, which is wat we encourage new users to adopt, so almost all changeset which apply to tree should include it as well.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -486,6 +486,18 @@ export class TreeCheckout implements ITreeCheckout {
...
+	 * True while a `changed` event is being emitted from {@link TreeCheckout.onAfterBranchChange}.
+	 * Used by {@link TreeCheckout.mountTransaction} to forbid re-entrant transactions started
+	 * from `changed` listeners — those would push label frames onto the outer transaction's
+	 * (not-yet-cleared) label tree, corrupting it.
+	 *
+	 * @remarks
+	 * The `editLock` only covers `nodeChanged`/`treeChanged` emissions inside `onAfterChange`.
+	 * `changed` fires from `onAfterBranchChange` outside that window, so a separate flag is needed.
+	 */
+	private isEmittingChangedEvent: boolean = false;
```
CraigMacomber: rather than introducing a new mechanism, I think it would make more sense to use the existing editLock. We should be able to lock the edit lock when applying changes (maybe in applySerializedChange? I'm not what the right spot is) and hold it until all our events are done. Maybe the way its used is insufficient to meet our needs: in such a case I'd rather see it fixed/improved that making a new system in parallel for the same purpose.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -791,16 +803,26 @@ export class TreeCheckout implements ITreeCheckout {
 					labels: buildLabelsSet(this.labelTreeNode),
 				};
 
-				this.#events.emit("changed", metadata, getRevertible);
+				this.isEmittingChangedEvent = true;
```
CraigMacomber: because your not reusing the editLock, and instead making your own thing, I think this failes to handle reentrancy (if isEmittingChangedEvent  is already true, no error) and non-transaction edits won't get blocked

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -791,16 +803,26 @@ export class TreeCheckout implements ITreeCheckout {
...
 				withinEventContext = false;
 			}
 		} else if (this.isRemoteChangeEvent(event)) {
 			// TODO: figure out how to plumb through commit kind info for remote changes
-			this.#events.emit("changed", {
-				isLocal: false,
-				kind: CommitKind.Default,
-				labels: new Set<unknown>(),
-			});
+			this.isEmittingChangedEvent = true;
```
CraigMacomber: if we need to do this samel logic around all cases which trigger the change event, might be better to factor out a private method rathewr than duplicate it.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -911,6 +933,19 @@ export class TreeCheckout implements ITreeCheckout {
...
+		// reason direct edits are: it would push a label frame and (typically) commit edits while
+		// the tree's invariants are mid-flight. Without this guard, nested transactions started
+		// from a listener leak label frames into the running outer transaction's label tree.
+		this.editLock.checkUnlocked("Running a transaction");
+		// Same concern for `changed` event listeners: `labelTreeNode` is still set when the
+		// `changed` event fires (the cleanup in runWithTransactionLabel's finally block runs
+		// after the emit), so a re-entrant runTransaction would push its label frame onto the
+		// outer's still-live label tree. `editLock` doesn't cover this path — it's only engaged
+		// around `nodeChanged`/`treeChanged` emissions in onAfterChange — so use a separate flag.
+		if (this.isEmittingChangedEvent) {
```
CraigMacomber: There are several other places which use editLock.checkUnlocked to check we are not inside of an event. We don't want to just add a bunch of logic and docs to this one. Again this is a reason to not introduce a separate mechanism. We don't want people doing non-transactional edits inside changed events either.

I think part of the issue here is you are fixing two issues:
1. Our validation for not editing in event callbacks is missing one of the callbacks.
2. Our validation for not editing in event callbacks doesn't trigger when starting a transaction.

These should be independent issues, with independent fixes and tests. Instead it seems like you added a new mechanism, which means there are more combinations of cases to handle then necessary, then forgot to handle one of them: when doing non transaction edits inside the changed event.

Given the confusion here, I think it would be better to have a separate PR for each of these.

.changeset/tough-lies-say.md
daesunp: I noticed that it was actually the `changed` event callback that had the biggest issue (which is now included in this PR), and for `treeChanged/nodeChanged` it was just no-op case that was silently passing. Updated the changeset to be more clear

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Updated with suggestion. Dropped the flag, now hold the existing editLock across the emit.

.changeset/tough-lies-say.md
daesunp: updated with `"fluid-framework": minor`

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Updated with suggestion. Dropped the flag, now hold the existing editLock across the emit.

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: factored out emitChangedLocked() helper

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Updated this PR to only fix issue 1. Will make a follow up PR for issue 2.

.changeset/tough-lies-say.md
```diff
@@ -0,0 +1,21 @@
...
+"fluid-framework": minor
+"__section": tree
+---
+The edit lock now also covers the `changed` event
+
+The existing edit-time lock that forbids tree mutations from inside a `nodeChanged` or `treeChanged` event listener now also covers the `changed` event.
```
CraigMacomber: What is this "changed" event on? Can you link the actual API docs for the event here?

.changeset/tough-lies-say.md
```diff
@@ -0,0 +1,21 @@
...
+Direct edits, branch operations, reverts, etc. attempted from inside a `changed` listener now throw the same canonical `UsageError` they already throw from the other change-event listeners.
+
+Previously the lock did not extend to `changed` emissions, so edits made from a `changed` listener ran to completion: they applied to the tree, their commits fired further `changed` events, and the listener could re-enter for those commits.
+This pattern could produce infinite edit loops, redundant work across clients, incorrect attribution, broken undo/redo grouping, and pollution of the outer commit's label data.
+
+The error message now names all three events:
+
+```
+Editing the tree is forbidden during a nodeChanged, treeChanged, or changed event
```
CraigMacomber: Again I don't thus its very clear which event "changed" is referring to. Tons of things could have a changed event, and many of them (such as events on UI content, ddetached nodes, other trees) are fine to edit the tree in (at least some of the time). There are also likely plenty more events this should apply to (like schema changes)

I think we want an error more like:

```suggestion
Editing a SharedTree is forbidden during its event callbacks for changes.
```

.changeset/tough-lies-say.md
```diff
@@ -0,0 +1,21 @@
...
+```
+
+Applications should not make edits in response to edits.
+If a derived edit is genuinely needed, drive it from a user action or apply it on a `TreeBranchAlpha` and merge from there.
```
CraigMacomber: I don't think this sentence makes a lot of sense. Using TreeBranchAlpha, which includes the main view which you presumable just got an event for isn't a solution. You could say make the event on a different branch, and drive merging that from a user action, but that's still likely to duplicate edits across all users and thus not a pattern we have good support for, and is also already covered by the suggestion to drive edits from user actions.


```suggestion
Generally edits should not be made in response to changes to the document: doing so tends to duplicate edits across multiple clients cause infinite edit loops, dirty last change attribution and be problematic in read only contexts.
Instead use alternative patterns like lazily fixing content when edited by a user, or asking a user if they want to repair the document.
If you are confident the desired change is atomic (even accounting for different application versions which might be involved) and/or won't trigger issues with other clients you can work around this restriction by asynchronously deferring the edit, but this pattern is not recommended.
```

Ideally though I'd like to find a place in our documentation covering this, which we can link to instead of just putting it here where it won't be seen much in the future.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -791,19 +791,41 @@ export class TreeCheckout implements ITreeCheckout {
...
+	 * @remarks
+	 * Shared by both the local and remote `changed` emission paths in {@link TreeCheckout.onAfterBranchChange}.
+	 * The `try`/`finally` ensures the lock is released even if a listener throws.
+	 */
+	private emitChangedLocked(emit: () => void): void {
+		this.editLock.lock();
+		try {
+			emit();
+		} finally {
+			this.editLock.unlock();
```
CraigMacomber, line 828: Since we don't have a properly scoped breakable we can use here, and this is really just an existing issue with EditLock, I guess we can just leave a TODO to clean this up.

```suggestion
			// TODO: any event that throws potentially leaves the code which triggered that event,
			// and thus this checkout (and likley more) in a broken state.
			// Unlocking this editLock prevents future use of this broken state from giving a confusing error in this case,
			// however, a better approach would probably be to put something (this checkout and/or the editLock)
			// into a broken state (using a properly scoped `Breakable`),
			// likely by moving emitChangedLocked into EditLock, and having EditLock get a Breakable,
			// and having the new emitChangedLocked use `Breakable.use`.
			this.editLock.unlock();
```

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
```diff
@@ -1341,6 +1341,27 @@ describe("SchematizingSimpleTreeView", () => {
...
 		});
 
+		it("direct (non-transaction) edit from a changed listener throws", () => {
+			// Engaging `editLock` around `changed` emissions means the lock check on the editor
+			// proxy also fires for direct edits — not just transactions. Without this, a consumer
+			// could bypass the transaction-time check by writing tree state directly inside a
+			// `changed` listener, with the same corruption potential.
+			const view = getTestObjectView();
+
+			view.checkout.events.on("changed", (meta) => {
```
CraigMacomber: Is this the "change" event you error message and changeset are referring to? That's not even a package exported/user facing API.

Its good to have a test at this level, but we should not document it for users at this level.

We should also likely have a test using the actual public API surface which can hit this.

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
```diff
@@ -1341,6 +1341,27 @@ describe("SchematizingSimpleTreeView", () => {
...
+				if (meta.isLocal && meta.kind === CommitKind.Default) {
```
CraigMacomber: why do we need this conditional in here? (Note if there is an actual reason for it, the reason should go in a code comment, not as a response to my PR thread)

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
```diff
@@ -1341,6 +1341,27 @@ describe("SchematizingSimpleTreeView", () => {
 			assert.equal(receivedLabels.tree.sublabels[1]?.sublabels.length, 0);
 		});
 
+		it("direct (non-transaction) edit from a changed listener throws", () => {
+			// Engaging `editLock` around `changed` emissions means the lock check on the editor
+			// proxy also fires for direct edits — not just transactions. Without this, a consumer
+			// could bypass the transaction-time check by writing tree state directly inside a
+			// `changed` listener, with the same corruption potential.
```
CraigMacomber: This comment seems a bit odd. I'm not sure why its here or what its trying to communicate. I don't get how "editor proxy" makes sense here nor why transactions are relevant.

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
```diff
@@ -1341,6 +1341,27 @@ describe("SchematizingSimpleTreeView", () => {
...
+			view.checkout.events.on("changed", (meta) => {
+				if (meta.isLocal && meta.kind === CommitKind.Default) {
+					view.root.content = view.root.content + 1;
+				}
+			});
+
+			assert.throws(
+				() => (view.root.content = 1),
+				validateUsageError(
+					/Editing the tree is forbidden during a nodeChanged, treeChanged, or changed event/,
```
CraigMacomber: if you are matching the entire error message, I think you should just use a string not a regex (which allows prefixes and suffixes). The point of using a regex here is if you want to make the check less specific so it won't break for unimportant changes to the message content: that doesn't make sense to use when validating the entire message

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
```diff
@@ -1300,36 +1300,38 @@ describe("sharedTreeView", () => {
 				args.duringEdit(view);
 			});
 
-			assert.throws(() => (view.root.number = 0), new RegExp(args.error));
+			assert.throws(() => (view.root.number = 0), validateUsageError(new RegExp(args.error)));
```
CraigMacomber: Do we need a RegExp here? Seems wrong (though preexisting)

CraigMacomber: If we do want to take in strings which can be regular expressions, it would be better to make the input type string or regular expression rather than blindly convert all strings to regular expressions without documenting the string will be used as one (basically use the API design used by validateUsageError and similar APIs: we did it that way for a reason)

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1687,8 +1709,9 @@ class EditLock {
 			// These type assertions ensure that the event name strings used here match the actual event names
 			const nodeChanged: keyof TreeChangeEvents = "nodeChanged";
 			const treeChanged: keyof TreeChangeEvents = "treeChanged";
+			const changed: keyof CheckoutEvents = "changed";
```
CraigMacomber: checkout is not a public facing concept. All this fancy typing checking is supposed to ensure the name we use is the thing its called int he public API (like TreeChangeEvents), so this pattern doesn't make sense to use with user facing errors for internal event names.

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
daesunp: Updated with suggestion

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: Removed type checking

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
daesunp: Updated with suggestion

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
daesunp: Simplified comment for clarification

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
daesunp: Yeah I this was unnecessary for this test, removed :)

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
daesunp: Added test that uses public api

packages/dds/tree/src/shared-tree/treeCheckout.ts
daesunp: updated with suggestion

.changeset/tough-lies-say.md
daesunp: Updated `events.mdx` file to include your suggestion in "Editing in response to change events" section, and removed `TreeBranchAlpha` workaround in the changeset

.changeset/tough-lies-say.md
daesunp: Updated with suggestion

.changeset/tough-lies-say.md
daesunp: The `changed` event doesn't seem to live on the API docs because it's internal, so I have reworded the changeset to not mention `changed` event, and applied your suggestions below.

.changeset/tough-lies-say.md
```diff
@@ -0,0 +1,20 @@
...
+Editing a `SharedTree` from inside one of its change-event callbacks has always been forbidden, but one path was not being caught: edits (along with branch operations, reverts, transactions, etc.) made while the tree was emitting its post-change notification ran to completion instead of throwing.
+
+Such edits would apply to the tree, trigger further change notifications, and could re-enter the same listener for the resulting commits.
+This can produce infinite edit loops, redundant work across clients, incorrect attribution, broken undo/redo grouping, and pollution of the outer commit's label data.
+
+This release closes that gap: editing the tree during a change-event callback now throws the same canonical `UsageError` as the other change-event callbacks:
+
+```
+Editing the tree is forbidden during a change event callback
+```
```
Josmithr: Nit: I think this would be better expressed using a quote block
```suggestion
> Editing the tree is forbidden during a change event callback
```

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
```diff
@@ -1341,6 +1341,36 @@ describe("SchematizingSimpleTreeView", () => {
...
+				view.root.content = view.root.content + 1;
+			});
+
+			assert.throws(
+				() => (view.root.content = 1),
+				validateUsageError("Editing the tree is forbidden during a change event callback"),
+			);
+		});
+
+		it("direct edit from a treeChanged listener throws (public API)", () => {
```
Josmithr: This test seems like it belongs in the `TreeNodeApi` test suite, rather than here. Would also probably be good to test both `nodeChanged` and `treeChanged` in isolation, just in case their implementations ever diverge from one another.

Josmithr (commented):
Left a couple of minor suggestions, but otherwise looks good to me!

.changeset/tough-lies-say.md
daesunp: Updated with suggestion

packages/dds/tree/src/test/shared-tree/schematizingTreeView.spec.ts
daesunp: Updated with suggestion

Josmithr (approved):

Josmithr (approved):

PR #27286, opened 2026-05-12
Add first-writer-wins claims to DataObject

kian-thompson:
## Description

Adds a new `SharedClaims` DDS (`@fluidframework/claims-dds`) that lets a data store publish first-writer-wins key/value entries. A claim is a small piece of immutable per-data-store state where the first client to write a given key wins; concurrent writers from other clients observe `"AlreadyClaimed"` and can branch their logic accordingly.

Use a claim (rather than writing to `DataObject.root`) when you specifically need first-writer-wins semantics — for example, when multiple clients race to designate themselves as the owner of a particular role within the data store and only one should succeed.

Every `DataObject` automatically gets a `SharedClaims` channel — no opt-in required. The DDS can also be consumed standalone by registering `SharedClaimsFactory` on any data store.

## New API surface (all `@internal`)

New package `@fluidframework/claims-dds`:
- `SharedClaims` / `ISharedClaims` — the DDS itself, exposing `trySetClaim(key, value)`, `getClaim(key)`, `hasClaim(key)`, and a `claims` iterator.
- `ClaimResult = "Success" | "AlreadyClaimed"`.
- `IClaimAttempt` — discriminated union returned from `trySetClaim` covering the synchronous and `Pending` (awaiting ack) cases.
- `SharedClaimsFactory`.

On `DataObject` (convenience helpers that forward to the auto-installed channel):
- `trySetClaim(key, value)`, `getClaim(key)`, `hasClaim(key)`.

## Implementation notes

- `DataObject.initializeInternal` creates the claims channel under id `"claims"`. A try/catch fallback handles legacy documents that predate the channel.
- `DataObjectFactory` auto-registers `SharedClaimsFactory` alongside `SharedDirectory` / `SharedMap`.
- Claim values may contain `IFluidHandle` instances; they flow through the standard `IFluidSerializer` (handle encoding + GC route collection come for free from `SharedObject`).
- Wire format: op = `{ type: "set"; key; value }`. Summary blob `header` records the sequenced entries.
- Detached writes go straight into local state and resolve synchronously to `Success`.
- Concurrent local attempts on the same key dedupe via a shared `Deferred`. On `runtime.dispose` all pending attempts resolve to `AlreadyClaimed`. Staging-mode rollback rejects the pending `Deferred`.

## Testing

`packages/dds/claims/src/test/claims.spec.ts` covers: single-client success, two-client race / first-writer-wins, repeated calls from winner and loser, detached set + persistence, GC route collection from embedded handles, reconnect/resubmit, snapshot rehydration via summarize / load, and rollback/dispose semantics. 9/9 passing.


ChumpChief:
This definitely feels like it's in the direction I would expect (right granularity and availability, right scenario support, etc.).  Very much prefer this direction over #27291.

My main feedback would be about encapsulation and reuse of existing concepts.  What you're building here is effectively a DDS - an improved and simplified version of ConsensusRegisterCollection.  I'd be interested to see this actually built as a new standalone DDS, which we can then leverage in the DataStoreRuntime to provide the functionality.  This would help keep the runtime from growing in direct responsibility, avoids expanding the op type vocabulary, probably simplifies your load flow and op buffering requirements, etc.  It probably also sets you up well to leverage existing DDS testing framework, and avoids being a special case for features that intersect with the op stream (e.g. staging mode, rollback, etc.)

Such a DDS could start out as `@internal` since we don't have direct goals of providing the raw functionality at the moment, but I could also see it eventually being a public replacement that lets us deprecate ConsensusRegisterCollection.

PR #27288, opened 2026-05-12
runtime: Add "retired" features implementation for Document Schema

scottn12:
## Description

Adds a "retired" feature mechanism to Document Schema. A retired feature is one that was previously toggleable via `IDocumentSchemaFeatures`  but is now hardcoded to a single value by the runtime.

Why we need this: existing documents may still carry a retired feature's key in their persisted
schema, so Document Schema must continue to recognize it (otherwise `checkRuntimeCompatibility`
would fail the session on an "unknown" runtime property). At the same time, we no longer want
that feature to be part of the public `IDocumentSchemaFeatures` surface, because callers can
no longer influence it.

Retired features are tracked in a separate `retiredDocumentSchemaFeatures` map. Each entry
bundles the property handler with the runtime's hardcoded value. They participate in the
normal validate / merge / schema-change-op flow exactly like non-retired features — the only
difference is that the value comes from the map instead of from the constructor's `features`
argument. A compile-time `satisfies` constraint ensures a key cannot appear in both
`IDocumentSchemaFeatures` and `retiredDocumentSchemaFeatures`.

Note: No features are retired yet. This PR just sets up the infrastructure for when features are retired in the future.

## Misc
[AB#72041](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/72041)

### DIFF
--- a/packages/runtime/container-runtime/src/summary/documentSchema.ts
+++ b/packages/runtime/container-runtime/src/summary/documentSchema.ts
@@ -304,6 +304,55 @@ const documentSchemaSupportedConfigs = {
 };
 
 /**
+ * Entry in {@link retiredDocumentSchemaFeatures}.
+ *
+ * - `handler` validates/merges the persisted value (same as a "non-retired" feature).
+ * - `value` is the hardcoded value that this runtime will always use for the desired schema.
+ */
+interface IRetiredFeatureEntry {
+	handler: IProperty;
+	value: DocumentSchemaValueType;
+}
+
+/**
+ * Retired runtime features. A retired feature is one this runtime version no longer toggles
+ * via {@link IDocumentSchemaFeatures}, but that older documents may still carry in their
+ * persisted schema. Each entry bundles the property handler with the hardcoded value for the feature.
+ *
+ * Retired features participate in the normal merge / schema-change-op flow exactly like
+ * non-retired features — the only difference is that their values are hardcoded.
+ */
+const retiredDocumentSchemaFeatures = {} satisfies Record<string, IRetiredFeatureEntry> & {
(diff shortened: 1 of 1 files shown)


packages/runtime/container-runtime/src/summary/documentSchema.ts
```diff
@@ -303,6 +303,55 @@ const documentSchemaSupportedConfigs = {
...
+
+/**
+ * Retired runtime features. A retired feature is one this runtime version no longer toggles
+ * via {@link IDocumentSchemaFeatures}, but that older documents may still carry in their
+ * persisted schema. Each entry bundles the property handler with the hardcoded value for the feature.
+ *
+ * Retired features participate in the normal merge / schema-change-op flow exactly like
+ * non-retired features — the only difference is that their values are hardcoded.
+ */
+const retiredDocumentSchemaFeatures = {} satisfies Record<string, IRetiredFeatureEntry> & {
```
ChumpChief: This structure/syntax combo confused me for a bit since there aren't any yet, maybe do this for clarity and give an example
```suggestion
const retiredDocumentSchemaFeatures = {
	// Note: There are currently no retired features.  When one is added it should look like
	// featureFoo: ...
} satisfies Record<string, IRetiredFeatureEntry> & {
```

ChumpChief (commented):
I like the idea of making it explicit!  Just a documentation/comment note.

packages/runtime/container-runtime/src/summary/documentSchema.ts
scottn12: Added in latest!

ChumpChief (approved):

PR #20565, opened 2024-04-09
[Data Virtualization & Offline]: Reimplement groupId offline with the loader

tyler-cai-microsoft:
Original PR: https://github.com/microsoft/FluidFramework/pull/20342/files

[AB#7244](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/7244)
[AB#7245](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/7245)
[AB#7588](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/7588)

# Description
For Data virtualization and offline, we need to build a solution that is robust when going offline and refreshing. The method we use here is to store any snapshots that we need in order to process the offline ops. The goal here is to provide basic Data Virtualization for offline.

## Storing
We need to store these snapshots separately when the snapshot was downloaded at some point between the base/refreshed snapshot and before the sequence number we went offline. This includes when we load from pending state as we can consider that "downloading".

The best way to do this is to put it in the pending state.

## Refresh
When we refresh the state of the snapshot, we need to make sure that all groupIds (virtualized snapshots that were not in the base snapshot) are downloaded so that the new snapshot is up to date.
 
## Loading
To load from these snapshots, we store them separately as pending state, so when we are apply stashed ops for offline, the right blobs are downloaded. We don't want to download any blobs from the future when applying stashed ops as that would lead to inconsistent/incorrect state.

PR #19697, opened 2024-02-20
Encapsulate container state attributes

dannimad:
There is a lot of code spread out in container.ts related to capturing the pending state and managing IPendingContainerState interface and related attributes. We can abstract all that manipulation into a separate class to simplify the container code and make future features like refreshing the base snapshot unit testable.

PR #27290, opened 2026-05-12
Remove deprecated ILoaderOptions.enableOfflineLoad

dannimad:
## Description

Removes the deprecated `enableOfflineLoad` field from `ILoaderOptions` (`@fluidframework/container-definitions`) and the legacy `Fluid.Container.enableOfflineLoad` feature-gate read from the config provider. The field was already marked `@deprecated Do not use.`; this completes its removal.

Offline-load default behavior is preserved: interactive clients still get offline load enabled by default. To opt out, set `Fluid.Container.enableOfflineFull: false` via the config provider — that is now the single control surface.

As a safety net for callers migrating from the legacy gate, `Container` load now throws a `UsageError` if `pendingLocalState` is provided while `Fluid.Container.enableOfflineFull` is explicitly set to `false`. This prevents silently loading from pending state with offline-load disabled.

Also drops the deprecated entry from the `test-service-load` options matrix and updates the stress-runner gate that referenced it.

## Breaking Changes

`ILoaderOptions.enableOfflineLoad` (`@legacy @beta`) is removed. Migration:

- Remove `enableOfflineLoad` from any `ILoaderOptions` passed to the `Loader`.
- If you previously set `Fluid.Container.enableOfflineLoad` via the config provider, set `Fluid.Container.enableOfflineFull` instead (same boolean value).

See [Breaking-vs-Non-breaking-Changes](https://github.com/microsoft/FluidFramework/wiki/Breaking-vs-Non-breaking-Changes).

## Reviewer Guidance

- This is a `@legacy @beta` break. Per the Beta-Break-Process it must land in a `.N0` minor and may need to be staged on `test/breaks/client/X.Y0/`. Confirm we're in the right release window before merging.
- `fluid-cr-api` will be auto-assigned as a required reviewer.
- The new throw-on-misconfig is a defensive guard while `enableOfflineFull` remains the control surface; happy to drop or relax it if reviewers prefer.

### DIFF
--- a/packages/common/container-definitions/api-report/container-definitions.legacy.beta.api.md
+++ b/packages/common/container-definitions/api-report/container-definitions.legacy.beta.api.md
@@ -436,7 +436,6 @@ export interface ILoader extends Partial<IProvideLoader> {
 export type ILoaderOptions = {
     cache?: boolean;
     client?: IClient;
-    enableOfflineLoad?: boolean;
     provideScopeLoader?: boolean;
     maxClientLeaveWaitTime?: number;
 };
--- a/packages/common/container-definitions/src/loader.ts
+++ b/packages/common/container-definitions/src/loader.ts
@@ -611,11 +611,6 @@ export type ILoaderOptions = {
 	client?: IClient;
 
 	/**
-	 * @deprecated Do not use.
-	 */
-	enableOfflineLoad?: boolean;
-
-	/**
 	 * Provide the current Loader through the scope object when creating Containers. It is added
 	 * as the `ILoader` property, and will overwrite an existing property of the same name on the
 	 * scope. Useful for when the host wants to provide the current Loader's functionality to
--- a/packages/loader/container-loader/src/container.ts
+++ b/packages/loader/container-loader/src/container.ts
@@ -961,11 +961,14 @@ export class Container
 			enableSummarizeProtocolTree,
 		);
 
-		const offlineLoadEnabled =
-			this.isInteractiveClient &&
-			(this.mc.config.getBoolean("Fluid.Container.enableOfflineLoad") ??
-				this.mc.config.getBoolean("Fluid.Container.enableOfflineFull") ??
-				options.enableOfflineLoad !== false);
+		const explicitOfflineFull = this.mc.config.getBoolean("Fluid.Container.enableOfflineFull");
+		// paranoic defense mechanism while enableOfflineFull still exists in config.
+		if (pendingLocalState !== undefined && explicitOfflineFull === false) {
+			throw new UsageError(
+				"Cannot load from pending local state when Fluid.Container.enableOfflineFull is explicitly disabled",
+			);
+		}
+		const offlineLoadEnabled = this.isInteractiveClient && (explicitOfflineFull ?? true);
 		this.serializedStateManager = new SerializedStateManager(
 			this.subLogger,
 			this.storageAdapter,
--- a/.changeset/remove-enable-offline-load.md
+++ b/.changeset/remove-enable-offline-load.md
@@ -1,0 +1,16 @@
+---
+"@fluidframework/container-definitions": minor
+"@fluidframework/container-loader": minor
+"__section": legacy
+---
+
+Remove deprecated `ILoaderOptions.enableOfflineLoad`
+
+The deprecated `enableOfflineLoad` property on `ILoaderOptions` has been removed, along with the legacy `Fluid.Container.enableOfflineLoad` feature gate read from the config provider. Offline load remains enabled by default for interactive clients.
+
+To opt out, set `Fluid.Container.enableOfflineFull` to `false` via the config provider. To prevent silent misconfiguration, container load now throws a `UsageError` if a `pendingLocalState` is provided while `Fluid.Container.enableOfflineFull` is explicitly set to `false`.
+
+#### Migration
+
+- Remove any use of `enableOfflineLoad` from `ILoaderOptions` you pass to the `Loader`.
+- If you previously set `Fluid.Container.enableOfflineLoad` via the config provider, set `Fluid.Container.enableOfflineFull` instead.
--- a/packages/test/test-service-load/src/optionsMatrix.ts
+++ b/packages/test/test-service-load/src/optionsMatrix.ts
@@ -34,7 +34,6 @@ const loaderOptionsMatrix: OptionsMatrix<ILoaderOptionsExperimental> = {
 	client: [undefined],
 	provideScopeLoader: booleanCases,
 	maxClientLeaveWaitTime: numberCases,
-	enableOfflineLoad: booleanCases,
 	enableOfflineSnapshotRefresh: booleanCases,
 	snapshotRefreshTimeoutMs: [undefined, 60 * 5 * 1000 /* 5min */],
 };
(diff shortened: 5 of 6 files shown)


anthony-murphy:

  - 🔄 mechanical (Anglerfish)
  - 🔄 mechanical (Fangtooth)
  - ✅ summarize (Anglerfish)
- 👁️ **Adversarial Filter** — puzzling over the find…
  - 🔄 prosecution (Anglerfish)
- ✅ Synthesis
- 📡 **Comments** — transmitting the discovery…
  - 🔄 comments (Anglerfish)

packages/common/container-definitions/api-report/container-definitions.legacy.beta.api.md
```diff
@@ -436,7 +436,6 @@ export interface ILoader extends Partial<IProvideLoader> {
 export type ILoaderOptions = {
     cache?: boolean;
     client?: IClient;
-    enableOfflineLoad?: boolean;
```
anthony-murphy, line 439: this is a legacy breaking change and needs to wait for the next legacy breaking release

packages/loader/container-loader/src/container.ts
```diff
@@ -961,11 +961,14 @@ export class Container
...
 		);
 
-		const offlineLoadEnabled =
-			this.isInteractiveClient &&
-			(this.mc.config.getBoolean("Fluid.Container.enableOfflineLoad") ??
-				this.mc.config.getBoolean("Fluid.Container.enableOfflineFull") ??
-				options.enableOfflineLoad !== false);
+		const explicitOfflineFull = this.mc.config.getBoolean("Fluid.Container.enableOfflineFull");
+		// paranoic defense mechanism while enableOfflineFull still exists in config.
+		if (pendingLocalState !== undefined && explicitOfflineFull === false) {
```
anthony-murphy, line 966: the user passing pending state is a signal they want to load from it, i don't think we should override that decision based on flag, and i don't think we did before. the flags are a bit of a mess, but i wouldn't expect new error cases

dannimad: It's not an override but a safeguard. In any case, given these change will need to wait I'd just ignore the depracated flag for now. I wanted to start with this one as it was the cleanest to remove. I agree flags are a mess right now.

anthony-murphy:
## Deep Review

*Reviewed commit `c61f1ef` on 2026-05-12.*

Not ready for sign-off. Two concerns flagged inline: a silent runtime semantic flip for consumers using the legacy `Fluid.Container.enableOfflineLoad` config-provider key (the new `UsageError` covers only the new key), and the Beta-Break-Process `.N0` release-window / staging check you flagged in the PR description. Both have local fixes.

### Path to Ready

- [ ] Resolve inline threads
- [ ] If Beta-Break-Process staging is required for `@fluidframework/container-definitions` at the current version, add the `packages/test/breaks/client/X.Y0/` entry in this PR

### Context for Reviewers

- This completes the long-promised removal flagged in #19306 (ChumpChief, 2024-02-02), where anthony-murphy blocked unilateral removal with "we need a transition path". #19306 cleaned the typed-API footgun and deprecated the field; this PR is the runtime cleanup ~27 months later. The transition-path obligation now applies to the equally-public config-provider key, not just the typed surface — that is what drives S1 below.
- Suggested reviewers based on blame/history: **ChumpChief** (owner of the #19306 typed-surface cleanup; api-report and Beta-Break-Process), **anthony-murphy** (cross-cutting; cited on #19306, #14502, #19697), **wes-carlson** (stress-test offline scheduling: #12214, #14502, #16031), **tyler-cai-microsoft** (#20565, pending-state + snapshot-refresh interaction).

- **Needs human judgment — release-window confirmation** — Is this PR landing in a `.N0` minor for `@fluidframework/container-definitions`, and is `test/breaks/client/X.Y0/` staging mandatory at the current version? You explicitly invited this confirmation in the PR description; the pipeline cannot answer it.
- **Needs human judgment — design call on the new `UsageError` guard** — You wrote it "happy to drop or relax it if reviewers prefer." Whether the guard should be permanent, temporary, or relaxed is a taste call for the area owners (ChumpChief / anthony-murphy).
- **Needs human judgment — strict vs conservative S1 fix** — One reviewer-proposal direction would keep both config keys live with precedence for one release rather than only logging/throwing. Worth a call on whether the partner-migration risk warrants the conservative path for one more release.

</details>

packages/loader/container-loader/src/container.ts
```diff
@@ -961,11 +961,14 @@ export class Container
...
+			throw new UsageError(
+				"Cannot load from pending local state when Fluid.Container.enableOfflineFull is explicitly disabled",
+			);
+		}
+		const offlineLoadEnabled = this.isInteractiveClient && (explicitOfflineFull ?? true);
```
anthony-murphy, line 971: Deep Review: The new code replaces `getBoolean("Fluid.Container.enableOfflineLoad") ?? getBoolean("Fluid.Container.enableOfflineFull") ?? options.enableOfflineLoad !== false` with a single read of `Fluid.Container.enableOfflineFull`. A consumer who today sets `Fluid.Container.enableOfflineLoad: false` via the config provider to opt out of offline-load gets the **opposite** behaviour after this PR — offline silently re-enabled. The new `UsageError` only fires on `explicitOfflineFull === false`, so this case produces no `UsageError`, no telemetry, and no compile break (the config provider is a dynamic string surface).

PR #19306 (ChumpChief, 2024-02-02) established the transition-path rule for this exact option — anthony-murphy: *"i like removing it, but i think its used today, so we need a transition path, i don't think we can just unilaterally remove"*. This PR satisfies that for the typed surface but not for the equally-public config-provider key. The changeset documents the rename but no runtime hint exists when the stale key is observed.

Two cheap local fixes, either works:

- (a) Emit a one-shot `mc.logger` telemetry warning when `Fluid.Container.enableOfflineLoad` is observed in the config provider, so partners notice they need to migrate before behaviour silently flips.
- (b) Extend the new `UsageError` guard to also throw when the legacy key is explicitly `false`, mirroring the `pendingLocalState + enableOfflineFull=false` safety net.

Either keeps the "single gate" goal of the PR while preserving a runtime signal for migrating consumers.

packages/common/container-definitions/src/loader.ts
```diff
@@ -610,11 +610,6 @@ export type ILoaderOptions = {
 	 */
 	client?: IClient;
 
-	/**
-	 * @deprecated Do not use.
-	 */
-	enableOfflineLoad?: boolean;
-
 	/**
 	 * Provide the current Loader through the scope object when creating Containers. It is added
```
anthony-murphy, line 614: Deep Review: This removes `enableOfflineLoad?: boolean` from the `@legacy @beta` `ILoaderOptions` surface (api-report change at `container-definitions.legacy.beta.api.md:436-440`). You flagged this yourself in the PR description: *"This is a `@legacy @beta` break. Per the Beta-Break-Process it must land in a `.N0` minor and may need to be staged on `test/breaks/client/X.Y0/`. Confirm we're in the right release window before merging."*

The diff contains no entry under `packages/test/breaks/client/...` and no version-bump-policy / legacy.beta rename. Precedent #19306 (ChumpChief, 2024-02-02) took two steps over a long window — deprecate-then-remove — with explicit framing debate on the API surface.

Before merge:

- Confirm with the release manager and ChumpChief that this lands in a `.N0` minor for `@fluidframework/container-definitions`.
- If the Beta-Break-Process harness requires it for the current version, add the corresponding `test/breaks/client/X.Y0/` staging entry in this PR.
- Verify CI's type-compat pipeline produces the correct `validate*Previous.generated.ts` artifacts (regenerate by hand if needed).

PR #27293, opened 2026-05-12
Add overrides to pnpm workspace for compat installs to address CVEs

alexvy86:
## Description

Adds a few overrides to the pnpm workspace for compat installs. We got an alert for axios 0.28.1, so that was the main thing to fix, but applied a few others that we also have in other places.

## Reviewer Guidance


Josmithr (approved):

packages/test/test-version-utils/compat-workspaces/full/pnpm-workspace.yaml
```diff
@@ -1,3 +1,6 @@
+
+packages:
+  - '*'
 # This file is scaffolded by scripts/updateCompatVersions.ts.
```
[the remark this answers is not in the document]
alexvy86: Resolved in latest.

kian-thompson (approved):

alexvy86:
FYI @ChumpChief @jason-ha . Where possible we've been starting to use `pnpm-workspace.yaml`  for overrides so we can comment them better. We couldn't do it in the client release group because syncpack didn't quite support catalogs and/or the `pnpm-workspace.yaml` file. Syncpack 15 seems to have added the necessary support so maybe we can do it now :).

packages/test/test-version-utils/compat-workspaces/full/pnpm-workspace.yaml
```diff
@@ -11,9 +15,14 @@ minimumReleaseAgeExclude:
   - "@fluid-private/*"
   - "@fluid-tools/*"
   - "fluid-framework"
+overrides:
+  axios@<1: ^0.31.1
```
Josmithr, line 24: Nit: might be worth a quick comment explaining why we need these new overrides. I'm assuming we don't have any reasonable mechanism for determining when these can be removed in the future?

Josmithr (approved):

packages/test/test-version-utils/compat-workspaces/full/pnpm-workspace.yaml
alexvy86: Added in latest.

Josmithr (approved):

alexvy86:
> FYI @ChumpChief @jason-ha . Where possible we've been starting to use `pnpm-workspace.yaml` for overrides so we can comment them better. We couldn't do it in the client release group because syncpack didn't quite support catalogs and/or the `pnpm-workspace.yaml` file. Syncpack 15 seems to have added the necessary support so maybe we can do it now :).

https://github.com/microsoft/FluidFramework/pull/27320

PR #26103, opened 2025-12-30
SnapshotRefresher creation and unit tests

dannimad:
This PR refactors snapshot refresh functionality by extracting it from SerializedStateManager into a new dedicated SnapshotRefresher class, improving separation of concerns,maintainability and testing. Current e2e snapshot refresh tests are disabled due to flakiness so this PR adds back some test coverage for those scenarios.

PR #27295, opened 2026-05-12
chore(container-loader): tighten snapshot refresh config flag gating

dannimad:
## Description

Removes the `Fluid.Container.enableOfflineFull` fallback from `SnapshotRefresher`'s `#snapshotRefreshEnabled` gating. Snapshot refresh now requires `Fluid.Container.enableOfflineSnapshotRefresh` to be explicitly set to `true`; without it, refresh stays disabled even when `enableOfflineFull` is on.

This decouples snapshot refresh from the broader offline-full kill switch so the two can be rolled forward and back independently.

Adds a unit test pinning the new gating: with only `enableOfflineFull=true`, the refresh timer does not fire and no storage calls are made.

## Reviewer Guidance

### DIFF
--- a/packages/loader/container-loader/src/snapshotRefresher.ts
+++ b/packages/loader/container-loader/src/snapshotRefresher.ts
@@ -97,8 +97,7 @@ export class SnapshotRefresher implements IDisposable {
 
 		this.#snapshotRefreshEnabled =
 			this.offlineLoadEnabled &&
-			(this.mc.config.getBoolean("Fluid.Container.enableOfflineSnapshotRefresh") ??
-				this.mc.config.getBoolean("Fluid.Container.enableOfflineFull")) === true;
+			(this.mc.config.getBoolean("Fluid.Container.enableOfflineSnapshotRefresh") ?? false);
 
 		this.refreshTimer = this.#snapshotRefreshEnabled
 			? new Timer(this.snapshotRefreshTimeoutMs, () => this.tryRefreshSnapshot())
--- a/packages/loader/container-loader/src/test/snapshotRefresher.spec.ts
+++ b/packages/loader/container-loader/src/test/snapshotRefresher.spec.ts
@@ -60,6 +60,12 @@ function enableOfflineSnapshotRefresh(logger: ITelemetryBaseLogger): ITelemetryB
 	}).logger;
 }
 
+function enableOfflineFullOnly(logger: ITelemetryBaseLogger): ITelemetryBaseLogger {
+	return mixinMonitoringContext(logger, {
+		getRawConfig: (name) => (name === "Fluid.Container.enableOfflineFull" ? true : undefined),
+	}).logger;
+}
+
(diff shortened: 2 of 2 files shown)


anthony-murphy:
## Deep Review

*Reviewed commit `3007392` on 2026-05-12.*

The code change is clean and the new unit test pins the intended contract. No correctness concerns. What's left is release-note hygiene, a design-history pointer in the description, and a deployer-side audit confirmation before this ships.

### Path to Ready

- [ ] Add a `.changeset/` entry (or repo-equivalent release note) calling out that `Fluid.Container.enableOfflineFull` no longer implies snapshot refresh — operators must set `Fluid.Container.enableOfflineSnapshotRefresh=true` explicitly to preserve prior refresh behavior.
- [ ] Update the PR description to cite #25605 and explicitly note that this narrows the umbrella semantics on the snapshot-refresh axis (so future readers of #25605's "unified offline" framing don't reach a stale conclusion).
- [ ] Confirm in the PR description (or a follow-up) that no live config-provider ring sets `enableOfflineFull=true` without `enableOfflineSnapshotRefresh=true` — or sequence the dedicated-flag rollout ahead of this PR. The only in-repo `enableOfflineFull`-only configuration is a stress test that doesn't opt into refresh, so the residual risk is downstream deployments.
- [ ] Optional: add a manager-level regression test in `packages/loader/container-loader/src/test/serializedStateManager.spec.ts` mirroring the new gate-site test (with `offlineLoadEnabled=true` and only `enableOfflineFull=true`, the embedded refresher does not start a timer or call `getLatestSnapshotInfo`). ~10 lines, guards against future re-coupling at the construction site in `SerializedStateManager`.

### Context for Reviewers

- Builds on the offline-flag taxonomy series: #21504 introduced periodic refresh; #25605 added `enableOfflineFull` as a unified umbrella with `?? enableOfflineFull` as the snapshot-refresh fallback; #25637 centralized/cached the enablement check while fixing timer leaks observed when `enableOfflineFull` was the experimental default; #26103 extracted `SnapshotRefresher` carrying the fallback verbatim. This PR drops that fallback, making `enableOfflineSnapshotRefresh` the sole opt-in.
- Suggested reviewer: **anthony-murphy** — authored #25605 (introduced the umbrella + fallback being removed) and #25637 (centralized the enablement computation), and is the only other recent contributor to `snapshotRefresher.ts`. Best fit to sign off on narrowing the umbrella semantics.

- **Owner sign-off on the umbrella reversal** — anthony-murphy designed `enableOfflineFull` as the unified umbrella in #25605; weigh in on whether narrowing it on the snapshot-refresh axis is acceptable design drift or whether the umbrella should be preserved.
- **Downstream rollout coordination** — confirming which config-provider rings (canary/test/production) currently set `enableOfflineFull` without `enableOfflineSnapshotRefresh` requires visibility into deployment configuration outside this repo.

</details>

markfields (approved):

PR #27300, opened 2026-05-13
build: tag untagged asserts for 2.101.0 release

jatgarg:
Generated by `pnpm run policy-check:asserts` as part of release prep.


CraigMacomber (approved):

CraigMacomber:
Seems odd there is only one new assert, but as this makes progress toward tagging, and doesn't remove any old ones oddly, it doesn't due any harm as is, but seems like it might not be everything.

Checked locally, and this does seem like all of them.

PR #27302, opened 2026-05-13
build(docs): apply prettier to docusaurus.config.ts

jatgarg:
## Description

The docs site CI runs `prettier --check` against the docs package and was failing on a long-line in docusaurus.config.ts that the client-release-group root format script doesn't cover (docs is a separate workspace package). Fix is the canonical prettier --write output: wrap the typesenseAllowedTypos chained calls.

### DIFF
--- a/docs/docusaurus.config.ts
+++ b/docs/docusaurus.config.ts
@@ -30,7 +30,9 @@ const typesenseSearchFields: [field: string, weight: number, allowedTypos: numbe
 ];
 const typesenseQueryBy = typesenseSearchFields.map(([field]) => field).join(",");
 const typesenseQueryByWeights = typesenseSearchFields.map(([, weight]) => weight).join(",");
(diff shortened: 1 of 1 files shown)


WillieHabi (approved):

alexvy86:
@Josmithr / @WayneFerrao we probably want to check how this made it in in the first place. Seems like we're not running all the correct checks when making changes to docs? Also, I see a [failed check](https://github.com/microsoft/FluidFramework/actions/runs/25761416589/job/75663254251) in Wayne's PR yesterday that touched this line and it didn't prevent merging 🤔 .

But also, @jatgarg , are we sure this is the problem? The [failure in your PR](https://github.com/microsoft/FluidFramework/actions/runs/25814146708/job/75840500571) is the same as in Wayne's, and it doesn't seem related to the long line...

PR #27311, opened 2026-05-14
test(matrix,tree): Clean up matrix/table benchmarks

Josmithr:
In `matrix`:
- Combines benchmark test modules

In `tree`:
- Unifies the 2 mostly duplicated performance benchmark suites for tables.
- Combines test modules


packages/dds/tree/src/test/tableSchema.bench.ts
```diff
@@ -0,0 +1,1073 @@
...
+	maxBenchmarkDurationSeconds,
+	title,
+	tableSize,
+	initialCellValue,
+	beforeOperation,
+	operation,
+	afterOperation,
+}: BenchmarkConfig): Test {
+	// eslint-disable-next-line unicorn/prefer-ternary -- The code is easier to read as-is.
+	if (mode === "execution-time") {
```
CraigMacomber: I think it would be simpler to delete the mode parameter and the if/else, and have this unconditionally generate both tests. We have no reason for the memory tests to be off in a special memory directory, and can simply have both sets included inside of one suite.

CraigMacomber (approved):

packages/dds/tree/src/test/tableSchema.bench.ts
Josmithr: I _think_ there are pipeline dependencies or something at play here. Or at least, that was the case when Tong was doing this work initially. I.e., the file paths were important in some way.

Or was all of that cleaned up with your script update work from a couple months back?

CraigMacomber: When I rewrote all the scripts and pipelines, I purged the last of the file name dependence. All filtering is done based on how the texts are named/tagged. And even if you screw that up, the worst you will get is the test run in the wrong batch (we try and split by runtime dependent and non-runtime dependent: we try and avoid overloading the CPU too bad in the runtime dependent one with a bit less parallelism, though currently we don't do a great job at that)

Josmithr: Excellent! Then I will go ahead and clean this up!

CraigMacomber: I still think you could simplify this by deleting the "mode" option, and the if/else block here, and just have runBenchmark setup both tests unconditionally (we use that pattern in the end to end tests somewhere).

Do that and you can also replace runBenchmarkTestSuite and the describe block which calls it twice with just the describe block inside of runBenchmarkTestSuite, but run at the top level of the file.

Overall the setup would be a lot simpler (one less indirection through a function, and one less layer of describe blocks, and no need to plumb "mode" around).

Same simplification could apply to the table code.

If you don't want to do that now, and instead just focus on getting the suites aligned that is fine.

CraigMacomber (approved):
See https://github.com/microsoft/FluidFramework/pull/27311/files#r3245067404 for possible simplification, though that does not need to be in this PR.

packages/dds/tree/src/test/tableSchema.bench.ts
Josmithr: No, that's fair, and easy to fix. Will update.

packages/dds/matrix/src/test/matrix.bench.ts
```diff
@@ -0,0 +1,1059 @@
...
+				// CAREFUL: usually beforeEach/afterEach hooks are used to initialize or interact with variables
+				// whose scope is the encompassing test suite, but that's a problem for memory-profiling tests.
+				// See the comment at the top of the test suite for more details.
+			});
+
+			afterEach(() => {
+				// CAREFUL: usually beforeEach/afterEach hooks are used to initialize or interact with variables
+				// whose scope is the encompassing test suite, but that's a problem for memory-profiling tests.
+				// See the comment at the top of the test suite for more details.
+			});
```
CraigMacomber: I think this is out of date with the new memory testing APIs and can just be removed

packages/dds/tree/src/test/tableSchema.bench.ts
```diff
@@ -0,0 +1,1078 @@
...
+	maxBenchmarkDurationSeconds,
+	title,
+	tableSize,
+	initialCellValue,
+	beforeOperation,
+	operation,
+	afterOperation,
+}: BenchmarkConfig): Test {
+	// eslint-disable-next-line unicorn/prefer-ternary -- The code is easier to read as-is.
+	if (mode === "execution time") {
```
CraigMacomber: This still has the mode paramater and internal `if` based on it which I suggested removing, as does the matrix tests

packages/dds/tree/src/test/tableSchema.bench.ts
```diff
@@ -0,0 +1,1078 @@
...
+				// CAREFUL: usually beforeEach/afterEach hooks are used to initialize or interact with variables
+				// whose scope is the encompassing test suite, but that's a problem for memory-profiling tests.
+				// See the comment at the top of the test suite for more details.
+			});
+
+			afterEach(() => {
+				// CAREFUL: usually beforeEach/afterEach hooks are used to initialize or interact with variables
+				// whose scope is the encompassing test suite, but that's a problem for memory-profiling tests.
+				// See the comment at the top of the test suite for more details.
+			});
```
CraigMacomber: I think this is out of date with the new memory testing APIs

CraigMacomber (approved):

CraigMacomber (approved):

anthony-murphy (approved):

PR #27312, opened 2026-05-14
ci: Stop running server pipelines for the lts branch

alexvy86:
## Description

Server releases will never be made from the LTS branch, so having the pipelines trigger for PRs to it / commits in it is not necessary. Arguably, we could remove the whole server/ folder in LTS, but I did not try to be that aggressive with this PR.

This just makes it so:
- Those pipelines don't run for the `lts` branch.
- The long-gone `next` branch is removed from the triggers section
- Pipelines that used to run for PRs into `release/*` branches but not on commits to those branches (historian, gitrest, gitssh), now also run for commits on the branches. The routerlicious pipeline already behaves like this. Not clear why we wouldn't want the others to do the same.

## Reviewer Guidance

### DIFF
--- a/tools/pipelines/server-gitrest.yml
+++ b/tools/pipelines/server-gitrest.yml
@@ -40,8 +40,7 @@ trigger:
   branches:
     include:
     - main
-    - next
-    - lts
+    - release/*
   paths:
     include:
     - .prettierignore
@@ -65,8 +64,6 @@ pr:
   branches:
     include:
     - main
-    - next
-    - lts
     - release/*
   paths:
     include:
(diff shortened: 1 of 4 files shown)


ChumpChief (approved):
Seems good to me, would you also merge this to the lts branch?

alexvy86:
> Seems good to me, would you also merge this to the lts branch?

For some reason I was under the impression that ADO only processed triggers based on the version of the yml in the default repo branch but [the docs](https://learn.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#branch-consideration-for-triggers-in-yaml-pipelines) I'm finding right now suggest it is indeed per-branch. So yeah, I'll cherry pick this to lts 👍 .

PR #27313, opened 2026-05-14
(tree) Update test that validates unknown op space ID bug

agarwal-navin:
Updated the test that was added by https://github.com/microsoft/FluidFramework/pull/27292:
- Removed usage of `any` when creating the shared tree.
- Disabled summarization in regular clients that could lead to interfering with summarizer clients.
- Added validation that summary does not reject so that it doesn't fail because of error thrown in test.

### DIFF
--- a/packages/test/test-end-to-end-tests/src/test/tree.spec.ts
+++ b/packages/test/test-end-to-end-tests/src/test/tree.spec.ts
@@ -21,7 +21,6 @@ import {
 	createContainerRuntimeFactoryWithDefaultDataStore,
 	createSummarizerFromFactory,
 	summarizeNow,
-	type SummaryInfo,
 } from "@fluidframework/test-utils/internal";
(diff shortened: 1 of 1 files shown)


packages/test/test-end-to-end-tests/src/test/tree.spec.ts
```diff
@@ -129,13 +118,16 @@ describeCompat(
...
 				defaultFactory,
-				registryEntries: [
-					[defaultFactory.type, Promise.resolve(defaultFactory)],
-					[treeOwningFactory.type, Promise.resolve(treeOwningFactory)],
-				],
+				registryEntries,
 				runtimeOptions: {
 					// SharedTree requires the runtime id compressor.
 					enableRuntimeIdCompressor: "on",
+					// Disbale summaries for regular clients so they don't interfere with on demand summaries.
```
Abe27342:
```suggestion
					// Disable summaries for regular clients so they don't interfere with on demand summaries.
```

Abe27342 (approved):

PR #27314, opened 2026-05-14
ci(server): Try explicit networkIsolationPolicy

alexvy86:
## Description

I haven't been able to figure out why the gitssh pipeline activates a different set of rules for network isolation policy, and the additional rules that get activated break the requests to `https://registry-1.docker.io` that happen during a docker build even if the docker images are baked into the build agent.

Trying out setting an explicit network isolation policy to see if it helps.

## Reviewer Guidance


alexvy86:
Build still failed, so it seems the setting is not the final decider for which network isolation rules get applied. Closing as failed experiment.

PR #27315, opened 2026-05-14
Update FluidClientVersion docs

CraigMacomber:
## Description

FluidClientVersion predates MinimumVersionForCollab: this gets its docs more aligned with MinimumVersionForCollab

## Reviewer Guidance

### DIFF
--- a/packages/runtime/runtime-definitions/src/compatibilityDefinitions.ts
+++ b/packages/runtime/runtime-definitions/src/compatibilityDefinitions.ts
@@ -20,6 +20,11 @@
  * Since this uses the semver notion of "greater" (which might not actually mean a later release, or supporting more features), care must be taken with how this is used.
  * See remarks for {@link @fluidframework/runtime-utils#MinimumMinorSemanticVersion} for more details.
  *
+ * This scheme assumes a single version is always enough to communicate compatibility, which requires that compatibility is strictly increasing across releases.
+ * If this is violated (for example, a subset of incompatible features from 3.x is back-ported to 2.x, or compatibility depends on a patch thats not in the next minor's first release),
+ * a more complex scheme may be needed;
+ * one can be added if/when needed since it would be opt-in and thus non-breaking.
+ *
  * Since this type is marked with `@input`, it can be generalized to allow more cases in the future as a non-breaking change.
  *
  * TODO: before stabilizing this further, some restrictions should be considered (since once stabilized, this can be relaxed, but not more constrained).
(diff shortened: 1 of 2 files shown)


packages/runtime/runtime-definitions/src/compatibilityDefinitions.ts
```diff
@@ -6,24 +6,33 @@
 /**
  * Oldest version of Fluid Framework client packages to support collaborating with.
  * @remarks
- * String in a SemVer format indicating a specific version of the Fluid Framework client package, or the special case of {@link @fluidframework/runtime-utils#defaultMinVersionForCollab}.
+ * A string in SemVer format indicating a specific version of the Fluid Framework client package, or the special case of {@link @fluidframework/runtime-utils#defaultMinVersionForCollab}.
  *
- * When specifying a given `MinimumVersionForCollab`, any client with a version that is greater than or equal to the specified version will be considered compatible.
+ * When specifying a given `MinimumVersionForCollab`, collaboration with any client with a version of Fluid Framework client packages that is greater is supported.
```
Josmithr: Greater than or equal?

packages/runtime/runtime-definitions/src/compatibilityDefinitions.ts
```diff
@@ -6,24 +6,33 @@
...
  *
- * Must be at least {@link @fluidframework/runtime-utils#lowestMinVersionForCollab} and cannot exceed the current version.
+ * Must be at least {@link @fluidframework/runtime-utils#lowestMinVersionForCollab} and cannot exceed the version of any Fluid Framework client package in use by the local client.
```
Josmithr, line 14: What do we do if this isn't the case? When do we (presumably) error?

CraigMacomber: Depends. We typically validate using validateMinimumVersionForCollab (linked below) in a few places.
If you are using our layering, and have multiple different versions of packages at once, things get complicated, but presumably checked somewhere inside the various layers using the correct version of validateMinimumVersionForCollab.

I don't know the exact details on where all uses of this type do validation, but generally APIs taking an untrusted MinimumVersionForCollab from users can call validateMinimumVersionForCollab if they want.

Josmithr: It would probably be nice if we failed fast consistently where possible, then we could document that plainly. But obviously beyond the scope of this PR.

Josmithr (approved):

PR #27317, opened 2026-05-14
fix(test-version-utils): drop transitive mocha from compat workspace

ChumpChief:
## Description

[AB#60288](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/60288)

The `compat-workspaces/full` nested workspace installs back-compat versions
of `@fluidframework/test-utils` so that `testApi.ts` can dynamically import
each one and harvest `TestFluidObjectFactory` for cross-version compat
tests. Old `test-utils` releases declare `mocha` as a runtime dependency,
which dragged `mocha@10.8.2` and its transitive `serialize-javascript@6.0.2`
into the lockfile across all 11 installed versions.

`serialize-javascript@6.0.2` is flagged by
[GHSA-5c6j-r48x-rmvq](https://github.com/advisories/GHSA-5c6j-r48x-rmvq)
(code injection via `RegExp.flags` / `Date.toISOString` during
`serialize()`). The advisory's only patched line is 7.x.

The rest of the repo already uses a `serialize-javascript@>=6 <7 → ^7.0.4`
override for the same CVE. That override could have been added here too,
but the cleaner option is to recognize that this workspace **never invokes
the legacy mocha runner** — `loadPackage` only reads each old test-utils'
primary export and dynamic-imports it. Mocha is dead weight in this
lockfile.

This PR adds a single pnpm override at the workspace root that drops mocha
entirely from the legacy `test-utils` snapshots:

```jsonc
"pnpm": {
  "overrides": {
    "@fluidframework/test-utils>mocha": "-"
  }
}
```

Net effect on `compat-workspaces/full/pnpm-lock.yaml`: 53 packages removed
(mocha, serialize-javascript, workerpool, yargs, chokidar, and assorted
CLI infra), `-506` / `+97` lines.

### Verification

1. Probe script that mirrors `testApi.ts`'s dynamic-import path was run
   against each installed back-compat version (0.56.0 through 2.93.0,
   17 total). Every entrypoint loaded successfully and exposed
   `TestFluidObjectFactory` as a constructor with mocha absent.
2. `pnpm test:realsvc:local` in `packages/test/test-end-to-end-tests`:
   **4872 passing, 0 failing, 504 pending** (~1 minute). All
   `compat <version>` and `compat cross-client` suites — which exercise
   the dynamic-import path — passed cleanly.

If this regresses for some version we don't have local coverage for, the
failure mode is loud: `Cannot find module 'mocha'` at compat test setup,
with a stack pointing at the dynamic `import()` of test-utils. Rolling
back is just dropping the `pnpm.overrides` block.

## Reviewer Guidance

- Sanity check that `@fluidframework/test-utils>mocha: "-"` syntax is the
  intended way to drop a transitive dep in pnpm. The repo already uses
  the same pattern for `oclif>@aws-sdk/client-cloudfront: "-"` in the
  root `package.json`.
- This only affects the *nested* `compat-workspaces/full` workspace, not
  the main repo lockfile. The main repo uses `mocha@11.x` (with the
  `serialize-javascript@>=6 <7 → ^7.0.4` override already in place).

### DIFF
--- a/packages/test/test-version-utils/compat-workspaces/full/package.json
+++ b/packages/test/test-version-utils/compat-workspaces/full/package.json
@@ -12,5 +12,13 @@
   "author": "Microsoft and contributors",
   "scripts": {
     "preinstall": "node ../scripts/only-pnpm.cjs"
+  },
+  "pnpm": {
+    "commentsOverrides": [
+      "@fluidframework/test-utils>mocha: dropped here because this workspace only loads test-utils' primary export at runtime (see test-version-utils/src/testApi.ts) — it never invokes the legacy test-utils' bundled test runner. Removing mocha eliminates serialize-javascript@6.0.2 (GHSA-5c6j-r48x-rmvq) which would otherwise be pulled in transitively across all back-compat versions of @fluidframework/test-utils."
+    ],
+    "overrides": {
+      "@fluidframework/test-utils>mocha": "-"
+    }
   }
 }
--- a/packages/test/test-version-utils/compat-workspaces/full/pnpm-lock.yaml
+++ b/packages/test/test-version-utils/compat-workspaces/full/pnpm-lock.yaml
@@ -4,6 +4,9 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+overrides:
+  '@fluidframework/test-utils>mocha': '-'
+
 importers:
 
   .: {}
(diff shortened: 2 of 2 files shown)


jason-ha (approved):
Love seeing unused deps get removed.
A mocha dependency doesn't seem like it should need to exist in a set of packages for compat testing. I suspect that there could be a refactor to split up duties for test-utils (1) helper to load some code and (2) "timeoutUtils.ts" injection.
cc @agarwal-navin, @scottn12

PR #23171, opened 2024-11-21
(docs): fix implementation issues with deployment and publish artifact pipeline

zhenmichael:
Corrected errors from previous imlementation of deploy-website, upload-json-steps, and publish-api-model-artifact pipelines. Also removed unused pipeline artifact downloads in upload-json-steps (only downloading from client build now).

PR #27319, opened 2026-05-15
build(common-utils): modernize build, lint, and test setup

ChumpChief:
## Description

Tracking: [AB#73445](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73445).

`@fluidframework/common-utils` is deprecated but lives in this repo as its own pnpm workspace. Its bespoke build/lint/test setup makes routine maintenance (CG alerts, pipeline rollouts) expensive — each change has to be done twice. This PR brings it in line with `client-utils` and `protocol-definitions`.

### Updates

- **Modern exports map.** Replaced the legacy `main`/`module`/`browser` field setup with a conditional `exports` map providing per-environment entrypoints. `indexBrowser.ts` and `indexNode.ts` now carry the full public API surface (was: thin env-specific files used only by the bundler `browser`-field remap).
- **Dropped the ESM dual-emit.** Removed `tsconfig.esnext.json` and the `lib/` output. Modern resolution already routed to CJS via `main` since there was no exports map; the ESM build was orphaned. attw 🟢 across all four modes. `client-utils` is the dual-emit replacement for downstream.
- **Node16 tsconfig layout.** Single root tsconfig extending `tsconfig.node16.json`, `.js`-extension relative imports, `tsconfig.test.node16.json`-based test configs (mirrors `client-utils`).
- **API extraction.** Replaced the single `api-extractor.json` + `api-extractor-lint.json` with a per-entrypoint `api-extractor/` config dir (browser/node lint + model-browser docs). Added `@arethetypeswrong/cli`, chained into `lint`.
- **Mocha setup.** Adopted `@fluid-internal/mocha-test-setup` via `.mocharc.cjs`, replacing the bespoke `test:report` / `test:mocha:multireport` chain.
- **Eslint config.** Replaced the hand-rolled flat config with `@fluidframework/eslint-config-fluid@^10` `recommended` (resolves [AB#59243](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/59243)). Violations silenced package-wide — deprecation mode, no source cleanups.
- **Tooling refresh.** Bumped `@fluid-tools/build-cli` + `@fluidframework/build-tools` to `^0.65.0`; refreshed `mocha`/`rewire`/`cross-env`/`jest-junit`/`ts-node`.
- **Misc cleanup.** `lodash` → `lodash.clonedeep` (named import broken under ESM); dropped `@types/base64-js` and the `bump-version` script; renamed `jest.config.js` / `jest-puppeteer.config.js` → `.cjs`.
- The `Buffer` re-export from `indexNode` is now type-only. `import { Buffer } from "@fluidframework/common-utils"; new Buffer(...)` will fail at runtime. Package is `@internal`/`@deprecated` with no in-repo consumers; Node global `Buffer` and `IsoBuffer` remain available.

## Reviewer Guidance

Commits are scoped to one logical change for bisecting. Verified locally: `build`, `test:mocha` (53), `test:jest` (19), `check:exports` (3/3), attw (🟢 all four modes), `eslint`.


common/lib/common-utils/src/base64Encoding.ts
```diff
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { IsoBuffer } from "./indexNode";
+import { IsoBuffer } from "./indexNode.js";
```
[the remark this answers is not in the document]
ChumpChief: Good catch — fixed in 490458eb9e. Splitting `base64Encoding.ts` into `base64EncodingBrowser.ts` and `base64EncodingNode.ts` (mirroring the client-utils pattern); each imports from the platform-correct `bufferBrowser`/`bufferNode` module, and `indexBrowser`/`indexNode` re-export from the matching half.

common/lib/common-utils/src/trace.ts
```diff
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { performance } from "./indexNode";
+import { performance } from "./indexNode.js";
```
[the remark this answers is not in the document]
ChumpChief: Good catch — fixed in 490458eb9e. `trace.ts` now imports `performance` directly from `./performanceIsomorphic.js`, which is already platform-isomorphic and avoids the entrypoint round-trip entirely.

common/lib/common-utils/src/index.ts
```diff
@@ -3,53 +3,8 @@
...
-	type IEventProvider,
-	type IEventThisPlaceHolder,
-	type IEventTransformer,
-	type ReplaceIEventThisPlaceHolder,
-	type TransformedEvent,
-	TypedEventEmitter,
-	type TypedEventTransform,
-} from "./typedEventEmitter";
-export { unreachableCase } from "./unreachable";
+export * from "./indexNode.js";
```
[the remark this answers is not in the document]
ChumpChief: Fixed in 9e721686be by adding `@packageDocumentation` (with a deprecation note pointing at `@fluid-internal/client-utils`) to `src/index.ts`, plus the same to `indexBrowser.ts` and `indexNode.ts` for consistency.

kian-thompson (commented):
Reviewed via Copilot CLI (deep mode). Verdict: **Approve with suggestions** — 0 CRITICAL, 0 HIGH, 4 MEDIUM.

### Findings

| Sev | Area | File | What | Fix |
|---|---|---|---|---|
| 🟡 | API Quality / Correctness | `src/indexNode.ts:17` | `Buffer` re-export changed from value+type to `export { type Buffer }`. Any `import { Buffer } from "@fluidframework/common-utils"; new Buffer(...)` will fail at runtime. No in-repo consumers and the package is `@internal`+`@deprecated`, so likely acceptable, but it isn't called out in the PR description as a behavior change. | Re-export `Buffer` as a value from `./bufferNode.js`, or explicitly note the symbol removal in the commit (`common-utils: stop pulling Node graph into browser builds`) and PR description. |
| 🟡 | API Quality | `src/base64EncodingBrowser.ts:15,25,37` (also pre-existing in `bufferBrowser.ts`, `hashFileBrowser.ts`, `bufferNode.ts`, etc.) | New file's `@deprecated` tags point at **`@fluidframework-internal/client-utils`**, which is not a real package. The real one (correctly used in the new `@packageDocumentation` block in `index*.ts`) is **`@fluid-internal/client-utils`**. `src/timer.ts:246,263` uses yet a third name (`@fluid-private/client-utils`). IntelliSense steers users to nonexistent packages. | At minimum fix the three new occurrences in `base64EncodingBrowser.ts`. Ideally do a sweep across the package: `@fluidframework-internal/client-utils` and `@fluid-private/client-utils` → `@fluid-internal/client-utils`. |
| 🟡 | Architecture | `src/index.ts:15` | Comment says "THIS FILE IS NOT ACTUALLY USED", but it remains the api-extractor `mainEntryPointFilePath` (inherited as `<projectFolder>/lib/index.d.ts` from `api-extractor-base.esm.json`, no override in `api-extractor.json`) and the type-test path `validateCommonUtilsPrevious.generated.ts` resolves `@fluidframework/common-utils` through the `exports` map. The comment will mislead future maintainers. | Update the comment to acknowledge the api-extractor + type-test role, or override `mainEntryPointFilePath` in `api-extractor.json` to point at a real shipped entrypoint (e.g. `lib/indexNode.d.ts`). |
| 🟡 | Correctness | `package.json:39-40` | Legacy fallback `main: "lib/indexBrowser.js"` / `types: "lib/indexBrowser.d.ts"` while `"type": "module"`. Resolvers that don't honor `exports` (older bundlers / legacy tooling) will try to `require()` ESM → `ERR_REQUIRE_ESM`. Previously these pointed at the CJS `dist/` build. | Either drop the legacy fields (exports-only) or point them at the CJS bundle (`./dist/indexBrowser.js` / `./dist/indexBrowser.d.ts`). |

### Suggestions (non-blocking)

- The `tsc` script copies `common/build/build-common/src/cjs/package.json` into `./dist` for CJS subpath resolution. Worth a quick parity check against `client-utils` to confirm both packages use the same mechanism.
- `c8` config in `package.json` still only includes `dist/**/*.*js` / `dist/test/**/*.*js`. With the dual ESM/CJS layout, coverage now misses `lib/**`. Consider broadening if coverage parity matters.

common/lib/common-utils/package.json
```diff
@@ -78,18 +103,17 @@
...
 		"buffer": "^6.0.3",
 		"events": "^3.1.0",
-		"lodash": "^4.18.1",
 		"sha.js": "^2.4.12"
 	},
 	"devDependencies": {
-		"@eslint/js": "^9.39.0",
-		"@fluid-tools/build-cli": "^0.63.0",
+		"@arethetypeswrong/cli": "^0.18.2",
+		"@fluid-internal/mocha-test-setup": "~2.83.0",
```
jason-ha: Quite an old version, right? Why not more recent?
And why not ^ dep?

common/lib/common-utils/package.json
```diff
@@ -11,46 +11,71 @@
...
+				"require": {
+					"types": "./dist/indexBrowser.d.ts",
+					"default": "./dist/indexBrowser.js"
+				}
+			}
+		}
 	},
-	"types": "dist/index.d.ts",
+	"main": "lib/indexBrowser.js",
+	"types": "lib/indexBrowser.d.ts",
```
jason-ha: I would not switch this around now since it is deprecated but keep with the same to reduce any risk.
In fact, if the ESM build was invalid which I have commonly found recently for packages that weren't dealt with in the large client migration, then it feels safe to drop the ESM build.
Node16 cleanup is good but leave out ESM.
"Restore" `"type": "commonjs"` (that was implicit before).
If there is a need for valid ESM build, then make ESM secondary using `fluid-tsc module`.

common/lib/common-utils/src/index.ts
```diff
@@ -6,50 +6,14 @@
...
- * NOTE: This export is remapped to export from "./indexBrowser" in browser environments via package.json.
- * Because the two files don't have fully isomorphic exports, using named exports for the full API surface
- * is problematic if that named export includes values not in their intersection.
- *
- * In a future breaking change of common-utils, we could use a named export for their intersection if we
- * desired.
- */
+// THIS FILE IS NOT ACTUALLY USED (see indexBrowser.ts and indexNode.ts for that).
+// It's only here so type-test-generator doesn't fail, until we update it to support packages that don't have an
+// index.ts file.
```
jason-ha: Seems like this file could be removed per this comment. The updated build-tools don't use index.ts. Type test generation only reads from the prior package. At build time (test time), current package's entrypoint is used.

common/lib/common-utils/package.json
```diff
@@ -11,46 +11,71 @@
...
+			"default": {
+				"import": {
+					"types": "./lib/indexBrowser.d.ts",
+					"default": "./lib/indexBrowser.js"
+				},
+				"require": {
+					"types": "./dist/indexBrowser.d.ts",
+					"default": "./dist/indexBrowser.js"
+				}
+			}
```
jason-ha, line 24: Unless a package is playing with export types like most client packages do for /beta, /internal, etc. The preference is to leave "types" out. Just specify the .js properties and let compiler do the sibling lookup. (There are cases where it does that anyway; so, just safer in general to not rely on "types" entries.)

common/lib/common-utils/package.json
```diff
@@ -11,46 +11,71 @@
...
-		"test:jest:report": "npm run test:jest -- --coverage",
-		"test:mocha": "mocha --unhandled-rejections=strict  --recursive \"dist/test/mocha/**/*.spec.*js\"",
-		"test:mocha:multireport": "cross-env FLUID_TEST_MULTIREPORT=1 npm run test:mocha",
-		"test:mocha:report": "npm run test:mocha -- -- --reporter xunit --reporter-option output=nyc/mocha-junit-report.xml",
-		"test:report": "npm run test:mocha:report && npm run test:jest:report",
-		"tsc": "tsc",
+		"test:mocha": "npm run test:mocha:esm && npm run test:mocha:cjs",
+		"test:mocha:cjs": "cross-env FLUID_TEST_MODULE_SYSTEM=CJS \"MOCHA_SPEC=dist/test/mocha/**/*.spec.*js\" mocha",
+		"test:mocha:esm": "cross-env \"MOCHA_SPEC=lib/test/mocha/**/*.spec.*js\" mocha",
+		"tsc": "fluid-tsc commonjs --project ./tsconfig.cjs.json && copyfiles -f ../../../common/build/build-common/src/cjs/package.json ./dist",
```
jason-ha: The package.json copied here is not likely correct since this has complicated export block. Better to just support CJS if the prior ESM was invalid anyway.

common/lib/common-utils/package.json
```diff
@@ -128,18 +149,23 @@
 		"tasks": {
 			"eslint": [
 				"tsc",
-				"build:test:mocha",
+				"build:esnext",
+				"build:test:mocha:esm",
+				"build:test:mocha:cjs",
 				"build:test:jest",
 				"build:test:types"
 			],
```
jason-ha, line 132: I expect this can be removed. The standard definition should work. The only real dependency is on the production build. eslint does not require that the tests builds have been done. (The production build is needed because the tests depend on it and eslint checks the tests.)

common/lib/common-utils/api-extractor.json
jason-ha, line 1: With the switch to node+browser, this should follow the client-utils pattern.
Perhaps better would be to just drop api-extractor model and report - all is deprecated and internal.

jason-ha (commented):
Main question is switch to ESM and whether instead ESM support should just be removed. More detailed ramblings in sub-issues.

common/lib/common-utils/package.json
ChumpChief: Fixed in latest

common/lib/common-utils/src/index.ts
ChumpChief: Fixed in latest

common/lib/common-utils/package.json
ChumpChief: Fixed in latest

common/lib/common-utils/package.json
ChumpChief: Fixed in latest

common/lib/common-utils/api-extractor.json
ChumpChief: Fixed in latest

common/lib/common-utils/package.json
ChumpChief: Switched to CJS primary in latest

common/lib/common-utils/package.json
ChumpChief: Switched to CJS primary in latest

packages/common/client-utils/package.json
ChumpChief, line 1: Also updated client-utils based on feedback that equally applied there.

tylerbutler (approved):

common/lib/common-utils/api-report/common-utils.legacy.beta.api.md
```diff
@@ -0,0 +1,7 @@
+## Beta API Report File for "@fluidframework/common-utils"
```
Josmithr: Given that we don't have any `@legacy` exports in this package, seems like we could simplify things by removing the legacy rollup generation and API report generation steps from the build, and remove those exports from the package.json.

ChumpChief: Fixed in latest

common/lib/common-utils/src/test/types/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-	"extends": "@fluidframework/build-common/ts-common-config.json",
+	"extends": "../../../../../../common/build/build-common/tsconfig.test.node16.json",
 	"exclude": ["dist", "node_modules"],
 	"compilerOptions": {
 		"rootDir": "./",
```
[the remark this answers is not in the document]
ChumpChief: Fixed in latest

common/lib/common-utils/package.json
```diff
@@ -126,20 +153,56 @@
...
+			"build:test:mocha:cjs": [
 				"tsc"
 			],
+			"build:test:mocha:esm": [
+				"build:esnext"
+			],
 			"build:test:types": [
-				"tsc"
+				"build:esnext"
+			],
```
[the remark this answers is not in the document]
ChumpChief: Fixed in latest

packages/common/client-utils/package.json
```diff
@@ -15,24 +15,12 @@
...
-				"import": {
-					"types": "./lib/indexBrowser.d.ts",
-					"default": "./lib/indexBrowser.js"
-				},
-				"require": {
-					"types": "./dist/indexBrowser.d.ts",
-					"default": "./dist/indexBrowser.js"
-				}
+				"import": "./lib/indexBrowser.js",
+				"require": "./dist/indexBrowser.js"
```
jason-ha: My advice was bad.
I forgot that the typetest gen code does custom resolution and currently getTypesPathFromPackage has a bug. It NEEDS to have the types entries. The failure won't show up until after publish and use of the prior package.
Instead of racing for build-tools fix, these need to go back for now. And common-utils too.

jason-ha (commented):
partial (quick) re-review

common/lib/common-utils/package.json
jason-ha: Doesn't actually switch to CJS primary for consumers in case that is what you are hoping for.
Once exports are added, then these properties are ignored by modern things.
So, as it stands ESM consumers will get ESM where they would have gotten CJS before.
I don't think we should add ESM here. Can discuss more next week.

common/lib/common-utils/pnpm-workspace.yaml
```diff
@@ -16,5 +16,27 @@ trustPolicy: no-downgrade
...
+  - "chokidar@4.0.3"
+  # semver@5.7.2 and semver@6.3.1 — legacy maintenance lines (last published 2023-07-10
+  # and 2023-04-26 respectively, by lukekarrys).
+  # Prior trusted: semver@7.5.4 (provenance, 2023-07-07, by npm-cli-ops).
+  # Different publisher account because the 5.x and 6.x maintenance lines are
+  # hand-published by lukekarrys (a long-time npm/node-semver maintainer), while the
+  # current 7.x line publishes through the npm-cli OIDC/Actions pipeline. The 5.x/6.x
+  # lines will not be retroactively re-published with provenance. Pulled in by widely-
+  # used legacy tooling that pins ^5 / ^6.
+  - "semver@5.7.2 || 6.3.1"
```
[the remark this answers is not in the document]
ChumpChief: Keeping the spaces form here — it matches the example in pnpm's official `trustPolicyExclude` documentation (`'webpack@4.47.0 || 5.102.1'`) as well as `docs/pnpm-workspace.yaml` elsewhere in the repo. Both forms are accepted by pnpm. Aligning the root `pnpm-workspace.yaml` and `build-cli`'s flag construction in the opposite direction is out of scope for this PR.

packages/common/client-utils/package.json
ChumpChief: Reverted in latest.

common/lib/common-utils/package.json
ChumpChief: Dropped ESM in latest

common/lib/common-utils/package.json
ChumpChief: Dropped ESM in latest - my main goal was to align with other packages as best we could (so by default, dual-building like most do) but the ESM build isn't crucial to most of the gains (aligning the dependencies, build tools, etc.).

common/lib/common-utils/package.json
```diff
@@ -11,46 +11,52 @@
...
-		"check:release-tags": "api-extractor run --config ./api-extractor-lint.json",
+		"check:are-the-types-wrong": "attw --pack .",
+		"check:exports": "concurrently \"npm:check:exports:*\"",
+		"check:exports:bundle-release-tags": "api-extractor run --config api-extractor/api-extractor-lint-bundle.json",
+		"check:exports:cjs:indexBrowser": "api-extractor run --config api-extractor/api-extractor-lint-indexBrowser.cjs.json",
+		"check:exports:cjs:indexNode": "api-extractor run --config api-extractor/api-extractor-lint-indexNode.cjs.json",
 		"ci:build": "fluid-build . --task ci:build",
-		"ci:build:docs": "api-extractor run --typescript-compiler-folder ./node_modules/typescript && copyfiles -u 1 \"./_api-extractor-temp/doc-models/*\" ../../../_api-extractor-temp/",
-		"ci:test": "npm run test:report",
+		"ci:build:docs": "api-extractor run --config api-extractor/api-extractor-model-browser.json",
```
jason-ha, line 44: Does the copy up to root not need to happen any more?

common/lib/common-utils/package.json
```diff
@@ -11,46 +11,52 @@
...
+		"ci:test": "npm test",
 		"ci:test:coverage": "npm run test:coverage",
-		"clean": "rimraf --glob _api-extractor-temp dist lib \"**/*.tsbuildinfo\" \"**/*.build.log\" nyc",
+		"clean": "rimraf --glob _api-extractor-temp dist \"**/*.tsbuildinfo\" \"**/*.build.log\" nyc",
```
jason-ha, line 47: An open discussion topic in my head, when removing an output folder from generation, should it be left in cleanup for some period of time?
If not, as soon as someone picks up the change and does clean, any remnants will remain.
I tend to leave them in the clean (least harm), but won't block a PR if it is removed.

jason-ha (approved):
Looks overall good - I will approve with one question about making sure ci:build:docs doesn't need to copy to root.

common/lib/common-utils/package.json
ChumpChief: Traced this end-to-end — the copy is safe to drop:

1. **common-utils CI artifact** is unaffected. `build-npm-package.yml` copies from `${buildDirectory}/_api-extractor-temp` (= `common/lib/common-utils/_api-extractor-temp/`), which is exactly where api-extractor writes the doc model natively. The copy-to-repo-root was just a duplicate this pipeline never reads.

2. **Published api-model artifact** is unaffected. `publish-api-model-artifact.yml` only downloads from `Build - client packages` — `Build - common-utils` was explicitly removed from that aggregation in #23171 (Nov 2024) as "unused", so the per-package copy step has effectively been dead code for CI since then.

3. **Local docs preview** is the only place it ever still mattered: with `LOCAL_API_DOCS=true`, the docs site picks up every `*.api.json` in repo-root `_api-extractor-temp/doc-models/` (no allowlist), and the root `build:gendocs:client:*` scripts don't glob `common/**` — so the per-package copy was the last remaining way `common-utils.api.json` could end up in a local preview. Since the package is deprecated and not part of the api-model artifact publish, we probably don't need to worry about the local preview.

common/lib/common-utils/package.json
ChumpChief: I usually go ahead with the alignment to current implementation just because I worry that we'll forget to come back and fix it up (and it's hard to know what's the appropriate waiting time to feel confident that everyone has gotten their `pnpm clean` in, maybe 2-3 weeks?).  Similar stale artifacts tend to exist when e.g. packages are removed entirely or file removals/renames paired with .gitignore changes, so it's already somewhat of a risk developers need to account for when sync'ing.

common/lib/common-utils/package.json
jason-ha: Sounds good. Seems like 3 remaining may have been an oversight during 2 change.

common/lib/common-utils/package.json
jason-ha: The .gitignore case has visibility and has well-known scoped, clean paths.
After I wrote my thoughts (usually helps to write things out), I was feeling better about leaving the old artifacts in the clean script even if they aren't representative in the current config. clean becomes more "timeless". Maybe we never trim and that is fine.

I was glad I didn't remove a folder from cleanup in recent change I was working on yesterday. I switched devices and didn't clean before switching to the branch. I was able to do scoped clean efficiently.

I think for the most part we expect clean to get us pretty clean. We have git clean to be more aggressive. If script clean gets old stuff, that is better than not. It would be a problem if we cleaned something that shouldn't be but probably, we'd discover that.

For this guy, totally good to drop lib. The whole thing is going away anyway and doesn't get touched often enough.

jatgarg (approved):

MarioJGMsoft (approved):

PR #27320, opened 2026-05-15
Update syncpack to v15 and move pnpm overrides to pnpm workspace file

alexvy86:
## Description

Update syncpack to v15 in the client release group so we can move pnpm overrides to `pnpm-workspace.yaml` file (also done here). Before that version, syncpack didn't fully support processing overrides in the workspace file.

## Reviewer Guidance

I confirmed that making changes to deps in `catalogs:` are flagged correctly per the syncpack configuration. Also of note, today syncpack config in the client release group already ignores everything in pnpm overrides. I confirmed that updating that config and doing something "bad" in those deps in their new location, gets flagged correctly.


.syncpackrc.yml
```diff
@@ -87,6 +87,9 @@ semverGroups:
   - label: "Must use exact dependency ranges"
     dependencies:
       - "@tiny-calc/*"
+      # @tylerbu/markdown-magic is currently pinned to a tagged prerelease (e.g. 2.4.0-tylerbu-1); a caret range on a
```
alexvy86, line 90: Note for reviewers: this seems to have been a bug fixed in syncpack 15, not sure why it wasn't getting flagged before.

jason-ha (approved):
:)
Note that #27307 also moves these and conflict will need resolved. But that is blocked per so I think this is good to proceed.

PR #27298, opened 2026-05-13
feat(build-cli): rework bundle size check for inner-dev-loop use

ChumpChief:
Adds `flub check bundleSize` so engineers can check how their local changes affect bundle sizes without waiting for CI. Implements [AB#73299](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73299), one step in the broader rework of bundle-size diffing for PRs ([AB#56981](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/56981)).

## Highlights

- **Inner-dev-loop bundle check.** New `flub check bundleSize` compares locally-built `analyzer.json` against the CI baseline and prints per-asset deltas (parsed + gzip). `--json` for tooling.
- **Auto-detected baseline.** No flags needed in the common case: walks `git config` for remotes pointing at `microsoft/FluidFramework` and picks the one whose `<remote>/main` tip is freshest locally — useful when a dev keeps separate `upstream` / `upstream_push` remotes and one falls behind.
- **`--target <ref>` for overrides.** Replaces the previous `--targetBranch`. Accepts `<remote>/<branch>`, a SHA, or anything `git merge-base` does; the baseline commit is `git merge-base <target> HEAD`.
- **ADO pipelines no longer compute the diff.** Permanent split of concerns: pipelines just publish `bundleAnalyzerJson`; the diff itself moves to the new CLI (local dev) and a followup GitHub Actions workflow (PR comments).
- **Library cleanup.** `library/bundleSizeDiff/` → `library/bundleSize/`; `getAzureDevopsApi` promoted to `library/azureDevops/` (generic ADO concern, was already imported by non-bundle consumers); drop legacy `BundleBuddy` branding.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #27321, opened 2026-05-15
chore(build-cli): reshape bundleSize library to pure-function primitives

ChumpChief:
Implements [AB#73449](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73449) — followup to PR #27298 ([AB#73299](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73299)), continuing the bundle-size rework. Replaces the class-based `ADOSizeComparator` with pure-function primitives, redesigns the comparison types, and pulls generic helpers out of the bundleSize library.

## Highlights

- **`bundleSize/ADO/` and `ADOSizeComparator` go away.** New primitives (`compareJsonReportsByPackage`, `extractAnalyzerJsonsFromArtifact`, `readAnalyzerJsonsFromFileSystem`, `sourcePackageFromAnalyzerPath`) compose inside `check bundleSize`.
- **Type redesign.** `BundleData` / `BundlesComparison` / `PackageComparison`. Field presence on each bundle entry encodes added vs removed vs both-sides — previously silently dropped asymmetric assets.
- **Generic helpers promoted out of bundleSize/.** `library/azureDevops/getArtifactForCommit.ts` (find a build for a commit, download its artifact). `library/git/pickFreshestRemote.ts` (PR #27298's canonical-remote auto-detect, now parameterized via a URL-matcher callback).
- **Drop the "command never throws" contract.** With the diff moving to a follow-up GH Actions workflow rather than a CI pipeline, the `kind: "error"` result variant was over-engineered. Use oclif's `this.error()` for fatals — matches the rest of the codebase (`bump.ts`, `list.ts`, etc.).
- **`downloadArtifact` HTTP-status check.** Missing artifacts now surface "Failed to download artifact X from build Y (HTTP 404)" instead of fflate's cryptic "invalid zip data."

## Followup
Converge `getArtifactForCommit` / `getBaselineBuildMetrics` / `getBuildArtifactForSpecificBuild` around a shared "find build by selector + download artifact" pattern ([AB#73595](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73595)).


build-tools/packages/build-cli/src/library/azureDevops/getArtifactForCommit.ts
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Wrapper around the unwieldy positional signature of ADO's `getBuilds`.
+ */
+async function getRecentBuilds(
+	adoApi: WebApi,
+	project: string,
+	definitionId: number,
+): Promise<Build[]> {
+	const buildApi = await adoApi.getBuildApi();
+	return buildApi.getBuilds(
```
TommyBrosman: This is a little hard to follow with the long list of `undefined` inputs. Can you use this wrapper? https://github.com/microsoft/FluidFramework/blob/abcec227ff7db77508b0b198f6fcfbfec4a2db07/build-tools/packages/build-cli/src/library/azureDevops/utils.ts#L42

build-tools/packages/build-cli/src/library/bundleSize/extractAnalyzerJsonsFromArtifact.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Walks a downloaded artifact's contents, finds every `analyzer.json`, parses
+ * it, and keys the results by source package.
+ */
+export function extractAnalyzerJsonsFromArtifact(
+	contents: ArtifactContents,
+): AnalyzerJsonByPackage {
+	const result: AnalyzerJsonByPackage = new Map();
+	for (const [relativePath, bytes] of Object.entries(contents)) {
+		const sourcePackage = sourcePackageFromAnalyzerPath(relativePath);
+		if (sourcePackage === undefined) continue;
```
TommyBrosman, line 24: Does it make sense to silently ignore this case? Are there legitimate cases where the relativePath is not an analyzer path?

build-tools/packages/build-cli/src/library/bundleSize/sourcePackageFromAnalyzerPath.ts
```diff
@@ -0,0 +1,21 @@
...
+ * package name. Returns `undefined` for paths that don't match.
+ *
+ * Slashes are normalized first so the same logic handles both Windows and
+ * POSIX path separators.
+ */
+export function sourcePackageFromAnalyzerPath(relativePath: string): string | undefined {
+	const normalized = relativePath.replace(/\\/g, "/");
+	if (!normalized.endsWith(analyzerJsonSuffix)) return undefined;
+	const sourcePackage = normalized.slice(0, -analyzerJsonSuffix.length);
+	return sourcePackage.length === 0 ? undefined : sourcePackage;
```
TommyBrosman, line 22: What case is this? If the sourcePackage was just analyzer.json, is that valid?

build-tools/packages/build-cli/src/library/azureDevops/getArtifactForCommit.ts
ChumpChief: Good catch — done in `f0535fe009`. Likely an oversight on my part when I was coalescing the packages.

build-tools/packages/build-cli/src/library/bundleSize/extractAnalyzerJsonsFromArtifact.ts
ChumpChief: The `bundleAnalyzerJson` artifact's "only analyzer.json files" contract is convention rather than pipeline-enforced — extra files getting published alongside (a manifest, a README, etc.) is plausible and shouldn't break the comparison. Added a TSDoc note explaining this in `f0535fe009`. The caller in `bundleSize.ts` validates the resulting map is non-empty (the `baselineJsons.size === 0` check), so a complete failure still surfaces clearly to the user.

build-tools/packages/build-cli/src/library/bundleSize/sourcePackageFromAnalyzerPath.ts
ChumpChief: The case is `relativePath === "/analyzer.json"` exactly — leading slash, no package directory. After normalization it still ends with `/analyzer.json`, so the `endsWith` check passes, then `slice` produces `""`. (Bare `analyzer.json` with no leading slash is rejected earlier by the `endsWith` check.)

A bare `/analyzer.json` isn't valid in the comparison: the directory IS the source-package key, and returning `""` would let it pollute the package-keyed Map alongside real packages (and produce nonsense like `  :` headers in the formatted output). Not expected from the artifact in practice, just defensive. Added a comment in `f0535fe009`.

build-tools/packages/build-cli/src/library/azureDevops/getArtifactForCommit.ts
```diff
@@ -0,0 +1,111 @@
...
+		maxBuildsPerDefinition: recentBuildsToFetch,
+	});
+	const buildId = findBuildIdForCommit(builds, commit);
+
+	try {
+		return await downloadArtifact(adoApi, project, buildId, artifactName);
+	} catch (e) {
+		throw new Error(`Could not download artifact "${artifactName}" for commit ${commit}`, {
+			cause: e,
+		});
```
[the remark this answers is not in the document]
ChumpChief: Confirmed — oclif's renderer only reads `.message`, ignores `.cause`. Dropped the wrap in `ce549598cb`; the `Baseline commit: <sha>` log line still gives commit context.

build-tools/packages/build-cli/src/commands/check/bundleSize.ts
```diff
@@ -59,76 +95,88 @@ export default class CheckBundleSize extends BaseCommand<typeof CheckBundleSize>
...
+		// Append the `pnpm bundle-analysis:collect` hint only on the default
+		// path — overrides are populated from some source we don't know about.
+		const hint =
+			localReportPath === defaultLocalReportPath
+				? " Run `pnpm bundle-analysis:collect` to populate it."
+				: "";
+		if (localResult.kind === "error") {
+			this.error(
+				`Local bundle report path "${localReportPath}" is missing or not a directory.${hint}`,
+			);
```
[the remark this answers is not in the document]
ChumpChief: Real concern, but the root cause is `BaseCommand.error()` returning under `--quiet` — a violation of oclif's throw/exit contract that affects ~93 call sites across build-cli. Filed AB#74372 to fix it at the base class. Leaving this PR's three sites alone since a local workaround here would just make `bundleSize` an island.

build-tools/packages/build-cli/src/commands/check/bundleSize.ts
```diff
@@ -59,76 +95,88 @@ export default class CheckBundleSize extends BaseCommand<typeof CheckBundleSize>
...
+			// Source-of-truth: tools/pipelines/build-bundle-size-artifacts.yml.
+			definitionId: 48,
+			project: "public",
+		});
+
+		const baselineJsons = extractAnalyzerJsonsFromArtifact(artifactContents);
+		if (baselineJsons.size === 0) {
+			this.error(
+				`Baseline artifact contains no analyzer.json files for commit ${baselineCommit}.`,
+			);
```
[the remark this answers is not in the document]
ChumpChief: Real concern, but the root cause is `BaseCommand.error()` returning under `--quiet` — a violation of oclif's throw/exit contract that affects ~93 call sites across build-cli. Filed AB#74372 to fix it at the base class. Leaving this PR's three sites alone since a local workaround here would just make `bundleSize` an island.

build-tools/packages/build-cli/src/commands/check/bundleSize.ts
```diff
@@ -59,76 +95,88 @@ export default class CheckBundleSize extends BaseCommand<typeof CheckBundleSize>
...
+		if (localResult.kind === "error") {
+			this.error(
+				`Local bundle report path "${localReportPath}" is missing or not a directory.${hint}`,
+			);
+		}
+		const compareJsons = localResult.data;
+		if (compareJsons.size === 0) {
+			this.error(
+				`Local bundle report directory "${localReportPath}" contains no analyzer.json files at the expected \`<package>/analyzer.json\` layout.${hint}`,
+			);
```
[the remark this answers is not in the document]
ChumpChief: Real concern, but the root cause is `BaseCommand.error()` returning under `--quiet` — a violation of oclif's throw/exit contract that affects ~93 call sites across build-cli. Filed AB#74372 to fix it at the base class. Leaving this PR's three sites alone since a local workaround here would just make `bundleSize` an island.

TommyBrosman (approved):

PR #25292, opened 2025-08-22
Make mocha configuration more consistent and centralized

CraigMacomber:
## Description

Currently a lot of mocha configuration logic is duplicated across many packages.
Additionally, a lot of this configuration is unnecessarily different in non-obvious ways.

This change adds some defaults to the shared mocha configuration, then relies on those defaults in nearly identical ways across the majority of packages.

This reduces the duplication of policy logic, as well as making the configurations much more consistent and concise.

### Changes to the Common Config

Mocha's `spec` now defaults to `lib/test` with option to override via "MOCHA_SPEC" environment variable (previously many packages implemented this same pattern). Packages with dual CommonJS and ESM builds set "MOCHA_SPEC" to distinguish runs.

This default spec does not include a filter to `*.spec.js` so many packages which previously had such a filter now load all files, except the newly ignored `*.tool.[m,c]?js` files. This is done to error on the side of loading too many files rather than too few to reduce the risk of tests not running due to different file names (there were 3 cases of this, found and fixed as part of this work). In the future, `*.spec.js` will be the default and enforced.

`getFluidTestMochaConfig` previously had no doc comment: one has been added, and its newly linked readme has been updated as well.

### Other Changes

Many redundant --recursive flags were removed.

Inline use of __dirname for more concise configs.

## Breaking Changes

For packages outside of client workspace which use this config, integrating the config into them may need updates.

There are three such usages:

- tinylicious using "~2.0.5"
- @fluid-tools/api-markdown-documenter using "~2.23.0"
- @fluid-tools/benchmark using "^2.51.0" (updated to use "~2.51.0" in this change)

None of these do this integration regularly and are all already well behind.

## Reviewer Guidance

The review process is outlined on [this wiki page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

PR #27322, opened 2026-05-15
test(client-local-server-*tests): test ESM

jason-ha:
Stop "lying" that ESM is tested (there was not CJS test label) and actually test ESM that we care more about. Note that all server packages (server workspaces) are CJS only.

1. Update packages to type "module" (ESM)
   - ran ts2esm that added missing .js extensions to import spec
   - dirname.cts removed in favor of ESM only import.meta.url use.
2. Adopt standardized test patterns
   - Move files spec from mocha command lines (was ignored) to config file.
   - Move test support files under src/test instead of just src/.
   - Move stress package tsconfig.json to under src/test.
3. Fix sequence build deps to allow build-and-test:unit:esm from stress package.


alexvy86:
> Move files spec from mocha command lines (was ignored) [...]

Curious, why was it ignored?

jason-ha:
> > Move files spec from mocha command lines (was ignored) [...]
> 
> Curious, why was it ignored?

Yes. As of #25292 most mocha command line test specifications are ignored.

packages/test/local-server-stress-tests/package.json
```diff
@@ -28,7 +29,8 @@
 		"lint:fix": "fluid-build . --task eslint:fix --task format",
 		"test": "npm run test:mocha",
 		"test:coverage": "c8 npm test",
-		"test:mocha": "mocha \"lib/test/**/*.spec.*js\"  --exit",
+		"test:mocha": "npm run test:mocha:esm",
+		"test:mocha:esm": "mocha",
```
MarioJGMsoft, line 33: I thought that mocha was only compatible with CJS? Why does this work here?

jason-ha: We use mocha to test both. jest has issues with ESM - it can't as easily inject / manipulate things.

MarioJGMsoft (approved):
LGTM!

PR #27324, opened 2026-05-15
test(tree): Fix and re-enable text op size benchmarks

Josmithr:
The allowed variance was not accounting for the largest test variant, which is not run in correctness mode (and therefore was not caught by CI).

### DIFF
--- a/packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
+++ b/packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
@@ -42,8 +42,7 @@ function makeTestString(length: number): string {
 	return "ab".repeat(Math.ceil(length / 2)).slice(0, length);
 }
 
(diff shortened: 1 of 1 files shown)


packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
```diff
@@ -340,7 +339,7 @@ describe.skip("TextDomain benchmarks", () => {
 							// TODO: Investigate whether this approximately-constant behavior is
```
CraigMacomber, line 339: Do we have a work item for this?

CraigMacomber (approved):

packages/dds/tree/src/test/shared-tree/textDomain.integration.bench.ts
Josmithr: Not to my knowledge. I can make one though.

PR #27327, opened 2026-05-16
feat(datastore): Decorate and log errors thrown during DataStore entrypoint initialization

markfields:
## Description

Component load failures are one of the top  scenarios our partners need to be able to monitor and diagnose in production. This PR closes a long-standing gap where errors thrown during entryPoint initialization (e.g. for `PureDataObject`, that includes the initalization hooks) are not logged to FF tables.

While host applications can and do log these errors, there is context that's only readily available within FF that is really important for investigation - All the common ContainerRuntime props (notably clientType - is it the Summarizer?), as well as the DataStore's packagePath (indicates the component type) and id.

This PR logs an error event with all these props, and additionally annotates the error object* with the dataStore package/id info.

\* _Note: In testing I found that these telemetry props get stripped off on the way up by roundtripping through `exceptionToResponse` and `responseToException`.  A bit curious to look into that flow and if we could cut out those tranforms._

### DIFF
--- a/packages/runtime/container-runtime/src/dataStoreContext.ts
+++ b/packages/runtime/container-runtime/src/dataStoreContext.ts
@@ -75,6 +75,7 @@ import { channelsTreeName } from "@fluidframework/runtime-definitions/internal";
 import {
 	addBlobToSummary,
 	isSnapshotFetchRequiredForLoadingGroupId,
+	dataStoreLoadTelemetryProps,
 } from "@fluidframework/runtime-utils/internal";
 import {
 	DataProcessingError,
@@ -587,12 +588,7 @@ export abstract class FluidDataStoreContext
 					error,
 					"realizeFluidDataStoreContext",
 				);
-				errorWrapped.addTelemetryProperties(
-					tagCodeArtifacts({
-						fullPackageName: this.pkg?.join("/"),
-						fluidDataStoreId: this.id,
-					}),
-				);
+				errorWrapped.addTelemetryProperties(dataStoreLoadTelemetryProps(this));
 				this.mc.logger.sendErrorEvent({ eventName: "RealizeError" }, errorWrapped);
 				throw errorWrapped;
 			});
--- a/packages/runtime/container-runtime/src/dataStoreContext.ts
+++ b/packages/runtime/container-runtime/src/dataStoreContext.ts
@@ -588,7 +588,9 @@ export abstract class FluidDataStoreContext
 					error,
 					"realizeFluidDataStoreContext",
 				);
-				errorWrapped.addTelemetryProperties(dataStoreLoadTelemetryProps(this));
+				errorWrapped.addTelemetryProperties(
+					dataStoreLoadTelemetryProps({ id: this.id, packagePath: this.pkg ?? [] }),
+				);
 				this.mc.logger.sendErrorEvent({ eventName: "RealizeError" }, errorWrapped);
 				throw errorWrapped;
 			});
--- a/packages/runtime/datastore/src/dataStoreRuntime.ts
+++ b/packages/runtime/datastore/src/dataStoreRuntime.ts
@@ -509,12 +509,18 @@ export class FluidDataStoreRuntime
 		this.entryPoint = new FluidObjectHandle<FluidObject>(
 			new LazyPromise(async () =>
 				provideEntryPoint(this).catch((error) => {
+					let packagePath: readonly string[] = [];
+					try {
+						packagePath = this.dataStoreContext.packagePath;
+					} catch {
+						// `packagePath` may not be available during early load failures.
+					}
 					const errorWrapped = DataProcessingError.wrapIfUnrecognized(
 						error,
 						"entryPointInitialization",
 					);
 					errorWrapped.addTelemetryProperties(
-						dataStoreLoadTelemetryProps(this.dataStoreContext),
+						dataStoreLoadTelemetryProps({ id: this.dataStoreContext.id, packagePath }),
 					);
 					this.mc.logger.sendErrorEvent(
 						{ eventName: "EntryPointInitializationFailure" },
--- a/packages/runtime/datastore/src/dataStoreRuntime.ts
+++ b/packages/runtime/datastore/src/dataStoreRuntime.ts
@@ -506,7 +506,25 @@ export class FluidDataStoreRuntime
 		}
 
 		this.entryPoint = new FluidObjectHandle<FluidObject>(
-			new LazyPromise(async () => provideEntryPoint(this)),
+			new LazyPromise(async () =>
+				provideEntryPoint(this).catch((error) => {
+					const errorWrapped = DataProcessingError.wrapIfUnrecognized(
+						error,
+						"entryPointInitialization",
+					);
+					errorWrapped.addTelemetryProperties(
+						tagCodeArtifacts({
+							fullPackageName: this.dataStoreContext.packagePath.join("/"),
+							fluidDataStoreId: this.id,
+						}),
+					);
+					this.mc.logger.sendErrorEvent(
+						{ eventName: "EntryPointInitializationFailure" },
+						errorWrapped,
+					);
+					throw errorWrapped;
+				}),
+			),
 			"",
 			this.objectsRoutingContext,
 		);
--- a/packages/runtime/datastore/src/dataStoreRuntime.ts
+++ b/packages/runtime/datastore/src/dataStoreRuntime.ts
@@ -78,6 +78,7 @@ import {
 	exceptionToResponse,
 	generateHandleContextPath,
 	processAttachMessageGCData,
+	dataStoreLoadTelemetryProps,
 	toFluidHandleInternal,
 	unpackChildNodesUsedRoutes,
 	toDeltaManagerErased,
@@ -513,10 +514,7 @@ export class FluidDataStoreRuntime
 						"entryPointInitialization",
 					);
 					errorWrapped.addTelemetryProperties(
-						tagCodeArtifacts({
-							fullPackageName: this.dataStoreContext.packagePath.join("/"),
-							fluidDataStoreId: this.id,
-						}),
+						dataStoreLoadTelemetryProps(this.dataStoreContext),
 					);
 					this.mc.logger.sendErrorEvent(
 						{ eventName: "EntryPointInitializationFailure" },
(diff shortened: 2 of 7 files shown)


packages/runtime/container-runtime/src/dataStoreContext.ts
```diff
@@ -587,12 +588,7 @@ export abstract class FluidDataStoreContext
 					error,
 					"realizeFluidDataStoreContext",
 				);
-				errorWrapped.addTelemetryProperties(
-					tagCodeArtifacts({
-						fullPackageName: this.pkg?.join("/"),
-						fluidDataStoreId: this.id,
-					}),
-				);
+				errorWrapped.addTelemetryProperties(dataStoreLoadTelemetryProps(this));
```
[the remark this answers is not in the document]
markfields: Fixed

packages/runtime/datastore/src/dataStoreRuntime.ts
```diff
@@ -506,7 +507,22 @@ export class FluidDataStoreRuntime
...
 
 		this.entryPoint = new FluidObjectHandle<FluidObject>(
-			new LazyPromise(async () => provideEntryPoint(this)),
+			new LazyPromise(async () =>
+				provideEntryPoint(this).catch((error) => {
+					const errorWrapped = DataProcessingError.wrapIfUnrecognized(
+						error,
+						"entryPointInitialization",
+					);
+					errorWrapped.addTelemetryProperties(
```
markfields, line 522: Note: In testing I found that these telemetry props get stripped off on the way up by roundtripping through `exceptionToResponse` and `responseToException`.  A bit curious to look into that flow and if we could cut out those tranforms.

shlevari: Does this have to do with privacy concerns? I know there can sometimes be policies to strip all error information away in case it contains user data or identifiable information.

markfields: Nope, it has to do with legacy architectural direction (URI-based routing within and across containers)

markfields: These properties are correctly tagged as "code artifacts", so hosts can decide if they're safe to log under different circumstances

packages/runtime/container-runtime/src/dataStoreContext.ts
ChumpChief: Passing `this` tends to be a code smell, probably prefer either passing the specific members needed or else make it a member method if it truly needs access to `this`.

packages/runtime/runtime-utils/src/dataStoreHelpers.ts
```diff
@@ -140,6 +145,30 @@ export function createResponseError(
 	};
 }
 
+/**
+ * Returns the canonical set of code-artifact-tagged telemetry properties identifying a data store.
+ * Use this anywhere a data store identity needs to appear in telemetry, so all such logs use
+ * consistent property names.
+ * @internal
+ */
+export function dataStoreLoadTelemetryProps(
```
ChumpChief: If this is bound to the context, probably keep it colocated with the context (same package)

packages/runtime/runtime-utils/src/dataStoreHelpers.ts
```diff
@@ -140,6 +145,30 @@ export function createResponseError(
...
+	context: IFluidDataStoreContext,
+): ITelemetryPropertiesExt {
+	// `packagePath` is typed as always defined, but its implementation asserts on the package being
+	// set — and during early load-failure paths (e.g. realize() rejecting before pkg is read) it
+	// can throw. Swallow that so error decoration never replaces the underlying error.
+	let fullPackageName: string | undefined;
+	try {
+		fullPackageName = context.packagePath.join("/");
+	} catch {
+		// `packagePath` is unset during early failures; leave fullPackageName undefined.
```
ChumpChief: This would read easier and be more self-documenting as an explicit check for that condition rather than try/catch.

packages/runtime/datastore/src/dataStoreRuntime.ts
```diff
@@ -506,7 +507,22 @@ export class FluidDataStoreRuntime
 		}
 
 		this.entryPoint = new FluidObjectHandle<FluidObject>(
-			new LazyPromise(async () => provideEntryPoint(this)),
+			new LazyPromise(async () =>
+				provideEntryPoint(this).catch((error) => {
+					const errorWrapped = DataProcessingError.wrapIfUnrecognized(
```
ChumpChief, line 518: Great idea!

ChumpChief (commented):
Love wrapping at the exit to custom code!  Few organizational/API suggestions.

packages/runtime/container-runtime/src/dataStoreContext.ts
markfields: Didn't want to muck up the legacy-beta IFluidDataStoreContext interface.  I can consider passing in just the props we need off the context.

packages/runtime/runtime-utils/src/dataStoreHelpers.ts
markfields: This is an idiosyncracy of this property, which is that it throws if the underlying property is undefined, and that underlying property is not exposed.

packages/runtime/container-runtime/src/dataStoreContext.ts
ChumpChief: That would be my preference too, passing big/powerful objects for small tasks is also usually a code smell :-D

packages/runtime/runtime-utils/src/dataStoreHelpers.ts
ChumpChief: Oh right, how awful :-P

packages/runtime/runtime-utils/src/dataStoreHelpers.ts
markfields: Discussed offline.  It's used in two different packages so going to leave in runtime-utils

packages/runtime/datastore/src/dataStoreRuntime.ts
agarwal-navin, line 518: I am wondering if `DataProcessingError` is appropriate here. This runs application entry point logic and has nothing to do with Fluid right? So, maybe this is a `UsageError` or just a generic error?

packages/test/test-end-to-end-tests/src/test/summarization/summaries.spec.ts
```diff
@@ -296,60 +296,70 @@ describeCompat("Summaries", "NoCompat", (getTestObjectProvider, apis) => {
...
-			dataStoreFactory1,
-			undefined,
-			ContainerRuntimeFactoryWithDefaultDataStore,
-			registryStoreEntries,
-		);
-		await assert.doesNotReject(
-			summarizeNow(createSummarizerResult.summarizer, "test"),
-			"Summarizing should not throw",
-		);
+			// Create a summarizer for the container and do a summary shouldn't throw.
```
agarwal-navin: Interesting test. I know you are not changing this but this comment isn't clear. I would suggest updating this to say that the data store is realized / loaded during summarization but is not initialized and hence summarization succeeds.

packages/test/test-end-to-end-tests/src/test/summarization/summaries.spec.ts
```diff
@@ -296,60 +296,70 @@ describeCompat("Summaries", "NoCompat", (getTestObjectProvider, apis) => {
...
-				const runtime = (createSummarizerResult.summarizer as any).runtime as ContainerRuntime;
-				const dsEntryPoint = await runtime.getAliasedDataStoreEntryPoint("default");
-				await dsEntryPoint?.get();
-			},
-			(e: Error) =>
-				e.message ===
-				"Non interactive/summarizer client's data object should not be initialized",
-			"Loading data store in summarizer did not throw as it should, or threw an unexpected error.",
-		);
+			// In summarizer, load the data store should fail.
```
agarwal-navin: Update this comment to say that this throws an error and also logs the error in `itExpects`

packages/test/test-end-to-end-tests/src/test/summarization/summaries.spec.ts
```diff
@@ -296,60 +296,70 @@ describeCompat("Summaries", "NoCompat", (getTestObjectProvider, apis) => {
...
+			);
 
-		// Load second container, load the data store will also call initializingFromExisting and succeed.
-		const container2 = await provider.loadContainer(runtimeFactory);
-		await assert.doesNotReject(
-			container2.getEntryPoint(),
-			"Initial creation of container and data store should succeed.",
-		);
-	});
+			// Load second container, load the data store will also call initializingFromExisting and succeed.
```
agarwal-navin, line 357: An observation, don't need to change this I guess - This step seems unnecessary.

packages/runtime/datastore/src/dataStoreRuntime.ts
agarwal-navin: Thinking more about this - we should separate out the path that loads the data store / DDS data from the path that loads the application logic. And we should be able to tell these apart from telemetry. I would suggest to not use data process error for the latter because that should not cause any problems with the data (it does fail summaries today but that is an unexpected scenario and will be fixed soon)

markfields: I think we have different impression of what DPE means. I think it's any error that happens on a critical load or process path, regardless of the source.  Asking Copilot to audit usage, it buckets about 70% as FF-root-cause and 30% as app callback or hook code.

Practically speaking it doesn't really matter, because we'd expect an application to want to monitor these closely.  Root cause is a second step.

PErhaps we can subdivide the errorType like you're thinking, but I think that's a bigger effort.

As for the feasibility of doing so here (capturing DDS load v. app code), I don't think the entryPoint abstraction affords us that.  That really depends on the implementation.  Even PureDataObject doesn't know.  It happens that we typically do DDS (at least root) loading in a shared base class like DataObject or MultiFormatDataObject (moved to another repo), but even that is not a hard requirement.

Let's keep talking and we can refine this, I'm probably glossing over something you're seeing, but I'm confident that this is a reasonable use of DPE for now.

packages/test/test-end-to-end-tests/src/test/summarization/summaries.spec.ts
markfields: Yeah I found these comments a little off too, happy to fix them up

markfields:
Marking as release-blocking, this is an important repair item from a recent incident faced by some first party partners.

agarwal-navin (approved):

PR #26617, opened 2026-03-02
build: update @biomejs/biome to 2.x

tylerbutler:
## Summary

- Update `@biomejs/biome` from `~1.9.x` to `~2.4.5` across all packages in both the client and build-tools release groups
- Remove temporary `@biomejs/biome` pnpm overrides from both root and build-tools `package.json` that were used to stage the upgrade
- Remove syncpack workarounds: the biome ignore rule in `build-tools/syncpack.config.cjs` and the biome entry from version compatibility workarounds in `syncpack.config.cjs`
- Apply biome 2.x formatting changes: removal of unused generic type parameters in `exposedInternalUtilityTypes.ts`, `typeUtils.ts`, and `latestValueTypes.ts`
- Collateral lockfile changes: `acorn` 8.15.0 -> 8.16.0 and `acorn-walk` 8.3.4 -> 8.3.5 resolved as part of the lockfile update

PR #27335, opened 2026-05-18
Tidy up root workspace

CraigMacomber:
## Description

Tidy up root workspace.

- Adds some blank lines for clarity
- Removes `"@biomejs/biome": "~2.4.5"` which was supposed to be removed when we adopted that version of biome ( https://github.com/microsoft/FluidFramework/pull/26617 )
- Fix trustPolicyExclude syntax for semver (see added comment for details)

## Reviewer Guidance


pnpm-workspace.yaml
```diff
@@ -92,83 +100,80 @@ catalogs:
...
+  qs: ^6.15.0
 
-  # simple-git: overridden to ^3.32.3 to resolve a CG alert.
-  "simple-git": "^3.32.3"
+  # Resolve a CG alert.
+  simple-git: ^3.32.3
 
-  # systeminformation: overridden to ^5.31.0 to resolve command injection vulnerabilities.
-  "systeminformation": "^5.31.0"
+  # Resolve command injection vulnerabilities.
```
CraigMacomber, line 132: Most of these entries redundantly state how it is an override (redundant with this being in the overrides section), what is being overridden, and to what version.

I shorted the phrasing of several of these to remove the redundant cruft.
These comments were written back when they lived in the package.json file and were much harder to associate with what they applied to. Now that these are in yaml, they can be simplified and I have done so.

pnpm-workspace.yaml
```diff
@@ -67,13 +72,16 @@ trustPolicyExclude:
   # current 7.x line publishes through the npm-cli OIDC/Actions pipeline. The 5.x/6.x
   # lines will not be retroactively re-published with provenance. Pulled in by widely-
   # used legacy tooling that pins ^5 / ^6.
-  - "semver@5.7.2"
-  - "semver@6.3.1"
+  # Note: multiple versions of the same package must be combined with "||" — pnpm's
+  # trust-policy evaluator returns on the first name match and does not aggregate
+  # subsequent entries for the same package.
+  - "semver@5.7.2||6.3.1"
```
alexvy86, line 78: Huh, but wouldn't that make is to semver 6.3.1 complained before? (Which we were not seeing AFAIK). If we're confident about the new approach, go for it.

alexvy86 (approved):

pnpm-workspace.yaml
CraigMacomber: pnpm has a known issue (fixed in 11.1.3) where it doesn't check already installed packages, and only gives trust errors when it visits them in some other operation, which let this slip through.

PR #27337, opened 2026-05-18
fix(build-cli): use node:zlib.gunzipSync for tarball extraction

ChumpChief:
## Description

The `fflate@0.8.3` release (2026-05-16) broke `flub publish tarballs` across all internal ADO publish pipelines — `Build - client packages`, `Build - build-tools`, and `server-routerlicious` have all been red since shortly after that publish. The publish jobs reinstall build-tools from artifact tarballs on every run without a lockfile, so they picked up the new minor patch transitively via build-cli's `^0.8.2` range. Example failure: https://dev.azure.com/fluidframework/internal/_build/results?buildId=400663&view=results

In `extractPackageJsonFromTarball`, the prior code used fflate's streaming `Gunzip` as a workaround for [fflate#207](https://github.com/101arrowz/fflate/issues/207) (where `gunzipSync` silently returns empty for some npm tarballs). In 0.8.3 the streaming callback now fires multiple times for a single `push(buf, /*final*/ true)` — first a real chunk, then zero-length terminator chunks — so the existing `unzipped = chunk` assignment ends up holding an empty buffer. `untar(...)` then returns `[]`, and `data[0].filename` throws `TypeError: Cannot read properties of undefined (reading 'filename')`.

Switching to `node:zlib.gunzipSync` removes the dependency on fflate's fragile gunzip implementations for this Node-only code path. `fflate` stays in deps for `unzipSync` in `downloadArtifact.ts`, which is unaffected. The bump to `^0.8.3` is included in a follow-up commit now that we no longer call its broken gunzip API.

### Reproducer

A minimal repro of the failure (same pattern as the prior `extractPackageJsonFromTarball`):

```js
import { readFile } from 'node:fs/promises';
import { Gunzip } from 'fflate';
import { untar } from '@andrewbranch/untar.js';
const tarball = new Uint8Array(await readFile(process.argv[2]));
let unzipped;
const gunzip = new Gunzip((chunk) => { unzipped = chunk; });
gunzip.push(tarball, true);
const data = untar(unzipped.buffer);
console.log(data.length, data[0]?.filename);
```

Run against any `.tgz`:
- fflate 0.8.2 → `1 package/package.json`
- fflate 0.8.3 → `0 undefined` (then production code throws the observed TypeError)

After this change, the same code path using `gunzipSync` from `node:zlib` works on both small (pool-shared `Buffer`) and large registry tarballs.

### Temporary publish-job override

The source fix alone is not enough to recover the publish pipelines. `Build - build-tools` consumes `buildTools-resource` (its own previous successful main run) for the publish stage, so the broken pre-fix `flub` is what actually runs `flub publish tarballs` even when this branch's fixed source is being built. A manual `Build - build-tools` queue against `test/build-failures` with only the source fix confirmed this — [build 400994](https://dev.azure.com/fluidframework/internal/_build/results?buildId=400994&view=results) failed with the same `TypeError` as the original.

To break that bootstrap cycle, this PR also adds a one-line override to `tools/pipelines/templates/include-publish-npm-package-deployment.yml` pinning `fflate` to `0.8.2` in the synthesized publish-job `package.json`. This uses the escape hatch already documented in the template (lines 61–69) for transitive-dependency-floats-and-breaks scenarios. It keeps the currently-published broken `flub` working long enough for one successful post-merge `Build - build-tools@main` run to refresh `buildTools-resource` with the fixed `flub`.

Validated by re-queueing `Build - build-tools` against `test/build-failures` with the override commit applied: [build 400999](https://dev.azure.com/fluidframework/internal/_build/results?buildId=400999&view=results) went green through `Publish internal test feed`.

### Follow-up

- Removing the fflate override once `buildTools-resource` has been refreshed by a successful post-merge `Build - build-tools@main` run (the fixed `flub` no longer depends on fflate's streaming `Gunzip`) is tracked as [AB#73634](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73634).
- The publish jobs do a lockfile-less fresh install of build-tools, which is what allowed an unrelated transitive minor release to break us. Tracked separately as [AB#73628](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73628).

[AB#73627](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73627)

## Reviewer Guidance


alexvy86 (approved):
The explanation in the PR description makes sense to me and the change does too as solution. Assuming some validation was done with a test branch on the pipelines, looks good to me 👍 .

Abe27342 (approved):
Looks like we might be able to remove the fflate dep altogether and use node's native lib as well? Only other usage is `unzipSync` in `downloadArtifact.ts` 👀 

(not asking as part of this PR :))

ChumpChief:
> Looks like we might be able to remove the fflate dep altogether and use node's native lib as well? Only other usage is `unzipSync` in `downloadArtifact.ts` 👀
> 
> (not asking as part of this PR :))

lol I just added that usage :-D afaik zlib doesn't support normal zips, just gzips unfortunately

alexvy86 (approved):

PR #27341, opened 2026-05-18
build: regenerate type tests for 2.101.0 and disable for fluid-framework

jatgarg:
## Description

 Post-2.101.0 type test baseline update on `main`. Bumps all client release group `*-previous` deps to
  the 2.101.0 publish and regenerates the `validate*Previous.generated.ts` files.

  Two packages can't use the standard pattern and get targeted fixes:

  ### `fluid-framework` — disabled
  Flub's default entrypoint is `legacy/alpha`, which is not exported by `fluid-framework@2.101.0` (or
  any earlier published version). Disabled via `typeValidation.disabled: true`, matching the prior fix
  from PR #26930.

  ### `@fluid-internal/presence-definitions` — disabled
  The renamed `@fluid-internal/presence-definitions@2.101.0` only exports `.` and `./internal*`, so
  `entrypoint: "beta"` can't resolve.


jzaffiro (approved):

packages/framework/tree-agent-langchain/package.json
```diff
@@ -83,8 +83,7 @@
 		"test:mocha:cjs": "cross-env FLUID_TEST_MODULE_SYSTEM=CJS mocha",
 		"test:mocha:esm": "mocha",
 		"test:mocha:verbose": "cross-env FLUID_TEST_VERBOSE=1 npm run test:mocha",
-		"tsc": "fluid-tsc commonjs --project ./tsconfig.cjs.json && copyfiles -f ../../../common/build/build-common/src/cjs/package.json ./dist",
-		"typetests:gen": "flub generate typetests --dir . -v"
```
CraigMacomber: There are a mix of changes unrelated to the upgrade in this PR that are just due to normalization (like this), and the apparent manual workarounds mentioned in the PR description (of which we should have none: we should keep our codebase in a state where when we run our automated normalization it is left in a working state so we don't have to manually fix things every time we normalizer the type tests.

I think it might be best to do a separate PR first which just fixes our settings and normalizes them such that running a normalize produces a working state with no additional changes.

On a related note, I think the diff to this package, and the change to fluid-framework are likely mistakes: tests with user facing APIs really should have type tests: this package simply left them disabled (likely in its initial version as there wasn't a version to test against): now that this package has a previous version it can be tested against, it should probably test against that rather than be disabled.

packages/framework/fluid-framework/package.json
```diff
@@ -119,12 +119,11 @@
...
 		"copyfiles": "^2.4.1",
 		"eslint": "catalog:eslint",
-		"fluid-framework-previous": "npm:fluid-framework@2.92.0",
 		"jiti": "^2.6.1",
 		"rimraf": "^6.1.3",
 		"typescript": "~5.4.5"
 	},
 	"typeValidation": {
-		"broken": {}
+		"disabled": true
```
CraigMacomber: I don't know what we want to do with type tests in for this package. Clearly they should be in a state where they are normalized (either properly disabled like you have made them) or properly enabled (and thus not missing the generation script and pointed at an existing entry point). I however am not clear on which entry points are supposed to exist in this package as the ones in the package.json do not match the api-extractor configs. Specifically we have an API extractor config for beta-legacy, which we don't export: I think that API extractor config probably shouldn't exist as  users of that API surface shouldn't use this package.

In that case we would want to type test the legacy entry point I think.

packages/framework/tree-agent-langchain/package.json
CraigMacomber: Maybe its fine to just do it in this PR: I don't actually see other places which are being changed. It is a pain to look for cases like this in review of a big PR like this though

PR #27345, opened 2026-05-19
build: normalize typetest script presence across client release group

jatgarg:
## Description

Output of `pnpm exec flub typetests -g client --normalize`. Reconciles each package's `typetests:gen` script presence with its `typeValidation.disabled` state:

- Removes the stale `typetests:gen` script from 10 packages whose `typeValidation.disabled` is already `true` (sequence-deprecated, ink, legacy-dds, attributor, fluid-telemetry, dds-interceptions, tree-agent and its variants, type-factory).

 - fluid-framework: no version of the package exposes flub's default "legacy/alpha" entrypoint; it's an aggregator and constituent packages already provide
  type-test coverage.
 - @fluid-internal/presence-definitions: test against "./public" entrypoint.


packages/framework/presence-definitions/package.json
```diff
@@ -104,14 +103,6 @@
 		}
 	},
 	"typeValidation": {
-		"entrypoint": "beta",
```
CraigMacomber, line 107: @jason-ha : This seems odd. I see the `"typetests:gen-disabled": "flub generate typetests --dir . -v"` script above.

Is this intended to actually be enabled?
Now that there is a release of it, we could enable this, and test against that?

CraigMacomber: @jatgarg  Loking at the entry points that do exist, I think this should target public

packages/framework/fluid-framework/package.json
```diff
@@ -119,12 +119,11 @@
...
 		"copyfiles": "^2.4.1",
 		"eslint": "catalog:eslint",
-		"fluid-framework-previous": "npm:fluid-framework@2.92.0",
 		"jiti": "^2.6.1",
 		"rimraf": "^6.1.3",
 		"typescript": "~5.4.5"
 	},
 	"typeValidation": {
-		"broken": {}
+		"disabled": true
```
CraigMacomber: I think this one should target legacy, which I think should actually work (we don't have legacy beta, so we can't target that, so its either beta or legacy, and I think legacy makes more sense). Both of those include public, which for sure should be included.

packages/framework/presence-definitions/package.json
jatgarg: Added it back.

packages/framework/fluid-framework/package.json
```diff
@@ -68,6 +68,9 @@
 		"build:compile": "fluid-build . --task compile",
 		"build:docs": "api-extractor run --local",
 		"build:esnext": "tsc --project ./tsconfig.json",
+		"build:test": "npm run build:test:esm && npm run build:test:cjs",
+		"build:test:cjs": "fluid-tsc commonjs --project ./src/test/tsconfig.cjs.json",
+		"build:test:esm": "tsc --project ./src/test/tsconfig.json",
```
CraigMacomber, line 73: I see why we didn't get this setup before: we were missing the whole test build setup for this package so they wouldn't do anything. I guess that was a bigger fix than I expected, but still nice to have (now we can write other tests if we want to easily: making writing of tests easy is a good thing!). Looks good.

CraigMacomber (approved):

PR #27346, opened 2026-05-19
ci(pipelines): route server base images through container registry mirrors

ChumpChief:
Adds an overridable `BASE_IMAGE_REGISTRY` build arg to the four `server-*` Dockerfiles (gitssh, gitrest, historian, routerlicious), defaulting to `docker.io` so local and external-contributor builds are unchanged. Each `server-*` pipeline overrides the arg via `additionalBuildArguments` to pull from a mirror reachable from its build pool, avoiding the 1ES network-isolation block on `registry-1.docker.io`.

The override is selected per project:

- **internal**: all four pipelines target our internal ACR mirror via `$(containerRegistryUrl)/mirror/docker` (from the `container-registry-info` variable group). This is the long-term home for our pinned base images.
- **public**:
  - `gitssh` targets `mcr.microsoft.com/mirror/docker`, which carries the pinned `alpine:3.16` digest.
  - `gitrest`, `historian`, `routerlicious` use Node 22, which MCR's mirror does not carry, so they fall back to the Dockerfile default of `docker.io`. They have not been observed hitting the netiso block on public to date; mirroring Node 22 into a public-accessible registry is tracked as part of the long-term fix below.

Also adds `tools/pipelines/README.md` documenting how to upgrade or add mirrored base images going forward.

Verified locally: `pnpm run build:docker` succeeds for gitrest/historian/routerlicious and `docker build` succeeds for gitssh, with all four resolving base images from `docker.io` via the Dockerfile default — confirming the fallback path for normal contributors.

This PR is the interim fix tracked by [AB#73954](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73954). The broader long-term migration (including a public-accessible Node 22 mirror) remains tracked by [AB#73353](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73353).

[AB#73954](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73954)

### DIFF
--- a/tools/pipelines/templates/include-test-real-service.yml
+++ b/tools/pipelines/templates/include-test-real-service.yml
@@ -263,7 +263,7 @@ stages:
           - task: Docker@2
             displayName: 'Login to container registry'
             inputs:
-              containerRegistry: 'Fluid Container Registry'
+              containerRegistry: $(containerRegistryConnection)
               command: 'login'
           - task: DockerCompose@1
             displayName: 'Start routerlicious environment in docker'
@@ -274,7 +274,7 @@ stages:
               dockerComposeCommand: 'up -d --wait'
               projectName: 'routerlicious'
             env:
-              REGISTRY_URL: fluidcr.azurecr.io/build
+              REGISTRY_URL: $(containerRegistryUrl)/build
               ALFRED_IMAGE_TAG: ${{ parameters.alfredImageTag }}
               HISTORIAN_IMAGE_TAG: ${{ parameters.historianImageTag }}
               GITREST_IMAGE_TAG: ${{ parameters.gitrestImageTag }}
--- a/tools/pipelines/test-real-service.yml
+++ b/tools/pipelines/test-real-service.yml
@@ -38,10 +38,9 @@ resources:
 variables:
 - template: /tools/pipelines/templates/include-vars-telemetry-generator.yml@self
 - group: prague-key-vault
-# Note: we tried to centralize the use of the 'container-registry-info' variable group (i.e. put it
-# inside templates we depend on) but ran into issues because if it exists in stage-level variables
-# instead of root-level pipeline variables, the value for the ACR service connection name passed
-# to Docker tasks isn't available before ADO performs checks for that task.
+# Loaded at the root level rather than as stageVariables inside include-test-real-service.yml so
+# that $(containerRegistryConnection) is available before ADO performs service-connection pre-checks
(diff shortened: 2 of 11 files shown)


tools/pipelines/test-real-service.yml
```diff
@@ -38,6 +38,11 @@ resources:
 variables:
 - template: /tools/pipelines/templates/include-vars-telemetry-generator.yml@self
 - group: prague-key-vault
+# Note: we tried to centralize the use of the 'container-registry-info' variable group (i.e. put it
+# inside templates we depend on) but ran into issues because if it exists in stage-level variables
+# instead of root-level pipeline variables, the value for the ACR service connection name passed
+# to Docker tasks isn't available before ADO performs checks for that task.
```
alexvy86: Hmm I remember writing this comment a while back somewhere, but it doesn't seem to apply here? We're putting the variable group in the root-level pipeline variables.

Abe27342 (approved):

tools/pipelines/templates/include-test-real-service.yml
```diff
@@ -263,7 +263,7 @@ stages:
           - task: Docker@2
             displayName: 'Login to container registry'
             inputs:
-              containerRegistry: 'Fluid Container Registry'
+              containerRegistry: $(containerRegistryConnection)
```
alexvy86, line 266: I think this will work with the variable group that contains the variable now being in the root-level pipeline variables, but would be great to double confirm with a test run.

tools/pipelines/test-real-service.yml
ChumpChief: Yes, it still applies — this PR moves the load from `stageVariables` (where it lived inside the `include-test-real-service.yml` template parameters, see the deletion at line 105) up to root level for exactly that reason. Tightened the wording in df10918e31 to make that historical move explicit instead of reading as a general warning.

tools/pipelines/templates/include-test-real-service.yml
ChumpChief: Confirmed working — e2e run [402070](https://dev.azure.com/fluidframework/internal/_build/results?buildId=402070) succeeded on this branch, including the Docker login + compose-up steps that exercise this path.

alexvy86 (approved):

PR #19854, opened 2024-02-28
Remove url-parse usage throughout repo

ChumpChief:
With #19840, we'll no longer use non-standard protocols in resolved URLs.  This was the remaining reason for using the url-parse library (aka Node's url support polyfill).  This library is deprecated.

This change removes all remaining usage of that package, replacing with standard `URL()`.

Aside from moving to a supported standard and away from a deprecated dependency, this also saves bundle size (looks to be about 50kb minified, comparing before/after of one of our examples).

This change revealed another defect: IParsedUrl.version claimed to permit `null` to indicate "don't load from a snapshot", but this was actually getting converted to `undefined` in `Loader.loadContainer()` (probably sensible, since we don't really support loading from no snapshot anymore or even guarantee we keep old snapshots around).  Updated the typing to reflect this reality - technically a breaking change but not really.  Fixing this is convenient, since we then don't need to worry about the return value of `URLSearchParams.get()` being `null` when a param is not found.

[AB#822](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/822)

PR #17850, opened 2023-10-18
Expose the parseUrl utility api for loop

jatgarg:
## Description

ADO item link: https://dev.azure.com/fluidframework/internal/_workitems/edit/5811/
Expose the parseUrl utility api for loop so that they can use it. Since requestResolvedObjectFromContainer() in loader.ts is deprecated and loop also implement their own routing logic instead of IFluidRouter on container, they can not use this api. In order to request a particular datastore, they need to get the deep link info from the url. That funtionality to get the deep link info from the URL is done in parseURl api present in container loader package.

PR #27211, opened 2026-04-30
feat(container-loader): add readOnly option to loadFrozenContainerFromPendingState

anthony-murphy:
## Description

Adds an optional `readOnly` parameter (default `true`, preserving existing behavior) to `loadFrozenContainerFromPendingState` and `createFrozenDocumentServiceFactory`. When `readOnly: false`, the frozen container loads as writable so the runtime accepts DDS submissions and accumulates them in `pendingStateManager` for `getPendingLocalState()` to capture — without publishing them.

The mechanism: a sibling `WritableFrozenDeltaStream` is the live connection (claims include `DocWrite`, mode stays `"read"`). `ConnectionManager.sendMessages` matches it via `isWritableFrozenDeltaStreamConnection` and short-circuits before the read-mode reconnect branch — outbound writes are dropped at the network layer instead of triggering a reconnect (which under `ReconnectMode.Never` would close the container).

For blob uploads, `FrozenDocumentStorageService.createBlob` hangs (returns a never-resolving promise) instead of throwing in the writable path. The BlobManager keeps the blob in `uploading` state, `getPendingBlobs` downgrades it to `localOnly` in pending state, and a subsequent live load runs `sharePendingBlobs` to complete the upload against real storage. Throwing instead would delete the cache entry and lose the blob.

Other notable changes:

- `FrozenDeltaStream` is split into two sibling classes (`FrozenDeltaStream` read-only, `WritableFrozenDeltaStream` writable) over a shared abstract base. Both internal-only. Type discrimination at consumer call sites uses predicate functions (`isFrozenDeltaStreamConnection`, `isWritableFrozenDeltaStreamConnection`) — no raw `instanceof`.
- `WritableFrozenDeltaStream` mints a per-instance `frozen-delta-stream/<uuid>` `clientId` to avoid `pendingStateManager` `0x173` (`replayPendingStates called twice for same clientId!`) on reconnect with dirty pending ops. The read-only variant keeps the historical constant `"storage-only client"`.
- `submitSignal` is now a silent no-op on both variants (was a 403 nack on the read-only variant). Signals are ephemeral and dropping is the right behavior with no upstream; the read-only variant's `submit` stays defensive (403 nack) since ops are durable state.
- `createFrozenDocumentServiceFactory` rewraps an already-wrapped factory if its `readOnly` differs from the caller — most recent intent wins.
[...]

PR #27347, opened 2026-05-19
feat(container-loader): support offline loads in loadFrozenContainerFromPendingState

anthony-murphy:
## Description

Adds an "offline" form of `loadFrozenContainerFromPendingState` that loads a frozen container entirely from the URL captured in `pendingLocalState` — no `IRequest`, `IUrlResolver`, or `IDocumentServiceFactory` is supplied by the caller.

`ILoadFrozenContainerFromPendingStateProps` becomes a discriminated union of two shapes, selected by the presence of the driver-wiring fields:

- **Online** (existing behavior): caller supplies `request`, `urlResolver`, and `documentServiceFactory`. The supplied factory is wrapped in a frozen factory and the resolver re-resolves the request just as today.
- **Offline** (new): caller omits all three. The entry point synthesizes an `IResolvedUrl` from `pendingLocalState.url`, an `IUrlResolver` whose `resolve()` returns that synthesized URL, and a no-inner frozen `IDocumentServiceFactory`. The existing pipeline (`loadExistingContainer` -> `Loader.resolve` -> `Container.load`) is reused unchanged — the identity guard in `Loader.resolveCore` is trivially satisfied because the same URL string is on both sides of the compare.

Mixing the two forms (some fields supplied, others omitted) is rejected at runtime with `UsageError`.

`IContainer.getAbsoluteUrl` throws on an offline-loaded container, since the resolver's external (request-form) URL cannot be reconstructed from the captured resolved URL alone. Fabricating a string in the wrong format would mislead any caller that passed it back into a real resolver.

No changes to `Container` or `Loader` were required; all new behavior is composed at the entry-point layer in `createAndLoadContainerUtils.ts`.

## Drive-by fix: `tryParseCompatibleResolvedUrl` non-throwing contract restored

The `utils.ts` hunk restores the original non-throwing contract of `tryParseCompatibleResolvedUrl`. The `try`-prefixed name implies a `=== undefined` signal on bad input — established by markfields in #17850 — and was silently regressed in #19854 when the body was rewritten to use an unguarded `new URL(url)` (which throws `TypeError` on relative paths, empty strings, or non-URL input). The offline path here is the first caller for which the gap mattered in practice; the fix is a try/catch wrap so the helper returns `undefined` for any input that fails to parse as an absolute URL. Treat it as a real bug fix, not cosmetic.

## Reviewer Guidance

- The `interface` -> `type` change on `ILoadFrozenContainerFromPendingStateProps` is required because TS does not allow widening an inherited required field (`request`/`urlResolver`/`documentServiceFactory`) to optional via `extends`. The API is `@legacy @alpha` and was introduced one PR ago, so consumer impact is essentially zero.
- The synthesized `request.url` carries a *resolved-form* URL where an *external-form* URL is normally expected. Nothing downstream inspects it before the synthesized resolver swallows it. Worth a sanity check that no future load-pipeline change starts depending on `request.url` having external-form structure.
- Three new tests in `loadFrozenContainerFromPendingState.spec.ts` cover: offline happy path, `getAbsoluteUrl` throws, mixed-driver-wiring is rejected.


packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -284,11 +329,73 @@ export interface ILoadFrozenContainerFromPendingStateProps
...
+				);
+			},
+		};
+		return loadExistingContainer({
+			...props,
+			// `request.url` carries the resolved-form URL here. Nothing downstream
+			// inspects it between this entry point and the synthesized resolver above.
+			request: { url: pending.url, headers: {} },
+			urlResolver: synthesizedUrlResolver,
+			documentServiceFactory: createFrozenDocumentServiceFactory(undefined, readOnly),
```
anthony-murphy, line 469: Deep Review: The offline branch passes `createFrozenDocumentServiceFactory(undefined, readOnly)` — no inner factory. In `frozenServices.ts:218-220`, `readBlob` then falls back to `frozenDocumentStorageServiceHandler`, which throws a raw `Error("Operations are not supported on the FrozenDocumentStorageService.")` — not a `UsageError`, bypassing the package's misuse-diagnosis machinery.

#25590 (anthony-murphy, 2025-10-01) explicitly flagged this gap: *"Pending local state does not currently capture blobs, so we need to support reading blobs from the network. This will prevent usages in full offline situations…"*. #27220 (markfields, 2026-05-08) then added `captureFullContainerState` as the inlined-blob counterpart that makes a truly-offline load safe, and described it as enabling fully-offline frozen-container scenarios *together with* `loadFrozenContainerFromPendingState`. This PR is the consuming end of #27220 but neither references it from the new JSDoc nor exercises a `captureFullContainerState` → offline-load round-trip — the new spec uses `getRequiredPendingLocalState(container)` (the standard, blob-omitting capture) and only sets primitive DDS values, so the gap is not surfaced.

The JSDoc on the offline form says "no real driver is contacted" without flagging the inlined-blob precondition.

Suggested fix — one or more of:

1. Document on `ILoadFrozenContainerFromPendingStateProps`'s offline form that `pendingLocalState` must be produced by `captureFullContainerState` (the only path that inlines blobs), and cross-link both APIs in the JSDoc.
2. At the offline entry point, detect pending-state payloads that lack the inlined-blob marker and reject with a `UsageError` instead of letting the failure surface mid-load as a generic `Error`.
3. Add a `captureFullContainerState` → offline-load round-trip test that exercises an attachment blob, locking the contract in.

Question: Is the intent that the offline branch accepts only `captureFullContainerState` output, or also `getRequiredPendingLocalState` output (in which case, what's the contract when the runtime calls `readBlob` during load)?

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -284,11 +329,73 @@ export interface ILoadFrozenContainerFromPendingStateProps
...
+		// Offline: synthesize the driver wiring from the URL captured in pending state.
+		// The container's load pipeline is reused unchanged — the synthesized resolver
+		// returns a resolved URL whose `url` equals `pendingLocalState.url`, so the
+		// identity guard in `Loader.resolveCore` is trivially satisfied.
+		const pending = getAttachedContainerStateFromSerializedContainer(pendingLocalState);
+		const parsed = tryParseCompatibleResolvedUrl(pending.url);
+		if (parsed === undefined) {
+			throw new UsageError(
+				`loadFrozenContainerFromPendingState: pending state URL is not in a parseable form (${pending.url})`,
+			);
```
anthony-murphy, line 425: Deep Review: This branch calls `const parsed = tryParseCompatibleResolvedUrl(pending.url)` and gates only on `parsed === undefined`, throwing `UsageError("…pending state URL is not in a parseable form (…)")`. But `utils.ts:83-100` shows `tryParseCompatibleResolvedUrl` begins with an unguarded `const parsed = new URL(url);` — for any non-absolute or syntactically invalid URL string, `new URL(...)` throws a built-in `TypeError: Invalid URL` *before* the function can ever return `undefined`. The intended `UsageError` jacket is unreachable for the broadest class of bad URLs (anything that isn't a well-formed absolute URL).

The helper's `Try…` name implies non-throwing on bad input; today it only handles the "parsed but wrong shape" case, not the "didn't parse" case.

Suggested fix — preferably harden `tryParseCompatibleResolvedUrl` itself so its name is honest for all callers:

```ts
export function tryParseCompatibleResolvedUrl(url: string): IResolvedUrl | undefined {
	let parsed: URL;
	try {
		parsed = new URL(url);
	} catch {
		return undefined;
	}
	// …existing shape checks…
}
```

Alternative: wrap just this call site in try/catch and rethrow as `UsageError`. The choice affects other consumers of the helper — which do you prefer?

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Addressed in 9213320:

- **Type**: rozenDocumentStorageServiceHandler now throws `UsageError` (not raw `Error`), and the no-inner-factory `readBlob` path gets a dedicated `frozenReadBlobOfflineHandler` whose message names `captureFullContainerState` explicitly.
- **Doc**: The offline form's JSDoc now spells out the inlined-blob precondition and cross-links to `captureFullContainerState`; `readOnly: false` semantics are documented for offline too.
- **Tests**: Positive — `captureFullContainerState` ΓåÆ offline-load round-trip in `loadFrozenContainerFromPendingState.spec.ts` (uploads a blob, captures, loads offline, dereferences handle). Negative — unit test in `frozenServices.spec.ts` verifies `readBlob` on no-inner storage throws `UsageError` with `/captureFullContainerState/` in the message.

Re: your question — offline accepts either capture path, but the new doc states the contract: any artifact whose dereferenced graph contains attachment blobs must come from `captureFullContainerState`. `getRequiredPendingLocalState` is fine for the no-blob case; the runtime never eagerly reads attachment blobs at load time, so loads always succeed and only `handle.get()` on a non-inlined blob fails with the named `UsageError`.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Addressed in 9213320 — hardened `tryParseCompatibleResolvedUrl` itself (option 1 from your suggestion), with a try/catch around the `new URL(url)` call so the function's `try`-prefixed name is honest for all callers. The `UsageError` jacket at line 358 is now reachable for the broadest class of bad URLs.

Surveyed the other two call sites:
- `loader.ts:331` — applied to `resolvedAsFluid.url` from a real resolver. The pre-existing `throw new Error("Invalid URL ")` now becomes the surfaced error (better than the bare `TypeError` that `new URL` threw before — the message now includes the offending URL).
- `loader.ts:337` — applied to `pendingLocalState.url`. A bad URL now yields `parsedPendingUrl === undefined`, which falls through to the existing identity-mismatch error path. Same observable behavior for the caller, just routed through the intended check.

No caller relied on the throw; all gate on `=== undefined`.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Confirmed resolved in `9213320` — `frozenReadBlobOfflineHandler` throws `UsageError` naming `captureFullContainerState`, the offline JSDoc spells out the inlined-blob precondition with the cross-link, the positive round-trip test (uploads blob → `captureFullContainerState` → offline-load → `handle.get()`) and the negative unit test in `frozenServices.spec.ts` are both in. The clarification on `getRequiredPendingLocalState` (no-blob case still supported; only `handle.get()` on a non-inlined blob fails with the named `UsageError`) answers the contract question. Thanks.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Confirmed resolved in `9213320` — option 1 (harden the helper itself) applied, the `try`-prefixed name is now honest for all callers. Call-site survey of `loader.ts:331` and `loader.ts:337` checks out: both gate on `=== undefined`, and `loader.ts:331`'s pre-existing `throw new Error("Invalid URL ")` now surfaces with the offending URL in the message instead of a bare `TypeError`. Thanks.

packages/loader/container-loader/api-report/container-loader.legacy.alpha.api.md
```diff
@@ -4,6 +4,11 @@
 
 ```ts
 
+// @alpha @legacy
+export type AllOrNothing<T> = T | {
+    [K in keyof T]?: never;
+};
+
 // @alpha @legacy
```
anthony-murphy, line 7: Deep Review: This PR newly exports `AllOrNothing<T>` on the `@legacy @alpha` surface (re-exported from `src/index.ts:266`) to express a one-off all-or-none constraint on the driver-wiring triplet of a single internal call site. `ILoadFrozenContainerFromPendingStateProps` is being reshaped for the third time in ~8 months by the same author (#25538 introduced it, #27211 modified it, this PR reshapes `interface` → `type` intersection). ChumpChief pushed back in #25513 on newly-exposed `asLegacyAlpha` helpers and on whether generic names scale on this surface — same area, same pattern.

The constraint can be expressed inline without a public export and without changing the runtime `UsageError` backstop at `createAndLoadContainerUtils.ts:348-360` or the mixed-wiring test at `loadFrozenContainerFromPendingState.spec.ts:1274-1295`:

```ts
export type ILoadFrozenContainerFromPendingStateProps = ILoadExistingContainerProps &
	(
		| { urlResolver: IUrlResolver; documentServiceFactory: IDocumentServiceFactory; request: IRequest }
		| { urlResolver?: never; documentServiceFactory?: never; request?: never }
	);
```

Static type-safety is preserved, no new generic helper enters the forever-public `@legacy @alpha` surface.

If `AllOrNothing<T>` stays exported as a deliberate decision (intended reuse on other props types, etc.), document the rationale in its JSDoc so future maintainers don't repeat the question — and worth getting markfields to explicitly sign off on the API shape since the consumer story intersects directly with `captureFullContainerState` (#27220).

Question: Was the inline anonymous discriminated union considered? If so, what tipped the decision toward the named export?

anthony-murphy: Moved AllOrNothing<T> to `@fluidframework/core-interfaces` and renamed it to `AllOrNone<T>` (better fit — the "all-or-none" framing reads cleanly and lines up with the JSDoc, and the original "AllOrNothing" had the wrong cadence for a type modifier).

Kept it as an exported `@legacy @alpha` symbol per the rationale you raised — see [`allOrNone.ts`](https://github.com/microsoft/FluidFramework/blob/66f1d3dac2/packages/common/core-interfaces/src/allOrNone.ts) and the `@remarks` block: this is a small, general, composable primitive that shows up whenever a feature is wired up by supplying a group of cooperating objects (driver triples, transport plus credentials, etc.), and naming it lets reviewers see intent at the declaration site instead of re-deriving it from a hand-rolled discriminated union.

In `ILoadFrozenContainerFromPendingStateProps` we now import `AllOrNone` from `@fluidframework/core-interfaces/internal` (eslint blocks `/legacy/alpha` per the existing repo convention) and continue to use it in the existing intersection. No public-API surface added to `container-loader`; the new symbol lives one layer down in `core-interfaces` where shared primitives belong.

Commit: 66f1d3dac2.

anthony-murphy: Deep Review: Confirmed resolved in `66f1d3d` — `AllOrNothing<T>` moved to `@fluidframework/core-interfaces` and renamed `AllOrNone<T>`, with `@remarks` documenting the cooperating-object-group rationale at the declaration site. `container-loader` imports from `/internal` per the existing eslint convention, so no new public surface lands on this package. The rename reads cleanly and the question is preserved on the human-reviewer agenda for markfields to weight the reuse case explicitly. Thanks.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -238,10 +244,40 @@ export async function loadExistingContainer(
...
+ * resolver's external URL format is unknown without a real `IUrlResolver`.
+ *
+ * **Offline form precondition.** With no driver wiring there is no live
+ * storage to read attachment blobs from, so any blob the runtime dereferences
+ * during load must already be inlined into `pendingLocalState`. The pending
+ * state must therefore be produced by {@link captureFullContainerState}
+ * (which inlines all referenced attachment blobs) rather than
+ * `IContainer.getPendingLocalState` / `getRequiredPendingLocalState` (which
+ * do not). If the runtime needs an attachment blob that is not inlined, the
+ * load fails with `UsageError` from the synthesized storage service.
```
anthony-murphy, line 330: Deep Review: The offline-form JSDoc on `ILoadFrozenContainerFromPendingStateProps` documents the no-driver-wiring contract, the inlined-blob precondition, and `readOnly` semantics — but doesn't surface the shape constraint that `pendingLocalState.url` must satisfy `tryParseCompatibleResolvedUrl`. The parser's contract (established by markfields in #17850, see `utils.ts:74-83`) is that the URL must be of shape `protocol://<string>/.../..?<querystring>` and "may not be compatible with any URL resolver." Drivers that emit non-standard resolved-URL shapes will surface as `UsageError` at load time (the throw at `createAndLoadContainerUtils.ts:348-353`), but a caller reading the offline JSDoc has no way to anticipate that.

Suggestion: add a one- or two-sentence note to the offline-form JSDoc stating that `pendingLocalState.url` must be a `tryParseCompatibleResolvedUrl`-compatible resolved URL (shape `protocol://…/…?querystring`), and that drivers emitting non-standard resolved-URL shapes will surface as a `UsageError` at load time. Cross-link `tryParseCompatibleResolvedUrl` if practical.

packages/loader/container-loader/api-report/container-loader.legacy.alpha.api.md
```diff
@@ -115,10 +115,14 @@ export interface ILoadExistingContainerProps extends ICreateAndLoadContainerProp
 }
 
 // @alpha @legacy
-export interface ILoadFrozenContainerFromPendingStateProps extends ILoadExistingContainerProps {
+export type ILoadFrozenContainerFromPendingStateProps = Omit<ILoadExistingContainerProps, "request" | "urlResolver" | "documentServiceFactory" | "pendingLocalState"> & {
```
anthony-murphy: i don't really like the large omit. can we restructure the interfaces to be more composable, like split the driver types out in the reusable type, similar with the other options, then recompose the exiting and types we need from those

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -284,11 +333,73 @@ export interface ILoadFrozenContainerFromPendingStateProps
...
+		const pending = getAttachedContainerStateFromSerializedContainer(pendingLocalState);
+		const parsed = tryParseCompatibleResolvedUrl(pending.url);
+		if (parsed === undefined) {
+			throw new UsageError(
+				`loadFrozenContainerFromPendingState: pending state URL is not in a parseable form (${pending.url})`,
+			);
+		}
+		const synthesizedResolvedUrl: IResolvedUrl = {
+			type: "fluid",
+			id: parsed.id,
```
anthony-murphy: Deep Review: The synthesized `IResolvedUrl` here assigns `id: parsed.id`, where `parsed = tryParseCompatibleResolvedUrl(pending.url)`. In `utils.ts:100-105`, `parsed.id` is `match[1]` from `/^\/([^/]*\/[^/]*)(\/?.*)$/` — the first **two** pathname segments joined by `/`, i.e. `tenantId/docId`.

But `IResolvedUrl.id` is documented in `packages/common/driver-definitions/src/urlResolver.ts:13-14` as *"the id of the container this resolved url is for,"* and every production resolver populates it with a single id segment:

- `localResolver.ts:58` — `id: documentId`
- `insecureUrlResolver.ts:132` — `id: documentId`
- `routerlicious-urlResolver/src/urlResolver.ts:134` — `id: documentId`
- `odspDriverUrlResolver.ts:167` — `id: hashedDocumentId`

The offline container exposes this synthesized object via `IContainer.resolvedUrl`, so any caller / telemetry / cache keyed on `resolvedUrl.id` sees `tenant/doc` on offline-loaded containers and a single id on online-loaded ones — a silent shape divergence with no test pinning the value.

Suggested fix: assign the document-id segment only (e.g. derive from the second path segment of `pending.url`, or extract from `parsed` if a helper is added), or thread the container id through pending state at capture time. Either way, add an offline-load assertion in `loadFrozenContainerFromPendingState.spec.ts` that pins `container.resolvedUrl.id` to the expected single-segment value so this contract can't regress.

Question: Is the `tenantId/docId` value now appearing as `IContainer.resolvedUrl.id` on offline-loaded containers intentional? If so, which downstream consumers (telemetry, cache keys, `redirectUrl` handling) were audited, and where is that decision documented?

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -139,6 +140,57 @@ export interface ICreateAndLoadContainerProps {
...
+	 * The request to resolve the container.
+	 */
+	readonly request: IRequest;
+}
+
+/**
+ * Properties necessary for creating and loading a container.
+ * @legacy @beta
+ */
+export interface ICreateAndLoadContainerProps
```
anthony-murphy, line 202: i think we should deprecate this, and move other usages to the new interfaces

anthony-murphy:
## Deep Review

*Reviewed commit `ce2ede7` on 2026-05-20.*

Not ready for sign-off. The `IContainer.resolvedUrl.id` shape divergence flagged on `faa6a1f` is still unaddressed at `ce2ede7` and the inline thread has no author reply — the offline path assigns the two-segment `tenantId/docId` string where every production resolver writes a single container-id segment. Tier 3 polish (resolved-form `request.url`, `@legacy @beta` reshape note, synthesized-resolver identity guard) is held until that lands.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add a one-line release note that `tryParseCompatibleResolvedUrl` no longer throws `TypeError` on malformed input (restored `try*` contract — behavior change for external `@legacy @beta` callers).

### Context for Reviewers

- Load-side bookend of the multi-PR frozen-container series (#25538, #25590, #25603, #25653, #27211, #27220). #27220 (markfields) introduced `captureFullContainerState` as the inlined-blob counterpart; this PR is its consuming end and the offline blob round-trip test (`loadFrozenContainerFromPendingState.spec.ts` lines 1298-1361) is the integration point that proves the series hangs together.
- `AllOrNone<T>` lives in `@fluidframework/core-interfaces` as a small composable `@legacy @alpha` primitive; `container-loader` imports it from `/internal` per the existing eslint convention. Author's own unresolved comment ([3276855511](https://github.com/microsoft/FluidFramework/pull/27347#discussion_r3276855511)) — *"i don't really like the large omit. can we restructure the interfaces to be more composable"* — suggests this shape may move again; deferring to markfields.
- `readOnly` naming was contested and settled in #27211 (markfields proposed `allowLocalChanges`/`disallowLocalChanges`; author reverted to `readOnly?: boolean` to match `IContainer.readOnlyInfo.readonly`). This PR inherits the name unchanged via `createFrozenDocumentServiceFactory(undefined, readOnly)` — don't relitigate.
- `tryParseCompatibleResolvedUrl` now honors its `try*` name. The non-throwing contract was suggested by markfields in #17850 (2023-10-19) but regressed in #19854 (2024-03-01) when the body was rewritten to use `new URL(url)` unguarded. The new offline path is the first caller for which the gap mattered in practice.
- Offline `getAbsoluteUrl` deliberately throws `UsageError`: pending state stores the resolved-form URL, not the external/request form `IUrlResolver.getAbsoluteUrl` is contractually required to produce. JSDoc at `createAndLoadContainerUtils.ts:268-276` documents it; spec at lines 1251-1271 asserts it.
- Suggested reviewers: **markfields** (author of #27220, owns the `captureFullContainerState` consumer story; persistent reviewer in #27211's `readOnly` debate; originator of `try*` naming in #17850; right voice on the `AllOrNone` shape question above), **dannimad** (authored `createFrozenDocumentServiceFactory` in #25653 and the test scaffold in #25603; natural fit for `frozenServices.spec.ts` and could verify the `resolvedUrl.id` fix below), **ChumpChief** (last touched `tryParseCompatibleResolvedUrl` in #19854; right sign-off for the utils.ts contract shift and release note when taken on), **jatgarg** (#23301 added the `@legacy` tags with the explicit rationale that Loop consumes these loader APIs in production; relevant for the held `@legacy @beta` `ICreateAndLoadContainerProps` reshape).

- **Needs human judgment (markfields)** — Whether `AllOrNone<T>` belongs as a publicly-exported `@legacy @alpha` primitive in `core-interfaces` or should be inlined at the single current call site. Author's own unresolved comment 3276855511 ("can we restructure the interfaces to be more composable") suggests the shape may move again; reconciling that with the published export is a taste/owner call.
- **Needs human judgment (markfields)** — Offline + `readOnly: false` semantics: local DDS submissions accrue without being published. The "capture → restart offline → keep editing → re-capture and replay online later" workflow is the supporting case. Owner sign-off on this contract is appropriate.
- **Needs human judgment (markfields / ChumpChief)** — Offline `getAbsoluteUrl` throws `UsageError` rather than fabricating a string. Well-reasoned, but a contract call worth confirming.
- **Needs human judgment (jatgarg)** — Reshape of `@legacy @beta` `ICreateAndLoadContainerProps` into `IContainerHostProps`/`IContainerDriverServices` bases: low-probability declaration-merging risk for Loop. Loop-prod usage check is human-scope.
- **Cannot be assessed by the pipeline** — Whether the end-to-end offline blob round-trip test catches realistic consumer workloads beyond a single primitive-DDS + single attachment-blob scenario; runtime determinism across resolver implementations; whether `Loader.resolveCore` / `Container.load` evolution will preserve the "`request.url` is opaque-until-resolver in offline mode" invariant.

</details>

- `faa6a1f` 2026-05-20 · **5/10** — `resolvedUrl.id` two-segment `tenantId/docId` shape divergence flagged inline on the offline path
- `66f1d3d` 2026-05-19 · **9/10** — `AllOrNothing` API-surface concern resolved (moved to core-interfaces, renamed `AllOrNone`); JSDoc shape-constraint Tier 3 flagged inline
- `9213320` 2026-05-20 · **8/10** — both prior correctness concerns resolved; `AllOrNothing` API-surface item flagged inline
- `2a9813f` 2026-05-19 · **5/10** — two correctness concerns flagged inline (offline blob-read footgun; unreachable `UsageError` jacket in `tryParseCompatibleResolvedUrl`)

</details>

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Addressed in faa6a1f8: the offline-form JSDoc on `ILoadFrozenContainerFromPendingStateProps` now documents that `pendingLocalState.url` must satisfy `tryParseCompatibleResolvedUrl`'s contract (shape `protocol://<string>/.../..?<querystring>`) and that drivers emitting non-standard resolved-URL shapes will surface as a `UsageError` at load time. The note is right above the existing offline-form section so callers see the constraint before consulting the parser.

packages/loader/container-loader/api-report/container-loader.legacy.alpha.api.md
anthony-murphy: Addressed in ce2ede7a: the prop types are now composed from reusable building blocks instead of the large `Omit<ILoadExistingContainerProps, ...>`.

- `IContainerHostProps` — code loader + optional policy/observability fields  
- `IContainerDriverServices` — `urlResolver` + `documentServiceFactory` pair  
- `IContainerLoadDriverProps` — driver services + `request` (the all-or-nothing group)

`ICreateAndLoadContainerProps` now extends `IContainerHostProps & IContainerDriverServices`, and `ILoadFrozenContainerFromPendingStateProps` becomes `IContainerHostProps & { pendingLocalState; readOnly? } & AllOrNone<IContainerLoadDriverProps>` — no more `Omit`. Also added `validateAllOrNone` in `@fluidframework/core-utils` so the runtime check uses a typed helper instead of hand-rolled all-defined / any-defined arithmetic.

f3ec8db4 follows up by also marking `ICreateAndLoadContainerProps` as `@deprecated` and re-extending `ILoadExistingContainerProps` / `ICreateDetachedContainerProps` / `IRehydrateDetachedContainerProps` from the building blocks directly.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Addressed in f3ec8db4: `IResolvedUrl.id` on offline-loaded containers now matches the single-doc-id-segment shape that production resolvers emit.

The synthesized `IResolvedUrl` takes the trailing path segment of `parsed.id` (which is `tenantId/docId` per the regex in `tryParseCompatibleResolvedUrl`) instead of the full two-segment value. Downstream consumers keyed on `IContainer.resolvedUrl.id` (telemetry, cache keys) now see the same shape on offline-loaded and online-loaded containers.

Also added a regression-pinning assertion in `loadFrozenContainerFromPendingState.spec.ts` (`offline (no driver wiring)` → `loads from pending state without...`) that asserts `frozenContainer.resolvedUrl?.id === "test"` — the doc-id the source container was attached with via `createCreateNewRequest("test")`. Any future regression that brings back the two-segment shape will trip this test.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Done in f3ec8db4: `ICreateAndLoadContainerProps` is now `@deprecated` with JSDoc pointing at the composable building blocks (`IContainerHostProps` for host-level config, `IContainerDriverServices` for the driver pair, or `IContainerHostProps & IContainerDriverServices` for the previous union).

The three remaining usages inside container-loader — `ILoadExistingContainerProps`, `ICreateDetachedContainerProps`, `IRehydrateDetachedContainerProps` — were migrated to extend the building blocks directly instead of `ICreateAndLoadContainerProps`. Member sets are unchanged so the interfaces remain structurally identical (type-tests pass). The deprecated interface is the only spot that still references `ICreateAndLoadContainerProps` as a name and is kept around as an alias for back-compat.

.changeset/quiet-shrews-marry.md
```diff
@@ -0,0 +1,29 @@
+---
+"@fluidframework/container-loader": minor
+"__section": deprecation
```
anthony-murphy, line 3: Deep Review: The changeset front-matter lists only `"@fluidframework/container-loader": minor`, but this PR adds new exports in two dependency packages that `container-loader` now imports across package boundaries:

- `validateAllOrNone` — new runtime export from `@fluidframework/core-utils` (`packages/common/core-utils/src/allOrNone.ts:40-57`, re-exported from `src/index.ts:37`).
- `AllOrNone` — new `@legacy @alpha` type export from `@fluidframework/core-interfaces` (`packages/common/core-interfaces/src/allOrNone.ts:90-92`, surfaced in `core-interfaces.legacy.alpha.api.md:7-10`).

`packages/loader/container-loader/src/createAndLoadContainerUtils.ts:19-20` imports both: `AllOrNone` from `@fluidframework/core-interfaces/internal` and `validateAllOrNone` from `@fluidframework/core-utils/internal`. If only `container-loader` is versioned, consumers installing the new `container-loader` minor can resolve older `core-utils` / `core-interfaces` packages that lack the new symbols — runtime `undefined is not a function` for `validateAllOrNone` and TS compile failures for `AllOrNone`. The new public `@legacy @alpha` `AllOrNone` surface also goes undocumented in release notes.

Fix: add changeset entries for `@fluidframework/core-utils` (minor — new `validateAllOrNone` runtime export) and `@fluidframework/core-interfaces` (minor — new `AllOrNone` `@legacy @alpha` type export). Either add separate `.changeset/*.md` files or extend this file's front matter to list all three packages, and add a brief note describing the new shared `AllOrNone` primitive.

While here, worth folding in the one-line note that `tryParseCompatibleResolvedUrl` no longer throws `TypeError` on malformed input — that's the same release-note channel and a behavior change for external `@legacy @beta` callers.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Confirmed resolved in `faa6a1f8` — the offline-form JSDoc on `ILoadFrozenContainerFromPendingStateProps` now documents the `tryParseCompatibleResolvedUrl` shape contract (`protocol://<string>/.../..?<querystring>`) and that drivers emitting non-standard resolved-URL shapes will surface as `UsageError` at load time. Thanks.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Confirmed resolved in `f3ec8db4` — synthesized `IResolvedUrl.id` now takes the trailing path segment of `parsed.id` and matches the single-doc-id shape every production resolver emits (`localResolver`, `insecureUrlResolver`, `routerlicious-urlResolver`, `odspDriverUrlResolver`). The new regression assertion `frozenContainer.resolvedUrl?.id === "test"` in `loadFrozenContainerFromPendingState.spec.ts` pins the contract so any future regression to the two-segment shape will trip the test. Thanks.

.changeset/quiet-shrews-marry.md
anthony-murphy: Addressed in 463f9aa1 with two new changesets:

- `sweet-spiders-grow.md` — `@fluidframework/core-interfaces` and `@fluidframework/core-utils` both minor, section: `legacy`. Documents the new shared `AllOrNone` primitive (the `@legacy @alpha` type from `core-interfaces` and the `@internal` `validateAllOrNone` runtime helper from `core-utils`) with usage examples for each, and notes that they ship together with the offline `loadFrozenContainerFromPendingState` form so the cross-package version sync is explicit.

- `harden-try-parse-resolved-url.md` — `@fluidframework/container-loader` minor, section: `fix`. Documents the behavior change: `tryParseCompatibleResolvedUrl` no longer lets `TypeError` escape on malformed input and returns `undefined` instead, plus a migration note for any caller that was relying on the throw.

Left the `quiet-shrews-marry.md` deprecation changeset focused purely on the `ICreateAndLoadContainerProps` deprecation since that's its own concern.

anthony-murphy:

- 🔦 **Mechanical Review** — testing the lights…
- 🌋 **Solution / Problem Space Review** — approaching a vent plume…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

packages/loader/container-loader/src/frozenServices.ts
```diff
@@ -165,7 +166,22 @@ class FrozenDocumentService
...
+ * whose pending state was produced by {@link getPendingLocalState} (which omits
+ * inlined attachment blobs) rather than {@link captureFullContainerState}. A
+ * generic "operation not supported" is technically true but unhelpful; this
+ * message names the missing precondition and the API that produces it.
+ */
+const frozenReadBlobOfflineHandler = async (): Promise<never> => {
+	throw new UsageError(
+		"Attempted to read an attachment blob from a frozen-loaded container without a live storage service. " +
+			"Fully-offline frozen loads must use pending state produced by `captureFullContainerState`, which inlines all referenced attachment blobs.",
+	);
```
anthony-murphy, line 184: Deep Review: This is an observable change to the `@legacy @alpha` surface that isn't called out in any changeset.

Pre-PR, `frozenDocumentStorageServiceHandler` threw `new Error("Operations are not supported on the FrozenDocumentStorageService.")`. Post-PR the generic handler throws `new UsageError(...)` (different class), and `readBlob` is rebound to a new `frozenReadBlobOfflineHandler` that throws a dedicated `UsageError` whose message names `captureFullContainerState` explicitly.

The three changesets in this PR cover URL parsing (`harden-try-parse-resolved-url.md`), the `ICreateAndLoadContainerProps` deprecation (`quiet-shrews-marry.md`), and the `AllOrNone` / `validateAllOrNone` primitive (`sweet-spiders-grow.md`) — none mention this `Error` → `UsageError` class change or the new `readBlob`-specific message.

Suggested fix: append a sentence to `quiet-shrews-marry.md` (or land a fourth changeset) noting that `FrozenDocumentStorageService` operations now throw `UsageError` instead of `Error`, and that `readBlob` on a no-inner-storage frozen container throws a dedicated `UsageError` whose message names `captureFullContainerState`.

.changeset/quiet-shrews-marry.md
anthony-murphy: Claude here — removing all four changesets on this PR in c53a55a. None of the touched surface is `@public`:

- `tryParseCompatibleResolvedUrl` → `@beta @legacy`
- `IContainer.getPendingLocalState` → `@beta @legacy` (the optional member on `IContainer`)
- `ICreateAndLoadContainerProps` / `IContainerHostProps` / `IContainerDriverServices` → `@legacy @beta`
- `AllOrNone` → `@legacy @alpha`; `validateAllOrNone` → `@internal`

Changesets aren't required for `@legacy`/`@alpha`/`@beta` surface, which covers both the cross-package `AllOrNone`/`validateAllOrNone` new-export concern raised here and the `tryParseCompatibleResolvedUrl` non-throwing fix you suggested folding in.

packages/loader/container-loader/src/frozenServices.ts
anthony-murphy: Claude here — this behavior change ships on `@legacy @alpha` surface only (`frozenDocumentStorageServiceHandler` is reached through `createFrozenDocumentServiceFactory`, which is `@alpha @legacy`), so it doesn't need changeset coverage.

Removed all four changesets on this PR in c53a55a since the same logic applies to every surface this PR touches — nothing `@public`.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -238,10 +305,46 @@ export async function loadExistingContainer(
...
+ * load fails with `UsageError` from the synthesized storage service.
+ *
+ * **URL shape requirement.** In the offline form the captured
+ * `pendingLocalState.url` is the only URL available; it is parsed in place of
+ * a real `IUrlResolver.resolve()` call, so it must satisfy
+ * {@link tryParseCompatibleResolvedUrl}'s contract — a resolved URL of shape
+ * `protocol://<string>/.../..?<querystring>`. This is the format that
+ * Fluid-shipped drivers emit; drivers that emit a non-standard resolved-URL
+ * shape will surface as a `UsageError` at load time. The online form has no
+ * such constraint because the supplied resolver controls URL parsing.
```
anthony-murphy, line 339: Deep Review: The offline form's props JSDoc here documents that `pendingLocalState.url` must satisfy `tryParseCompatibleResolvedUrl`'s shape contract (`protocol://<string>/.../..?<querystring>`), and the offline branch at lines 403-414 throws `UsageError` on parse failure. That neutralizes the doc-omission half of the original concern.

Residual gap: a partner with a non-conforming `IUrlResolver` will successfully `captureFullContainerState` but only learn at offline-load time (in a different process, possibly days later) that the resulting pending state isn't rehydratable. The PR has no capture-time shape check and no type-level discriminator on the pending-state blob — the prose precondition is the only signal. `tryParseCompatibleResolvedUrl`'s own JSDoc (`utils.ts:74-83`) still warns it "may not be compatible with any Url Resolver"; the `try*` prefix was deliberately chosen in #17850 (markfields) to signal that partial-compatibility status.

Suggested directions, in roughly increasing scope:

1. Validate the resolved-URL shape eagerly when `captureFullContainerState` writes the URL into pending state, so misuse surfaces at capture rather than at load in another process.
2. Introduce a branded `OfflineCapableUrl` discriminator on the pending-state blob and consume it at the offline entry point — gives a type-level signal that survives serialization.
3. If neither is in scope here, file a follow-up issue and link it from the `@remarks` block so the type-level hardening isn't lost.

Not a blocker — the runtime path is correct and `frozenReadBlobOfflineHandler` already pins the inlined-blob half of the precondition. This is about moving the failure earlier for partners wiring custom resolvers.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -139,11 +140,73 @@ export interface ICreateAndLoadContainerProps {
...
+/**
+ * Properties necessary for creating and loading a container.
+ *
+ * @deprecated
+ * Use the composable building blocks instead: extend
+ * {@link IContainerHostProps} for the code-loader / policy / observability
+ * surface and {@link IContainerDriverServices} for the
+ * `urlResolver` / `documentServiceFactory` pair, or compose them inline as
+ * `IContainerHostProps & IContainerDriverServices`. This interface is kept
+ * as an alias for back-compat and will be removed in a future release.
```
anthony-murphy, line 198: Deep Review: The `@deprecated` JSDoc here names the migration targets (`IContainerHostProps`, `IContainerDriverServices`, and their composition), which is the important half. What's missing is a concrete removal milestone — the current wording effectively relies on "future release," which historically calcifies on `@legacy` surfaces. #23301 added `@legacy` on top of `@alpha` precisely because `@alpha` alone wasn't gating consumer adoption; the same risk applies here once the deprecation has no fixed sunset.

Suggested fix: extend the `@deprecated` block with a target version or LTS window — e.g. "remove in 2.x+N" or "deprecated for one LTS window, target removal in <version>" — so the props-hierarchy refactor doesn't lose its motivating cleanup. The structurally-identical alias is fine for the deprecation window; the milestone is what closes the loop.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -284,11 +389,79 @@ export interface ILoadFrozenContainerFromPendingStateProps
...
+				throw new UsageError(`${fnName}: getAbsoluteUrl requires ${driverPropKeys.join("/")}`);
+			},
+		};
+		return loadExistingContainer({
+			...props,
+			// `request.url` carries the resolved-form URL here. Nothing downstream
+			// inspects it between this entry point and the synthesized resolver above.
+			request: { url: pending.url, headers: {} },
+			urlResolver: synthesizedUrlResolver,
+			documentServiceFactory: createFrozenDocumentServiceFactory(undefined, readOnly),
```
anthony-murphy, line 469: Deep Review: The offline branch synthesizes `request: { url: pending.url, headers: {} }` and an `IUrlResolver` whose `resolve()` returns an `IResolvedUrl` whose `.url` is the same `pending.url` string — a resolved-form URL fed into a request slot whose other callers carry external-form URLs. The PR description itself flags the smell: *"Worth a sanity check that no future load-pipeline change starts depending on `request.url` having external-form structure."* #5175 (ChumpChief, 2021-02-19) deliberately walked the codebase the other direction, pulling `IRequest` interpretation out of `Container`.

Today the invariant is locked down only by JSDoc prose — no assertion, no test, no type-level discriminator. Cheapest lockdown is a one-line `assert(request.url === resolvedUrl.url)` inside the synthesized resolver's `resolve()`, which converts the prose precondition into a runtime guard that fires the moment a future refactor breaks it. Alternatives: a JSDoc note on `IContainerLoadDriverProps` that the request/resolver pair may carry resolved-form internally only because both sides cooperate; or a regression test in `loader.spec.ts` asserting `Loader.resolveCore` does not inspect `request.url` past the identity guard.

Not a blocker — no present consumer reads `request.url` in the offline path. This is purely latent hardening against a future refactor.

anthony-murphy:
## Deep Review

*Reviewed commit `9558d1f` on 2026-05-26.*

Not ready for sign-off. The PR newly documents a `UsageError`-on-non-compatible-URL contract for the offline form and restores the non-throwing contract of `tryParseCompatibleResolvedUrl` — but adds no regression test pinning either. This is the same gap pattern (#19854) that silently broke the helper in the first place and is the reason this PR's own `utils.ts` drive-by fix is needed. Three prior Tier 3 inline threads also remain open.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add a one-line citation to the PR description's "Drive-by fix" subsection (or to the `utils.ts` commit message) referencing #17850 (origin of the non-throwing `try*` contract) and #19854 (the PR that silently broke it).

### Context for Reviewers

- Load-side bookend of the multi-PR frozen-container series (#25538, #25590, #25603, #25653, #27211, #27220). #27220 (markfields) introduced `captureFullContainerState` as the inlined-blob counterpart; this PR is its consuming end, and the offline blob round-trip in `loadFrozenContainerFromPendingState.spec.ts` is the integration point proving the series hangs together.
- `tryParseCompatibleResolvedUrl`'s non-throwing contract — established by markfields in #17850 (2023-10-19), silently regressed in #19854 (2024-03-01) when the body was rewritten to use unguarded `new URL(url)` — is restored by this PR's `utils.ts` hunk. The offline path is the first caller for which the gap mattered in practice; treat the change as a real bug fix, not cosmetic.
- `ILoadFrozenContainerFromPendingStateProps` turns the soft URL-shape compatibility warning from #17850 into a hard runtime precondition on the offline form: a host wiring a custom `IUrlResolver` whose resolved-URL shape differs from `protocol://<string>/.../..?<querystring>` will work online but fail offline with `UsageError`. JSDoc documents this (thread `3271021816` confirmed wording in `faa6a1f`); driver-compatibility scrutiny should focus here.
- `IResolvedUrl.id` single-segment convention is encoded as a contract for the first time. Every in-tree production resolver (`localResolver`, `insecureUrlResolver`, `routerlicious-urlResolver`, `odspDriverUrlResolver`) populates `IResolvedUrl.id` with only the trailing doc-id segment; the offline path takes that trailing segment and a regression assertion `frozenContainer.resolvedUrl?.id === "test"` pins it (resolved thread `3276943638`).
- `IContainerHostProps` / `IContainerDriverServices` / `IContainerLoadDriverProps` decomposition is broader than the offline-load feature strictly required, but is shape-preserving; `ICreateAndLoadContainerProps` is kept as a `@deprecated` alias for back-compat. The decomposition makes `AllOrNone<IContainerLoadDriverProps>` clean to apply.
- Offline implementation preserves the Container/Loader layering boundary established in #5175 (ChumpChief, 2021-02-19): offline behavior is composed entirely at the entry-point layer (synthesizes `IUrlResolver`, `IResolvedUrl`, `IRequest`, frozen `IDocumentServiceFactory`, then delegates to `loadExistingContainer`) — no changes to `Container` or `Loader`.
- `AllOrNone<T>` lands as a `@legacy @alpha` primitive in `core-interfaces` paired with `@internal` `validateAllOrNone` in `core-utils`. Today's tagging is internally consistent (only one alpha caller), but #23301's precedent (load-pipeline shapes touched by Loop prod required `@legacy @beta`) means a future prod consumer would need a retag PR.
- Precondition that pending state was produced by `captureFullContainerState` is enforced lazily at the storage layer: `frozenReadBlobOfflineHandler` throws `UsageError` naming `captureFullContainerState` only when the runtime dereferences a non-inlined blob. There is no eager check because there is no robust marker on pending state indicating its producer (resolved thread `3270280561`).
- Suggested reviewers: **markfields** (`tryParseCompatibleResolvedUrl` naming and URL-shape caveat in #17850; authored `captureFullContainerState` in #27220; right voice on the `AllOrNone` shape), **ChumpChief** (last touched `tryParseCompatibleResolvedUrl` in #19854 and established Container/Loader layering in #5175; right sign-off for the synthetic-resolver pattern), **jatgarg** (added the `@legacy` tags in #23301; relevant for the `ICreateAndLoadContainerProps` `@deprecated` tag and the `AllOrNone` tagging question), **dannimad** (authored `createFrozenDocumentServiceFactory` in #25653 and the test scaffold in #25603; natural fit for `frozenServices.spec.ts`).

- **Needs human judgment (ChumpChief)** — Sign-off on the synthetic-resolver pattern as an acceptable internal caller of the `request.url ≡ resolvedUrl.url` identity guard. #5175 (2021-02-19) walked the codebase the other direction; the author's own Reviewer Guidance flags this as worth a sanity check.
- **Needs human judgment (jatgarg)** — Reshape of `@legacy @beta` `ICreateAndLoadContainerProps` into `IContainerHostProps` / `IContainerDriverServices` bases plus the `@deprecated` tag. Structurally identical so existing Loop call sites still compile, but TS will emit deprecation diagnostics. #23301 explicitly carved this surface out for Loop; a Loop-prod usage check and migration-window confirmation are human-scope.
- **Needs human judgment (markfields)** — Whether `AllOrNone<T>` belongs as a publicly-exported `@legacy @alpha` primitive in `core-interfaces` or should be inlined at the single current call site.
- **Needs human judgment (markfields)** — Offline + `readOnly: false` semantics: local DDS submissions accrue without being published. Owner sign-off on the "capture → restart offline → keep editing → re-capture and replay online later" contract is appropriate.
- **Needs human judgment (markfields / ChumpChief)** — Offline `getAbsoluteUrl` deliberately throws `UsageError` rather than fabricating a string. Well-reasoned but a contract call worth confirming.
- **Cannot be assessed by the pipeline** — Whether the end-to-end offline blob round-trip test catches realistic consumer workloads beyond a single primitive-DDS + single attachment-blob scenario; runtime determinism across resolver implementations; whether `Loader.resolveCore` / `Container.load` evolution will preserve the "`request.url` is opaque-until-resolver in offline mode" invariant.

</details>

<details><summary>Review history (10 prior reviews)</summary>

- `86d11ed` 2026-05-26 · **8/10** — Tier 3 polish items remain (URL-shape capture-time hardening, `@deprecated` milestone, three new inline)
- `9f26493` 2026-05-26 · **8/10** — four Tier 3 polish items flagged inline on the `@legacy @alpha` surface
- `2bcc5c0` 2026-05-26 · **9/10** — all prior inline threads resolved; changeset removal in c53a55a closes release-notes polish items
- `463f9aa` 2026-05-21 · **8/10** — three Tier 3 polish items remain — flagged inline
- `5efb38e` 2026-05-21 · **5/10** — `quiet-shrews-marry.md` cross-package changeset gap for `AllOrNone` / `validateAllOrNone`
- `ce2ede7` 2026-05-20 · **5/10** — `resolvedUrl.id` two-segment shape divergence still unaddressed at this commit
- `faa6a1f` 2026-05-20 · **5/10** — `resolvedUrl.id` two-segment `tenantId/docId` shape divergence flagged inline on the offline path
- `9213320` 2026-05-20 · **8/10** — both prior correctness concerns resolved; `AllOrNothing` API-surface item flagged inline
- `66f1d3d` 2026-05-19 · **9/10** — `AllOrNothing` API-surface concern resolved (moved to core-interfaces, renamed `AllOrNone`); JSDoc shape-constraint Tier 3 flagged inline
- `2a9813f` 2026-05-19 · **5/10** — two correctness concerns flagged inline (offline blob-read footgun; unreachable `UsageError` jacket in `tryParseCompatibleResolvedUrl`)

</details>

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -284,11 +389,79 @@ export interface ILoadFrozenContainerFromPendingStateProps
...
+	}
+
+	if (driverWiring === "none") {
+		// Offline: synthesize the driver wiring from the URL captured in pending state.
+		// The container's load pipeline is reused unchanged — the synthesized resolver
+		// returns a resolved URL whose `url` equals `pendingLocalState.url`, so the
+		// identity guard in `Loader.resolveCore` is trivially satisfied.
+		const pending = getAttachedContainerStateFromSerializedContainer(pendingLocalState);
+		const parsed = tryParseCompatibleResolvedUrl(pending.url);
+		if (parsed === undefined) {
```
anthony-murphy, line 422: Deep Review: The offline form newly documents (lines 328-335) that `pendingLocalState.url` must satisfy `tryParseCompatibleResolvedUrl`'s shape (`protocol://<string>/.../..?<querystring>`) and that *"drivers that emit a non-standard resolved-URL shape will surface as a UsageError at load time."* The throw site here at 403-414 enforces it. The `utils.ts:83-95` drive-by hunk in the same PR restores the non-throwing contract of `tryParseCompatibleResolvedUrl` that markfields established in #17850 and that #19854 silently broke when `parse(url, true)` was swapped for unguarded `new URL(url)`.

Neither contract has a regression test. The new `loadFrozenContainerFromPendingState.spec.ts` covers offline happy path, `getAbsoluteUrl` throws, and mixed-driver-wiring rejection — but no case feeds a non-parseable string (`""`, `"not-a-url"`, `"/relative/path"`) as `pendingLocalState.url`. The absence of such a test is exactly the pattern that let #19854 regress the helper unnoticed: the `try*` prefix and the `=== undefined` signal were a contract, but no test pinned them, so a refactor to `new URL(url)` slipped through review.

Suggested fix: add a 5–10 line test in `loadFrozenContainerFromPendingState.spec.ts` (or a unit test next to `tryParseCompatibleResolvedUrl`) that constructs pending state with a non-compatible `url` value and asserts the offline `loadFrozenContainerFromPendingState` throws `UsageError`. One case per representative shape (`""`, `"not-a-url"`, `"/relative/path"`) pins both the new offline contract *and* the restored non-throwing `tryParseCompatibleResolvedUrl` contract — closing the gap that allowed the original silent regression.

Related to unresolved thread `3306477725` (capture-time URL-shape hardening), which addresses the *when* of detection; this thread addresses the *whether* it's locked down at all.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Claude here — addressed in e7236fc. `captureFullContainerState` now validates the resolver's URL shape at capture time via `tryParseCompatibleResolvedUrl` and throws `UsageError` naming the shape contract. A partner with a non-conforming resolver now fails fast at the `captureFullContainerState` call instead of producing a poisoned pending-state artifact that fails later at offline-load (in a different process). Regression test pins it in `captureFullContainerState.spec.ts` ("rejects at capture time when the resolver emits a non-conforming URL shape").

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Claude here — addressed in e7236fc. The synthesized `request.url` is now a recognizable opaque placeholder (`"frozen-load-from-pending-state://offline-placeholder"`) rather than `pending.url`. The synthesized `IUrlResolver.resolve()` ignores its argument and returns the `IResolvedUrl` derived from `pending.url` regardless, so functionally nothing changes — but the placeholder removes the smell of routing a resolved-form URL into a request slot whose other callers carry external-form URLs. Comment block on the new constant explains the layering.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Claude here — addressed in e7236fc. Added a regression test in `loadFrozenContainerFromPendingState.spec.ts` ("rejects offline load when pendingLocalState.url is non-conforming") that mutates a captured pending state's URL to a single-segment path, calls the offline form, and pins the `UsageError` with the documented "pending state URL is not in a parseable form" message. The companion capture-time test in `captureFullContainerState.spec.ts` closes the same gap on the producer side, so the contract is pinned at both ends of the round-trip.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Claude here — addressed in e7236fc via a new changeset (`deprecate-create-and-load-container-props.md`, `__section: deprecation`) that surfaces the deprecation in release notes and names the migration targets (`IContainerHostProps` / `IContainerDriverServices` / their composition) explicitly. Concrete removal-version milestone deferred: the changeset is the gating signal for consumer adoption, and a hand-rolled `@deprecated since` tag in JSDoc historically drifts from the actual removal schedule. The structural alias remains so existing callers compile unchanged; removal can be scheduled when the major-release plan that drops it is in flight.

anthony-murphy:

- 🔧 **Mechanical Review** — prepping the manipulator…
- 🌊 **Historical / Contextual Review** — feeling the pressure mount…
- 🔦 **Solution / Problem Space Review** — sweeping the seafloor…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Capture-time hardening confirmed in `e7236fc` — `captureFullContainerState` now throws `UsageError` when `tryParseCompatibleResolvedUrl(resolvedUrl.url)` returns `undefined`, and the new test in `captureFullContainerState.spec.ts` pins it. That closes the *when* of detection.

Residual gap (docs/changeset angle of the same concern): the gate is added unconditionally at `createAndLoadContainerUtils.ts:568-572`, but the surrounding `captureFullContainerState` docstring at `:528-530` (unchanged by this PR) still advertises the output as intended for "state relay, inspection, and durable-state snapshot use cases" — broader than offline rehydration. `tryParseCompatibleResolvedUrl`'s own JSDoc (`utils.ts:74-83`) explicitly warns it "may not be compatible with any Url Resolver"; the `try*` prefix was chosen deliberately in #17850 precisely because the parser is resolver-specific. The offline-load branch independently enforces the same URL-shape requirement at `:416-421`, so the capture-side gate is strictly broader than the code path that strictly requires the shape. `.changeset/deprecate-create-and-load-container-props.md` documents only the `ICreateAndLoadContainerProps` deprecation and doesn't surface this resolver-shape narrowing to partner consumers (Loop, per #23301).

Two minimum-cost remedies, either resolves the gap:

1. Move the gate from `captureFullContainerState` to the offline branch in `loadFrozenContainerFromPendingState` — the only path that strictly requires the parser shape; `:416-421` already enforces it on `pendingLocalState.url`. Keeps `captureFullContainerState` usable for inspection / state-relay against resolvers whose URL shape doesn't satisfy `tryParseCompatibleResolvedUrl`.
2. Keep the capture-time gate but update the `captureFullContainerState` docstring at `:528-530` to name the new supported-resolver constraint, and add a one-line entry to `.changeset/deprecate-create-and-load-container-props.md` surfacing the resolver-shape narrowing to partner consumers.

Question that picks between them: is `captureFullContainerState` intended *only* to feed `loadFrozenContainerFromPendingState`, or do partners (Loop, per #23301) consume the captured state for inspection / state-relay independent of offline rehydrate?

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Confirmed in `e7236fc`. The new `deprecate-create-and-load-container-props.md` changeset surfaces the deprecation through the release-notes channel and names the migration targets (`IContainerHostProps` / `IContainerDriverServices` / their composition) explicitly. Reasonable call to defer the concrete removal-version milestone to the major-release plan that drops the alias — `@deprecated since` tags in JSDoc do drift from the actual schedule. Thanks.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Confirmed in `e7236fc`. The synthesized `request.url` is now the recognizable opaque placeholder `"frozen-load-from-pending-state://offline-placeholder"` and the synthesized `IUrlResolver.resolve()` ignores its argument, returning the `IResolvedUrl` derived from `pending.url` regardless. The behavior is unchanged but the layering smell is gone — no resolved-form URL routed through the request slot. Comment block on the constant captures the rationale. Thanks.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Confirmed in `e7236fc`. The new regression test in `loadFrozenContainerFromPendingState.spec.ts` ("rejects offline load when pendingLocalState.url is non-conforming") pins the offline-form `UsageError` contract on a mutated single-segment URL, and the companion capture-time test in `captureFullContainerState.spec.ts` pins the producer side. The contract is now locked at both ends of the round-trip — exactly the gap that let #19854 silently regress the helper. Thanks.

packages/loader/container-loader/src/frozenServices.ts
anthony-murphy: Deep Review: Acknowledged — agreed that with all four changesets removed in c53a55a, the `Error` → `UsageError` change ships purely on `@legacy @alpha` surface (`createFrozenDocumentServiceFactory` is `@alpha @legacy`) and doesn't need changeset coverage under the repo's convention. Thanks for the cleanup.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Claude here on behalf of @anthony-murphy — pushing back on this one: the only point of `captureFullContainerState` is to eventually feed `loadFrozenContainerFromPendingState`. There is no separate "inspection / state-relay" consumer; the docstring framing of those terms is all in service of eventual rehydration. Under that design, the capture-time gate is exactly as broad as the function's actual contract: surfacing the shape requirement at the producer call is strictly better than letting a partner serialize and persist a poisoned artifact only to learn at load time (different process, possibly days later) that it isn't rehydratable. So no docstring narrowing or changeset note is needed — the constraint isn't a *new* narrowing of the function, just enforced at the earliest point in the round-trip.

anthony-murphy:
## Deep Review

*Reviewed commit `6ad3168` on 2026-05-26.*

No actionable code findings remain. The prior Tier 2 on the capture-time URL-shape gate (thread `3306477725`) was rebutted by the author with a substantive contract argument — `captureFullContainerState` has no consumer other than `loadFrozenContainerFromPendingState`, so the eager gate matches the actual contract — and the thread is now resolved. The remaining concerns (single-entry vs sibling-function design, `AllOrNone<T>` placement in `core-interfaces`, `getAbsoluteUrl` throws-vs-undefined, `request.url` placeholder) are soft design-preference items with explicit "keep PR" recommendations and don't block sign-off. Remaining gate is reviewer sign-off — primarily markfields on the capture-side contract framing.

### Context for Reviewers

- `tryParseCompatibleResolvedUrl`'s non-throwing contract — established by markfields in #17850 (2023-10-19), silently regressed in #19854 (2024-03-01) when the body was rewritten to use unguarded `new URL(url)` — is restored by this PR's `utils.ts` hunk. The offline path is the first caller for which the gap mattered in practice; treat the change as a real bug fix, not cosmetic.
- Load-side bookend of the multi-PR frozen-container series (#25538, #25590, #25603, #25653, #27211, #27220). #27220 (markfields) introduced `captureFullContainerState` as the inlined-blob counterpart; this PR is its consuming end, and the offline blob round-trip in `loadFrozenContainerFromPendingState.spec.ts` is the integration point proving the series hangs together.
- The capture-time URL-shape gate in `captureFullContainerState` (`createAndLoadContainerUtils.ts:568-572`) is now the eager half of the contract that the offline-load branch (`:416-421`) also enforces. Author has stated on thread `3306477725` that `captureFullContainerState` exists only to feed `loadFrozenContainerFromPendingState`, so the producer-side gate matches the function's actual contract; markfields (owner per #27220) is the right voice on whether that framing is the intended one.
- `IContainerHostProps` / `IContainerDriverServices` / `IContainerLoadDriverProps` decomposition continues the layering trajectory of #3945 (2020) and #5175 (2021); #23301 (2024-12-12) promoted these props to `@legacy @beta` so Loop could consume them in production. `ICreateAndLoadContainerProps` is kept as a `@deprecated` alias for back-compat — TS will emit deprecation diagnostics in partner builds.
- Offline implementation preserves the Container/Loader layering boundary established in #5175: offline behavior is composed entirely at the entry-point layer (synthesizes `IUrlResolver`, `IResolvedUrl`, `IRequest`, frozen `IDocumentServiceFactory`, then delegates to `loadExistingContainer`). The synthesized `request.url` carries a recognizable opaque placeholder (`"frozen-load-from-pending-state://offline-placeholder"`), neutralizing the smell of routing a resolved-form URL through the request slot.
- `AllOrNone<T>` lands as a `@legacy @alpha` primitive in `core-interfaces` paired with `@internal` `validateAllOrNone` in `core-utils`. Today's tagging is internally consistent (one alpha caller). #23301's precedent means a future prod consumer would need a retag PR; if a second consumer doesn't materialize within a release or two, the primitive is worth re-evaluating for inlining. Last to touch both package index files: DLehenbauer (#19783 / #19784).
- Suggested reviewers: **markfields** (`tryParseCompatibleResolvedUrl` naming in #17850; authored `captureFullContainerState` in #27220 — load-bearing on the capture-time gate framing and `AllOrNone` placement), **ChumpChief** (last touched `tryParseCompatibleResolvedUrl` in #19854 and established Container/Loader layering in #5175; right sign-off for the synthetic-resolver pattern), **jatgarg** (added `@legacy` tags in #23301; relevant for the `ICreateAndLoadContainerProps` `@deprecated` reshape), **dannimad** (authored `createFrozenDocumentServiceFactory` in #25653 and the test scaffold in #25603; natural fit for `frozenServices.spec.ts`), **DLehenbauer** (cross-package `AllOrNone` additions if CODEOWNERS doesn't auto-trigger).

- **Needs human judgment (markfields)** — Owner of `captureFullContainerState` (#27220) and the `tryParseCompatibleResolvedUrl` contract (#17850). Right voice on whether the eager capture-time gate matches the intended contract of `captureFullContainerState` — author asserts in `3307998521` that the function has no non-offline consumer; pipeline cannot independently disprove the no-other-consumer claim.
- **Needs human judgment (markfields)** — Whether `AllOrNone<T>` belongs as a publicly-exported `@legacy @alpha` primitive in `core-interfaces` with one in-tree consumer, or should remain inlined at the call site until a second consumer materializes.
- **Needs human judgment (markfields)** — Offline + `readOnly: false` semantics: local DDS submissions accrue without being published. Owner sign-off on the "capture → restart offline → keep editing → re-capture and replay online later" contract.
- **Needs human judgment (markfields / ChumpChief)** — Offline `getAbsoluteUrl` deliberately throws `UsageError` rather than fabricating a string. Well-reasoned but a contract call worth confirming.
- **Needs human judgment (ChumpChief)** — Sign-off on the synthetic-resolver pattern as an acceptable internal caller of the `request.url ≡ resolvedUrl.url` identity guard in `Loader.resolveCore`; #5175 walked the codebase the other direction.
- **Needs human judgment (jatgarg)** — `@deprecated` reshape of `ICreateAndLoadContainerProps` into `IContainerHostProps` / `IContainerDriverServices` bases. Structurally identical so existing Loop call sites still compile, but TS will emit deprecation diagnostics; #23301 explicitly carved this surface for Loop and a Loop-prod usage check / migration-window confirmation is human-scope.
- **Cannot be assessed by the pipeline** — Whether the offline blob round-trip E2E test catches realistic consumer workloads beyond a single primitive-DDS + single attachment-blob scenario; runtime determinism across resolver implementations; whether future `Loader.resolveCore` / `Container.load` evolution will preserve the "`request.url` is opaque-until-resolver in offline mode" invariant.

</details>

<details><summary>Review history (10 prior reviews)</summary>

- `e7236fc` 2026-05-26 · **6/10** — capture-time URL-shape gate in `captureFullContainerState` flagged as Tier 2 (resolver-shape narrowing)
- `9558d1f` 2026-05-26 · **6/10** — missing regression tests for restored `tryParseCompatibleResolvedUrl` non-throwing contract and offline URL-shape precondition
- `86d11ed` 2026-05-26 · **8/10** — Tier 3 polish items remain (URL-shape capture-time hardening, `@deprecated` milestone, three new inline)
- `9f26493` 2026-05-26 · **8/10** — four Tier 3 polish items flagged inline on the `@legacy @alpha` surface
- `2bcc5c0` 2026-05-26 · **9/10** — all prior inline threads resolved; changeset removal in c53a55a closes release-notes polish items
- `463f9aa` 2026-05-21 · **8/10** — three Tier 3 polish items remain — flagged inline
- `5efb38e` 2026-05-21 · **5/10** — `quiet-shrews-marry.md` cross-package changeset gap for `AllOrNone` / `validateAllOrNone`
- `ce2ede7` 2026-05-20 · **5/10** — `resolvedUrl.id` two-segment shape divergence still unaddressed at this commit
- `faa6a1f` 2026-05-20 · **5/10** — `resolvedUrl.id` two-segment `tenantId/docId` shape divergence flagged inline on the offline path
- `9213320` 2026-05-20 · **8/10** — both prior correctness concerns resolved; `AllOrNothing` API-surface item flagged inline

</details>

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Accepting the rebuttal — if `captureFullContainerState` has no consumer outside the offline-rehydration round-trip, surfacing the resolver-shape requirement at the producer is strictly better than letting a partner persist a poisoned artifact and discover it at load time in a different process. Pipeline can't independently confirm the no-other-consumer claim; flagging that contract framing for **markfields** (owner per #27220) for explicit sign-off and dropping the docstring/changeset narrowing ask. Thanks for the pushback.

dhr-verma (approved):

jzaffiro (approved):

.changeset/deprecate-create-and-load-container-props.md
```diff
@@ -0,0 +1,29 @@
+---
+"@fluidframework/container-loader": minor
+"__section": deprecation
+---
+Deprecate `ICreateAndLoadContainerProps` in favor of composable building blocks
```
Josmithr: Our changeset guidelines call to avoid using `codeSyntax` in changeset titles (everywhere else is fine and encouraged)
```suggestion
Deprecate ICreateAndLoadContainerProps in favor of composable building blocks
```

packages/common/core-interfaces/src/allOrNone.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @remarks
+ * The "all-or-none" shape is a small, general, composable primitive — it
+ * shows up wherever a feature is wired up by supplying a group of cooperating
+ * objects (driver triples, transport plus credentials, etc.) and is awkward
+ * to spell out inline. Naming it lets callers reuse the same pattern and lets
+ * reviewers see the intent at the declaration site instead of re-deriving it
+ * from a hand-rolled discriminated union. The mapped form
+ * `{ [K in keyof T]?: never }` is mechanical, so downstream consumers do not
+ * need an unusual TS feature to use it.
```
Josmithr:
```suggestion
 * need any unusual TS feature to use it.
```

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -78,22 +85,16 @@ interface SummarizerLike {
 }
 
 /**
- * Properties necessary for creating and loading a container.
+ * Host-level container loader properties — the code loader plus all the
+ * optional policy / observability fields that aren't tied to driver wiring.
+ *
+ * Extracted as a reusable building block so other props types (create,
```
Josmithr, line 92: Nit
```suggestion
 * @remarks
 * Extracted as a reusable building block so other props types (create,
```

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -139,11 +140,73 @@ export interface ICreateAndLoadContainerProps {
 	readonly clientDetailsOverride?: IClientDetails | undefined;
 }
 
+/**
+ * The driver-services pair — `urlResolver` plus `documentServiceFactory` —
+ * required to wire a container to a real driver at create or load time.
+ *
+ * Extracted as a reusable building block so the `request` field can be
```
Josmithr, line 149: Nit
```suggestion
 * @remarks
 * Extracted as a reusable building block so the `request` field can be
```

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -139,11 +140,73 @@ export interface ICreateAndLoadContainerProps {
...
+	 * for communication with the container's server.
+	 */
+	readonly documentServiceFactory: IDocumentServiceFactory;
+}
+
+/**
+ * The load-time driver wiring trio — `request`, `urlResolver`, and
+ * `documentServiceFactory` together.
+ *
+ * Reused as the all-or-nothing group for entry points (e.g. the frozen-load
```
Josmithr, line 176: Nit
```suggestion
 * @remarks
 * Reused as the all-or-nothing group for entry points (e.g. the frozen-load
```

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -238,10 +305,46 @@ export async function loadExistingContainer(
 
 /**
  * Properties required to load a frozen container from pending state.
+ *
+ * Two forms are supported and selected by the presence of the driver-wiring
```
Josmithr, line 313: Nit
```suggestion
 * @remarks
 * Two forms are supported and selected by the presence of the driver-wiring
```

Josmithr (approved):
Left a few docs suggestions. Otherwise, API changes look good.

dhr-verma (approved):

.changeset/deprecate-create-and-load-container-props.md
```diff
@@ -0,0 +1,29 @@
...
+// Add only what you need
+interface MyHostOnlyProps extends IContainerHostProps {
+    readonly extraOption: boolean;
+}
+```
+
+- `IContainerHostProps` covers the code loader plus optional policy / observability fields (`options`, `scope`, `logger`, `configProvider`, `protocolHandlerBuilder`, `allowReconnect`, `clientDetailsOverride`).
+- `IContainerDriverServices` covers the `urlResolver` + `documentServiceFactory` pair.
+
+`ICreateAndLoadContainerProps` will be removed in a future major release.
```
jason-ha, line 29: All deprecations should have a tracking issue. Please use the API deprecation guide.

PR #15550, opened 2023-05-10
Improve telemetry in fluid cache and snapshot fetch in odsp driver

jatgarg:
## Description

Record pkg version in cache.
Record trees, blobs, size in ObtainSnapshot event.

PR #15549, opened 2023-05-10
Improve odsp driver fluid cache perf

jatgarg:
## Description
Item: https://dev.azure.com/fluidframework/internal/_workitems/edit/4040
Improve odsp driver fluid cache performance. Avoid closing db after each access and rather schedule close so that it can be used many times in between closes.
Schedule idle task after 2 sec for first time.

PR #14315, opened 2023-02-24
Fluid Cache: Close DB more consistently, add optional delete DB callbacks

JasonGore:
## Description

Consistently close DB after operations in FluidCache. Allow optional listeners on DB deletion to check for blocked situations as needed.

FluidCache return paths can leave DB open, delaying and potentially blocking DB deletions. Additionally, having delete callbacks can help identify when these situations will occur ahead of time so the delete caller can act appropriately.

I would like to add unit tests to ensure DB closure but also have an urgency to consume this as quickly as possible. I will follow up in another PR if needed.

PR #27348, opened 2026-05-19
feat(driver-web-cache): add update for atomic read-modify-write

anthony-murphy:
## Description

Adds `update(entry, updater)` to `FluidCache`, an atomic read-modify-write. The currently-cached value is read and the updater callback decides whether — and what — to write, inside a single IndexedDB `readwrite` transaction. This gives consistent update semantics across consumers sharing the same underlying IndexedDB instance (for example, multiple browser tabs racing to persist offline pending state).

The updater is invoked with `(existing, set)`. To commit a write, call `set(value)`; to leave the cache untouched, return without calling `set`. This shape covers both the simple conditional-overwrite case and read-modify-write where the new value is derived from the existing one — the latter is not expressible with a predicate-only API.

```ts
// Conditional overwrite (active tab wins): only write if our revision is higher.
const wrote = await fluidCache.update(entry, (existing, set) => {
    const existingRev = (existing as { rev?: number } | undefined)?.rev ?? -1;
    if (mine.rev > existingRev) {
        set(mine);
    }
});

// Read-modify-write: increment a counter atomically.
await fluidCache.update(entry, (existing, set) => {
    const prev = (existing as { count: number } | undefined)?.count ?? 0;
    set({ count: prev + 1 });
});
```

Contract details:

- `existing` is `undefined` in any case where the row would not be visible to `get`: no entry exists for the key, the existing entry belongs to a different partition, or the existing entry is older than `maxCacheItemAge`.
- Both the updater body and the `set` call must run synchronously. IndexedDB transactions close automatically on any non-IDB await, which would silently break the atomicity that makes the update correct. Calling `set` after the updater has returned throws a `UsageError` at the call site so misuse (for example, invoking `set` from a `setTimeout`) is visible rather than silently lost.
- If the updater calls `set` more than once, the last value wins.
- If the updater throws — including after calling `set` — the transaction is aborted, the existing row is preserved, and the error is logged under the dedicated `FluidCacheUpdateCallbackError` telemetry event (distinct from IDB-write failures under `FluidCachePutError`). The call returns `false`.
- When `set` is called, the write atomically replaces whatever row exists at the key, including cross-partition or stale rows the updater saw as `undefined` (matching the unconditional overwrite behavior of `put`).
- Returns `true` if `set` was called and the write committed, `false` otherwise. IDB errors are logged and not thrown, matching `put`.

API surface is additive on the existing `@legacy @beta` class. Per-package type tests mark `Class_FluidCache forwardCompat: false` to acknowledge that adding methods to a class is a forward-compat type-break by definition.

## Reviewer Guidance

Specific things I'd appreciate eyes on:

- The IDB-transaction atomicity argument. The crucial invariant is that the updater body — including the `set` call — runs synchronously between `tx.store.get` and `tx.store.put`, so the transaction never yields to the event loop between the read and the decision. If you see a way that any await could sneak in there (via the updater, the `set` closure, the `idb` library's promise wrapping, or the `existing?.partitionKey` access), please flag it.
- The `set`-after-return guard: `set` captures an `updaterReturned` flag from the enclosing scope and throws `UsageError` if invoked after the updater returns. This is meant to make deferred-`set` misuse (for example, calling `set` from a `setTimeout`) loud at the misuse site rather than silently writing into a dead transaction.
- The visibility rules `update` applies to `existing` (different partition / past `maxCacheItemAge` are reported as `undefined`). The intent is that the updater sees the same view it would under a `get`+`put`, so cross-partition and stale rows are not exposed to updater logic, but the write still atomically replaces them when `set` is called (matching the unconditional overwrite of `put`).
- The throw-after-`set` policy: if the updater calls `set` then throws, the transaction is aborted and the existing row is preserved. The reasoning is that the updater's intent is ambiguous in that case and preserving existing data is the safer default. Worth confirming this matches host expectations.


packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -168,6 +214,96 @@ export class FluidCache implements IPersistedCache {
...
+	 * because `removeEntries` deletes rows regardless of partition.
+	 *
+	 * Note: `BroadcastChannel` does not echo a message back to the instance that posted
+	 * it, so writes performed by *this* `FluidCache` do not trigger its own listeners.
+	 * Other `FluidCache` instances (including ones in the same tab) will receive them.
+	 *
+	 * @returns a function that unregisters the listener. Idempotent.
+	 */
+	public onChange(listener: (event: FluidCacheChangeEvent) => void): () => void {
+		if (this.disposed) {
```
anthony-murphy: Deep Review: Post-dispose contract is asymmetric — only `onChange` throws; mutations and reads silently reopen the DB.

`onChange` here checks `this.disposed` and throws `UsageError`. None of `get`, `put`, `putIf` (line 538), `removeEntry` (line 385), `removeEntries` (line 355) check `this.disposed` — they all call `await this.openDb()`, which (per the companion thread on line 314) will lazily create a fresh DB connection and arm a new close timer even on a disposed instance. The JSDoc at lines 260–263 says "other methods are unsafe to call" but the actual behavior is *silent resurrection*, which is worse than either throwing or being a no-op.

The PR description itself flags this as an open design question ("is that the right contract, or should `put`/`get`/`removeEntry` also throw after dispose?"). Combined with the in-flight `openDb()` race, this makes silent post-dispose mutations a real lifecycle hazard, not a stylistic preference.

**Fix**: add an early `if (this.disposed) throw new UsageError("FluidCache disposed")` guard to `put`, `putIf`, `removeEntry`, and `removeEntries`. For `get`, decide deliberately and document — either throw to match the mutating path, or resolve `undefined` so shutdown read paths degrade safely. Whichever you pick, replace the JSDoc "unsafe to call" wording at lines 260–263 with the actual contract. Pair with the line-314 guard so the resurrection race cannot fire from a post-dispose caller. Add tests for each public method exercising the post-dispose path.

anthony-murphy: Addressed in d4e7a629e8.

- `put`, `putIf`, `get`, `removeEntry`, `removeEntries` now throw `UsageError` after `dispose` (uniform contract, matches `onChange`). Catch blocks rethrow `UsageError` when `this.disposed` is true so dispose-mid-flight rejects cleanly instead of silently swallowing.
- `openDb` now checks `this.disposed` immediately after awaiting `getFluidCacheIndexedDbInstance` and closes the just-acquired connection + throws, so an in-flight `openDb` can no longer resurrect the DB or arm a fresh close timer past `dispose()`.
- `dispose()` JSDoc updated to state the actual contract (throw on all methods, in-flight calls reject, no resurrection). README and changeset updated to match.
- Tests added for each public method's post-dispose throw (both `immediateClose` modes) plus an in-flight `dispose` race against an awaiting `put` ΓÇö 74 passing.

anthony-murphy: Deep Review: Confirmed fixed on 4b389ab — `throwIfDisposed()` is now uniform across `put`/`putIf`-success/`removeEntry`/`removeEntries` after the awaited write, the `broadcast()` early-return is solid defence in depth, and the idle-task callbacks now guard against the construct-then-dispose sequence. Resolving this thread; two new Tier 2 gaps in the same dispose-contract family at different anchors are posted as fresh inline threads.

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -360,5 +578,113 @@ export class FluidCache implements IPersistedCache {
...
+			// than `maxCacheItemAge`. Cross-partition and stale entries are treated as
+			// absent so the predicate sees the same view it would under `get`+`put`.
+			const existingVisible =
+				existing?.partitionKey === this.partitionKey &&
+				Date.now() - existing.createdTimeMs <= this.maxCacheItemAge;
+			const existingValue = existingVisible ? existing?.cachedObject : undefined;
+
+			if (!shouldWrite(existingValue, value)) {
+				await transaction.done;
+				return false;
```
anthony-murphy, line 505: Deep Review: `putIf`'s rejected-write branch breaks the in-flight dispose contract.

The reject branch does:

```ts
if (!shouldWrite(existingValue, value)) {
    await transaction.done;
    return false;
}
```

It awaits `transaction.done` and returns `false` with no `this.throwIfDisposed()` check. The symmetric successful-`putIf` branch (a few lines below) deliberately calls `this.throwIfDisposed()` immediately after `await transaction.done`; `put` does the same after its awaited write; `removeEntry` does the same after `await transaction.done`. The changeset (`.changeset/driver-web-cache-onchange.md`) and JSDoc promise: *"Operations that were already in flight when `dispose` was called also reject with a `UsageError`."*

A caller whose predicate returned `false` while `dispose()` ran in another microtask sees `false` (a valid contract value meaning *predicate rejected*) — indistinguishable from a genuine cooperative rejection — instead of the promised `UsageError`.

**Fix**: add `this.throwIfDisposed()` immediately after `await transaction.done` in the reject branch, before `return false`. Add an in-flight test mirroring the existing `does not broadcast a change event when dispose runs mid-flight` test but where `shouldWrite` returns `false` — assert the awaiting `putIf` rejects with `UsageError` rather than resolving `false`.

packages/drivers/driver-web-cache/api-report/driver-web-cache.legacy.beta.api.md
```diff
@@ -10,16 +10,31 @@ export function deleteFluidCacheIndexDbInstance(deleteDBCallbacks?: DeleteDBCall
...
+    putIf(entry: ICacheEntry, value: unknown, shouldWrite: (existing: unknown, proposed: unknown) => boolean): Promise<boolean>;
     // (undocumented)
     removeEntries(file: IFileEntry): Promise<void>;
     // (undocumented)
     removeEntry(entry: ICacheEntry): Promise<void>;
 }
 
+// @beta @legacy
+export type FluidCacheChangeEvent = {
+    readonly op: "put" | "remove";
```
anthony-murphy:
```suggestion
    readonly type: "put" | "remove";
```

packages/drivers/driver-web-cache/api-report/driver-web-cache.legacy.beta.api.md
```diff
@@ -10,16 +10,31 @@ export function deleteFluidCacheIndexDbInstance(deleteDBCallbacks?: DeleteDBCall
...
+    readonly partitionKey: string | null;
+    readonly fileId: string;
+    readonly type: string;
+    readonly cacheItemId: string;
+} | {
+    readonly op: "removeFile";
```
anthony-murphy:
```suggestion
    readonly type: "removeFile";
```

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -69,6 +69,48 @@ export interface FluidCacheConfig {
...
+			readonly op: "removeFile";
+			readonly fileId: string;
+	  };
+
+/**
+ * Name of the `BroadcastChannel` used to deliver cache-change notifications between
+ * `FluidCache` instances. A single channel is used for the entire driver cache; partition
+ * scoping is applied by the receiving listener.
+ */
+const FluidCacheBroadcastChannelName = "fluid-driver-cache";
```
anthony-murphy: add a guid to this as a post-fix, so it will be unique

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -168,14 +236,135 @@ export class FluidCache implements IPersistedCache {
...
+		// available (e.g. older environments), notifications silently degrade to a no-op.
+		if (typeof BroadcastChannel !== "undefined") {
+			this.broadcastChannel = new BroadcastChannel(FluidCacheBroadcastChannelName);
+			this.broadcastChannel.addEventListener("message", (messageEvent: MessageEvent) => {
+				this.dispatchChangeEvent(messageEvent.data as FluidCacheChangeEvent);
+			});
+			// In Node, BroadcastChannel keeps the event loop alive until closed. Calling
+			// `unref` (when available) lets test processes and short-lived scripts exit
+			// even if a FluidCache instance was never disposed.
+			(this.broadcastChannel as unknown as { unref?: () => void }).unref?.();
```
anthony-murphy: gross. we should just ensure our tests dispose

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -85,6 +127,10 @@ export class FluidCache implements IPersistedCache {
 	private dbCloseTimer: ReturnType<typeof setTimeout> | undefined;
 	private dbReuseCount: number = -1;
 
+	private readonly broadcastChannel: BroadcastChannel | undefined;
+	private readonly changeListeners = new Set<(event: FluidCacheChangeEvent) => void>();
```
anthony-murphy: there is a internal eventing library. can we re-use that instead of rolling out own

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -168,14 +236,135 @@ export class FluidCache implements IPersistedCache {
...
+	 * because `removeEntries` deletes rows regardless of partition.
+	 *
+	 * Note: `BroadcastChannel` does not echo a message back to the instance that posted
+	 * it, so writes performed by *this* `FluidCache` do not trigger its own listeners.
+	 * Other `FluidCache` instances (including ones in the same tab) will receive them.
+	 *
+	 * @returns a function that unregisters the listener. Idempotent.
+	 */
+	public onChange(listener: (event: FluidCacheChangeEvent) => void): () => void {
+		if (this.disposed) {
```
anthony-murphy: use throwIfDisposed

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -168,14 +236,135 @@ export class FluidCache implements IPersistedCache {
...
 
 	private async openDb(): Promise<IDBPDatabase<FluidCacheDBSchema>> {
+		this.throwIfDisposed();
 		if (this.closeDbImmediately) {
-			return getFluidCacheIndexedDbInstance(this.logger);
+			const dbInstance = await getFluidCacheIndexedDbInstance(this.logger);
+			// Dispose may have run while we awaited the IDB open. Close the just-acquired
+			// connection and reject so callers do not silently operate on a disposed cache.
+			if (this.disposed) {
+				dbInstance.close();
```
anthony-murphy: should we make throw if disposed take an optional lambda that allows local clean up. seems we have a few cases

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -360,5 +583,118 @@ export class FluidCache implements IPersistedCache {
...
+	): Promise<boolean> {
+		this.throwIfDisposed();
+		let db: IDBPDatabase<FluidCacheDBSchema> | undefined;
+		let wrote = false;
+		try {
+			db = await this.openDb();
+
+			const key = getKeyForCacheEntry(entry);
+			const transaction = db.transaction(FluidDriverObjectStoreName, "readwrite");
+			const existing = await transaction.store.get(key);
```
anthony-murphy, line 438: should we be usings a cursor rather than a transaction? i'm not sure

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Deep Review: Confirmed fixed on 658765a — the reject branch now calls `this.throwIfDisposed()` after `await transaction.done` before `return false`, symmetric with the success branch. Resolving.

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -259,9 +466,19 @@ export class FluidCache implements IPersistedCache {
 		} finally {
 			this.closeDb(db);
 		}
+		if (removed) {
+			this.broadcast({
+				op: "remove",
+				partitionKey: this.partitionKey,
```
anthony-murphy: Deep Review: Cross-partition writes silently invalidate other partitions' cached rows without notifying them.

`getKeyForCacheEntry(entry)` does not include `partitionKey` in the IDB key, yet every per-entry broadcast here (and in `put` ~line 589, `putIf` ~line 676) stamps `partitionKey: this.partitionKey`, and `dispatchChangeEvent` drops events whose `event.partitionKey !== this.partitionKey`. So a partition-A `put` / successful `putIf` / `removeEntry` overwrites or deletes a partition-B row in IDB, but partition B's listener never receives the event because it is filtered as partition A. Partition B's next `get` returns `undefined` for what was previously a present row with zero notification, defeating the point of `onChange` for that listener.

The cross-partition overwrite *policy* for `putIf` is settled (issue-comment thread, point 3 — *"`putIf` is a guarded `put`, not a more conservative one"*) and documented in the README/changeset/tests. The unstated edge case is the *broadcast routing* on that cross-partition path.

**Fix**: inside the same `readwrite` transaction (mirroring `putIf`'s read-then-write), read the existing record's `partitionKey` before overwriting/deleting. If it differs from `this.partitionKey`, also broadcast an invalidation event stamped with the *existing* partition's key, so the partition that actually cached the row is notified. Same shape as the fix for the `removeEntry` over-broadcast (separate thread below). No storage-key change required.

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -360,5 +583,118 @@ export class FluidCache implements IPersistedCache {
...
+			await transaction.store.put(this.buildRecord(entry, value), key);
+			await transaction.done;
+			// Dispose may have landed during the awaits above. Reject before reporting
+			// success or broadcasting so the caller observes the disposed state cleanly
+			// and no `put` event escapes after teardown.
+			this.throwIfDisposed();
+			wrote = true;
+		} catch (error: any) {
+			this.throwIfDisposed();
+			this.logger.sendErrorEvent(
```
anthony-murphy, line 533: Deep Review: `throwIfDisposed()` in `catch` arms discards the underlying IDB error.

This pattern (also at lines ~432, ~460, ~547, ~578) does:

```ts
} catch (error: any) {
    this.throwIfDisposed();
    this.logger.sendErrorEvent(...);
}
```

When a real IDB failure (e.g. a quota error) and `dispose()` race, the catch-arm `throwIfDisposed()` re-throws `UsageError` *before* `sendErrorEvent` runs — the original IDB error is silently dropped. The success-path post-await guards are fine (no error to lose); the catch-path guards are the issue.

**Fix**: log the original error (or attach it as `{ cause: error }` on the `UsageError`) *before* `throwIfDisposed()` throws, so the diagnostic survives the dispose-race path. Apply to each of the five catch arms.

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -360,5 +583,118 @@ export class FluidCache implements IPersistedCache {
...
+			// Surface the cached value to the predicate only when the existing entry is
+			// visible under the same rules `get` applies: same partition and not older
+			// than `maxCacheItemAge`. Cross-partition and stale entries are treated as
+			// absent so the predicate sees the same view it would under `get`+`put`.
+			const existingVisible =
+				existing?.partitionKey === this.partitionKey &&
+				Date.now() - existing.createdTimeMs <= this.maxCacheItemAge;
+			const existingValue = existingVisible ? existing?.cachedObject : undefined;
+
+			if (!shouldWrite(existingValue, value)) {
```
anthony-murphy: Deep Review: `putIf` predicate exceptions are logged under `FluidCachePutError`, conflating host-bug telemetry with IDB-write failures.

`shouldWrite()` is invoked inside the `try` here, so any predicate throw lands in the same catch that handles genuine IDB write failures — it logs `FluidCachePutError` with the predicate error and returns `false`. The test at `FluidCache.test.ts:824-838` documents the swallow-and-return-`false` behavior but does not assert what event is logged. `putIf` is brand-new ingress for caller-supplied callbacks, so the wrong observability default will stick.

**Fix**: either invoke `shouldWrite` outside the `try/catch` (it's synchronous by design — predicate exceptions then propagate naturally without polluting the IDB-error event) and `tx.abort()` if it throws, or wrap the predicate call in its own try and log under a distinct telemetry event (e.g. `FluidCachePutIfPredicateError`).

packages/drivers/driver-web-cache/.mocharc.cjs
```diff
@@ -10,4 +10,9 @@ const getFluidTestMochaConfig = require("@fluid-internal/mocha-test-setup/mochar
 // Tests use fake-indexeddb v3 (CJS-only) and require browser-like globals.
 // setup.cjs installs window/self aliases and fake-indexeddb before tests run.
 const config = getFluidTestMochaConfig(__dirname, [`${__dirname}/src/test/setup.cjs`]);
+// Node's BroadcastChannel (used by FluidCache for cross-instance change notifications)
```
anthony-murphy: Deep Review: `config.exit = true` reverses PR #27188's deliberate test-hygiene invariant for this exact package.

The diff adds `config.exit = true;` to `.mocharc.cjs`. PR #27188 (Josmithr, merged 2026-04-30) deliberately engineered the migrated Mocha tests for this package *not* to need `exit: true` — Josmithr's own review comment on #27188 line 204 reads: *"This was needed in order to not specify `exit: true` in the mocha config."* The current PR re-introduces it because Node's `BroadcastChannel` keeps the event loop alive after `close()`. A package-wide `exit: true` will mask any unrelated future handle leak in this package's tests, not just the BroadcastChannel one — exactly the regression #27188 worked to prevent.

The test scaffold already includes an `unref()` cast (PR description §"Reviewer Guidance" bullet 4) and `afterEach` already calls `dispose()` (`FluidCache.test.ts:84-88`), so a narrower escape hatch is achievable. Your own inline self-comment on this file (*"gross. we should just ensure our tests dispose"*) reads the same way.

**Fix** — pick one:

1. `.unref()` the production `BroadcastChannel` handle inside `FluidCache` itself (so production code doesn't keep the Node event loop alive — the `unref()` cast in test setup is the existing precedent), then drop `config.exit = true`. Cleanest.
2. Restrict `exit: true` to a sub-config used only by the BroadcastChannel-using `describe` block (env flag or sub-pattern in `.mocharc.cjs`).
3. If neither is feasible, document at `.mocharc.cjs:13` and in the changeset why `unref()` alone is insufficient (e.g. fake-indexeddb still holds a refcount that overrides `BroadcastChannel.unref`), and tag Josmithr explicitly so the #27188 invariant owner signs off on the reversal.

anthony-murphy: Fixed in `9ea3fa2` — option (1) from your list. Production `BroadcastChannel` now `unref()`s itself (typed cast because the web interface has no `unref`; Node's implementation does, browsers no-op via the optional chain). `config.exit = true` dropped from `.mocharc.cjs`. 88 tests pass and mocha exits cleanly without the flag, restoring #27188's invariant.

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Fixed in `9ea3fa2`. `put`/`putIf`/`removeEntry` each now wrap the read+write/delete in a single `readwrite` transaction, capture `existing.partitionKey` when it differs from `this.partitionKey`, and broadcast an additional `remove` event stamped with the displaced partition's key after the primary broadcast. `removeEntry` also no longer broadcasts when no row existed at the key (previously over-broadcast). Three new tests in `change event notifications` cover the cross-partition routing for `put`/`putIf`/`removeEntry` plus a no-row-no-broadcast test.

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Sticking with the readwrite transaction for now — rationale captured in the JSDoc on `putIf` (the get/put pair inside one `readwrite` transaction gives the same atomicity as a cursor for a single-key CAS, and is materially simpler to reason about). A cursor would be the right tool for range scans or iteration; for a known key it is just extra ceremony. Happy to revisit if you want the cursor form — this is the only place in the file that does a read-then-write on a known key.

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: The five catch arms (`putIf`/`put`/`removeEntry`/`removeEntries`/`getItemFromCache`) already call `logger.sendErrorEvent(..., error)` before `throwIfDisposed()` — e.g. `put` at lines 599-608, `putIf` at 725-732, `getItemFromCache` at 563-575. So the original IDB error survives the dispose-race path via telemetry. The pattern the review snippet showed (throw-first-then-log) does not appear on this SHA. Resolving.

packages/drivers/driver-web-cache/api-report/driver-web-cache.legacy.beta.api.md
anthony-murphy: Accepted; api report now shows `readonly type: "put" | "remove"`.

packages/drivers/driver-web-cache/api-report/driver-web-cache.legacy.beta.api.md
anthony-murphy: Accepted; api report now shows `readonly type: "removeFile"`.

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Done — channel name is `fluid-driver-cache:b3e4f5a6-7c8d-49e0-a1b2-c3d4e5f6a7b8` (fixed namespace suffix so every `FluidCache` in the browsing context joins the same channel while staying unlikely to collide with unrelated host code). Rationale captured in the JSDoc on `FluidCacheBroadcastChannelName`.

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Resolved as part of the `.mocharc.cjs` thread fix — tests already dispose via `afterEach` and the production `BroadcastChannel` now `unref()`s itself, so `config.exit = true` is gone.

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Done — `throwIfDisposed` is applied at the top of every public method, after every post-await success guard, and inside `openDb` after the IDB-open await.

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Done — `throwIfDisposed` accepts an optional `cleanup` callback (FluidCache.ts:335-340) used by `openDb` to close a just-acquired IDB connection before throwing.

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Done — `shouldWrite` is invoked inside its own `try` (FluidCache.ts:689-706), and predicate exceptions are logged under `FluidCachePutIfPredicateError` (distinct from `FluidCachePutError` for IDB-write failures), after aborting the transaction so the existing row is preserved.

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -318,47 +595,240 @@ export class FluidCache implements IPersistedCache {
...
+			// (e.g. the database version having upgraded underneath us), so we
+			// still want the original failure visible in telemetry.
 			this.logger.sendErrorEvent(
 				{ eventName: FluidCacheErrorEvent.FluidCacheGetError, pkgVersion },
 				error,
 			);
+			// If dispose ran during the operation, surface that to the caller instead of
+			// silently returning undefined, matching the post-dispose throw contract.
+			this.throwIfDisposed();
 			this.closeDb(db);
```
anthony-murphy, line 327: Deep Review: `getItemFromCache`'s catch arm leaks the IndexedDB connection under a dispose race in `closeDbImmediately: true` mode (the default per `FluidCache.ts:147`).

The catch arm reads:

```ts
} catch (error: any) {
    this.logger.sendErrorEvent(
        { eventName: FluidCacheErrorEvent.FluidCacheGetError, pkgVersion },
        error,
    );
    this.throwIfDisposed();   // throws UsageError → next line never runs
    this.closeDb(db);
    return undefined;
}
```

When the success-path `throwIfDisposed()` (post `await db.get` at line 568) fires under a dispose race, control enters the catch with `error = UsageError`; the catch then re-throws via `this.throwIfDisposed()` *before* `this.closeDb(db)` runs. With `closeDbImmediately: true`, `openDb` returns a fresh per-call `IDBPDatabase` that `dispose()` never tracks — the handle is leaked.

`getItemFromCache` is the only public method using inline `closeDb` calls. `put`, `putIf`, `removeEntry`, and `removeEntries` all wrap `closeDb(db)` in a `finally` block (`FluidCache.ts:462-464`, `508-510`, `651-653`, `793-795`), so they are immune to this failure mode. Historical context: #14315 (JasonGore) specifically added `try { … } finally { db?.close(); }` because *"FluidCache return paths can leave DB open, delaying and potentially blocking DB deletions"*; christiango on #15549 cited host signout hangs as the observable symptom. This regression reintroduces that exact failure class.

**Fix**: wrap the body in `try/finally` and move `this.closeDb(db)` into the `finally` clause, mirroring `put`/`putIf`/`removeEntry`/`removeEntries`. This both fixes the leak and brings `getItemFromCache` into structural alignment with the file's own established invariant. A targeted regression test that fires `dispose()` after `await this.openDb()` returns but before `closeDb` runs — asserting the handle is closed even under the dispose race — would lock this down.

anthony-murphy: Addressed in 53a815d.

**Refactored `getItemFromCache` to `try / catch / finally`** so `closeDb(db)` always runs, regardless of how the method exits. Removed the four redundant inline `closeDb(db)` calls in the success returns and the one in the catch — `finally` now owns the cleanup. Under `closeDbImmediately: true` (the default) each `get` opens a fresh per-call connection, and the previous catch-then-`throwIfDisposed`-then-`closeDb` ordering leaked it on every disposed-mid-get race.

**Reordered the catch arms** of `get` / `put` / `putIf` / `removeEntry` / `removeEntries` to call `throwIfDisposed()` *before* `sendErrorEvent`. The error in that branch is the `UsageError` thrown by our own post-await `throwIfDisposed` — not an IDB failure — so logging it under `FluidCache*Error` mis-categorized the dispose race. Real IDB errors still log unchanged (post-`throwIfDisposed` is a no-op when `disposed` is false).

**Regression test** (`get rejected mid-flight closes its IDB connection (no leak)`): seeds a row, then races `dispose()` against five back-to-back `get` calls, then calls `indexedDB.deleteDatabase`. `fake-indexeddb` resolves `deleteDatabase` only when no open connections remain — a leak would fire `blocked` instead of `success`. Test passes in `closeDbImmediately: true` mode and is skipped in shared-connection mode (where `dispose()` itself closes `this.db`).

anthony-murphy:
Round 4 — fixed in 53a815d.

**Tier 2: `getItemFromCache` IDB connection leak on dispose race**
- Refactored `getItemFromCache` to `try / catch / finally`. `closeDb(db)` now lives in `finally` and always runs. Removed four redundant inline `closeDb(db)` calls in the success returns and one in the catch arm.
- Under `closeDbImmediately: true` (the default), every `get` opens a fresh per-call `IDBPDatabase`. The previous ordering (catch → log → `throwIfDisposed` → `closeDb`) short-circuited at `throwIfDisposed`, leaking the connection on every disposed-mid-`get` race.

**Tier 3: catch-arm telemetry mis-categorization**
- All five catch arms (`get` / `put` / `putIf` / `removeEntry` / `removeEntries`) now call `throwIfDisposed()` *before* `sendErrorEvent`. The error in that branch is the `UsageError` our own post-await `throwIfDisposed` threw — not an IDB failure — so logging it under `FluidCache*Error` mis-categorized the dispose race. Real IDB errors still log normally (post-`throwIfDisposed` is a no-op when `disposed` is false).

**New test**
- `get rejected mid-flight closes its IDB connection (no leak)` — seeds a row, races `dispose()` against five back-to-back `get` calls, then calls `indexedDB.deleteDatabase`. `fake-indexeddb` resolves `deleteDatabase` only when no open connections remain (a leak fires `blocked`). Passes in `closeDbImmediately: true`, skipped in shared-connection mode (`dispose()` closes `this.db` there).

89 passing, 1 properly skipped. Build, biome, eslint all clean.

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -303,62 +580,255 @@ export class FluidCache implements IPersistedCache {
...
+	 *
+	 * @remarks
+	 * The implementation uses `transaction.store.get` + `transaction.store.put` rather
+	 * than an IDB cursor. Both run inside the same `readwrite` transaction, so the
+	 * atomicity guarantee is identical, and the get/put pair is materially simpler
+	 * to reason about for a single-key compare-and-swap. A cursor would be the right
+	 * tool if we needed to iterate or range-scan; for a known key we don't.
+	 *
+	 * @param entry - cache entry; identifies the file and the key within that file.
+	 * @param value - the proposed JSON-serializable value to write if `shouldWrite` returns true.
```
anthony-murphy: Deep Review: `putIf`'s `@param value` JSDoc narrows the value contract to "JSON-serializable", but the actual storage path stores it via IDB structured clone — same requirements as `put`.

The JSDoc reads:

```
* @param value - the proposed JSON-serializable value to write if shouldWrite returns true.
```

But `value: unknown` flows through `buildRecord` (FluidCache.ts:712-716, 776-777) into an IDB schema where `cachedObject: any` (`FluidCacheIndexedDb.ts:111-114`). That is structured-clone storage with no JSON constraint, and matches `put` exactly. The README's `putIf` section (README.md:69-92) describes the proposed value with no JSON-serializability restriction, which is consistent with the actual code but contradicts this line.

Historical context: PR #15550 (vladsud, 2023-05-12) explicitly pushed back on the JSON-serializable framing for `FluidCache` values, citing MFS-reported non-serializable content (array buffers, blobs). The same framing is wrong for the same reason here.

```suggestion
     * @param value - the proposed value to write if `shouldWrite` returns true. Stored via IndexedDB structured clone, with the same value requirements as {@link FluidCache.put}.
```

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -303,62 +580,255 @@ export class FluidCache implements IPersistedCache {
...
+	 *
+	 * The existing entry is read and (if `shouldWrite` returns true) the new entry is
+	 * written inside a single IndexedDB `readwrite` transaction. This provides
+	 * compare-and-swap semantics for callers sharing the same underlying IndexedDB
+	 * instance (e.g. multiple browser tabs racing to persist pending state).
+	 *
+	 * @remarks
+	 * The implementation uses `transaction.store.get` + `transaction.store.put` rather
+	 * than an IDB cursor. Both run inside the same `readwrite` transaction, so the
+	 * atomicity guarantee is identical, and the get/put pair is materially simpler
```
anthony-murphy, line 376: Deep Review: `putIf`'s CAS API shape cannot express read-modify-write — flagging as a design discussion before this `@legacy @beta` surface solidifies, not as a blocking concern.

The current signature is `putIf(entry, value, shouldWrite(existing, proposed): boolean): Promise<boolean>` — the proposed value is fixed at call time and the predicate can only accept or reject it. That covers the revision-gating use case in this PR ("active tab wins") but cannot express the canonical CAS shape where the new value is *derived* from the persisted value — e.g. a monotonic sequence-number bump, or merging the existing record into the new one. The two independent blind redesigns run during review both landed on an updater-callback shape (`update(entry, updater)` / `updateUsingPrevious(entry, update(previous): unknown | noUpdate)`) where the callback returns the value to write inside the transaction, and the cross-evaluator scoring put the updater shape at least equal to the predicate on expressiveness (5 vs. 3 on edge cases).

This matters now because the class is `@legacy @beta` and `Class_FluidCache forwardCompat: false` is *already* being taken in `validateDriverWebCachePrevious.generated.ts`. Reshaping after ship costs both backward- and forward-compat; reshaping while the break is already being taken is much cheaper. An additive `update(entry, updater)` overload alongside the existing predicate `putIf` remains forward-compat-compatible as a later addition, so this is not a now-or-never decision — but the next CAS consumer requirement is the natural time to take it, and it is worth surfacing before that consumer needs to work around a missing shape.

**Resolutions either way close this:**
- (a) Add a sibling `update(entry, updater)` overload now while the type-validation break is already being taken; or
- (b) Document in the `putIf` JSDoc that the predicate-only shape is intentional and an updater variant is a planned follow-up if needed.

Flagging **shubhi1092** / **markfields** for the human design call.

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -168,14 +259,132 @@ export class FluidCache implements IPersistedCache {
...
+				subCategory: FluidCacheEventSubCategories.FluidCache,
+				pkgVersion,
+			});
+		} else {
+			this.broadcastChannel = new BroadcastChannel(FluidCacheBroadcastChannelName);
+			// Don't keep the Node event loop alive on account of an idle cache. The web
+			// `BroadcastChannel` interface has no `unref`; Node's implementation does, and
+			// the cast is intentional. In browsers this is a no-op (the property is absent
+			// and the optional-chained call is skipped).
+			(this.broadcastChannel as unknown as { unref?: () => void }).unref?.();
```
anthony-murphy: Deep Review: The `BroadcastChannel.unref()` reliance on an optional chain is not asserted anywhere — if Node ever stops exposing `unref` on `BroadcastChannel`, the call silently no-ops and the package's mocha config tips back into needing `exit: true`.

The call is `(this.broadcastChannel as unknown as { unref?: () => void }).unref?.();` — the optional chain is necessary in browsers, but it also masks regressions on Node. PR #27188 (Josmithr, 2026-04-30) migrated `driver-web-cache` from jest to mocha and explicitly preserved an `exit: true`-free invariant — *"This was needed in order to not specify `exit: true` in the mocha config."* The PR description acknowledges that dependency. The new test file adds 647 lines that construct many `FluidCache` instances under mocha-on-Node; the `afterEach` cleanup at `FluidCache.test.ts:73-88` disposes every cache (closing each `BroadcastChannel` via `dispose`), which mitigates the symptom but does not lock in the contract.

**Resolutions either way close this:**
- (a) Add a Node-only `typeof (channel as { unref?: unknown }).unref === "function"` check around the call that asserts/logs when absent — locks in the contract PR #27188 deliberately preserved; or
- (b) Add a smoke test asserting the mocha process exits cleanly without `exit: true` after constructing and disposing a representative `FluidCache`.

Low-probability gap on this SHA — flagging **Josmithr** (PR #27188 owner) as the minimum reviewer.

anthony-murphy: No longer applicable on `c835c9f`. The events / `BroadcastChannel` surface (and the `unref()` cast that prompted this thread) has been removed from the PR; the change is now scoped to `putIf` only. PR #27188's `exit: true`-free invariant is preserved trivially since no `BroadcastChannel` is ever constructed. Resolving.

— Claude

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -361,4 +361,144 @@ export class FluidCache implements IPersistedCache {
...
+			let valueToWrite: unknown;
+			let setCalled = false;
+			let updaterReturned = false;
+			const set = (value: unknown): void => {
+				if (updaterReturned) {
+					throw new UsageError("FluidCache.update: set called after updater returned");
+				}
+				valueToWrite = value;
+				setCalled = true;
+			};
```
anthony-murphy, line 464: Deep Review: `set`-after-return throws `UsageError` but never logs `FluidCacheUpdateUsageError`.

The `set` closure throws `new UsageError("FluidCache.update: set called after updater returned")` when invoked after `updaterReturned === true`. The PR description frames this guard as *"loud at the misuse site rather than silently lost"*, with the motivating case being a `setTimeout`-deferred `set`.

But by definition that fires *after* `await fluidCache.update(...)` has already resolved — `update`'s own outer `try/catch` (the one that maps to `FluidCachePutError`) has already returned `false` and cannot capture the throw. The exception surfaces as an unhandled rejection on a random event-loop tick with no correlation to the originating call.

The telemetry enum `FluidCacheUpdateUsageError` is defined in `fluidCacheTelemetry.ts:11-18`, but a repository search finds the symbol only in the enum definition — it is dead code. This conflicts with the #15549 norm (vladsud) of capturing misuse/edge states centrally rather than relying on host-side noise.

**Fix**: before the `throw`, call `this.logger.sendErrorEvent({ eventName: FluidCacheErrorEvent.FluidCacheUpdateUsageError, pkgVersion }, error)` (with the constructed `UsageError` as the second arg) so deferred-`set` misuse is recorded centrally regardless of which event-loop tick catches the throw. Add a test that forces a late `set` (e.g. via `setTimeout`) and asserts the event is emitted.

If the unused enum entry is genuinely intended to remain unused, delete it and document the design choice — but the more likely reading is that the `sendErrorEvent` call was lost during a refactor.

anthony-murphy: this is a usage error, and handling is owned by the consumer. we don't need to deeply embed misuse of this api into our logging. it is unlikely any consumer will do this, and this error should be enough for them to realize the pattern isn't supported

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -361,4 +361,144 @@ export class FluidCache implements IPersistedCache {
...
+
+			if (!setCalled) {
+				await transaction.done;
+				return false;
+			}
+
+			const currentTime = Date.now();
+			await transaction.store.put(
+				{
+					cachedObject: valueToWrite,
```
sonalideshpandemsft: nit: if `valueToWrite` is undefined, we can probably call `removeEntry`?

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Deep Review: Acknowledged — consumer-owned-error handling is a reasonable stance for a `@legacy @beta` API, and the `UsageError` thrown at the call site is sufficient signal to the misusing consumer.

Given that decision, the matching cleanup is to remove the `FluidCacheUpdateUsageError` entry from `fluidCacheTelemetry.ts:17`. It is currently declared but never emitted — a grep across `packages/drivers/driver-web-cache` finds only the declaration. Leaving it advertises a telemetry event that will never fire, which can mislead downstream pipelines wiring alarms off the enum. One-line deletion closes the loop on this thread.

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -361,4 +361,144 @@ export class FluidCache implements IPersistedCache {
...
+	 * key, including cross-partition or stale rows that the updater saw as `undefined`.
+	 * This matches the unconditional overwrite behavior of `put`. Callers that must
+	 * preserve cross-partition rows should not use `update`.
+	 *
+	 * Exceptions thrown by `updater` are logged under the dedicated
+	 * `FluidCacheUpdateCallbackError` telemetry event (distinct from IDB write errors)
+	 * and surfaced to the caller as a `false` return value, after aborting the
+	 * transaction so the existing row is preserved — even if `set` was called before
+	 * the throw.
+	 * @returns `true` if `updater` called `set` and the write committed; `false` if
```
anthony-murphy, line 424: Deep Review: `Promise<boolean>` collapses three distinct outcomes — committed-write, lost-race (updater returned without `set`), updater-threw, and IDB-error — into one bit.

The TSDoc at lines 407-409 covers this in a parenthetical: *"`true` if `updater` called `set` and the write committed; `false` if `updater` returned without calling `set`, threw, or an IDB error occurred."* The body confirms three `return false` paths: updater threw (line 470), updater returned without `set` (line 476), outer catch on IDB error (line 499). Telemetry distinguishes the latter two (`FluidCacheUpdateCallbackError` vs `FluidCachePutError`) but the boolean return cannot.

A compare-and-set caller ("did my write win the race?") cannot distinguish lost-race from IDB-error from updater-throw via the return value alone — they must cross-reference telemetry events on a per-call basis, which is awkward at best and impossible in non-telemetry-equipped environments. This is consistent with `put`'s log-don't-throw convention (`FluidCache.ts:332-362`) so the contract is fine; only the documentation could be louder.

**Minimal fix**: promote the caveat out of the parenthetical into a named *Compare-and-set callers* note that explicitly states callers cannot distinguish lost-race from IDB-error from updater-throw via the boolean alone, and must consult `FluidCacheUpdateCallbackError` / `FluidCachePutError` telemetry to disambiguate. Mirror to README at 108-109.

**More decisive fix while still `@beta`**: replace the return with a discriminated union — `{kind:"committed"|"noop"|"callbackError"|"idbError"}` — so the disambiguation is part of the API contract rather than a side channel. See the companion thread on the callback shape for a unified refactor that folds this in.

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -361,4 +361,144 @@ export class FluidCache implements IPersistedCache {
...
+	 * `updater` returned without calling `set`, threw, or an IDB error occurred. IDB
+	 * errors are logged and not thrown, matching the behavior of `put`.
+	 */
+	public async update(
```
anthony-murphy, line 428: Deep Review: The callback-with-`set` shape carries a runtime-guard tax that a return-value shape would eliminate.

The shipped signature is `update(entry, updater: (existing, set: (v) => void) => void): Promise<boolean>`. The `set` closure (lines 438-447) captures three locals (`valueToWrite`, `setCalled`, `updaterReturned`) and throws `UsageError` if `set` is invoked after the updater returns (lines 442-444). A dedicated test exists solely to pin down that guard.

A return-value shape — `updater: (existing) => {kind:"write"; value: unknown} | {kind:"unchanged"} | {kind:"delete"}` — would:

- eliminate the three captured locals and the `setCalled`/`updaterReturned` flags;
- remove the runtime `UsageError` guard by making deferred-`set` misuse *structurally impossible* (no callable is handed out);
- make `set(undefined)`-as-write unambiguous via the explicit `kind` (a plain `T | undefined` sentinel cannot — that legitimate counter does not apply to the tagged-union form);
- fold atomic-delete (the open question on sonalideshpandemsft's thread at line 482) into the same primitive at zero marginal cost, instead of requiring a separate non-atomic `removeEntry` call;
- compose with the discriminated-union return suggested on the sibling thread at line 407, so all four outcomes are part of the contract rather than split between return value and telemetry.

The current design is defensible, and `set(undefined)` distinguishability is real for the *sentinel* form earlier proposals suggested. But it does not refute a tagged-union form, and the runtime guard + test exist only because the callable form leaves the misuse path expressible.

**Recommendation**: while the API is still `@legacy @beta`, decide explicitly — either replace the signature with the discriminated-union return (which also closes the sibling `Promise<boolean>` thread and the atomic-delete scope question in one move), or close as "considered and declined" with a one-line TSDoc rationale. Either way, decide before `@beta` graduation.

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Done in `f9d01c364b` (now `935554d03f` after rebase on the merged main). `update` now treats `set(undefined)` as an atomic delete: same `readwrite` transaction switches to `transaction.store.delete(key)` instead of `put`. Rationale matches your nit — `get` already collapses "no entry" and "entry stored as undefined" into the same observable result, so callers get an atomic conditional-delete for free with no meaningful ambiguity. JSDoc/README/changeset updated, three tests added (existing-row delete verified by raw IDB read, absent-key no-op, cross-partition delete). 60 tests passing.

— Claude

anthony-murphy:
## Deep Review

*Reviewed commit `935554d` on 2026-05-26.*

A new Tier 2 correctness gap surfaced on this commit: the `update(entry, updater)` callback is typed `(...) => void`, which TypeScript structurally accepts `async` functions for. The implementation never inspects the return value for thenable-ness, so an async updater can call `set(value)` synchronously, return a pending Promise, have the write commit, and then reject unhandled — violating the PR's documented contract that *"if the updater throws, including after calling `set`, the transaction is aborted."* Fix is mechanical (thenable guard + abort + telemetry) and no async-updater test exists. Three prior design threads (return-shape, `IPersistedCache` placement, partition-mismatch telemetry parity) are held until this closes.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Decide the S1 async-updater mitigation: either runtime thenable-detection on the updater return value (`transaction.abort()` + `FluidCacheUpdateCallbackError` log + `return false`, or throw `UsageError` synchronously to mirror the existing `updaterReturned` guard), or adopt a structural API change (e.g. the discriminated-return reshape on thread 3284824393/3284824477) that makes the misuse unreachable
- [ ] Add a test covering `async updater calls set then awaits then throws` — assert (i) existing row preserved, (ii) `update` returns `false`, (iii) `FluidCacheUpdateCallbackError` fires, (iv) no unhandled rejection escapes
- [ ] Acknowledge the chosen async-callback mitigation in the PR description's *things to look for* section so future reviewers can verify it

### Context for Reviewers

- Atomicity hinges on the delayed-close `openDb()` pattern from #15549 (jatgarg). The `no-await between get and put` argument holds *only because* `openDb()` resolves synchronously to the same `IDBPDatabase` for the transaction's duration; if `versionchange`/`close` fires between `openDb()` and `db.transaction(...)`, the call throws and is swallowed as `FluidCachePutError`, returning `false`.
- DB-close discipline established in #14315: `try { ... } finally { closeDb(db); }` on every method. `update()` follows it, including the updater-throw branch which uses `await transaction.done.catch(() => {})` before returning so `finally` always runs.
- `forwardCompat: false` on the `Class_FluidCache` type test follows the precedent #26065 set for `removeEntry` (shubhi1092). Additive method on a `@legacy @beta` class is a forward-compat type-break by definition; consumers typed only as `IPersistedCache` will not see `update`.
- API stays at `@legacy @beta` (not `@alpha`) per PR #25256 (chentong7) — `@beta` symbols cannot reference `@alpha` symbols, so `update` lands in `driver-web-cache.legacy.beta.api.md`.
- New 224-line test block (concurrent-update, transaction-abort) pushes extra `FluidCache` instances onto `extraCaches` and drains them in `afterEach` — the fake-indexeddb v3 leak-avoidance pattern Josmithr established in #27188.
- sonalideshpandemsft's nit on `set(undefined)` → atomic-delete was addressed in `f9d01c3`/`935554d`.
- Suggested reviewers: **jatgarg** (openDb/delayed-close, #15549) and **vladsud** for atomicity / abort-on-shared-connection; **shubhi1092** / **markfields** for the API-shape decision driven by S1; **JasonGore** for DB-lifetime sign-off on `update`'s exit paths; **Josmithr** / **alexvy86** for test cleanup hygiene; **chentong7** for API report / type-validation surface; **christiango** for DB-handle leakage history.

- **API shape decision (still owner-call)** — `Promise<boolean>` vs discriminated-union return, and whether the callback-with-`set` should become a return-value form. Both are already filed as inline threads (3284824393, 3284824477); adopting the tagged-union return form would structurally eliminate the S1 async-callback gap as a side effect. Best for **shubhi1092** / **markfields**.
- **Interface placement on `IPersistedCache`** — two independent blind redesigns added `update?` to the shared interface citing the `removeEntry?` precedent from #26065. Defensible scope choice to defer; cheap to extend additively while `@beta`. Best for **jatgarg** / **vladsud** / **christiango**.
- **Atomic-delete as a first-class outcome** — sonalideshpandemsft's open thread on `FluidCache.ts:482` independently surfaces the same gap the blind redesigns folded into a tagged-union primitive. Overlaps the discriminated-return reshape above. Best for **jatgarg** / **shubhi1092**.
- **DB-lifetime scrutiny on `update`'s exit paths** — confirm transaction settlement + abort path + `closeDb(db)` ordering on every `update` exit, including the `versionchange`/close mid-transaction → `false` outcome. The existing test matrix covers the abort-on-keep-alive case; an area expert should still sign off on the `versionchange` interpretation. Best for **JasonGore** or **vladsud**.
- **Multi-`set` last-wins semantics** — PR ships and tests last-wins; one blind redesign argued a second `set` should throw `UsageError` to match the loud-at-misuse posture. Worth surfacing for **shubhi1092** / **markfields** while `@beta`.
- **Throw-after-`set` policy** — PR aborts and preserves the existing row when the updater throws after `set`. PR description explicitly asks reviewers to confirm this matches host expectations.

</details>

- `824e638` 2026-05-26 · **7/10** — primitive mechanically sound; surviving items polish + scope/owner calls
- `1f3eb05` 2026-05-26 · **5/10** — telemetry-hygiene gap flagged inline; primitive otherwise mechanically sound

</details>

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -361,4 +361,159 @@ export class FluidCache implements IPersistedCache {
...
+
+			// Invoke the updater in its own try/catch so a host-supplied callback
+			// throwing does not get logged under `FluidCachePutError` (which is for
+			// IDB-write failures). On updater throw we abort the transaction so the
+			// existing row is preserved — even if `set` was called before the throw —
+			// log under the updater-specific event, and return `false` (matching the
+			// documented "errors are logged, not thrown" contract).
+			try {
+				updater(existingValue, set);
+			} catch (updaterError: any) {
```
anthony-murphy, line 487: Deep Review: An async updater can commit a write and then reject unhandled, contradicting the documented abort-on-throw contract.

The public signature is `updater: (existing: unknown, set: (value: unknown) => void) => void` (lines 417-420). TypeScript structurally accepts `async (...) => Promise<void>` for that type. The implementation invokes `updater(existingValue, set);` inside a synchronous `try`/`catch` and immediately flips `updaterReturned = true` — it never inspects the return value for thenable-ness.

An async updater that calls `set(value)` synchronously, then `await`s and throws, will:

1. synchronously populate `valueToWrite`/`setCalled`,
2. return a still-pending Promise,
3. be treated as a successful synchronous return by the `try` block,
4. reach the `setCalled` branch and `await transaction.store.put(...)`,
5. commit the write and `return true`.

The eventual rejection is unhandled and never reaches the inner `catch (updaterError)` that the JSDoc (lines 397-412) and the changeset both claim aborts the transaction on *"updater throws — including after calling `set`."* No test in `FluidCache.test.ts:316-596` exercises an async-updater path. The PR description explicitly invites this class of finding: *"If you see a way that any await could sneak in there ... please flag it."*

**Recommended fix**: capture the return value of `updater(existingValue, set)` and, if it is thenable, treat it as misuse symmetric with the existing `updaterReturned` guard — `transaction.abort()`, log under a `FluidCacheUpdateCallbackError` event distinguishing async-callback misuse from a thrown error, and return `false` (or throw `UsageError` synchronously). Either way, add a test for `async updater that calls set, awaits, then throws` asserting (i) the existing row is preserved, (ii) `update` returns `false`, (iii) telemetry fires, (iv) no unhandled rejection escapes.

A structural alternative is the discriminated-return reshape already raised on threads 3284824393 / 3284824477 — replacing the `(existing, set) => void` callable with `(existing) => {kind: "write"; value} | {kind: "unchanged"} | {kind: "delete"}` makes the async-callback misuse unreachable by construction, since no callable is handed out and the result type is not a Promise.

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Declining the reshape, keeping the callback form. Reasoning:

**The structural-misuse argument is the strongest part, and the T3 fix on `6e80dc24a8` neutralizes it.** Deferred-`set` was already loud (the `updaterReturned` guard throws `UsageError` at the misuse site); async-updater misuse is now also loud (`isPromiseLike` check after `updater` returns aborts the tx, logs `FluidCacheUpdateCallbackError`, returns false). Two tests cover the async-misuse paths. The discriminated union would make these *structurally impossible* rather than *runtime-loud* — strictly stronger, but the runtime guards are now non-leaky in observable behavior.

**Call-site ergonomics regress materially.** The two motivating use cases:

```ts
// callback form (shipped)
await cache.update(entry, (existing, set) => {
  if (mine.rev > (existing?.rev ?? -1)) set(mine);
});
await cache.update(entry, (existing, set) => {
  const prev = (existing as { count?: number } | undefined)?.count ?? 0;
  set({ count: prev + 1 });
});

// discriminated-union form
await cache.update(entry, (existing) => {
  if (mine.rev > (existing?.rev ?? -1)) return { kind: "write", value: mine };
  return { kind: "unchanged" };
});
await cache.update(entry, (existing) => {
  const prev = (existing as { count?: number } | undefined)?.count ?? 0;
  return { kind: "write", value: { count: prev + 1 } };
});
```

Every conditional now has to either name a kind or return an `unchanged` shape — even the "do nothing" path becomes a typed return. The callback form's implicit "return without calling set ⇒ unchanged" is genuinely lighter.

**Atomic-delete fits the callback form cleanly.** `set(undefined)` already means delete (landed in `f9d01c364b` on sonalideshpandemsft's thread, and verified by raw-IDB-read in tests). No marginal cost vs. the discriminated union; the `get`-collapses-undefined argument applies equally in both shapes.

**Three captured locals is not load-bearing complexity.** `valueToWrite` / `setCalled` / `updaterReturned` are all tied to a single lexical scope and dead the moment `update` returns. They aren't a maintenance hazard.

**Forward-compat note.** If a future caller genuinely needs to distinguish all four outcomes synchronously without consulting telemetry — the implicit ask behind the sibling T1 thread — the right move is a `Promise<UpdateResult>` discriminated-union *return type* on top of the callback shape, not a reshape of the updater itself. That refactor remains additive whether we ship the callback form or the union form today.

Resolving as not-planned-on-this-PR.

— Claude

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Took the doc fix on `6e80dc24a8`. JSDoc on `update` and README now carry a "Compare-and-set callers" note explicitly calling out that `false` collapses (a) lost race (updater returned without `set`), (b) updater-throw — now including the async-updater misuse case from sibling thread T3 — and (c) IDB-write failure. Telemetry mapping is spelled out: updater-side failures log `FluidCacheUpdateCallbackError`; IDB-write failures log `FluidCachePutError`; lost races are not logged. Promoted out of the parenthetical as you suggested.

Going with the doc-only fix rather than the discriminated-union return reshape. If a future caller actually needs to distinguish all four outcomes synchronously without consulting telemetry, the additive move is a `Promise<UpdateResult>` return type on top of the existing callback shape — it stays forward-compat-compatible from here. The full update-shape redesign is the discussion on sibling T2, which I've replied to and resolved.

— Claude

packages/drivers/driver-web-cache/src/FluidCache.ts
anthony-murphy: Fixed on `6e80dc24a8`. Took the recommended approach: capture the updater's return value and, if `isPromiseLike` (the `@fluidframework/core-utils` helper — thanks for the pointer), abort the transaction and treat it symmetrically with the existing `updaterReturned` guard. Specifically:

```ts
const updaterResult = updater(existingValue, set);
updaterReturned = true;
if (isPromiseLike(updaterResult)) {
    throw new UsageError(
        "FluidCache.update: updater must be synchronous (returned a thenable)",
    );
}
```

The throw lands in the existing updater try/catch, which aborts the transaction (preserving the existing row even if `set` was called sync before the throw), logs under `FluidCacheUpdateCallbackError`, and returns `false`. The pending Promise the updater returned is orphaned by design — the updater contract is sync.

Two tests added covering the cases from your write-up:
- `rejects an async updater that calls set then awaits and throws` — verifies the documented commit-then-reject race no longer commits; existing row preserved
- `rejects an async updater that returns a promise without throwing` — verifies a non-throwing async updater is still misuse (the structural intent is "do more work after returning," which we can't honor)

JSDoc and README both updated to call out the second guard explicitly alongside the existing late-`set` guard. The new T1 "Compare-and-set callers" doc note (sibling thread) folds the async-updater misuse into the broader `false`-collapses-N-outcomes story.

— Claude

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -361,4 +361,182 @@ export class FluidCache implements IPersistedCache {
...
+				// type, but an async updater that calls `set` synchronously and then
+				// awaits would let the IDB write commit before its eventual rejection
+				// surfaced — contradicting the "throw aborts the transaction" contract.
+				// Detect a thenable return and treat it as misuse symmetric with the
+				// late-`set` guard.
+				if (isPromiseLike(updaterResult)) {
+					throw new UsageError(
+						"FluidCache.update: updater must be synchronous (returned a thenable)",
+					);
+				}
```
anthony-murphy, line 486: Deep Review: The new `isPromiseLike` guard closes the structural half of the prior async-updater concern (commit-then-reject can no longer slip through to a successful write), but the detected thenable is never attached a rejection handler.

At line 473 the implementation does `const updaterResult = updater(existingValue, set);`. If the host passes an `async` updater (structurally accepted because the declared parameter type is `(...) => void`), `updaterResult` is a `Promise`. Lines 482-486 detect it and throw `UsageError` synchronously — the inner `catch` at 487-501 then aborts the transaction, swallows `transaction.done`, logs `FluidCacheUpdateCallbackError`, and returns `false`. Good.

But the `updaterResult` promise itself is orphaned. The new test `"rejects an async updater that calls set then awaits and throws"` (`FluidCache.test.ts:483-505`) exercises exactly the path where that orphaned promise rejects *after* `update` has returned — the in-test comment even calls it out as "orphaned by design." Node ≥ 15 default behavior is to terminate the process on unhandled rejections; FluidFramework's mocha config and typical production telemetry hosts install strict `unhandledRejection` handlers. A documented benign misuse path that can crash the host isn't the loud-at-misuse-site posture this guard was added for.

Fix is a one-liner: attach a no-op handler to the detected thenable before throwing. Converts the misuse from a process-level crash to the intended synchronous `UsageError` + `FluidCacheUpdateCallbackError` telemetry, with the existing test row preservation. Worth extending the test to assert no unhandled rejection escapes (e.g. via a `process.on("unhandledRejection")` listener bracketing the call).

The structural alternative — replacing `(existing, set) => void` with a discriminated-union return shape (open design thread 3307031902) — would make this unreachable by construction, but is out of scope for the Tier 2 fix.

```suggestion
            if (isPromiseLike(updaterResult)) {
                void Promise.resolve(updaterResult).catch(() => {});
                throw new UsageError(
                    "FluidCache.update: updater must be synchronous (returned a thenable)",
                );
            }
```

anthony-murphy: won't fix. again, this is trying to handle api misue. if the use passes a promise that is an error is their code they should fix, we don't need to handled the case completely, throwing an error is enough convivence .

anthony-murphy:

- ✅ Mechanical Review
- 👻 **Solution / Problem Space Review** — tracing ghost currents…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -361,4 +361,182 @@ export class FluidCache implements IPersistedCache {
...
+	 * logged under `FluidCachePutError`. A lost compare-and-set race (the updater
+	 * returned without calling `set`) is not logged.
+	 * @returns `true` if `updater` called `set` and the write committed; `false` if
+	 * `updater` returned without calling `set`, threw, or an IDB error occurred. IDB
+	 * errors are logged and not thrown, matching the behavior of `put`.
+	 */
+	public async update(
+		entry: ICacheEntry,
+		updater: (existing: unknown, set: (value: unknown) => void) => void,
+	): Promise<boolean> {
```
anthony-murphy, line 431: Deep Review: The atomic `update` primitive lives only on the concrete `FluidCache` class — `IPersistedCache` (the interface most consumers are typed against) is untouched.

This PR adds `update(...)` to the `FluidCache` class and flips `Class_FluidCache: {forwardCompat: false}` in `validateDriverWebCachePrevious.generated.ts` to acknowledge the type-break. The canonical `IPersistedCache` in `packages/common/driver-definitions/src/cacheDefinitions.ts` and its api-report are not touched. The motivating consumer named in `.changeset/optimistic-driver-web-cache.md` and the README — cross-tab pending-state coordination in drivers such as `odsp-driver` — is typed against `IPersistedCache`, so it cannot reach `update` without a cast or a class-level feature-detect.

The `removeEntry?` precedent from #26065 (shubhi1092) shows the additive-optional-member shape works for exactly this case: add the new capability as an optional member on the shared interface, ship the implementation only on `FluidCache`, and consumers feature-detect via `cache.update?.(…)`. That route also sidesteps the class-level `forwardCompat: false` break this PR currently accepts, because adding an optional member to an interface is backward-compat for implementers and forward-compat for callers.

**Recommendation**: decide explicitly while the API is still `@legacy @beta`, either —

- (a) add `update?(entry, updater): Promise<boolean>` as an optional member of `IPersistedCache` in `packages/common/driver-definitions/src/cacheDefinitions.ts` (and the matching api-report) in this PR. Keeps the implementation on `FluidCache` exactly as written; source-compatible for existing external implementers; avoids the class-level `forwardCompat: false` flip. Materially cheaper than retrofitting it once `odsp-driver` has consumed the primitive against the concrete class.
- (b) record a one-line rationale in the changeset and/or the `update` JSDoc that the class-only placement is intentional and the interface addition is deferred to the consumer PR, so reviewers and downstream consumers see the decision in writing.

Flagging while the surface is still `@beta` and the consumer hasn't landed yet.

anthony-murphy:
## Deep Review

*Reviewed commit `47f975b` on 2026-05-26.*

The `update(entry, updater)` primitive is mechanically sound: the get→updater→put sequence runs synchronously inside one IDB `readwrite` transaction, the `try { … } finally { closeDb(db) }` invariant from #14315 is preserved on every exit, and the new tests exercise the documented misuse paths. One new Tier 3 polish item is flagged inline (telemetry-label gap on `update`'s outer catch); a prior design-shape thread on `Promise<boolean>` outcome conflation remains open as the only thing pinning the score at the bottom of this band.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Request at least one reviewer from the suggested list — the `Requested Reviewers` field is still empty.

### Context for Reviewers

- `forwardCompat: false` / optional-on-interface / required-on-class follows the `removeEntry?` precedent from #26065 (shubhi1092): additive on `IPersistedCache`, non-optional on the `FluidCache` class, `Class_FluidCache: { forwardCompat: false }` in `package.json` `typeValidation.broken`.
- DB-close discipline from #14315 (JasonGore): `try { … } finally { closeDb(db) }` on every public method. `update()` follows it on every exit — success, no-op, updater-throw, `transaction.abort()`, and IDB-error.
- Atomicity-argument scrutiny is the established review culture for this file (#15549 thread between christiango/vladsud/JasonGore). The PR description's "Reviewer Guidance" explicitly invites line-by-line scrutiny of the `get` → updater → `set` → `put` sequence and the `isPromiseLike` / `updaterReturned` guards.
- The test file was just migrated from jest to mocha in #27188 (Josmithr). New tests use `node:assert`, shared `fluidCache` + `afterEach` cleanup, and explicit `db.close()` after the one raw `openDB` call to avoid fake-indexeddb v3 leak/deadlock hazards.
- Suggested reviewers: atomicity / lifecycle — **jatgarg**, **JasonGore**, **vladsud**, **christiango**; test wiring — **Josmithr**, **alexvy86**; API-shape precedent — **shubhi1092**, **markfields**.

- **Async-updater unhandled-rejection policy** (thread `3307434191`) — Author declined the one-line `void Promise.resolve(updaterResult).catch(() => {})` with "throwing an error is enough convenience." Two independent reviewers argued the fix is worth taking because the failure mode (global `unhandledrejection`, Node `--unhandled-rejections=strict` termination) escapes the caller's frame entirely. Worth a ratify-or-overturn call.
- **`Promise<boolean>` outcome-conflation** (thread `3307031797`) — Author took the doc-only fix; design-shape question (`{ committed; reason }` discriminated return) deferred. While `@beta`, worth one explicit owner decision on whether the cross-tab compare-and-set motivation in the JSDoc would be better served by a tagged-union return. Best for **shubhi1092** / **markfields** / **jatgarg**.
- **Interface placement on `IPersistedCache`** (thread `3307615764`) — `update` lives only on the concrete `FluidCache` class; the `removeEntry?` precedent from #26065 argues for `update?` on `IPersistedCache` so cross-tab consumers (`odsp-driver`) can feature-detect without casts. Defensible to defer; cheap to extend additively while `@beta`. Best for **shubhi1092** / **markfields** / **jatgarg**.
- **DB-lifetime sign-off on `update`'s exit paths** — confirm `transaction.done`, `transaction.abort()`, and `closeDb(db)` ordering on every exit. Best for **JasonGore** / **vladsud**.
- **Cross-tab atomicity on the shared cached connection** — walk the `versionchange` / close-mid-transaction interpretation under #15549's delayed-close `openDb()` pattern. Best for **jatgarg** / **vladsud**.
- **Throw-after-`set` policy** — PR aborts and preserves the existing row when the updater throws after `set`; explicitly flagged in the PR description for confirmation against host expectations.
- **Multi-`set` last-wins semantics** — PR ships last-wins; one blind redesign argued the second `set` should throw `UsageError` to match the loud-at-misuse posture elsewhere. Worth one explicit decision while `@beta`. Best for **shubhi1092** / **markfields**.

</details>

- `7cba23e` 2026-05-27 · **6/10** — primitive sound; interface-placement design question surfaced
- `6e80dc2` 2026-05-26 · **5/10** — async-updater unhandled-rejection Tier 2 surfaced (later author-declined)
- `935554d` 2026-05-26 · **5/10** — async-updater Tier 2 surfaced; three design threads held
- `824e638` 2026-05-26 · **7/10** — primitive mechanically sound; surviving items polish + scope/owner calls
- `1f3eb05` 2026-05-26 · **5/10** — telemetry-hygiene gap flagged inline; primitive otherwise mechanically sound

</details>

anthony-murphy:
Deep Review: Acknowledged — routing this to human-reviewer ratification rather than re-anchoring the readiness score on a disputed item. Surfaced under **For human reviewer** in the current summary so a human can ratify or overturn the policy call. No further automated nudge planned absent new evidence.

anthony-murphy:
Deep Review: The doc-only fix on `6e80dc24a8` (promoted "Compare-and-set callers" note in JSDoc + README, telemetry mapping spelled out) addresses the documentation half cleanly. The design-shape question — whether `Promise<{ committed; reason }>` discriminated return would better serve the cross-tab compare-and-set use case — is held open under **For human reviewer** in the current summary for **shubhi1092** / **markfields** / **jatgarg** to weigh while `@beta`. A one-line "considered and declined; matches `put` semantics" closure would move this to fully resolved.

packages/drivers/driver-web-cache/src/FluidCache.ts
```diff
@@ -361,4 +361,182 @@ export class FluidCache implements IPersistedCache {
...
+				);
+			}
+			await transaction.done;
+			return true;
+		} catch (error: any) {
+			this.logger.sendErrorEvent(
+				{ eventName: FluidCacheErrorEvent.FluidCachePutError, pkgVersion },
+				error,
+			);
+			return false;
```
anthony-murphy, line 537: Deep Review: The outer `catch` at lines 532-537 unconditionally emits `FluidCacheErrorEvent.FluidCachePutError`, but this catch covers (a) `openDb()` failure at line 434, (b) `transaction.store.get(key)` failure at line 438, and (c) `transaction.done` rejection at line 504 (the no-`set`-called path that still awaits transaction settlement) — none of which are puts. The JSDoc at lines 420-422 also tells telemetry users "IDB-write failures are logged under `FluidCachePutError`", which understates what the code actually does.

This PR already split out `FluidCacheUpdateCallbackError` precisely because commingling distinct error classes harms triage; the same reasoning applies symmetrically to infra failures from `update` vs `put`. An operator chasing a spike in `FluidCachePutError` today cannot tell whether it's a real `put()` site, an `openDb` regression hit through `update`, or a `transaction.done` rejection on the no-op path.

**Fix** (pick one):

1. Tag the existing `sendErrorEvent` call with an `op: "update"` property so operators can distinguish `put()`-origin from `update()`-origin storage failures during triage.
2. Add a dedicated `FluidCacheUpdateStorageError` sibling event and emit it from the outer catch in `update`.

Either way, update the JSDoc at lines 420-422 to match whichever path is chosen.

sonalideshpandemsft (approved):

jatgarg (approved):

PR #27353, opened 2026-05-20
[release/client/2.100] fix(tree): Add option to heal unresolvable identifier IDs at doc load time

Abe27342:
Cherry-pick of #27281 into 2.100

## Description

Behavior:
- `chunkDecoding.readValue` (the `SpecialField.Identifier` path) now wraps the `normalizeToSessionSpace` + `decompress` call in try/catch. On failure, when `healUnresolvableIdsOnDecode` is enabled and a `sharedObjectId` is available, it returns a deterministic stable UUID computed as `uuidv5("${sharedObjectId}|${opSpaceId}", healingNamespace)`. All clients loading the same blob agree on the resulting value, and the existing chunked-forest encode path (also from d43d50d7563) writes the string back out at the next summary so the healing does not need to be re-applied on subsequent loads.

Plumbing:
- `FieldBatchEncodingContext` and `IdDecodingContext` gain `healUnresolvableIdsOnDecode?: boolean` and `sharedObjectId?: string`.
- `ChangeEncodingContext` and `EditManagerEncodingContext` carry the same fields so the flag reaches chunked-forest decode for builds/refreshers inside trunk-commit changes (via `modularChangeCodecV1.decodeDetachedNodes`) in addition to `ForestSummarizer.load`.
- `EditManagerSummarizer` accepts the flag and shared-object id as constructor args.
- `SharedTreeOptionsBeta` exposes `healUnresolvableIdsOnDecode?: boolean` (default `false`) via `configuredSharedTreeBetaLegacy`. Off by default so it does not mask genuine decode failures in documents that aren't actually corrupt.

The `RevisionTagCodec` and `DetachedFieldIndex` v1/v2 codecs are left unchanged: v2 already emits long IDs when summarizing in detached state, so those codecs do not produce unresolvable persisted IDs to heal.

Adds an end-to-end test that reproduces the
SharedTree-attaches-to-attached- container scenario and verifies healed loading succeeds when the option is enabled.

## Breaking Changes

N/A -- all behavioral changes are opt-in

### DIFF
--- a/.changeset/petite-pillows-wave.md
+++ b/.changeset/petite-pillows-wave.md
@@ -1,0 +1,17 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": fix
+---
+Add SharedTreeOptionsBeta.healUnresolvableIdentifiersOnDecode to recover documents with corrupted identifiers
+
+A SharedTree bug can result in corrupted documents due to their attach summary compressing identifier-field values in a way that cannot be uncompressed.
+This bug manifested as remote clients processing the op throwing an error with the message "Unknown op space ID.".
+
+This change adds an option (`healUnresolvableIdentifiersOnDecode`) to `configuredSharedTreeBetaLegacy` which will allow documents affected by this bug to load again when enabled.
+Enabling this option carries some risk, see documentation on the interface itself for more details.
+
+#### Who is affected
+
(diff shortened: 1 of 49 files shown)


markfields (approved):

frankmueller-msft (approved):

alexvy86 (approved):
Approving for docs

noencke (approved):

PR #26295, opened 2026-01-27
Duplicate batch detector optimization and BatchId benchmark tests.

dannimad:
This PR adds benchmarking coverage for batch ID tracking and optimizes the duplicate batch detection logic to use a bidirectional mapping between batch IDs and sequence numbers.

Changes:

- Extend the op critical path benchmark to run both with and without Fluid.ContainerRuntime.enableBatchIdTracking enabled, by parameterizing container setup and benchmark titles over a small set of configs.
- Refactor DuplicateBatchDetector to replace the batchIdsAll set with a seqNumByBatchId map, enabling O(1) duplicate detection.
- Add more benchmark testing to have a baseline on enabling batch id overhead. By the benchmark results on `batchIdTracking.time.spec.ts` and analyzing DuplicateBatchDetector logic, there should be no significant impact enabling it.

PR #22501, opened 2024-09-13
ContaineRuntime: Refactor batch processing code to support either op-by-op or batch-all-at-once semantics

markfields:
## Description

Step one of [AB#15221](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/15221)

No behavior change here, just refactoring to be able to express the result of RemoteMessageProcessor in terms of ops or batches.

Follow-up change to adjust the logic will be much more scoped then.

## Reviewer Guidance

No behavior change.  Notice that even in test files, the change is strictly refactoring, test results/asserts are unchanged.

PR #27355, opened 2026-05-20
feat(container-runtime): enhance DuplicateBatch telemetry diagnostics

anthony-murphy:
## Description

The `DuplicateBatch` `dataCorruptionError` event recorded the new (duplicate) batch's identity but not the original's, and its `messageClientSequenceNumber` was the synthetic counter produced by `OpGroupingManager.ungroupOp` rather than the real outer-envelope csn. Both shortcomings made it impossible to tell from telemetry whether a production duplicate was caused by a service-side double-sequence, a client-side resubmit collision, or another path.

### Changes

- **`opLifecycle/duplicateBatchDetector.ts`** — Exported `RecordedBatchInfo` (`clientId`, `batchStartCsn`, `batchIdExplicit`); renamed `batchIdsBySeqNum` → `batchesBySeqNum` (now `Map<number, { batchId, info? }>`); `processInboundBatch` returns the original's `RecordedBatchInfo` (or `undefined` when the original came from a snapshot) on duplicate match. Snapshot wire format `[[sequenceNumber, batchId]]` preserved for backward compatibility.
- **`containerRuntime.ts`** — `DuplicateBatch` event now logs `batchIdExplicit`, `otherClientId`, `otherBatchStartCsn`, `otherBatchIdExplicit`, `otherFromSnapshot`, and overrides `messageClientSequenceNumber` with the real outer-envelope `batchStart.batchStartCsn` (with a code comment explaining the `ungroupOp` fake-csn artifact).
- **Tests** — All existing assertions updated; 2 new tests cover the runtime-info path and explicit-vs-synthesized batchId detection. 11/11 unit + 3/3 integration tests pass.

### Motivation — what production telemetry will now reveal

Across 50 production events over 7 days, every `DuplicateBatch` showed `DupClientId == MsgClientId`, `BatchStartCsn == 1`, `MessageClientSequenceNumber == 1`, and an empty `details.batchId` — which leaves multiple very different root-cause hypotheses (service-side double-sequence vs. client-side resubmit collision vs. another path) indistinguishable. With the new fields, the next event will tell us which it is:

| Field combination on next event                                                       | Likely root cause                                |
| ------------------------------------------------------------------------------------- | ------------------------------------------------ |
| `batchIdExplicit == false` **and** `otherBatchIdExplicit == false`, both csn equal    | Service-side double-sequence — escalate to ODSP. |
| Exactly one of the two `*Explicit` flags is `true`                                    | Client-side resubmit colliding with fresh submit. |

## Reviewer Guidance

- **`DuplicateBatchDetector` is internal** (exported from `opLifecycle/index.ts` but not in any `api-report.md`). No public API surface change, no changeset required.
- **Snapshot wire format unchanged**: `getRecentBatchInfoForSummary` still emits the same `[number, string][]` tuples; a round-trip integration test exercises this.
- **Worth a careful look**: the `messageClientSequenceNumber` override is correctness-tied — please confirm `batchStart.batchStartCsn` is the right substitute (it's set from the outer-envelope `clientSequenceNumber` before `ungroupOp` runs; see `remoteMessageProcessor.ts`).
- **Out of scope, but related**: the same fake-csn artifact exists in other call sites that consume grouped sub-messages via `extractSafePropertiesFromMessage` (`channelCollection.ts:480/526/578`, `dataStoreContext.ts:761`, `pendingStateManager.ts:729`, `inboundBatchAggregator.ts:183`). Deliberately left for a follow-up — flagging for awareness.

Tagged `deep-review` because the fix is bound up with a live production data-corruption investigation; we want to be sure the new telemetry is correct before the next event lands.

### DIFF
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -3141,11 +3141,26 @@ export class ContainerRuntime
 							eventName: "DuplicateBatch",
 							details: {
 								batchId: batchStart.batchId,
+								batchIdExplicit: batchStart.batchId !== undefined,
 								clientId: batchStart.clientId,
 								batchStartCsn: batchStart.batchStartCsn,
 								size: inboundResult.length,
 								duplicateBatchSequenceNumber: result.otherSequenceNumber,
+								// Identifying info for the ORIGINAL occurrence of this batch, so we can
+								// disambiguate the duplicate's source (e.g. resubmit vs fresh submit, same
+								// vs different wire clientId). Undefined fields indicate the original was
+								// loaded from a summary snapshot rather than seen at runtime.
+								otherClientId: result.otherBatchInfo?.clientId,
+								otherBatchStartCsn: result.otherBatchInfo?.batchStartCsn,
+								otherBatchIdExplicit: result.otherBatchInfo?.batchIdExplicit,
+								otherFromSnapshot: result.otherBatchInfo === undefined,
 								...extractSafePropertiesFromMessage(batchStart.keyMessage),
+								// For grouped batches, `keyMessage` is one of the sub-messages produced by
+								// `OpGroupingManager.ungroupOp`, which overwrites `clientSequenceNumber`
+								// with a synthetic counter (1, 2, 3, ...). Override with the real outer
+								// envelope's clientSequenceNumber so downstream telemetry doesn't get a
+								// misleading "fake csn" value.
+								messageClientSequenceNumber: batchStart.batchStartCsn,
 							},
 						},
 						error,
--- a/packages/runtime/container-runtime/src/opLifecycle/duplicateBatchDetector.ts
+++ b/packages/runtime/container-runtime/src/opLifecycle/duplicateBatchDetector.ts
@@ -10,6 +10,43 @@ import { getEffectiveBatchId } from "./batchManager.js";
 import type { BatchStartInfo } from "./remoteMessageProcessor.js";
 
 /**
+ * Identifying info for a previously-recorded batch that we can include in DuplicateBatch telemetry
+ * to help diagnose where the duplicate came from.
+ *
+ * @remarks `batchIdExplicit` distinguishes the two main duplicate-source scenarios:
+ * - `true`: the batchId was stamped on the wire as explicit metadata, indicating a resubmit
+ *   (PendingStateManager.replayPendingStates).
+ * - `false`: the batchId was derived from the wire `clientId` and `batchStartCsn`, indicating a
+ *   fresh, non-resubmit batch.
+ */
+export interface RecordedBatchInfo {
+	/**
+	 * Wire clientId on the message that started the batch (NOT necessarily the `originalClientId`
+	 * encoded in the batchId for resubmits).
+	 */
+	readonly clientId: string;
+	/**
+	 * Wire client sequence number at the start of the batch.
+	 */
+	readonly batchStartCsn: number;
+	/**
+	 * True if the batchId came from explicit metadata on the wire (i.e. a resubmit),
+	 * false if it was derived from clientId + batchStartCsn (i.e. a fresh submit).
+	 */
+	readonly batchIdExplicit: boolean;
+}
+
+interface RecordedBatch {
+	readonly batchId: string;
+	/**
+	 * Identifying info for the batch as observed at runtime.
+	 * `undefined` if the batch was loaded from a summary snapshot (where only the
+	 * `[seqNum, batchId]` pair is persisted).
+	 */
+	readonly info: RecordedBatchInfo | undefined;
+}
+
+/**
  * Detects duplicate batches that can arise from the "parallel fork" scenario:
  * Container 1 is serialized, and Containers 2 and 3 are rehydrated from that state.
  * They both catch up and (re)connect in parallel (at the same time), submitting the same local state,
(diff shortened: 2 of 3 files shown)


anthony-murphy:
## Deep Review

*Reviewed commit `4230044` on 2026-05-19.*

The detector-layer refactor and the `batchStartCsn` override are well-designed and well-tested at the unit layer. No correctness defects found. Three polish items keep this short of sign-off: a missing event-site test that pins the grouped-batch `messageClientSequenceNumber` override (the PR's highest-value correctness line), a silent semantic change to an already-emitted telemetry property that production dashboards depend on, and an undocumented rollout/opt-in story given that #27351 reverted `enableBatchIdTracking` to opt-in only ~2h before this PR opened.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add an event-site test in `containerRuntime.spec.ts` that synthesizes a grouped duplicate batch with an outer-envelope `clientSequenceNumber` distinct from the post-`ungroupOp` sub-message counter and asserts the emitted `DuplicateBatch` event's `messageClientSequenceNumber` equals `batchStartCsn` (not `1`). Without this, a future refactor that reorders the object literal, drops the override, or renames `batchStartCsn` would silently reintroduce the exact `MessageClientSequenceNumber == 1` symptom this PR is shipping to fix.

### Context for Reviewers

- The `messageClientSequenceNumber: batchStart.batchStartCsn` override uses the field markfields introduced in #21714 precisely to surface the real outer-envelope CSN past `OpGroupingManager.ungroupOp`. The substitution is correctness-preserving and is using the field as #21714 intended.
- The snapshot wire format `[number, string][]` is deliberately preserved per #22454 (markfields); the new `info` field is intentionally not persisted, and snapshot-loaded entries record `info: undefined`. Expect `otherFromSnapshot: true` (and missing `other*` runtime fields) for originals loaded from summaries.
- **markfields** authored the foundational PRs here (#22310, #22454, #22501, #21714) — right person for the `batchStartCsn` substitution sign-off the PR description already requests. **kian-thompson** reviewed #22454 and historically audits telemetry-key/format changes in this area. **dannimad** authored #26295 (bidirectional-map perf optimization) if a perf delta is wanted. **daesunp** authored #27351 (the opt-in revert ~2h before this PR) — right person to confirm the rollout story.

- **markfields sign-off on the `batchStartCsn` substitution** — the PR description explicitly requests this ("Worth a careful look ... please confirm `batchStart.batchStartCsn` is the right substitute"). The originator's confirmation is the canonical sign-off.
- **kian-thompson telemetry-shape review** — the new field names (`batchIdExplicit`, `otherClientId`, `otherBatchStartCsn`, `otherBatchIdExplicit`, `otherFromSnapshot`) plus the silent semantic change to `messageClientSequenceNumber` are exactly the surface kian-thompson historically audits (see #22454 review).
- **Two unresolved threads from `copilot-pull-request-reviewer`** worth a human pass alongside the inline threads below: `duplicateBatchDetector.spec.ts:33` (the `makeBatch` helper now accepts `batchId`/`clientId`/`batchStartCsn` all optional, weakening type safety against invalid combinations) and `duplicateBatchDetector.ts:124` (the `assert(other?.batchId === batchId, ...)` pattern relies on optional-chaining for type narrowing; also applies to line 133).
- **Production rollout / opt-in coverage decision** — whether to set `enableBatchIdTracking` for the affected tenants, and on what timeline, is an operational call that belongs with the on-call investigator.

</details>

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3141,11 +3141,26 @@ export class ContainerRuntime
...
+								otherBatchStartCsn: result.otherBatchInfo?.batchStartCsn,
+								otherBatchIdExplicit: result.otherBatchInfo?.batchIdExplicit,
+								otherFromSnapshot: result.otherBatchInfo === undefined,
 								...extractSafePropertiesFromMessage(batchStart.keyMessage),
+								// For grouped batches, `keyMessage` is one of the sub-messages produced by
+								// `OpGroupingManager.ungroupOp`, which overwrites `clientSequenceNumber`
+								// with a synthetic counter (1, 2, 3, ...). Override with the real outer
+								// envelope's clientSequenceNumber so downstream telemetry doesn't get a
+								// misleading "fake csn" value.
+								messageClientSequenceNumber: batchStart.batchStartCsn,
```
anthony-murphy, line 3163: Deep Review: This override silently changes the meaning of an already-emitted telemetry property. `...extractSafePropertiesFromMessage(batchStart.keyMessage)` already emits `messageClientSequenceNumber` (the post-`ungroupOp` synthetic counter); the subsequent `messageClientSequenceNumber: batchStart.batchStartCsn` replaces that value with the outer-envelope CSN.

The PR description confirms the field is in production dashboards today: *"Across 50 production events over 7 days, every `DuplicateBatch` showed ... `MessageClientSequenceNumber == 1`"*. Any KQL query, alert, or dashboard tile keyed off `DuplicateBatch.MessageClientSequenceNumber == 1` will stop matching once this ships.

Before merging, either:

(a) notify the on-call/telemetry owner (cc kian-thompson — historical reviewer on telemetry shape, see #22454) that the field's semantics are changing, and call out the back-compat impact in the PR description; or

(b) emit the corrected value under a distinct field name (e.g. `outerBatchClientSequenceNumber`) and leave the inherited spread value untouched.

(a) is preferred — the existing value is known-misleading and is exactly what this PR is trying to stop surfacing as authoritative — but it requires the explicit communication step. Also worth extending the inline comment at the override site to flag that this *overrides* an inherited field rather than adding a new one.

anthony-murphy, line 3163: Deep Review: The rollout/opt-in path for the new diagnostic fields is undocumented. PR #27351 (daesunp, merged 2026-05-20T00:20:38Z — ~2h before this PR was created) reverted #27262 and shifted `Fluid.ContainerRuntime.enableBatchIdTracking` back to explicit opt-in. The `DuplicateBatch` event only fires inside the duplicate-detection path that is now opt-in.

The PR description's success criterion — *"the next event will tell us which it is"* — implicitly assumes the tenant population emitting these events is opted in. The gated path is reachable via `Fluid.Container.enableOfflineFull` or `Fluid.ContainerRuntime.enableBatchIdTracking`, so this is a documentation gap, not a code defect. But the diagnostic value of the new fields is contingent on opt-in coverage; reviewers and the on-call investigator need that context.

Suggested fix: add a one-paragraph note to the PR description stating which enablement gate will be set for the affected tenant population and roughly when — either the existing flags will be turned on for those tenants, or this PR is staged ahead of a separate re-enablement.

dannimad (approved):
Only needs reformat

PR #27363, opened 2026-05-20
test(client-presence): cleanup type test generation

jason-ha:
- use root entrypoint (empty string - there isn't public versus beta, etc. here)
- remove dead script entry


packages/framework/presence-definitions/package.json
```diff
@@ -105,7 +104,7 @@
 		}
 	},
 	"typeValidation": {
-		"entrypoint": "public",
+		"entrypoint": "",
```
CraigMacomber, line 107: I checked the implementation, and confirmed that this is valid, and does the same thing as public (but I guess makes more sense for this package).

CraigMacomber (approved):

jatgarg (approved):

PR #27364, opened 2026-05-20
ci: fail fast when npmjs publish auth is broken

ChumpChief:
[AB#73812](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73812)

## Description

Adds an `npm whoami` pre-flight check before the `Publish Packages` step in `include-publish-npm-package-steps.yml`, scoped (via `feedKind == 'public'`) to the npmjs.org publish job only. If the service-connection token can't authenticate against the registry, the check fails the step with a clear error pointing at the token, instead of letting the publish proceed and surface the misleading `404 Not Found - PUT` response that npm returns to unauthenticated publishers.

Motivation: build [401538](https://dev.azure.com/fluidframework/internal/_build/results?buildId=401538) (release 2.101.1) failed exactly this way — every retry got a 404 PUT against `@fluidframework/tool-utils`, which reads as a missing-package error but is npm's standard response when the publisher isn't authorized. Surfacing this earlier as an auth error should save time on future incidents.

Caveat: this catches the case where the token is **unusable** (expired/revoked/missing). It does **not** catch the case where the token authenticates fine but lacks publish rights for a specific package (e.g., maintainer demotion or granular-token-scope drift). For that, an `npm publish --dry-run` pre-flight would be required — out of scope here.


tools/pipelines/templates/include-publish-npm-package-steps.yml
```diff
@@ -34,6 +34,24 @@ steps:
...
+    workingDirectory: $(Pipeline.Workspace)/pack/${{ parameters.artifactPath }}
+    script: |
+      set -eu -o pipefail
+      # Fail fast if the service connection's credentials can't authenticate
+      # to the target registry. Without this, an expired/revoked token
+      # surfaces as a misleading "404 Not Found - PUT" from npm during publish.
+      # --userconfig points npm at the .npmrc generated by npmAuthenticate@0,
+      # the same generated content the publish step uses (it copies this file
+      # to ~/.npmrc before running).
+      if ! npm whoami --registry="${{ parameters.feedUrl }}" --userconfig .npmrc; then
```
CraigMacomber, line 51: Seems like the `!` suppresses the errexit opted into by -e, so I believe this should work correctly. That said, I suck at bash and might be missing something.

CraigMacomber (approved):

tools/pipelines/templates/include-publish-npm-package-steps.yml
ChumpChief: You're right — `set -e` is explicitly suppressed inside an `if` condition. From `bash(1)`:

> The shell does not exit if the command that fails is part of the test following the `if` or `elif` reserved words...

So `if ! npm whoami ...; then` works as intended: the `if` test exempts the command from errexit, and the `!` inverts whoami's exit status so the `then` block fires on whoami's failure (rather than its success). We then `exit 1` explicitly inside the block.

PR #27367, opened 2026-05-20
Factor out tree identifier compression and healing utils

CraigMacomber:
## Description

Factor out tree identifier compression and healing utils.

Remove the originator id (which is incorrect) that is passed down into the field batch codec, factoring the encode and decode codecs into different types which only take in exactly what they should have.

This intentionally changes `IIdCompressor` from `@public` to `@public @sealed`. While in some ways this is a breaking change, custom implementations of `IIdCompressor` didn't actually work due to us already having code which assumes they are our implementation and down casts them. Therefor adding `@sealed` is more of a bug fix than a break.

## Reviewer Guidance


packages/dds/tree/src/util/compressedIds.ts
```diff
@@ -0,0 +1,216 @@
...
+ * These are only safe to use in contexts where the encoding id compressor is available
+ * and the session id of the encoding id compressor is known.
+ * Currently the only such case is in ops, excluding attach summary ops.
+ */
+export type OriginatorDependentEncodedId = OpSpaceCompressedId;
+
+/**
+ * Selects how identifiers are encoded for transport/persistence.
+ */
+export enum EncodedIdType {
```
CraigMacomber, line 38: I plan to, in future PRs, update the places using isSummary to control if encoding should permit OriginatorDependentEncodedId to use this enum instead of the isSummary Boolean for a more semantic control, or perhaps wrap their id compressors in a more opinioned simpler type which bakes in the option. I'll see what is cleanest in all the cases when I go to update them.

packages/runtime/id-compressor/package.json
```diff
@@ -162,7 +162,15 @@
 		"typescript": "~5.4.5"
 	},
 	"typeValidation": {
-		"broken": {},
+		"broken": {
+			"Interface_IIdCompressorCore": {
```
CraigMacomber: This type is internal, but selected entry point is legacy. That's concerning.

CraigMacomber: Oh, its unneeded: not sure how this change happened then, but removed.

packages/runtime/id-compressor/src/idCompressor.ts
daesunp, line 438: Should we return `undefined` instead of erroring, since we added the "try" prefix to the method name?

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -58,32 +55,31 @@ import {
 	supportsIncrementalEncoding,
 } from "./format/index.js";
 
+/**
+ * Context for decoding identifiers.
+ * @remarks
+ * See {@link FieldBatchDecodingContext} for the production implementation of this.
+ *
+ * This intentionally avoids exposing anything which depends on the underlying id compressor's session id to avoid confusion with the session if of the compressor which encoded the data.
```
daesunp:
```suggestion
 * This intentionally avoids exposing anything which depends on the underlying id compressor's session id to avoid confusion with the session id of the compressor which encoded the data.
```

packages/dds/tree/src/feature-libraries/chunked-forest/codec/compressedEncode.ts
```diff
@@ -540,7 +548,27 @@ export class EncoderContext implements NodeEncodeBuilder, FieldEncodeBuilder {
 		 * See {@link FieldBatchEncodingContext.isSummary}.
 		 */
 		public readonly isSummary: boolean,
-	) {}
+	) {
+		// Currently we never include originator-dependent identifiers in summaries
+		// (there was a bug which violated this for attach-summaries: it has been fixed, see `IdentifierHealingConfig`.
```
daesunp:
```suggestion
		// (there was a bug which violated this for attach-summaries: it has been fixed, see `IdentifierHealingConfig`).
```

packages/runtime/id-compressor/src/idCompressor.ts
CraigMacomber: No. I have now added more docs both here and in the APi docs for this function to better clarify this.

.changeset/add-originatorless-id-normalization.md
```diff
@@ -0,0 +1,24 @@
+---
+"@fluidframework/id-compressor": minor
+"__section": feature
+---
+Add originatorless normalization for op-space IDs in id-compressor
+
+`IIdCompressor` now includes `tryNormalizeToSessionSpaceWithoutSession(id)`.
+This API supports recovery scenarios where an op-space identifier must be decoded
```
jzaffiro: What does op-space mean?

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -278,45 +274,20 @@ export function readValue(
...
+			// Op-space compressed ids are resolved by the caller-supplied
+			// `resolveEncodedId`, which encapsulates the originator lookup, finalized-id
+			// normalization, and (for the forest summarizer's heal path) UUIDv5 synthesis.
+			const sessionIdOrString: SessionSpaceCompressedId | string =
+				typeof streamValue === "string"
+					? streamValue
+					: idDecodingContext.resolveEncodedId(streamValue as OpSpaceCompressedId);
+			// Performance:
+			// Currently, we just fully expand the identifier here rather than keeping it in the SessionSpaceCompressedId format.
+			// Avoiding this expansion, and keeping the in memory format using SessionSpaceCompressedId would be a good optimization for the future.
```
jzaffiro, line 291: Is this tracked in ADO? Might be good to have a work item linked

packages/dds/tree/src/util/compressedIds.ts
```diff
@@ -0,0 +1,216 @@
...
+} from "@fluidframework/id-compressor";
+import { isFinalId, isStableId } from "@fluidframework/id-compressor/internal";
+import { v5 as uuidV5 } from "uuid";
+
+/**
+ * An encoded identifier that can be decoded without an originator session id.
+ * A finalized compressed id — the same numeric value in op-space and session-space.
+ * @remarks
+ * These are safe to use in contexts where the encoding id compressor is available to decode them,
+ * even when the session id of the encoding id compressor not known.
```
daesunp: nit:
```suggestion
 * even when the session id of the encoding id compressor is not known.
```

daesunp (approved):

packages/runtime/id-compressor/src/test/idCompressor.spec.ts
```diff
@@ -847,6 +847,87 @@ describe("IdCompressor", () => {
...
+				const opSpaceId = compressor.normalizeToOpSpace(localId);
+				assert(isFinalId(opSpaceId));
+				const sessionSpaceId = compressor.tryNormalizeToSessionSpaceWithoutSession(opSpaceId);
+				// Round-trips through normalizeToSessionSpace with the local session,
+				// proving the value is correct.
+				assert.equal(
+					sessionSpaceId,
+					compressor.normalizeToSessionSpace(opSpaceId, compressor.localSessionId),
+				);
+				// And specifically, it maps back to the original local id, not the final id.
```
jzaffiro:
```suggestion
				// Specifically, it maps back to the original local id, not the final id.
```
nit

jzaffiro (approved):
Overall, I'm a bit confused on the expected usage and flow of the APIs, but it's probably best to talk about that offline. Docs look good though, so approving for that.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -58,32 +55,31 @@ import {
 	supportsIncrementalEncoding,
 } from "./format/index.js";
 
+/**
+ * Context for decoding identifiers.
+ * @remarks
+ * See {@link FieldBatchDecodingContext} for the production implementation of this.
+ *
+ * This intentionally avoids exposing anything which depends on the underlying id compressor's session id to avoid confusion with the session id of the compressor which encoded the data.
+ * If the session id of the encoder which encoded the data is known, that information is baked into `resolveEncodedId`.
```
Josmithr: Nit: we should capitalize "ID" (except when referring to the package name and such, I guess)
```suggestion
 * This intentionally avoids exposing anything which depends on the underlying id-compressor's session ID to avoid confusion with the session ID of the compressor which encoded the data.
 * If the session ID of the encoder which encoded the data is known, that information is baked into `resolveEncodedId`.
```

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
```diff
@@ -58,32 +55,31 @@ import {
...
-	isSummary: boolean;
 	/**
-	 * See {@link FieldBatchEncodingContext.healUnresolvableIdentifiersOnDecode}.
+	 * Compressor which can decompress session-space identifiers from {@link resolveEncodedId} as needed.
 	 */
-	healUnresolvableIdentifiersOnDecode?: boolean;
+	readonly idCompressor: Pick<IIdCompressor, "decompress">;
 	/**
-	 * See {@link FieldBatchEncodingContext.sharedObjectId}.
+	 * Resolves an encoded op-space identifier to either a session-space id
```
Josmithr:
```suggestion
	 * Resolves an encoded op-space identifier to either a session-space ID
```

packages/runtime/id-compressor/api-report/id-compressor.public.api.md
```diff
@@ -13,6 +13,7 @@ export interface IIdCompressor {
     normalizeToOpSpace(id: SessionSpaceCompressedId): OpSpaceCompressedId;
     normalizeToSessionSpace(id: OpSpaceCompressedId, originSessionId: SessionId): SessionSpaceCompressedId;
     recompress(uncompressed: StableId): SessionSpaceCompressedId;
+    tryNormalizeToSessionSpaceWithoutSession(id: OpSpaceCompressedId): SessionSpaceCompressedId | undefined;
```
Josmithr, line 16: This interface wasn't previously sealed, and was public. Are we absolutely confident there couldn't be implementors of this interface out there? Otherwise, I don't think we can reasonably make this change now. 

Can we make this method optional? Plan on making it required in 3.0? We should also probably wait until then to mark this as sealed, so we don't think we can make other breaking changes in the interim.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
CraigMacomber: I don't think we have work planned as far into the future as doing this optimization will likely be.

This note is mostly here to help people understand this code, and not break this possible optimization unintentionally, while making it clear that if they are profiling and or trying to optimize this due to it being a problem, that how they might want to do it.

Personally I think its unlikely that putting an item in ADO for this would improve discoverability or usefulness of this note as its importance is small enough that its likely not worth caring about unless you are already looking at this code and our workflows for optimizing stuff generally consist of profiling or benchmarking code, and then looking at the places that are slow for optimizations, and not searching the ADO backlog for optimization tasks.

That said, tracking things like this in ADO can be done, and my decision not to is clearly a subjective trad-off. I just personally think this is not worth tracking there.

packages/runtime/id-compressor/api-report/id-compressor.public.api.md
CraigMacomber: Due to the existence of `toIdCompressorWithCore` which down casts, it is already effectively sealed, and implementing it yourself can cause that conversion to assert.

This type simple predates us having a "sealed" tag, and was documented poorly about that aspect.

If we make this method optional, then we would either need to use toIdCompressorWithCore, which would assert if someone had their own implementation, or continue to rely on implementation details of the id-compressor indirectly, like how it encodes final ids.

That said there are possible layer compat implications. What if someone uses a IIdCompressor from before this change with a tree after it? TypeScript will block them since we made it required, but if they defeat the type safety (as layer compat stuff often seems to) maybe there is a real risk here.

.changeset/add-originatorless-id-normalization.md
CraigMacomber: Link added.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/chunkDecoding.ts
CraigMacomber: If there is a existing item in ADO which is tracking work that this would be part of, then referencing it there (directly or as a child) could be useful, but I don't think we have such an item.

taylorsw04 (approved):

packages/runtime/id-compressor/api-report/id-compressor.public.api.md
Josmithr: Okay, in that case my ask would just be to note this tag change and why it's okay in the PR description + in a changeset.

CraigMacomber: Done.

Josmithr (approved):

PR #27368, opened 2026-05-20
perf(tree): lazy KernelEventBuffer + shared flush dispatcher

Josmithr:
## Overview

Defer `KernelEventBuffer` allocation to first `.events` access on a `TreeNodeKernel`, and replace the per-buffer `flushEventsEmitter` subscription with a single shared `pendingFlushBuffers` set that only enrolls a buffer when it actually buffers an event. The kernel additionally tracks its current event source so the buffer can migrate without being constructed up front.

This removes the up-front cost of constructing an event buffer, its inner `CustomEventEmitter`, and its flush subscription on every `TreeNodeKernel` — none of which were used unless the caller subscribed to `nodeChanged` / `treeChanged` on that node, which is the minority case in typical tree workloads.

No public API surface change; internal-only refactor of the kernel.

## Performance impact

Measured on the `Tree event benchmarks` suite (see how-measured below):

- **Kernel construction (object node) –31 %** (≈ 48 µs/op → ≈ 33 µs/op).
- **Retained kernel memory –22 %** (≈ 2 KiB/kernel saved on small object nodes).
- **`Tree.on` round-trip –5 to –8 %** across hydrated/unhydrated, object/array/map.
- **Overall benchmark geomean –6.5 %** (7,024 ns → ≈ 6,568 ns).

These benefits scale linearly with tree size: applications building trees with many nodes get the proportional construction and memory savings whether or not they subscribe to events.

## How measured

- Bench harness: `packages/dds/tree/src/test/simple-tree/treeEvents.bench.ts` (CPU + retained-memory benches over hydrated and unhydrated `object`, `array` and `map` nodes).
- Runner: `FLUID_TEST_PERF_MODE=1 npx mocha --no-parallel --fgrep "Tree event benchmarks" --reporter @fluid-tools/benchmark/dist/mocha/Reporter.js`.
- Compared against a baseline captured on the same machine immediately before applying the change. Margins of error 0.2–1.0 % per case; reported deltas exceed margin of error in every case noted above.
- Functional verification: full tree unit-test suite (13,914 passing) on the change.

### DIFF
--- a/packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
+++ b/packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
@@ -124,7 +124,7 @@ export class TreeNodeKernel {
 	 * Per-kernel event buffer.
 	 * @remarks
 	 * Allocated lazily on first access to {@link TreeNodeKernel.events}.
-	 * The vast majority of nodes never subscribe to any events, so deferring this allocation
+	 * We expect that the vast majority of nodes will never have associated event subscriptions, so deferring this allocation
 	 * (and the associated source-listener bookkeeping) saves both time and memory.
 	 */
 	#eventBuffer: KernelEventBuffer | undefined = undefined;
--- a/packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
+++ b/packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
@@ -5,6 +5,7 @@
 
 import { createEmitter } from "@fluid-internal/client-utils";
 import type { HasListeners, Listenable, Off } from "@fluidframework/core-interfaces/internal";
+
 import {
 	assert,
 	fail,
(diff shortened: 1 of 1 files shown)


packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -120,14 +121,20 @@ export class TreeNodeKernel {
...
 	/**
-	 * Events registered before hydration.
+	 * Per-kernel event buffer.
 	 * @remarks
-	 * Since these are usually not used, they are allocated lazily as an optimization.
-	 * The laziness also avoids extra forwarding overhead for events from this kernel's anchor node and also avoids registering for events that are unneeded.
-	 * This means optimizations like skipping processing data in subtrees where no subtreeChanged events are subscribed to would be able to work,
-	 * since the kernel does not unconditionally subscribe to those events (like a design which simply forwards all events would).
+	 * Allocated lazily on first access to {@link TreeNodeKernel.events}.
+	 * The vast majority of nodes never subscribe to any events, so deferring this allocation
```
```suggestion
	 * We expect that the vast majority of nodes will never have associated event subscriptions, so deferring this allocation
```

PR #27369, opened 2026-05-20
perf(tree): return bound off() from KernelEventBuffer.on

Josmithr:
## Overview

Replace the per-subscription `() => this.off(eventName, listener)` arrow returned by `KernelEventBuffer.on` with `this.off.bind(this, eventName, listener)`.

A V8 `JSBoundFunction` stores `(target, thisArg, ...boundArgs)` in a fixed shape and is significantly more compact than a fresh arrow + Context pair, which captures the surrounding lexical scope. Functionally identical — both forms invoke `this.off(eventName, listener)` with no other state.

No public API surface change; internal-only.

### DIFF
--- a/packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
+++ b/packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
@@ -447,7 +447,7 @@ class KernelEventBuffer implements Listenable<KernelEvents> {
 
 		this.#events.on(eventName, listener);
 		// Return a bound method instead of an arrow closure. A bound function captures
-		// (target, thisArg, ...boundArgs) in a fixed shape that V8 can optimise more
+		// (target, thisArg, ...boundArgs) in a fixed shape that V8 can optimize more
 		// uniformly than a closure that captures its lexical context.
 		return this.off.bind(this, eventName, listener);
 	}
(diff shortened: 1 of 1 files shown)


noencke (approved):
TIL. Though, I'm still surprised this is anything more than a micro-op. Is this premature optimization? Or is the rationale that because it's the same amount of lines, we may as well just do it the faster way?

Do you think it's something we might employ elsewhere? If we do, we wouldn't want to repeat these same three lines of comments at every instance. We _could_ make it a helper - that would add a micro-cost at registration time, but wouldn't add any additional cost at "off" time. Or, just leave it as is, but remove the comment and instead put a new entry in our coding best practices document that says to use `bind` when you only need `this` and no other state, as a micro-op?

This is fine as is, just trying to sniff out how this technique generalizes...

Josmithr:
> TIL. Though, I'm still surprised this is anything more than a micro-op. Is this premature optimization? Or is the rationale that because it's the same amount of lines, we may as well just do it the faster way?
> 
> Do you think it's something we might employ elsewhere? If we do, we wouldn't want to repeat these same three lines of comments at every instance. We _could_ make it a helper - that would add a micro-cost at registration time, but wouldn't add any additional cost at "off" time. Or, just leave it as is, but remove the comment and instead put a new entry in our coding best practices document that says to use `bind` when you only need `this` and no other state, as a micro-op?
> 
> This is fine as is, just trying to sniff out how this technique generalizes...

This was something Claude identified. It ran its own perf test suite which seemed to show this change having a noticeable impact, given how hot this code path is. I omitted the results from the description because they referred to a perf suite that hasn't been checked in yet. I have a draft PR up for those here:  https://github.com/microsoft/FluidFramework/pull/27372. Waiting for the PR build to pass before promoting it for review.

PR #27377, opened 2026-05-21
build(build-tools): rename typeCompatibility to .cts for future ESM flip

tylerbutler:
## Description

Renames `build-tools/packages/build-tools/src/common/typeCompatibility.ts` → `typeCompatibility.cts` so it always emits CommonJS output (`.cjs` + `.d.cts`) regardless of the package's module type.

### Why now (when the package is still CJS)

This is preparatory work for an eventual flip of `@fluidframework/build-tools` to `"type": "module"`. Today the rename is a no-op behaviorally:
- The package is still CJS, so emit ends up at the same location either way.
- The single intra-package re-export in `index.ts` is updated to `from "./common/typeCompatibility.cjs"` (TypeScript's `node16` moduleResolution maps the `.cjs` specifier to the `.cts` source).

But it lets us land the `.cts` rename + the long explanatory docstring on its own, separately from the eventual `package.json` `type`/`exports` flip, which will rely on this file being the `require` condition entry point for type-only consumers.

### Background

The generated type test files across the repo do:

```ts
import type { TypeOnly, MinimalType, FullType, requireAssignableTo } from "@fluidframework/build-tools";
```

These tests are dual-compiled (ESM + CJS). When the CJS pass runs against an ESM-only `build-tools`, TypeScript throws TS1479 ("ESM module cannot be imported with require"). The standard fix is to expose a CJS-typed entry point via `exports.require`. Producing one requires a `.cts` source file — done here. The matching `package.json` `exports` map will come with the ESM flip PR.

A long docstring at the top of the renamed file explains all of the above for future readers.

### Validation

From `build-tools/packages/build-tools`:
- `pnpm run clean && pnpm run tsc` — clean; emit contains only `typeCompatibility.cjs` / `.d.cts` (no stale `.js`).
- `pnpm run eslint` — clean.
- `pnpm run build:test && pnpm test:mocha` — 137/137 passing.
- `pnpm build core-utils` from the repo root — full client-release-group build still succeeds; type-test compilation in downstream packages is unaffected.

### Minor incidental change

`tsconfig.json` — `"include"` now lists both `.ts` and `.cts`. Required because TypeScript's default glob does not pick up `.cts` files.

## Reviewer Guidance

The file content is unchanged apart from the leading docstring. The only behavioral change is the emit extension. The follow-up ESM-flip PR will add the matching `exports` map entries.


build-tools/packages/build-tools/src/index.ts
```diff
@@ -21,11 +21,12 @@ export type {
 	requireAssignableTo,
 	SkipUniqueSymbols,
 	TypeOnly,
-} from "./common/typeCompatibility.js";
+} from "./common/typeCompatibility.cjs";
 export { getTypeTestPreviousPackageDetails } from "./common/typeTests.js";
 export { type IFluidBuildConfig } from "./fluidBuild/fluidBuildConfig.js";
 export { type IFluidCompatibilityMetadata } from "./fluidBuild/fluidCompatMetadata.js";
 export { FluidRepo } from "./fluidBuild/fluidRepo.js";
+
```
tylerbutler:
```suggestion
```

alexvy86 (approved):
Not an expert on this space, but the explanation in JSDoc makes sense to me. It makes me wonder of the best way of enforcing the condition like "CJS consumers should only `import type`". If someone were using a value import, would things fail for them at compile time or runtime?

tylerbutler:
> Not an expert on this space, but the explanation in JSDoc makes sense to me. It makes me wonder of the best way of enforcing the condition like "CJS consumers should only `import type`". If someone were using a value import, would things fail for them at compile time or runtime?

I think we might be able to make it fail at compile time by omitting everything except the types declaration in exports, but I don't want to risk breaking something else. Since these imports are only used by generated code, I think this is safe enough.

PR #27381, opened 2026-05-21
feat(container-runtime): enhance DuplicateBatch telemetry diagnostics

dannimad:
> Re-opens [#27355](https://github.com/microsoft/FluidFramework/pull/27355) by @anthony-murphy with the biome formatting fix applied so CI can pass. All code is Tony's; only a whitespace fix was added on top.

## Description

The `DuplicateBatch` `dataCorruptionError` event recorded the new (duplicate) batch's identity but not the original's, and its `messageClientSequenceNumber` was the synthetic counter produced by `OpGroupingManager.ungroupOp` rather than the real outer-envelope csn. Both shortcomings made it impossible to tell from telemetry whether a production duplicate was caused by a service-side double-sequence, a client-side resubmit collision, or another path.

### Changes

- **`opLifecycle/duplicateBatchDetector.ts`** — Exported `RecordedBatchInfo` (`clientId`, `batchStartCsn`, `batchIdExplicit`); renamed `batchIdsBySeqNum` → `batchesBySeqNum` (now `Map<number, { batchId, info? }>`); `processInboundBatch` returns the original's `RecordedBatchInfo` (or `undefined` when the original came from a snapshot) on duplicate match. Snapshot wire format `[[sequenceNumber, batchId]]` preserved for backward compatibility.
- **`containerRuntime.ts`** — `DuplicateBatch` event now logs `batchIdExplicit`, `otherClientId`, `otherBatchStartCsn`, `otherBatchIdExplicit`, `otherFromSnapshot`, and overrides `messageClientSequenceNumber` with the real outer-envelope `batchStart.batchStartCsn` (with a code comment explaining the `ungroupOp` fake-csn artifact).
- **Tests** — All existing assertions updated; 2 new tests cover the runtime-info path and explicit-vs-synthesized batchId detection. 11/11 unit + 3/3 integration tests pass.

### Motivation — what production telemetry will now reveal

Across 50 production events over 7 days, every `DuplicateBatch` showed `DupClientId == MsgClientId`, `BatchStartCsn == 1`, `MessageClientSequenceNumber == 1`, and an empty `details.batchId` — which leaves multiple very different root-cause hypotheses (service-side double-sequence vs. client-side resubmit collision vs. another path) indistinguishable. With the new fields, the next event will tell us which it is:

| Field combination on next event                                                       | Likely root cause                                |
| ------------------------------------------------------------------------------------- | ------------------------------------------------ |
| `batchIdExplicit == false` **and** `otherBatchIdExplicit == false`, both csn equal    | Service-side double-sequence — escalate to ODSP. |
| Exactly one of the two `*Explicit` flags is `true`                                    | Client-side resubmit colliding with fresh submit. |

## Reviewer Guidance

- **`DuplicateBatchDetector` is internal** (exported from `opLifecycle/index.ts` but not in any `api-report.md`). No public API surface change, no changeset required.
- **Snapshot wire format unchanged**: `getRecentBatchInfoForSummary` still emits the same `[number, string][]` tuples; a round-trip integration test exercises this.
- **Worth a careful look**: the `messageClientSequenceNumber` override is correctness-tied — please confirm `batchStart.batchStartCsn` is the right substitute (it's set from the outer-envelope `clientSequenceNumber` before `ungroupOp` runs; see `remoteMessageProcessor.ts`).
- **Out of scope, but related**: the same fake-csn artifact exists in other call sites that consume grouped sub-messages via `extractSafePropertiesFromMessage` (`channelCollection.ts:480/526/578`, `dataStoreContext.ts:761`, `pendingStateManager.ts:729`, `inboundBatchAggregator.ts:183`). Deliberately left for a follow-up — flagging for awareness.

Tagged `deep-review` because the fix is bound up with a live production data-corruption investigation; we want to be sure the new telemetry is correct before the next event lands.

### DIFF
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -3141,11 +3141,26 @@ export class ContainerRuntime
 							eventName: "DuplicateBatch",
 							details: {
 								batchId: batchStart.batchId,
+								batchIdExplicit: batchStart.batchId !== undefined,
 								clientId: batchStart.clientId,
 								batchStartCsn: batchStart.batchStartCsn,
 								size: inboundResult.length,
 								duplicateBatchSequenceNumber: result.otherSequenceNumber,
+								// Identifying info for the ORIGINAL occurrence of this batch, so we can
+								// disambiguate the duplicate's source (e.g. resubmit vs fresh submit, same
+								// vs different wire clientId). Undefined fields indicate the original was
+								// loaded from a summary snapshot rather than seen at runtime.
+								otherClientId: result.otherBatchInfo?.clientId,
+								otherBatchStartCsn: result.otherBatchInfo?.batchStartCsn,
+								otherBatchIdExplicit: result.otherBatchInfo?.batchIdExplicit,
+								otherFromSnapshot: result.otherBatchInfo === undefined,
 								...extractSafePropertiesFromMessage(batchStart.keyMessage),
+								// For grouped batches, `keyMessage` is one of the sub-messages produced by
+								// `OpGroupingManager.ungroupOp`, which overwrites `clientSequenceNumber`
+								// with a synthetic counter (1, 2, 3, ...). Override with the real outer
+								// envelope's clientSequenceNumber so downstream telemetry doesn't get a
+								// misleading "fake csn" value.
+								messageClientSequenceNumber: batchStart.batchStartCsn,
 							},
 						},
 						error,
--- a/packages/runtime/container-runtime/src/opLifecycle/duplicateBatchDetector.ts
+++ b/packages/runtime/container-runtime/src/opLifecycle/duplicateBatchDetector.ts
@@ -82,7 +82,7 @@ export class DuplicateBatchDetector {
 	/**
 	 * Records this batch's batchId, and checks if it's a duplicate of a batch we've already seen.
 	 * If it's a duplicate, also return the sequence number of the other batch (and identifying info,
-	 * if the other batch was seen at runtime rather than loaded from snapshot) for logging.
+	 * if the other batch was seen during this container session rather than loaded from snapshot) for logging.
 	 *
 	 * @remarks We also use the minimumSequenceNumber to clear out old batchIds that are no longer at risk for duplicates.
 	 */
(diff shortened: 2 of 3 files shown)


packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3141,11 +3141,26 @@ export class ContainerRuntime
...
+								otherBatchStartCsn: result.otherBatchInfo?.batchStartCsn,
+								otherBatchIdExplicit: result.otherBatchInfo?.batchIdExplicit,
+								otherFromSnapshot: result.otherBatchInfo === undefined,
 								...extractSafePropertiesFromMessage(batchStart.keyMessage),
+								// For grouped batches, `keyMessage` is one of the sub-messages produced by
+								// `OpGroupingManager.ungroupOp`, which overwrites `clientSequenceNumber`
+								// with a synthetic counter (1, 2, 3, ...). Override with the real outer
+								// envelope's clientSequenceNumber so downstream telemetry doesn't get a
+								// misleading "fake csn" value.
+								messageClientSequenceNumber: batchStart.batchStartCsn,
```
markfields, line 3163: nit: I'd just omit it rather than remapping it

packages/runtime/container-runtime/src/opLifecycle/duplicateBatchDetector.ts
```diff
@@ -24,31 +59,40 @@ export class DuplicateBatchDetector {
...
 				this.seqNumByBatchId.set(batchId, seqNum);
 			}
 		}
 	}
 
 	/**
 	 * Records this batch's batchId, and checks if it's a duplicate of a batch we've already seen.
-	 * If it's a duplicate, also return the sequence number of the other batch for logging.
+	 * If it's a duplicate, also return the sequence number of the other batch (and identifying info,
+	 * if the other batch was seen at runtime rather than loaded from snapshot) for logging.
```
markfields: nit

```suggestion
	 * if the other batch was seen during this container session rather than loaded from snapshot) for logging.
```

packages/runtime/container-runtime/src/opLifecycle/duplicateBatchDetector.ts
```diff
@@ -64,21 +108,36 @@ export class DuplicateBatchDetector {
...
-			!this.batchIdsBySeqNum.has(sequenceNumber),
+			!this.batchesBySeqNum.has(sequenceNumber),
 			0xce1 /* seqNumToBatchId and batchIdToSeqNum should be in sync */,
 		);
 
-		// Add new batch
-		this.batchIdsBySeqNum.set(sequenceNumber, batchId);
+		// Add new batch. Record identifying info so we can report it if a future duplicate matches us.
+		const info: RecordedBatchInfo | undefined =
+			batchStart.clientId === undefined || batchStart.batchStartCsn === undefined
```
markfields: This is overly-defensive IMO.  We can trust the TypeScript types here, these will always be defined.

packages/runtime/container-runtime/src/opLifecycle/duplicateBatchDetector.ts
```diff
@@ -108,16 +167,18 @@ export class DuplicateBatchDetector {
...
 		telemetryContext?.set(
 			"fluid_DuplicateBatchDetector_",
 			"recentBatchCount",
-			this.batchIdsBySeqNum.size,
+			this.batchesBySeqNum.size,
 		);
 
-		return [...this.batchIdsBySeqNum.entries()];
+		return [...this.batchesBySeqNum.entries()].map(
+			([seqNum, recorded]) => [seqNum, recorded.batchId] as [number, string],
```
markfields: Remove the `as` if not required?

packages/runtime/container-runtime/src/opLifecycle/duplicateBatchDetector.ts
```diff
@@ -64,21 +108,33 @@ export class DuplicateBatchDetector {
...
 		assert(
-			!this.batchIdsBySeqNum.has(sequenceNumber),
+			!this.batchesBySeqNum.has(sequenceNumber),
 			0xce1 /* seqNumToBatchId and batchIdToSeqNum should be in sync */,
 		);
 
-		// Add new batch
-		this.batchIdsBySeqNum.set(sequenceNumber, batchId);
+		// Add new batch. Record identifying info so we can report it if a future duplicate matches us.
+		const info: RecordedBatchInfo | undefined = {
```
markfields, line 130: super nit!  Totally optional
```suggestion
		const info: RecordedBatchInfo = {
```

markfields (approved):

PR #27382, opened 2026-05-21
perf(tree): Lazily initialize event buffer

Josmithr:
Most nodes likely don't get any event subscriptions. Lazy initialization avoids unnecessary allocations and internal event registrations..


noencke (approved):
Nice!

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -232,6 +234,7 @@ export class TreeNodeKernel {
 	}
 
 	public get events(): Listenable<KernelEvents> {
+		assert(!this.disposed, "Cannot register events on a disposed node");
```
noencke, line 237: Why doesn't this also use `this.assertNotDisposed()`?

noencke (approved):

PR #27386, opened 2026-05-26
merge chunks during attach/detach edits

daesunp:
## Description

Follow-up PR to #27153. Adds `coalesceAroundSplice`, the inverse of `splitFieldAtIndex`: merges adjacent same-shape `UniformChunk`s along the seams a splice could have created, so repeated mid-field attach/detach operations don't leave the field permanently fragmented.

`splitFieldAtIndex` (from #27153) lets attach/detach land in the middle of a multi-node `UniformChunk` by splitting the chunk. Without a corresponding merge, every mid-chunk edit fragments the field — repeated same-shape inserts would produce an ever-growing run of small adjacent chunks. The `uniformChunkNodeCountDynamicTargetMax` policy field added in #27153 explicitly anticipated this work; this PR provides the merge logic it was waiting for.

### The win - chunk count stays batched after editing

`insertAt` typing one character at a time, then counting `UniformChunk`s in the forest. Without coalescing every character stays a single-node chunk (~N); with coalescing same-shape neighbors merge (~N/10).

| characters typed | UniformChunks (this change) | without coalescing |
|---|---|---|
| 100 | 11 | ~100 |
| 1,000 | 101 | ~1,000 |
| 10,000 | 1,001 | ~10,000 |

### The cost - per-edit overhead

Each case performs a real insert edit (split a multi-node `UniformChunk`, splice in a node, then optionally coalesce the seam), mirroring the forest's `attachEdit`. The baseline does the split+splice **without** coalescing, so the delta above it is the coalescing overhead.

These results were rerun after the `TreeShape.equals()` identity and type-mismatch short circuits landed. Values are the median of three runs and use microseconds throughout.

| case | time / edit | overhead vs. baseline |
|---|---:|---:|
| insert edit, no coalesce (baseline) | 0.74 us | - |
| insert edit + coalesce, shallow homogeneous | 1.24 us | 0.50 us |
| insert edit + coalesce, deep homogeneous | 2.15 us | 1.41 us |
| insert edit + coalesce, deep heterogeneous | 1.59 us | 0.85 us |

The heterogeneous case is no longer the worst case: because the shapes have different top-level types, `TreeShape.equals()` now rejects them before walking the nested structure. The largest measured overhead is the deep homogeneous case at ~1.41 us per edit, still a small fraction of a keystroke's end-to-end cost (~77 us geomean in the whole-document edit benchmark).

### Memory reclaimed by coalescing

A field fragmented into single-node `UniformChunk`s vs. the same field after coalescing. Coalescing removes per-chunk overhead (not stored values), so a light (plain-leaf) shape reclaims a large fraction while a heavy (deep object) shape reclaims less.

| shape | count | fragmented | coalesced | reclaimed |
|---|---:|---:|---:|---:|
| light (plain) | 1,000 | 99.4 KiB | 36.9 KiB | ~62% |
| heavy (deep node) | 1,000 | 99.2 KiB | 83.5 KiB | ~16% |

### The boundary - in-place edits are not addressed by coalescing

A large array of same-shape objects batches into a few multi-node `UniformChunk`s. A single in-place field edit shatters **only the one chunk enclosing that node** into per-node `BasicChunk`s (`enterNode`); coalescing (which acts on attach/detach seams) neither causes nor undoes this pre-existing behavior. Included to bound where the win applies.

| array | basicChunks | largestUniformChunkNodes | totalChunks |
|---|---:|---:|---:|
| fresh, 100 objects | 2 | 100 | 3 |
| after 1 middle in-place edit | 102 | 1 | 203 |
| fresh, 1,000 objects | 2 | 200 | 7 |
| after 1 middle in-place edit | 202 | 200 | 407 |

For the 1,000-object array, `largestUniformChunkNodes` stays 200 after the edit - the shatter is localized to the single enclosing chunk; the other chunks are untouched.

### Summary

- **No regression** in correctness, memory, or execution time.
- **Win:** content stays batched after editing (~10x fewer chunks) instead of fragmenting to one chunk per node.
- **Cost:** at most ~1.41 us per edit in these cases - negligible relative to keystroke latency.
- **Boundary:** the win applies to attach/detach edits (for example, text insert/remove); in-place field edits on composite nodes still shatter, which is pre-existing and unchanged by this PR.


packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -640,6 +640,98 @@ export function splitFieldAtIndex(
 	return chunks.length;
 }
 
+/**
+ * Keeps a field's chunks from accumulating same-shape fragments across repeated edits.
```
CraigMacomber: Generally, the top level documentation for a function should say what it does. This is instead documenting what its used to accomplish. Additionally, this doesn't prevent there from being same shaped fragments, so it not quite accurate either.

Right now the API and docs are focused around this being used in exactly one use-case, when a `TreeChunk[]` which holds the contents of a field has undergone a splice.

That is way more specific that this should be, making this code way more fragile than it needs to be. If we refactor or add some logic which needs to work on a TreeChunk[] thats not a field (like an ArrayChunk) or wasn't spliced (maybe was appended to), this code could still be valid if you just change its documentation and parameter names to be more general.

I'd suggest something like:

```suggestion
 * Coalesce adjacent small uniform chunks with the same shape in part of the provided array. 
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -640,6 +640,98 @@ export function splitFieldAtIndex(
...
+ *
+ * @remarks
+ * Merges adjacent {@link UniformChunk}s of matching shape along the seams a splice could have
+ * created. Acts as the inverse of {@link splitFieldAtIndex}: without it, repeated mid-field
+ * attach/detach against the same field would leave it permanently fragmented into ever-smaller
+ * adjacent chunks. Cost is proportional to `insertedCount`.
```
CraigMacomber: Generally it seems like this function's runtime O(insertedCount) when it is a no-op, and worst case O(field length * insertedCount)

I'm not sure what "cost" you are referring to (memory, time, dollars?): use a more specific term. Assuming that was supposed to be time, I think you missed this can do O(insertedCount) splices which are O(field length) in cost.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -640,6 +640,98 @@ export function splitFieldAtIndex(
...
+ * @privateRemarks
+ * Caps merged chunks at {@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax}, matching the
+ * bisect threshold used by {@link splitFieldAtIndex} so split-and-coalesce pairs don't oscillate.
+ *
+ * @param field - The field's chunks array, modified in place.
+ * @param spliceStart - The index passed to the originating `splice` call.
+ * @param insertedCount - The number of chunks the splice inserted (0 for a pure detach).
+ * @param policy - The {@link ChunkPolicy} supplying the per-chunk cap.
+ */
+export function coalesceAroundSplice(
```
CraigMacomber: Since there are multiple ways one could coalesce chunks (like put them into ArrayChunks, or merge ArrayChunks) and this seems to only handle UniformChunks, despite taking in a ChunkPolicy which includes details about how to use ArrayChunks, I think it should have a more specific name.

I think it might also make sense for it to make the sub-range which it focuses on optional, and default to processing the whole array. Maybe take in a `range?: {start: number, length: number}`.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -640,6 +640,98 @@ export function splitFieldAtIndex(
...
+ * created. Acts as the inverse of {@link splitFieldAtIndex}: without it, repeated mid-field
+ * attach/detach against the same field would leave it permanently fragmented into ever-smaller
+ * adjacent chunks. Cost is proportional to `insertedCount`.
+ *
+ * @privateRemarks
+ * Caps merged chunks at {@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax}, matching the
+ * bisect threshold used by {@link splitFieldAtIndex} so split-and-coalesce pairs don't oscillate.
+ *
+ * @param field - The field's chunks array, modified in place.
+ * @param spliceStart - The index passed to the originating `splice` call.
```
CraigMacomber: You should clarify this is a chunk index not a node index. Might even be good to call it something like spliceStartChunkIndex

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -640,6 +640,98 @@ export function splitFieldAtIndex(
...
+ * @param insertedCount - The number of chunks the splice inserted (0 for a pure detach).
+ * @param policy - The {@link ChunkPolicy} supplying the per-chunk cap.
+ */
+export function coalesceAroundSplice(
+	field: TreeChunk[],
+	spliceStart: number,
+	insertedCount: number,
```
CraigMacomber: it might be nice for the name to clarify this is chunks not nodes (its good the docs are clear this time though)

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -640,6 +640,98 @@ export function splitFieldAtIndex(
...
+
+/**
+ * Attempts to merge `field[i]` and `field[i + 1]` into a single {@link UniformChunk}. Skips if
+ * either is not a {@link UniformChunk}, the shapes differ, the chunks carry different non-undefined
+ * {@link UniformChunk.idCompressor}s, or the combined topLevelLength would exceed
+ * {@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax}.
+ *
+ * @returns `true` if the merge occurred (and `field` was mutated), `false` otherwise.
+ */
+function tryMergeAt(field: TreeChunk[], i: number, policy: ChunkPolicy): boolean {
```
CraigMacomber: this function seems like a good target for unit testing, but seeing as its not exported, it can't have any tests.

CraigMacomber: Avoid single letter name like "i" and use more descriptive ones made of English words.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -640,6 +640,98 @@ export function splitFieldAtIndex(
...
+	const left = field[i];
+	const right = field[i + 1];
+	if (!(left instanceof UniformChunk) || !(right instanceof UniformChunk)) {
+		return false;
+	}
+
+	const leftTreeShape = left.shape.treeShape;
+	const rightTreeShape = right.shape.treeShape;
+	// Identity is the fast path (chunkers cache shapes); equals() is the fallback.
+	if (leftTreeShape !== rightTreeShape && !leftTreeShape.equals(rightTreeShape)) {
```
CraigMacomber: If we need to optimize equality here, I think it would be better to move the fast path into the implementation of TreeShape.equals (which should probably also move its O(1) check at the bottom to before its fields check).

You can make that optimization to TreeShape.equals in its own PR, and remove the workaround for the lack of that optimization from here.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -640,6 +640,98 @@ export function splitFieldAtIndex(
...
+		return false;
+	}
+
+	// Refuse to merge chunks whose values were compressed under different idCompressors;
+	// the merged chunk can carry only one, so the other's compressed-id values would
+	// decompress to incorrect values when read.
+	const leftCompressor = left.idCompressor;
+	const rightCompressor = right.idCompressor;
+	if (
```
brrichards: can we have the scenario where chunks in the same forest have a different idCompressor? Do all chunks created in the same forest not reference the same idCompressor?

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -640,6 +640,98 @@ export function splitFieldAtIndex(
...
+	if (combinedTopLevel > policy.uniformChunkNodeCountDynamicTargetMax) {
+		return false;
+	}
+
+	const merged = new UniformChunk(
+		leftTreeShape.withTopLevelLength(combinedTopLevel),
+		[...left.values, ...right.values],
+		leftCompressor ?? rightCompressor,
+	);
+	field.splice(i, 2, merged);
```
CraigMacomber: I don't think the algorithm you are using here id ideal.

Rather than calling tryMergeAt, and passing in the index, I think it would make more sense to:
1. make this function into something like "tryCoalesceChunks" which takes in two chunks and the chunk policy, and returns a chunk (replacing the two inputs) or undefined (the two inputs should still be used).
2. The outer function can then scan over the relevant range, creating a new array of chunks for the impacted range. It can then do a single splice operation before returning, avoiding the worse case of doing many splices. Take care to avoid doing the splice at all if there are no changes.

If you take this approach, it fixes the worst case performance case, and also makes the two function much more usable. For example the new inner function I propose could be used directly when inserting content to try first to merge it with the chunk before it, then the one after it, and only insert to the array if both merges fail. In the case where this does a merge, it prevents the insert from having to splice the array at all, saving the possible O(n) cost of moving everything in the array.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -640,6 +640,98 @@ export function splitFieldAtIndex(
...
+	) {
+		return false;
+	}
+
+	const combinedTopLevel = left.topLevelLength + right.topLevelLength;
+	if (combinedTopLevel > policy.uniformChunkNodeCountDynamicTargetMax) {
+		return false;
+	}
+
+	const merged = new UniformChunk(
```
CraigMacomber, line 775: When merging chunks, ideally instead of always creating a new chunk, if the first/left chunk only as ref count 1, just append the second chunk's content array to it, and update the chunk shape to one with the new top level length. That way if you try and merge a bunch of chunks together in a row, instead of getting O(n^2) array data allocated and filled for the values arrays, you only get O(n) as you grow a single chunk.

Note: because of this issue, I think  coalesceAroundSpliceis actually has worst case runtime of O(Max(insertedCount^2, insertedCount * fields.length))

Splitting out a dedicated coalesceUniformChunks which takes in two uniform chunks, asserts the shapes are the same, then does this merging approach could be good, so it can have its own targeted tests about getting the refcounts and top level lengths correct.

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -272,5 +272,111 @@ describe("ChunkedForest", () => {
 			}
 			assert.deepEqual(values, [0, 1, 99, 2, 3, 4]);
 		});
+
+		/**
+		 * Manually seeds the root field with the requested sequence of single-shape UniformChunks.
+		 * Sidesteps the default chunker, which would produce one UniformChunk per top-level node,
```
CraigMacomber: I don't think this is accurate anymore (@brrichards fixed it I believe)

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
```diff
@@ -872,4 +883,239 @@ describe("chunkTree", () => {
...
+			);
+			const field: TreeChunk[] = [leftWithCompressor, rightWithOtherCompressor];
+			const snapshot = [...field];
+
+			coalesceAroundSplice(field, 1, 1, policy);
+
+			assertChunksUnchanged(field, snapshot);
+		});
+
+		it("keeps chunk count bounded under repeated mid-field edits", () => {
```
brrichards: nit: This test looks like it rewrites the splitFieldAtIndex logic which using might simplify/help with readability, but the test itself looks good!

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
brrichards: Yeah the default chunker now produces multi-node UniformChunks

daesunp: Updated stale comment

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated so when `!left.isShared()`, `tryCoalesceUniformChunks` extends `left.values` in place, updates `left.shape`, releases `right`, and returns `left`. Otherwise it falls back to allocating a new chunk.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated to:
- new `tryCoalesceUniformChunks(left, right, policy)` inner helper and exported
- outer `coalesceUniformChunks` scans the range, builds a result and does splice at end (if there are changes)

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: I was about to remove it, but since the function is exported now, I updated it to an assert as a guard :)

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: removed optimization and added TODO

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: updated to `chunkIndex`

daesunp: inner helper renamed to `tryCoalesceUniformChunks` and added unit tests in `chunkTree.spec.ts`

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: `insertedCount` removed after suggested refactor

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: removed `spliceStart` and updated to your suggestion above

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Renamed `coalesceAroundSplice` to `coalesceUniformChunks` and updated to take in `(chunks, policy, range?: { start, length })`

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated and clarified what is being measured

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated with suggestion

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
daesunp: Updated with suggestion

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+ *
+ * @remarks
+ * Walks the range from left to right, attempting to merge each chunk with its left neighbor via
+ * {@link tryCoalesceUniformChunks}. Performs at most one in-place `splice` on `chunks`. Non-mergeable
+ * chunks are passed through unchanged.
+ *
+ * Time: O(range_length) for the walk, plus the cost of one splice (O(chunks.length)) and the
+ * cost of the merges themselves. Per-chunk merge cost is O(rightChunk.values.length) when the
+ * surviving chunk can be mutated in place (see {@link tryCoalesceUniformChunks}), so the total
+ * work of merging a run of `k` chunks together is O(total values), not O(total values squared).
```
Josmithr: Nit: I'm not sure we need to include this. I don't think we generally document complexity like this. Seems like this could get out of date easily.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+ *
+ * @privateRemarks
+ * Caps merged chunks at {@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax}, matching the
+ * bisect threshold used by {@link splitFieldAtIndex} so split-and-coalesce pairs don't oscillate.
```
Josmithr: I'm not sure I understand this. Could we expand on this a bit? Also, does this need to be in a `@privateRemarks` block? Those are usually reserved for docs we don't want end-users to see, and this note seems relevant.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+ *
+ * @param chunks - The chunks array, modified in place.
```
Josmithr, line 663: Thanks for calling out that the array is modified in place! In my opinion, this is one of the most critical things to document in an API, and it frequently isn't.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+ * @param policy - The {@link ChunkPolicy} supplying the per-chunk cap.
+ * @param range - Sub-range of `chunks` (by chunk index) to consider for merging. Defaults to the
+ * whole array. `length` is clamped to the available chunks.
+ */
+export function coalesceUniformChunks(
+	chunks: TreeChunk[],
+	policy: ChunkPolicy,
+	range?: { readonly start: number; readonly length: number },
```
Josmithr: We actually have an `IndexRange` util type we could leverage here. I think it lives in `ArrayUtilities.ts`.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+ * work of merging a run of `k` chunks together is O(total values), not O(total values squared).
+ *
+ * @privateRemarks
+ * Caps merged chunks at {@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax}, matching the
+ * bisect threshold used by {@link splitFieldAtIndex} so split-and-coalesce pairs don't oscillate.
+ *
+ * @param chunks - The chunks array, modified in place.
+ * @param policy - The {@link ChunkPolicy} supplying the per-chunk cap.
+ * @param range - Sub-range of `chunks` (by chunk index) to consider for merging. Defaults to the
+ * whole array. `length` is clamped to the available chunks.
```
Josmithr: Is it important that we do this clamping on the caller's behalf? Or would it be reasonable to just assert that the provided `length` is valid? Assuming this is a hot code path, we could use a `debugAssert` to avoid the comparison cost in production.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+): void {
+	const rangeStart = range?.start ?? 0;
+	const rangeLength = range?.length ?? chunks.length - rangeStart;
+	const rangeEnd = Math.min(rangeStart + rangeLength, chunks.length);
+	if (rangeEnd - rangeStart < 2) {
+		return;
+	}
+
+	// Build the resulting chunk list for [rangeStart, rangeEnd). Passthrough chunks keep their
+	// identity; merged pairs are replaced by the chunk returned from tryCoalesceUniformChunks,
```
Josmithr: Nit
```suggestion
	// identity; merged pairs are replaced by the chunk returned from `tryCoalesceUniformChunks`,
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+	// which also handles ref-count bookkeeping for chunks that drop out of the array.
+	const result: TreeChunk[] = [];
+	let mutated = false;
+	for (let chunkIndex = rangeStart; chunkIndex < rangeEnd; chunkIndex++) {
+		const next = chunks[chunkIndex] ?? oob();
+		if (result.length === 0) {
+			result.push(next);
+			continue;
+		}
+		const last = result[result.length - 1] ?? oob();
```
Josmithr: Nit: "next" and "last" are a bit confusing for this, in my opinion. Could we do "current" and "previous", or "next" and "current"?

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+	}
+
+	if (mutated) {
+		chunks.splice(rangeStart, rangeEnd - rangeStart, ...result);
+	}
+}
+
+/**
+ * Attempts to combine two adjacent {@link UniformChunk}s into a single {@link UniformChunk}. Returns
+ * the merged chunk on success, or `undefined` when the pair is not mergeable.
```
Josmithr: Nit: The second sentence here is already captured by the `@returns` block at the bottom. I think we can remove it.
```suggestion
 * Attempts to combine two adjacent {@link UniformChunk}s into a single {@link UniformChunk}.
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+ * @remarks
+ * Skips if either input is not a {@link UniformChunk}, the {@link TreeShape}s differ, or the
+ * combined `topLevelLength` would exceed
+ * {@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax}.
+ *
+ * Asserts that the two inputs do not carry different non-undefined
+ * {@link UniformChunk.idCompressor}s: that case would silently produce a merged chunk whose
+ * compressed-id values decompress to incorrect strings under the surviving compressor. All
+ * chunks in a single {@link ChunkedForest} share that forest's idCompressor, so the assertion
+ * documents the invariant rather than guarding a path callers should take.
```
Josmithr: Nit: I don't think this note is necessary.
```suggestion
 * compressed-id values decompress to incorrect strings under the surviving compressor.
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+ *
+ * @remarks
```
Josmithr: There should only be a single `@remarks` block. I'm actually a little surprised our linter didn't complain about this.
```suggestion
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+	right: TreeChunk,
+	policy: ChunkPolicy,
+): UniformChunk | undefined {
+	if (!(left instanceof UniformChunk) || !(right instanceof UniformChunk)) {
+		return undefined;
+	}
+	const leftTreeShape = left.shape.treeShape;
+	const rightTreeShape = right.shape.treeShape;
+	// TODO: TreeShape.equals could short-circuit on identity for the common
+	// (same-chunker) case; until then this is a full structural check on every merge.
```
Josmithr: Should we do this in this PR?

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+	assert(
+		leftCompressor === undefined ||
+			rightCompressor === undefined ||
+			leftCompressor === rightCompressor,
+		"tryCoalesceUniformChunks: left and right carry different idCompressors",
+	);
+	const combinedTopLevel = left.topLevelLength + right.topLevelLength;
+	if (combinedTopLevel > policy.uniformChunkNodeCountDynamicTargetMax) {
+		return undefined;
+	}
```
Josmithr, line 762: Nit: A quick comment explaining this might be useful.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +641,148 @@ export function splitFieldAtIndex(
...
+
+	if (!left.isShared()) {
+		// In-place: grow `left` to absorb `right`. `left`'s sole array-slot ref is preserved
+		// and returned; `right`'s slot ref is released.
+		left.values.push(...right.values);
+		left.shape = leftTreeShape.withTopLevelLength(combinedTopLevel);
+		if (leftCompressor === undefined && rightCompressor !== undefined) {
```
Josmithr: We assert above that `leftCompressor` cannot be `undefined`. Do we need this?

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -272,5 +272,110 @@ describe("ChunkedForest", () => {
...
 			assert.deepEqual(values, [0, 1, 99, 2, 3, 4]);
 		});
+
+		/**
+		 * Manually seeds the root field with the requested sequence of single-shape UniformChunks.
+		 * Lets tests position attach/detach boundaries on existing chunk seams and exercise
+		 * coalesceUniformChunks without splitFieldAtIndex having to bisect first.
+		 */
+		function setupForestWithChunks(chunkSizes: readonly number[]): {
+			forest: ReturnType<typeof setupForest>;
```
Josmithr: Any reason this needs to return an object containing the `forest` property, rather than just returning the forest?

Josmithr: It looks like callers are doing `const { forest } = ...` anyways.

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -272,5 +272,110 @@ describe("ChunkedForest", () => {
 			}
 			assert.deepEqual(values, [0, 1, 99, 2, 3, 4]);
 		});
+
+		/**
+		 * Manually seeds the root field with the requested sequence of single-shape UniformChunks.
+		 * Lets tests position attach/detach boundaries on existing chunk seams and exercise
+		 * coalesceUniformChunks without splitFieldAtIndex having to bisect first.
```
Josmithr: Nit: This documentation contains some useful notes about the setup here, but they don't document what the function _does_. These contents look like they belong in a `@remarks` block, and we still need the semantic summary.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Simplified and removed complexity comments

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
```diff
@@ -1030,4 +1042,267 @@ describe("chunkTree", () => {
...
+
+				const result = tryCoalesceUniformChunks(left, right, policy);
+
+				assert(result !== undefined);
+				assert(result !== left, "result should be a fresh chunk, not left");
+				assertNumbersChunk(result, [1, 2, 3, 4]);
+				// left's slot-ref was released (only the extra ref remains).
+				assert.equal(left.isShared(), false);
+				assert.equal(right.isShared(), false);
+				left.referenceRemoved();
```
Josmithr: What does this call do?

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: moved to `@remarks` and elaborated for clarification

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
```diff
@@ -1030,4 +1042,267 @@ describe("chunkTree", () => {
 			assert.equal(chunks.length, 0);
 		});
 	});
+
+	describe("coalesceUniformChunks", () => {
```
Josmithr, line 1046: Given the ref-counting aspect of the chunks, it seems like we should have some explicit test coverage of how the counts are affected by calling these functions.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Removed `math.min()`, added assert, and updated docs

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated with suggestions

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: updated with suggestions

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated with suggestions

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated with suggestions

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: removed extra remarks tag

Josmithr (commented):
Left some questions and suggestions.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: It was originally part of this PR, but Craig suggested that we split it up into two

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated with suggestion

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: The above assert checks if either right or left compressor is undefined, so this was trying to set it to the rightCompressor if it is undefined (similarly to `leftCompressor ?? rightCompressor` used below it). But the if statement seemed unnecessary and updated it to `left.idCompressor ??= rightCompressor`

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
daesunp: Updated with suggestion

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
daesunp: Updated doc with a semantic summary and moved contents to remarks block

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
daesunp: This was just to balance the left.referenceAdded() above, but realized it is not needed for the test, since we aren't doing any checks. Removed :)

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
daesunp: Updated 2 of our tests to have additional checks with asserts for ref-counts

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
Josmithr: Do we have a task tracking this next step? Let's make sure we have something tracking this.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Added explanation

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Added [task](https://dev.azure.com/fluidframework/internal/_workitems/edit/75982) :)

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
CraigMacomber: Its unlikely, but chunks can live outside of forests (for example in edits) and get added to a chunked forest regardless of where they are from, so someday it could theoretically happen in a case like copy paste from one fluid container into a new one. No current public api can hit that, but if we optimize a few things, it might eventually become possible.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +645,145 @@ export function splitFieldAtIndex(
...
+	left: TreeChunk,
+	right: TreeChunk,
+	policy: ChunkPolicy,
+): UniformChunk | undefined {
+	if (!(left instanceof UniformChunk) || !(right instanceof UniformChunk)) {
+		return undefined;
+	}
+	const leftTreeShape = left.shape.treeShape;
+	const rightTreeShape = right.shape.treeShape;
+	// TODO: AB#75982 TreeShape.equals could short-circuit on identity for the common
```
CraigMacomber: given the change is trivial (At least I think it should be) and it will help with the perf here, why not make another PR which fixes this instead of adding documentation about it. The diff is similar is size to this comment anyway.

daesunp: Merged PR for short-circuit optimization in TreeShape.equals, and removed stale TODO comment

.changeset/tall-tips-teach.md
```diff
@@ -0,0 +1,9 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+Reduce optimized SharedTree memory use after sequence edits
```
CraigMacomber: I don't think this is a great description. The name "sequence" is not used in our user facing API surface, and should prabably say reduced or optimized, but both is confusing unless we refer to ForestTypeOptimized directly, which we are not when we say "optimized SharedTree"

```suggestion
Optimize memory use of arrays when using ForestTypeOptimized
```

.changeset/tall-tips-teach.md
```diff
@@ -0,0 +1,9 @@
...
+
+`ForestTypeOptimized` now coalesces adjacent uniformly shaped chunks after sequence insertions and removals.
+For sequences of small, uniformly shaped subtrees, such as plain text, this reduces fragmentation and can reduce memory use by approximately 60%.
```
CraigMacomber: "Chunks" "shapes" and "sequences" are not user facing concepts. Also link relevant APIs. Linking relevant APIs can also help ensure you document this in terms of user facing concepts.

```suggestion
[`ForestTypeOptimized`](https://fluidframework.com/docs/api/fluid-framework#foresttypeoptimized-variable) now more efficiently deduplicates structural information for adjacent children in [array nodes](https://fluidframework.com/docs/api/tree/treearraynode-interface) after edits.
For arrays of small, uniformly shaped subtrees, such as [`PlainText`](https://fluidframework.com/docs/api/fluid-framework/plaintext-namespace), this reduces fragmentation and can reduce memory use by approximately 60%.
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +645,143 @@ export function splitFieldAtIndex(
...
+ *
+ * The per-chunk cap ({@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax}) intentionally
+ * matches the threshold {@link splitFieldAtIndex} uses when bisecting a chunk: if coalescing
+ * produced chunks larger than that threshold, a subsequent split would immediately re-divide
+ * them, so matching the two keeps repeated split/coalesce cycles from oscillating.
+ *
+ * @param chunks - The chunks array, modified in place.
+ * @param policy - The {@link ChunkPolicy} supplying the per-chunk cap.
+ * @param range - Half-open `[start, end)` sub-range of `chunks` (by chunk index) to consider for
+ * merging. Defaults to the whole array. `end` must not exceed `chunks.length`.
```
CraigMacomber: Since this can coalesce within the whole array by default, calling out the sub-range case in the middle of this without saying what its an alternative complicates things. Also lets simplify the top level description and leave the details to the remarks and params sections.

We should also move implementation details into the function itself as implementation comments, or as privateRemarks.

```suggestion
/**
 * Coalesce adjacent small same-shape {@link UniformChunk}s into larger {@link UniformChunk}s.
 *
 * @param chunks - The chunks array, modified in place.
 * @param policy - The {@link ChunkPolicy} supplying the per-chunk cap.
 * @param range - Half-open `[start, end)` sub-range of `chunks` (by chunk index) to consider for
 * merging. Defaults to the whole array. `end` must not exceed `chunks.length`.
 *
 * @remarks
 * Size capped at {@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax} top-level nodes per chunk.
 * @privateRemarks
 * Walks the range from left to right, attempting to merge each chunk with its left neighbor via
 * {@link tryCoalesceUniformChunks}. Performs at most one in-place `splice` on `chunks`. Non-mergeable
 * chunks are passed through unchanged.
 *
 * The per-chunk cap ({@link ChunkPolicy.uniformChunkNodeCountDynamicTargetMax}) intentionally
 * matches the threshold {@link splitFieldAtIndex} uses when bisecting a chunk: if coalescing
 * produced chunks larger than that threshold, a subsequent split would immediately re-divide
 * them, so matching the two keeps repeated split/coalesce cycles from oscillating.
 */
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +645,143 @@ export function splitFieldAtIndex(
...
+	policy: ChunkPolicy,
+	range?: IndexRange,
+): void {
+	const rangeStart = range?.start ?? 0;
+	const rangeEnd = range?.end ?? chunks.length;
+	debugAssert(
+		() =>
+			(rangeStart <= rangeEnd && rangeEnd <= chunks.length) ||
+			"coalesceUniformChunks: invalid range",
+	);
```
CraigMacomber: This seems like a common thing to do (convert an optional range, and a length into either a valid range or an error).

Do we already have shared code for this? If not, I think making this into its own function would be good (it can live next to IndexRange) as that would be trivial to test and nice to share in the future.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +645,143 @@ export function splitFieldAtIndex(
...
+	if (rangeEnd - rangeStart < 2) {
+		return;
```
CraigMacomber, line 679:
```suggestion
		// Zero and 1 chunks are common cases, so as an optimization we return early as there is never anything to do for them.
		return;
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +645,143 @@ export function splitFieldAtIndex(
...
+	}
+
+	// Build the resulting chunk list for [rangeStart, rangeEnd). Passthrough chunks keep their
+	// identity; merged pairs are replaced by the chunk returned from `tryCoalesceUniformChunks`,
+	// which also handles ref-count bookkeeping for chunks that drop out of the array.
```
CraigMacomber: SInce this is an implementation comment, not a comment on some abstraction, its more important to document why this works this way here, and the high-level approach to how its implemented than what is being done exactly, since what is being done is covered by the actual code.

```suggestion
	// Updating `chunks` as we go could incur a lot of overhead if there are a lot of chunks after the location we are editing,
	// so we instead build up this separate array which is spliced over the selected range.
	// As we traverse, ownership of the chunks (tracked via the refcounts) is moved to this array.
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +645,143 @@ export function splitFieldAtIndex(
...
+	const result: TreeChunk[] = [];
+	let mutated = false;
+	for (let chunkIndex = rangeStart; chunkIndex < rangeEnd; chunkIndex++) {
+		const current = chunks[chunkIndex] ?? oob();
+		if (result.length === 0) {
+			result.push(current);
+			continue;
+		}
+		const previous = result[result.length - 1] ?? oob();
```
CraigMacomber: I think with our client 3.0 update to use es2022, we can now use:

 ```suggestion
		const previous = result.at(-1) ?? oob();
```

I think that is cleaner.

Comment (CraigMacomber) [reply packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts]: I can't get github to let me put correct indentation in there, so I guess that will cause a formatting error if you apply it.

File: packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts [hunk 38]
```diff
@@ -641,6 +645,143 @@ export function splitFieldAtIndex(
...
+	// identity; merged pairs are replaced by the chunk returned from `tryCoalesceUniformChunks`,
+	// which also handles ref-count bookkeeping for chunks that drop out of the array.
+	const result: TreeChunk[] = [];
+	let mutated = false;
+	for (let chunkIndex = rangeStart; chunkIndex < rangeEnd; chunkIndex++) {
+		const current = chunks[chunkIndex] ?? oob();
+		if (result.length === 0) {
+			result.push(current);
+			continue;
+		}
```
CraigMacomber: This approach adds overhead to each loop iteration, and makes the code a bit harder to read. I think it would be better to:

```suggestion
	const result: TreeChunk[] = [chunks[rangeStart] ?? oob()];
	let mutated = false;
	for (let chunkIndex = rangeStart + 1; chunkIndex < rangeEnd; chunkIndex++) {
		const current = chunks[chunkIndex] ?? oob();
```

That takes advantage of the early return above which ensure we never go through the loop just once.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -641,6 +645,143 @@ export function splitFieldAtIndex(
...
+		if (coalesced === undefined) {
+			result.push(current);
+		} else {
+			result[result.length - 1] = coalesced;
+			mutated = true;
+		}
+	}
+
+	if (mutated) {
+		chunks.splice(rangeStart, rangeEnd - rangeStart, ...result);
```
CraigMacomber: This could actually be an issue: if result has thousands of items in it, this could hit runtime limits on argument counts. Not sure we have a nice alternative, but might be worth consideration, and maybe a comment.

daesunp: Updated to use a loop for replacements by index, and does a delete-only splice. Also added regression test

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated with suggestion

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated with suggestion

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated with suggestion

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated with suggestion

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated to use `validateIndexRange`

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: Updated with suggestion

.changeset/tall-tips-teach.md
daesunp: Updated with suggestion

.changeset/tall-tips-teach.md
daesunp: Updated with suggestion

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
```diff
@@ -703,7 +697,10 @@ export function coalesceUniformChunks(
 	}
 
 	if (mutated) {
-		chunks.splice(rangeStart, rangeEnd - rangeStart, ...result);
+		for (let resultIndex = 0; resultIndex < result.length; resultIndex++) {
+			chunks[rangeStart + resultIndex] = result[resultIndex] ?? oob();
+		}
+		chunks.splice(rangeStart + result.length, rangeEnd - rangeStart - result.length);
```
CraigMacomber: I think this is just a reimplementation of splice which avoids the argument spread problem right? If so, would be good to move it to a function in out package level utils, and include a test suite for it (should be easy to test it really well in not very much code).

Also come comments in the implementation would be good. It's not clear to me exactly that that loop is doing from a quick glance. Looks like its copying result over some subrange of chunks. Seems like that might corrupt things if the result range is longer than the rangeEnd - rangeStart? If its relying on that, maybe this isn't a general splice replacement, but should very much have some more comments and an assert, and probably still be factored out into a generic testing function, just one in this file instead of at the package level.

I imagine splice is much faster than this when its valid to do a splice: a simple perf tests comparing the two would be nice. Depending on the results we might want to default to splice unless result.length is larger than some threshold. If going that route, writing and using a general splice alternative with that fast path might make more sense again.

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
```diff
@@ -1086,4 +1097,297 @@ describe("chunkTree", () => {
...
+
+			coalesceUniformChunks(field, policy, { start: 0, end: 2 });
+
+			assert.equal(field.length, 3);
+			assertNumbersChunk(field[0], [1, 2]);
+			assertNumbersChunk(field[1], [3]);
+			assertNumbersChunk(field[2], [4]);
+		});
+
+		it("supports coalescing fields larger than the function argument limit", () => {
```
CraigMacomber, line 1339: A comment about why this test has value would be good.


```suggestion
		// Its tempting to use splice with a spread in the implementation,
		// which can hit the argument limit for long sequences of unmergeable chunks
		// This ensures we go not regress our support for such cases
		it("supports coalescing fields larger than the function argument limit", () => {
```

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
```diff
@@ -1086,4 +1097,297 @@ describe("chunkTree", () => {
...
+			const blocker = new BasicChunk(numberType, new Map(), 99);
```
CraigMacomber, line 1340: Some comments in this code for why this test sets up its chunks this way, and how that ensures this tests the case we care about would be good.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkTree.ts
daesunp: moved to a separate function `replaceArrayRange`, and added perf tests to determine threshold to use splice vs the argument limit safe method. Added unit tests as well.

Set the threshold to below 500, based on these perf test results:

| replacement items | native `splice` | argument-safe path | faster strategy |
|---:|---:|---:|---|
| 100 | 1.18 us | 1.28 us | native `splice` |
| 250 | 2.23 us | 3.25 us | native `splice` |
| 500 | 3.94 us | 3.55 us | argument-safe path |
| 750 | 5.58 us | 4.55 us | argument-safe path |
| 1,000 | 8.59 us | 3.85 us | argument-safe path |
| 10,000 | 64.16 us | 15.06 us | argument-safe path |

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
daesunp: Updated with suggestion

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkTree.spec.ts
daesunp: Added comment explaining only first 2 chunks are mergable, while remaining entries are non-uniform blockers, which keeps the replacement array above the argument limit while still requiring coalescing to update the field.

packages/dds/tree/src/util/arrayUtilities.ts
```diff
@@ -97,6 +97,57 @@ export interface IndexRange {
...
+	}
+	array.copyWithin(startIndex + replacementItems.length, endIndex, originalLength);
+	array.length = newLength;
+
+	for (
+		let replacementIndex = 0;
+		replacementIndex < replacementItems.length;
+		replacementIndex++
+	) {
+		array[startIndex + replacementIndex] = replacementItems[replacementIndex] as T;
```
CraigMacomber: The type cast here isn't ideal, and could use a comment noting why its safe, and why we didn't pick one of the more preferred options. Maybe something like:

```suggestion
		// Index is known to be in range due to bounds on replacementIndex, so this cast to remove undefined is safe.
		// ?? oob() must not be used here as it would mishandle values in the array which are null or undefined.
		// `for ... of` over `array.entries()` is cleaner but was measure but was measured to be roughly ???? times slower.
		array[startIndex + replacementIndex] = replacementItems[replacementIndex] as T;
```

Of course if "`for ... of` over `array.entries()`" isn't slower, just use that.

packages/dds/tree/src/util/arrayUtilities.ts
```diff
@@ -97,6 +97,57 @@ export interface IndexRange {
...
+	const newLength = originalLength + replacementItems.length - replacedLength;
+
+	// Extend before moving the suffix when growing; truncate after moving it when shrinking.
+	if (newLength > originalLength) {
+		array.length = newLength;
+	}
+	array.copyWithin(startIndex + replacementItems.length, endIndex, originalLength);
+	array.length = newLength;
+
+	for (
```
CraigMacomber: all the other high level parts of this have a nice line comment (which is great btw), so lets match:


```suggestion

	// copy the items after the splice to their new location
	array.copyWithin(startIndex + replacementItems.length, endIndex, originalLength);
	array.length = newLength;

	// overwrite the old copies (from above) with the inserted content, completing the splice
	for (
```

packages/dds/tree/src/test/util/arrayUtilities.spec.ts
```diff
@@ -169,6 +170,83 @@ describe("arrayUtilities unit tests", () => {
...
+		const cases: readonly ReplaceCase[] = [
+			["inserts into an empty range", [0, 3], 1, 1, [1, 2]],
+			["deletes a range", [0, 1, 2, 3], 1, 3, []],
+			["grows a middle range", [0, 1, 4], 1, 2, [1, 2, 3]],
+			["shrinks a middle range", [0, 1, 2, 3, 4], 1, 4, [9]],
+			["replaces the whole array", [0, 1, 2], 0, 3, [3, 4]],
+			["is a no-op for two empty ranges", [0, 1, 2], 1, 1, []],
+		];
+
+		for (const [name, input, start, end, replacement] of cases) {
```
CraigMacomber: while I really like your testing approach here (nice table driven testing, having a type for ReplaceCase etc) , all of these cases are implemented using array.splice, then tested against array.splice, which is not very useful.

Either you need to factor out the non array.splice fallback for large data into its own function which you test this way (which would also make benchmarking it directly easier, which is useful for reevaluating the size cutoff), or you need to find some other approach.

PR #27388, opened 2026-05-26
ci(client): remove CJS compile to run ESM tests

jason-ha:
Leverage Node 22 https://nodejs.org/learn/typescript/run-natively mode to avoid need to compile compatOptions.ts for CJS to run ESM tests.

This addresses pipeline failure from recent scoped build improvements in #27265.

### DIFF
--- a/packages/test/test-version-utils/mocharc-common.cjs
+++ b/packages/test/test-version-utils/mocharc-common.cjs
@@ -4,6 +4,7 @@
  */
 
 "use strict";
+// Note: this uses Node 22+'s native TS support. See https://nodejs.org/learn/typescript/run-natively.
 const options = require("./src/compatOptions.ts");
 const getFluidTestMochaConfig = require("@fluid-internal/mocha-test-setup/mocharc-common");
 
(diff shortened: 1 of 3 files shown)


packages/test/test-version-utils/mocharc-common.cjs
```diff
@@ -4,7 +4,7 @@
  */
 
 "use strict";
-const options = require("./dist/compatOptions.js");
+const options = require("./src/compatOptions.ts");
```
CraigMacomber, line 8: A comment on this line noting the new and usual pattern would be good, but with the larger cleanup coming, no need to block on it.

CraigMacomber (approved):

CraigMacomber:
Looks like that really does fix it: that's a very minimal fix :)

packages/test/test-version-utils/mocharc-common.cjs
jason-ha: Well, I ran the perf benchmarks pipeline just to be sure and there was a test failure. This fix is good, but a test case failed. I am not sure how to get around that to be allowed to merge. So, might be making a change anyway.
LMK if you know how to get past failed pipeline showing in PR status.

Abe27342 (approved):

PR #27389, opened 2026-05-26
Add Claims DDS

kian-thompson:
## Summary

Adds a new `@fluidframework/claims` DDS implementing first-writer-wins claim semantics with compare-and-swap (CAS) support. All exports are `@internal` for now.

### Key changes

- **New `Claims` DDS** — `trySetClaim(key, value)` returns `Accepted`, `AlreadyClaimed`, or `Pending` (with a promise). Emits typed `"claimed"` events with key.
- **Compare-and-swap** — optional `expectedValue` parameter enables CAS semantics: claim succeeds only if the current value matches `expectedValue`. Omitting it preserves write-once behavior.
- **Fuzz/stress tests** — 200-seed fuzz suite (default + rebasing configs) using `createDDSFuzzSuite`, covering concurrent claims, CAS, and read operations with consistency validation.
- **Local-server-stress integration** — exports `baseClaimsModel` from `./internal/test` and registers it in the local-server-stress-tests `ddsModelMap`.

### Why a new DDS?
 
 Our long-term direction is consolidating DDSes into `SharedTree`, but `Claims` is strictly internal — it helps implement "aliasing" functionality at the dataobject/datastore level and is not intended for direct external use. It will become available to consumers via a scoped API surface.
 
 A DDS was chosen because:
 
 - **New op types are required.** Claim semantics (first-writer-wins, compare-and-set) need dedicated ops. Encapsulating these in a self-contained DDS is a natural fit.
 - **Avoids runtime complexity.** Adding new op handling across various runtime layers (container, data store, etc.) would overcomplicate those components with concerns that are better isolated.
 - **Clean separation.** A dedicated DDS keeps the claim logic self-contained — easier to reason about, test, and maintain without cross-cutting changes to the runtime.

### What's NOT in this PR

- No aqueduct/PureDataObject integration (deferred to a followup PR per reviewer feedback)
- Example package to follow separately

```mermaid
flowchart TD
    A["trySetClaim(key, value)"] --> B{Key already\ncommitted?}

    B -->|Yes| C["AlreadyClaimed"]

    B -->|No| D{Detached?}

    D -->|Yes| E["Accepted"]

    D -->|No| F["Pending { promise }"]

    F --> G{Op roundtrips}
    G -->|Won| H["Accepted"]
    G -->|Lost| I["AlreadyClaimed"]
    G -->|Cancelled / Disposed| J["Aborted"]

    style C fill:#f9e79f
    style E fill:#abebc6
    style H fill:#abebc6
    style I fill:#f9e79f
    style J fill:#f5b7b1
```

[AB#74045](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/74045)


packages/framework/aqueduct/api-report/aqueduct.legacy.beta.api.md
```diff
@@ -94,10 +117,11 @@ export interface IDataObjectProps<I extends DataObjectTypes = DataObjectTypes> {
 }
 
 // @beta @legacy
-export abstract class PureDataObject<I extends DataObjectTypes = DataObjectTypes> extends TypedEventEmitter<I["Events"] & IEvent> implements IFluidLoadable, IProvideFluidHandle {
+export abstract class PureDataObject<I extends DataObjectTypes = DataObjectTypes, TClaimValue = unknown> extends TypedEventEmitter<I["Events"] & IEvent> implements IFluidLoadable, IProvideFluidHandle {
```
ChumpChief: Two main thoughts on integration:
1. I suspect we'll eventually want to consider integrating/exposing this through the datastore runtime (thereby giving aliasing parity at the datastore runtime layer as compared to the container runtime layer), rather than the aqueduct layer.  Probably a discussion to have but worth thinking about pros/cons of each.
    * Either way - I'd probably recommend your first PR be just the DDS plus unit tests, and then take integration as a followup.  Just from the perspective of work staging and throughput.
2. I strongly recommend writing an example/ package that leverages the new DDS as a concrete and speedy way to experiment with the usage patterns and get a feel for the ergonomics of its use without needing to worry about legacy API impact/compat/rollout.
    * It can be a toy/fake scenario (e.g. two clients trying to write their name into a field), no need to devise a complex real-world app.  This can still help highlight if we've made good choices around eventing, promises, etc.
    * You might even consider doing it as a "consensus-comparison" example comparing against CRC, similar to how the task-selection example compares OldestClientObserver vs. TaskManager or tree-comparison compares legacy SharedTree against new SharedTree.
    * Totally OK if this is a followup PR, but I do think it will be useful.

packages/dds/claims/internal.d.ts
ChumpChief, line 1: This file seems unusual, probably don't want it?

packages/dds/claims/src/index.ts
ChumpChief, line 1: Package is missing a README?

packages/dds/claims/src/claimsFactory.ts
```diff
@@ -0,0 +1,88 @@
...
+
+	public create(document: IFluidDataStoreRuntime, id: string): ISharedClaims {
+		const sharedClaims = new SharedClaimsImpl(id, document, this.attributes);
+		sharedClaims.initializeLocal();
+		return sharedClaims;
+	}
+}
+
+/**
+ * A distributed data structure providing first-writer-wins claim semantics.
```
ChumpChief, line 62: For naming, I'd suggest dropping "Shared", we don't typically include it anymore and it doesn't really add anything.

We don't have to belabor naming yet (fine to defer until we're about to expose for customer use, and options might become clearer as the functionality is iterated on) but I'd be curious to know how you chose "Claims" as the name.

packages/dds/shared-claims/src/sharedClaimsFactory.ts
```diff
@@ -0,0 +1,88 @@
...
+ * ### Creation
+ *
+ * ```typescript
+ * const claims = SharedClaims.create(this.runtime, id);
+ * ```
+ *
+ * ### Usage
+ *
+ * ```typescript
+ * const result = await claims.trySetClaim("singleton-component", componentHandle);
```
ChumpChief: This is a good example of a thing to experiment with in an examples/ package - whether a promise-based API adds any value to make it worth the extra effort in implementation (e.g. pending claim tracking, resolve, abort) as compared to fire-and-forget + events.

This is why I made [`.set()` be `void` for `PactMap`](https://github.com/microsoft/FluidFramework/blob/main/packages/dds/pact-map/src/interfaces.ts#L82) - to correctly use the DDS the caller must comprehend that they risk a race against other clients.  Through experimentation I found it easier to separately fire-and-forget plus listen for any incoming ops (including your own) rather than try to reconcile an await callstack against an event handler for remote ops (AB#1602).

Plus it's not too hard for a customer to Promise-ify an event-based API themselves if they really feel they have a need.

packages/dds/shared-claims/src/interfaces.ts
```diff
@@ -0,0 +1,130 @@
...
+ * SharedClaims acts as a scoped aliasing mechanism. Once a key is claimed, it cannot be
+ * overwritten. The `trySetClaim` method returns a synchronous result indicating whether
+ * the key is already claimed, or a pending status with a promise that resolves after the
+ * op roundtrips.
+ *
+ * @internal
+ */
+export interface ISharedClaims<T = unknown> extends ISharedObject<ISharedClaimsEvents<T>> {
+	/**
+	 * Attempts to claim a key with the given value. If the key is already claimed,
```
ChumpChief: We should give this more of a CAS semantic to be more generally useful (would love to see it completely replace CRC eventually).  At the integration layer (i.e. `dataStoreRuntime.trySetAlias()` or wherever it goes) we can still elect to restrict this to "only write once" as a policy decision.

packages/dds/shared-claims/src/interfaces.ts
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Events emitted by {@link ISharedClaims}.
+ *
+ * @internal
+ */
+export interface ISharedClaimsEvents<T = unknown> extends ISharedObjectEvents {
+	/**
+	 * Notifies when a claim has been accepted for a key.
+	 */
+	(event: "claimed", listener: (key: string, value: T) => void): void;
```
ChumpChief: It's generally best to keep event parameters absolutely minimal if they can just be looked up by the event handler off the DDS directly.


```suggestion
	(event: "claimed", listener: (key: string) => void): void;
```

ChumpChief (commented):
Looks like a good start!  Mostly just high-level feedback on API shape, functionality, staging, exposure -- I didn't dive into the details yet.

anthony-murphy:
adding a new dds is a big deal. i don't think we should group adding to pure data object, and public exposures into the same pr. i think the dds itself, if we go this route, should be committed as internal. it then needs to have a dds stress added, like we do for all dds, then you need to export a model and integrate into local server test

anthony-murphy (dismissed):
blocked on splitting between adding a new dds, and only integrating after a full test suite exists for the new dds

packages/test/local-server-stress-tests/src/test/localServerStress.spec.ts
```diff
@@ -40,8 +40,8 @@ describe("Local Server Stress", () => {
 		},
 		// Minimization is slow with many seeds; use only to minimize specific failing seeds.
 		skipMinimization: true,
-		// Pre-existing DDS bugs: seed 54 (ConsensusOrderedCollection consistency).
-		skip: [54],
+		// Pre-existing DDS bugs: seed 54 (ConsensusOrderedCollection), seeds 0 and 92 (Matrix).
+		skip: [0, 54, 92],
```
kian-thompson, line 44: @anthony-murphy @ChumpChief This was a worrying one, but I triple checked and these failures aren't caused by the new DDS (I copied the seed json and replayed strictly against main). I had copilot go all night to find root causes for each and fix locally, I'll put some of the fixes in a separate PR (might skip the ConsensusOrderedCollection fix). 

**Seed 92 — Matrix (SharedMatrix consensus update)**
**File:** packages/dds/matrix/src/matrix.ts
**Bug:** When multiple local set ops are pending and an ack arrives, the consensus value on remaining pending cells wasn't updated. Rolling back remaining local changes restored incorrect server state.
**Fix:** After clearing the acked pending cell, update pendingCell.consensus to the acked value — but only in LWW mode (FWW mode has its own handling).

**Seed 0 — ConsensusOrderedCollection (missed quorum events)**
**File:** packages/dds/ordered-collection/src/consensusOrderedCollection.ts
**Bug:** When a channel is loaded from a snapshot, the constructor's removeMember quorum listener misses departure events that already fired. Catch-up ops can create jobTracking entries for departed clients, causing permanent data inconsistency.
**Fix:** Dual reconciliation:
 - onConnect(): Reconciles for lazy loading (where onConnect fires after pending replay)
 - processMessagesCore() with queueMicrotask: Reconciles for non-lazy loading (where onConnect fires before catch-up ops)

**Seed 54 — fuzzUtils bug**
**File:** packages/dds/ordered-collection/src/test/fuzzUtils.ts
**Fix:** Changed aData.size to aData.size() (method call, not property access).

packages/dds/shared-claims/src/sharedClaimsFactory.ts
kian-thompson: I'm not sure how I feel about a fire-and-forget plus listening to events. Having a returned value and choosing whether to await it or not feels much more explicit without the consumer needing to reason much about current state.

packages/dds/claims/src/claimsFactory.ts
kian-thompson: I went with "Claims" because this DDS is all about claiming keys

packages/dds/claims/src/interfaces.ts
```diff
@@ -0,0 +1,162 @@
...
+	 */
+	compareAndSetClaim(key: string, value: T, expectedValue: T): ClaimResult<T>;
+
+	/**
+	 * Gets the current claimed value for a key, or `undefined` if the key has not been claimed.
+	 *
+	 * @param key - The claim key to look up.
+	 * @returns The claimed value, or `undefined` if unclaimed.
+	 */
+	getClaim(key: string): T | undefined;
```
ChumpChief: If `trySetClaim(key, undefined)` is valid (as under the current implementation, then this kind of begs for a `.has()`-style method too, to distinguish unset from set-as-`undefined`

packages/dds/claims/src/interfaces.ts
```diff
@@ -0,0 +1,162 @@
...
+	 * Only succeeds if the key has not already been claimed.
+	 *
+	 * @param key - The claim key to reserve.
+	 * @param value - The value to associate with the claim.
+	 * @returns The claim result — synchronous for known states, or "Pending" with a promise.
+	 * @throws Will throw a {@link @fluidframework/telemetry-utils#UsageError} if the
+	 * container is not attached and connected, or if a claim for this key is already
+	 * pending locally.
+	 */
+	trySetClaim(key: string, value: T): ClaimResult<T>;
```
ChumpChief, line 133: If you add a `has()`-style method as mentioned in the other comment, this is trivially implemented by the caller with just a `has()` check and should be dropped to be as lean as possible.

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,305 @@
...
+
+	/**
+	 * {@inheritDoc IClaims.compareAndSetClaim}
+	 */
+	public compareAndSetClaim(key: string, value: T, expectedValue: T): ClaimResult<T> {
+		this.guardConnected();
+
+		// CAS: reject if key doesn't exist or current value doesn't match expected.
+		const currentValue = this.claims.get(key);
+		if (!this.claims.has(key) || currentValue !== expectedValue) {
```
ChumpChief: I'd expect a normal case for CAS is to "try to set only if it's unset", so a "has" check is probably not quite right here.

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,305 @@
...
+	createSingleBlobSummary,
+} from "@fluidframework/shared-object-base/internal";
+import { UsageError } from "@fluidframework/telemetry-utils/internal";
+
+import type { ClaimConfirmation, ClaimResult, IClaims, IClaimsEvents } from "./interfaces.js";
+
+/**
+ * Op format for Claims operations.
+ */
+type IClaimOperation<T> =
```
ChumpChief: I don't think you need the bifurcation here on mode - both have the same root semantic of "write only if the value is still what I thought it was", I don't think you need to special-case "I thought it was unset".

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,305 @@
...
+		this.pendingClaims.delete(op.key);
+		pending.resolve({ status: "Aborted" });
+	}
+
+	/**
+	 * {@inheritDoc @fluidframework/shared-object-base#SharedObjectCore.reSubmitCore}
+	 */
+	protected reSubmitCore(content: unknown, _localOpMetadata: unknown): void {
+		this.submitLocalMessage(content);
+	}
```
ChumpChief: This is effectively the default implementation, you can remove it here unless you need to do something to handle disconnect -> reconnect state gracefully.

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,305 @@
...
+	/**
+	 * Guards that the container is attached and connected.
+	 */
+	private guardConnected(): void {
+		if (!this.isAttached()) {
+			throw new UsageError(
+				"Claims cannot be modified while the container is detached or in staging mode",
+			);
+		}
+		if (!this.runtime.connected) {
```
ChumpChief: This is probably a pretty rough user experience for a dev using the API - either they have to wrap all API calls with connected checks or else try/catch blocks to avoid app crashes.

Typical convention in our DDSs today is to permit the modification attempt to go through, and either (1) reconcile state in resubmit (doing op fixup or dropping them) or (2) include enough info in the op to resolve it on the receiver side.

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,305 @@
...
+			mode: "cas",
+			expectedValue,
+		});
+	}
+
+	/**
+	 * Guards that the container is attached and connected.
+	 */
+	private guardConnected(): void {
+		if (!this.isAttached()) {
```
ChumpChief: Detached might be trivial to implement - you just set the value, skipping all checks and return success immediately?

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,305 @@
...
+	): void {
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
+		if (messageEnvelope.type === MessageType.Operation) {
+			const op = messageContent.contents as IClaimOperation<T>;
+
+			assert(op.type === "claim", "Claims: unexpected op type");
+
+			const isAccepted =
+				op.mode === "cas"
+					? this.claims.has(op.key) && this.claims.get(op.key) === op.expectedValue
```
ChumpChief: This probably doesn't work for non-primitives, right?  If the newly-deserialized `op.expectedValue` is an object, array, etc. it won't necessarily have the same object identity as what's stored in `this.claims`?

Regardless, I think it's probably better to avoid direct object comparison and instead lean on sequence number ("Set if the value is the same as it was at sequence number 23" rather than "Set if the value is the same as 'foobar'").  This gives it more of an ETag semantic and therefore a more conservative resolution for e.g. the case that a value is changed A -> B -> A while someone else's A -> C op is in-flight, in case the decision to revert to A was meaningful.  And as a collateral benefit you don't have to worry about object equality or deep comparison or anything.

packages/dds/claims/src/test/fuzzUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+		};
+	}
+
+	async function cas(state: FuzzTestState): Promise<CasOperation> {
+		const key = randomKey(state);
+		const currentValue = (state.client.channel as IClaims<string>).getClaim(key);
+		return {
+			type: "cas",
+			key,
+			newValue: `v${state.random.integer(0, 100)}`,
```
ChumpChief, line 101: I think the top scenario is setting a handle as a value for aliasing purposes, probably want to include that here.

packages/dds/claims/src/test/claims.spec.ts
```diff
@@ -0,0 +1,450 @@
...
+
+	const claims = new Claims(id, dataStoreRuntime, ClaimsFactory.Attributes);
+	claims.connect(services);
+	return claims;
+}
+
+const createLocalClaims = (id: string): Claims =>
+	new Claims(id, new MockFluidDataStoreRuntime(), ClaimsFactory.Attributes);
+
+describe("Claims", () => {
```
ChumpChief, line 36: Tests for setting handles would go well here too.

packages/test/local-server-stress-tests/src/test/localServerStress.spec.ts
ChumpChief: Not sure if there are already issues filed for any of these (is the new 54 failure the same as the old 54 failure?) but probably make sure there's something tracking it - consider putting up a quick fix on the side if you have a good handle on the fixes as suggested in this comment, but totally optional.

ChumpChief (commented):
Bunch of comments - probably there's a good bit you could plagiarize from CRC to resolve some of these.  CRC does some silly stuff (e.g. versions is pretty useless, should probably proactively drop doomed ops on resubmit rather than send them, leftover op format scars from ValueTypes, etc.) but there's still some good ideas in there to consider.

packages/dds/claims/src/claims.ts
kian-thompson: Added a `remark` here and marked it as `@experimental`. This method is not required for the feature I'm working on, so this can be revisited later if needed.

packages/test/local-server-stress-tests/src/test/localServerStress.spec.ts
kian-thompson: Yes, I was planning to do the quick fixes in a follow up PR

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,330 @@
...
+} from "@fluidframework/shared-object-base/internal";
+import { UsageError } from "@fluidframework/telemetry-utils/internal";
+
+import type { ClaimConfirmation, ClaimResult, IClaims, IClaimsEvents } from "./interfaces.js";
+
+/**
+ * Op format for Claims operations.
+ *
+ * Both write-once and compare-and-swap share the same op shape.
+ * `refSeq` of `undefined` means "set only if key is unset" (trySetClaim).
```
ChumpChief: You don't need this special-casing - the client can know authoritatively that the key is unset at e.g. sequence number 123, so sending the op with a refSeq of 123 is equivalent.

packages/dds/claims/src/test/claims.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+		});
+	});
+
+	describe("Summary round-trip", () => {
+		it("Can summarize and load claims", async () => {
+			const containerRuntimeFactory = new MockContainerRuntimeFactory();
+			const claims = createConnectedClaims("claims", containerRuntimeFactory);
+
+			// Set a claim
+			const claimResult = claims.trySetClaim("persistKey", "persistValue");
```
ChumpChief, line 211: Recommend also testing round-trip of other types, esp. those that serialize differently (e.g. `undefined`, `null`, number, handles, object, array).

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,330 @@
...
+	 */
+	public has(key: string): boolean {
+		return this.claims.has(key);
+	}
+
+	// #region SharedObject overrides
+
+	protected summarizeCore(serializer: IFluidSerializer): ISummaryTreeWithStats {
+		const entries = [...this.claims.entries()].map(
+			([key, entry]) => [key, entry.value, entry.sequenceNumber] as [string, T, number],
```
ChumpChief: Prefer a dictionary entry rather than a tuple to make future changes simpler if needed.


```suggestion
			([key, entry]) => { k: key, v: entry.value, s: entry.sequenceNumber },
```

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,330 @@
...
+		const blob = await storage.readBlob(snapshotFileName);
+		const content = new TextDecoder().decode(blob);
+		const entries = this.serializer.parse(content) as [string, T, number][];
+		for (const [key, value, sequenceNumber] of entries) {
+			this.claims.set(key, { value, sequenceNumber });
+		}
+	}
+
+	protected initializeLocalCore(): void {}
+
```
ChumpChief: Has a default impl
```suggestion
```

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,330 @@
...
+		if (messageEnvelope.type === MessageType.Operation) {
+			const op = messageContent.contents as IClaimOperation<T>;
+
+			assert(op.type === "claim", "Claims: unexpected op type");
+
+			const entry = this.claims.get(op.key);
+			const isAccepted =
+				op.refSeq === undefined
+					? entry === undefined // trySetClaim: key must not exist
+					: op.refSeq === entry?.sequenceNumber; // CAS: writer saw latest
```
ChumpChief: With non-optional refSeq then this is just
```suggestion
			const isAccepted = entry === undefined || op.refSeq >= entry.sequenceNumber;
```

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,330 @@
...
+
+		this.submitLocalMessage(op);
+
+		return { status: "Pending", promise };
+	}
+
+	/**
+	 * {@inheritDoc IClaims.getClaim}
+	 */
+	public getClaim(key: string): T | undefined {
```
ChumpChief: Align naming with "has" (or alternatively align "has" -> "hasClaim", but probably prefer the terser form).


```suggestion
	public get(key: string): T | undefined {
```

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,330 @@
...
+
+	/**
+	 * {@inheritDoc IClaims.compareAndSetClaim}
+	 *
+	 * @experimental
+	 */
+	public compareAndSetClaim(
+		key: string,
+		value: T,
+		expectedValue: T | undefined,
```
ChumpChief: You should use an implicit expectedValue here to easily dodge object comparison issues.  It also removes a little boilerplate from the caller-side too:

Current:
```ts
const currentVal = claims.get(key);
if (!someEqualityComparison(currentVal, desiredVal)) {
    claims.compareAndSetClaim(key, desiredVal, currentVal);
}
```

Impilcit:
```ts
const currentVal = claims.get(key);
if (!someEqualityComparison(currentVal, desiredVal)) {
    claims.compareAndSetClaim(key, desiredVal);
}
```

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,330 @@
...
+			this.claims.set(key, { value, sequenceNumber: 0 });
+			return { status: "Accepted", currentValue: value };
+		}
+
+		return this.submitClaim(key, value, {
+			type: "claim",
+			key,
+			value,
+			refSeq: undefined,
+		});
```
ChumpChief: If you want to keep this method is fine but we should unify the op format and ensure CAS is working right.  Recommend replacing this with just a call to CAS then so you only have to implement once:


```suggestion
		return this.compareAndSetClaim(key, value);
```

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,330 @@
...
+	 * Shared submit logic for both write-once and CAS operations.
+	 */
+	private submitClaim(key: string, value: T, op: IClaimOperation<T>): ClaimResult<T> {
+		// If there is already a pending local claim for this key, reject.
+		if (this.pendingClaims.has(key)) {
+			throw new UsageError(`Claims: a claim for key "${key}" is already pending locally`);
+		}
+
+		const promise = new Promise<ClaimConfirmation<T>>((resolve) => {
+			this.pendingClaims.set(key, { value, resolve });
```
ChumpChief, line 147: I'd still suggest dropping the promise return, but if you stick with it we should avoid smuggling the resolveFunc out of the executor (same reasons that Deferred is not a good pattern, losing code locality/ownership of promise resolution).  This exacerbates as more reasons for resolution are added.

I typically like using an `internalEvents` emitter to keep a clean separation of concerns - the Promise's entire lifecycle is documented locally in the executor, and the emit locations don't need to maintain knowledge of all the downstream effects.  ConsensusRegisterCollection and BlobManager are good examples.  Also lets you drop the pendingClaims member.

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,330 @@
...
+	resolve: (result: ClaimConfirmation<T>) => void;
+}
+
+const snapshotFileName = "header";
+
+/**
+ * {@inheritDoc IClaims}
+ * @internal
+ */
+export class Claims<T = unknown> extends SharedObject<IClaimsEvents> implements IClaims<T> {
```
ChumpChief: API council has discussed moving public events to a member (rather than class inheritance, and also using the new emitter type and interfaces), it's probably a good idea to get ahead of that here to avoid a breaking change later.

```suggestion
export class Claims<T = unknown> extends SharedObject implements IClaims<T> {
	private readonly _events = createEmitter<IClaimsEvents>();
	public get events(): Listenable<IClaimsEvents> {
		return this._events;
	}
```

packages/dds/claims/src/claims.ts
ChumpChief: Fine to leave marked as `@experimental` or even make it `private` as an implementation detail of `trySetClaim` initially (see other comment), but we need to resolve anything that puts eventual consistency at risk.  This should just be a refSeq check though I think.

ChumpChief (commented):
Submitting another round - still need to look at resubmit/rollback a bit more but wanted to get this to you

packages/dds/claims/src/claims.ts
kian-thompson: But then this DDS has to reach to the DeltaManager to get the current sequence number. I'm not a huge fan of that and would rather it be self-contained.

Edit: changes `refSeq` to be required and use `IDeltaManager.lastSequenceNumber`

packages/dds/claims/src/claims.ts
kian-thompson: I'm a fan of the current return type. It's explicit and is directly tied to each call, whereas events are not. Let's discuss over a call.

packages/dds/claims/internal.d.ts
ChumpChief: I see this comment keeps getting resolved but I still don't understand what this file is doing :) Could you please explain?  I'd recommend in general when resolving feedback to add a response with the resolution (i.e. "changed in latest" or otherwise explaining why rejected or an alternative was selected), this makes review faster because the reviewer doesn't need to scan through all their prior feedback, deducing whether their feedback was taken and/or what changes were made in response.

If not needed it should be deleted, if needed then add an inline comment in this file explaining why it's deviating from the other DDS package structures.

packages/dds/claims/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Compiled TypeScript and CSS
+dist
+lib
+
+# specialized Node10 support
+!internal.d.ts
```
ChumpChief, line 6: If deleted then this also goes

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,309 @@
...
+			key,
+			value,
+			refSeq: entry?.sequenceNumber ?? this.deltaManager.lastSequenceNumber,
+		});
+	}
+
+	/**
+	 * Shared submit logic for both write-once and CAS operations.
+	 */
+	private submitClaim(key: string, value: T, op: IClaimOperation<T>): ClaimResult<T> {
```
ChumpChief, line 140: Nit: this now has a single call-site, consider collapsing into the call-site.  Optional

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,309 @@
...
+
+	/**
+	 * Resolves the deferred promise for a pending claim.
+	 *
+	 * @remarks
+	 * The pending entry may have a no-op resolve for stashed ops re-submitted
+	 * during rehydration, or may be absent in unexpected edge cases.
+	 */
+	private resolvePending(key: string, isWinner: boolean): void {
+		const pending = this.pendingClaims.get(key);
```
ChumpChief, line 239: I'd still recommend doing this via internalEvents as mentioned in the other feedback, but optional

packages/dds/claims/src/claims.ts
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	protected applyStashedOp(content: unknown): void {
+		const op = content as IClaimOperation<T>;
+		assert(op.type === "claim", "Claims: only claim ops should be stashed");
+		// Track stashed ops as pending so the key is guarded against duplicate
+		// trySetClaim/compareAndSetClaim calls and handles are visited by GC.
+		// No external caller awaits this promise, so resolve is a no-op.
+		this.pendingClaims.set(op.key, { value: op.value, resolve: () => {} });
+		this.submitLocalMessage(op);
```
ChumpChief, line 305: Before submitting the op, would be good to check locally if we believe it has any hope of succeeding.  I.e. if by the time we're applying it the current value already has a more-recent sequenceNumber then there's no point in submitting the op.

This is also true (and probably more commonly occurring) in the resubmit flow.

Optional for this PR but a good improvement to include in a followup.

ChumpChief (approved):
Thank you for making the updates; wire format, summary format, and eventual consistency are looking good now.  More comments here but OK to consider them in a followup.

PR #27394, opened 2026-05-27
bug fix for chunked forest enterNode

daesunp:
## Description

Fixes `ChunkedForest`'s visitor `enterNode` when a field's `chunks` array contains multiple multi-node `UniformChunk`s of differing sizes. The loop that walks chunks to find the target index reassigned its `chunk` variable at the top of the loop body (before `indexOfChunk++`), so the next iteration's condition check used the prior chunk. 

With all chunks having `topLevelLength === 1` the off-by-one coincidentally produced the correct behavior

### DIFF
--- a/packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
+++ b/packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
@@ -245,12 +245,12 @@ export class ChunkedForest implements IEditableForest {
 				let indexOfChunk = 0;
 				let chunk = chunks[indexOfChunk] ?? oob();
 				while (indexWithinChunk >= chunk.topLevelLength) {
-					chunk = chunks[indexOfChunk] ?? oob();
 					indexWithinChunk -= chunk.topLevelLength;
 					indexOfChunk++;
 					if (indexOfChunk === chunks.length) {
 						fail(0xaf7 /* missing edited node */);
 					}
+					chunk = chunks[indexOfChunk] ?? oob();
 				}
 				let found = chunks[indexOfChunk] ?? oob();
 				if (!(found instanceof BasicChunk)) {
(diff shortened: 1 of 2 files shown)


packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -245,12 +245,12 @@ export class ChunkedForest implements IEditableForest {
 				let indexOfChunk = 0;
 				let chunk = chunks[indexOfChunk] ?? oob();
 				while (indexWithinChunk >= chunk.topLevelLength) {
-					chunk = chunks[indexOfChunk] ?? oob();
```
CraigMacomber, line 248: This does look like a bug to me. Good find, and thanks for including a regression test.

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -207,6 +207,40 @@ describe("ChunkedForest", () => {
...
+			]);
+
+			const visitor = forest.acquireVisitor();
+			visitor.enterField(rootFieldKey);
+			visitor.enterNode(6);
+			visitor.exitNode(6);
+			visitor.exitField(rootFieldKey);
+			visitor.free();
+
+			// enterNode shatters the targeted UC(5) into 5 BasicChunks; the field's chunk
```
CraigMacomber: Please avoid uncommon abbreviations. I assume UC(5) means the uniform chunk with top level length 5?

daesunp: Removed abbreviation

brrichards (approved):
LGTM! Thanks for finding this and the test looks good!

CraigMacomber (approved):

PR #12273, opened 2022-10-05
Local Rollback for Cell

clarenceli-msft:
## Description

Implement local rollback on SharedCell to roll back edits when the orderSequentially callback fails. Please also see local rollback epic #8545 and infra implementation #9345.

### PR Checklist

-   [ ] I have updated the documentation accordingly.
-   [X] I have added tests to cover my changes.
-   [X] All new and existing tests passed.
-   [X] My code follows the code style of this project.
-   [X] I ran the lint checks which produced no new errors nor warnings for my changes.
-   [X] I have checked to ensure there aren't other open Pull Requests for the same update/change.

### Does this introduce a breaking change?

-   [ ] Yes
-   [X] No

### Testing

Call orderSequentially with the config set to enable local rollback, make changes to a SharedCell and throw in the callback.

PR #27403, opened 2026-05-27
feat(cell): implement reSubmitSquashed hook with last-write-wins semantics

anthony-murphy:
## Description

Implements `SharedObjectCore.reSubmitSquashed` on `SharedCell` with last-write-wins semantics: any pending op other than the latest is dropped (its value never reaches the wire), and the final pending op is resubmitted as the cell's tip state.

## Why

The base `reSubmitSquashed` falls back to `reSubmitCore` only while the `Fluid.SharedObject.AllowStagingModeWithoutSquashing` flag is true; otherwise it throws. Explicitly opting in protects this DDS from breaking when the flag is flipped, and pre-lands the cell-specific squash semantics so they are not part of the broader squash PR.

## Notes

Prep for upcoming DDS-wide staging-mode squash work — this PR does not add any new squash machinery, it only implements the hook for `SharedCell`.

### DIFF
--- a/packages/dds/cell/src/cell.ts
+++ b/packages/dds/cell/src/cell.ts
@@ -335,6 +335,23 @@ export class SharedCell<T = any>
 	}
 
 	/**
+	 * Cell is last-write-wins: any pending op other than the latest is superseded.
+	 * We drop superseded ops entirely (no wire emission) and resubmit only the final pending op,
+	 * which by construction represents the cell's tip state.
+	 */
+	protected override reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
+		const cellOpMetadata = localOpMetadata as ICellLocalOpMetadata;
+		const lastPendingMessageId = this.pendingMessageIds[this.pendingMessageIds.length - 1];
+		if (cellOpMetadata.pendingMessageId === lastPendingMessageId) {
+			this.submitLocalMessage(content, localOpMetadata);
+		} else {
+			const index = this.pendingMessageIds.indexOf(cellOpMetadata.pendingMessageId);
+			assert(index !== -1, 0xd01 /* Pending message id missing from queue during squash */);
+			this.pendingMessageIds.splice(index, 1);
+		}
+	}
+
+	/**
 	 * Rollback a local op.
 	 *
 	 * @param content - The operation to rollback.


anthony-murphy:
Closing: this implements 

anthony-murphy:
**Deep Review** has launched the submersible.

- ✅ Mechanical Review
- 👻 **Solution / Problem Space Review** — tracing ghost currents…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:
## Deep Review

*Reviewed commit `47dd067` on 2026-05-27.*

PR is closed and will be folded into the broader staging-mode squash work. Implementation is clean and matches what an independent design would produce, but a behavior-changing op-lifecycle hook ships against an unmodified `cell.spec.ts` — see the inline thread on `cell.ts`. Items below target the follow-up branch.

### Path to Ready

- [ ] Resolve inline threads
- [ ] When this work returns, decide whether the cell-specific LWW test matrix lives in `packages/dds/cell/src/test/cell.spec.ts` or folds into `packages/test/local-server-tests/src/test/stagingMode.spec.ts`

### Context for Reviewers

- The `pendingMessageIds` queue lifecycle (push in `createLocalOpMetadata`, `shift` on ACK, `pop` in `rollback`, now `splice` here) was shaped by review feedback on #12273 — the new `splice(index, 1)` for superseded ops is consistent with that contract (would otherwise leak, since dropped ops never ACK).
- This continues the op-lifecycle alignment work the same author landed in #19518 for `applyStashedOps`; framed as prep for upcoming DDS-wide staging-mode squash work.
- Surface-level review tax from #12273 is already paid: `protected override` ✓, cast named `cellOpMetadata` ✓, queue-leak handled ✓ — reviewer effort belongs on test coverage and the squash/rollback invariant.
- For the follow-up squash PR, suggested reviewers based on file history: **scarlettjlee** (set the `pendingMessageIds` lifecycle rules and gated the prior hook on tests in #12273), **clarenceli-msft** as secondary, plus runtime/staging-mode owners.

- **Squash/rollback runtime contract** — confirm whether `rollback` can ever be invoked against a queue that has been through a squashed resubmit. Current runtime flow (`containerRuntime.ts:3810-3829`, `pendingStateManager.ts:774-831`, `containerRuntime.ts:5012-5145`) does not interleave them — staging discard rolls back, staging commit squash-replays — but this is unwritten. Decide whether the invariant deserves codification in `rollback` or just a comment on `reSubmitSquashed`.
- **Test scope** — decide whether the cell-specific LWW matrix lives in `cell.spec.ts` or in the staging-mode integration suite. Both defensible.

</details>

packages/dds/cell/src/cell.ts
```diff
@@ -334,6 +334,23 @@ export class SharedCell<T = any>
...
+	protected override reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
+		const cellOpMetadata = localOpMetadata as ICellLocalOpMetadata;
+		const lastPendingMessageId = this.pendingMessageIds[this.pendingMessageIds.length - 1];
+		if (cellOpMetadata.pendingMessageId === lastPendingMessageId) {
+			this.submitLocalMessage(content, localOpMetadata);
+		} else {
+			const index = this.pendingMessageIds.indexOf(cellOpMetadata.pendingMessageId);
+			assert(index !== -1, 0xd01 /* Pending message id missing from queue during squash */);
+			this.pendingMessageIds.splice(index, 1);
+		}
```
anthony-murphy, line 351: Deep Review: This new override has three branches — (a) tail match → `submitLocalMessage`; (b) `indexOf` found → `splice(index, 1)` *silent drop, no wire emission*; (c) `index === -1` → tagged assert `0xd01`. None of them are exercised: `cell.spec.ts` is untouched (no matches for `reSubmit|squash|Squash`), existing reconnection coverage at `cell.spec.ts:401-488` only hits ordinary disconnect/reconnect resend, and `packages/test/local-server-tests/src/test/stagingMode.spec.ts:614-669` verifies staging-mode calls `reSubmitSquashed` for a root map, not `SharedCell`'s LWW behavior.

The PR description frames this PR as existing to "pre-land the cell-specific squash semantics so they are not part of the broader squash PR" — i.e. the squash behavior is meant to be validated here. In-area precedent: PR #12273 (Local Rollback for Cell) shipped with tests, and the rollback hook — which mutates the same `pendingMessageIds` queue — was gated on test coverage.

When this work returns, add a focused `describe("reSubmitSquashed", ...)` in `packages/dds/cell/src/test/cell.spec.ts` covering:
- multiple pending `set`s squashed → only the last reaches the wire; final remote value matches the last local set;
- a `set` followed by a `delete` squashed → only the `delete` is emitted;
- ACK ordering after squash — the `pendingMessageIds[0]` invariant in `processCore` still holds for the surviving op;
- the `0xd01` missing-from-queue assert path.

Use `MockContainerRuntime` / `reconnectAndSquash` helpers consistent with sibling DDS test suites.

Two related polish items on the same `pendingMessageIds` contract are worth picking up in the same follow-up: (1) the surviving op's `localOpMetadata.previousValue` reflects the *tail* op, not the pre-batch value — latent, since today's runtime does not interleave `rollback` with squashed resubmit, but worth either a one-line `assert` in `rollback` or a comment on `reSubmitSquashed` codifying the invariant; (2) `pendingMessageIds` now has four mutation sites (push/shift/pop/splice) with no comment at the declaration (`cell.ts:88`) and no length guard before the tail read here — a one-line enumeration comment plus `assert(this.pendingMessageIds.length >= 1)` at the top of this method would document the contract.

PR #3345, opened 2020-08-26
Add copy button to QuickStart code blocks

skylerjokiel:
Adds a script that will go through a hugo markdown page and add a copy button next to a code block. I'm using this for the Quick Start page only to make it easier for people getting started.

Thanks to @tylerbutler for helping me make this a js file that is re-useable and not just some in-lined script

PR #2538, opened 2020-06-15
Docs/reorg part 3

tylerbutler:

PR #27405, opened 2026-05-27
feat(ink): implement reSubmitSquashed hook (identity transform)

anthony-murphy:
## Description

Implements `SharedObjectCore.reSubmitSquashed` on `Ink` as the identity transform — each pending op (`createStroke` / `stylus` / `clear`) is resubmitted unchanged via `reSubmitCore`.

## Why

Ink ops are append-style: each op records an intentional user action and is never superseded by a later pending op. The base `reSubmitSquashed` falls back to `reSubmitCore` only while the `Fluid.SharedObject.AllowStagingModeWithoutSquashing` flag is true; otherwise it throws. Explicitly opting in protects this DDS from breaking when the flag is flipped.

## Notes

### DIFF
--- a/packages/dds/ink/src/ink.ts
+++ b/packages/dds/ink/src/ink.ts
@@ -260,6 +260,15 @@ export class Ink extends SharedObject<IInkEvents> implements IInk {
 	}
 
 	/**
+	 * Ink ops (createStroke / stylus / clear) are append-style: each op records an intentional
+	 * user action and is never superseded by a later pending op. Squash is therefore the identity
+	 * transform — we just replay the original ops via reSubmitCore.
+	 */
+	protected override reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
+		this.reSubmitCore(content, localOpMetadata);
+	}
+
+	/**
 	 * Update the model for a clear operation.
 	 * @param operation - The operation object
 	 */


anthony-murphy:
Closing: this implements 

anthony-murphy:
## Deep Review

*Reviewed commit `a57ed0e` on 2026-05-27.*

Not ready for sign-off. The identity `reSubmitSquashed` and its "never superseded by a later pending op" JSDoc rationale are contradicted by Ink's own `clear` op, which destroys all prior staged `createStroke`/`stylus` content — exactly the insert-then-remove case the squash contract names as the canonical thing to elide. Eventual convergence bounds the user-visible damage, but the rationale shipped in the diff is provably false. Detail inline. (Noted the PR is already closed with intent to revisit during the actual squash work — option (c) below aligns with that.)

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- Ink area is essentially frozen — 6-month churn of 1 on `ink.ts`, no inline comments on `ink.ts` in any of the related PRs (#210, #629, #2538, #3345). Treat this as DDS-base-class-compliance work, not Ink-domain work; the substantive review surface is the `SharedObject` / staging-mode contract.
- Identity-shape `reSubmitSquashed` precedent already exists in `packages/dds/sequence/src/sequence.ts` (note: `sequence` passes a third `true` argument to `reSubmitCore`; the Ink override does not — minor signature drift between siblings).
- Before `Fluid.SharedObject.AllowStagingModeWithoutSquashing` is flipped off, the other in-tree DDSes (`map`, `directory`, `matrix`, `cell`, `counter`, `register-collection`, `ordered-collection`, `task-manager`, `migration-shim`, `tree`) likely need the same opt-in audit — sibling-PR question, not a blocker here.
- PR #629 (ChumpChief, 2019-11-21) framed Ink ops as "op concealment" to protect future op-pattern changes. The general append-style framing matches that; the narrower "never superseded" sub-claim is the part that fails for `clear`.

- **Owner sign-off on squash-contract interpretation** — Whoever owns `reSubmitSquashed` and `AllowStagingModeWithoutSquashing` in `packages/dds/shared-object-base/src/sharedObject.ts` (and the contract text in `runtime-definitions/src/stagingMode.ts`, `datastore-definitions/src/channel.ts`) should rule on whether an identity passthrough is acceptable for an "append-style intent" DDS that also exposes a destructive `clear` op. Framework-policy call, not author-resolvable.
- **Imminence of the flag flip** — The risk profile depends entirely on how soon `AllowStagingModeWithoutSquashing` will be turned off in default config. Near-term: the inline concern matters. Distant: deferring (option c) is essentially free.
- **Cross-DDS coverage** — Confirm whether other DDSes also need the opt-in before the flag flip. The pipeline did not survey all sibling DDSes.

</details>

packages/dds/ink/src/ink.ts
```diff
@@ -259,6 +259,15 @@ export class Ink extends SharedObject<IInkEvents> implements IInk {
 		return;
 	}
 
+	/**
+	 * Ink ops (createStroke / stylus / clear) are append-style: each op records an intentional
+	 * user action and is never superseded by a later pending op. Squash is therefore the identity
+	 * transform — we just replay the original ops via reSubmitCore.
+	 */
+	protected override reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
```
anthony-murphy, line 267: Deep Review: The new override is an unconditional identity passthrough, justified by JSDoc claiming Ink ops are "append-style ... never superseded by a later pending op." That claim is provably false for `clear`: `Ink.clear()` (see `ink.ts:173-179` and `:275-277`) submits an `IClearOperation` whose apply step calls `this.inkData.clear()`, destroying every prior stroke. A staged `createStroke` → N×`stylus` → `clear` sequence is exactly the "inserting some content then removing it" case the squash contract names — see `packages/runtime/runtime-definitions/src/stagingMode.ts:14-22` and `packages/runtime/datastore-definitions/src/channel.ts:160-164`. The framework routes `squash=true` to `reSubmitSquashed` and falls back to `reSubmitCore` only while `Fluid.SharedObject.AllowStagingModeWithoutSquashing` is on (`packages/dds/shared-object-base/src/sharedObject.ts:490-500, 663`); otherwise it throws. So the override is asking each DDS to declare *real* squash semantics, not silently passthrough.

Convergence mitigates the impact: the replayed `clear` still removes the replayed strokes on remote peers, so final state converges. But the intermediate `createStroke`/`stylus` events are still broadcast, wasting bandwidth proportional to stylus volume and potentially flickering remote renderers that listen on `"stylus"`/`"clear"` (see `inkCanvas.ts:130-131`).

Pick one before this lands:

1. **Honest-JSDoc path (cheapest)** — Rewrite the JSDoc to drop the "never superseded" claim and document a deliberate trade-off, e.g. "Ink preserves the full op stream through squash because each op records intentional user input; the runtime accepts the resulting redundancy across staged `clear` for the sake of user-intent fidelity." Keep the body unchanged.
2. **Real-squash path** — In `reSubmitSquashed`, walk the pending op queue and drop staged `createStroke`/`stylus` ops whose target stroke is removed by a later staged `clear`, then resubmit the residue. Requires access to the staged queue beyond the single `(content, localOpMetadata)` pair — likely a follow-up.
3. **Don't opt in yet** — Defer the override until Ink can satisfy the contract; rely on `AllowStagingModeWithoutSquashing=true` in the interim. Aligns with your closing comment that the real squash work will be revisited later.

If the override stays (paths 1 or 2), add a flag-off (`allowStagingModeWithoutSquashing: false`) regression test in `ink.spec.ts` covering `createStroke` → `stylus` (→ `clear`) under reconnect-with-squash. The existing reconnection tests at `ink.spec.ts:345-488` use the default `MockFluidDataStoreRuntime` and never configure the flag off, so they exercise the base `reSubmitCore` fallback rather than this override.

PR #27406, opened 2026-05-27
feat(pact-map): implement reSubmitSquashed hook (identity transform)

anthony-murphy:
## Description

Implements `SharedObjectCore.reSubmitSquashed` on `PactMap` as the identity transform — each pending op is resubmitted unchanged via `reSubmitCore`.

## Why

`PactMap` implements consensus over proposed values, and `set` already drops a new write early if there is an outstanding pending proposal for the same key, so the squash surface is small in practice. Collapsing pending proposals beyond the existing rule would change observable consensus semantics, so identity is the right squash.

## Notes

### DIFF
--- a/packages/dds/pact-map/src/pactMap.ts
+++ b/packages/dds/pact-map/src/pactMap.ts
@@ -368,6 +368,17 @@ export class PactMapClass<T = unknown>
 	protected onDisconnect(): void {}
 
 	/**
+	 * PactMap implements consensus over proposed values: each set is a proposal that must be
+	 * accepted by all connected clients at the time it was sequenced. Collapsing pending
+	 * proposals would change observable consensus semantics (and PactMap already drops a new
+	 * set early if there is an outstanding pending proposal for the same key, so the squash
+	 * surface is small in practice). Squash on resubmit is the identity transform.
+	 */
+	protected override reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
+		this.reSubmitCore(content, localOpMetadata);
+	}
+
+	/**
 	 * {@inheritDoc @fluidframework/shared-object-base#SharedObjectCore.reSubmitCore}
 	 */
 	protected reSubmitCore(content: unknown, localOpMetadata: unknown): void {


anthony-murphy:
Closing: this implements 

anthony-murphy:
## Deep Review

*Reviewed commit `a928793` on 2026-05-27.*

The 11-line identity `reSubmitSquashed` override on `PactMapClass` is the right shape: mechanical, historical, and solution-space angles all agree, and the override inherits #10449's selective-resubmit filter by delegating straight to `reSubmitCore`. One Tier 3 polish flagged inline — a DDS-level regression test pinning the `reSubmitSquashed === reSubmitCore` contract, which the PR write-up already concedes is a reasonable critique. Noting the PR is closed as draft; this review is for whenever the work is carried into the actual squash effort.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- ChumpChief is the de facto owner of `pactMap.ts` reconnect/resubmit semantics (#10449 established the selective-resubmit filter, #13594 ratified the two-stage `set` pending+acceptance semantics) — tag on review.
- The identity squash inherits #10449's selective-resubmit filter inside `reSubmitCore` (pactMap.ts:384-401): pending `set` is dropped on resubmit if another value is pending or a newer value was accepted. Future squash-related changes must not bypass this.
- The doc-comment rationale ("collapsing pending proposals would change observable consensus semantics") matches #13594's decision to preserve both pending and acceptance resolutions of `set`.
- Sibling consensus DDSes are *not* identity transforms and need separate analysis when `Fluid.SharedObject.AllowStagingModeWithoutSquashing` flips: `TaskManager.reSubmitCore` is conditional on `submitVolunteerOp` (taskManager.ts:676-694), and `ConsensusRegisterCollection` reassigns pending message ids for stashed ops (consensusRegisterCollection.ts:434-440). Out of scope here; belongs to the flag-flip PR.

- **Owner sign-off** — ChumpChief should confirm that identity-squash is the intended shape and that existing `reSubmitCore` reconnect coverage at `pactMap.spec.ts:352-447` is acceptable as the regression backstop until the inline test gap is closed.
- **Rollout coordination** — When `AllowStagingModeWithoutSquashing` flips globally, the flag-flip PR needs matched overrides (or staging) for every remaining consensus-style DDS whose resubmit is not identity — at minimum `TaskManager` and `ConsensusRegisterCollection`. Flag for the follow-up; not actionable here.

</details>

packages/dds/pact-map/src/pactMap.ts
```diff
@@ -367,6 +367,17 @@ export class PactMapClass<T = unknown>
...
+	/**
+	 * PactMap implements consensus over proposed values: each set is a proposal that must be
+	 * accepted by all connected clients at the time it was sequenced. Collapsing pending
+	 * proposals would change observable consensus semantics (and PactMap already drops a new
+	 * set early if there is an outstanding pending proposal for the same key, so the squash
+	 * surface is small in practice). Squash on resubmit is the identity transform.
+	 */
+	protected override reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
+		this.reSubmitCore(content, localOpMetadata);
+	}
```
anthony-murphy, line 379: Deep Review: No DDS-level regression test pins the `reSubmitSquashed === reSubmitCore` identity contract. The override's entire purpose per the PR description is forward compatibility once `Fluid.SharedObject.AllowStagingModeWithoutSquashing` is flipped to false — without the override the base method throws (`shared-object-base/src/sharedObject.ts:490-500`). Existing PactMap reconnect coverage at `pactMap.spec.ts:352-381` (futile pre-offline proposals not resubmitted) and `:383-399`, `:431-447` (still-valid proposals accepted after reconnect) exercises `reSubmitCore`'s consensus semantics but does not route through the squashed-dispatch entry point with the compat flag off, so a future "helpful" change to either method could silently diverge the two paths — a critique the PR write-up itself concedes.

Suggest a small `pactMap.spec.ts` test that (a) drives `reSubmitSquashed` directly with `Fluid.SharedObject.AllowStagingModeWithoutSquashing` disabled to assert it does not throw, and (b) verifies it produces the same submitted op as `reSubmitCore` for a pending `set` and filters a superseded `set` identically. This pins the identity-delegation contract against silent future divergence.

Is there a tracking issue or follow-up planned to add this test before the flag flip ships?

PR #27408, opened 2026-05-27
feat(register-collection): implement reSubmitSquashed hook (identity transform)

anthony-murphy:
## Description

Implements `SharedObjectCore.reSubmitSquashed` on `ConsensusRegisterCollection` as the identity transform — each pending write is replayed in submit order via `reSubmitCore`.

## Why

`ConsensusRegisterCollection` writes participate in server-side consensus, and the version history is exposed via `readVersions()`. Collapsing pending writes would change observable semantics, so identity is the right squash.

## Notes

Updates `register-collection.legacy.beta.api.md` to reflect the new `protected` member. These api-report files are normally generated; if CI's api-extractor disagrees with the change, regenerate via `build:api-reports` and push a fixup.

### DIFF
--- a/packages/dds/register-collection/src/consensusRegisterCollection.ts
+++ b/packages/dds/register-collection/src/consensusRegisterCollection.ts
@@ -275,6 +275,19 @@ export class ConsensusRegisterCollection<T>
 
 	protected onDisconnect(): void {}
 
+	/**
+	 * ConsensusRegisterCollection writes participate in server-side consensus and the version
+	 * history is exposed via readVersions(). Collapsing pending writes would change observable
+	 * semantics, so squash on resubmit is intentionally the identity transform (replay each
+	 * pending write in order via reSubmitCore).
+	 *
+	 * If callers need staging-mode-like behavior with no intermediate version exposure, they
+	 * should hold writes locally and apply only the final value when committing.
+	 */
+	protected override reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
+		this.reSubmitCore(content, localOpMetadata);
+	}
+
 	protected override processMessagesCore(messagesCollection: IRuntimeMessageCollection): void {
 		const { envelope, local, messagesContent } = messagesCollection;
 		for (const messageContent of messagesContent) {
--- a/packages/dds/register-collection/api-report/register-collection.legacy.beta.api.md
+++ b/packages/dds/register-collection/api-report/register-collection.legacy.beta.api.md
@@ -25,6 +25,7 @@ export class ConsensusRegisterCollectionClass<T> extends SharedObject<IConsensus
     read(key: string, readPolicy?: ReadPolicy): T | undefined;
     // (undocumented)
     readVersions(key: string): T[] | undefined;
+    protected reSubmitSquashed(content: unknown, localOpMetadata: unknown): void;
     // @sealed
     protected rollback(content: unknown, localOpMetadata: unknown): void;
     // (undocumented)


anthony-murphy:
Closing: this implements 

anthony-murphy:
## Deep Review

*Reviewed commit `6899559` on 2026-05-27.*

The identity-transform override is the right call — four independent analyses (including two blind solution proposers given only the problem statement) landed on the same one-liner with the same `readVersions`/server-side-FWW justification. One inline thread flags the missing regression test for when this work is revived.

### Path to Ready

- [ ] Resolve inline thread (add a unit test pinning the `readVersions()` order-preservation contract under `reSubmitSquashed`; optionally flip `testSquashResubmit: true` in the CRC fuzz spec)

### Context for Reviewers

- Identity resubmit is sound because CRC defers mutation to ack — pending writes resolve via `localOpMetadata` on ack/rollback, so there's no local data state to compensate for (established in #24734).
- Override style matches the convention standardized in #26526 (CraigMacomber): `protected override` with substantive JSDoc, no `{@inheritDoc}`.
- The api-report hand-edit follows the pattern of every prior behavior-affecting CRC PR (#8754, #24734, #25745, #26526); CI's api-extractor is the authoritative check.
- Once `Fluid.SharedObject.AllowStagingModeWithoutSquashing` flips to false, the base `reSubmitSquashed` throws — this override is what keeps CRC working past that flip.
- Suggested reviewers: **ChumpChief** (primary — owns #24734 rollback/resubmit semantics on this exact class), **CraigMacomber** (override convention, last api-report touch), **agarwal-navin** (SharedObject hook-implementation pattern across DDSes).

- **Needs human judgment** — ChumpChief should confirm the identity-transform interpretation of "squashed resubmit" matches the staging-mode squash design intent from #24734, in particular that no consumer relies on `reSubmitSquashed` collapsing per-key writes for CRC.
- **Cannot be assessed by the pipeline** — runtime behavior of CRC under `AllowStagingModeWithoutSquashing=false`; confirm no other CRC code path would still throw once the flag flips.

</details>

packages/dds/register-collection/src/consensusRegisterCollection.ts
```diff
@@ -275,6 +275,19 @@ export class ConsensusRegisterCollection<T>
...
+	 * history is exposed via readVersions(). Collapsing pending writes would change observable
+	 * semantics, so squash on resubmit is intentionally the identity transform (replay each
+	 * pending write in order via reSubmitCore).
+	 *
+	 * If callers need staging-mode-like behavior with no intermediate version exposure, they
+	 * should hold writes locally and apply only the final value when committing.
+	 */
+	protected override reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
+		this.reSubmitCore(content, localOpMetadata);
+	}
```
anthony-murphy, line 289: Deep Review: The new JSDoc advertises a behavioral contract — "replay each pending write in order … so that `readVersions()` history is preserved under staging-mode squash" — that's novel to CRC and isn't exercised by existing tests, because `reSubmitCore` coverage doesn't enter through the `reSubmitSquashed` hook. This override becomes load-bearing once `Fluid.SharedObject.AllowStagingModeWithoutSquashing` flips to false.

Precedent: PR #24734 ("Enable rollback for ConsensusRegisterCollection") added e2e tests when CRC gained a new lifecycle semantic — the same shape of change as this one.

When this PR is revived as part of the broader squash work, add a focused unit test in `packages/dds/register-collection/src/test/consensusRegisterCollection.spec.ts` that:

1. Drives ≥2 pending writes to the same key through `reSubmitSquashed`.
2. Asserts `readVersions(key)` returns every value in submit order (non-collapse).
3. Asserts `ReadPolicy.LWW` returns the last write.

Optionally flip `testSquashResubmit: true` in `consensusRegisterCollection.fuzz.spec.ts` to get randomized coverage of the new path under the existing harness.

PR #25388, opened 2025-09-04
feat(task-manager): Add rollback support for TaskManager

scottn12:
## Description

This PR adds support for rollback for TaskManager (including across remote ops). 

## Implementation

Since TaskManager is a fully consensus-based DDS, we don't need to do much to support rollback. We mainly do two things:
1. Ensure the op we are trying to rollback exists in the pending data (and cleanup pending data since we will no longer process the op).
2. Use a new `rollbackWatcher` event emitter to ensure we properly resolve promises that were waiting on the pending op to be ack'd.

Note: There are a few miscellaneous changes in this PR that are not directly related to rollback, but were exposed when enabling local server stress tests (and fixed as part of this PR).

## Testing

This PR also adds additional testing:
1. Added rollback specific testing in `taskManager.rollback.spec.ts`
4. Enabled rollback in TaskManager fuzz testing
5. Enabled local server stress tests to use TaskManager

## Misc
[AB#44704](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/44704)

PR #27409, opened 2026-05-27
feat(task-manager): implement reSubmitSquashed hook via existing reSubmitCore

anthony-murphy:
## Description

Implements `SharedObjectCore.reSubmitSquashed` on `TaskManager` by delegating to the existing `reSubmitCore`, which already collapses volunteer+abandon pairs (it was designed for the disconnect case, where the server drops the client from queues).

## Why

The collapse `TaskManager.reSubmitCore` already performs is the correct squash for a staging-mode commit: redundant pending ops are not re-emitted, and a still-pending volunteer is re-emitted to express the final intent.

One known limitation, consistent with TaskManager's overall queue-position-vs-connection-state design: an in-staging abandon for a task the client previously volunteered to (pre-staging) is not propagated through this path, because `reSubmitCore` assumes the server has already removed the client from the queue.

## Notes

### DIFF
--- a/packages/dds/task-manager/src/taskManager.ts
+++ b/packages/dds/task-manager/src/taskManager.ts
@@ -673,6 +673,21 @@ export class TaskManagerClass
 	}
 
 	/**
+	 * TaskManager's reSubmitCore already collapses volunteer+abandon pairs (it was designed for
+	 * the disconnect case, where the server drops the client from queues). The same collapse is
+	 * the correct squash for staging-mode commit: redundant pending ops are not re-emitted, and
+	 * a still-pending volunteer is re-emitted to express the final intent.
+	 *
+	 * One known limitation: an in-staging abandon for a task the client previously volunteered
+	 * to (pre-staging) is not propagated by this path, because reSubmitCore assumes the server
+	 * has already removed the client from the queue. TaskManager's queue position is inherently
+	 * tied to live connection state, so this is consistent with its overall design.
+	 */
+	protected override reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
+		this.reSubmitCore(content, localOpMetadata as number);
+	}
+
+	/**
 	 * Override resubmit core to avoid resubmission on reconnect.  On disconnect we accept our removal from the
 	 * queues, and leave it up to the user to decide whether they want to attempt to re-enter a queue on reconnect.
 	 * However, we do need to update latestPendingOps to account for the ops we will no longer be processing.


anthony-murphy:
Closing: this implements 

anthony-murphy:
## Deep Review

*Reviewed commit `7b06d2e` on 2026-05-27.*

Not ready for sign-off. The PR is already closed for redesign ("will be revisited as part of the actual squash work later"), which matches the analysis: delegating squashed resubmission to the disconnect-shaped `reSubmitCore` silently drops pending `complete` ops, over-emits on `[volunteer, abandon, volunteer]`, and leaves the documented lone-abandon limitation unenforced — all without tests on a file whose recent history (#25388) set spec + fuzz + stress as the bar. Two correctness concerns flagged inline; the same checklist applies to the follow-up PR.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add `taskManager.squash.spec.ts` modelled on `taskManager.rollback.spec.ts` covering volunteer/abandon/complete combinations (lone volunteer, `[v, a]`, `[v, complete]`, lone abandon via pre-staging volunteer, `[v, a, v]`, multi-task interleaving, the documented limitation) and wire a `createSquashFuzzSuite` variant into the existing TaskManager fuzz/stress configs; confirm with scottn12 that the staging path is covered
- [ ] Update the JSDoc on `reSubmitSquashed` to acknowledge the `complete` op type (the current rationale predates #25388, which introduced `complete`)

### Context for Reviewers

- Pending-op symmetry across `volunteer` / `abandon` / `complete` and `latestPendingOps` message-id bookkeeping is the historical hotspot in this file — see #25388 review threads where ChumpChief pressed on the resubmit-not-at-head case "at least for staging mode commit," which is exactly the surface this PR opens.
- Best-positioned reviewers: **scottn12** owns the `complete` op, `completedWatcher`, the disconnect-shaped `reSubmitCore` doc this PR reuses, and the fuzz+stress+spec triad added in #25388 (correctness on the `complete` drop and on test coverage). **ChumpChief** authored the original DDS and the "queue position is tied to live connection state" design from which the disconnect-only `reSubmitCore` derives (design call on whether the documented lone-abandon limitation is acceptable for first staging-mode rollout).

- **Needs human judgment — design call** — Is the documented lone-abandon-of-pre-staging-volunteer limitation acceptable as a first-rollout posture, or must it be implemented before TaskManager opts into squashed submit? ChumpChief is best positioned.
- **Needs human judgment — correctness call** — Does the `complete` op's externally observable semantics (broadcast to all clients, queue clear, `"completed"` event) actually require on-wire delivery at commit, or is there a server-side recovery path that makes silent drop acceptable? scottn12 owns this design.
- **Needs human judgment — test scope** — Is the existing TaskManager fuzz/stress harness (added in #25388) already exercising squash paths in CI, or does staging-mode squash need explicit wiring? scottn12 can confirm.
- **Cannot be assessed by the pipeline** — Whether the `[v, a, v]` over-emission has knock-on effects on queue-position determinism across forks at squash time; needs human runtime verification.

</details>

packages/dds/task-manager/src/taskManager.ts
```diff
@@ -672,6 +672,21 @@ export class TaskManagerClass
...
+	 * the correct squash for staging-mode commit: redundant pending ops are not re-emitted, and
+	 * a still-pending volunteer is re-emitted to express the final intent.
+	 *
+	 * One known limitation: an in-staging abandon for a task the client previously volunteered
+	 * to (pre-staging) is not propagated by this path, because reSubmitCore assumes the server
+	 * has already removed the client from the queue. TaskManager's queue position is inherently
+	 * tied to live connection state, so this is consistent with its overall design.
+	 */
+	protected override reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
+		this.reSubmitCore(content, localOpMetadata as number);
```
anthony-murphy, line 687: Deep Review: Staged `complete` ops are silently dropped on commit.

`reSubmitSquashed` unconditionally delegates to `reSubmitCore`, which splices the matching pending op from `latestPendingOps` but only re-emits when `content.type === "volunteer"` (line 705). That is correct for the disconnect contract (lines 691–693: "we accept our removal from the queues"), but on a staging-mode commit the server has observed none of the staged ops, so a staged `complete` is lost on the wire.

`complete` is how task completion is broadcast: `submitCompleteOp` records a pending `complete` (lines 287–304), attached `complete()` calls into it (lines 588–605), and processing a `complete` op deletes the task queue and emits `"completed"` (lines 193–212). Concrete failing sequence:

1. Client volunteers pre-staging.
2. Client calls `complete(taskId)` while staged.
3. Commit fires `reSubmitSquashed` → `reSubmitCore` splices the pending `complete` and the `type !== "volunteer"` guard at line 705 returns without re-emitting.
4. Other clients never observe completion; the queue never clears; the `"completed"` event never fires.

The new JSDoc (lines 676–684) discusses only volunteer+abandon and never mentions `complete` — that rationale predates #25388 (scottn12, 2025-09-09), which introduced the `complete` op type.

Recommendation: don't delegate squashed resubmission of `complete` to `reSubmitCore`. Either (a) implement staging-specific handling that preserves `complete` as a final intent (splice the pending op and call `submitCompleteOp(content.taskId)` for non-superseded completes while keeping volunteer/abandon collapse), or (b) keep the throwing fallback (`Fluid.SharedObject.AllowStagingModeWithoutSquashing` off) until staged `complete` handling is implemented. Loop in scottn12 (owns the `complete` op and rollback design).

anthony-murphy, line 687: Deep Review: Trivial delegation violates the "single final intent" squash contract on `[volunteer, abandon, volunteer]` and leaves the documented lone-abandon limitation unenforced.

`reSubmitCore` is dispatched per pending op by `SharedObject.reSubmitSquashed` (`shared-object-base/src/sharedObject.ts:653-666` — one call per message). For pending sequence `[volunteer#1, abandon, volunteer#2]`:

- `v#1` splices itself, sees tail `volunteer#2 !== "abandon"` (guard at line 706), so `submitVolunteerOp` fires (line 708).
- The `abandon` resubmit is a no-op.
- `v#2` also passes the tail guard and `submitVolunteerOp` fires again.

Result: two `volunteer` ops on the wire instead of the squash contract's single final intent. Each `submitVolunteerOp` (lines 251–266) appends a fresh pending volunteer, compounding queue churn.

Separately, the JSDoc (lines 681–684) admits a lone in-staging `abandon` of a pre-staging volunteer is silently dropped — but the override does not assert this precondition, leaving the client in the queue on the wire when `AllowStagingModeWithoutSquashing` flips. The doc claim that `reSubmitCore` semantics "match" squash semantics is overstated.

Recommendation:

- Preferred: replace the trivial delegation with a per-task collapse (lowest-risk variant: emit only when no other pending op for the same `taskId` remains after splice — strictly stronger than the PR for `[v, a, v]`, reuses existing helpers, no new state).
- Alternative: keep the delegation as a deliberate transitional step but (i) correct the JSDoc to acknowledge the contract gap, (ii) add an `assert` that the pending shape matches the documented narrow precondition (lone-volunteer / volunteer+abandon-pair / volunteer+complete), and (iii) file a follow-up to track real squash semantics before the flag flips.

Current state (trivial delegation + doc claiming equivalence + removed throwing fallback) is the worst combination. Tag ChumpChief on the design call.

PR #27410, opened 2026-05-27
refactor(matrix): extract reSubmitInternal with squash parameter from reSubmitCore

anthony-murphy:
## Description

Extracts `reSubmitInternal(incoming, localOpMetadata, squash: boolean)` from `SharedMatrix.reSubmitCore`. `reSubmitCore` becomes a one-line delegator passing `squash = false`. The two hardcoded `false` arguments to `regeneratePendingOp` (for `SnapshotPath.cols` / `SnapshotPath.rows`) are replaced with the new `squash` parameter so the value can flow through to merge-tree.

## Why

Sets up a clean seam for the upcoming `reSubmitSquashed` override without changing observable behavior: `squash` is always `false` from the only current call site, so output is byte-identical. Pre-landing the seam removes the parameter-threading refactor from the eventual squash PR.

## Notes

Prep for upcoming DDS-wide staging-mode squash work. This PR does not add `reSubmitSquashed` or any per-cell LWW logic.

### DIFF
--- a/packages/dds/matrix/src/matrix.ts
+++ b/packages/dds/matrix/src/matrix.ts
@@ -821,6 +821,14 @@ export class SharedMatrix<T = any>
 	}
 
 	protected reSubmitCore(incoming: unknown, localOpMetadata: unknown): void {
+		this.reSubmitInternal(incoming, localOpMetadata, false);
+	}
+
+	private reSubmitInternal(
+		incoming: unknown,
+		localOpMetadata: unknown,
+		squash: boolean,
+	): void {
 		const originalRefSeq = this.inFlightRefSeqs.shift();
 		assert(
 			originalRefSeq !== undefined,
@@ -868,13 +876,13 @@ export class SharedMatrix<T = any>
 			switch (content.target) {
 				case SnapshotPath.cols: {
 					this.submitColMessage(
-						this.cols.regeneratePendingOp(content, localOpMetadata, false),
+						this.cols.regeneratePendingOp(content, localOpMetadata, squash),
 					);
 					break;
 				}
 				case SnapshotPath.rows: {
 					this.submitRowMessage(
-						this.rows.regeneratePendingOp(content, localOpMetadata, false),
+						this.rows.regeneratePendingOp(content, localOpMetadata, squash),
 					);
 					break;
 				}


anthony-murphy:
Closing: this implements 

anthony-murphy:
## Deep Review

*Reviewed commit `eb682ae` on 2026-05-27.*

Behavior-preserving extraction of `reSubmitInternal(..., squash)` from `SharedMatrix.reSubmitCore`. The new seam preserves the `inFlightRefSeqs.shift()` invariant, deliberately bypasses the cell-set (FWW/LWW) branch, and the `squash=true` path is currently unreachable but harmless because the base `SharedObject.reSubmitSquashed` falls back to `reSubmitCore`. No correctness defects. Three documentation polish items below — noted that the PR is closed pending the actual squash work; these apply to the eventual replacement PR.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add a tracking-issue or follow-up-PR link in the description for the `reSubmitSquashed` override + per-cell LWW work, so reviewers can see where the `squash=true` path becomes reachable

### Context for Reviewers

- Sequence DDS adopted the opposite seam shape for the same migration: it widened `reSubmitCore` directly with `squash: boolean = false` and shipped its `reSubmitSquashed` override in #24424. Matrix keeps the inherited two-parameter `reSubmitCore` and extracts a `private reSubmitInternal` — `SharedObject`, `map`, `directory`, and `task-manager` all preserve the two-parameter shape, so sequence is the outlier.
- The `MatrixOp.set` cell-write branch and the FWW/LWW policy were shaped in #19658; the `inFlightRefSeqs.shift()` assert (0x8b9) was added in #19302 to fix an applyStashedOp empty-queue regression. Both are invariants this PR correctly preserves.
- PR is closed (draft) pending revisit with the actual squash work.

- **Needs human judgment** — Owner sign-off on the seam shape: `private reSubmitInternal` here vs. widening `reSubmitCore` (sequence's choice in #24424). Both shapes are legitimate; the area experts for cross-DDS staging/squash semantics (Abe27342, Mark Fields) should consciously accept the matrix/sequence divergence rather than have it slip in by accident.
- **Cannot be assessed by the pipeline** — End-to-end behavior of the `squash=true` branch is currently unreachable, so the seam is not behaviorally tested. Confirming that the follow-up override + tests are queued is a process check.

</details>

anthony-murphy:
Deep Review: Noted — the PR is being parked for the actual squash work. The three inline comments below and the description-link item still apply to the eventual replacement PR (seam-contract notes for `inFlightRefSeqs.shift()`, the cell-set asymmetry, and a tracking link so the `squash=true` path is reviewable end-to-end). Leaving them here so the follow-up has the traceability.

packages/dds/matrix/src/matrix.ts
```diff
@@ -821,6 +821,14 @@ export class SharedMatrix<T = any>
 	}
```
anthony-murphy, line 821: Deep Review: This PR ships only the plumbing half of the squash seam. `reSubmitInternal` accepts `squash: boolean`, but the only caller is `reSubmitCore` passing `false`, and there is no `SharedMatrix.reSubmitSquashed` override — so the `squash=true` branch is unreachable and untested. Sequence DDS pre-landed both halves together in #24424 (`packages/dds/sequence/src/sequence.ts:796` adds the one-line override alongside the widened `reSubmitCore`).

Behavior is preserved today: the base `SharedObject.reSubmitSquashed` (`packages/dds/shared-object-base/src/sharedObject.ts:490-499`) falls back to `reSubmitCore` when staging-without-squash is allowed, so this is not a broken production path. But the seam ships untested at `squash=true`, and the PR description states "This PR does not add `reSubmitSquashed` or any per-cell LWW logic" without a tracking link.

Two ways to close this:

(a) Add the one-line override now so the `squash=true` path becomes reachable and reviewable end-to-end:

```ts
protected reSubmitSquashed(content: unknown, localOpMetadata: unknown): void {
    this.reSubmitInternal(content, localOpMetadata, true);
}
```

(b) Add a tracking-issue / follow-up-PR link in the description pointing to where the override and per-cell LWW design will land.

packages/dds/matrix/src/matrix.ts
```diff
@@ -821,6 +821,14 @@ export class SharedMatrix<T = any>
...
```
anthony-murphy, line 822: Deep Review: The `squash` asymmetry between this helper's branches is intentional but undocumented. `reSubmitInternal` threads `squash` into the `SnapshotPath.cols` / `SnapshotPath.rows` branches (where merge-tree's coalescing-segment squash semantics apply), but the `MatrixOp.set` cell-write branch (~`matrix.ts:839-874`) does not — cell writes re-run FWW/LWW conflict resolution via `fwwPolicy.cellLastWriteTracker` and `sendSetCellOp` (`matrix.ts:482-501`, `ops.ts:34-47`), which are invariant under merge-tree squash.

A brief comment at the `squash` parameter declaration (or above the `MatrixOp.set` case) noting that cell-set ops are FWW/LWW-governed and intentionally not threaded with `squash` would prevent a plausible wrong-fix in the queued `reSubmitSquashed` follow-up — someone reading the helper could easily try to "complete" the parameter threading by passing it into `sendSetCellOp`.

packages/dds/matrix/src/matrix.ts
```diff
@@ -821,6 +821,14 @@ export class SharedMatrix<T = any>
...
+		this.reSubmitInternal(incoming, localOpMetadata, false);
+	}
+
+	private reSubmitInternal(
+		incoming: unknown,
+		localOpMetadata: unknown,
+		squash: boolean,
+	): void {
 		const originalRefSeq = this.inFlightRefSeqs.shift();
 		assert(
```
anthony-murphy, line 833: Deep Review: The `const originalRefSeq = this.inFlightRefSeqs.shift(); assert(originalRefSeq !== undefined, 0x8b9 …)` invariant deserves an in-code contract note. This assert was added in #19302 (2024-01-23) because the applyStashedOp codepath previously called `.shift()` on an empty queue. Today's `reSubmitCore → reSubmitInternal` path is safe, and the base `SharedObject.reSubmitSquashed` fallback routes through `reSubmitCore`, so the invariant holds.

The risk is future-facing: a hand-rolled `reSubmitSquashed` override that submits row/col messages directly (bypassing `reSubmitInternal`) would silently regress #19302's invariant, and there is no `reSubmitSquashed`-path test today because the override doesn't exist yet.

A one-line comment above `reSubmitInternal` stating that any future `reSubmitSquashed` override MUST delegate through this helper to preserve the `inFlightRefSeqs` invariant (ref #19302) would lock the contract in source.

PR #22787, opened 2024-10-11
merge-tree: Add regression tests for bugs with zero length obliterate 

jzaffiro:
Using the sided obliterate function in shared string fails currently with two different errors:

- `0x2d8: Empty change event should not be emitted`
- `0x045: On insert, seq number already assigned!`

Tests in `sharedString.spec.ts` currently repro the `0x2d8` error only. 
[AB#19742](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/19742)

PR #27411, opened 2026-05-27
feat(sequence): record propertyKeys on interval Add/Change local metadata at submit

anthony-murphy:
## Description

Adds an optional `propertyKeys: ReadonlySet<string>` field to both `IntervalAddLocalMetadata` and `IntervalChangeLocalMetadata` in `intervalCollectionMapInterfaces.ts`, and populates it at the two inline submit sites in `IntervalCollection.add` and `IntervalCollection.change` (`props === undefined ? undefined : new Set(Object.keys(props))`).

## Why

Records which user-defined property keys an interval add/change op carried at submit time. The metadata is small, isolated, and read-only at the consumer; landing it independently keeps the data-layout change separate from the squash semantics that will eventually consume it.

## Notes

Prep for upcoming staging-mode squash work for `SharedString` intervals. This PR does not add the squash filter helpers, the squash branch in `resubmitMessage`, or any new walker logic.


packages/dds/sequence/src/intervalCollection.ts
```diff
@@ -1225,6 +1225,7 @@ export class IntervalCollection
 						type: "add",
 						localSeq,
 						interval,
+						propertyKeys: props === undefined ? undefined : new Set(Object.keys(props)),
```
anthony-murphy: Deep Review: Contract mismatch between the new JSDoc and this populate predicate.

The JSDoc added to `IntervalAddLocalMetadata.propertyKeys` (and the matching one on `IntervalChangeLocalMetadata` at the line 1357 site) states the field carries *"user-defined property keys"*. But `props` here is the raw caller-supplied `PropertySet`, and `new Set(Object.keys(props))` snapshots every key including reserved ones — `reservedIntervalIdKey` and `reservedRangeLabelsKey`.

The rest of the module is careful to segregate those reserved keys from user properties: `getSerializedProperties` (`intervals/sequenceInterval.ts:85-102`) destructures both reserved keys out before exposing `properties`, and `createSequenceInterval` strips them from stored interval properties (`intervals/sequenceInterval.ts:995-997`). On the `change` path at line 1357, `reservedRangeLabelsKey` is rejected upstream (`intervalCollection.ts:1322-1328`), but `reservedIntervalIdKey` can still flow through here on both paths, and label keys can still slip through the add path.

The PR intentionally lands `propertyKeys` as a stable data-layout contract for a future squash consumer. Any consumer that reads the JSDoc at face value will mis-handle ops that touched reserved metadata — label changes that look like user-property deltas, or vice versa. The fix is two lines while the field has zero readers; deferring locks the wrong semantic into a permanent contract.

Two valid resolutions:

1. Filter reserved keys at both submit sites — either inline (`Object.keys(props).filter(k => k !== reservedIntervalIdKey && k !== reservedRangeLabelsKey)`) or by sourcing from the already-computed serialized bag: `Object.keys(getSerializedProperties(serializedInterval).properties)`, which performs the same stripping the rest of the module relies on.
2. Keep raw caller keys and reword the JSDoc on both interfaces to say "property keys present on this op's `props` bag at submit, including any reserved keys" — so the consumer knows it must filter.

Either resolution closes the finding, but the choice is a contract decision worth a sign-off (Abe27342 or pleath).

packages/dds/sequence/src/intervalCollection.ts
```diff
@@ -1353,6 +1354,7 @@ export class IntervalCollection
 					type: "change",
 					localSeq,
 					interval: newInterval ?? interval,
+					propertyKeys: props === undefined ? undefined : new Set(Object.keys(props)),
```
anthony-murphy: Deep Review: Same contract mismatch as the inline at line 1228 — `new Set(Object.keys(props))` snapshots reserved keys despite the "user-defined" JSDoc on `IntervalChangeLocalMetadata.propertyKeys`.

On this `change` path, `reservedRangeLabelsKey` is rejected upstream at intervalCollection.ts:1322-1328, but `reservedIntervalIdKey` is not, so the leak path is real here too. Fix at both sites together (filter the predicate, or reword the JSDoc on both interfaces) — see the line 1228 thread for the two resolution options and the rationale.

packages/dds/sequence/src/intervalCollection.ts
anthony-murphy: Deep Review: Resolved in `eae55e7`. Both populate sites now source from `new Set(Object.keys(getSerializedProperties(serializedInterval).properties))`, which strips `reservedIntervalIdKey` and `reservedRangeLabelsKey` centrally — matches the "user-defined property keys" JSDoc. New tests in `sharedString.spec.ts:1084-1108` and `:1140-1163` assert reserved-key exclusion on both add and change paths.

packages/dds/sequence/src/intervalCollection.ts
anthony-murphy: Deep Review: Resolved in `eae55e7` alongside the line 1228 site — same `getSerializedProperties(...).properties` sourcing closes the `reservedIntervalIdKey` leak on the change path too.

packages/dds/sequence/src/test/sharedString.spec.ts
```diff
@@ -989,3 +989,177 @@ describe("Shared String Obliterate", () => {
 		assert.equal(sharedString2.getText(), "01234BBB56789", "obliterate failed");
 	});
```
anthony-murphy, line 990: Deep Review: This new 174-line `describe("IntervalCollection local op metadata propertyKeys", …)` block exercises `getIntervalCollection().add/change` exclusively and asserts nothing SharedString-specific — the `SharedString` is only used to instantiate the harness. The established home for IntervalCollection behavior tests is the sibling `packages/dds/sequence/src/test/intervalCollection.spec.ts` (alongside `intervalCollection.rollback.spec.ts`, `intervalCollection.events.spec.ts`, `intervalCollection.detached.spec.ts`), and `intervalCollection.spec.ts:87-98` already instantiates `SharedStringClass` for interval-collection tests, so the harness is portable.

`sharedString.spec.ts`'s historical scope (PRs #22787, #23011) is SharedString text/obliterate regressions, not interval-collection metadata. Move the new `describe` block into `packages/dds/sequence/src/test/intervalCollection.spec.ts`, or create `intervalCollection.localMetadata.spec.ts` following the `intervalCollection.rollback.spec.ts` pattern. The relocation is mechanical — the harness only depends on `SharedStringClass` and `MockContainerRuntimeFactory`, both already in scope of the other interval specs.

If the area owners prefer the current location (harness closer to the SharedString integration boundary), that overrides the convention argument — worth a sign-off either way.

anthony-murphy:
## Deep Review

*Reviewed commit `cf41585` on 2026-06-02.*

No new code changes since the prior round; the reserved-key leak fix in `eae55e7` still holds and tests still lock it in. Remaining items are documentation/discoverability polish — the PR description still shows the old `Object.keys(props)` snippet rather than the shipped `getSerializedProperties(...).properties` expression, the `propertyKeys` JSDoc doesn't document the load-bearing `undefined` vs empty-`Set` distinction, the squash-consumer follow-up link isn't filled in, and the open test-placement thread is awaiting an author response.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Replace the PR description with the "Suggested description" block below — fixes the `Object.keys(props)` vs `getSerializedProperties(...).properties` drift in one edit
- [ ] Fill in the squash-consumer issue/PR link in the "Notes" section of that replacement description before un-drafting
- [ ] Reply on thread `3314861225` — either move the new `describe` block to `intervalCollection.spec.ts` (or a new `intervalCollection.localMetadata.spec.ts`), or record why the current `sharedString.spec.ts` location is preferred

### Context for Reviewers

- `resubmitMessage` re-forwards the same `localOpMetadata` object via `submitDelta` and never reconstructs metadata from `value`, so `propertyKeys` survives resubmit/rebase unchanged — the intended invariant for the future squash consumer ("keys the user originally asked to write" should not change because position rebased).
- `intervalCollectionMapInterfaces.ts` is unusually low-churn (1 edit / 6mo) and recent history (#24829, #24867 — both anthony-murphy) tightened/replaced existing fields rather than adding optional ones. Adding `propertyKeys?` as a fourth top-level optional is defensible for a metadata-only prep change, but is a stylistic departure from the "replace, don't accumulate" trend the consumer PR should revisit.
- Reconnect/rebase has historically been the sharp edge for interval change metadata (#10342, #24867); the consumer PR will need explicit rebase coverage on change ops even though this PR is only recording submit-time keys.
- PR #24867 deliberately slimmed `IntervalAddLocalMetadata` / `IntervalChangeLocalMetadata` from serialized op data (`original`) down to a live `LocalReferencePosition` (`interval`); this PR reintroduces a small property-derived slice (keys only, not values) onto the same shape.
- `IntervalDeleteLocalMetadata` is intentionally omitted — delete ops carry no `props` bag in the post-#24186 / post-#24867 design.

- **Test-placement convention call (thread `3314861225`)** — the mechanical relocation argument to `intervalCollection.spec.ts` is strong, but area owners may prefer the harness closer to the SharedString integration boundary. Suggested sign-off: **jzaffiro** (authored SharedString test PR #22787).
- **Defer-vs-pin resubmit-preservation regression test** — judgment call on whether to pin the `propertyKeys`-survives-resubmit invariant here or in the consumer PR. Suggested sign-off: **Abe27342** (reviewed pending-state lifecycle on #24307).
- **`ReadonlySet<string>` vs `readonly string[]` vs full `PropertySet`** — the O(1) `has()` justification is defensible for the future squash walker's expected access pattern, but only the consumer-PR drafter can confirm.
- **Field name `propertyKeys` vs `propertyDeltaKeys`** — taste call; JSDoc disambiguates today. Revisit if consumer-PR call sites read better with the more specific name.
- **Draft-PR landing discipline** — confirm the squash consumer is at least drafted before merging so `propertyKeys` does not become orphaned dead data on the submit path.

</details>

### Suggested description

```markdown
## Description

Adds an optional `propertyKeys: ReadonlySet<string>` field to both `IntervalAddLocalMetadata` and `IntervalChangeLocalMetadata` in `intervalCollectionMapInterfaces.ts`, and populates it at the two inline submit sites in `IntervalCollection.add` and `IntervalCollection.change`:

```ts
propertyKeys:
    props === undefined
        ? undefined
        : new Set(Object.keys(getSerializedProperties(serializedInterval).properties)),
```

Sourcing from `getSerializedProperties(serializedInterval).properties` (rather than raw `Object.keys(props)`) centralizes reserved-key stripping in `getSerializedProperties`, which already destructures `reservedIntervalIdKey` and `reservedRangeLabelsKey` out of the returned `properties`. The resulting `propertyKeys` contains only user-defined keys — verified by the new test cases for both `add` and `change` paths.

`undefined` means the op was submitted without any `props` bag at all; an empty `Set` means an explicit empty `{}` was passed. The distinction is preserved because the future squash consumer needs to tell "this op did not touch user properties" from "the user explicitly cleared/no-op'd the bag."

## Why

Records which user-defined property keys an interval add/change op carried at submit time. The metadata is small, isolated, and read-only at the consumer; landing it independently keeps the data-layout change separate from the squash semantics that will eventually consume it. This reintroduces a small property-derived slice onto interfaces that #24867 had deliberately slimmed — a knowingly additive step.

## Notes

Prep for upcoming staging-mode squash work for `SharedString` intervals. This PR does not add the squash filter helpers, the squash branch in `resubmitMessage`, or any new walker logic. The resubmit path preserves `propertyKeys` for free because `resubmitMessage` re-forwards the same `localOpMetadata` object via `submitDelta` (no recompute) — pinning a regression test for that round-trip is deferred to the consumer PR.

Follow-up: link the squash consumer issue/PR here before un-drafting (TODO: #<issue-or-pr-number>).
```

- `eae55e7` 2026-05-28 · **8/10** — reserved-key leak fixed; remaining items are docs/discoverability polish
- `2f02e9f` 2026-05-27 · **5/10** — contract mismatch flagged inline at both submit sites: JSDoc promised "user-defined property keys" but populate predicate snapshotted raw `Object.keys(props)`

</details>

File: packages/dds/sequence/src/intervalCollectionMapInterfaces.ts [hunk 4]
```diff
@@ -24,12 +24,24 @@ export interface IntervalAddLocalMetadata {
...
+	 */
+	propertyKeys?: ReadonlySet<string>;
 }
 export interface IntervalChangeLocalMetadata {
 	type: typeof IntervalDeltaOpType.CHANGE;
 	localSeq: number;
 	endpointChangesNode?: ListNode<IntervalChangeLocalMetadata | IntervalChangeLocalMetadata>;
 	interval: SequenceIntervalClass;
+	/**
+	 * Set of user-defined property keys present at submit (reserved keys such as the interval id
```
anthony-murphy, line 40: Deep Review: The JSDoc here (and the matching one on `IntervalChangeLocalMetadata.propertyKeys` around line 52) only states `"Undefined when the op was submitted without any user-supplied property bag."` That documents `undefined` but leaves the empty-`Set` case implicit.

The tests at `sharedString.spec.ts:1084-1108` (add) and `:1140-1163` (change) lock in that `props: {}` produces an empty `Set` (defined, `size === 0`), not `undefined`. The implementation expression `props === undefined ? undefined : new Set(...)` (rather than the simpler `props ? new Set(...) : undefined`) deliberately preserves a three-valued state — `undefined` / empty `Set` / populated `Set`.

With no consumer in-tree yet, a future squash-mode reader writing the natural `if (!propertyKeys?.size)` would silently fold "no `props` field passed" and "explicit `props: {}`" together, and the producer-side tests in this PR would not catch the regression. Extend the JSDoc on both declarations to call out: `undefined` = caller did not pass a `props` field; empty `Set` = caller passed `props: {}`; these states are intentionally distinct and consumers must not collapse them.

PR #25390, opened 2025-09-04
Toggle and toggleMove rollback

dannimad:
Adding toggle and toggleMove ops rollback in SharedArray. Logic is pretty much identical to their respective methods, since this ops are undoing previous insert/move operations, so toggling twice would undo the undo. Only difference is seeing respective local delete and move flags to zero.

Adding unit and fuzz tests, with the comment of disabling toggle rollback fuzz tests for now since there is an existing bug in toggle that makes rollback fail a bunch of seeds. Enabling back once that bug is solved.

PR #27412, opened 2026-05-27
legacy-dds: SharedArray pending-ops ledger + submit funnel

anthony-murphy:
## Description

Bundles three structural changes for `SharedArray` in `legacy-dds`:

1. **FIFO `pendingOps` ledger** backed by `DoublyLinkedList<SharedArrayPendingOp<T>>` — adds the `SharedArrayPendingOp<T>` interface, a private `pendingOps` list, a `buildPendingOp(op)` builder switch, a push at each of the 6 local submit sites (`insertEntry` / `delete` / `move` / `toggle` / `toggleMove` / `applyStashedOp` tail), and a shift+assert on local-ack in `processMessage`.
2. **`submitArrayOp` funnel** — routes those 6 sites through a single private `submitArrayOp(op: ISharedArrayOperation<T>)` helper instead of calling `submitLocalMessage` directly.
3. **`findEntryById` via `idToEntryMap`** — a new helper that consults the already-maintained id → entry Map for O(1) lookup. Converts `prepareToMakeEntryIdLive`'s neighbor lookups (no longer touches `sharedArray`) and `handleStashedInsert`'s entry lookup. `handleInsertOp` is intentionally not converted because it feeds the index to `getInternalInsertIndexByIgnoringLocalPendingInserts`.

## Why

- The pendingOps ledger gives in-flight `SharedArray` ops a first-class, legible record (today they are inferred from per-entry counters scattered across the skip list); the FIFO shift+assert documents the invariant that local acks arrive in submit order. Backing the ledger with `DoublyLinkedList` makes the head shift O(1) rather than O(n).
- The `submitArrayOp` funnel gives a single hook site for any future per-op routing or telemetry.
- `findIndexOfEntryId` was a linear scan over `sharedArray` despite a Map sitting next to it. For the 100K-entry scenarios called out in the SharedArray docstring, every remote insert / move paid 100K comparisons.

## Notes

Pure structural prep — no `reSubmitSquashed` override, no squash logic, no tests touched, no api-report changes.


packages/dds/legacy-dds/src/array/sharedArray.ts
```diff
@@ -618,7 +667,14 @@ export class SharedArrayClass<T extends SerializableTypeForSharedArray>
...
 				}
 			}
-			if (!local) {
+			if (local) {
+				// FIFO consume: the head of pendingOps is the ack target.
+				const ackedPendingOp = this.pendingOps.shift();
+				assert(
+					ackedPendingOp !== undefined && ackedPendingOp.entryId === op.entryId,
+					"pendingOps head must match acked op entryId",
+				);
```
anthony-murphy, line 719: Deep Review: The new FIFO `pendingOps` shift+assert here collides with the already-shipped `rollback()` at `sharedArray.ts:428` (added in PR #25234). `rollback()` reverts a local in-flight op without calling `submitLocalMessage` and — grep-confirmed — without touching `pendingOps`. After a rollback, the original op never gets acked, but its ledger record stays in `pendingOps`. The next locally-acked op shifts the stale record off the head, the `entryId === op.entryId` comparison fails, this assert fires, and the container disconnects/closes.

The failing sequence `submit → rollback → submit → ack` is exercised today by `src/test/array/array.rollback.spec.ts:71-426` (insert/delete/move/toggle/toggleMove via `containerRuntime.rollback?.()`) and by the rollback fuzz suite from #25234, so this regresses shipped coverage. Framing the change as "pure structural prep" doesn't insulate it: `rollback()` is not deferred.

Additional wrinkle: `rollback` runs LIFO while `pendingOps` is FIFO, so the rolled-back op is generally *not* at the head — a naive `shift()` inside `rollback` would not fix it. Removal must target the matching record by op identity (same `ISharedArrayOperation<T>` reference) or by shape `(type, entryId[, targetEntryId])`.

**Recommended fix:** In `rollback()`, splice the matching `pendingOps` record by identity/shape *before* the inverse mutation, and assert exactly one record was removed. Add a unit test that submits, rolls back, then submits and acks another op (per op family) to prove the FIFO `entryId` assert no longer fires, and re-run the #25234 rollback fuzz against this branch.

Alternatives to weigh: (b) loosen this ack assert to a non-FIFO membership check, or (c) explicitly override `rollback()` as a no-op until pendingOps integration lands and flag-gate this assert — but (a) preserves the FIFO invariant the PR is documenting.

Related items being held until this is resolved and will re-emerge alongside the fix: tag this new `assert` with a `0xb9x` code consistent with the file's other asserts (lines 254, 280, 331, 790, 843, 856, 919, 941, 946); strengthen the assert to also check `ackedPendingOp.type === op.type` and (for move/toggleMove) `ackedPendingOp.targetEntryId === op.changedToEntryId` — the ledger already stores both; and document at the `pendingOps` declaration that on reconnect/resubmit the base-class path is intentionally not threaded through `submitArrayOp` and records persist for the original ack to drain.

anthony-murphy:

- ✅ Mechanical Review
- 👁️ **Adversarial Filter** — puzzling over the find…
- ⬜ Synthesis
- ⬜ Comments

packages/dds/legacy-dds/src/array/sharedArray.ts
```diff
@@ -618,7 +700,25 @@ export class SharedArrayClass<T extends SerializableTypeForSharedArray>
...
+				const ackedTargetEntryId =
+					op.type === OperationType.moveEntry || op.type === OperationType.toggleMove
+						? op.changedToEntryId
+						: undefined;
+				assert(
+					acked !== undefined &&
+						acked.type === op.type &&
+						acked.entryId === op.entryId &&
+						acked.targetEntryId === ackedTargetEntryId,
+					0xb97 /* pendingOps head must match acked op shape */,
```
anthony-murphy, line 718: Deep Review: This new assert hand-picks short code `0xb97`, which `packages/runtime/test-runtime-utils/src/assertionShortCodesMap.ts:1645` still binds to the `"Invalid index"` assert removed from `prepareToMakeEntryIdLive` in this same diff (`0xb98` at line 1646 is the other removed tag). Short codes in this repo are append-only canonical identifiers stamped at release time by the assert-tagging pipeline (PR #24711, tylerbutler, 2025-05-27), so re-binding `0xb97` to a new, semantically unrelated message will either fail the policy/check step in CI or — if it slips past — alias post-release crash telemetry from shipped 2.x containers onto this new ledger assert, mislabeling field triage as `"Invalid index"`.

The other new assert in `removeMatchingPendingOp` a few lines above is correctly untagged (string only) and demonstrates the safer pattern — let the next `tag asserts for release` run mint a fresh code.

**Recommended fix:** drop the `0xb97` literal here; leave the message string only. Retire `0xb97` and `0xb98` rather than recycling them.

```ts
assert(
    acked !== undefined &&
        acked.type === op.type &&
        acked.entryId === op.entryId &&
        acked.targetEntryId === ackedTargetEntryId,
    "pendingOps head must match acked op shape",
);
```

Question: was the reuse intentional (e.g., to preserve a known telemetry signal) or an oversight?

packages/dds/legacy-dds/src/array/sharedArray.ts
```diff
@@ -635,6 +734,14 @@ export class SharedArrayClass<T extends SerializableTypeForSharedArray>
 			this.getEntryForId(entryId).isAckPending = false;
```
anthony-murphy, line 734: Deep Review: `handleInsertOp` is the per-remote-insert hot path the PR description explicitly calls out as the perf motivation — *"`findIndexOfEntryId` was a linear scan over `sharedArray` despite a Map sitting next to it. For the 100K-entry scenarios called out in the SharedArray docstring, every remote insert / move paid 100K comparisons."* But this PR converts only `prepareToMakeEntryIdLive` and `handleStashedInsert` (neither on the remote-op hot path) and leaves `handleInsertOp`'s `this.findIndexOfEntryId(insertAfterEntryId)` in place. The inline comment justifies the deferral on the grounds that converting it would require restructuring local-pending-state tracking (the index feeds `getInternalInsertIndexByIgnoringLocalPendingInserts`).

That's a defensible engineering call, but it leaves the PR's stated scope mismatching what shipped: the description sells a perf fix the diff doesn't deliver on the hot path it names. Two structurally cheaper options would close the gap: an in-memory `arrayIndex?: number` field on `SharedArrayEntry` maintained by `addEntry`/`loadCore` (splice-tail repaint is asymptotically free for append-only bulk insert), or a less-invasive `idToIndexMap` alongside the existing `idToEntryMap`. Either makes the 100K-entry hot path O(1) without touching the local-pending consumer's contract — the consumer reads `.arrayIndex` (or `idToIndexMap.get(id)`) instead of computing it.

**Two acceptable paths — pick one:**

(a) Land an `arrayIndex` field on `SharedArrayEntry` (or an `idToIndexMap`) in this PR and convert `handleInsertOp` and `getInternalInsertIndexByIgnoringLocalPendingInserts` to use it. Delivers the perf win the description promises.

(b) Retitle / rescope the PR ("structural prep for ledger/funnel; partial perf wins, hot-path follow-up") and file a tracking issue for the `arrayIndex` work, so the gap between claimed scope and delivered scope is explicit.

Either is fine. The current shape — description sells a 100K-entry perf fix the diff doesn't deliver on the named hot path — is not.

packages/dds/legacy-dds/src/array/sharedArray.ts
```diff
@@ -572,6 +606,54 @@ export class SharedArrayClass<T extends SerializableTypeForSharedArray>
...
+	 * Locate and remove the {@link pendingOps} ledger record that corresponds to
+	 * the supplied rolled-back op. The ledger is FIFO but rollback can target any
+	 * in-flight op, so we walk the list and match by op-shape: `type` + `entryId`,
+	 * plus `targetEntryId` for {@link OperationType.moveEntry} and
+	 * {@link OperationType.toggleMove}. Asserts exactly one match is removed so the
+	 * ledger stays in sync with the locally-applied op set.
+	 */
+	private removeMatchingPendingOp(op: ISharedArrayOperation<T>): void {
+		const expected = this.buildPendingOp(op);
+		const match = this.pendingOps.find(
```
anthony-murphy, line 644: we can make this O(1) if we store the ListNode as the localOpMetadata

anthony-murphy:
## Deep Review

*Reviewed commit `17681ea` on 2026-06-02.*

Not ready for sign-off. The rollback × ledger collision from prior rounds is resolved by `removeMatchingPendingOp` plus the five per-op-family regression tests in `array.rollback.spec.ts`. Four Tier 2 concerns remain and they compound: (1) the new ledger's match key `(type, entryId, targetEntryId)` does not include `isDeleted`, so two pending same-entry toggles with opposite `isDeleted` are indistinguishable on both rollback (`removeMatchingPendingOp` head-matches the wrong entry) and local-ack (the new `0xb97` head-shape assert passes anyway and silently corrupts the ledger); (2) the previously-flagged `0xb97` shipped-tag reuse is still open inline on line 718; (3) no test exercises the multi-pending / duplicate-shape rollback path — the case that would catch (1) and validate the load-bearing `0xb97` assert; (4) no regression covers the new `applyStashedOp` ledger push, which is historically the most fragile local site in this file (#25140, #25281, #25980). Three Tier 3 polish items are held this round.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Update the PR description: drop the "no tests touched" line (the diff adds +99 lines to `array.rollback.spec.ts`) and tighten the perf framing so it matches the delivered scope — `handleInsertOp` (the per-remote-insert hot path the description names as the 100K-entry motivation) is intentionally left on the linear scan in this PR. Either narrate the deferral inline or rescope the headline; current text overstates what shipped.
- [ ] Re-run the rollback fuzz suite and the `applyStashedOp` stress suite from PR #25281 against this branch before merge; confirm the new `0xb97` head-shape assert in `processMessagesCore` stays quiet on both. (Carried from prior round; promoted into the inline S4 thread.)

### Context for Reviewers

- `applyStashedOp`'s tail now participates in the strict FIFO ack ledger for the first time. Stashed/reconnect is the most historically fragile local site in this file — ChumpChief flagged it on #25140 ("no direct mapping of op to normal-ish local method"), #25281 patched it after stress exposure, #25980 patched it again after a rollback × toggle bug. Layering a hard assert here is the right direction but warrants re-running the stress/fuzz suite enabled by #25281.
- The rollback design intentionally rejects a naive LIFO `pop()` on `pendingOps`: rollback is LIFO across the whole runtime but other DDSes' rollbacks can interleave, and within SharedArray the rolled-back op isn't always the most recent. The op-shape `find` in `removeMatchingPendingOp` handles multi-DDS interleaving correctly where blind `pop()` would silently corrupt the ledger.
- `processMessagesCore` is recent op-bunching boilerplate from PR #25745 (agarwal-navin, 2025-10-24); the SharedArray-specific FIFO shift+assert is layered inside that loop.
- Short codes in this repo are append-only canonical identifiers stamped at release time by the assert-tagging pipeline (PR #24711, tylerbutler, 2025-05-27) — the basis for the `0xb97` reuse concern on line 718.
- The `removeMatchingPendingOp` assert (`rollback target must have a matching pendingOps ledger entry`) follows the author's own past review guidance on PR #25234 ("should this throw, rather than break, since rollback didn't actually occur") — the philosophy is consistent.
- Suggested reviewers: **dannimad** (owns `rollback()` and `applyStashedOp` — #24661, #25140, #25234, #25281, #25390, #25980), **ChumpChief** (design reviewer on #25140 and #25390), **Abe27342** (toggle/toggleMove review on #25140), **agarwal-navin** (author of #25745), **tylerbutler** (owns the assert-tagging policy from #24711).

- **dannimad** — design call on S1 (toggle uniqueness): include `isDeleted` in the `SharedArrayPendingOp` toggle record + match keys, or attach a `localSeq` and round-trip via `localOpMetadata` (your own line-644 note `3320014698` already hints at the latter). Also the line-734 thread on the `handleInsertOp` scope call — land an `arrayIndex` / `idToIndexMap` in this PR vs. rescope the PR description and file a tracking issue.
- **ChumpChief** — evaluate the new `submitArrayOp` passthrough and the extended `applyStashedOp` ledger push against the recurring "hard to inspect with confidence" feedback from #25140. Note that `pendingOps.push` happens at six call sites outside the funnel and `applyStashedOp` is the one site without an `isAttached()` guard.
- **Abe27342** — sanity-check S1 against the toggle/toggleMove semantics you reviewed on #25140 — specifically whether two same-entry toggles with opposite `isDeleted` can actually be in flight simultaneously under any user-visible path.
- **agarwal-navin** — confirm the local-ack shift+assert in `processMessagesCore` composes correctly with batched/bunched message processing (#25745); the pipeline cannot assess batching semantics at runtime.
- **tylerbutler** — confirm whether `0xb9x` short codes will be auto-assigned by the next release-branch tagger run once the literal is dropped, and whether `0xb97`/`0xb98` should be removed from `assertionShortCodesMap.ts` manually now.
- **Scope call** — land the ledger now with `rollback()` integrated but `reSubmitSquashed` / squash logic deferred (current shape), or hold until both lifecycle paths are integrated. Current shape is internally consistent if the four Tier 2 items above land.

</details>

- `9735af6` 2026-05-28 · **5/10** — rollback × pendingOps collision resolved; `0xb97` reuse + line-734 scope mismatch open
- `30171cb` 2026-05-27 · **5/10** — rollback × pendingOps collision resolved; `0xb97` tag-reuse flagged inline
- `8ef3a20` 2026-05-27 · **5/10** — rollback × pendingOps collision flagged inline
- `8e4862b` 2026-05-27 · **2/10** — rollback × pendingOps collision flagged inline
- `1203744` 2026-05-27 · **2/10** — rollback × pendingOps collision flagged inline

</details>

packages/dds/legacy-dds/src/array/sharedArray.ts
anthony-murphy: Deep Review: Picking up on the `localOpMetadata` direction here — it also closes a concrete identity-key gap a few lines down. `removeMatchingPendingOp` (lines 642-654) and the new `0xb97` head-shape assert (lines 703-719) both match on `(type, entryId, targetEntryId)` only. `IToggleOperation` carries an `isDeleted` field that is *not* in the match key, and `buildPendingOp` for `OperationType.toggle` (lines 612-618) does not store it. `toggle()` derives `isDeleted` from current state and does not reject a second local toggle on the same live entry while earlier toggles remain unacked.

Failing sequence: submit toggle A on entry X (sets `isDeleted=true`), submit toggle A' on entry X with opposite `isDeleted`, rollback A' → `find(...)` head-matches and removes A's record instead. When A then acks, the (now-stale) head still passes the reduced shape check, the `0xb97` assert is silent, and the ledger is silently corrupted. The PR writeup acknowledges this: *"There is a latent uniqueness concern … correctness depends on `(type, entryId, targetEntryId)` being unique among in-flight ops."*

**Two acceptable fixes — pick one:**

(a) **Minimal** — include `isDeleted` in `SharedArrayPendingOp` for `OperationType.toggle`, in the `removeMatchingPendingOp` predicate, and in the `0xb97` head-shape assert. Self-contained at the build/match-key seam.

(b) **Stronger (subsumes this concern + future op-shape evolution)** — exactly the `localOpMetadata` route you mention here. Attach a monotonic per-submit `localSeq` to the ledger entry, round-trip it through `localOpMetadata` (SharedCell / TaskManager precedent), assert on `localSeq` for ack and rollback, and keep the shape check as a sanity assertion. Storing the `ListNode` as `localOpMetadata` (your suggestion) gives O(1) `removeMatchingPendingOp` *and* a unique handle in one move.

Either must ship with the regression test called out separately on `array.rollback.spec.ts` — submit toggle A, submit toggle A' with opposite `isDeleted`, rollback A', ack A, assert ledger drains cleanly. Without that test the fix can't be proven and the new load-bearing `0xb97` assert isn't exercised against the case it's there to catch.

packages/dds/legacy-dds/src/test/array/array.rollback.spec.ts
```diff
@@ -423,4 +423,103 @@ describe("SharedArray rollback", () => {
...
+			// Second toggleMove on a different prior move.
+			revertibles[4].revert();
+			containerRuntime.flush();
+			containerRuntimeFactory.processAllMessages();
+			assert.deepStrictEqual(
+				sharedArray.get(),
+				[1, 2, 0],
+				"Second toggleMove should ack cleanly",
+			);
+		});
```
anthony-murphy, line 523: Deep Review: The five new tests (one per op family, lines 427-523) each do: submit one op → rollback → submit a second op of the same family → ack. With only one op in flight, `removeMatchingPendingOp`'s `find(...) + assert exactly one match` is functionally indistinguishable from a `shift()` — the test the PR doesn't have is the test that exercises the new logic. The PR writeup admits: *"The tests do not exercise rollback across multiple in-flight ops simultaneously, nor mixed-family rollback orderings — they're per-family smoke regressions."*

The new `0xb97` head-shape assert on `sharedArray.ts:718` is load-bearing (disconnect-on-trip on container lifetime), and the conditions under which it most likely fires are exactly the ones not covered.

**Recommended additions:**

1. **Duplicate same-entry toggle rollback** — submit toggle A on entry X (e.g. live → deleted), submit toggle A' on entry X (deleted → live), rollback A', ack A. Assert `pendingOps` drains cleanly and the `0xb97` assert doesn't fire. This is the exact scenario the inline thread on line 644 / `removeMatchingPendingOp` flags as the concrete identity-key gap, and the test that would catch it.
2. **Multi-op-in-flight rollback targeting a non-head entry** — submit ops A, B, C of mixed families, rollback B, ack A then C. Validates that `find(...)`-by-shape (not `shift()`) actually does the right thing when the rolled-back op isn't at the head.
3. **Post-drain ledger invariant in the connected / fuzz suites** — bolt a `pendingOps.length === 0` assertion after each drain into `arrayFuzzTests.spec.ts` (or the connected harness used here). Cheap, catches a whole class of leak-the-ledger regressions, and is exactly the kind of invariant prior rollback-area work (#25234, #25390) has reached for after the fact.

packages/dds/legacy-dds/src/array/sharedArray.ts
```diff
@@ -1024,6 +1156,7 @@ export class SharedArrayClass<T extends SerializableTypeForSharedArray>
 				unreachableCase(op);
 			}
 		}
-		this.submitLocalMessage(op);
+		this.pendingOps.push(this.buildPendingOp(op));
+		this.submitArrayOp(op);
```
anthony-murphy, line 1160: Deep Review: `applyStashedOp` now pushes onto the new `pendingOps` ledger and routes through `submitArrayOp` — the first time the stashed-op tail participates in the strict FIFO ack invariant. Two structural points and a coverage gap:

1. **No `isAttached()` guard here.** The five public mutators (`insert`/`delete`/`move`/`toggle`/`toggleMove`) all gate ledger push on `isAttached()`; `applyStashedOp` does not. This is the single piece of the prior "centralize the push inside `submitArrayOp`" feedback that survives factually — whether intentional or not, it's worth a comment in the code so the asymmetry is legible, and an argument for moving the push into `submitArrayOp` so the guard lives in one place.
2. **No targeted regression for this path through the new ledger.** The five new rollback tests (`array.rollback.spec.ts:427-523`) cover normal local submit→rollback→submit→ack flows only — none drive an op through `applyStashedOp` and back out through the local-ack drain or rollback.
3. **Historically the most fragile local site in this file.** PR #25281 (dannimad, 2025-09-02) added stashed-op fuzz tests *after* discovering `applyStashedOp` was setting internal pending flags incorrectly and had ordering bugs. PR #25980 (dannimad, 2025-12-05) patched it again after a rollback × toggle bug. Layering a hard assert here is the right direction, but warrants empirical soak.

**Recommended:** add one focused regression that drives an op through `applyStashedOp` and exercises the new ledger interaction with the local-ack drain (or rollback), and re-run the rollback fuzz suite + the PR #25281 `applyStashedOp` stress suite against this branch before merge — confirm the new `0xb97` head-shape assert on line 718 does not fire from the stashed-op path.

PR #24253, opened 2025-04-04
feat(merge-tree): Support resubmission of obliterates

Abe27342:
## Description

This updates merge-tree's `resubmit` to support resubmission for obliterates. It takes generally the same strategy that we have used in `intervalCollection`, with a few subtleties around updating additional bookkeeping we have in `ObliterateInfo` and on segments.

As part of this PR, I've added some unit tests to the obliterate reconnect file and turned on reconnection in the obliterate fuzz test suite.

PR #27414, opened 2026-05-27
merge-tree: split reset arms + previousProps accessor backed by WeakMap

anthony-murphy:
## Description

Bundles three merge-tree structural changes:

1. **Extract `resetAnnotateOp` / `resetInsertOp` helpers** from the long switch in `Client.resetPendingDeltaToOps` (`client.ts`). Pure code motion: the ANNOTATE and INSERT arms become one-line calls to the new private helpers, with their original control flow, parameters, and assertions preserved byte-identical.
2. **`SegmentGroupCollection.previousPropsForSegment` accessor** — encapsulates the "what's the previousProps entry for this collection's segment in the given SegmentGroup" lookup.
3. **`SegmentGroup.previousProps: WeakMap<ISegmentLeaf, PropertySet>`** — replaces the array (paired by index with `SegmentGroup.segments`) with a WeakMap keyed by segment. Removes the fragile "i-th `previousProps` entry pairs with i-th `segments` entry" invariant that comments had to call out. The `previousPropsForSegment` accessor becomes a clean `return segmentGroup.previousProps?.get(this.segment);` one-liner.

Sites converted for the WeakMap migration:

- `mergeTreeNodes.ts`: type declaration.
- `mergeTree.ts:addToPendingList`: `previousProps = new WeakMap()`.
- `mergeTree.ts` write site: `set(segment, propertySet)` instead of array push.
- `mergeTree.ts:2479` (rollback ANNOTATE): `previousProps!.get(segment)` instead of `previousProps![i]`. The `i` counter is removed.
- `client.ts` reissue path: the original `.slice(0)` was a defensive copy of the props array, but the surrounding loop only ever wrote a single segment to the new group, so the array shape was already mispaired post-copy (latent inconsistency). Replaced with a new `WeakMap` containing only the entry for the current segment.
- `segmentGroupCollection.ts:enqueueOnCopy`: drops the `indexOf(sourceSegment)`; maps source's WeakMap value onto destination segment in the destination map.

Truthy-presence semantics (e.g. `op.type === ANNOTATE && !pendingSegmentGroup.previousProps`) work unchanged — a `WeakMap` is truthy.

## Why

`Client.resetPendingDeltaToOps` is ~360 lines with a 4-way switch and triple-nested ternaries in the ANNOTATE arm. Extracting the arms makes future per-arm changes localized to small focused helpers.

The implicit "i-th entry pairs with i-th segment" invariant in `previousProps` was a known source of bugs and forced comment annotations. Keying on the segment directly via `WeakMap` makes the pairing inherent in the type. As a bonus, two array `indexOf` calls in `segmentGroupCollection.ts` are now O(1).

## Notes

`previousProps` is purely internal merge-tree bookkeeping (not on the wire, not public API). Pure structural prep — no behavior change, no squash logic, no tests touched, no api-report changes.


anthony-murphy:
## Deep Review

*Reviewed commit `8d34c55` on 2026-05-27.*

Pure-refactor prep PR. The `resetAnnotateOp` / `resetInsertOp` extraction is behaviour-preserving (each `break;` maps 1:1 to `return undefined;`, asserts and side effects ordered identically), and the additive `previousPropsForSegment` accessor is consistent with the established no-clone, package-internal `SegmentGroupCollection` contract. One contained polish item — no direct test for the new accessor — flagged inline.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- The ANNOTATE arm's `// unless the remove was local…` comment and `!isRemovedAndAcked(segment)` guard trace to anthony-murphy's review pushback on PR #9944 (wes-carlson's original used `removedSeq === undefined && localRemovedSeq === undefined`, which would have incorrectly skipped locally-removed segments). Both are carried byte-identical into `resetAnnotateOp`.
- The INSERT arm preserves the `mergeTreeEnableObliterateReconnect`-gated block from PR #22080 and the `0xbaf` "Expected only one local remove" squash assert from PR #24253. Asserts `0x036`, `0x037`, `0xb5c`, `0xbaf` are tagged for cross-version telemetry stability (#24197, #24349) and must not be renumbered.
- The accessor's shared-reference (no-clone) return matches the invariant established in PR #11961: `previousProps` is immutable by construction ("changing it would mean somehow changing history"). `SegmentGroup` / `SegmentGroupCollection` are intentionally outside the supported internal API surface per PR #22696 / PR #23401 — the new accessor must remain package-internal.
- `resetPendingDeltaToOps`'s INSERT arm has historically attracted subtle correctness bugs: PR #9944 (wrong remove condition), PR #22069 (duplicated `break;` in a near-identical extraction), PR #22080 (obliterate reconnect), PR #24253 (obliterate resubmission). The new helpers replace four switch exits — future edits should re-verify the exit mapping before merging.

- **Suggested extra eyes** — Abe27342 (caught the duplicated `break;` in PR #22069's near-identical extraction; added the squash/`0xbaf` path in PR #24253), scarlettjlee (introduced the positional-pairing invariant in PR #11961 that `previousPropsForSegment` formalises), wes-carlson (author of the PR #9944 "skip annotate regen on remotely removed segments" behaviour preserved verbatim in `resetAnnotateOp`).
- **Re-audit the four control-flow exits** — ANNOTATE `!isRemovedAndAcked` → `createAdjustRangeOp` / `createAnnotateRangeOp` vs `undefined`; INSERT `isSquashedOp` → `undefined`; INSERT squash+remove → `undefined`; INSERT remote-obliterate → `undefined`; INSERT fall-through → `createInsertSegmentOp`. Equivalence already confirmed; a fresh eye on the merge gate is cheap insurance given the area's bug history.
- **Confirm the staging-mode squash PR scope** — whether the cross-file callsite migrations (`mergeTree.ts` rollback at ~2479, `client.ts` rebased-group construction at ~1277) land soon enough that `previousPropsForSegment` doesn't become unused public surface (PR #23401 trimmed `SegmentGroup` from the internal API specifically to avoid that).
- **Latent over-copy at `client.ts:~1277`** — the rebased `SegmentGroup` is constructed with `segmentGroup.previousProps?.slice(0)` even though only one segment is enqueued into the rebased group at a time. Pre-existing and out-of-scope for this PR (rollback and `enqueueOnCopy` tolerate the over-long array via `indexOf`), but worth fixing in the squash PR by constructing with `[ previousPropsForSegment(originalGroup) ]` — design call on whether that belongs in the squash PR or a separate follow-up.

</details>

### Questions for the author

1. Will the upcoming staging-mode squash PR migrate `enqueueOnCopy` (and the rebased-group construction at `client.ts:~1277` / rollback at `mergeTree.ts:~2479`) to use `previousPropsForSegment`, or will the `this.segment` vs `sourceSegment` asymmetry require a second accessor variant? Answer would confirm the new method is the right shape before downstream callers depend on it.
2. Would you accept a ~5-line unit test (per the inline thread on `segmentGroupCollection.ts`) in this PR to lock in the silent-`undefined`-over-assert contract before the squash PR migrates callers?

packages/dds/merge-tree/src/segmentGroupCollection.ts
```diff
@@ -48,6 +49,23 @@ export class SegmentGroupCollection {
...
+	 */
+	public previousPropsForSegment(segmentGroup: SegmentGroup): PropertySet | undefined {
+		if (segmentGroup.previousProps === undefined) {
+			return undefined;
+		}
+		const index = segmentGroup.segments.indexOf(this.segment);
+		if (index === -1) {
+			return undefined;
+		}
+		return segmentGroup.previousProps[index];
```
anthony-murphy: Deep Review: `previousPropsForSegment` adds three observable branches — `previousProps === undefined` → `undefined`, `indexOf === -1` → `undefined`, and the happy path returning `segmentGroup.previousProps[index]` — but the PR touches only production code; there's no `segmentGroupCollection.spec.ts` hunk and no other test exercises the new method directly.

The silent-`undefined`-over-assert shape is a deliberate design choice (consistent with `SegmentGroup` being intentionally internal per PR #22696 / PR #23401), but right now only the docstring locks it in. A future drift toward an assert-on-miss variant would go uncaught until a downstream caller broke, and the upcoming squash PR is about to start depending on this contract.

A ~5-line direct test covering the three branches — `previousProps === undefined` → `undefined`, segment not in group → `undefined`, happy-path positional pairing returns `segmentGroup.previousProps[index]` for the correct index — closes the gap before the first migrated caller lands. Either in this PR or paired with the first caller migration in the squash PR.

anthony-murphy:
**Deep Review** is diving into the unknown.

- ✅ Mechanical Review
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

packages/dds/merge-tree/src/segmentGroupCollection.ts
```diff
@@ -48,13 +49,21 @@ export class SegmentGroupCollection {
 		walkList(this.segmentGroups, (sg) => segmentGroups.enqueueOnCopy(sg.data, this.segment));
 	}
 
+	/**
+	 * Returns the previousProps entry paired with this collection's segment within the given
+	 * segmentGroup, or undefined if the group has no previousProps or no entry exists for the segment.
+	 */
+	public previousPropsForSegment(segmentGroup: SegmentGroup): PropertySet | undefined {
+		return segmentGroup.previousProps?.get(this.segment);
+	}
```
anthony-murphy, line 58: Deep Review: The PR description states the accessor *"encapsulates the … lookup"* and that callers become a clean one-liner, but the two natural call sites that landed in the same PR continue to access `previousProps` directly — `mergeTree.ts:2479` uses `pendingSegmentGroup.previousProps!.get(segment)` and `client.ts:~1286` uses `segmentGroup.previousProps.get(segment)`. `grep previousPropsForSegment` across `packages/dds/merge-tree` returns only this declaration; the encapsulation goal isn't realized in this commit.

Blast radius is narrow — `index.ts` does not re-export `SegmentGroupCollection`, so this isn't package-public surface — but the description-vs-diff inconsistency is real and PR #11961's review precedent was to keep copy-only `previousProps` plumbing private until a consumer exists.

Either (a) migrate `mergeTree.ts:2479` and `client.ts:~1286` to `segment.segmentGroups.previousPropsForSegment(segmentGroup)` so the encapsulation lands here, or (b) drop the accessor until the consumer lands in the squash PR.

packages/dds/merge-tree/src/mergeTree.ts
```diff
@@ -2476,7 +2475,11 @@ export class MergeTree {
 						{ op: removeOp, rollback: true },
 					);
 				} /* op.type === MergeTreeDeltaType.ANNOTATE */ else {
-					const props = pendingSegmentGroup.previousProps![i];
+					const props = pendingSegmentGroup.previousProps!.get(segment);
+					assert(
+						props !== undefined,
+						"Segment missing previousProps entry on annotate rollback",
+					);
```
anthony-murphy, line 2482: Deep Review: This new `assert` uses a raw string, but every other `assert(...)` in `mergeTree.ts` uses a numbered hex tag (`0x036`, `0x037`, `0x043`–`0x04f`, `0x751`, `0x86c`, `0xa40`, `0xa6e`, `0xb5c`, `0xb5d`, `0xb6e`–`0xb73`, `0xbaf` — 16+ examples). This very PR preserves `0x036`, `0x037`, `0xbaf`, `0xb5c` byte-identical in the extracted `client.ts` helpers, so the convention is intentional.

The repo ships `policy-check:asserts` (`flub generate assertTags`) that rewrites raw strings to hex tags, so this either gets rewritten before merge or trips the policy gate. Tags are how `assertionShortCodesMap` maps production failures back to messages without shipping the strings.

Run `npm run policy-check:asserts` before merge, or tag manually now:

```suggestion
            const props = pendingSegmentGroup.previousProps!.get(segment);
            assert(
                props !== undefined,
                0xXXXX /* Segment missing previousProps entry on annotate rollback */,
            );
```

packages/dds/merge-tree/src/client.ts
```diff
@@ -1270,11 +1283,19 @@ export class Client extends TypedEventEmitter<IClientEvents> {
...
 
 			if (newOp) {
+				let newPreviousProps: WeakMap<ISegmentLeaf, PropertySet> | undefined;
+				if (segmentGroup.previousProps) {
+					const sourceProps = segmentGroup.previousProps.get(segment);
+					newPreviousProps = new WeakMap();
+					if (sourceProps !== undefined) {
+						newPreviousProps.set(segment, sourceProps);
+					}
+				}
```
anthony-murphy, line 1293: Deep Review: This hunk replaces `previousProps: segmentGroup.previousProps?.slice(0)` with a freshly constructed `WeakMap` carrying only the current segment's entry — the PR description identifies the old `.slice(0)` as a latent inconsistency (defensive copy of the props array, but the surrounding loop only ever wrote a single segment to the new group, so the array shape was already mispaired post-copy). The PR ships with no test changes.

Closest existing coverage (`resetPendingSegmentsToOp.spec.ts:189-211`, `client.rollback.spec.ts:206-245` and `:658-688`) doesn't inspect per-segment `previousProps` entries. Prior bugs in this exact reset/rebase path were fuzz-discovered, not unit-discovered (#11946, #22069), so existing unit tests are an unreliable safety net for semantic regressions here.

Add a focused regression in `resetPendingSegmentsToOp.spec.ts` (or `client.rollback.spec.ts`) that constructs a pending annotate group with multiple segments, triggers `resetPendingDeltaToOps` reissue, and asserts each new per-segment `SegmentGroup` has a `previousProps` WeakMap containing exactly the entry for its own segment.

anthony-murphy:
Deep Review: The missing-tests concern is resolved — this commit adds three direct cases for `previousPropsForSegment` in `test/segmentGroupCollection.spec.ts` (the `undefined` / not-in-group / happy-path branches called out above). Remaining accessor concern — that the two production lookup sites at `mergeTree.ts:2479` and `client.ts:~1286` still bypass the accessor — is tracked separately in the inline thread on `segmentGroupCollection.ts:58`.

packages/dds/merge-tree/src/segmentGroupCollection.ts
```diff
@@ -48,13 +49,21 @@ export class SegmentGroupCollection {
...
 		if (segmentGroup.previousProps) {
-			// duplicate the previousProps for this segment
-			const index = segmentGroup.segments.indexOf(sourceSegment);
-			if (index !== -1) {
-				segmentGroup.previousProps.push(segmentGroup.previousProps[index]);
+			// duplicate the previousProps entry for the destination segment, keyed off the source's entry
+			const sourceProps = segmentGroup.previousProps.get(sourceSegment);
+			if (sourceProps !== undefined) {
+				segmentGroup.previousProps.set(this.segment, sourceProps);
 			}
```
anthony-murphy, line 67: Deep Review: `enqueueOnCopy` is the structural successor to PR #11961 (scarlettjlee, 2022-09-15), which fixed the exact split-after-annotate rollback crash: *"when a segment is split after an annotation, there was no corresponding property set for the new segment, causing rollback to crash."* The rewritten copy path — `segmentGroup.previousProps.get(sourceSegment)` → `segmentGroup.previousProps.set(this.segment, sourceProps)` — is the load-bearing replacement for the prior `indexOf` + parallel-array `push` logic.

The new `.copyTo` spec in `test/segmentGroupCollection.spec.ts` only enqueues groups with `previousProps === undefined` (`segmentGroups.enqueue({ segments: [], localSeq: 1, refSeq: 0 })`). Three new tests were added for the trivial `previousPropsForSegment` read accessor; zero for the nontrivial copy logic. Prior bugs on this path (#11946, #22069, #24253) were fuzz-discovered, not unit-discovered — existing unit tests aren't a reliable safety net for a silent regression on the split-then-rollback case.

Distinct from the reissue-path concern on `client.ts:1293` — that thread covers `resetPendingDeltaToOps`; this is the split-via-`copyTo` path.

Add a `.copyTo` (or sibling) case here that constructs a source `SegmentGroup` with a populated `previousProps` WeakMap entry for the source segment, runs the enqueue-on-copy path, and asserts the destination's `SegmentGroup.previousProps.get(destSegment) === sourceProps`. Ideally also exercise `MergeTree.rollback` ANNOTATE end-to-end against the split case PR #11961 originally regressed on.

anthony-murphy:
## Deep Review

*Reviewed commit `3138872` on 2026-06-02.*

Structural-prep refactor (helper extraction + `WeakMap` migration of `previousProps`) remains mechanically sound on this SHA. Score is unchanged from the prior review because no new commits since `c976b85` addressed the three live inline threads: untagged rollback assert (`mergeTree.ts:2479`), `previousPropsForSegment` accessor that no production call site uses, and missing regression coverage on the two behavior-bearing migration hunks (reissue `client.ts:1293` and `enqueueOnCopy` split-path `segmentGroupCollection.ts:67`). These three cluster around the same migration surface and need to land together — routing `mergeTree.ts:2479` and `client.ts:~1286` through `segmentGroups.previousPropsForSegment(...)` simultaneously closes the encapsulation gap and deletes the untagged assert.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- The `WeakMap` migration preserves the `previousProps` reference-sharing invariant established in PR #11961 (alexvy86's question, scarlettjlee + anthony-murphy's answer: *"changing it would mean somehow changing history"*). Reviewers flagging reference-sharing or split-alignment concerns on the new map should be pointed at #11961.
- The reset/rebase path has a recurring history of fuzz-discovered semantic bugs: #9944 (ANNOTATE `!isRemovedAndAcked` guard, preserved verbatim in `resetAnnotateOp`), #11946 (reorder, the original motivation for killing the index-paired `previousProps[]`), #22069 (immediate ancestor of `resetInsertOp`; duplicated `break;` caught in review), #22080 (obliterate reconnect), #24253 (`0xbaf` squash assert). Tags `0x036`, `0x037`, `0xb5c`, `0xbaf` are stable for cross-version telemetry (#24197, #24349) and must not be renumbered — this PR preserves them byte-identical.
- PR #9944's *"don't regenerate annotate op on removed segment"* guard was specifically corrected so the op *is* still regenerated when the local remove came after the annotate; `resetAnnotateOp` preserves that guard and its comment verbatim.
- Both blind reviewer proposals reached for `Map<ISegmentLeaf, PropertySet>` and rejected `WeakMap` ("iteration", "easier to inspect in tests"). Neither rationale holds — rollback always looks up by key from `pendingSegmentGroup.segments`, never iterates the map. The PR's `WeakMap` choice is the better-reasoned one and additionally ties entry lifetime to segment liveness.
- Churn index (6-month window): `client.ts` 1, `mergeTree.ts` 1, `mergeTreeNodes.ts` / `segmentGroupCollection.ts` / `test/segmentGroupCollection.spec.ts` 0. Low-churn area; the bug history motivating parallel-array pairing (PR #11961 split-then-rollback crash) sits in a code path with no direct new test for the populated-`previousProps` case in this PR.
- Suggested extra eyes: **Abe27342** (caught the duplicated `break;` in PR #22069's near-identical extraction; added `0xbaf` in PR #24253; added the PR #11946 reorder step that first put the i-th pairing at risk), **scarlettjlee** (original `previousProps` design owner, PR #11961), **connorskees** (obliterate branch in `resetInsertOp`, PR #18037 / #12578), **wes-carlson** (author of the PR #9944 guard preserved verbatim), **alexvy86** (raised the deep-copy-vs-reference question on PR #11961 that the WeakMap design now answers).

- **Re-audit the four control-flow exits** in `resetAnnotateOp` / `resetInsertOp`: ANNOTATE `!isRemovedAndAcked` → `createAdjustRangeOp` / `createAnnotateRangeOp` vs `undefined`; INSERT `isSquashedOp` → `undefined`; INSERT squash+remove → `undefined`; INSERT remote-obliterate → `undefined`; INSERT fall-through → `createInsertSegmentOp`. Equivalence already confirmed mechanically; fresh eyes in the most bug-prone area of this package are cheap insurance.
- **Confirm staging-mode squash PR scope** — whether the cross-file callsite migrations land soon enough that `previousPropsForSegment` doesn't sit as unused internal surface (PR #23401 trimmed `SegmentGroup` from the internal API specifically to avoid that). Owner-level scope call; affects disposition of the accessor-encapsulation inline thread.
- **Cross-fork / fuzz determinism** — the merge-tree fuzz suite has historically been the safety net for this code path (PR #11946, #22069, #24253). Confirm it still passes on this branch. Cannot be assessed by the pipeline.

</details>

- `c976b85` 2026-05-27 · **6/10** — load-bearing `enqueueOnCopy` rewrite lacks a populated-`previousProps` test; three Tier 3 inline items still live
- `7ac273f` 2026-05-27 · **7/10** — three polish items live inline; PR Notes "no behavior change" contradicts description bullet 3
- `0c08285` 2026-05-27 · **8/10** — pure structural-prep refactor; three inline polish items (accessor encapsulation, untagged assert, missing reissue regression test)
- `8d34c55` 2026-05-27 · **9/10** — pure-refactor prep; one inline polish item on missing accessor test

</details>

PR #27417, opened 2026-05-27
Remove lockfile-less install from include-test-real-service.yml

Abe27342:
## Description

Restructures include-test-real-service.yml so that it no longer runs lockfile-less installations. This also removes the wonky artificial package.json constructed for this pipeline.

Instead, the general approach here is to:

- Check out the FluidFramework repository unconditionally
- Download the full build artifact that we already produce as part of the CI build and unzip it to avoid necessitating a full client build
- Run the tests in the context they normally would be run in microsoft/FluidFramework

There are some small complications to get our logger package (which is imported at test runtime) working.

This change should also make it unnecessary to publish "test artifacts" from a number of our packages. Removing those (and simplifying CI build) will be done in a follow-up.


jenn-le (approved):

tools/pipelines/templates/include-test-real-service.yml
```diff
@@ -287,109 +285,71 @@ stages:
...
         - task: Bash@3
-          displayName: Install base dependencies
+          displayName: Install ff_pipeline_host dependencies
+          retryCountOnTaskFailure: 4
           inputs:
             targetType: 'inline'
             workingDirectory: $(Pipeline.Workspace)/$(FFPipelineHostDirectory)
             script: |
               set -eu -o pipefail
               pnpm install
```
alexvy86: The new install below explicitly uses `--frozen-lockfile`. I don't recall if that's default behavior in recent pnpm, or if we set it in ff_pipeline_host, but maybe good to do it explicitly here anyway?

alexvy86 (approved):
One comment below, but overall I think the approach is good.

alexvy86 (approved):

alexvy86:
Linking AB#6200 just for reference.

jenn-le (approved):

PR #27418, opened 2026-05-27
perf(legacy-dds): use idToEntryMap for entryId lookup in SharedArray

anthony-murphy:
## Description

Adds `findEntryById(entryId): SharedArrayEntry<T> | undefined` that consults the already-maintained `idToEntryMap` (kept correct on every load and `addEntry`; never deleted from). Converts hot-path callers that need an entry — not an array index — to use it.

Sites converted:

- `prepareToMakeEntryIdLive` — neighbor lookups (`prevEntryId`, `nextEntryId`) no longer touch `sharedArray`; they mutate the entries directly via `findEntryById`.
- `handleStashedInsert` — uses `findEntryById`, then `sharedArray.indexOf(entry)` only for the splice position (still O(n), but the lookup itself is O(1)).

`findIndexOfEntryId` is retained for `handleInsertOp`, which feeds the index to `getInternalInsertIndexByIgnoringLocalPendingInserts(index)`; rewriting that helper around an entry reference would require restructuring local-pending tracking and is out of scope.

## Why

Today every remote insert / move / stashed op walks `sharedArray` linearly via `findIndexOfEntryId` despite an already-maintained id → entry Map sitting next to it. For the 100K-entry scenarios called out in the SharedArray docstring, each remote op pays 100K comparisons. This change is independent of squash and stands alone on perf grounds.

## Notes

Pure perf prep — no `reSubmitSquashed`, no squash logic, no tests touched, no api-report changes.

### DIFF
--- a/packages/dds/legacy-dds/src/array/sharedArray.ts
+++ b/packages/dds/legacy-dds/src/array/sharedArray.ts
@@ -635,6 +635,14 @@ export class SharedArrayClass<T extends SerializableTypeForSharedArray>
 			this.getEntryForId(entryId).isAckPending = false;
 		} else {
 			if (insertAfterEntryId !== undefined) {
+				// This call site still uses the index-returning helper because the
+				// result feeds `getInternalInsertIndexByIgnoringLocalPendingInserts`,
+				// which iterates `sharedArray` from a starting array position to skip
+				// over locally-pending inserts. Rewriting that helper in terms of an
+				// entry reference would require restructuring how local-pending state
+				// is tracked, which is out of scope for this change. Switching to
+				// `findEntryById` + `indexOf` here would not be an improvement since
+				// `indexOf` is itself O(n).
 				index = this.findIndexOfEntryId(insertAfterEntryId) + 1;
 			}
 			const newEntry = this.createNewEntry(entryId, value);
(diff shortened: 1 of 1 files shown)


anthony-murphy:

anthony-murphy:
## Deep Review

*Reviewed commit `2e4e99c` on 2026-05-27.*

Clean, well-motivated perf prep — routes entry-only callers through the existing `idToEntryMap` via a new `findEntryById`. No correctness or API concerns. Two Tier 3 polish items inline. Findings apply to the consolidated prep-<dds> PR you're cherry-picking this commit into.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Either land a tail-append fast path in `handleInsertOp` or amend the PR description to call out that `handleInsertOp` is explicitly out of scope, so the 100K bulk-append motivation isn't read as fully delivered
- [ ] Add the two targeted regressions called out inline (neighbor rewire with missing prev/next id; stashed insert with missing `insertAfterEntryId`), or run PR #25281's stress suite against this change and note the result in the description

### Context for Reviewers

- SharedArray was migrated from the Loop repo in #24661 (dannimad) with intentionally preserved shape; the `idToEntryMap` this PR leans on was introduced by that migration and is kept in sync on load and `addEntry`, never deleted from. Review as a behavior-preserving legacy perf change.
- Low-churn area: `sharedArray.ts` has 2 commits in the last 6 months across 3 related PRs — #24661 (dannimad, migration), #24711 (tylerbutler, assert tagging), #25281 (dannimad, local-server stress harness for SharedArray).

- **dannimad sign-off** — original migrator (#24661) and author of `handleStashedInsert` plus the stress harness (#25281); best-positioned to vet the map↔array invariant this PR now leans on and the stashed-insert equivalence.
- **tylerbutler awareness** — this PR retires tagged asserts 0xb97/0xb98; the short-code registry already records them, but the assert-tagging pipeline owner may want visibility on the retirement pattern.
- **Stress-run judgment** — whether running PR #25281's stress suite locally or in CI is an acceptable substitute for adding the two targeted regressions; that's a maintainer taste call on test-debt accounting in legacy-dds.

</details>

packages/dds/legacy-dds/src/array/sharedArray.ts
```diff
@@ -635,6 +635,14 @@ export class SharedArrayClass<T extends SerializableTypeForSharedArray>
...
 			if (insertAfterEntryId !== undefined) {
+				// This call site still uses the index-returning helper because the
+				// result feeds `getInternalInsertIndexByIgnoringLocalPendingInserts`,
+				// which iterates `sharedArray` from a starting array position to skip
+				// over locally-pending inserts. Rewriting that helper in terms of an
+				// entry reference would require restructuring how local-pending state
+				// is tracked, which is out of scope for this change. Switching to
+				// `findEntryById` + `indexOf` here would not be an improvement since
+				// `indexOf` is itself O(n).
 				index = this.findIndexOfEntryId(insertAfterEntryId) + 1;
```
anthony-murphy, line 646: Deep Review: The in-line comment justifies deferring `handleInsertOp` on the grounds that downstream `getInternalInsertIndexByIgnoringLocalPendingInserts` is also O(n). That holds in general but misses the bulk-append fast path: when the after-entry is the current tail (`sharedArray[length-1]`), the index is simply `length`, and on a fresh remote receiver during 100K bulk append the local-pending queue is empty/tiny, so end-to-end the named scenario becomes O(1).

`handleInsertOp` is the dominant remote-insert path and is what the PR description's "100K comparisons per remote op" motivation actually points at. Two options:

1. Land a follow-up tail check using the already-introduced `findEntryById` — e.g. `if (insertAfterEntry === this.sharedArray[this.sharedArray.length - 1]) index = this.sharedArray.length; else index = this.sharedArray.indexOf(insertAfterEntry) + 1;` — which closes the headline scenario.
2. Amend the PR description to acknowledge that `handleInsertOp` is explicitly out of scope so the 100K motivation isn't read as fully delivered by this change.

Non-blocking — defensible at the line level; flagged so the perf opportunity isn't lost when this commit lands in the consolidated prep-<dds> PR.

PR #25126, opened 2025-07-31
feat(directory): Directory rollback across remote ops (Part 1 -  storage ops only)

scottn12:
## Description

This PR is a prototype for rollback across remote ops for directory. This PR is based off of https://github.com/microsoft/FluidFramework/pull/25070 and implements rollback for storage ops (set, delete, clear).

Note: Rollback acrros remote ops for subdirectories is **not** implemented in this PR and will be included in a follow up.

## Misc
[AB#43738](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/43738)
Prototype: https://github.com/microsoft/FluidFramework/pull/25092
Will fix [AB#38522](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/38522)

PR #27419, opened 2026-05-27
perf(dds-map): use metadata identity for local set/delete ack lookups

anthony-murphy:
## Description

Eliminates the `pendingData.findIndex(entry => entry.type !== "clear" && entry.key === op.key)` scans at four local-op-ack sites in `MapKernel` and `SubDirectory`. The `localOpMetadata` already carries the relevant `PendingKeyDelete` / `PendingKeySet`; the new code uses reference identity instead.

Sites converted:

- `mapKernel.ts` delete-ack: `localOpMetadata` IS the `PendingKeyDelete`; locate with `pendingData.indexOf(localOpMetadata)`.
- `mapKernel.ts` set-ack: `localOpMetadata` is the `PendingKeySet`; locate its parent lifetime via `find(entry => entry.type === "lifetime" && entry.keySets[0] === localOpMetadata)`.
- Same conversions in `directory.ts` for both ack paths.

The empty-lifetime splice still falls back to `pendingData.indexOf(pendingEntry)`, but only after the lifetime has been located by reference equality.

## Why

The read/process counterpart to the rollback fix already landed on `prep-map`. Today each local-op ack walks `pendingData` linearly by key; with the metadata in hand, the lookup is direct.

A future follow-up that adds a `lifetime` back-pointer on `PendingKeySet` (or that hands the ListNode through `localOpMetadata`) makes the set-ack lookup O(1) too — out of scope here.

## Notes

Pure perf prep — no `reSubmitSquashed`, no squash logic, no tests touched, no api-report changes.

### DIFF
--- a/packages/dds/map/src/directory.ts
+++ b/packages/dds/map/src/directory.ts
@@ -1974,15 +1974,14 @@ class SubDirectory extends TypedEventEmitter<IDirectoryEvents> implements IDirec
 			return;
 		}
 		if (local) {
-			const pendingEntryIndex = this.pendingStorageData.findIndex(
-				(entry) => entry.type !== "clear" && entry.key === op.key,
-			);
-			const pendingEntry = this.pendingStorageData[pendingEntryIndex];
 			assert(
-				pendingEntry !== undefined &&
-					pendingEntry.type === "delete" &&
-					pendingEntry.key === op.key,
-				0xc05 /* Got a local delete message we weren't expecting */,
+				localOpMetadata !== undefined && localOpMetadata.type === "delete",
+				"Got a local delete message we weren't expecting",
+			);
+			const pendingEntryIndex = this.pendingStorageData.indexOf(localOpMetadata);
+			assert(
+				pendingEntryIndex !== -1,
+				"Local delete metadata not found in pending storage data",
 			);
 			this.pendingStorageData.splice(pendingEntryIndex, 1);
 			this.sequencedStorageData.delete(op.key);
@@ -2033,20 +2032,27 @@ class SubDirectory extends TypedEventEmitter<IDirectoryEvents> implements IDirec
 		const { key } = op;
 
 		if (local) {
-			const pendingEntryIndex = this.pendingStorageData.findIndex(
-				(entry) => entry.type !== "clear" && entry.key === key,
-			);
-			const pendingEntry = this.pendingStorageData[pendingEntryIndex];
 			assert(
-				pendingEntry !== undefined && pendingEntry.type === "lifetime",
-				0xc06 /* Couldn't match local set message to pending lifetime */,
+				localOpMetadata !== undefined && localOpMetadata.type === "set",
+				"Got a local set message we weren't expecting",
+			);
+			// Locate the parent lifetime by reference identity: it is the entry whose
+			// keySets array contains this exact PendingKeySet instance as its head.
+			const pendingEntry = this.pendingStorageData.find(
+				(entry): entry is PendingKeyLifetime =>
+					entry.type === "lifetime" && entry.keySets[0] === localOpMetadata,
+			);
+			assert(
+				pendingEntry !== undefined,
+				"Couldn't match local set message to pending lifetime",
 			);
 			const pendingKeySet = pendingEntry.keySets.shift();
 			assert(
-				pendingKeySet !== undefined && pendingKeySet === localOpMetadata,
-				0xc07 /* Got a local set message we weren't expecting */,
+				pendingKeySet === localOpMetadata,
+				"Got a local set message we weren't expecting",
 			);
 			if (pendingEntry.keySets.length === 0) {
+				const pendingEntryIndex = this.pendingStorageData.indexOf(pendingEntry);
 				this.pendingStorageData.splice(pendingEntryIndex, 1);
 			}
 			this.sequencedStorageData.set(key, pendingKeySet.value);
--- a/packages/dds/map/src/mapKernel.ts
+++ b/packages/dds/map/src/mapKernel.ts
@@ -767,16 +767,14 @@ export class MapKernel {
 				const { key } = op;
 
 				if (local) {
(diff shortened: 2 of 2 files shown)


anthony-murphy:

anthony-murphy:
## Deep Review

*Reviewed commit `c81ef4c` on 2026-05-27.*

Well-scoped perf-prep refactor — no Tier 1 or Tier 2 issues. Noted the PR is closed for consolidation into the per-DDS prep PR; the items below still apply to whatever sha the commit lands on after cherry-pick.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Decide on the `PendingKeySet.lifetime` back-pointer — either roll the one-field addition into this commit (eliminates both remaining O(n) walks on the set-ack hot path) or link an explicit follow-up issue from the PR description. Both independent design passes reached for the same back-pointer.
- [ ] Add a PR-description note naming the six retired short codes (`0xbf7`–`0xbf9` from `mapKernel.ts`, `0xc05`–`0xc07` from `directory.ts`) and that the predicate restructuring is why they could not be preserved verbatim, so the next `policy-check:asserts` pass and `assertionShortCodesMap.ts` cleanup is unambiguous.
- [ ] Confirm the Map and Directory fuzz suites (with reconnect + rebasing) ran green on this branch and note that in the PR description or on the consolidating PR — given PR #25066's stress-test-driven revert history in this same machinery.
- [ ] Note in the PR description that delete-ack also remains O(n) (not just set-ack) so reviewers calibrate expectations.

### Context for Reviewers

- The FIFO-ack invariant (`keySets[0] === localOpMetadata`) is now load-bearing for the set-ack *lookup itself*, not just a downstream sanity check. Confirm it still holds across `prep-map`'s resubmit/squash paths.
- Untagged string-literal asserts are the established author-time convention (PRs #24973, #25150, #25225); the next `policy-check:asserts` run retags new messages and prunes stale `assertionShortCodesMap.ts` entries.
- This is the partial follow-through on anthony-murphy's own 2025-06-24 review comment on PR #24902 ("use a linked list … O(1) vs O(n)"). The O(1) goal is explicitly deferred here; both ack paths stay O(n) with better constants.
- Pending-lifetime changes have prior stress-test-driven revert history (#24961 → reverted by #25066 → restored by #25070), and PR #25126 ported the model to `SubDirectory`. This PR changes both kernels' ack-removal logic and adds no new tests.
- Suggested reviewers: **ChumpChief** (owns `PendingDataEntry`/`PendingKeyLifetime`/`PendingKeySet` from #24902, #25070), **scottn12** (directory rollback, #25126), **alexvy86** (raised "deduping the asserts is not correct" on #25150 — courtesy ping before retiring the six tagged codes).

- **Needs human judgment** — Scope call on the `PendingKeySet.lifetime` back-pointer (S2 above). Design-deferral disagreement, not a bug. Two independent design passes added the back-pointer; the author's stated "no shape change to pending entry types" rationale is reasonable but in tension with this PR's perf-prep goal. Owner call.
- **Cannot be assessed by the pipeline** — Whether the FIFO-ack invariant holds across `prep-map`'s resubmit/squash paths now that it drives the lookup itself. Needs ChumpChief / scottn12.
- **Cannot be assessed by the pipeline** — Whether retiring `0xbf7`–`0xbf9` and `0xc05`–`0xc07` is acceptable to alexvy86 given his prior position on assert dedupe (#25150).

</details>

packages/dds/map/src/mapKernel.ts
```diff
@@ -808,21 +806,27 @@ export class MapKernel {
...
 					const pendingKeySet = pendingEntry.keySets.shift();
 					assert(
-						pendingKeySet !== undefined && pendingKeySet === localOpMetadata,
-						0xbf9 /* Got a local set message we weren't expecting */,
+						pendingKeySet === localOpMetadata,
+						"Got a local set message we weren't expecting",
 					);
 					if (pendingEntry.keySets.length === 0) {
+						const pendingEntryIndex = this.pendingData.indexOf(pendingEntry);
 						this.pendingData.splice(pendingEntryIndex, 1);
```
anthony-murphy, line 830: Deep Review: The empty-lifetime cleanup here uses `pendingStorageData.indexOf(pendingEntry)` then `splice(pendingEntryIndex, 1)` with no `!== -1` guard. The sibling delete-ack path added in this same PR (and the `directory.ts` equivalents) do assert `pendingEntryIndex !== -1` before splicing. By construction `pendingEntry` was just returned from `find` over the same array with no intervening mutation, so `indexOf` cannot return `-1` today — but `splice(-1, 1)` silently removes the last element if it ever did, which is the hard-to-diagnose corruption mode the belt-and-suspenders pattern from PR #24973 was designed to catch.

Add the matching assert here (and at the parallel site in `directory.ts`) so all four shift/remove sites in this diff use the same guard.

packages/dds/map/src/directory.ts
```diff
@@ -2033,20 +2032,27 @@ class SubDirectory extends TypedEventEmitter<IDirectoryEvents> implements IDirec
...
 			assert(
-				pendingKeySet !== undefined && pendingKeySet === localOpMetadata,
-				0xc07 /* Got a local set message we weren't expecting */,
+				pendingKeySet === localOpMetadata,
+				"Got a local set message we weren't expecting",
 			);
 			if (pendingEntry.keySets.length === 0) {
+				const pendingEntryIndex = this.pendingStorageData.indexOf(pendingEntry);
 				this.pendingStorageData.splice(pendingEntryIndex, 1);
 			}
```
anthony-murphy, line 2057: Deep Review: Same asymmetry as the `mapKernel.ts` empty-lifetime branch — `indexOf(pendingEntry)` then `splice` with no `!== -1` guard, while the delete-ack path added in this same PR does assert it. Safe today (no intervening mutation between `find` and `indexOf`), but matching the established guard pattern across all four shift/remove sites makes the next refactor's failure mode loud instead of silent.

PR #19862, opened 2024-02-28
fix(matrix): Properly detect row/cell deletion when resubmitting a set op

Abe27342:
## Description

This adjusts the matrix resubmit logic for its `setCell` ops. Prior changes didn't correctly rebase over row/column changes in a sensible way.

The general problem here is that when resubmitting a setCell operation, matrix must correctly resolve the cell's location using its row/col permutation vectors. There was one simple thing wrong with this logic: if the row or column for that cell has since been removed, there's nothing to resubmit.

There's also another trickier bug: if the setCell operation was for an unacked row or column, it's possible that while resubmitting the insertion of that row/col, we discovered that it was inserted into a removed range. In this case, for consistency across other clients, merge-tree normalizes the location of the inserted segment to the start of the removed range (that's where the insertion will happen on non-local clients). This means that positional reasoning involving that segment over the window between the original setCell op and the current sequence number won't always resolve correctly.

This PR takes the approach of adding local reference positions on the row/col segments of the setCell op for use during reconnect.

PR #27422, opened 2026-05-27
perf(matrix): use DoublyLinkedList + side-map for pending cell changes

anthony-murphy:
## Description

Replaces `PendingCellChanges<T>.local: { localSeq, value }[]` with `DoublyLinkedList<{ localSeq, value }>` and adds a sibling side-map `localByLocalSeq: Map<number, ListNode<{ localSeq, value }>>` for O(1) `findIndex`-by-`localSeq` lookup.

Sites converted:

- `sendSetCellOp` push: insert into list, register node in side-map.
- `reSubmitCore`: was `findIndex + splice` per pending op; now `localByLocalSeq.get(localSeq)` → `list.remove(node)` → `localByLocalSeq.delete(localSeq)`. O(1).
- ACK shift: `list.shift()?.data` + side-map delete.
- Rollback `pop()` + peek-tail `list.last!.data.value`.
- Length probes use `list.length`.

## Why

For a hot cell with k pending writes, the previous reSubmit-per-op was O(k) find + O(k) splice (O(k²) total) and per-ACK was O(k) shift. For n ops to a single hot cell, the whole rebase pass was O(n²). The list-plus-side-map makes every per-op step O(1).

## Notes

Snapshot/wire format unaffected (`_pendingCliSeqData` is `unknown`). Pure perf prep — no `reSubmitSquashed`, no squash logic, no tests touched, no api-report changes.


anthony-murphy:
## Deep Review

*Reviewed commit `b065a1c` on 2026-05-27.*

Faithful, well-scoped perf prep: `PendingCellChanges<T>.local: []` becomes a `DoublyLinkedList` + `Map<localSeq, ListNode>` side index, turning the per-cell reSubmit hot path from O(k²) to O(k). No correctness defects found. Three polish items remain — the most material being the absence of a focused hot-cell reconnect regression that exercises the path this PR exists to optimize.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add a hot-cell reconnect regression in `packages/dds/matrix/src/test/matrix.reconnect.spec.ts`: disconnect, write the same `(row, col)` N times (e.g. N=200), reconnect, assert convergence. This exercises the exact `localByLocalSeq.get → list.remove(node)` mid-list path the PR exists to optimize and lives in the harness already owned from #24604.

### Context for Reviewers

- Direct continuation of #24604 (same-author follow-up touching the `PendingCellChanges<T>` shape #24604 introduced). Wire/snapshot format is untouched (`_pendingCliSeqData: unknown`); the three rewritten call sites — `reSubmitCore` `findIndex+splice`, `rollback` `pop()`+peek-tail, ACK `shift()`+empty-cell cleanup — all date to #24604.
- Recurring failure modes in this file's history to keep in mind: pending-op identity vs position (Copilot flagged `pop()` and a wrong-index `splice` on #24604), pending-cell order sensitivity (#18018 vladsud), and resubmit rebase over deleted rows/cols (#19862 Abe27342, adjacent to the new `localByLocalSeq.get → remove(node)` in `reSubmitCore`).
- Suggested domain reviewers: **Josmithr** (#24604 design owner, primary inline reviewer), **Abe27342** (resubmit correctness, #19862), **jatgarg** (FWW policy, #18018/#18709), optional **vladsud** and **DLehenbauer**.

- **Encapsulation taste call** — JSDoc-documenting the `local` ↔ `localByLocalSeq` lock-step invariant vs. wrapping the two fields in a small private `PendingLocalChanges<T>` class. Both are correctness-equivalent; this is a maintainability call for the matrix area owner (Josmithr). Two independent design passes both reached for the wrapper.
- **Perf-direction validation** — once a hot-cell regression test exists, confirm the measured improvement matches the O(k²) → O(k) claim under the #24604 stress workloads. The pipeline cannot measure runtime.
- **FWW ACK glance** — the ACK-path conversion from array `shift()` to linked-list `shift()?.data` (matrix.ts:1093-1097) is mechanically equivalent, but a domain-expert look from the FWW owner (jatgarg) is worthwhile given #18709's history around pending-write sequencing.
- **Resubmit-over-deleted-row/col semantics** — Abe27342's #19862 fix area sits adjacent to the new `localByLocalSeq.get → remove(node)` in `reSubmitCore`; the local-ref-position dance is unchanged but worth a confirming glance from the original owner.

</details>

- `09b3415` 2026-05-27 · **6/10** — perf prep with a single hot-cell regression-test blocker

</details>

packages/dds/matrix/src/matrix.ts
```diff
@@ -226,7 +231,12 @@ interface PendingCellChanges<T> {
...
 	 * The local changes including the local seq, and the value set at that local seq.
 	 */
-	local: { localSeq: number; value: MatrixItem<T> }[];
+	local: DoublyLinkedList<{ localSeq: number; value: MatrixItem<T> }>;
+	/**
+	 * Side-map from localSeq to the corresponding node in `local`, enabling O(1) lookup
+	 * during reSubmit instead of a linear `findIndex` scan.
+	 */
+	localByLocalSeq: Map<number, ListNode<{ localSeq: number; value: MatrixItem<T> }>>;
 	/**
```
anthony-murphy, line 295: Deep Review: `localByLocalSeq` is introduced as a sibling field to `local`, and the JSDoc states the *purpose* (O(1) lookup) but not the *invariant* — that `localByLocalSeq` has a key for exactly the nodes in `local` and that every mutation of either must update both. Dual maintenance is now duplicated at five sites: the initializer, `sendSetCellOp` push, `reSubmitCore` removal, rollback, and ACK shift. The PR's own Risks section calls this out: *"every mutation site now has to update both `local` and `localByLocalSeq`. A future contributor could add a new path and forget to update the map."*

The wrapper risk is gated — the type is file-private and every consumer asserts `localSeq` equality on the popped/shifted/looked-up node (`0xba6`/`0xba7`/`0xbaa`/`0xbab`), so a desync would surface as an assertion failure rather than silent corruption. Still, two independent design passes both produced a small private `PendingLocalChanges<T>` wrapper exposing `push` / `removeBySeq` / `shift` / `popLast` / `peekLast` / `length`, and Josmithr previously asked for invariants to be documented on this exact interface (#24604).

Pick one:

- (a) **Minimum** — add a `@remarks` paragraph on `PendingCellChanges<T>` stating the lock-step invariant.
- (b) **Proposer-converged** — wrap `local` + `localByLocalSeq` in a small private class inside `matrix.ts` so the five mutation sites become method calls.

packages/dds/matrix/src/matrix.ts
```diff
@@ -905,12 +919,14 @@ export class SharedMatrix<T = any>
 				const pendingCell = this.pending.getCell(setMetadata.rowHandle, setMetadata.colHandle);
 				assert(pendingCell !== undefined, 0xba9 /* must have pending */);
 
-				const change = pendingCell.local.pop();
+				const change = pendingCell.local.pop()?.data;
 				assert(change?.localSeq === setMetadata.localSeq, 0xbaa /* must have change */);
+				pendingCell.localByLocalSeq.delete(setMetadata.localSeq);
```
anthony-murphy: Deep Review: `pop()` here is correct — rollback is LIFO, the operation being rolled back is always the tail, and assert `0xbaa` enforces the match. But on #24604 the Copilot bot flagged this exact construct (*"Using pop() to remove the pending operation may remove the most recently added op rather than the one corresponding to the rollback"*), and now that this PR adds the `localByLocalSeq` index — making targeted removal trivial and applying it in `reSubmitCore` — a reader will reasonably ask why rollback didn't switch too. Switching would be semantically wrong (rollback must restore the previous cell value, which requires popping the tail — see `matrix.rollback.spec.ts:163-187`, `:232-290`), so the code is right; only the intent is implicit.

A one-line comment above the `pop()` pre-empts the same #24604 round-trip.

```suggestion
			// Rollback is LIFO: the operation being rolled back is always the tail. `pop()` is correct; assert 0xbaa enforces the match.
			const change = pendingCell.local.pop()?.data;
```

packages/dds/matrix/src/matrix.ts
anthony-murphy: Deep Review: The wrapper-class direction (option b) is in the PR — `PendingLocalCellChanges<T>` now owns the `list`/`index` lock-step invariant, and the five mutation sites are method calls on the wrapper. Resolves this thread. One ergonomic leak still open on the wrapper's public surface (`get last` returns `ListNode`, forcing the PR's only `eslint-disable`) — tracked in a new thread.

anthony-murphy:

- ✅ Mechanical Review
- 🌊 **Historical / Contextual Review** — passing through the thermocline…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:
## Deep Review

*Reviewed commit `0a7f3fe` on 2026-06-02.*

Faithful, well-scoped perf prep: `PendingCellChanges<T>.local` is a private `PendingLocalCellChanges<T>` wrapper around `DoublyLinkedList` + `Map<localSeq, ListNode>`, turning the per-cell reSubmit hot path from O(k²) to O(k). The hot-cell reconnect regression test landed at `matrix.reconnect.spec.ts:234-281` (N=200) and the lock-step invariant is resolved by the wrapper-class adoption. No correctness defects. Six Tier 3 polish items remain — none individually material, but they accumulate: the `last`-leaks-`ListNode` ergonomic gap (the PR's only `eslint-disable`), the untagged assert at `matrix.ts:250`, the rollback `pop()`-is-LIFO-only note (open thread `3314510178`), undocumented squash-prep / explicit-removal rationale on the wrapper class, a final-state-only assertion in the new reconnect test, and missing rollback-of-many-pending + FWW-with-many-pending coverage.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Amend PR description Notes to acknowledge the new regression test (`matrix.reconnect.spec.ts`, N=200 same-cell writes is the worst-case O(n²) path this change targets) — current Notes say "no tests touched"

### Context for Reviewers

- Third step on a 5-year `localSeq`-keyed pending-op trajectory: DLehenbauer #2540 (2020) introduced `localOpMetadata`/`localSeq`; Abe27342 #19862 (2024) hardened resubmit with local reference positions; anthony-murphy #24604 (2025) introduced `PendingCellChanges<T>`. This PR is the internal data-structure swap inside that container, not a re-opening of the pending-cell design.
- Wire/snapshot format untouched (`_pendingCliSeqData: unknown`); evaluate as incremental optimisation.
- The prior `pop()` / by-`localSeq` removal concern from #24604 (Copilot reviewer flagged both rollback `pop()` and ACK `pending.splice(1)` as potentially removing the wrong op) is structurally addressed here: `removeByLocalSeq(localSeq)` + dedicated `shift()`/`pop()` semantics and per-call asserts (`0xba6`/`0xba7`/`0xbaa`/`0xbab`) make the by-`localSeq` removal explicit and O(1).
- Rollback (`matrix.ts:974-979`) and ACK (`:1142-1146`) share the changed structure — validation should not focus only on the reSubmit hot path.
- Recurring failure modes in this file's history: pending-op identity vs position (#24604 Copilot), pending-cell order sensitivity on resubmit (#18018 vladsud), resubmit rebase over deleted rows/cols (#19862 Abe27342, adjacent to the new `removeByLocalSeq → list.remove(node)` in `reSubmitCore`), and FWW eventual-consistency surfaced by fuzz (#24652).
- Suggested domain reviewers: **Josmithr** (#24604 design owner; cares about `last`-shape and member docs), **jatgarg** (bulk-tagged matrix asserts in #24797 establishing the convention the new assert breaks; also owns FWW interacting with ACK), **Abe27342** (resubmit correctness #19862), **DLehenbauer** (foundational `localSeq` design #2540), optional **vladsud** (#18018 pending-cell ordering and change-event-sequence contract — relevant to the open test-coverage thread on `cellsChanged` sequence).

- **Encapsulation taste call (Josmithr)** — the wrapper-class direction was already adopted at the design-owner's request (thread `3314510112`, resolved). The remaining `last`-shape ergonomic gap is matrix-area taste; if `peekLastValue()` is acceptable, the inline thread's recommendation lands cleanly.
- **Perf-direction validation** — the pipeline cannot measure runtime. Once a hot-cell micro-bench lands in `packages/dds/matrix/src/test/matrix.bench.ts`, confirm the measured improvement matches the O(k²) → O(k) claim under #24604's stress workloads.
- **FWW ACK glance (jatgarg)** — the ACK-path conversion from array `shift()` to `list.shift()?.data` (`matrix.ts:1142-1146`) is mechanically equivalent but worth a domain-expert look given #18709's pending-write sequencing history.
- **Resubmit-over-deleted-row/col (Abe27342)** — #19862's fix sits adjacent to the new `removeByLocalSeq → list.remove(node)` in `reSubmitCore`. The local-ref-position dance is unchanged but worth a confirming glance.
- **Change-event-sequence contract (vladsud)** — the open inline thread on the new reconnect test asks whether the rewritten resubmit path still exposes every intermediate `cellsChanged` state. Domain-expert confirmation that the contract is unchanged would close the residual risk that the test gap signals.

</details>

- `abd04d4` 2026-05-27 · **8/10** — wrapper-class adoption resolved lock-step invariant; five Tier 3 polish items remained
- `42d0e76` 2026-05-27 · **7/10** — perf prep verified; four polish items, hot-cell regression test most material
- `b065a1c` 2026-05-27 · **7/10** — perf prep verified; three polish items, hot-cell regression test most material
- `09b3415` 2026-05-27 · **6/10** — perf prep with a single hot-cell regression-test blocker

</details>

packages/dds/matrix/src/matrix.ts
```diff
@@ -219,14 +224,74 @@ type FirstWriterWinsPolicy =
...
+	private readonly index = new Map<
+		number,
+		ListNode<{ localSeq: number; value: MatrixItem<T> }>
+	>();
+
+	public get length(): number {
+		return this.list.length;
+	}
+
+	public get last(): ListNode<{ localSeq: number; value: MatrixItem<T> }> | undefined {
```
anthony-murphy, line 245: Deep Review: `PendingLocalCellChanges.last` returns `ListNode<{ localSeq: number; value: MatrixItem<T> }> | undefined`, exposing the internal `DoublyLinkedList` node type as part of the wrapper's public surface. The single call site in `setCellCore` writes `pendingCell.local.last!.data.value` with `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion` — the only such disable in this PR. Both the eslint suppression and the leaky `ListNode` shape come from the same accessor.

Replace `get last(): ListNode<…>` with `peekLastValue(): MatrixItem<T> | undefined` returning `this.list.last?.data.value`. The call site collapses to `pendingCell.local.peekLastValue() ?? pendingCell.consensus`, dropping the `!.data.value` chain and the eslint-disable, and centralising the internal-structure dependency inside the wrapper where it belongs. Josmithr's pattern on #24604 was to push back on exactly this kind of optional/ambiguity surface on the pending-cell interface family.

packages/dds/matrix/src/matrix.ts
```diff
@@ -219,14 +224,74 @@ type FirstWriterWinsPolicy =
...
+		return this.list.last;
+	}
+
+	public push(localSeq: number, value: MatrixItem<T>): void {
+		assert(!this.index.has(localSeq), "duplicate localSeq in PendingLocalCellChanges");
```
anthony-murphy, line 250: Deep Review: `assert(!this.index.has(localSeq), "duplicate localSeq in PendingLocalCellChanges")` is the only plain-string `assert` in the changed file. Every other `assert(...)` in `matrix.ts` carries a hex short-code tag, including the three asserts this PR modifies (`0xba4`–`0xba7`) and the adjacent `0xba8`–`0xbab`. The convention was established by the bulk-tagging pass in #24797 (jatgarg, 2025-06-09); the release-time `assert-short-code` tagger converts plain strings to numeric codes. If that policy check runs in CI for this package, it will fail.

Run the `assert-short-code` tagger (or confirm the intent is to let the next release-time pass tag it). Don't invent a code by hand.

packages/dds/matrix/src/matrix.ts
```diff
@@ -219,14 +224,74 @@ type FirstWriterWinsPolicy =
 			cellLastWriteTracker: SparseArray2D<CellLastWriteTrackerItem>;
 	  };
 
+/**
+ * Encapsulates the list of pending local changes for a single cell, plus the side-index from
+ * `localSeq` to the corresponding `DoublyLinkedList` node enabling O(1) lookup during reSubmit
+ * (instead of a linear `findIndex` scan).
+ *
+ * The list and the index are kept in lock-step: every mutator updates both.
+ */
```
anthony-murphy, line 233: Deep Review: The class JSDoc states the *purpose* (O(1) lookup) but not the *motivation* for the side-map shape. Under today's runtime contract `reSubmitCore` always receives a `localSeq` equal to the head of `pendingCell.local` (because `sendSetCellOp` assigns a fresh `nextLocalSeq()` and pushes to tail, and the runtime walks pending ops FIFO). A `Deque` + head-shift + head-match assert would deliver identical O(1) behavior with no side-map — which is exactly the question the next reader will ask, and the question this dossier converged on.

The justification — explicit removal by `localSeq` (a) avoids encoding an implicit FIFO dependency in a per-cell data structure, and (b) is staged for follow-up squash work that needs arbitrary-position removal — is reasonable but lives only in PR discussion, not in code. Expand the `PendingLocalCellChanges` class comment to state explicitly: (1) why the side-map exists (avoids relying on FIFO/head invariant at call sites; supports upcoming squash/arbitrary-position removal), and (2) the lock-step invariant the wrapper enforces between `list` and `index`. Without this, the next reader will re-derive the "why not a Deque?" critique.

packages/dds/matrix/src/test/matrix.reconnect.spec.ts
```diff
@@ -231,4 +231,54 @@ describe("SharedMatrix reconnect", () => {
...
+		assert.deepEqual(extract(matrix1), [[undefined]]);
+		assert.deepEqual(extract(matrix2), [[N - 1]]);
+
+		// Reconnect matrix2 to force the resubmit path to re-emit all N pending writes.
+		containerRuntime2.connected = true;
+		containerRuntimeFactory.processAllMessages();
+
+		// Both clients should converge on the last-written value.
+		const expected = [[N - 1]];
+		assert.deepEqual(extract(matrix1), expected);
```
anthony-murphy, line 281: Deep Review: `"resubmits N writes to the same cell after reconnect"` stages N=200 pending writes, reconnects, and asserts only `assert.deepEqual(extract(matrix1), [[N - 1]])` and `extract(matrix2), [[N - 1]]` — final-state convergence. The test's own framing ("re-emit all N pending writes") implies a per-op regression target, but no `cellsChanged` observation is wired up. On #18018, vladsud explicitly objected to resubmit logic preserving final state while skipping intermediate set-cell states, because matrix exposes each intermediate state through change events. The PR adds no batching/squash path so the risk is narrowed today — but the gap remains real for the path being rewritten, and a future squash follow-up could regress this silently.

Attach a `cellsChanged` consumer (or a counter on the open-handler) and verify both clients observe all N values in order during the reconnect drain — not just the final one.

packages/dds/matrix/src/test/matrix.reconnect.spec.ts
```diff
@@ -231,4 +231,54 @@ describe("SharedMatrix reconnect", () => {
 		assert.deepEqual(extract(matrix1), expected);
 		assert.deepEqual(extract(matrix3), expected);
 	});
+
```
anthony-murphy, line 234: Deep Review: This PR introduces three new code paths on `PendingLocalCellChanges`: `pop()` (rollback, `matrix.ts:974`), `last` peek (FWW previous-value decision in `setCellCore`), and `shift()` (ACK). The single new test exercises only the reSubmit path. Existing tests cover *single-pending* rollback and *single-pending* FWW, but the "many pending locals to one cell → rollback / FWW conflict" combinations — exactly the paths that depend on the new `last` accessor and on `pop`'s side-map cleanup — are not directly exercised.

Add two tests under the same harness:

- A rollback test that stacks ≥3 pending writes to one cell then undoes them in sequence (exercises `pop` + `last` + side-map cleanup; pairs with the existing `matrix.rollback.spec.ts` patterns).
- An FWW-conflict test where a remote winner arrives while multiple local pendings exist on the same cell (exercises the `length > 0 ? local.last!.data.value : pendingCell.consensus` branch).

Both fit the existing `MockContainerRuntime` harness.

PR #22717, opened 2024-10-02
Optimize anchorset and flex tree traversal

CraigMacomber:
## Description

Optimizations based on a profile of BubbleBench.

This resulted in about 5% more bubbles (80 vs 76).

Given this application's read costs scale with bubbles squared (for the collision logic), contains writes, and most reads are leaf values which are not impacted by this, this seems like a pretty good win. Some real apps could probably show significantly larger gains if they are less leaf heavy or less write heavy.

## Reviewer Guidance

The review process is outlined on [this wiki page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

PR #12338, opened 2022-10-07
Tree: Fix anchor set move

yann-achard-MS:
Fixes a bug in the implementation of `AnchorSet.moveChildren`.
Add doc comments for the invariants that were being violated by the bug.
Updates tests to validate fix.

PR #11284, opened 2022-07-26
Tree: implement AnchorSet.moveChildren

CraigMacomber:
Fill out AnchorSet.moveChildren, and tweak its API to handle insert and delete.

Still needs tests.

PR #27424, opened 2026-05-27
perf(tree): binary-search by parentIndex in PathNode/SparseNode removeChild

anthony-murphy:
## Description

Replaces the linear `indexOf` in `PathNode.removeChild` (`anchorSet.ts`) and `SparseNode.removeChild` (`sparseTree.ts`) with binary search by `child.parentIndex()`. The field arrays are already kept sorted by `parentIndex` (the existing `binaryFind` helpers on the same structs rely on it).

A local `binaryFindIndex` helper is added in each file. I did not extend the existing `binaryFind` because it returns the element (not the index) and has a tuned "guess" fast-path for dense arrays that doesn't apply to the index-by-parentIndex case. Two small focused helpers leave the benchmarked `binaryFind` untouched.

The existing TODOs at both `removeChild` sites are removed. The `splice(idx, 1)` after the search is unchanged (array remove is still O(n) but the dominant cost was the scan).

## Why

The TODO at both sites already called this out: "should do more optimized search (ex: binary search or better) using `child.parentIndex()`." This is invoked from anchor lifecycle (`considerDispose`) — every anchor free in workloads with many siblings paid the scan.

## Notes

Pure perf refactor — no behavior change. The sort-by-`parentIndex` invariant is already maintained by every field mutator (`getOrCreateChild` sorts after push; `coupleNodes` / `decoupleNodes` / `offsetChildren` preserve the order). No tests, changesets, or api-report changes.

### DIFF
--- a/packages/dds/tree/src/core/tree/anchorSet.ts
+++ b/packages/dds/tree/src/core/tree/anchorSet.ts
@@ -1226,9 +1226,8 @@ class PathNode extends ReferenceCountedBase implements AnchorNode {
 		assert(this.status === Status.Alive, 0x40e /* PathNode must be alive */);
 		const key = child.parentField;
 		const field = this.children.get(key);
-		// TODO: should do more optimized search (ex: binary search or better) using child.parentIndex()
-		// Note that this is the index in the list of child paths, not the index within the field
-		const childIndex = field?.indexOf(child) ?? -1;
+		// Note that this is the index in the list of child paths, not the index within the field.
+		const childIndex = field === undefined ? -1 : binaryFindIndex(field, child.parentIndex);
 		assert(childIndex !== -1, 0x35c /* child must be parented to be removed */);
 		field?.splice(childIndex, 1);
 		if (field?.length === 0) {
@@ -1308,6 +1307,32 @@ function binaryFind(sorted: readonly PathNode[], index: number): PathNode | unde
 		}
 	}
 	return undefined; // If we reach here, target is not in array (or array was not sorted)
+}
+
+/**
+ * Find the array index of a child PathNode by its parentIndex using a binary search.
+ * @param sorted - array of PathNode's sorted by parentIndex.
+ * @param index - parentIndex being looked for.
+ * @returns array index of the child with the requested parentIndex, or -1 if not found.
+ */
+function binaryFindIndex(sorted: readonly PathNode[], index: number): number {
+	// inclusive
+	let min = 0;
+	// exclusive
+	let max = sorted.length;
+
+	while (min !== max) {
+		const mid = Math.floor((min + max) / 2);
+		const found = sorted[mid]!.parentIndex;
+		if (found === index) {
+			return mid;
+		} else if (found > index) {
+			max = mid;
+		} else {
+			min = mid + 1;
+		}
+	}
+	return -1;
 }
 
 interface BufferedEvent {
--- a/packages/dds/tree/src/core/tree/sparseTree.ts
+++ b/packages/dds/tree/src/core/tree/sparseTree.ts
@@ -111,9 +111,8 @@ export class SparseNode<TData> implements UpPath {
 	public removeChild(child: SparseNode<TData>): void {
 		const key = child.parentField;
 		const field = this.children.get(key);
-		// TODO: should do more optimized search (ex: binary search or better) using child.parentIndex()
-		// Note that this is the index in the list of child paths, not the index within the field
-		const childIndex = field?.indexOf(child) ?? -1;
+		// Note that this is the index in the list of child paths, not the index within the field.
+		const childIndex = field === undefined ? -1 : binaryFindIndex(field, child.parentIndex);
 		assert(childIndex !== -1, 0x4a5 /* child must be parented to be removed */);
 		field?.splice(childIndex, 1);
 		if (field?.length === 0) {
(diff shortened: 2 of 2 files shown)


anthony-murphy:

anthony-murphy:
## Deep Review

*Reviewed commit `3adfaae` on 2026-05-27.*

Not ready for sign-off. The swap from `indexOf(child)` to `binaryFindIndex(field, child.parentIndex)` drops a free identity check at both `removeChild` sites — a future sort-by-`parentIndex` invariant slip would silently splice the wrong sibling instead of tripping the existing `childIndex !== -1` assert. One-line fix flagged inline at both call sites. PR is closed pending cherry-pick into a consolidated prep PR; the concern applies to that target.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- Builds on the perf-design lineage from #22717 / #23908 (CraigMacomber). The author respected #22717's "do not touch `binaryFind`" warning by adding a separate `binaryFindIndex` rather than generalising the existing helper — that trade-off was already defended on #22717.
- The sort-by-`parentIndex` invariant the new search depends on is real and historically maintained: documented at `sparseTree.ts:23`, actively restored after insert at `sparseTree.ts:88`, and preserved across structural edits by the move/offset bookkeeping from #12338 (yann-achard-MS) and #11284.
- Suggested reviewers: **CraigMacomber** (perf-design owner), **yann-achard-MS** (move-children invariant the safety case rests on), **taylorsw04** (raised the binary-search duplication question on #22717 — likely relevant to a follow-up).

- **Needs human judgment** — whether restoring the identity assert at the two `removeChild` sites is the right invariant boundary, or whether yann-achard-MS prefers the slip surface elsewhere.
- **Cannot be assessed by the pipeline** — microbenchmark evidence on the small-array (`length` 0 / 1) case that #22717 flagged as benchmark-sensitive; CraigMacomber can confirm whether the `removeChild` workload is outside that regime.

</details>

packages/dds/tree/src/core/tree/anchorSet.ts
```diff
@@ -1226,9 +1226,8 @@ class PathNode extends ReferenceCountedBase implements AnchorNode {
 		assert(this.status === Status.Alive, 0x40e /* PathNode must be alive */);
 		const key = child.parentField;
 		const field = this.children.get(key);
-		// TODO: should do more optimized search (ex: binary search or better) using child.parentIndex()
-		// Note that this is the index in the list of child paths, not the index within the field
-		const childIndex = field?.indexOf(child) ?? -1;
+		// Note that this is the index in the list of child paths, not the index within the field.
+		const childIndex = field === undefined ? -1 : binaryFindIndex(field, child.parentIndex);
 		assert(childIndex !== -1, 0x35c /* child must be parented to be removed */);
 		field?.splice(childIndex, 1);
```
anthony-murphy, line 1232: Deep Review: The swap from `field?.indexOf(child) ?? -1` to `binaryFindIndex(field, child.parentIndex)` loses a free identity check. The old form returned -1 unless the exact `child` reference was present, so any future sort-by-`parentIndex` invariant slip (transient sort violation, duplicated `parentIndex` from a bookkeeping bug elsewhere, etc.) tripped the existing `assert(childIndex !== -1, 0x35c, …)` noisily. The new form returns whichever entry matches by key and the subsequent `splice(childIndex, 1)` removes it regardless of identity — converting a fail-fast assertion into silent wrong-sibling removal.

The `decoupleNodes` doc comment a few lines down notes decoupled children are "left at their existing values," so transient inconsistency windows exist by design in this code's history; the identity check was the backstop that made those windows safe.

Restoring the check after the search is a one-line cost (one property compare per `removeChild`, which is not on the hot read path) and keeps the fail-fast behaviour:

```suggestion
		// Note that this is the index in the list of child paths, not the index within the field.
		const childIndex = field === undefined ? -1 : binaryFindIndex(field, child.parentIndex);
		assert(childIndex !== -1, 0x35c /* child must be parented to be removed */);
		assert(field![childIndex] === child, "removeChild: parentIndex matched a different sibling — sort-by-parentIndex invariant violated");
```

packages/dds/tree/src/core/tree/sparseTree.ts
```diff
@@ -111,9 +111,8 @@ export class SparseNode<TData> implements UpPath {
 	public removeChild(child: SparseNode<TData>): void {
 		const key = child.parentField;
 		const field = this.children.get(key);
-		// TODO: should do more optimized search (ex: binary search or better) using child.parentIndex()
-		// Note that this is the index in the list of child paths, not the index within the field
-		const childIndex = field?.indexOf(child) ?? -1;
+		// Note that this is the index in the list of child paths, not the index within the field.
+		const childIndex = field === undefined ? -1 : binaryFindIndex(field, child.parentIndex);
 		assert(childIndex !== -1, 0x4a5 /* child must be parented to be removed */);
 		field?.splice(childIndex, 1);
```
anthony-murphy, line 117: Deep Review: Same identity-check regression as the matching site in `anchorSet.ts` (`PathNode.removeChild`). Pre-PR, `indexOf(child)` returned -1 unless the exact `child` reference was in `field`, so any future sort-by-`parentIndex` invariant slip tripped the existing `assert(childIndex !== -1, 0x4a5, …)` noisily. After the PR, `binaryFindIndex(field, child.parentIndex)` returns whichever entry matches by key, and the `splice(childIndex, 1)` below removes it regardless of identity — converting a fail-fast assertion into silent wrong-sibling removal on the `SparseNode` side too.

The sort invariant is documented at `sparseTree.ts:23` and actively restored at `sparseTree.ts:88`, so the new assert nearly always holds — which is exactly why the check is cheap insurance rather than a hot-path cost. Restore it after the search:

```suggestion
		// Note that this is the index in the list of child paths, not the index within the field.
		const childIndex = field === undefined ? -1 : binaryFindIndex(field, child.parentIndex);
		assert(childIndex !== -1, 0x4a5 /* child must be parented to be removed */);
		assert(field![childIndex] === child, "removeChild: parentIndex matched a different sibling — sort-by-parentIndex invariant violated");
```

PR #25572, opened 2025-09-29
tree: Tolerate concurrent shared branch merges in resubmit

alex-pardes:
## Description

This refactors DefaultResubmitMachine and changes the ResubmitMachine interface to tolerate local revisions being sequenced by a non-local change (due to concurrent shared branch merges).

PR #23732, opened 2025-01-31
(tree) Added op bunching processing to shared tree

agarwal-navin:
## Description
Added op bunching processing capability to shared tree. With op buching, a continguous bunch of ops for a DDS from the same grouped batch is sent to it together. Tree uses this to optimize op processing by rebasing the bunch of ops over the local branch in one go instead of doing it one by one prior to this change. This results in asymptomatic performance improvements as described in the "Benchmarks" section below.

In addition to the tree changes, it also changes the following:
- Extended runtime mocks in shared tree to support op bunching.
- Makes some minor changes to the runtime mocks to support extending it for op bunching.

## Benchmarks
The results from the bechmark test added in `sharedTree.bench.ts`. There is an asymptomatic performance improvement which increases with the number of incoming ops and the number of local ops. In the following runs, the performance improvement ranges from 71% to 94%.
```
Without op bunching:
status  name                                               total time (s)  Batch Count  Iterations per Batch  Period (ns/op)  Margin of Error  Relative Margin of Error
------  -------------------------------------------------  --------------  -----------  --------------------  --------------  ---------------  ------------------------
    ✔   Rebase 10 local commits over 10 inbound commits    5.01                    511                     1    4,107,336.96           ±0.00%                    ±2.20%
    ✔   Rebase 100 local commits over 10 inbound commits   5.04                     78                     1   53,541,414.53           ±0.00%                    ±1.85%
    ✔   Rebase 10 local commits over 100 inbound commits   5.04                    101                     1   36,775,637.37           ±0.00%                    ±3.28%
    ✔   Rebase 100 local commits over 100 inbound commits  5.25                      9                     1  495,604,537.11           ±0.01%                    ±2.14%

With op bunching:
status  name                                               total time (s)  Batch Count  Iterations per Batch  Period (ns/op)  Margin of Error  Relative Margin of Error
------  -------------------------------------------------  --------------  -----------  --------------------  --------------  ---------------  ------------------------
    ✔   Rebase 10 local commits over 10 inbound commits    5.01                    755                     1    1,200,198.57           ±0.00%                    ±2.49%
    ✔   Rebase 100 local commits over 10 inbound commits   5.02                    260                     1   10,675,092.18           ±0.00%                    ±2.11%
    ✔   Rebase 10 local commits over 100 inbound commits   5.01                    298                     1    6,828,631.68           ±0.00%                    ±2.60%
[...]

PR #22921, opened 2024-10-29
Fix bug in EditManager fast forwarding step

noencke:
## Description

This fixes a bug in which the sequence numbers associated with local branches could be advanced when they shouldn't have been. This allowed commits from those branches to be garbage collected while the branches were still alive.

PR #27425, opened 2026-05-27
perf(tree): O(1) revision lookup in editManager via side-map

anthony-murphy:
## Description

Adds a side-map `Map<RevisionTag, number>` on `SharedBranch` (alongside the existing `localCommits: GraphCommit[]`) so `sharedTreeCore.ts:getEnrichedCommit`'s `localCommits.findIndex(c => c.revision === revision)` becomes O(1). The `localCommits` array and `getLocalCommits()` shape are unchanged.

Maintenance:

- On `localCommits.push` (the `afterChange` listener): `localCommitsByRevision.set(commit.revision, localCommits.length + headOffset)`.
- On `localCommits.shift` (in `sequenceLocalCommit`): delete the shifted revision and increment a `localCommitsHeadOffset` counter so other map entries do not need to be touched. The offset resets to 0 when the array fully drains (so it cannot grow unboundedly).
- On bulk rebuild (rebase): clear and refill from the new array.

Exposes `SharedBranch.getLocalCommitIndexByRevision(revision)` and a delegating `EditManager.getLocalCommitIndexByRevision(branchId, revision)`. `sharedTreeCore.ts` uses the new method.

## Why

Burst submit / resubmit storms accumulate large local-commit batches. The `findIndex` per resubmit was O(n) → O(n²) per full reSubmit. The shift-on-ack path was already O(n) for the array; the side-map keeps the lookup itself O(1) without amortizing the shift cost over every other entry.

The head-offset approach was chosen over "decrement-all on shift" because the latter would be O(m) per ACK where m is the depth — defeating the side-map's purpose for the burst case.

## Notes

`getLocalCommits()` return shape unchanged (still `readonly GraphCommit[]`). Pure perf prep — no squash logic, no tests touched, no api-report changes.

### DIFF
--- a/packages/dds/tree/src/shared-tree-core/editManager.ts
+++ b/packages/dds/tree/src/shared-tree-core/editManager.ts
@@ -425,6 +425,19 @@ export class EditManager<
 	}
 
 	/**
+	 * @returns the index of the local commit with the given revision in {@link getLocalCommits} for the given branch,
+	 * or `undefined` if no such commit exists.
+	 * @remarks O(1) alternative to scanning the array returned by {@link getLocalCommits}.
+	 */
+	public getLocalCommitIndexByRevision(
+		branchId: BranchId,
+		revision: RevisionTag,
+	): number | undefined {
+		const branch = this.getSharedBranch(branchId);
+		return branch.getLocalCommitIndexByRevision(revision);
+	}
+
+	/**
 	 * Gets the length of the longest branch maintained by this `EditManager`.
 	 * This may be the length of a peer branch or the local branch.
 	 *
@@ -677,6 +690,21 @@ class SharedBranch<TEditor extends ChangeFamilyEditor, TChangeset> {
 	private readonly localCommits: GraphCommit<TChangeset>[] = [];
 
 	/**
+	 * Side-map for O(1) revision -> {@link localCommits} index lookup.
+	 * @remarks
+	 * Stored values are "absolute" indices that are NOT affected by shifts off the front of {@link localCommits}.
+	 * The actual array index is computed as `value - localCommitsHeadOffset`.
+	 * This avoids having to update every entry in the map when a commit is sequenced (shifted) off the front.
+	 */
+	private readonly localCommitsByRevision = new Map<RevisionTag, number>();
+
+	/**
+	 * The number of commits that have been shifted off the front of {@link localCommits} since the last bulk rebuild.
+	 * Used together with {@link localCommitsByRevision} to translate absolute indices into current array indices.
+	 */
+	private localCommitsHeadOffset = 0;
+
+	/**
 	 * Records extra data associated with sequenced commits.
 	 * This does not include an entry for the {@link trunkBase}.
 	 */
@@ -709,6 +737,10 @@ class SharedBranch<TEditor extends ChangeFamilyEditor, TChangeset> {
 		this.localBranch.events.on("afterChange", (event) => {
 			if (event.type === "append") {
 				for (const commit of event.newCommits) {
+					this.localCommitsByRevision.set(
+						commit.revision,
+						this.localCommits.length + this.localCommitsHeadOffset,
+					);
 					this.localCommits.push(commit);
 				}
 			} else {
@@ -717,6 +749,11 @@ class SharedBranch<TEditor extends ChangeFamilyEditor, TChangeset> {
 					[this.localBranch.getHead(), this.localCommits],
 					this.trunk.getHead(),
 				);
+				this.localCommitsByRevision.clear();
+				this.localCommitsHeadOffset = 0;
+				for (let i = 0; i < this.localCommits.length; i += 1) {
+					this.localCommitsByRevision.set(this.localCommits[i].revision, i);
+				}
 			}
 		});
 	}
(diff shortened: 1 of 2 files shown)


anthony-murphy:

anthony-murphy:
## Deep Review

*Reviewed commit `4ef110d` on 2026-05-27.*

Not ready for sign-off. The algorithm is sound — two independent blind proposers converged on the same revision→absolute-index + head-offset shape — but no targeted unit tests cover the three new mutation modes (append, non-append rebuild, front-shift drain-reset) on a method (`sequenceLocalCommit`) that was rewritten in #22921 to fix a fast-forward correctness bug. Four held-back polish items (bench table, TSDoc precision, duplicate-revision assert, drain-reset invariant comment) sit behind the testing gap. PR is closed for consolidation into the prep-`<dds>` PR; this review applies to whichever PR receives the commit.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Confirm in the consolidated prep-`<dds>` PR whether the test additions, bench table, and TSDoc tightening ride along with this commit or land as a follow-up

### Context for Reviewers

- The `localCommits` cache + `afterChange` `if append / else rebuild` pattern was established by yann-achard-MS in #19272 (2024-01-18); this PR extends that cache with a sibling side-map.
- The `findIndex` lookup this PR replaces was installed by alex-pardes in #25572 (2025-10-01) as the deliberate revision-keyed replacement for a prior sequential-queue approach; the new `assert(... !== undefined, 0xbdb)` preserves the same telemetry tag.
- `sequenceLocalCommit` was rewritten by noencke in #22921 (2024-10-29) to fix a fast-forward correctness bug; noencke's own review note flagged the area as fragile. The new offset/delete bookkeeping in this PR lands directly on top of that previously-broken invariant.
- Prior PRs touching this code path (#19235, #19272, #23732, #24964) led with before/after numbers from `EditManager - Bench No concurrency Many local edits`; the bench rig at `editManager.bench.ts:250-300` already exists.

- **noencke** — owner of the `sequenceLocalCommit` correctness fix (#22921). Best fit to judge whether the testing gap inside that method is acceptable or wants a `debugAssert` backstop inside `getLocalCommitIndexByRevision`.
- **yann-achard-MS** — historical owner of the `localCommits` cache and bench rig (#19235, #19272). Best fit to decide whether the bench table is required-for-merge or sufficient-as-followup for this perf-prep change.
- **alex-pardes** — author of the revision-keyed `findIndex` (#25572). Best fit to confirm the new `!== undefined` failure mode preserves the intended resubmit semantics across concurrent shared-branch-merge cases.
- **Cannot be assessed by the pipeline** — constant-factor and burst-resubmit-storm measurements. The asymptotic O(n)→O(1) argument is sound, but the actual perf win needs either a bench artifact (see Context bullet on #19272's table) or human judgment that the algorithmic argument suffices.

</details>

packages/dds/tree/src/shared-tree-core/editManager.ts
```diff
@@ -709,6 +737,10 @@ class SharedBranch<TEditor extends ChangeFamilyEditor, TChangeset> {
 		this.localBranch.events.on("afterChange", (event) => {
 			if (event.type === "append") {
 				for (const commit of event.newCommits) {
+					this.localCommitsByRevision.set(
+						commit.revision,
+						this.localCommits.length + this.localCommitsHeadOffset,
```
anthony-murphy, line 742: Deep Review: No targeted unit tests cover the three new mutation modes introduced here, and this method has a prior correctness-bug history (#22921, noencke) that makes the gap material rather than polish.

The PR threads three coordinated pieces of mutable state through a non-obvious "absolute index minus head offset" scheme:

1. **Append branch (this site)** — `localCommitsByRevision.set(commit.revision, this.localCommits.length + this.localCommitsHeadOffset)` runs *before* `localCommits.push(commit)`. The read-before-write of `length` is correct only because the stored value is absolute, not relative.
2. **Non-append branch** — clears the map, resets `headOffset = 0`, and refills using the relative index `i`. Correct only because the array is freshly rebuilt and the offset is also reset.
3. **`sequenceLocalCommit` (editManager.ts:965-975)** — `delete(firstLocalCommit.revision)` + `headOffset += 1`, with a drain-reset `if (this.localCommits.length === 0) { this.localCommitsHeadOffset = 0; }`. The drain-reset arm is what keeps `headOffset` from growing unboundedly, but no test exercises it directly.

Existing public-flow tests (`editManagerCorrectness.test.ts:641-690`, `sharedTreeCore.spec.ts:728-763`) transitively exercise append/drain/rebuild, and `defaultResubmitMachine.ts:66-89` asserts `staleChanges[0].newCommit.revision === revision` as a downstream tripwire — that's why this is a Tier 2 testing gap, not a Tier 1 correctness claim. But none of those validate `getLocalCommitIndexByRevision` against `getLocalCommits().findIndex` after a sequence of mixed operations, and none specifically hit the drain-reset arm.

**Suggested fix**, either of:

- Add a focused test in `editManager.spec.ts` that drives appends + `addSequencedChanges` (including a full drain that triggers `headOffset = 0`) + a rebase, asserting `branch.getLocalCommitIndexByRevision(r)` agrees with `branch.getLocalCommits().findIndex(c => c.revision === r)` after each step.
- Or add a `debugAssert` inside `getLocalCommitIndexByRevision` that cross-validates against the array in dev builds, giving every existing public-flow test the cross-check for free.

Worth pulling noencke in for review of the bookkeeping inside `sequenceLocalCommit` regardless of which option lands.

PR #27426, opened 2026-05-27
refactor(merge-tree): use WeakMap for SegmentGroup.previousProps keyed by segment

anthony-murphy:
## Description

Replaces `SegmentGroup.previousProps: PropertySet[]` (paired by index with `SegmentGroup.segments`) with `WeakMap<ISegmentLeaf, PropertySet>`. Kills the two `segments.indexOf(segment)` calls in `segmentGroupCollection.ts` and removes the fragile "i-th `previousProps` entry pairs with i-th `segments` entry" invariant that comments had to call out.

Sites converted:

- `mergeTreeNodes.ts`: type declaration.
- `mergeTree.ts:addToPendingList`: `previousProps = new WeakMap()`.
- `mergeTree.ts` write: `set(segment, propertySet)` instead of array push.
- `mergeTree.ts:2479` (rollback ANNOTATE): `previousProps!.get(segment)` instead of `previousProps![i]`. `i` counter removed.
- `client.ts` (reissue path): the original `.slice(0)` was a defensive copy of the props array, but the surrounding loop only ever wrote a single segment to the new group, so the array shape was already mispaired post-copy (latent inconsistency). Replaced with a new `WeakMap` containing only the entry for the current segment, looked up from the source.
- `segmentGroupCollection.ts:enqueueOnCopy`: drops the `indexOf(sourceSegment)`; maps source's WeakMap value onto destination segment in the destination map.

Truthy-presence semantics (e.g. `op.type === ANNOTATE && !pendingSegmentGroup.previousProps`) work unchanged — a `WeakMap` is truthy.

## Why

The implicit "i-th entry pairs with i-th segment" invariant was a known source of bugs and forced comment annotations. Keying on the segment directly makes the pairing inherent in the type. As a bonus, the two array `indexOf` calls in `segmentGroupCollection.ts` are now O(1).

## Notes

`previousProps` is purely internal merge-tree bookkeeping (not on the wire, not public API). Pure refactor — no behavior change. No tests, changesets, or api-report changes.

### DIFF
--- a/packages/dds/merge-tree/src/client.ts
+++ b/packages/dds/merge-tree/src/client.ts
@@ -1270,11 +1270,19 @@ export class Client extends TypedEventEmitter<IClientEvents> {
 			}
 
 			if (newOp) {
+				let newPreviousProps: WeakMap<ISegmentLeaf, PropertySet> | undefined;
+				if (segmentGroup.previousProps) {
+					const sourceProps = segmentGroup.previousProps.get(segment);
+					newPreviousProps = new WeakMap();
+					if (sourceProps !== undefined) {
+						newPreviousProps.set(segment, sourceProps);
+					}
+				}
 				const newSegmentGroup: SegmentGroup = {
 					segments: [],
 					localSeq: segmentGroup.localSeq,
 					refSeq: this.getCollabWindow().currentSeq,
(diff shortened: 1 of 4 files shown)


anthony-murphy:

PR #20675, opened 2024-04-15
tree: Resubmit encodes with correct schema

taylorsw04:
## Description

This PR fixes tree such that resubmitted ops are encoded with the schema that was in place at the time of their original submission, rather than the current schema.

PR #19272, opened 2024-01-17
Solo editing perf track local branch array

yann-achard-MS:
## Description

Before this PR, the "EditManager - Bench No concurrency Many local edits" benchmark reports the following times on a dev workstation:
```
status  name                                          period (ns/op)  relative margin of error  iterations per batch  batch count  total time (s)
------  --------------------------------------------  --------------  ------------------------  --------------------  -----------  --------------
    ✔   Process the sequencing of 1 local commits           1,317.27                    ±1.42%                     1       21,286            5.00
    ✔   Process the sequencing of 10 local commits          9,143.64                    ±1.00%                     1       11,520            0.92
    ✔   Process the sequencing of 100 local commits       356,412.61                    ±1.00%                     1          912            0.45
    ✔   Process the sequencing of 1000 local commits   27,710,768.29                    ±0.98%                     1           41            1.24
```

After this PR:
```
status  name                                          period (ns/op)  relative margin of error  iterations per batch  batch count  total time (s)
------  --------------------------------------------  --------------  ------------------------  --------------------  -----------  --------------
    ✔   Process the sequencing of 1 local commits           1,163.39                    ±2.21%                     1       19,962            5.00
    ✔   Process the sequencing of 10 local commits          4,308.33                    ±1.94%                     1       20,516            5.00
    ✔   Process the sequencing of 100 local commits        44,794.80                    ±1.00%                     1        5,599            0.95
    ✔   Process the sequencing of 1000 local commits      574,640.66                    ±1.00%                     1        2,319            3.26
```

## Changes

The `EditManager` now maintains an array of commits that represents the changes on the local branch that are not yet on the trunk. This saves the `EditManager` manager from having to re-crawl the local branch every time a commit from the local branch is sequenced.

# Breaking Change
No breaking changes

PR #15746, opened 2023-05-27
Fix bug in SharedTreeBranch replace event check

noencke:
## Description

This PR fixes a mistake in the logic that detects if a `replace` event emitted by a branch originated from a rebase operation or the commit of a transaction. It also does some minor cleanup in related areas.

PR #15177, opened 2023-04-19
Prevent SharedTree trunk commits from being evicted if they are referenced by branches

noencke:
## Description

This PR updates the EditManager's trunk eviction code to retain all commits on the trunk that are currently the base for any outstanding branch/fork. It attempts to do so while preserving separation-of-concerns by adding additional events to the Branch class, but not putting any trunk-specific code into the Branch class (e.g. sequence numbers). It also:

* Deletes the EditManager module in Core and moves EditManager to be part of SharedTreeCore. This is done since EditManager now has a dependency on Branch, which would otherwise introduce a cyclic dependency. Regardless, it is probably an improvement to the module structure even without the motivation provided by this PR.
* Made some methods on SharedTreeCore protected instead of public to reduce the API surface for eventual/future subclassers of SharedTreeCore (e.g. "other SharedTrees"). These methods are now tested via a test subclass in shared-tree-core/utils.ts

PR #27428, opened 2026-05-27
perf(tree): binary-search anchor removeChild + O(1) revision lookup in editManager

anthony-murphy:
## Description

Bundles two tree structural changes:

1. **Binary-search by `parentIndex` in `removeChild`** (`anchorSet.ts` and `sparseTree.ts`) — replaces the linear `indexOf` with binary search by `child.parentIndex()`. The field arrays are already kept sorted by `parentIndex` (the existing `binaryFind` helpers on the same structs rely on it). The existing TODOs at both sites are removed.

2. **`Map<RevisionTag, number>` side-map on `SharedBranch`** (`editManager.ts` + `sharedTreeCore.ts`) — so `sharedTreeCore.ts:getEnrichedCommit`'s `localCommits.findIndex(c => c.revision === revision)` becomes O(1). The `localCommits` array and `getLocalCommits()` shape are unchanged. Maintenance uses a head-offset counter on shift so other map entries don't need to be touched (avoids O(m) decrement-all per ACK).

A local `binaryFindIndex` helper is added in each removeChild file. The existing benchmarked `binaryFind` is left untouched — it returns the element (not the index) and has a tuned dense-array fast-path that doesn't apply to the index-by-parentIndex case.

## Why

- The TODO at both `removeChild` sites called this out: "should do more optimized search (ex: binary search or better) using `child.parentIndex()`." This is invoked from anchor lifecycle (`considerDispose`) — every anchor free in workloads with many siblings paid the scan.
- Burst submit / resubmit storms accumulate large local-commit batches. `findIndex` per resubmit was O(n) → O(n²) per full reSubmit. The side-map keeps the lookup O(1) without amortizing the shift cost over every other entry.

## Notes

`getLocalCommits()` return shape unchanged (`readonly GraphCommit[]`). The sort-by-`parentIndex` invariant on anchor fields is already maintained by every field mutator (`getOrCreateChild` sorts after push; `coupleNodes` / `decoupleNodes` / `offsetChildren` preserve order). Pure perf prep — no squash logic, no tests touched, no api-report changes.


packages/dds/tree/src/core/tree/anchorSet.ts
```diff
@@ -1310,6 +1309,32 @@ function binaryFind(sorted: readonly PathNode[], index: number): PathNode | unde
...
+		if (found === index) {
+			return mid;
+		} else if (found > index) {
+			max = mid;
+		} else {
+			min = mid + 1;
+		}
+	}
+	return -1;
+}
```
anthony-murphy, line 1336: Deep Review: `binaryFindIndex` here is a textbook `[min, max)` search with no fast-path. The sibling `binaryFind` immediately above (anchorSet.ts:1284-1310) opens with `const guess = sorted[index]; if (guess?.parentIndex === index) return guess;`, and its `privateRemarks` (anchorSet.ts:1273-1283) explicitly warns *"current usages tends to fully populate the anchor tree leading the correct array index to be the requested parent index … care and benchmarking should be used when messing with this function."*

The PR description says the fast-path "doesn't apply to the index-by-parentIndex case", but when the field is dense `sorted[index].parentIndex === index`, so the array index of the hit equals `index` — one read instead of `log n` iterations. `removeChild` runs from `considerDispose` on every anchor free, which is exactly the path CraigMacomber tuned in #22717 (*"I started with a more generic binary search which returned the index, but this was so dominated by small data sizes (0 and 1 are the common array sizes here) that it had a lot of overhead…"*, 2024-10-03).

Suggest adding the guess fast-path before the binary loop and updating `binaryFind`'s `privateRemarks` to also cover the new index variant. Re-run the BubbleBench-style anchor benchmarks #22717 used to confirm parity.

```suggestion
function binaryFindIndex(sorted: readonly PathNode[], index: number): number {
	const guess = sorted[index];
	if (guess !== undefined && guess.parentIndex === index) {
		return index;
	}
	let min = 0;
	let max = sorted.length;
	while (min < max) {
		const mid = (min + max) >>> 1;
		const current = sorted[mid].parentIndex;
		if (current === index) {
			return mid;
		} else if (current < index) {
			min = mid + 1;
		} else {
			max = mid;
		}
	}
	return -1;
}
```

packages/dds/tree/src/core/tree/sparseTree.ts
```diff
@@ -145,6 +144,32 @@ export class SparseNode<TData> implements UpPath {
...
+		if (found === index) {
+			return mid;
+		} else if (found > index) {
+			max = mid;
+		} else {
+			min = mid + 1;
+		}
+	}
+	return -1;
+}
```
anthony-murphy, line 173: Deep Review: Same missing fast-path as the `anchorSet.ts` copy. When the field is dense `sorted[index].parentIndex === index`, so the answer is `index` — one read instead of `log n` iterations. The sibling `binaryFind` (and `binaryFind` in `anchorSet.ts`) opens with exactly this guess to avoid the loop on the common dense-array shape that #22717's benchmarks were tuned against. Suggest mirroring the fast-path here and re-running the relevant benchmark to confirm parity.

```suggestion
function binaryFindIndex<TData>(sorted: readonly SparseNode<TData>[], index: number): number {
	const guess = sorted[index];
	if (guess !== undefined && guess.parentIndex === index) {
		return index;
	}
	let min = 0;
	let max = sorted.length;
	while (min < max) {
		const mid = (min + max) >>> 1;
		const current = sorted[mid].parentIndex;
		if (current === index) {
			return mid;
		} else if (current < index) {
			min = mid + 1;
		} else {
			max = mid;
		}
	}
	return -1;
}
```

packages/dds/tree/src/core/tree/anchorSet.ts
```diff
@@ -1226,9 +1226,8 @@ class PathNode extends ReferenceCountedBase implements AnchorNode {
...
 		const key = child.parentField;
 		const field = this.children.get(key);
-		// TODO: should do more optimized search (ex: binary search or better) using child.parentIndex()
-		// Note that this is the index in the list of child paths, not the index within the field
-		const childIndex = field?.indexOf(child) ?? -1;
+		// Note that this is the index in the list of child paths, not the index within the field.
+		const childIndex = field === undefined ? -1 : binaryFindIndex(field, child.parentIndex);
 		assert(childIndex !== -1, 0x35c /* child must be parented to be removed */);
 		field?.splice(childIndex, 1);
 		if (field?.length === 0) {
```
anthony-murphy, line 1233: Deep Review: The previous `field?.indexOf(child)` implicitly cross-checked object identity (it returns `-1` if the exact `child` reference is absent). The replacement `binaryFindIndex(field, child.parentIndex)` returns a hit whenever *some* node with that `parentIndex` exists — if the sort/uniqueness invariant ever drifted (duplicate `parentIndex`, stale `child` reference), the PR would silently splice the wrong node.

Extend the assert to also verify identity. Zero runtime cost in non-debug builds; restores the implicit invariant `indexOf` provided.

```suggestion
		assert(childIndex !== -1 && field![childIndex] === child, 0x35c /* child must be in the field */);
```

packages/dds/tree/src/core/tree/sparseTree.ts
```diff
@@ -111,9 +111,8 @@ export class SparseNode<TData> implements UpPath {
 	public removeChild(child: SparseNode<TData>): void {
 		const key = child.parentField;
 		const field = this.children.get(key);
-		// TODO: should do more optimized search (ex: binary search or better) using child.parentIndex()
-		// Note that this is the index in the list of child paths, not the index within the field
-		const childIndex = field?.indexOf(child) ?? -1;
+		// Note that this is the index in the list of child paths, not the index within the field.
+		const childIndex = field === undefined ? -1 : binaryFindIndex(field, child.parentIndex);
 		assert(childIndex !== -1, 0x4a5 /* child must be parented to be removed */);
 		field?.splice(childIndex, 1);
```
anthony-murphy, line 117: Deep Review: Same identity-loss concern as the `anchorSet.ts` assert. The prior `field?.indexOf(child)` returned `-1` unless the exact `child` reference was present; `binaryFindIndex(field, child.parentIndex)` matches any node with that `parentIndex`. If the sorted-by-`parentIndex` invariant ever drifted (duplicate, stale reference), this would splice the wrong node silently.

```suggestion
		assert(childIndex !== -1 && field![childIndex] === child, 0x4a5 /* child must be in the field */);
```

packages/dds/tree/src/shared-tree-core/editManager.ts
```diff
@@ -424,6 +424,19 @@ export class EditManager<
...
 
+	/**
+	 * @returns the index of the local commit with the given revision in {@link getLocalCommits} for the given branch,
+	 * or `undefined` if no such commit exists.
+	 * @remarks O(1) alternative to scanning the array returned by {@link getLocalCommits}.
+	 */
+	public getLocalCommitIndexByRevision(
+		branchId: BranchId,
+		revision: RevisionTag,
+	): number | undefined {
```
anthony-murphy, line 435: Deep Review: `getLocalCommitIndexByRevision(branchId, revision): number | undefined` leaks the storage shape — callers need to know the result is an index into `localCommits` *and* must reason about the new `headOffset` index space the side-map maintains. The only call site (`sharedTreeCore.ts:558-566`) immediately does `localCommits.slice(revisionIndex)`, so the index is a means to a suffix slice.

Both solution-space reviewers independently designed a `getLocalCommitsFrom(branchId, revision): readonly GraphCommit<TChangeset>[]` returning the suffix slice (or `undefined`/empty). That matches the symmetry of the existing `getLocalCommits(branchId)` accessor at `editManager.ts:422-425`, removes the `undefined`-vs-`-1` sentinel split the PR's own "Risks and tradeoffs" calls out, and keeps `EditManager`'s storage substitutable.

The fair counter is that `getLocalCommits()` already exposes the array shape — but it does not expose `headOffset`, which this accessor effectively does. If you'd rather keep the index API for callers we haven't seen yet, a one-line `// returns an index into getLocalCommits(branchId); undefined if not present` comment here would be enough to ward off the next reader.

packages/dds/tree/src/core/tree/anchorSet.ts
```diff
@@ -1310,6 +1309,32 @@ function binaryFind(sorted: readonly PathNode[], index: number): PathNode | unde
 	return undefined; // If we reach here, target is not in array (or array was not sorted)
 }
 
+/**
+ * Find the array index of a child PathNode by its parentIndex using a binary search.
+ * @param sorted - array of PathNode's sorted by parentIndex.
+ * @param index - parentIndex being looked for.
+ * @returns array index of the child with the requested parentIndex, or -1 if not found.
+ */
+function binaryFindIndex(sorted: readonly PathNode[], index: number): number {
```
anthony-murphy, line 1318: Deep Review: This `binaryFindIndex` and the copy in `sparseTree.ts:150-172` are byte-identical apart from the element type parameter (`PathNode` vs `SparseNode<TData>`), and both rely only on `.parentIndex` — a property on both via `UpPath`. Unlike `binaryFind`, this helper has no tuned fast-path or benchmark surface, so the "don't extract benchmarked code" rationale in the PR description does not apply here. Drift risk is real: a fast-path or bug-fix added to one copy (see the adjacent thread asking for the dense-array guess) will silently miss the other.

Suggest extracting `function binaryFindIndexByParentIndex<T extends { parentIndex: number }>(sorted: readonly T[], index: number): number` to a shared module alongside `UpPath` (or `core/tree/`) and calling it from both `removeChild` sites. If the fast-path lands, the shared helper carries it once — one code change, two beneficiaries.

anthony-murphy:
## Deep Review

*Reviewed commit `a0d2d19` on 2026-06-02.*

Not ready for sign-off. The two perf-prep changes (binary-search `removeChild`; O(1) revision→index side-map on `SharedBranch`) remain structurally correct, but the new `getLocalCommitIndexByRevision` translation arithmetic (`absoluteIndex - localCommitsHeadOffset` at `editManager.ts:935-937`) — the only non-trivial branch the PR adds — is uncovered by the two new tests; both exercise only `offset == 0` paths. A wrong index there silently resubmits the wrong commit suffix via `sharedTreeCore.ts:558-564`. Previously flagged polish (dense-array fast-path, identity assert, accessor shape, helper duplication, benchmark numbers) all stand under their existing inline threads.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Update the PR description — `"no tests touched"` is stale (the anchor sparse-removal test and two new side-map tests already live in this PR, and the offset>0 test from the new inline thread will too)

### Context for Reviewers

- **Side-map continues a deliberate sequencing-perf thread.** #19272 introduced `localCommits` to kill quadratic work in `EditManager`; #20270 added the `getEnrichedCommit` callsite this PR is now optimizing. The `headOffset` trick mirrors the `localCommits.shift()` cache pattern from #19272.
- **Bookkeeping side-structures parallel to `localCommits` are the recurring bug class here.** #15746 (noencke, 2023-05-30) fixed an `afterChange` discriminator bug in `SharedTreeBranch` — this PR's side-map is updated inside a very similar `afterChange` handler (`editManager.ts:737-758`) using `event.type === "append"` as the discriminator. #15177 / #19272 reviews both pushed back on parallel-to-`localCommits` naming and ordering.
- **Resubmit path is correctness-sensitive, not only perf-sensitive.** #20675 (taylorsw04, 2024-04-17) established that resubmit must encode with the schema in place at original submission. A wrong index out of `getLocalCommitIndexByRevision` would slice the wrong commit range under the wrong schema — the same path the new inline thread protects.
- **PR is correctly broader than either solution-space proposer designed.** Both blind proposers scoped `removeChild` only to `anchorSet.ts`; the PR also converts the matching `SparseNode.removeChild` / `getChild` linear scans in `sparseTree.ts`, which share the same sorted-by-`parentIndex` invariant.

- **CraigMacomber / taylorsw04** — judgment call on the missing dense-array fast-path on the two new `binaryFindIndex` copies and on whether they should be unified via a structurally-typed `<T extends { parentIndex: number }>` helper. See inline threads `3314527974`, `3314528053`, `3314863782`.
- **yann-achard-MS / alex-pardes** — confirm `localCommitsByRevision` invariants (append, non-append rebuild, drain reset) match the `localCommits` semantics established in #19272/#19235, and that the `headOffset` design is preferable to decrement-all on `shift`.
- **taylorsw04 / noencke** — confirm the `sharedTreeCore.ts:558-566` resubmit slice is behaviorally identical to the prior `findIndex`-based code under all rebase/fork states (per #20675 schema-encoding sensitivity).
- **Benchmark validation** — re-run the BubbleBench-style anchor harness (#22717) and `EditManager - Bench No concurrency Many local edits` (#19272 harness) to confirm parity / gain on the two paths this PR touches. Cannot be assessed by the pipeline.

</details>

- `87be3cb` 2026-05-27 · **7/10** — binary-search removeChild + side-map both correct; polish on a small diff
- `20bb41a` 2026-05-27 · **7/10** — same two perf changes; dense-array fast-path and identity-assert flagged
- `9b64f11` 2026-05-27 · **7/10** — same two perf changes; five Tier 3 follow-ups flagged

</details>

packages/dds/tree/src/test/shared-tree-core/edit-manager/editManagerCorrectness.test.ts
```diff
@@ -846,6 +846,72 @@ export function testCorrectness(): void {
...
+
+					// After the rebase, the local commits remain (with the same revisions) but the
+					// side-map has been cleared and rebuilt by the clear-and-rebuild branch.
+					const localCommitsAfterRebase = manager.getLocalCommits("main");
+					assert.equal(localCommitsAfterRebase.length, 2);
+					assert.equal(localCommitsAfterRebase[0].revision, local1.revision);
+					assert.equal(localCommitsAfterRebase[1].revision, local2.revision);
+					assert.equal(manager.getLocalCommitIndexByRevision("main", local1.revision), 0);
+					assert.equal(manager.getLocalCommitIndexByRevision("main", local2.revision), 1);
+				});
```
anthony-murphy, line 913: Deep Review: Both new tests cover only the `localCommitsHeadOffset == 0` path, so the new translation arithmetic at `editManager.ts:935-937` (`absoluteIndex - this.localCommitsHeadOffset`) — the only non-trivial branch this PR adds — is never exercised.

- Test 1 ("returns the correct index after all local commits are sequenced and new commits are appended") sequences *every* local commit, which hits the drain-reset branch at `editManager.ts:971-974` and sets `localCommitsHeadOffset = 0`. Subsequent lookups are on freshly appended commits where offset is again 0.
- Test 2 ("returns the correct index after a non-append afterChange event rebuilds the side-map") hits the clear-and-rebuild branch at `editManager.ts:752-756`, which also resets `localCommitsHeadOffset = 0`.

No test reaches the state where `localCommitsHeadOffset > 0` AND `localCommits.length > 0` and then asserts that `getLocalCommitIndexByRevision` returns the right *array* index. That state is the entire reason the translation exists.

This is correctness-sensitive, not just coverage. The production caller `sharedTreeCore.ts:558-564` (`getEnrichedCommit`) does `localCommits.slice(revisionIndex)` — a wrong index here silently resubmits the wrong commit range. #20675 established that resubmit is schema-encoding-sensitive, so a wrong slice means wrong commits resubmitted under the wrong schema. The historical pattern (#15746 `afterChange` discriminator bug; #15177 / #19272 reviewer pushback on parallel-to-`localCommits` structures) is exactly this side-structure-shadowing-an-array bug class.

Suggested test (~15 lines, using primitives already in the file): `applyLocalCommit × 3`, sequence only the first, assert `getLocalCommitIndexByRevision` returns 0 and 1 for the two remaining commits (matching `getLocalCommits()` array indices). Then `applyLocalCommit` a 4th and assert it lands at index 2.

PR #22839, opened 2024-10-17
Op bunching 1: Bunch contiguous ops for data store in a batch - Runtime part

agarwal-navin:
## Reviewer guidance
This is part 1 or 3 of the op bunching feature. The PR with the end-to-end feature is https://github.com/microsoft/FluidFramework/pull/22686. It has been broken down for simpler review process.
This part focuces on the changes in the Runtime layer.

Note - This change breaks the snapshot tests because old snapshots with merge tree had "catchupOps" blobs which contained ops. Now that "metadata" property is not sent to DDS, comparing snapshots fails because of the metadata property not present in latest snapshots for these merge tree instances. A code change is done in snapshot normalizer for this.

## Problem
During op processing, container runtime sends ops one at a time to data stores to DDSes. If a DDS has received M contiguous ops as part of a batch, the DDS is called M times to process them individually. This has performance implications for some DDSes and they would benefit from receiving and processing these M ops together.

Take shared tree for example:
For each op received which has a sequenced commit, all the pending commits are processed by the rebaser. So, as the number of ops received grows, so does the processing of pending commits. The following example describes this clearly:
Currently if a shared tree client has N pending commits which have yet to be sequenced, each time that client receives a sequenced commit authored by another client (an op), it will update each of its pending commits which takes at least O(N) work.
Instead, if it receives M commits at once, it could do a single update pass on each pending commit instead of M per pending commit. 
It can compose the M commits together into a single change to update over, so it can potentially go from something like O (N * M) work to O (N + M) work with batching.

## Solution - op bunching
The solution implemented here is a feature called "op bunching".
With this feature, contiguous ops in a grouped op batch that belong to a data store / DDS will be bunched and sent to it in an array - The grouped op is sent as an `ISequencedMessageEnvelope` and the individual message `contents` in it are sent as an array along with the `clientSequenceNumber` and `localOpMetadata`.
The container runtime will send bunch of contiguous ops for each data store to it. The data store will send bunch of contiguous ops for each DDS to it. The DDS can choose how to process these ops. Shared tree for instance, would compose the commits in all these ops and update pending commits with it.
Bunching only contiguous ops for a data store / DDS in a batch preserves the behavior of processing ops in the sequence it was received.

Couple of behavior changes to note:
[...]

PR #15944, opened 2023-06-10
Preserve batch boundaries upon resubmit

kian-thompson:
## Description

On op resubmit, we drop the opMetadata (which is what holds the batch metadata). This can potentially cause problems with batches getting grouped together into one mega batch.

[AB#3217](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/3217)

PR #27429, opened 2026-05-27
feat(container-runtime): bunch reSubmit dispatch across same-DDS runs

anthony-murphy:
## Description

Extend the bunched dispatch pattern from inbound `processMessages` to the outbound `reSubmit` path. On reconnect or staging-exit, a batch of N contiguous ops targeting the same DDS now makes one trip through ContainerRuntime → ChannelCollection → DataStoreContext → DataStoreRuntime → ChannelContext → DDS instead of N. This mirrors the existing inbound bunching (`ChannelCollection.processChannelMessages`) and gives DDSes that can resubmit a contiguous run efficiently (e.g. SharedTree, MergeTree in follow-ups) the same opportunity on the outbound side.

New `@legacy @beta` surface, additive and backwards compatible:

- `IRuntimeResubmitMessage`, `IRuntimeResubmitMessageCollection` (`runtime-definitions/src/protocol.ts`) — the bunched envelope, with a shared `squash` flag.
- Optional `IFluidDataStoreChannel.reSubmitMessages(type, collection)` (`runtime-definitions`) — bunched alongside the existing `reSubmit`.
- Optional `IDeltaHandler.reSubmitMessages(collection)` (`datastore-definitions`) — bunched alongside the existing `reSubmit`.

`ContainerRuntime.reSubmitBatch` now collects runs of `FluidDataStoreOp` entries and dispatches each run to a new `ChannelCollection.reSubmitContainerMessages`, which sub-bunches by `(address, ddsType)` exactly like `processChannelMessages`. Non-`FluidDataStoreOp` types (Attach, Alias, GC, IdAllocation, BlobAttach, Rejoin, DocumentSchemaChange) continue to use the existing single-op `reSubmit` path. `ChannelDeltaConnection.reSubmitMessages` fans out stashed-op metadata pairs across bunch entries before dispatching, mirroring `getContentsWithStashedOpHandling` for `processMessages`.

DDSes that don't implement `reSubmitMessages` automatically fall back to per-message `reSubmit`. `SharedObject` provides a default that loops over existing `reSubmitCore` / `reSubmitSquashedCore`, so all current DDSes keep working unchanged. Individual DDSes (SharedTree, MergeTree) can override in follow-up PRs to take advantage of seeing the full run together.

Out of scope per scope decision: bunching for `rollback` (constrained by reverse-order checkpoint API) and `applyStashedOp` (async per-message return tangles with PendingStateManager realization loop), and DDS-side overrides of `reSubmitMessages` for SharedTree/MergeTree.

## Reviewer Guidance

- The bunching loop in `ChannelCollection.reSubmitContainerMessages` and the `(address, ddsType)` key intentionally mirrors `processChannelMessages` — worth confirming the symmetry is what we want.
- `ChannelDeltaConnection.reSubmitMessages` flattens stashed-op metadata across entries before dispatching to the handler. The behavior should match `processMessages` + `getContentsWithStashedOpHandling`.
- New tests in `containerRuntime.spec.ts` (Staging Mode describe) exercise contiguous-same-DDS bunching, address-change splits, and `squash` propagation. Full container-runtime suite (968 tests) passes locally.
- Draft pending deep-review sign-off — no DDS currently overrides `reSubmitMessages`, so observable behavior is unchanged but the plumbing exercises the new path through `SharedObject`'s default loop.

### DIFF
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -118,9 +118,11 @@ import type {
 	IGarbageCollectionData,
 	CreateChildSummarizerNodeParam,
 	IDataStore,
+	IEnvelope,
 	IFluidDataStoreContextDetached,
 	IFluidDataStoreRegistry,
 	IFluidParentContext,
+	FluidDataStoreMessage,
 	ISummarizeInternalResult,
 	InboundAttachMessage,
 	NamedFluidDataStoreRegistryEntries,
@@ -5030,9 +5032,36 @@ export class ContainerRuntime
 			: this.reSubmit.bind(this);
 
 		this.batchRunner.run(() => {
-			for (const message of batch) {
-				resubmitFn(message);
+			// Collect contiguous runs of FluidDataStoreOp entries and dispatch each run through
+			// the bunched path so the channel collection can group them by (address, ddsType) for
+			// efficient single-call resubmit at the DDS layer. Non-FluidDataStoreOp entries flush
+			// the current run and use the existing single-op path (squash-aware via resubmitFn).
+			let currentBunch: {
+				envelope: IEnvelope<FluidDataStoreMessage>;
+				localOpMetadata: unknown;
+			}[] = [];
+
+			const flushBunch = (): void => {
+				if (currentBunch.length === 0) {
+					return;
+				}
+				this.channelCollection.reSubmitContainerMessages(currentBunch, squash);
+				currentBunch = [];
+			};
+
+			for (const data of batch) {
+				const message = data.runtimeOp;
+				if (message.type === ContainerMessageType.FluidDataStoreOp) {
+					currentBunch.push({
+						envelope: message.contents,
+						localOpMetadata: data.localOpMetadata,
+					});
+				} else {
+					flushBunch();
+					resubmitFn(data);
+				}
 			}
+			flushBunch();
 		}, resubmitInfo);
 
 		this.flush(resubmitInfo);
--- a/packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
+++ b/packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
@@ -190,6 +190,21 @@ function stubChannelCollection(
 		.stub<Parameters<ChannelCollection["reSubmitContainerMessage"]>, void>()
 		.callsFake(containerRuntime.submitMessage.bind(containerRuntime));
 
+	// Bunched form: replay each entry as a singleton FluidDataStoreOp through submitMessage.
+	const reSubmitBunchedFake = sandbox
+		.stub<Parameters<ChannelCollection["reSubmitContainerMessages"]>, void>()
+		.callsFake((entries, squash) => {
+			for (const { envelope, localOpMetadata } of entries) {
+				containerRuntime.submitMessage(
+					{
+						type: ContainerMessageType.FluidDataStoreOp,
+						contents: envelope,
+					},
+					localOpMetadata,
+				);
+			}
+		});
+
 	const stub = Sinon.createStubInstance(
 		// createSubInstance does not work with property methods (which are
 		// used for stricter typing); so, override via a subclass here.
@@ -199,6 +214,10 @@ function stubChannelCollection(
 				..._args: Parameters<ChannelCollection["reSubmitContainerMessage"]>
 			): void {}
 			// @ts-expect-error -- redefine as instance method for stubbing
+			public reSubmitContainerMessages(
+				..._args: Parameters<ChannelCollection["reSubmitContainerMessages"]>
+			): void {}
+			// @ts-expect-error -- redefine as instance method for stubbing
 			public rollbackDataStoreOp(
 				..._args: Parameters<ChannelCollection["rollbackDataStoreOp"]>
 			) {}
@@ -206,6 +225,7 @@ function stubChannelCollection(
 		{
 			setConnectionState: sandbox.stub(),
 			reSubmitContainerMessage: reSubmitFake,
+			reSubmitContainerMessages: reSubmitBunchedFake,
 			rollbackDataStoreOp: sandbox.stub(),
 			notifyStagingMode: sandbox.stub(),
 			dispose: sandbox.stub(),
@@ -1540,6 +1560,18 @@ describe("Runtime", () => {
 					setConnectionState: (_connected: boolean, _clientId?: string) => {},
 					// Pass data store op right back to ContainerRuntime
 					reSubmitContainerMessage: containerRuntime.submitMessage.bind(containerRuntime),
+					// Bunched form: replay each entry as a singleton through submitMessage.
+					reSubmitContainerMessages: (entries, _squash) => {
+						for (const { envelope, localOpMetadata } of entries) {
+							containerRuntime.submitMessage(
+								{
+									type: ContainerMessageType.FluidDataStoreOp,
+									contents: envelope,
+								},
+								localOpMetadata,
+							);
+						}
+					},
 				} satisfies Partial<ChannelCollection>;
 
 				return patched;
@@ -4487,19 +4519,20 @@ describe("Runtime", () => {
 				controls.commitChanges();
 
 				assert(
-					channelCollectionStub.reSubmitContainerMessage.calledOnce,
-					"Expected reSubmit to be called once. Prestaging op should not be resubmitted",
+					channelCollectionStub.reSubmitContainerMessages.calledOnce,
+					"Expected reSubmitContainerMessages to be called once. Prestaging op should not be resubmitted",
 				);
 				assert(
-					channelCollectionStub.reSubmitContainerMessage.calledWithExactly(
-						{
-							type: ContainerMessageType.FluidDataStoreOp,
-							contents: { address: "2", contents: stagedOpContents },
-						},
-						"LOCAL_OP_METADATA",
+					channelCollectionStub.reSubmitContainerMessages.calledWithExactly(
+						[
+							{
+								envelope: { address: "2", contents: stagedOpContents },
+								localOpMetadata: "LOCAL_OP_METADATA",
+							},
+						],
 						/* squash: */ false, // False by default on commitChanges
 					),
-					"Unexpected args for reSubmit",
+					"Unexpected args for reSubmitContainerMessages",
 				);
 				assert(
 					channelCollectionStub.notifyStagingMode.getCall(1)?.calledWithExactly(false),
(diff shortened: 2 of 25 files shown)


packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -5030,9 +5032,36 @@ export class ContainerRuntime
...
+			// Collect contiguous runs of FluidDataStoreOp entries and dispatch each run through
+			// the bunched path so the channel collection can group them by (address, ddsType) for
+			// efficient single-call resubmit at the DDS layer. Non-FluidDataStoreOp entries flush
+			// the current run and use the existing single-op path (squash-aware via resubmitFn).
+			let currentBunch: {
+				envelope: IEnvelope<FluidDataStoreMessage>;
+				localOpMetadata: unknown;
+			}[] = [];
+
+			const flushBunch = (): void => {
```
anthony-murphy, line 5044: can't we use the bunching helper here

anthony-murphy:
## Deep Review

*Reviewed commit `fa31a15` on 2026-05-27.*

No correctness or security blockers. The outbound `(address, ddsType)` sub-bunching mirrors the inbound `processChannelMessages` pattern cleanly and the optional-method fallbacks keep existing DDSes working unchanged. Six polish items remain inline — mostly doc/naming on the new `@sealed @legacy @beta` shapes (hard to fix after release), one test-coverage gap on the new sub-bunching layer, and one missing protected `SharedObject` seam that SharedTree/MergeTree will hit at merge time.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Answer your own pending question on `containerRuntime.ts:5044` (`forEachContiguousBunch` in `reSubmitBatch`) — if yes, land it here so the inbound/outbound symmetry is complete in one PR
- [ ] Loop in vladsud on the outbound naming decision before merge (owns the inbound `IRuntimeMessageCollection` naming convention this PR diverges from)

### Context for Reviewers

- Completes the outbound half of the inbound op-bunching series #22839/#22840/#22841 — `forEachContiguousBunch` is extracted into `runtime-utils/src/bunching.ts` and reused on the outbound path, the symmetry vladsud asked for during the inbound work.
- Resubmit-path `pendingOpCount` accounting (`dataStoreRuntime.ts:1434-1465`) is the area #15944 had to clean up once; the new bunched path is `old - N + M` and the public contract (`channel.ts:155-158`, `171-182`) permits same/different/multiple/no replacement ops — worth a careful pass when SharedTree/MergeTree later override.
- Kernel-based DDSes inherit the new default through `SharedObjectFromKernel extends SharedObject`; the inherited `attachDeltaHandler` carries the `reSubmitMessages` loop, which reaches the kernel override via `reSubmit` → `reSubmitCore`. No parallel change needed in `sharedObjectKernel.ts`.

- **vladsud sign-off on bunched naming** — relitigated the inbound naming convention on #22839; the outbound naming decision (mirror `messagesContent`/`IRuntimeMessagesContent` vs. deliberate semantic divergence) is their call.
- **agarwal-navin sign-off on the bunching contract symmetry** — owns the inbound bunching contract this PR mirrors; worth confirming the `(address, ddsType)` sub-bunching key and the stashed-op fan-out semantics are intentional.
- **abe27342 on the stashed-op expansion seam** — raised the original fan-out asymmetry concern on #19518; well-placed to review the new bunched interface contract.
- **SharedTree / MergeTree owners** — named as future consumers in the PR description; their input on whether the protected `SharedObject.reSubmitMessagesCore` seam lands here or in follow-ups is the design decision the pipeline cannot make.
- **Runtime-perf assessment** — whether contiguous-bunch dispatch on the outbound path wins back the O(N·M) cost the motivation cites requires real DDS overrides and benchmarks, neither of which exist in this PR.

</details>

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: Good question — `forEachContiguousBunch` would parallel the `channelCollection.ts:847-873` usage and complete the inbound/outbound symmetry the rest of the design rests on. Recommend landing the conversion here rather than as a follow-up, since `reSubmitBatch`'s ad-hoc accumulation is the one place in the outbound path that doesn't yet route through the shared helper.

packages/runtime/runtime-definitions/src/protocol.ts
```diff
@@ -141,6 +141,44 @@ export interface IRuntimeMessageCollection {
...
+	readonly localOpMetadata: unknown;
+}
+
+/**
+ * A collection of messages to be resubmitted together as a "bunch".
+ * @remarks
+ * All messages in a resubmit collection share the same target — that is, they are
+ * for the same DDS — and share a single `squash` setting. This mirrors the inbound
+ * "bunch" shape of {@link IRuntimeMessageCollection} for the outbound resubmit path,
+ * allowing DDSes to handle a contiguous run of resubmits in one call.
```
anthony-murphy, line 166: Deep Review: `IRuntimeResubmitMessageCollection` remarks overstate the same-DDS guarantee at this layer.

The remarks here (and the matching `IFluidDataStoreChannel.reSubmitMessages` doc at `dataStoreContext.ts:484-497`) say all messages "are for the same DDS" / "Ask the DDS". But grouping happens at two distinct layers:

- `channelCollection.ts:847-873` groups by `{ address: e.envelope.address, type: e.envelope.contents.type }` (i.e. `(address, ddsType)`) before calling `context.reSubmitMessages(key.type, { squash, messages })`.
- `dataStoreRuntime.ts:1418-1454` then re-splits ChannelOp bunches by channel address before delegating to `channelContext.reSubmitMessages`.

So at the `IFluidDataStoreChannel` boundary (where this doc lives) the collection is grouped by `(address, ddsType)` *within one data store* — the per-DDS run is established downstream by `FluidDataStoreRuntime` before reaching `IDeltaHandler.reSubmitMessages`.

Recommend rewriting both doc sites to scope the guarantee to the correct layer. Prevents misuse by third-party `IFluidDataStoreChannel` implementers that read the contract literally.

packages/runtime/runtime-definitions/src/protocol.ts
```diff
@@ -141,6 +141,44 @@ export interface IRuntimeMessageCollection {
...
+	/**
+	 * If true, the DDS should avoid resubmitting any "unnecessary intermediate state" created
+	 * by these messages. Applies uniformly to every message in the collection.
+	 */
+	readonly squash: boolean;
+	/**
+	 * The messages to resubmit, in original submission order.
+	 */
+	readonly messages: readonly IRuntimeResubmitMessage[];
+}
```
anthony-murphy, line 180: Deep Review: Bunched-shape naming diverges from the inbound `IRuntimeMessageCollection` precedent.

Inbound shape (protocol.ts:129-142): `IRuntimeMessageCollection { envelope; messagesContent: readonly IRuntimeMessagesContent[]; local; }`.

Outbound shape here: `IRuntimeResubmitMessageCollection { squash; messages: readonly IRuntimeResubmitMessage[]; }` — items field is `messages` (not `messagesContent`), per-item type is `IRuntimeResubmitMessage` (not `…MessagesContent`). The doc comment claims this "mirrors the inbound `IRuntimeMessageCollection`" but the field/type names don't.

This matters because these are `@sealed @legacy @beta` types and renaming after consumers adopt is a breaking change. vladsud relitigated the inbound naming convention on #22839 and produced the current convention — worth their sign-off here before merge.

Two paths:
- Align outbound naming with the inbound precedent (`messagesContent` / `IRuntimeResubmitMessagesContent`), or
- Keep the divergence (defensible: outbound entries genuinely lack `clientSequenceNumber`/`envelope`/`local` and carry `localOpMetadata` + collection-level `squash`) and update the "mirrors `IRuntimeMessageCollection`" doc to call out the intentional semantic distinction.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4515,6 +4548,84 @@ describe("Runtime", () => {
...
+				assert(
+					channelCollectionStub.reSubmitContainerMessages.calledOnce,
+					"Expected one bunched reSubmitContainerMessages call",
+				);
+				const [, squash] = channelCollectionStub.reSubmitContainerMessages.firstCall.args;
+				assert.strictEqual(squash, true, "squash flag should propagate to the bunched call");
+			});
+
 			it("discardChanges drops staged ops", () => {
 				const channelCollectionStub = stubChannelCollection(containerRuntime);
```
anthony-murphy, line 4630: Deep Review: No direct test for `ChannelCollection.reSubmitContainerMessages` `(address, type)` sub-bunching.

The new test "commitChanges splits bunches when the data store address changes" asserts only that `channelCollection.reSubmitContainerMessages` is called once carrying all five entries in submission order (`["ds1","ds1","ds2","ds1","ds1"]`) — the stub short-circuits back to `containerRuntime.submitMessage`, so the `(address, type)` sub-bunching at `channelCollection.ts:847-873` is not actually exercised in the integration path.

It's covered indirectly via `forEachContiguousBunch` unit tests (`bunching.spec.ts:33-67, 90-109`), but nothing exercises the `ChannelCollection`-layer split, and nothing exercises the case where two same-address entries differ on the inner `FluidDataStoreMessage.type` (uncovered at any layer).

Recommend a direct unit test against `ChannelCollection.reSubmitContainerMessages` (or via a richer stub) asserting:
- (a) contiguous same-`(address, type)` runs collapse into one `context.reSubmitMessages` call,
- (b) address transitions split,
- (c) same-address entries with different inner `FluidDataStoreMessage.type` also split.

The `(address, type)` tuple choice is this layer's central new behavior; the type-change case especially deserves a guard.

packages/runtime/datastore/src/channelDeltaConnection.ts
```diff
@@ -123,6 +125,39 @@ export class ChannelDeltaConnection implements IDeltaConnection {
...
+						localOpMetadata: expandedMetadata,
+					});
+				},
+			);
+		}
+
+		const expandedCollection: IRuntimeResubmitMessageCollection = {
+			squash: collection.squash,
+			messages: flattened,
+		};
```
anthony-murphy, line 148: Deep Review: Stashed-op expansion makes the bunch size invariant implicit.

This loop runs `processWithStashedOpMetadataHandling` over every input entry and pushes into `flattened`, so a `collection.messages.length === N` call from above can fan out to `expandedCollection.messages.length !== N` before reaching `handler.reSubmitMessages`. The inline comment captures the symmetry intent ("mirroring how processMessages expands messagesContent"), but the `IDeltaHandler.reSubmitMessages` interface contract over in `datastore-definitions/src/channel.ts` is silent on whether observed `bunch.length` corresponds 1:1 to the entries the caller passed in.

abe27342 raised this same fan-out asymmetry on #19518 (channelDeltaConnection.ts line 111) for the inbound path; the bunched outbound interface should document it explicitly so SharedTree/MergeTree don't write overrides that assume 1:1.

Recommend adding a doc comment on `IDeltaHandler.reSubmitMessages` (and the bunched `IFluidDataStoreChannel.reSubmitMessages`) clarifying that stashed-op expansion may make the observed bunch length differ from the caller's input length, and that `bunch.length` is post-expansion, not 1:1 with caller pending entries. Three-line fix that closes the contract footgun before consumers start implementing the override.

PR #27434, opened 2026-05-28
fix(tree): Fix event buffer leak

Josmithr:
Updates internal `withBufferedEvents` to simplify its state management to not leverage event listeners, and to ensure that nodes are not leaked.

[AB#71782](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/71782)


packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -336,30 +336,53 @@ export function withBufferedTreeEvents(callback: () => void): void {
...
 /**
- * Event emitter to notify subscribers when tree events buffered due to {@link withBufferedTreeEvents} should be flushed.
+ * Set of {@link KernelEventBuffer}s that have accumulated buffered events during the current
+ * {@link withBufferedTreeEvents} window and therefore need to be flushed when it ends.
+ *
+ * @remarks
+ * Storing the buffers directly (rather than subscribing each one to a global "flush" event)
+ * avoids leaking buffers - and everything their listeners capture - in the case where a
+ * {@link TreeNodeKernel} is dropped without being explicitly disposed. The set is empty
+ * whenever no buffering window is in progress.
```
```suggestion
 * The set should be empty whenever no buffering window is in progress.
```

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -336,30 +336,53 @@ export function withBufferedTreeEvents(callback: () => void): void {
...
-}>();
+const activeBuffers: Set<KernelEventBuffer> = new Set();
+
+/**
+ * Test-only accessor for the current size of {@link activeBuffers}.
+ *
+ * @remarks
+ * Exposed so regression tests can verify that {@link KernelEventBuffer}s are not retained
+ * by module-level state outside an active {@link withBufferedTreeEvents} window.
+ * Do not use outside of tests.
```
```suggestion
```

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -336,30 +336,45 @@ export function withBufferedTreeEvents(callback: () => void): void {
...
+			// Snapshot-and-clear before flushing to safely handle reentrant `withBufferedTreeEvents`
+			// calls made by listeners that fire during `buffer.flush()` below:
+			// - Iterating an array means a reentrant call's `clear()` cannot truncate our loop
+			//   and cause buffers later in `activeBuffers` to be skipped (and their events dropped).
+			// - Clearing up front means the reentrant call starts from an empty set, so its own
+			//   finally block only flushes what it buffered - not a re-flush of our remaining buffers.
+			const toFlush = [...activeBuffers];
+			activeBuffers.clear();
+			for (const buffer of toFlush) {
+				buffer.flush();
```
CraigMacomber: Its not new in this PR, so not a blocker, but we really shouldn't be running anything that might touch the tree, especially user events, in a finally block. Doing so is a violation of our own policies around error handling, and thus can end up with UsageErrors implying that the user incorrectly caught a fatal error and continued to use the tree, and losing the original error (well, if you get the use of broken object error we include it, but other places might fail to do so. This pattern of running code which might throw when another exception was already thrown is what's broken with our deprecated transaction API for example.).

I think this might be another example of an API which isn't actually practical to implement due to issues with exceptions.

Clearing the statics to avoid the leak in the error cases shouldn't throw, and should be safe but moving the actual flushing out of the finally block might be the right choice so it only runs when there was no error.

Ideally we would put the tree into a broken state if a user tries to bubble an exception (which may or may not have been from tree originally) back into a tree call stack from a callback, but as this is in a static context, we can't actually implement that (it might even be for unhydrated nodes which I think we can't put into a broken state)

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
```diff
@@ -336,30 +336,45 @@ export function withBufferedTreeEvents(callback: () => void): void {
...
+ */
+const activeBuffers: Set<KernelEventBuffer> = new Set();
+
+/**
+ * Test-only accessor for the current size of {@link activeBuffers}.
  */
-const flushEventsEmitter = createEmitter<{
-	flush: () => void;
-}>();
+export function TEST_activeBufferCount(): number {
```
CraigMacomber, line 370: Is this a naming convention we use elsewhere? I haven't seen it in this codebase before but it seems useful. Is it documented anywhere?

packages/dds/tree/src/test/simple-tree/core/treeNodeKernel.spec.ts
```diff
@@ -188,6 +189,68 @@ describe("withBufferedTreeEvents", () => {
...
+
+			assert.equal(
+				TEST_activeBufferCount(),
+				before,
+				"Unbuffered edits must not register the buffer with module-level state",
+			);
+		});
+
+		it("nested withBufferedTreeEvents windows release the buffer when the outer window ends", () => {
+			const before = TEST_activeBufferCount();
```
CraigMacomber: Shouldn't these tests assert this is zero (ensuring we don't have leaks from previous tests, which would likely screw up these tests) rather than explicitly tolerating it?

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
Josmithr: Not a convention that I am aware of, and I am not aware of an existing convention. I needed something in the moment, so I started with this. If we like it, I'd be happy to document it.

packages/dds/tree/src/test/simple-tree/core/treeNodeKernel.spec.ts
Josmithr: Yeah, good call. Will update.

packages/dds/tree/src/simple-tree/core/treeNodeKernel.ts
Josmithr: Makes sense. I'll update.

CraigMacomber (approved):

PR #27437, opened 2026-05-28
test(test-version-utils) Removed convertRegistry from e2e test infra that implicitly converts DDS registries

agarwal-navin:
## Description

The `convertRegistry` function in the test infra that replaces DDS registries passed in by tests with a versioned one as per the compat version the test is running in. For example, if the test is running in layer compat version mode where `DataStore` version is 1.4.0, it replaces the registry with its 1.4.0 version.
This works in most cases but has a few shortcomings:
1. If a test creates a custom DDS registry (say a tree registry via `configuredSharedTree`), this is replaced with the base registry losing all the custom configurations.
2. This happens explicitly and the test authors may not know about this. This can lead to unintended consequences where the DDS is versioned but the other objects like container runtime, data store, loader, etc. may not be. The tests can fail with seemingly unrelated and random errors.

This change fixes some of the above problems by doing the following:
- Removed the `convertRegistry` function.
- Updated almost all tests to use the correct version of the APIs exposed by `describeCompat`. There are a few tests which still use the latest APIs - added comments / tagged task items for them.
- There is a lint rule that notify test authors if they don't use the compat APIs. It was not working for all cases. Updated it to work for all knowns cases.
- Added documentation on the correct pattern to use for various scenarios. Also, pointed the lint rule error message to this document.

[AB#74635](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/74635)

## Future work

The test infra still implcitily uses compat versions of other objects - driver, loader, runtime, data store, DDS, etc. when tests don't create these objects explicitly. But these don't happen for all patterns - for example, a test can create a loader of a particular version and use it to create different versions of container runtimes and other objects leading to an unexpected mismatch.
One approach to fix this is to have APIs that explicitly tell the test which API is being used - old / new, for create / load, etc. That combined with the lint rules should make it clearer and more intuitive to test authors when they use incorrect APIs or incorrect version combinations of objects.

### DIFF
--- a/packages/test/test-end-to-end-tests/WritingCompatCorrectTests.md
+++ b/packages/test/test-end-to-end-tests/WritingCompatCorrectTests.md
@@ -1,0 +1,367 @@
+# Writing Compat-Correct Tests
+
+This document explains how to write Fluid Framework end-to-end tests that correctly exercise compatibility across versions. Tests that bypass these patterns may silently pass under the current version and break when run against older or cross-client compat configurations.
+
+If you arrived here from an ESLint `@typescript-eslint/no-restricted-imports` error — typically of the form
+_"`'@fluidframework/<pkg>/internal'` import is restricted from being used by a pattern. Rather than import this Fluid package directly, use the 'apis' argument of describeCompat."_
+or _"`<Name>` import from `'@fluidframework/<pkg>/internal'` is restricted from being used by a pattern. Use `apis.<layer>.<Name>` from describeCompat instead."_ — this is the doc to read.
+
+## Why this matters
+
+`describeCompat` parameterizes a test suite over a **compatibility matrix**: the same test runs against multiple combinations of Loader / Container Runtime / Data Runtime / DDS package versions. The compat matrix is what catches version-skew bugs before they reach production.
+
+The matrix only works if the test code uses the **version-aware factories and classes** that `describeCompat` provides via its `apis` argument. A statically imported `SharedMap` (for example) is fixed to the current version. When the matrix tries to run the test against an older container-runtime + older Data Runtime combination, that test still uses the current-version `SharedMap` — silently testing only the current version under the guise of the older config.
+
+For background on the compat matrix itself, see [test-version-utils's README.md](../test-version-utils/README.md) and [Compatibility.md](../../../docs/content/docs/deep/compatibility.md).
+
+## Using `apis`
+
+`describeCompat`'s callback receives `apis` as its second argument. **Use it for every value reference to a compat-versioned type** — never import these statically. Destructure what you need at the top of the callback so the rest of the test body reads naturally:
+
+```ts
+describeCompat("My Test", "FullCompat", (getTestObjectProvider, apis) => {
+    const { SharedMap, SharedString } = apis.dds;
+    const { DataObject, DataObjectFactory } = apis.dataRuntime;
+    const { ContainerRuntimeFactoryWithDefaultDataStore } = apis.containerRuntime;
+    const { Loader } = apis.loader;
+    // ...
+});
+```
+
+The `apis` object is organized by layer:
+
+| Path | What's there |
+|---|---|
+| `apis.dds.*` | DDS factory classes: `SharedMap`, `SharedDirectory`, `SharedCell`, `SharedCounter`, `SharedMatrix`, `SharedString`, `SharedTree`, `ConsensusQueue`, `ConsensusRegisterCollection`, `SparseMatrix`, `SharedArray`, `SharedSignal` |
+| `apis.dataRuntime.*` | `DataObject`, `DataObjectFactory`, `FluidDataStoreRuntime`, `TestFluidObjectFactory` |
+| `apis.dataRuntime.packages.<pkg>` | Less common APIs by package: `cell`, `counter`, `datastore`, `map`, `matrix`, `orderedCollection`, `registerCollection`, `sequence`, `sequenceDeprecated`, `agentScheduler`, `tree` |
+| `apis.containerRuntime.*` | `ContainerRuntime`, `BaseContainerRuntimeFactory`, `ContainerRuntimeFactoryWithDefaultDataStore` |
+| `apis.loader.*` | `Loader` |
+
+`apis.<layer-name>.packages.<package-name>` (package names are unscoped and camelCased) holds less common exports such as helper functions and constants. Keep in mind that if these APIs change over time, tests depending on them will either need to have a reduced compat matrix or include back-compat logic. See `"Change contents of dds, then rehydrate and then check summary"` for an example of such a test.
+
+### ❌ Static import — DON'T
+
+```ts
+import { SharedString, createOverlappingIntervalsIndex } from "@fluidframework/sequence";
+
+const registry: ChannelFactoryRegistry = [["sharedString", SharedString.getFactory()]];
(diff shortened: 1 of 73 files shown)


packages/test/test-end-to-end-tests/WritingCompatCorrectTests.md
```diff
@@ -0,0 +1,367 @@
...
+
+describeCompat("My SharedTree compat test", "FullCompat", (getTestObjectProvider, apis) => {
+    beforeEach(function () {
+        // SharedTree was added in version 2.0.0 — skip cross-client configs whose
+        // create or load version predates that.
+        if (apis.mode === "CrossClientCompat") {
+            const version = apis.dataRuntime.version;
+            const versionForLoading = apis.dataRuntimeForLoading?.version;
+            assert(versionForLoading !== undefined, "versionForLoading must be defined in cross-client");
+            if (lt(version, "2.0.0") || lt(versionForLoading, "2.0.0")) {
```
Abe27342: when `version` or `versionForLoading` is prerelease (happens in CI builds), funky stuff can happen with the comparison here. I think the way you have it written ends up working, but if you do `gt("2.0.0", version)` instead, it's wrong. I'd include a warning here mentioning to take special care for this case, and probably also recommend using `lt(version, "2.0.0", { includePrerelease: true })` (or maybe re-exporting utils like that from test-version-utils? We already take a dep on semver there)

packages/test/test-end-to-end-tests/WritingCompatCorrectTests.md
```diff
@@ -0,0 +1,367 @@
...
+        }
+    });
+
+    // The test body can now safely assume apis.dds.SharedTree / apis.dataRuntime.packages.tree.* exist.
+});
+```
+
+This keeps the compat matrix honest: a config that genuinely couldn't run the test gets `skip()` instead of being silently rerouted to current-version code.
+
+### Option B: Fall back to the current-version import (when the test is *not* about compat of this API)
```
Abe27342: won't this pattern end up running duplicate tests? When do we actually want to use it?

Abe27342 (commented):
reviewed docs changes--coming back for code later

packages/test/test-end-to-end-tests/WritingCompatCorrectTests.md
agarwal-navin: Yes, and those tests would actually not be testing what the combination is supposed to test. We need a pattern which will correctly skip compat combinations where a versioned object does not exist.
Do you reckon a pattern like, if a symbol doesn't exist in CompatApis, skip the test would be better? It will also likely get rid of the pattern in your other comment where the test skips for particular versions.

agarwal-navin: Well, it won't be that simple actually. The API vesions for loading will only exist in cross-client compat mode, so we don't want to always skip tests when those don't exist. We only want to do them in cross-client compat mode. This makes it complicated.

agarwal-navin: I am wondering if we should make the API versions for loading always exist. When running in non cross-client compat mode, they are the same as the create APIs. That way, when an API doesn't exist, the test should simply be skipped. At least in theory, sounds like it should work.

packages/test/test-version-utils/src/compatUtils.ts
```diff
@@ -120,32 +116,8 @@ function createGetDataStoreFactoryFunction(
 		}
 	}
 
-	const registryMapping = {};
```
Abe27342, line 117: 🥳

packages/test/test-end-to-end-tests/src/test/gc/gcIncrementalSummaries.spec.ts
```diff
@@ -12,7 +12,10 @@ import {
 	itExpects,
 } from "@fluid-private/test-version-utils";
 import { IContainer } from "@fluidframework/container-definitions/internal";
-import { ContainerRuntime, ISummarizer } from "@fluidframework/container-runtime/internal";
+import {
```
Abe27342: nit: here and in a few other files, seems like you should be able to make the whole thing an `import type` statement?

`import type { ContainerRuntime, ISummarizer } from "@fluidframework/container-runtime/internal"`?

packages/test/test-end-to-end-tests/src/test/offline/stashedOps.spec.ts
```diff
@@ -135,8 +137,20 @@ const pendingBlobsFromPendingState = (pendingState: string): Record<string, any>
...
 // REVIEW: enable compat testing
 describeCompat("stashed ops", "NoCompat", (getTestObjectProvider, apis) => {
-	const { SharedMap, SharedDirectory, SharedCounter, SharedString, SharedCell } = apis.dds;
+	const {
+		SharedMap,
+		SharedDirectory,
+		SharedCounter,
+		SharedString,
+		SharedCell,
+		// SharedArray, SharedSignal, and SharedTree were added recently and may not exist on apis.dds
```
Abe27342: seems smelly that this pattern is used in a test marked `NoCompat` -- what's up with that?

Abe27342: are we still doing cross-version testing on tests marked NoCompat or something like that?

agarwal-navin: It doesn't need to do this for "NoCompat" and cross-compat won't run. However, if "FullCompat" were to be turned on, these tests will start failing for versions where these APIs don't exist. That's the reason to do this pattern here.

packages/test/test-end-to-end-tests/WritingCompatCorrectTests.md
agarwal-navin: https://github.com/microsoft/FluidFramework/pull/27463

packages/test/test-version-utils/src/compatUtils.ts
agarwal-navin: 🥳

packages/test/test-end-to-end-tests/src/test/gc/gcIncrementalSummaries.spec.ts
agarwal-navin: Good point. I wonder why lint doesn't catch this.

packages/test/test-end-to-end-tests/src/test/offline/stashedOps.spec.ts
Abe27342: ok, if it's just pre-emptive defensive coding, thats fine.

packages/test/test-end-to-end-tests/WritingCompatCorrectTests.md
agarwal-navin: Replaced this to check for tree package is undefined.

packages/test/test-end-to-end-tests/src/test/offline/stashedOps.spec.ts
agarwal-navin: I removed this from NoCompat tests. They can add it if and when they conver to FullCompat.

Abe27342 (approved):

PR #27442, opened 2026-05-29
Bundle size reduction: Shared Tree: Make TypeBox imports more granular

TommyBrosman:
Rewrites TypeBox imports to avoid `import { Type } from "@sinclair/typebox";`. This avoids pulling in TypeBox dependencies we don't actually use.

## Bundle size impact

Measured with `examples/utils/bundle-size-tests/scripts/collectAndCompareBundles.ts` (note: not yet merged, see #27151) against the merge-base of `main` (`e7fa1dbb87e`). All other entrypoints are byte-for-byte unchanged — only bundles that pull in `@fluidframework/tree` see savings.

Bundle size comparison vs `main` (merge-base `e7fa1dbb87e`):

| Bundle                     |     Base |  Current |     Diff |     % |
| -------------------------- | -------: | -------: | -------: | ----: |
| `sharedTree.js`            |  418,159 |  389,125 |  -29,034 | -6.9% |
| `fluidFramework.js`        |  428,772 |  399,738 |  -29,034 | -6.8% |
| `sharedTree.js` (gzip)     |  116,224 |  109,645 |   -6,579 | -5.7% |
| `fluidFramework.js` (gzip) |  118,823 |  112,224 |   -6,599 | -5.6% |

## Reviewer Guidance

- Feedback on the ergonomics of the new pattern.


Abe27342:
Can we redefine `Type` using the approach you've used once in a single file in SharedTree (where we import every feature of typebox we actually use), then reference that everywhere else?

TommyBrosman:
@Abe27342:
> Can we redefine `Type` using the approach you've used once in a single file in SharedTree (where we import every feature of typebox we actually use), then reference that everywhere else?

I like that idea, it seems like it could be much cleaner.

packages/dds/tree/src/codec/versioned/format.ts
```diff
@@ -3,7 +3,8 @@
  * Licensed under the MIT License.
  */
 
-import { type Static, Type } from "@sinclair/typebox";
+import { type Static, Number as _typebox_Number, Object as _typebox_Object, String as _typebox_String, Union as _typebox_Union } from "@sinclair/typebox";
```
Josmithr: We should be sure to add comments explaining this pattern to the relevant files. Otherwise, devs are likely to make unwanted changes without understanding the impact.

TommyBrosman: See util/typebox.ts for the doc comment in the new version. One downside: it will only show up for `Type`.

packages/dds/tree/src/codec/codec.ts
```diff
@@ -7,8 +7,8 @@ import type { ErasedType } from "@fluidframework/core-interfaces/internal";
 import { assert, fail } from "@fluidframework/core-utils/internal";
 import type { MinimumVersionForCollab } from "@fluidframework/runtime-definitions/internal";
 import { cleanedPackageVersion as runtimeUtilsCleanedPackageVersion } from "@fluidframework/runtime-utils/internal";
-import type { Static, TAnySchema, TSchema } from "@sinclair/typebox";
 
+import type { Static, TAnySchema, TSchema } from "../util/index.js";
```
CraigMacomber: I don't see any reason to change `type` imports, as they don't impact bundle size.

packages/dds/tree/src/util/typebox.ts
```diff
@@ -0,0 +1,72 @@
...
+ * tree-shaking because it is exported as a single namespace object that
+ * pulls in every TypeBox builder. By re-exporting only the kinds we use
+ * as a plain object, bundlers can drop the rest, significantly reducing
+ * bundle size for consumers of `@fluidframework/tree`.
+ *
+ * If you need a TypeBox kind not listed here, add it to this object
+ * (and the corresponding granular import above) rather than importing
+ * `Type` directly from `@sinclair/typebox`.
+ */
+export const Type = {
```
CraigMacomber: With no changes to our lint rules, this is likely to just regress again.

Can we just add a lint rule banning imports of `type` from `"@sinclair/typebox"` with a nice explanation of what to do instead (import the specific type in question directly).

That is better than this pattern since this breaks tree shaking as well (just at a smaller scale), meaning that if you only depended on some of the code in this package, and thus some of these, it will force pulling all of them in

packages/dds/tree/src/codec/codec.ts
TommyBrosman: I included the type imports for consistency. Importing types from the original package would be a smaller diff, sure. My rationale was that this way we never reference @sinclair/typebox in our source. That is easier to enforce than "type imports are fine, importing Type is not."

packages/dds/tree/src/util/typebox.ts
TommyBrosman: I added a broader import restriction lint rule that also excludes type imports for consistency.

packages/dds/tree/src/codec/versioned/format.ts
Josmithr: I think that comment plus the linter rule sufficiently cover my concerns here. And the new state looks much better IMO. Thanks!

packages/dds/tree/src/codec/codec.ts
TommyBrosman: This is no longer an issue now that I'm using the `import * as Type` pattern.

packages/dds/tree/src/util/index.ts
```diff
@@ -21,6 +21,14 @@ export {
 	unbrand,
 } from "./brand.js";
 export { brandedNumberType, brandedStringType } from "./typeboxBrand.js";
+export type {
```
CraigMacomber: I think this can be removed and they can be imported from typebox normally like they were before.

packages/dds/tree/eslint.config.mts
```diff
@@ -36,6 +36,24 @@ const config: Linter.Config[] = [
 			"@typescript-eslint/no-unsafe-assignment": "off",
 			"jsdoc/require-description": "warn",
 			"unicorn/no-null": "off",
+			// Enforce the granular TypeBox import pattern: import the `Type` namespace
+			// directly via `import * as Type from "@sinclair/typebox"` so the bundler can
+			// tree-shake unused TypeBox builders. Importing the `Type` value from the
+			// `util/index.js` barrel re-introduces a runtime object that defeats
+			// tree-shaking, so it is forbidden here.
```
CraigMacomber: This seems like you are banning the importing of the reexport that you have removed in an earlier version of this PR. We don't want to ban importing things that don't exist, we want to ban importing the problematic value from Typebox.

TommyBrosman: Leftover from old version. I have the correct rule up now.

packages/dds/tree/src/core/schema-stored/formatV2.ts
```diff
@@ -3,11 +3,14 @@
...
+import type { ObjectOptions, Static } from "@sinclair/typebox";
+import * as Type from "@sinclair/typebox";
 
 import { unionOptions } from "../../codec/index.js";
-import type { JsonCompatibleReadOnlyObject } from "../../util/index.js";
-import { JsonCompatibleReadOnlySchema } from "../../util/index.js";
+import {
+	type JsonCompatibleReadOnlyObject,
+	JsonCompatibleReadOnlySchema,
+} from "../../util/index.js";
```
CraigMacomber: This change seems unrelated and unnecessary (and likely undesired, though not important).

packages/dds/tree/eslint.config.mts
```diff
@@ -36,6 +36,33 @@ const config: Linter.Config[] = [
...
+			// than combining several, so this list must contain every selector for this
+			// package: the general selectors plus the tree-specific TypeBox selector.
+			"no-restricted-syntax": [
+				"error",
+				{
+					selector: "ExportAllDeclaration",
+					message:
+						"Exporting * is not permitted. You should export only named items you intend to export.",
+				},
+				"ForInStatement",
```
CraigMacomber, line 52: I think this is trying to say that the defaults from `recommended` above are getting overwritten not extended, so this code is duplicating them here. The comment here though fails to communicate that these settings are duplicated from the recommended values, and it also likely to result in them getting out of date if those inherited defaults are changed.

Anyway, its surprising to me they are not merged properly: have you confirmed that's actually the case?
If so, rather than duplicating configuration we should probably extend the recommended one manually with something like `[...recommended.findLast((rules) => "no-restricted-syntax" in rules.rules).rules["no-restricted-syntax"], youExtraRuleHere]`

TommyBrosman: Doc: https://eslint.org/docs/latest/use/configure/rules#use-configuration-files

Take a look at the example with after the text:

>When more than one configuration object specifies the same rule, the rule configuration is merged with the later object taking precedence over any previous objects.

Then they go on to show that their definition of "merge" is "last element wins." The part about extending recommended explicitly makes sense to me, I'll fix that next. If the documentation is accurate, I am a little surprised that we haven't accidentally overwritten rules from the recommended set already at some point.

packages/dds/tree/src/core/schema-stored/formatV2.ts
TommyBrosman: Reverted.

packages/dds/tree/eslint.config.mts
TommyBrosman: PR to allow more reuse of restricted syntax matchers: #27479

TommyBrosman: Updated with a todo for now. Per discussion, we want to lock in the gain now and worry about the base ESLint later.

packages/dds/tree/eslint.config.mts
```diff
@@ -36,6 +36,36 @@ const config: Linter.Config[] = [
 			"@typescript-eslint/no-unsafe-assignment": "off",
 			"jsdoc/require-description": "warn",
 			"unicorn/no-null": "off",
+			// TODO:AB#74822: Remove the no-restricted-syntax block once updated to eslint-config-fluid@13.0.0
```
CraigMacomber, line 39: if you are going to say this, might make sense to add this rule to eslint-config-fluid in this PR to ensure it gets into the next build after this is merge. THat said, as long as you get around to it, no need to block this PR if you prefer doing it independently.

CraigMacomber (approved):

PR #27449, opened 2026-06-01
Update root workspace to pnpm 11

CraigMacomber:
## Description

Update root workspace to pnpm 11.

This is mostly based on the instructions in https://pnpm.io/migration

Since this update was somewhat complex (a lot more changes than most pnpm updates, some in other PRs), a limited change with just the root workspace is being done first.

## Reviewer Guidance


tools/pipelines/templates/include-install-pnpm.yml
```diff
@@ -114,5 +114,7 @@ steps:
       echo "Pnpm user config location: $(pnpm config get userconfig)"
       pnpm config set store-dir ${{ parameters.pnpmStorePath }}
       echo "Pnpm store: ${{ parameters.pnpmStorePath }}"
-      # workspace-concurrency 0 means use the CPU core count. This is better than the default (4) for larger agents.
-      pnpm config set -g workspace-concurrency 0
```
CraigMacomber, line 118: Setting this errors with pnpm 11 since its not allowed as a global setting. We already set this in workspaces where it matters, so this is mostly redundant. If there is any workspace where we forgot it we could always add if desired. Since setting it in the workspaces directly also benefits devX for building locally, that setting is preferred anyway.

packages/tools/changelog-generator-wrapper/.npmrc
```diff
@@ -1,6 +0,0 @@
-engine-strict=true
```
CraigMacomber, line 1: I'm pretty sure this file was accidently preserved from when this was an independent package, and all these settings are already in the workspace which contains this.

tools/pipelines/templates/include-install-pnpm.yml
```diff
@@ -114,5 +114,7 @@ steps:
...
+      # CI uses mirrored registries that may not preserve trust metadata consistently.
+      # Disable pnpm trust downgrade checks in CI as a workaround.
+      echo "Disabling pnpm trust policy as workaround for npm mirror limitations"
+      echo "##vso[task.setvariable variable=PNPM_CONFIG_TRUST_POLICY]off"
```
alexvy86, line 120: Is there a behavior change in pnpm 11 that makes this necessary? Since installs in CI should happen with a lockfile I wouldn't expect the trust policy settings to flag any issues.

.gitignore
```diff
@@ -70,3 +70,8 @@ CLAUDE.local.md
 
 # Output from benchmarks
 **/benchmark*Output.json
+
+# npm configuration: Currently only used by CI to redirect to an alternative registry.
+# Once all workspaces are updated to pnpm ll and thus have moved the pnpm config out of these files,
+# this can be made global (remove the leading slash)
+/.npmrc
```
alexvy86, line 77: Do we really need it ignored? What CI does to use alternate registries isn't changing (and auth tokens end up in a different .npmrc file IIRC) so I don't quite see why we'd need this.

packages/test/test-version-utils/compat-workspaces/full/package.json
```diff
@@ -12,5 +12,10 @@
   "author": "Microsoft and contributors",
   "scripts": {
     "preinstall": "node ../../../../../scripts/only-pnpm.cjs"
+  },
+  "packageManager": "pnpm@11.5.0+sha512.dbfcc4f81cf48597afd4bc391ffdf12c11f1a9fb83a395bfa6b0a2d9cc2fd8ffebafdb1ccbd529632153f793904c2615b7f09fe1a345473fd1c35845172a8eb1",
+  "engines": {
+    "node": ">=22.22.2",
+    "pnpm": "11"
```
alexvy86, line 19: This package is part of the client release group so it doesn't need these explicitly, no?

pnpm-workspace.yaml
```diff
@@ -128,10 +148,41 @@ packages:
...
+    # markdown-magic is provided by another dependency so is ignored here.
+    - markdown-magic
+    # oclif includes some AWS-related features, but we don't use them, so we ignore @aws-sdk peer
+    # dependencies.
+    - '@aws-sdk/*'
+
+# Hoist these dependencies to the root workspace
+publicHoistPattern:
+  - '@arethetypeswrong/cli'
+  # ['*eslint*', '*prettier*'] is the default, so we add those as well
```
ChumpChief, line 176: Looks like this comment predates this change, probably remove the comment and optionally the hoist? https://github.com/pnpm/pnpm/pull/8621

ChumpChief (approved):
Approved with a nit, thank you for doing this!

tools/pipelines/templates/include-install-pnpm.yml
ChumpChief: You might consider adding it in all workspaces in this PR just to prove there's no risk of losing concurrency

.gitignore
ChumpChief: I thought about this too, but since the remaining purpose of .npmrc is just supposed to be registries/auth it seems like maybe a nice default to help devs accidentally avoid committing auth they don't intend to.

tools/pipelines/templates/include-install-pnpm.yml
CraigMacomber: We already have it in all the places that matter. Aside from test data inside of build-tools (which doesn't have any pipelines and is thus not impacted), the only workspaces with multiple packages we have are:

- (the root)
- build-tools
- server/gitrest
- server/historian
- server/routerlicious

These all already have this setting.
As this setting only impacts how pnpm schedules recursive work across multiple packages, no other cases matter.

Note that this matters in a lot fewer cases that you might expect.
This setting only matters for CI if the build machine has more than 4 cores, and more than 4 packages (as 4 is the default). This leaves it only actually mattering for the root and build-tools. Also I think it doesn't actually help build tools since last I checked it had issues with mutating its test data and couldn't run its tests in parallel, and its build can't ever build all 5 packages at once due to dependencies. Thus I think the only place this does anything is the client test pipeline job (the build one uses fluid-build, which does not care about this value, and I think the other client pipelines which might matter likely have CI images with 4 cores, are not CI long poles, and likely don't use pnpm to parallelize).

tools/pipelines/templates/include-install-pnpm.yml
CraigMacomber: Yes. Our old version of pnpm did not do any trust policy checking unless modifying the lock file. The new version of pnpm actually validates that everything in the lock file meets trust policy, which causes these issues to start impacting our CI.

.gitignore
CraigMacomber: @jason-ha suggested I do this in https://github.com/microsoft/FluidFramework/pull/27307#discussion_r3249779603 and included some motivation for it there which seemed reasonable.

This is not required, but it apparently makes the git status when CI is running cleaner and will allow simplifying some things once applied to all workspaces.

packages/test/test-version-utils/compat-workspaces/full/package.json
CraigMacomber: Not really. This is the root of a workspace with its own lock-file. It gets installed in a install triggered script by client, and is needed to make end-to-end package installs auditable with a lock file I believe.

CraigMacomber: In my opinion, anything with a lock file should probably pin the version of the tool which made that lock file so it is locked as well, which was missing before.

.gitignore
jason-ha: Yes. This is a nice simplification. Currently we have checks that make sure there aren't any files changed except this one. Once we don't have these checked in then we don't need special git status checks.

jason-ha (approved):

pnpm-workspace.yaml
CraigMacomber: Goot point. As this change is just preserving the existing settings, I'll leave it in this PR and see if I can remove them all in a followup.

alexvy86 (approved):

tools/pipelines/templates/include-install-pnpm.yml
ChumpChief: Ah I meant in all pnpm-workspace.yaml files, but point taken about it likely having no effect.  I do think it would be good if our pnpm-workpace.yaml files all look as-identical-as-makes-sense across the repo after this update is done (including workspace-concurrency 0 even if likely meaningless) just for consistency.

Issue #26910, opened 2026-03-31
Breaking Change: `@fluidframework/telemetry-utils` — Type-erase `ITelemetryLoggerExt`


jason-ha:
## Deprecated API

The external-facing `ITelemetryLoggerExt` interface's methods and its associated telemetry event types are deprecated. The `ITelemetryLoggerExt` interface will continue to exist such that it can be passed across FF API surface as needed.

The following types and methods from `@fluidframework/telemetry-utils` are deprecated:

| API | Kind | Package Export |
|-----|------|----------------|
| `TelemetryEventCategory` | Type alias | `@legacy` `@beta` |
| `ITelemetryGenericEventExt` | Interface | `@legacy` `@beta` |
| `ITelemetryErrorEventExt` | Interface | `@legacy` `@beta` |
| `ITelemetryPerformanceEventExt` | Interface | `@legacy` `@beta` |
| `ITelemetryLoggerExt.sendTelemetryEvent()` | Method | `@legacy` `@beta` |
| `ITelemetryLoggerExt.sendErrorEvent()` | Method | `@legacy` `@beta` |
| `ITelemetryLoggerExt.sendPerformanceEvent()` | Method | `@legacy` `@beta` |

**Context**

`ITelemetryLoggerExt` was an internal interface that leaked into the public API surface. While method access continues the internal implementation cannot safely evolve.

**Approach**

The type will be "erased" so that internals can evolve. The erased type will continue to inherit from `ITelemetryBaseLogger` and its `send` method will remain usable.

**Dependencies**

None.

**Compatibility Concerns**

There are no runtime changes at this time.
If, externally, `ITelemetryLoggerExt` interface has been implemented to pass to FF, instead implement `ITelemetryBaseLogger` from `@fluidframework/core-interfaces` (simple `send` method) and call `@fluidframework/telemetry-utils`'s `createChildLogger({ logger: your_logger_impl })` to safely create the expected `ITelemetryLoggerExt`.

**Expected Timeline**

Removal of methods and related interfaces are expected in v2.110.

PR #27462, opened 2026-06-01
feat(telemetry-utils): escalate selected performance events to essential

MarioJGMsoft:
## Description

Adds opt-in `PerformanceEvent` markers for promoting selected performance telemetry to essential:

- `endEventEssentialDurationThresholdMs` logs `_end` events as essential when their duration exceeds the configured threshold.
- `logCancelAsEssential` logs `_cancel` events as essential while preserving the configured cancel category.

This lets callers keep normal performance events at their configured verbosity while ensuring long-running operations and configured cancel events are not filtered out. Tests cover threshold behavior, cancel escalation, unchanged default log-level behavior, and existing error-category essential behavior.

## Reviewer Guidance

Please focus on whether the marker names and escalation semantics are clear, especially using the existing undefined-log-level path to preserve essential-event behavior.

Fixes: [AB#74788](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/74788)


packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -895,7 +903,27 @@ export class PerformanceEvent {
...
 		}
 
-		this.logger.sendPerformanceEvent(event, error, this.logLevel);
+		this.logger.sendPerformanceEvent(event, error, this.getLogLevel(eventNameSuffix, event));
+	}
+
+	private getLogLevel(
+		eventNameSuffix: string,
+		event: ITelemetryPerformanceEventExt,
+	): typeof LogLevel.verbose | typeof LogLevel.info | undefined {
```
markfields, line 915: Why are we using undefined to signal essential? Seems like unnecessarily tight coupling with the default specified elsewhere

packages/utils/telemetry-utils/src/test/performanceEvent.spec.ts
```diff
@@ -34,6 +38,10 @@ describe("PerformanceEvent", () => {
 	const callback = (): void => {
 		callbackCalls++;
 	};
+	const delay = async (): Promise<void> =>
```
markfields: See if you can quickly switch to using Sinon fake timers, which lets you test the delay without taking any real time. Hopefully you can one-shot it with Copilot.

These little ms setTimeouts can quickly add up and unnecessarily bloat unit test run times.

markfields: Admittedly 1ms is probably not make-or-break but still good practice to avoid real delays. More reliable way to test too

packages/utils/telemetry-utils/src/test/performanceEvent.spec.ts
```diff
@@ -95,6 +103,100 @@ describe("PerformanceEvent", () => {
...
+			assert.equal(logger.events[0]?.eventName, "AboveThreshold_end");
+		});
+
+		it("Logs cancel events as essential when configured", () => {
+			const perfEvent = PerformanceEvent.start(
+				logger,
+				{ eventName: "EssentialCancel" },
+				{
+					cancel: "generic",
+					logCancelAsEssential: true,
```
markfields: Can we make this the default? Feels likely we'll always want this

markfields (commented):
Thanks for knocking this out! Left some small suggestions

packages/utils/telemetry-utils/src/test/performanceEvent.spec.ts
MarioJGMsoft: Applied change

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: because the sendPerformanceEvent can only receive LogLevel.info, LogLevel.verbose or undefined. So we need it to send undefined so that it defaults to LogLevel.essential

markfields: Oh ok. Might be clearer to update that to take essential at some point

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -895,7 +899,27 @@ export class PerformanceEvent {
...
+		if (eventNameSuffix === "cancel") {
+			return undefined;
```
markfields, line 918: Can you add a comment at these `return undefined` lines explaining that this is translated to `essential` elsewhere?

packages/utils/telemetry-utils/src/test/performanceEvent.spec.ts
MarioJGMsoft: I assumed that when you made this the default, I'm guessing you were referring to the `logCancelAsEssential` code. That change was made

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: I added the comment

packages/utils/telemetry-utils/src/logger.ts
MarioJGMsoft: So the reason why I made it that way is because `sendPerformanceEvent` can only take `LogLevel.verbose`, `LogLevel.info` or undefined. With the breaking change happening in #26910, that should be fixable and then it can take esential.

markfields (approved):

markfields (approved):

PR #27466, opened 2026-06-02
refactor(tree): clarify move logic

yann-achard-MS:
## Description

Clarifies the logic of `DefaultEditBuilder.move`  and adds more test coverage.

## Breaking Changes

None


packages/dds/tree/src/feature-libraries/default-schema/defaultEditBuilder.ts
```diff
@@ -331,44 +332,52 @@ export class DefaultEditBuilder implements ChangeFamilyEditor, IDefaultEditBuild
...
+			const fieldOnAttachPathUnderLCA =
+				attachPath[sharedDepth]?.parentField ?? destinationField.field;
 			let adjustedAttachField = destinationField;
-			// After the above loop, `sharedDepth` is the number of elements, starting from the root,
-			// that both paths have in common.
-			if (sharedDepth === detachPath.length) {
-				const attachField = attachPath[sharedDepth]?.parentField ?? destinationField.field;
-				if (attachField === sourceField.field) {
-					// The detach occurs in an ancestor field of the field where the attach occurs.
-					let attachAncestorIndex = attachPath[sharedDepth]?.parentIndex ?? sourceIndex;
```
yann-achard-MS, line 344: This line should have been:
```javascript
let attachAncestorIndex = attachPath[sharedDepth]?.parentIndex ?? destIndex;
```
instead of
```javascript
let attachAncestorIndex = attachPath[sharedDepth]?.parentIndex ?? sourceIndex;
```
Thankfully the right side of the `??` was never needed because `attachPath[sharedDepth]` was guaranteed to be defined in that code path (which is why this change is just a refactor instead of a fix).

packages/dds/tree/src/test/shared-tree/editing.spec.ts
```diff
@@ -1627,13 +1627,6 @@ describe("Editing", () => {
 			expectJsonTree(tree, expectedState);
 		});
 
-		it("can move a node out from a field and into a field under a sibling", () => {
```
yann-achard-MS, line 1630: This scenario is now covered in src/test/feature-libraries/default-field-kinds/defaultChangeFamily.spec.ts.

packages/dds/tree/src/feature-libraries/default-schema/defaultEditBuilder.ts
```diff
@@ -331,44 +332,52 @@ export class DefaultEditBuilder implements ChangeFamilyEditor, IDefaultEditBuild
...
 			const attachPath = topDownPath(destinationField.parent);
+			/**
+			 * The number of elements, starting from the root, that both paths have in common.
+			 * This defines the lowest common ancestor node (LCA) of the source and destination fields.
+			 */
 			const sharedDepth = getSharedPrefixLength(detachPath, attachPath);
+			const fieldOnDetachPathUnderLCA =
+				detachPath[sharedDepth]?.parentField ?? sourceField.field;
+			const fieldOnAttachPathUnderLCA =
+				attachPath[sharedDepth]?.parentField ?? destinationField.field;
```
jason-ha: These are never used unless `sharedDepth === detachPath.length`.
Recommend populating these only under that condition - split the
```ts
			if (
				sharedDepth === detachPath.length &&
				fieldOnDetachPathUnderLCA === fieldOnAttachPathUnderLCA
			) {
```
condition with a nested block.

Inside that block we also see that `sharedDepth` is out of bounds for `detachPath` (`sharedDepth` is not less than `detachPath.length`); so, `fieldOnDetachPathUnderLCA` assignment simplifies to `sourceField.field`.

packages/dds/tree/src/feature-libraries/default-schema/defaultEditBuilder.ts
```diff
@@ -331,44 +332,52 @@ export class DefaultEditBuilder implements ChangeFamilyEditor, IDefaultEditBuild
...
-						const parentField = attachPath[sharedDepth] ?? oob();
+			// If the attach path runs through the field where the detach occurs...
+			if (
+				sharedDepth === detachPath.length &&
+				fieldOnDetachPathUnderLCA === fieldOnAttachPathUnderLCA
+			) {
+				const firstDifferentAttachAncestor = attachPath[sharedDepth];
+				// We know that the attach path runs deeper because both paths would otherwise bottom-out in the same field,
+				// which is handled in a separate code path above.
+				assert(firstDifferentAttachAncestor !== undefined, "Unexpected identical paths");
```
jason-ha: Do you like that assert or should `oob()` be used like before? (What are the package guidelines?)
This is what I had in this space when I was playing locally with Alex's original changes substituting some of current naming (but I didn't use `fieldOnDetachPathUnderLCA` for `sourceField.field`)
```ts
				const firstDifferentAttachAncestor = attachPath[sharedDepth] ?? oob();
				const fieldOnAttachPathUnderLCA = firstDifferentAttachAncestor.parentField;
				if (fieldOnAttachPathUnderLCA === sourceField.field) {
					// The detach occurs in an ancestor field of the field where the attach occurs.
					const attachAncestorIndex = firstDifferentAttachAncestor.parentIndex;
```

packages/dds/tree/src/test/feature-libraries/default-field-kinds/defaultChangeFamily.spec.ts
```diff
@@ -824,6 +824,107 @@ describe("DefaultEditBuilder", () => {
...
+							},
+						},
+					],
+				},
+			});
+			builder.move(
+				{ parent: root, field: fooKey },
+				1,
+				3,
+				{
```
jason-ha, line 901:
```suggestion
				// Path to root.foo.
				{ parent: root, field: fooKey },
				1,
				3,
				// Path to root.foo[4].foo, which is after the moved nodes in their original location.
				{
```

packages/dds/tree/src/test/feature-libraries/default-field-kinds/defaultChangeFamily.spec.ts
```diff
@@ -824,6 +824,107 @@ describe("DefaultEditBuilder", () => {
...
+						{ type: brand(numberSchema.identifier), value: 2 },
+						{ type: brand(numberSchema.identifier), value: 3 },
+					],
+				},
+			});
+			builder.move(
+				{ parent: root, field: fooKey },
+				1,
+				3,
+				{ parent: root_foo0, field: fooKey },
```
jason-ha, line 957:
```suggestion
				// Path to root.foo.
				{ parent: root, field: fooKey },
				1,
				3,
				// Path to root.foo[0].foo, which is before the moved nodes in their original location.
				{ parent: root_foo0, field: fooKey },
```

jason-ha (approved):
Cool - additional flow/clarity suggestions

packages/dds/tree/src/feature-libraries/default-schema/defaultEditBuilder.ts
yann-achard-MS: `oob()` is fine. I updated the code to use it.
A note on the logic you had when you were playing locally with Alex's original changes: I think the `oob()` would have been hit when moving elements from on field of a given parent to another.

packages/dds/tree/src/feature-libraries/default-schema/defaultEditBuilder.ts
yann-achard-MS: Nice, thank you.

packages/dds/tree/src/feature-libraries/default-schema/defaultEditBuilder.ts
jason-ha: Okay and yes, it would have. (I honestly don't remember the health of the code I was playing with.)
What I can see from the code now is that 
```ts
				const fieldOnDetachPathUnderLCA = sourceField.field;
				const fieldOnAttachPathUnderLCA =
					attachPath[sharedDepth]?.parentField ?? destinationField.field;
				if (fieldOnDetachPathUnderLCA === fieldOnAttachPathUnderLCA) {
					// We know that the attach path runs deeper because both paths would otherwise bottom-out in the same field,
					// which is handled in a separate code path above.
					const firstDifferentAttachAncestor = attachPath[sharedDepth] ?? oob();
					if (firstDifferentAttachAncestor.parentIndex < sourceIndex) {
```
has some nonsense.
1. `fieldOnAttachPathUnderLCA` is never used past the conditional.
2. When `attachPath[sharedDepth]?.parentField` is out of bounds, the comparison is against `destinationField.field` and will never succeed as we know that `destinationField.field` != `sourceField.field` (aka `fieldOnDetachPathUnderLCA`).
3. And `fieldOnDetachPathUnderLCA` is not `undefined`.

which means that `fieldOnAttachPathUnderLCA` could just be `attachPath[sharedDepth]?.parentField` and be `undefined` to get the same !== result.
If we capture `attachPath[sharedDepth]` into `firstDifferentAttachAncestor` first and compare that directly, then TypeScript will know that inside the condition `firstDifferentAttachAncestor` is defined. The code simplifies to:
```ts
				const firstDifferentAttachAncestor = attachPath[sharedDepth];
				const fieldOnDetachPathUnderLCA = sourceField.field;
				if (fieldOnDetachPathUnderLCA === firstDifferentAttachAncestor?.parentField) {
					// We know that the attach path runs deeper because both paths would otherwise bottom-out in the same field,
					// which is handled in a separate code path above.
					if (firstDifferentAttachAncestor.parentIndex < sourceIndex) {
```
(with option to skip `const fieldOnDetachPathUnderLCA = sourceField.field;`; unfortunately if we capture `fieldOnAttachPathUnderLCA = firstDifferentAttachAncestor?.parentField` to use in condition, TypeScript 5.4.5 won't track that `firstDifferentAttachAncestor` must be defined inside the condition).

packages/dds/tree/src/feature-libraries/default-schema/defaultEditBuilder.ts
```diff
@@ -331,31 +332,35 @@ export class DefaultEditBuilder implements ChangeFamilyEditor, IDefaultEditBuild
...
-					let attachAncestorIndex = attachPath[sharedDepth]?.parentIndex ?? sourceIndex;
-					if (attachAncestorIndex < sourceIndex) {
+				const fieldOnDetachPathUnderLCA = sourceField.field;
+				const fieldOnAttachPathUnderLCA =
+					attachPath[sharedDepth]?.parentField ?? destinationField.field;
+				if (fieldOnDetachPathUnderLCA === fieldOnAttachPathUnderLCA) {
+					// We know that the attach path runs deeper because both paths would otherwise bottom-out in the same field,
+					// which is handled in a separate code path above.
+					const firstDifferentAttachAncestor = attachPath[sharedDepth] ?? oob();
+					if (firstDifferentAttachAncestor.parentIndex < sourceIndex) {
```
jason-ha, line 365: I commented elsewhere about how to simplify/streamline the logic. Here is my overall suggestion including additional comments that I think would help any future readers.
```suggestion
			// Check that the detach parent (location of detach) is along the
			// attach path, and thus might need to be adjusted.
			// (This is synonymous with checking that the attach parent is below
			// or the same as the detach parent.)
			// If the attach parent is above the detach parent, then the move
			// does not need to be adjusted, as the attach path will not be
			// affected by the detach.
			if (sharedDepth === detachPath.length) {
				const fieldOnDetachPathUnderLCA = sourceField.field;
				const firstDifferentAttachAncestor = attachPath[sharedDepth];
				// Check that the detach location uses the same field as in
				// the attach path.
				// Note that when `firstDifferentAttachAncestor` is undefined,
				// that means the parents are the same. Since earlier call
				// to `compareFieldUpPaths` returned false, we know that the
				// fields must be different AND thus no adjustment is needed.
				if (fieldOnDetachPathUnderLCA === firstDifferentAttachAncestor?.parentField) {
					// Now working at the level where detach happens along the
					// attach path.
					if (firstDifferentAttachAncestor.parentIndex < sourceIndex) {
```

packages/dds/tree/src/feature-libraries/default-schema/defaultEditBuilder.ts
```diff
@@ -331,31 +332,35 @@ export class DefaultEditBuilder implements ChangeFamilyEditor, IDefaultEditBuild
...
-						const parentField = attachPath[sharedDepth] ?? oob();
+						let parent = firstDifferentAttachAncestor.parent;
 						parent = {
 							parent,
-							parentIndex: attachAncestorIndex,
-							parentField: parentField.parentField,
+							parentIndex: firstDifferentAttachAncestor.parentIndex - count,
+							parentField: firstDifferentAttachAncestor.parentField,
 						};
 						for (let i = sharedDepth + 1; i < attachPath.length; i += 1) {
```
jason-ha, line 382: Some clarifying comments:
```suggestion
						// Start with path to root that remains intact.
						let parent = firstDifferentAttachAncestor.parent;
						// Extend with index-adjusted parent.
						parent = {
							parent,
							parentIndex: firstDifferentAttachAncestor.parentIndex - count,
							parentField: firstDifferentAttachAncestor.parentField,
						};
						// Extend with the rest of the attach path, which is unaffected
						// apart from parentage replacements.
						for (let i = sharedDepth + 1; i < attachPath.length; i += 1) {
```

packages/dds/tree/src/feature-libraries/default-schema/defaultEditBuilder.ts
yann-achard-MS: I applied it then made some tweaks (in a separate commit to make reviewing them easier):
1. Grammar nit: I reserve "check that" for validation and use "check if" otherwise
2. I spelled out what "detach parent" means a little differently because the location of the detach or attach would be the cell(s) within the field.
3. I didn't like "If the attach parent is above the detach parent" because that's not the only case where the detach parent is not above the attach parent: it could be some random relative like a cousin.

packages/dds/tree/src/feature-libraries/default-schema/defaultEditBuilder.ts
yann-achard-MS: Applied

packages/dds/tree/src/test/feature-libraries/default-field-kinds/defaultChangeFamily.spec.ts
```diff
@@ -732,6 +732,56 @@ describe("DefaultEditBuilder", () => {
...
+							},
+						},
+					],
+				},
+			});
+			builder.move(
+				{ parent: root, field: fooKey },
+				1,
+				3,
+				{ parent: root_bar0, field: barKey },
```
jason-ha, line 759: nit: would be nice to name the path: `root.bar[0].bar`, yes?

jason-ha (approved):

packages/dds/tree/src/test/feature-libraries/default-field-kinds/defaultChangeFamily.spec.ts
yann-achard-MS: Nah... None of the other `FieldUpPath` are defined out of line. Only the nodes' `UpPath`.

PR #27467, opened 2026-06-02
perf(tree): avoid per node NodePositionInfo allocation in UniformChunks

brrichards:
## Motivation

A `UniformChunk` traverses its flat `values` array like a tree by pairing it with a `ChunkShape`. Previously, each `ChunkShape` stored its own fully-expanded `positions` array `clonePositions` was called with the chunk's `topLevelLength`, so it materialized one `NodePositionInfo` per node, per top-level instance, an array of size `1 + topLevelLength × nodesPerTree`. A chunk of 100 points (3 nodes each) allocated ~301 `NodePositionInfo` objects, and a new array was built for every distinct `topLevelLength`.

This was pure duplication: every top-level instance held its own copy of the same structural template, differing only by index and value offset. Each entry was a complete `UpPath` (real `parent` pointers and per-instance `parentIndex`), which made the cursor cheap. `getPath()` just returned the stored entry and `fieldIndex` read its `.parentIndex` directly but the memory cost scaled with chunk size.

## The offset design

The nodePositionInfo is identical across top-level instances, so it only needs to be stored once. `positions` now lives on `TreeShape` (not `ChunkShape`) and holds a single instance's worth of entries that is shared by identity across every chunk of that shape and every top-level index. `ChunkShape` shrinks to a tiny `(treeShape, topLevelLength)` value object that no longer grows in size as the chunk grows.

The cursor recovers each concrete node by combining the shared template with the node's top-level
index, using a flat `positionIndex`:

- Flat layout: position `0` is the root-field marker; a node lives at:
  `1 + topLevelIndex × nodeLength + withinTree`.
- `moveToPosition` decomposes it back:
  `withinTree = (positionIndex - 1) % nodeLength`,
  `topLevelIndex = (positionIndex - 1 - withinTree) / nodeLength`,
  and selects the shared entry `treeShape.positions[withinTree]`.
- A node's value is `entry.valueOffset + topLevelIndex × valuesPerTopLevelNode`.

A fast path handles leaf shapes (`nodeLength === 1`): `topLevelIndex = positionIndex - 1` directly, with no modulo/division.

## Performance tradeoff vs. memory payoff

Cost: navigation now does a little arithmetic (a modulo/division in `moveToPosition`, plus additions for value/position offsets) where it previously did a single array index into a precomputed per-node array. So each cursor step is marginally more compute.

Payoff: position-info memory drops from `O(topLevelLength × nodesPerTree)` per chunk to`O(nodesPerTree)` per shape, shared. Large chunks and many chunks of the same shape no longer allocate redundant `NodePositionInfo` objects at all, and `ChunkShape` becomes a trivially small object. The leaf fast path keeps the most common case free of the extra division.

## UpPath is now materialized on demand

Because a shared template entry is only the `topLevelIndex = 0` base, it is no longer a valid standalone path for a given node. `getPath()` / `getFieldPath()` therefore build a fresh `UpPath` on demand via `materializePath(positionIndex)`, which walks the shared template and re-applies the top-level index at each level. This is an `O(depth)` allocation that happens only when a path is actually requested, rather than being stored for every node up front. 

## New tests

In `uniformChunk.spec.ts`:

- `validate shape for <name>` — walks each shape's shared `positions` array and checks every entry's parent linkage plus per-field `parentIndex`/`parentField`, confirming the single-instance template is built correctly (the original structural check, now reading from `TreeShape`).
- `shares one NodePositionInfo prototype (size 3) and derives values by offset` — verifies a point shape's `positions` has exactly 3 entries, is pinned by identity to the `TreeShape` across chunk lengths, and that the cursor reads each point's `x`/`y` from the correct flat slot.
- `shares one NodePositionInfo prototype (size 1) and derives leaf values by offset` — the same for a leaf shape, exercising the `nodeLength === 1` fast path.
- `does not store a per-node UpPath: paths only exist once materialized` — confirms the chunk keeps only the base (`topLevelIndex 0`) entries and that `getPath()` returns a fresh standalone path each call carrying the real top-level index, without mutating the shared template.
- `getPath/getFieldPath, with and without a PathRootPrefix (chunk nested in a forest)` — checks materialized paths at the root, a top-level node, a nested field, and a nested leaf, and that a prefix rewrites only the uppermost ancestor (`indexOffset`/`rootFieldOverride`/`parent`) while deeper nodes are untouched.
- `fork() at a nested position is independent and reproduces path/value` — forks the cursor at a deep node and verifies the fork reproduces the same value/path and stays correct after the original cursor moves away, proving its derived state is rebuilt from the copied `positionIndex`.

In `chunkTree.spec.ts`:

- `chunks 10 points into UniformChunks of topLevelLength 4 and reads values via the cursor` — an end-to-end test that runs 10 points through the real chunker with a policy capping nodes-per-chunk so it splits into UniformChunks of `[4, 4, 2]`, then reads every value back through the cursor's offset lookup.


packages/dds/tree/src/test/feature-libraries/chunked-forest/uniformChunk.spec.ts
```diff
@@ -34,22 +43,22 @@ import { cursorToJsonObject, singleJsonCursor } from "../../json/index.js";
...
 
 // Validate a few aspects of shapes that are easier to verify here than via checking the cursor.
+// After the offset rework the position info lives on the shared per-tree `TreeShape.positions`
+// rather than on a per-chunk array, so read it from there.
 function validateShape(shape: ChunkShape): void {
-	for (const [positionIndex, info] of shape.positions.entries()) {
-		if (info === undefined) {
-			continue;
-		}
+	const positions = shape.treeShape.positions;
```
brrichards, line 49: This test now basically checks that the chunk points to the correct treeShape.positions array (which they all share). So all these ChunkShapes should have the same positions array. I'm wondering if we want to re-write this test to just check that each ChunkShape points to the correct TreeShape?

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -113,7 +113,11 @@ export class TreeShape {
 	public readonly fieldsOffsetArray: readonly OffsetShape[];
 	public readonly valuesPerTopLevelNode: number;
 
-	// TODO: this is only needed at chunk roots. Optimize it base on that.
+	/**
+	 * Position info for the nodes of a single top-level tree of this shape, shared by every chunk
+	 * that uses this shape. The cursor derives each node's actual position info from this shared
+	 * array plus the node's top-level index within the chunk.
```
CraigMacomber: A bit of a nuanced point here, which isn't too important in this case, but a nice example of the idea: its nice to separate remakes about how something can be used (or is used) vs what its invariants are or what it is.  That helps make it clear what it would mean to change something (is a change non breaking for example), or which part is of the code is bugged when things don't work. Also, it makes the codebase easier to learn, as it removes the cycles from the graph of things you have to read to understand what's going on (basically avoiding the iterative challenges with the [Hermeneutic_circle](https://en.wikipedia.org/wiki/Hermeneutic_circle ) )

As this has only one user, and is mostly just to do whatever that user needs, I don't think it matters much here, but we can still try and preserve TreeShape as a logical thing which chunks build on rather than just one interconnected part of the chunk code.

I think this proposed comment is accurate, but you should verify it.

```suggestion
	 * Information about every node in this shape.
	 * The root is first, and all indexes it uses to refer to other nodes are indexes into this array.
	 * Beyond that the ordering is an implementation detail of this shape.
	 * @remarks
	 * Use of this in contexts where there might be multiple top-level nodes requires some additional care.
	 * For example {@link Cursor} derives each node's actual position info from this shared
	 * array plus the node's top-level index within the chunk.
```

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -322,16 +322,25 @@ class NodePositionInfo implements UpPath {
...
  *
- * Works by tracking its location in the chunk's `positions` array.
+ * Tracks a flat `positionIndex` and derives each node's position info from the shape's shared
+ * {@link TreeShape.positions} plus the node's top-level index.
  */
 class Cursor extends SynchronousCursor implements ChunkedCursor {
 	private positionIndex!: number; // When in fields mode, this points to the parent node.
-	// Undefined when in root field
+
+	// Shared position info for the current node: the same entry from the shape's per-tree
```
CraigMacomber: should use TS-doc style comments for improved intelisense support.

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
brrichards: Yeah thats accurate. Thanks for the example! I'vve updated the docs to mirror the `@ remarks` feedback

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -295,17 +297,18 @@ class OffsetShape {
...
+	 * @param parentField - The {@link FieldKey} of the field this node occupies within its parent.
 	 * @param parentIndex - indexWithinParentField
-	 * @param indexOfParentField - which field of the parent `parentIndex` is indexing into to locate this.
-	 * @param indexOfParentPosition - Index of parent NodePositionInfo in positions array. TODO: use offsets to avoid copying at top level?
+	 * @param indexOfParentField - Which field of the parent `parentIndex` is indexing into to locate this.
+	 * @param indexOfParentPosition - Index of this node's parent in {@link TreeShape.positions}
 	 * @param shape - Shape of the top level sequence this node is part of
-	 * @param valueOffset - TODO
+	 * @param topLevelLength - Number of siblings in this node's field. For a root this is unused
+	 * @param valueOffset - Offset of this node's value within one top-level tree's slice of the chunk's flat `values` array
```
CraigMacomber: Not all nodes have values. Would be good to clarify what happens in the case where there is no value (I think this is just the offset to where we would have put the value if there was one, and should not be used if the shape indicate there is no value, but I'm not confident in that)

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -322,16 +325,24 @@ class NodePositionInfo implements UpPath {
...
-	private readonly shape: ChunkShape;
-	private readonly positions: readonly (NodePositionInfo | undefined)[];
+	/** Which top-level node of the chunk the current position is within. Valid when nodePositionInfo !== undefined. */
+	private topLevelIndex: number = 0;
+
+	// Cached constants for faster access.
+	private readonly shape: ChunkShape; /** */
+	private readonly treeShape: TreeShape; /** The chunk's per-tree shape (shape of each top-level tree). */
+	private readonly nodeLength: number; /** Number of positions in one top-level tree (treeShape.positions.length). */
+	private readonly stride: number; /** Number of values per top-level node (treeShape.valuesPerTopLevelNode). */
```
CraigMacomber: Put these commends before the thing they are documenting., and fix the one on shape being empty.

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -377,18 +390,61 @@ class Cursor extends SynchronousCursor implements ChunkedCursor {
...
-		if (this.nodePositionInfo === undefined) {
-			assert(positionIndex === 0, 0x561 /* expected root at start */);
+		if (positionIndex === 0) {
+			this.nodePositionInfo = undefined;
 			assert(this.mode === CursorLocationType.Fields, 0x562 /* expected root to be a field */);
+			return;
+		}
+		const offset = positionIndex - 1;
+		if (this.nodeLength === 1) {
+			// Single-node-shape (leaf) fast path: no division needed.
```
CraigMacomber: while the `%` and `/` operations can be slow (though not enough I'd worry about it here), most CPUs should be able to divide by one quite fast (the cycle count for the instruction can vary based on the devisor), and that is likely less overhead than is caused by this branch given that eventually this branch will commonly go either way, making it poorly predicted by the branch predictor and a likely source of pipeline stalls.

Given this is a web app and bundle size is important, and having a single version here would have less code to download and parse, less branch mispredictions (thus might be faster), and less code maintenance costs, I think I'd prefer just having a single case.

If this were not a very performance sensitive code path, then I'd absolutely be for the single code path approach: its only in places like this where we want to consider such optimization. In this case however, I suspect is better to do the simpler smaller code size easier to maintain approach. Feel free to benchmark it if you want though. If doing that, just be sure your test data has a mix of 1 and non-1 lengths to properly account for the branch misprediction costs.

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -377,18 +390,61 @@ class Cursor extends SynchronousCursor implements ChunkedCursor {
...
+			this.topLevelIndex = offset;
+			this.nodePositionInfo = this.treeShape.positions[0];
+		} else {
+			const withinTree = offset % this.nodeLength;
+			this.topLevelIndex = (offset - withinTree) / this.nodeLength;
```
CraigMacomber: Again I don't think this is really worth optimizing, but I suspect you can do this in a faster way.

Sadly javascript doesn't provide a easy way to do integer division giving both the quotient and remainder (often called divmod).

I think this is a bit easier to read

```suggestion
			const withinTree = offset % this.nodeLength; // remainder 
			this.topLevelIndex = Math.trunc(offset / this.nodeLength); // quotient 
```

Might also be faster as it removes the dependency on the first operation from the second so the CPU can run the two divides in parallel, and also makes it clearer the two divides are the same which might allow the jit to optimize it.

Since the values are positive and generally less than 2^32, there are also other hacks for this that are event faster, like the tricks in https://stackoverflow.com/a/17218003

I don't think we should do it (too confusing/fragil) but my guess is something like this might be fastest (this is basically using the hacks that asmjs uses for 32 bit integer math):

```
			const length = this.nodeLength | 0;
			const offset2 = offset | 0;
			const withinTree = offset2 % length; // remainder 
			this.topLevelIndex = (offset2 % length) | 0; // quotient 
```

If we wanted to try for that level of optimization though, I don't think we should be using JavaScript.

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -377,18 +390,61 @@ class Cursor extends SynchronousCursor implements ChunkedCursor {
...
+	 */
+	private materializePath(positionIndex: number): UpPath | undefined {
+		if (positionIndex === 0) {
+			return undefined;
 		}
+		const offset = positionIndex - 1;
+		const withinTree = this.nodeLength === 1 ? 0 : offset % this.nodeLength;
+		const topLevelIndex =
+			this.nodeLength === 1 ? offset : (offset - withinTree) / this.nodeLength;
+		const info = this.treeShape.positions[withinTree] ?? oob();
```
CraigMacomber: This seems like its duplicating logic in moveToPosition. I think we should share a utility for both. Maybe a getInfo method that takes a positionIndex

packages/dds/tree/src/test/feature-libraries/chunked-forest/uniformChunk.spec.ts
```diff
@@ -143,6 +166,252 @@ describe("uniformChunk", () => {
...
+			cursor.enterNode(pointCount - 1);
+			cursor.enterField(xField);
+			cursor.enterNode(0);
+
+			// getPath materializes a standalone path on demand: a fresh object each call,
+			// carrying the real position that was never stored anywhere.
+			const path1 = cursor.getPath();
+			const path2 = cursor.getPath();
+			assert(path1 !== undefined && path2 !== undefined);
+			assert.notEqual(path1, path2);
```
CraigMacomber: Since UpPath is immutable, I don't think its helpful to specify and test that we fail cache/reuse them.

If we wanted to add some caching, it would be nicer to have tests which can confirm that didn't break the required behavior rather than have to update these tests since they ban caching.

brrichards: Makes sense, I'll go ahead and remove this section of the test

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
brrichards: I ended up benchmarking the optimizations you suggested. The fast path doesn't win by enough to make it a needed addition (so I'll go ahead and remove it), but it seemed Math.trunc() was a little slower than the original equation, but this hack allowed for the divides to run in parallel - `this.topLevelIndex = (offset / length) | 0; // quotient `. it wasn't much of a win for mixed forests, but for leaf nodes specifically it gave a small win. To be fair its not even a nanosecond difference. Same for a mixed forest. Since this is on the hot path it still might be worth, but I don't think we would want to add the hack (the linter alone hates it). I believe it would be safe since its bound is on topLevelLength and as long as we don't make that > 2^31. 

So I think it would be best to remove the fast path and then keep the current equation. Maybe leave a comment briefly describing what is happening to make it more readable.

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
brrichards: updated to the comment to match this

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
brrichards: Did some benchmarking and wrote about it in the convo below this one. Ended up removing the fast path

packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
brrichards: Moved to a new method `decodePosition(positionIndex: number)`, along with a comment that hopefully helps with the readability of the offset equation. If we want to name this method something else let me know.

CraigMacomber (approved):
Looks good to me

PR #27468, opened 2026-06-02
Align ADO pipeline package-manager install settings

ChumpChief:
## Description

Followup to #27456 (which removed `--unsafe-perm` from the server Dockerfiles): the pipeline templates still passed `--unsafe-perm` on the `npm ci` install command, though in practice that command wasn't actually getting used. Cleaning that up grew naturally into aligning all the package-manager install settings across pipelines.

Pure refactor — every pipeline runs the same install behavior it ran before. As a forward-looking benefit, flipping the template defaults from `npm` / `'npm ci --unsafe-perm'` to `pnpm` / `'pnpm i --frozen-lockfile'` reduces the risk that a future caller accidentally falls back to npm or `--unsafe-perm` by omitting these parameters.

- Every active caller of `build-docker-service.yml`, `build-npm-package.yml`, `build-npm-client-package.yml`, and `include-install.yml` already set `packageManager: pnpm` and `packageManagerInstallCommand: 'pnpm i --frozen-lockfile'` (or an equivalent). Flipping the templates' defaults to match lets the explicit per-caller settings be removed.
- For `server-*` and `test-perf-benchmarks` the install command gains `--frozen-lockfile`. This is a no-op at runtime because every affected workspace `.npmrc` (root, `ff_pipeline_host`, `server/gitrest`, `server/historian`, `server/routerlicious`) already has `frozen-lockfile=true`; the CLI flag is documentation of intent.
- `server-gitssh` now explicitly sets `packageManager: none`, a new enumerated value on `build-docker-service.yml` that means "no package manager setup" (for Dockerfile-only services with no `package.json`). Behavior matches the previous `npm` default for gitssh because (a) the pnpm-gated template paths skip for any non-`pnpm` value and (b) `setVersion: false` already skipped the install step regardless.

### Mechanical removals

- 11 canonical callers drop both `packageManager` and `packageManagerInstallCommand` (`build-api-markdown-documenter`, `build-benchmark-tool`, `build-build-common`, `build-build-tools`, `build-bundle-size-artifacts`, `build-client`, `build-common-utils`, `build-eslint-config-fluid`, `build-eslint-plugin-fluid`, `build-protocol-definitions`, `build-test-tools`).
- 3 `server-*` docker callers drop `packageManager` and align on `'pnpm i --frozen-lockfile --workspace-root'` (`server-gitrest`, `server-historian`, `server-routerlicious`).
- `test-perf-benchmarks` drops per-call overrides; relies on the new `include-install.yml` defaults.

[AB#74733](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/74733)

### DIFF
--- a/tools/pipelines/build-api-markdown-documenter.yml
+++ b/tools/pipelines/build-api-markdown-documenter.yml
@@ -86,8 +86,6 @@ extends:
     publishOverride: ${{ parameters.publishOverride }}
     releaseBuildOverride: ${{ parameters.releaseBuildOverride }}
     buildDirectory: ${{ variables.FluidFrameworkDirectory }}/tools/api-markdown-documenter
-    packageManagerInstallCommand: 'pnpm i --frozen-lockfile'
-    packageManager: pnpm
     testCoverage: ${{ variables.testCoverage }}
     tagName: api-markdown-documenter
     taskBuild: build
--- a/tools/pipelines/build-benchmark-tool.yml
+++ b/tools/pipelines/build-benchmark-tool.yml
@@ -88,9 +88,7 @@ extends:
     publishOverride: ${{ parameters.publishOverride }}
     releaseBuildOverride: ${{ parameters.releaseBuildOverride }}
     buildToolsVersionToInstall: ${{ parameters.buildToolsVersionToInstall }}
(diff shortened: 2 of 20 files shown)


MarioJGMsoft (approved):
LGTM

PR #27469, opened 2026-06-02
(tree): Added end-to-end tests for incremental summarization

agarwal-navin:
## Description

Added end-to-end test coverage for SharedTree's incremental summarization, exercising the full container + summarizer pipeline. These are the e2e counterpart to the `forestSummarizer.spec.ts` unit tests: where the unit tests drive the `ForestSummarizer` directly and assert on handle counts/paths, these run the real container + summarizer flow (where the stale-handle-path bug surfaced as an actual crash) and assert that every summary succeeds and the document round-trips correctly when loaded into a fresh client.

The new `treeIncrementalSummary.spec.ts` defines a 4-depth nested schema where each level's map field carries `incrementalSummaryHint`, creating four independent incremental chunks, and adds:

- **Smoke test** — produces an incremental summary, loads a new container/summarizer from it, and validates the loaded view matches. Runs against all services and in full compat mode.
- **4-depth schema suite** — parameterized over ordered sequences of depth changes (ascending, descending, repeated depths, and the copy-propagation cases that triggered the stale-path bug). Each round takes an incremental summary and validates the loaded document. Runs only against local, since running the full matrix against all services is expensive. Also, runs in no compat mode because running against all compat is expensive.
- **Sibling + descendant handles** — verifies a single summary that simultaneously contains a sibling chunk handle and a descendant chunk handle.

Test-only change; no production code is modified. The package (`@fluid-private/test-end-to-end-tests`) is private, so no changeset is required.

## Reviewer Guidance

### DIFF
--- a/packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
+++ b/packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
@@ -21,7 +21,6 @@ import {
 import {
 	FluidClientVersion,
 	ForestTypeOptimized,
-	incrementalEncodingPolicyForAllowedTypes,
 	TreeCompressionStrategy,
 	type ITree,
 } from "@fluidframework/tree/alpha";
@@ -54,6 +53,7 @@ async function make4DepthTreeView(
 		TreeViewConfigurationAlpha,
 		incrementalSummaryHint,
 		configuredSharedTree,
+		incrementalEncodingPolicyForAllowedTypes,
 	} = dataRuntimeApi.packages.tree;
 
 	const sf = new SchemaFactoryAlpha("incrementalSummary4DepthE2E");
@@ -101,7 +101,7 @@ async function make4DepthTreeView(
 	const SharedTree = configuredSharedTree({
 		forest: ForestTypeOptimized,
 		treeEncodeType: TreeCompressionStrategy.CompressedIncremental,
-		minVersionForCollab: FluidClientVersion.v2_74,
+		minVersionForCollab: FluidClientVersion.v2_80,
 		shouldEncodeIncrementally: incrementalEncodingPolicyForAllowedTypes(viewConfig),
 	});
 
@@ -255,7 +255,7 @@ async function loadContainerAndGetTreeView(
 	summaryVersion?: string,
 ) {
 	return make4DepthTreeView(
-		apis.dataRuntimeForLoading ?? apis.dataRuntime,
+		apis.dataRuntimeForLoading,
 		provider,
 		"load",
 		undefined,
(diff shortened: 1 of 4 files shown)


packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+ * - Depth 4: each item's `tags` map (innermost chunk).
+ *
+ * The root field `version` (depth 0) is non-incremental and does not belong to any chunk.
+ *
+ * @param variant - `"single"` initializes one document/section/item/tag; `"double"` adds a
+ * second sibling section (`Sec2`) so that sibling chunks become handles alongside descendant
+ * handles in the same summary.
+ */
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
Josmithr: Nit: any reason we can't define a return type? That would give us a place to document the properties of the returned property bag.

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+ * A 4-depth nested schema where each level's map field carries `incrementalSummaryHint`,
+ * creating 4 independent incremental chunks:
+ * - Depth 1: the `documents` map (outermost chunk).
+ * - Depth 2: each document's `sections` map.
+ * - Depth 3: each section's `items` map.
+ * - Depth 4: each item's `tags` map (innermost chunk).
+ *
+ * The root field `version` (depth 0) is non-incremental and does not belong to any chunk.
+ *
+ * @param variant - `"single"` initializes one document/section/item/tag; `"double"` adds a
```
Josmithr, line 66: Nit: can we document the other params as well?

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+			custom: { [incrementalSummaryHint]: true },
+		}),
+	}) {}
+
+	const viewConfig = new TreeViewConfigurationAlpha({ schema: Workspace });
+
+	const SharedTree = configuredSharedTree({
+		forest: ForestTypeOptimized,
+		treeEncodeType: TreeCompressionStrategy.CompressedIncremental,
+		minVersionForCollab: FluidClientVersion.v2_80,
```
Josmithr: Is this min version required for anything specific? If so, should we leave a comment?

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+		shouldEncodeIncrementally: incrementalEncodingPolicyForAllowedTypes(viewConfig),
+	});
+
+	const treeId = "sharedTree";
+	const testContainerConfig: ITestContainerConfig = {
+		fluidDataObjectType: DataObjectFactoryType.Test,
+		runtimeOptions: {
+			enableRuntimeIdCompressor: "on",
+			summaryOptions: {
+				summaryConfigOverrides: { state: "disabled" },
```
Josmithr, line 156: Nit: I'm assuming it's important that state is disabled here? Should we add a comment?

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+
+async function createTestSummarizer(
+	provider: ITestObjectProvider,
+	container: IContainer,
+	testContainerConfig: ITestContainerConfig,
+	summaryVersion?: string,
+): Promise<ISummarizer> {
+	const summarizerContainerConfig: ITestContainerConfig = {
+		...testContainerConfig,
+		runtimeOptions: { ...testContainerConfig.runtimeOptions, summaryOptions: undefined },
```
Josmithr, line 318: Nit: worth a comment?

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+			// These tests only run on local service.
+			// They are exhaustive tests of the incremental summary feature. Running on all
+			// services is expensive and takes a long time.
+			if (provider.driver.type !== "local") {
+				this.skip();
+			}
+		});
+
+		/**
+		 * Ordered sequences of depth changes, mirroring the parameterized unit test cases. Each change
```
Josmithr, line 418: Sharing these test cases / infra with the unit tests seems like way more effort than is merited here. But it might be worth adding a note to the top of each of the relevant test modules noting that changes to one should be mirrored in the other?

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+					assert(summaryInfo !== undefined, `Round ${round} should produce a summary`);
+					await assert.doesNotReject(
+						validateContainerLoad(summaryInfo.summaryVersion),
+						`Loading of summary version ${summaryInfo.summaryVersion} should succeed`,
+					);
+				}
+			});
+		}
+
+		it("simultaneous sibling and descendant handles when only one section's items change", async () => {
```
Josmithr: Nit: this test name is hard to parse (at least for me). Might be worth re-phrasing.

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
```diff
@@ -0,0 +1,505 @@
...
+			// Doc1 has two sections (Sec1 and Sec2). Changing Item1.itemName in Sec1 (a depth-3
+			// change) re-encodes the documents, Doc1.sections, and Sec1.items chunks, while Sec2.items
+			// (a sibling at depth 3) and Item1.tags (a descendant at depth 4) become handles in the
+			// same summary. Repeating the change exercises the stale-path bug for both handles.
```
Josmithr: Stale path bug? Can we elaborate on this?

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
agarwal-navin: The schema is created inside this and creating a return type for it isn't straightforward. I don't think it's worth writing a complex type for it.

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
agarwal-navin: I should probably write a doc / section about writing tests that summarize. It is complicated and not apparent.

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
agarwal-navin: Added that. However, I don't think all the tests need to be mirrored. These are basic tests that give good e2e coverage. Unit tests can be richer and cover more scenarios - we don't have e2e tests to run as many scenarios because they are expensive.

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
agarwal-navin: Added a description and link to the PR.

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
agarwal-navin: Updated. Please take a look.

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
Josmithr: Probably worth a quick comment. We have an upcoming linter rule that will require comments to accompany disables anyways.

Josmithr (approved):
Left one more nitpick suggestions. Otherwise looks good to me!

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
agarwal-navin: The comment is in the block above - I will move it here.

agarwal-navin: done

Josmithr (approved):

packages/test/test-end-to-end-tests/src/test/treeIncrementalSummary.spec.ts
Josmithr: Ah, I missed that. My bad. But the new position does likely make it a bit more discoverable. Thanks!

PR #27472, opened 2026-06-03
Remove no longer needed pnpm hoist settings for eslint and prettier, and a few stray prettier dependencies

CraigMacomber:
## Description

Our top level workspace no longer uses prettier, so we don't need to hoist its packages. I also removed some unused deps on prettier.

We still do use eslint, but the need for its package hoisting was related to issues with how it found plugins in older configuration which no longer appear to apply.

## Reviewer Guidance

### DIFF
--- a/packages/framework/tree-agent-langchain/package.json
+++ b/packages/framework/tree-agent-langchain/package.json
@@ -139,7 +139,6 @@
 		"jiti": "^2.6.1",
 		"mocha": "^11.7.5",
 		"mocha-multi-reporters": "^1.5.1",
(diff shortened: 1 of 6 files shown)


ChumpChief (commented):
Seems good but assume we should resolve the CI issue first.

CraigMacomber:
> Seems good but assume we should resolve the CI issue first.

Issues should be resolved now. https://github.com/microsoft/FluidFramework/pull/27478 is in. Rerunning PR CI now after pulling that in.

ChumpChief (approved):

PR #27474, opened 2026-06-03
Suppress minimumReleaseAge check on CI to workaround bad publish times

CraigMacomber:
## Description

Our internal CI pipelines use a package mirror with its own package ingestion delay/quarantine, so this check is rather redundant there. It also doesn't work, and having users apply the check when installing new package versions should also be sufficient.

Therefor suppressing this check on CI seems reasonable and necessary.

## Reviewer Guidance


CraigMacomber:
Validation for this fix running in https://dev.azure.com/fluidframework/internal/_build/results?buildId=404166&view=results

Edit: confirmed to fix the issue!

CraigMacomber:
Looks like docker image stuff is broken again, so this PR is failing on CI despite fixing the break to internal pipelines from the pnpm 11 update

PR #27475, opened 2026-06-03
Suppress minimumReleaseAge check on CI to workaround bad publish times

CraigMacomber:
## Description

Our internal CI pipelines use a package mirror with its own package ingestion delay/quarantine, so this check is rather redundant there. It also doesn't work, and having users apply the check when installing new package versions should also be sufficient.

Therefor suppressing this check on CI seems reasonable and necessary.

## Reviewer Guidance


CraigMacomber:
This is from https://github.com/microsoft/FluidFramework/pull/27474 but moved to an upstream test branch to work around known docker CI issue.

tools/pipelines/templates/include-install-pnpm.yml
```diff
@@ -118,3 +118,7 @@ steps:
       # Disable pnpm trust downgrade checks in CI as a workaround.
       echo "Disabling pnpm trust policy as workaround for npm mirror limitations"
       echo "##vso[task.setvariable variable=PNPM_CONFIG_TRUST_POLICY]off"
+      # Some mirrored registries also have unreliable publish timestamps.
```
CraigMacomber:
```suggestion
      # Our mirrored registries also have unreliable publish timestamps.
```

Abe27342 (approved):

PR #27479, opened 2026-06-03
refactor(eslint-config-fluid): no-restricted-syntax selector in base rule set that flags the named Type import from TypeBox

TommyBrosman:
Note: Closed in favor of PR #27482 due to CI issues with remote branches.

## Description

Adds a `no-restricted-syntax` selector to the base rule set that flags the named
`Type` import from TypeBox:

```ts
import { Type } from "@sinclair/typebox"; // now an error
```

See #27442 for the motivating PR.

The named `Type` export is the monolithic `TypeBuilder` aggregate, so importing it
pulls in every builder and defeats tree-shaking. The rule directs consumers to the
namespace form instead, which lets bundlers prune unused builders:

```ts
import * as Type from "@sinclair/typebox";
Type.Object({ /* ... */ });
```

This can't be expressed with `no-restricted-imports` / `importNames`, because that
would also report the desired `import * as Type` namespace form. A syntax selector
targets only the named specifier.

The selector is added inline to the base `no-restricted-syntax` rule (alongside the
existing `ExportAllDeclaration` and `ForInStatement` entries) rather than being
factored into a shared exported constant, keeping the change minimal. The checked-in
printed configs have been regenerated to match.


tylerbutler:
You may want to update the changelog with this change, since presumably you want to release it to take advantage of it in client.

alexvy86 (approved):
The change looks good to me. The issue in server/routerlicious is a known problem (see recent Teams threads in our Dev channel), the workaround is to create the PR from a branch in upstream. So my approval here is probably moot in practice, but take it in spirit 😄 .

PR #27488, opened 2026-06-04
test (docs) Guide for writing e2e tests that take summaries

agarwal-navin:
## Description

Adds a new doc, `WritingTestsThatTakeSummaries.md`, to the `test-end-to-end-tests` package documenting the pattern for writing end-to-end tests that drive summaries on demand. It covers:

- The two container configurations — interactive containers (with the automatic summarizer disabled) vs. the dedicated summarizer.
- How to take a summary with `summarizeNow` and inspect the result.
- How to load a new container or summarizer from a specific summary version.
- The rules that keep these tests deterministic (`syncSummarizer: true`, `ensureSynchronized()` before summarizing, chaining via `summaryVersion`, closing one summarizer before starting another, etc.).

It also adds a Table of Contents and an Introduction section to the package `README.md` and `WritingCompatCorrectTests.md` for consistency and navigation, links the new doc from the README, and fixes a heading typo ("Enpdoint" → "Endpoint") in the README.

Documentation-only change — no code or public API is affected.

### DIFF
--- a/packages/test/test-end-to-end-tests/WritingTestsThatTakeSummaries.md
+++ b/packages/test/test-end-to-end-tests/WritingTestsThatTakeSummaries.md
@@ -22,21 +22,25 @@
 
 ## Introduction
 
-This document explains the pattern for writing end-to-end (e2e) tests that generate **summaries**. The canonical examples live under [`src/test/summarization/`](./src/test/summarization/).
+This document explains the pattern for writing end-to-end (e2e) tests that generate **summaries**.
+The canonical examples live under [`src/test/summarization/`](./src/test/summarization/).
 
 ## Why a dedicated summarizer
 
-In production, summaries are produced by a single elected **summarizer client** - a non-interactive container that the runtime spins up in the background and generates summaries based on heuristics. In a test you want to control _exactly when_ a summary happens and _what_ goes into it, so you:
+In production, summaries are produced by a single elected **summarizer client** - a non-interactive container that the runtime spins up in the background and generates summaries based on heuristics.
+In a test you want to control _exactly when_ a summary happens and _what_ goes into it, so you:
 
 1. Create your normal interactive container(s) with the runtime's automatic summarizer **disabled**, so nothing summarizes behind your back.
 2. Create a separate **summarizer container** that summarizes only when you call `summarizeNow` on it.
 3. Synchronize all clients before each summary so the summary is deterministic.
 
-This separation is the heart of the pattern. Everything below follows from it.
+This separation is the heart of the pattern.
+Everything below follows from it.
 
 ## How to configure the regular (interactive) containers
 
-Every interactive container the test creates or loads must have the runtime's automatic summarizer turned **off**. Otherwise a background summary can race your `summarizeNow` calls and your assertions, making the test flaky.
+Every interactive container the test creates or loads must have the runtime's automatic summarizer turned **off**.
+Otherwise a background summary can race your `summarizeNow` calls and your assertions, making the test flaky.
 
 Disable it via `summaryConfigOverrides: { state: "disabled" }` in the container config:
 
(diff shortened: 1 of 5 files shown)


packages/test/test-end-to-end-tests/src/test/summarization/README.md
```diff
@@ -0,0 +1,19 @@
...
+## Writing a summarization test
+
+If you're here to learn how to write a test that takes summaries, read
+**[WritingTestsThatTakeSummaries.md](../../../WritingTestsThatTakeSummaries.md)**. It covers the
+container configurations to use, how to take and inspect a summary with `summarizeNow`, how to
+load from a specific `summaryVersion`, and the rules that keep these tests deterministic
+(`syncSummarizer: true`, calling `ensureSynchronized()` before summarizing, closing one
+summarizer before starting another, etc.).
+
+For worked examples, see the specs in this folder — for instance
```
markfields: nit?
```suggestion
For working examples, see the specs in this folder — for instance
```

markfields (approved):
This is great!

packages/test/test-end-to-end-tests/src/test/summarization/README.md
```diff
@@ -0,0 +1,18 @@
+# @fluidframework/test-end-to-end-tests/test/src/test/summarization
```
Josmithr, line 1: Nit: our best practices docs for Markdown recommend line-breaking along sentence boundaries. This helps make future changes easier to review.
https://github.com/microsoft/FluidFramework/wiki/Markdown-Best-Practices#line-breaks-along-sentence-boundaries

packages/test/test-end-to-end-tests/WritingTestsThatTakeSummaries.md
```diff
@@ -0,0 +1,268 @@
...
+interface SummaryInfo {
+	summaryTree: ISummaryTree; // the generated summary tree — inspect it for handles/blobs
+	summaryVersion: string;    // the acked summary handle — use this to load from this summary
+	summaryRefSeq: number;     // reference sequence number of this summary
+}
+```
+
+The minimal round looks like this:
+
+```ts
```
Josmithr, line 110: Since this is the first code example demonstrating the test flow, it might be worth including additional boilerplate beforehand, including where `provider` comes from (or at least a quick note about it)

Josmithr: Actually, having read the rest of the document, including more code here probably doesn't make sense. But a quick note that `provider` will be explained more later in the document might be useful.

packages/test/test-end-to-end-tests/WritingTestsThatTakeSummaries.md
```diff
@@ -0,0 +1,268 @@
...
+	summaryVersion,
+);
+await summarizeNow(summarizer2);
+```
+
+## A complete minimal example
+
+Putting it together — create, summarize, load-and-validate, then summarize from a new summarizer:
+
+```ts
```
Josmithr: A thought: since we don't want this to get out-of-date, it might be worth checking this in as an actual test in code (if it isn't already). Then we could use our markdown-magic pragmas to embed the code here, rather than copying it. That way, if we end up having to update this example for one reason or another, our docs won't become out-of-date.

Josmithr: An example of this pattern can be found here: https://github.com/microsoft/FluidFramework/blob/main/tools/api-markdown-documenter/README.md?plain=1#L225

Josmithr (approved):
I left a couple nitpicks and mechanical suggestions, but the docs look great! Thanks for doing this!

Josmithr (approved):

PR #27490, opened 2026-06-04
Bump eslint-config-fluid version to 13.0

TommyBrosman:
This is a version bump for the changes in #27482 which add TypeBox import pattern syntax exclusions.


common/build/eslint-config-fluid/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@fluidframework/eslint-config-fluid",
-	"version": "12.0.0",
+	"version": "13.0.0",
```
CraigMacomber, line 3: I screwed up with the 11 release. After all the struggles failing to create it due to bad repo rules, when I did eventually manage to create it, I apparently made it from the wrong commit, which was after the version got bumped to 12, so version 11 got skipped and 12 was published from the 11 branch.

In https://github.com/microsoft/FluidFramework/pull/27482/changes#diff-eab6625198198b0f9c22279da80ea1591528fbcf8e4111f10908161dc70f9142R3 you said those changes will be in version 13. However only changed which merge after this PR would normally be in 13: this PR is supposed to be the cutoff for what is in 12 vs 13.

In the end this works out for you though, as we can't release 12 from the proper commit, as that already happened, so we have to skip it and it in 13 anyway, making your changelog correct.

So yes, this PR is good to go, but annoyingly we will need to follow it up with another similar one bumping to 14 to delineate 13 vs 14 separation.

CraigMacomber (approved):

common/build/eslint-config-fluid/package.json
TommyBrosman: v14 bump: #27522

PR #27492, opened 2026-06-05
tree: Fix bug with node existence revert constraint

alex-pardes:
## Description

Fixed a bug where if a node existence revert constraint was placed on a node moved in the same transaction, the revert constraint would be violated when rebasing over any change. This PR changes the implementation of the node existence revert constraint to  detect violations during invert instead of during rebasing.


packages/dds/tree/src/test/shared-tree/editing.spec.ts
```diff
@@ -3218,6 +3218,34 @@ describe("Editing", () => {
 
 				stack.unsubscribe();
 			});
+
+			it("inverse node existence constraint not violated by unrelated change on moved node", () => {
```
yann-achard-MS:
```suggestion
			it("inverse node existence constraint on moved node not violated by unrelated change", () => {
```

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
```diff
@@ -1410,43 +1411,30 @@ export class ModularChangeFamily
...
-		parentOutputAttachState: NodeAttachState,
 		constraintState: ConstraintState,
-		revertConstraintState: ConstraintState,
 		nodes: ChangeAtomIdBTree<NodeChangeset>,
 	): void {
 		for (const field of fields.values()) {
 			const handler = getChangeHandler(this.fieldKinds, field.fieldKind);
-			for (const [nodeId, inputIndex, outputIndex] of handler.getNestedChanges(field.change)) {
+			for (const [nodeId, inputIndex, _outputIndex] of handler.getNestedChanges(
+				field.change,
```
yann-achard-MS, line 1420: Looks like this was the only code path where the output index was used. You could remove it from the `getNestedChanges` contract (and all computations of it) as part of this PR, or we could do that in a separate PR, I don't mind.

alex-pardes: I think that can be a separate PR.

yann-achard-MS (approved):

PR #27497, opened 2026-06-05
improvement(client-ordered-collection): ConsensusQueueFactory generic

jason-ha:
- Remove unused `IConsensusOrderedCollectionFactory`
- Make `ConsensusQueueFactory` generic (assumes `unknown`)
  - Get specific on `create` & `load` return types
- Remove most intended internal class use in test


packages/dds/ordered-collection/src/index.ts
```diff
@@ -8,7 +8,6 @@ export {
 	ConsensusResult,
 	type IConsensusOrderedCollection,
 	type IConsensusOrderedCollectionEvents,
-	type IConsensusOrderedCollectionFactory,
 	type IOrderedCollection,
 	type ISnapshotable,
 } from "./interfaces.js";
```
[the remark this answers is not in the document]
jason-ha: `IConsensusOrderedCollectionFactory` was `@internal` and changes to it including removal do not break public semantic versioning.

packages/dds/ordered-collection/src/consensusOrderedCollectionFactory.ts
```diff
@@ -23,7 +25,9 @@ import { pkgVersion } from "./packageVersion.js";
  *
  * @internal
  */
-export class ConsensusQueueFactory implements IConsensusOrderedCollectionFactory {
+export class ConsensusQueueFactory<T>
+	implements IChannelFactory<IConsensusOrderedCollection<T>>
+{
```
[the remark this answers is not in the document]
jason-ha: nah

packages/dds/ordered-collection/src/test/consensusOrderedCollection.spec.ts
```diff
@@ -32,7 +34,7 @@ import {
 /**
  * Test class that exposes protected applyStashedOp method for testing
  */
-class TestConsensusQueue extends ConsensusQueueClass {
+class TestConsensusQueue extends ConsensusQueueClass<unknown> {
```
CraigMacomber, line 36: Personally, I do not like the pattern of using generics such that they can cause incorrectly typed code to silently compile, especially when there is no runtime check involved.

There is no way for the TypeScript compiler to validate that the content stored in all possible current and future documents makes the type provided here, and not even a way for a careful developer to ensure its correct either. When importing persisted data, I don't think we should use generics to simply assume its what ever type the code feels like inferring, removing the compiler enforced validation of the data and hiding where the unsafe type conversion is occurring.

If you store Foos in your document, then a dev changes the Foo type in an incompatible way, or even uses a totally different type, there is nothing we can do at compile time to stop them, checked runtime conversion is required, and developers know where to do that because get type Unknown, and are forced to convert.

Anything that implicitly type converts content from persistence should require a runtime check: this is why we use strongly type schema like Tree schema and TypeBox. I'd rather not add violations to this typer safety practice in legacy code to make it slightly more ergonomic to write type unsafe usages of it.

jason-ha: I don't think that is an opinion unique to you. I was equally put off when I saw what is going on here.
Unfortunately, we don't have all the time to go address past silliness. This at least moves the generic to `unknown` here instead of `any`.

packages/dds/ordered-collection/src/consensusOrderedCollectionFactory.ts
```diff
@@ -43,24 +47,26 @@ export class ConsensusQueueFactory implements IConsensusOrderedCollectionFactory
...
-	 * {@inheritDoc @fluidframework/datastore-definitions#IChannelFactory.load}
-	 */
 	public async load(
 		runtime: IFluidDataStoreRuntime,
 		id: string,
 		services: IChannelServices,
 		attributes: IChannelAttributes,
-	): Promise<IConsensusOrderedCollection> {
+	): Promise<
+		IConsensusOrderedCollection<T> & SharedObjectCore<IConsensusOrderedCollectionEvents<T>>
```
ChumpChief: In other places we extend ISharedObject (and IConsensusOrderedCollection already extends that).  Seems a little odd to include the abstract class SharedObjectCore here, and I don't see any other precedent for this pattern?  What's the motivation?

jason-ha: The local test code wants it.
packages/dds/ordered-collection/src/test/consensusOrderedCollection.spec.ts does a number of `handle.absolute` checks.

ChumpChief: It would be preferable then to convert the tests to use `toFluidHandleInternal()` or `compareFluidHandles()`, which are the recommended way of accessing the `absolutePath` or doing handle comparison.

ChumpChief (approved):
Nice improvement, thanks :)

PR #27500, opened 2026-06-05
fix(datastore): Log when a DDS fails to load, including dataStore metadata

markfields:
## Description

When a DDS fails to load, we need to know what DataStore it was under for diagnostic purposes.  But in some cases (e.g. if it's the Summarizer client), the stack that first awaits the `getChannel` call doesn't afford any insight into which DataStore it was, and we just get the raw error at a higher level choke point (e.g. Summarizer failing and closing), with no helpful context besides what we can glean from the minified stack.

This PR adds a catch-log-throw block around the internal `channelP` promise so as soon as it rejects, we get a log that includes DataStore metadata (its ID and package path).


markfields:
Tagging as `release-blocking`, this is blocking the safe rollout of a feature in a 1P partner

agarwal-navin (approved):

PR #27505, opened 2026-06-08
Simplify pipelines

CraigMacomber:
## Description

Simplify pipelines a bit. This includes doing part of AB#:75329.

## Reviewer Guidance


tools/pipelines/templates/build-docker-service.yml
```diff
@@ -316,22 +314,27 @@ extends:
               buildDirectory: ${{ parameters.buildDirectory }}
               requireNotice: ${{ parameters.shouldReleaseDockerImage }}
 
-        - ${{ if eq(parameters.packageManager, 'pnpm') }}:
+        - ${{ if ne(parameters.skipPackageInstall, true) }}:
```
ChumpChief: Does this work?

```suggestion
        - ${{ if not(parameters.skipPackageInstall) }}:
```

ChumpChief: Same for other `eq` and `ne` locations below

tools/pipelines/templates/build-docker-service.yml
```diff
@@ -301,8 +299,8 @@ extends:
                 fi
               fi
 
-              if [[ "${{ parameters.packageManager }}" == "none" ]] && [[ "${{ parameters.setVersion }}" == "True" ]]; then
-                echo "##vso[task.logissue type=error]packageManager: 'none' is incompatible with setVersion: true (version-setting transitively requires a package manager)."
+              if [[ "${{ parameters.skipPackageInstall }}" == "true" ]] && [[ "${{ parameters.setVersion }}" == "True" ]]; then
```
ChumpChief: Should this be `True` rather than `true`?

tools/pipelines/templates/build-npm-client-package.yml
```diff
@@ -37,10 +52,6 @@ parameters:
   type: boolean
   default: false
 
-- name: isBundleSizeArtifactsPipeline
```
ChumpChief, line 40: Might be preferable to leave the bundle size stuff for now, Tommy and I are making changes in this area and some of this is probably going to change anyway.  Since the bundle size artifacts live in the client-side of the world it may make more sense for them to be produced over here...?

tools/pipelines/deploy-website.yml
```diff
@@ -130,22 +130,10 @@ stages:
 
         - template: /tools/pipelines/templates/include-use-node-version.yml@self
 
-        - template: /tools/pipelines/templates/include-install-pnpm.yml@self
+        - template: /tools/pipelines/templates/include-install.yml@self
           parameters:
             buildDirectory: ${{ variables.FluidFrameworkDirectory }}/docs
 
-        - task: Bash@3
```
ChumpChief, line 137: Was this just duplicate?

tools/pipelines/templates/build-npm-client-package.yml
CraigMacomber: Recently you updated the bundle size stuff in the other pipeline, the one we actually use for it, and left this one out of date. I find having the code duplicated here where it is dead code that's not fully up to date confusing. I do agree that it is probably better to not touch that in this PR though, as that's a whole different mess that the rest of these changes.

tools/pipelines/deploy-website.yml
CraigMacomber: This pretty much just an inline copy of include-install.yml (except for that pwd statement I added at some point) which it now just uses.

tools/pipelines/templates/build-npm-client-package.yml
CraigMacomber: I have reverted my changes in this area

tools/pipelines/templates/build-docker-service.yml
CraigMacomber: The code above uses both, which I suppose is likely a bug. I'll look into this more.

CraigMacomber: Looks like the code using lowercase was bugged, and that validation could never trigger. Docs for this are in https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#boolean-1 . Fixed here and there.

tools/pipelines/templates/build-docker-service.yml
CraigMacomber: Correct. Fixed them all.

tools/pipelines/templates/build-docker-service.yml
```diff
@@ -283,7 +281,7 @@ extends:
               fi
 
               if [[ "$(release)" == "prerelease" ]]; then
-                if [[ "${{ parameters.buildNumberInPatch }}" == "true" ]]; then
+                if [[ "${{ parameters.buildNumberInPatch }}" == "True" ]]; then
```
CraigMacomber, line 284: This was a bug: this would never be hit due to incorrect case.

ChumpChief: Is there a good reason here and other places are doing string comparison, can we just evaluate the boolean directly?

Not needed for this PR but wondering if we should file a followup bug.

tools/pipelines/templates/build-docker-service.yml
```diff
@@ -65,17 +65,15 @@ parameters:
...
 
-  # The package manager (if any) used by this pipeline.
-  - name: packageManager
-    type: string
-    default: pnpm
-    values:
-      - pnpm
-      - none
+  # If true, skip package-manager setup/install entirely (used by packages without package.json, e.g. gitssh).
+  - name: skipPackageInstall
```
ChumpChief: Nit for consideration but optional for this PR:

Consider inverting to `installPackages` with a `true` default, most conditionals below are `not()`ing it so this would probably read more easily:

```
${{ if not(parameters.skipPackageInstall) }}
```

```
${{ if parameters.installPackages }}
```

ChumpChief (approved):
Nit and a question, but doesn't have to block checkin.  If you want to take the nit here let me know and I'll reapprove.

tools/pipelines/templates/build-docker-service.yml
CraigMacomber: Do you mean refactor this to move the `if` condition out to the ADO template level instead of in the bash script? As it's in the middle of a nested if condition in the bash script, that seems difficult.

Perhaps you mean some way to have bash test these environment variable strings without comparing them to a literal? Or is this even using the environment variable ADO generates, it is it interpolating string here... So many layers of interpolation and variables in ADO pipelines, not to mention how much of a mess bash is.... Maybe there is a better way, but I don't know it off hand.

Also, not sure what bug we would file. I fixed the bug here. We could file a task to refactor our scripts though.

ChumpChief: Yeah I think you said what I was thinking :-D

Basically I'm wondering why this isn't

```
if [[ ${{ parameters.buildNumberInPatch }} ]]; then
```

CraigMacomber: Since "False" is more than zero characters, it evaluates to true if you do that as far as I can tell, see https://www.gnu.org/software/bash/manual/bash.html#Bash-Conditional-Expressions

CraigMacomber: You can write `if` conditions without using `[[` for true and false (lowercase) since those exist as binaries on the path which get executed and have the corresponding return codes, but since ADO uppercases them, that won't work either. If the value were a string which was either empty or True, your suggestion would work, but I believe ADO hadn't even handle empty strings properly, and uses False for booleans, so I think thats not an option either.

ChumpChief: Ok no worries - wasn't sure if there was any way to get around the conversion to capitalized strings.  Thank you for the detailed explanation!

tools/pipelines/templates/build-docker-service.yml
CraigMacomber: Good idea. Done.

ChumpChief (approved):

PR #27508, opened 2026-06-08
Remove unused credential flows in e2e tests and examples

jzaffiro:
Now that the new credential format has been implemented (#26714) and tested, remove the old formats from both the real service tests and the Fluid examples. 

[AB#74331](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/74331)

### DIFF
--- a/examples/utils/example-webpack-integration/src/tenantUtils.ts
+++ b/examples/utils/example-webpack-integration/src/tenantUtils.ts
@@ -50,14 +50,14 @@ export async function getTestTenantCheckoutClient(): Promise<TestTenantCheckoutC
 	const packageImportLocation = process.env.token__package__import__location;
 	if (packageImportLocation === undefined) {
 		throw new Error(
-			'The FIC credential flow relies on a test tenant checkout client, but no client was found. Ensure that the environment variable "token__package__import__location" is set to the location of a package that exports a compatible client.',
+			"The FIC credential flow relies on a test tenant checkout client, but no client was found. Populate this environment variable by running the @ff-internal/tenant-setup script.",
 		);
 	}
 
 	const pkg = (await import(packageImportLocation)) as TestTenantCheckoutClient;
 	if (typeof pkg.fetchFicTokens !== "function") {
 		throw new TypeError(
-			`Expected package at '${packageImportLocation}' to export fetchFicTokens.`,
+			`Expected package at '${packageImportLocation}' to export fetchFicTokens. Run the @ff-internal/tenant-setup script to populate this environment variable.`,
 		);
 	}
 	// eslint-disable-next-line require-atomic-updates
@@ -79,7 +79,9 @@ export function getOdspCredentials(
 ): LoginCredentials[] {
 	const ficAccounts = process.env.login__odsp__fic__test__users;
 	if (ficAccounts === undefined) {
-		throw new Error("login__odsp__fic__test__users is not defined.");
+		throw new Error(
+			"login__odsp__fic__test__users is not defined. Run the @ff-internal/tenant-setup script to populate this environment variable.",
+		);
 	}
 	const { usernames } = JSON.parse(ficAccounts) as {
 		usernames: string[];
@@ -87,7 +89,7 @@ export function getOdspCredentials(
 
 	if (usernames.length === 0) {
 		throw new Error(
-			"login__odsp__fic__test__users was defined but does not have any valid usernames.",
+			"login__odsp__fic__test__users was defined but does not have any valid usernames. Run the @ff-internal/tenant-setup script to populate this environment variable.",
 		);
 	}
 	return usernames.map((username) => getFicLoginCredentials(username, odspEndpointName));
@@ -108,12 +110,14 @@ const getFicLoginCredentials = (
 			// This error indicates a mismatch between the dynamically imported token fetcher package and this code.
 			// Double-check that the package specified in 'token__package__import__location' is up to date and its entrypoint
 			// matches the typing of `fetchFicTokens` as defined in `TestTenantCheckoutClient`.
-			throw new TypeError("Expected fetchFicTokens to return an array of tokens.");
+			throw new TypeError(
+				"Expected fetchFicTokens to return an array of tokens. Run the @ff-internal/tenant-setup script to populate this environment variable.",
+			);
 		}
 		const token = tokens.find((a) => a.UserPrincipalName === username);
 		if (!token) {
 			throw new Error(
-				`Unable to fetch token for user '${username}' and scope '${scopeEndpoint}'`,
+				`Unable to fetch token for user '${username}' and scope '${scopeEndpoint}'. Run the @ff-internal/tenant-setup script with the correct endpoint for the desired environment.`,
 			);
 		}
 		return token.Token;
--- a/examples/utils/example-webpack-integration/src/tenantUtils.ts
+++ b/examples/utils/example-webpack-integration/src/tenantUtils.ts
@@ -90,10 +90,10 @@ export function getOdspCredentials(
 			"login__odsp__fic__test__users was defined but does not have any valid usernames.",
 		);
 	}
-	return usernames.map((username) => ficLoginCredentials(username, odspEndpointName));
+	return usernames.map((username) => getFicLoginCredentials(username, odspEndpointName));
 }
 
-const ficLoginCredentials = (
+const getFicLoginCredentials = (
 	username: string,
 	odspEndpointName: OdspEndpoint,
 ): LoginCredentials => {
--- a/examples/utils/example-webpack-integration/src/tenantUtils.ts
+++ b/examples/utils/example-webpack-integration/src/tenantUtils.ts
@@ -1,0 +1,127 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Types of Odsp endpoints.
+ * @internal
+ */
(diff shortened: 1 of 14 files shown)


packages/utils/tool-utils/src/odspTokenManager.ts
```diff
@@ -272,21 +223,6 @@ export class OdspTokenManager {
...
-		clientConfig: IPublicClientConfig,
-		username: string,
-		password: string,
-	): Promise<IOdspTokens> {
-		const credentials: TokenRequestCredentials = {
-			grant_type: "password",
-			username,
-			password,
-		};
-		return fetchTokens(server, scope, clientConfig, credentials);
```
Abe27342, line 287: is `fetchTokens` used still? (fine if so, but check please :))

jzaffiro: No longer used - that and its caller (`refreshTokens`) have been removed.

packages/utils/tool-utils/src/odspTokenManager.ts
```diff
@@ -160,8 +146,8 @@ export class OdspTokenManager {
 
 	private async getTokens(
 		isPush: boolean,
-		server: string,
-		clientConfig: IPublicClientConfig,
+		_server: string,
+		_clientConfig: IPublicClientConfig,
```
alexvy86: These two aren't used at all anymore, right? Should they just be removed?

alexvy86 (commented):
The changes to examples/ concern me a bit... without understanding them 100%, it seems like things are changing such that only Microsoft devs would be able to run examples targeting odsp? (Because running tenant-setup requires interaction with internal Microsoft systems). cc @ChumpChief for thoughts on that part since he refactored things recently-ish in that space.

examples/utils/example-webpack-integration/src/tenantUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+export async function getTestTenantCheckoutClient(): Promise<TestTenantCheckoutClient> {
+	if (testTenantCheckoutClientCached !== undefined) {
+		return testTenantCheckoutClientCached;
+	}
+	// An internal package checks out test tenants, populates user information in the environment, and makes an entrypoint available
+	// at this location (token__package__import__location) which supports fetching tokens for those users.
+	const packageImportLocation = process.env.token__package__import__location;
+	if (packageImportLocation === undefined) {
+		throw new Error(
+			'The FIC credential flow relies on a test tenant checkout client, but no client was found. Ensure that the environment variable "token__package__import__location" is set to the location of a package that exports a compatible client.',
```
ChumpChief: This error and the other errors below are hard to understand to the average user who is not immersed in FIC world :)

Recommend only giving directives that a customer who hits this will know how to do (e.g. "run this command" or "see this documentation page on what to do").

Also nit that we use double-quotes.

```suggestion
			"The FIC credential flow relies on a test tenant checkout client, but no client was found. Run my really cool command to provide it",
```

jzaffiro:
> The changes to examples/ concern me a bit... without understanding them 100%, it seems like things are changing such that only Microsoft devs would be able to run examples targeting odsp? (Because running tenant-setup requires interaction with internal Microsoft systems). cc @ChumpChief for thoughts on that part since he refactored things recently-ish in that space.

I don't think this is much different than requiring a script to get an old tenant to target ODSP, which was the case before. Please correct me if I'm wrong though.

examples/utils/example-webpack-integration/src/tenantUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+
+	if (usernames.length === 0) {
+		throw new Error(
+			"login__odsp__fic__test__users was defined but does not have any valid usernames.",
+		);
+	}
+	return usernames.map((username) => ficLoginCredentials(username, odspEndpointName));
+}
+
+const ficLoginCredentials = (
```
ChumpChief: Nit, prefer a verb name


```suggestion
const getFicLoginCredentials = (
```

ChumpChief:
> > The changes to examples/ concern me a bit... without understanding them 100%, it seems like things are changing such that only Microsoft devs would be able to run examples targeting odsp? (Because running tenant-setup requires interaction with internal Microsoft systems). cc @ChumpChief for thoughts on that part since he refactored things recently-ish in that space.
> 
> I don't think this is much different than requiring a script to get an old tenant to target ODSP, which was the case before. Please correct me if I'm wrong though.

That side of things is probably fine, afaict this is mostly just trading one MSFT-only approach for another as you say.  However please make sure the errors and documentation are sufficient for MSFT devs to figure this out - based on my read here, I don't know where the appropriate package is or how to set it up.

examples/utils/webpack-fluid-loader/README.md
```diff
@@ -101,13 +101,9 @@ npm run start:r11s --env mode=r11s \
 
 To use a SharePoint server, run `npm run start:spo-df` if your OneDrive is on the DogFood server, and `npm run start:spo` if it is not.
 
-The following environment variables must be set:
+If you're a Microsoft developer, use the `@ff-internal/tenant-setup` package to populate this variable for `start:spo`:
 
--   local\_\_testing\_\_clientId
```
jzaffiro, line 106: @ChumpChief @alexvy86 is `getkeys` still in use? I'm pretty sure it's not and that the test tenant variables have been removed from the key vault but I'm not positive. If this is still a valid use case I'll revert this change.

ChumpChief: I'm not personally aware, but maybe check with @sonalivdeshpande about if there's anything important for OdspClient/SPE testing

examples/utils/example-webpack-integration/src/tenantUtils.ts
```diff
@@ -0,0 +1,131 @@
...
+
+let testTenantCheckoutClientCached: TestTenantCheckoutClient | undefined;
+
+export async function getTestTenantCheckoutClient(): Promise<TestTenantCheckoutClient> {
+	if (testTenantCheckoutClientCached !== undefined) {
+		return testTenantCheckoutClientCached;
+	}
+	// An internal package checks out test tenants, populates user information in the environment, and makes an entrypoint available
+	// at this location (token__package__import__location) which supports fetching tokens for those users.
+	const packageImportLocation = process.env.token__package__import__location;
```
ChumpChief: I'd strongly recommend moving all process.env access/validation up as early as possible in its respective flow (e.g. to the entry point of createOdspMiddlewares like the prior implementation did) and then disallow any other location to access env.  This lets you resolve bad-env failures and get concrete types early and in a single location that's easy to scan later.  Pass the resolved values like package import location through to these functions rather than letting them access the global environment.

It's especially valuable in this instance since this failure won't be hit until someone actually tries to get a token, whereas we can know much earlier that it won't work and fail faster.

examples/utils/webpack-fluid-loader/README.md
alexvy86: Ran getkeys just now and it still gave me `login__odsp__test__tenants` and `login__odspdf__test__tenants`, from the key vault. I don't know if they still work with webpack-fluid-loader though, I imagine not?

alexvy86 (approved):
I don't feel like I have *all* the context in my head to guarantee this change won't break anything, but it seems safe enough to me and in line with the direction we should be going anyway.

tools/pipelines/test-perf-benchmarks.yml
```diff
@@ -80,7 +80,9 @@ resources:
   - repository: ff_pipeline_host
     type: git
     name: ff_pipeline_host
-    ref: main
+    # Note: this is a test branch that is only used for staging purposes. Once the ff_pipeline_host main branch has the most recent
+    # package version, this change will be reverted.
+    ref: refs/heads/jzaffiro/updateTripsPkgVersion
```
ChumpChief: Is the plan to merge the change to ff_pipeline_host first?  I think this needs to be reverted before checkin, otherwise it complicates other possible maintenance needs on ff_pipeline_host.

examples/utils/example-webpack-integration/src/tenantUtils.ts
```diff
@@ -0,0 +1,131 @@
...
+	odspEndpointName: OdspEndpoint,
+	tenantIndex: number,
+): LoginCredentials[] {
+	const ficAccounts = process.env.login__odsp__fic__test__users;
+	if (ficAccounts === undefined) {
+		throw new Error(
+			"login__odsp__fic__test__users is not defined. Run the @ff-internal/tenant-setup script to populate this environment variable.",
+		);
+	}
+	const { usernames } = JSON.parse(ficAccounts) as {
```
ChumpChief: The prior implementation had type checking rather than casting, consider retaining a strong type-check here to cover scenarios besides just length === 0.

examples/utils/webpack-fluid-loader/README.md
jzaffiro: I would bet that they don't work - I think keyvault was populated with tenants that have since been phased out for security reasons. The tenants it gave you are probably well over a year old.

examples/utils/example-webpack-integration/src/tenantUtils.ts
jzaffiro: Fixed in the examples, but the env vars have to be read lazily during pipeline runs since the runtime exceeds the time for which a token is valid.

ChumpChief (approved):
LGTM pending the pipeline host change!

PR #27509, opened 2026-06-08
Update build-tools to pnpm 11

CraigMacomber:
## Description

Update build-tools to pnpm 11

## Reviewer Guidance


scripts/pack-packages.sh
```diff
@@ -19,15 +19,23 @@ if [ -f ".releaseGroup" ]; then
   if [ "$PACKAGE_MANAGER" == "pnpm" ]; then
     flub exec --no-private --concurrency=1 --releaseGroup $RELEASE_GROUP -- "pnpm --if-present pack:tests"
   fi
-  flub exec --no-private --concurrency=1 --releaseGroup $RELEASE_GROUP -- "$PACKAGE_MANAGER pack" && \
-  flub exec --no-private --concurrency=1 --releaseGroup $RELEASE_GROUP -- "mv -t $STAGING_PATH/pack/tarballs/ ./*.tgz" && \
+  flub exec --no-private --concurrency=1 --releaseGroup $RELEASE_GROUP -- "$PACKAGE_MANAGER pack"
+  flub exec --no-private --concurrency=1 --releaseGroup $RELEASE_GROUP -- "mv -t $STAGING_PATH/pack/tarballs/ ./*.tgz"
```
CraigMacomber, line 23: This use of && seemed harmful to me, as it obscured the error (this script did not fail). Claude agreed:

> These are separate statements (not chained with `&&`) so that `set -e` aborts the script on the first
> failure. Under `set -e`, a failing command that is not the last link of an `&&` chain is exempt from
> exit-on-error, so chaining would silently swallow a failed `pnpm pack` and let the script continue.

The && below is also removed for the same reason.

scripts/pack-packages.sh
```diff
@@ -19,15 +19,23 @@ if [ -f ".releaseGroup" ]; then
...
   flub exec --no-private --releaseGroup $RELEASE_GROUP -- "[ ! -f ./*test-files.tar ] || (echo 'test files found' && mv -t $STAGING_PATH/test-files/ ./*test-files.tar)"
 
+  # Clean up generated files that are listed in a package's "files" array (e.g. oclif.manifest.json) after packing.
+  # These are produced during build and intentionally shipped in the published tarball, but should not linger in
+  # the working tree. This cleanup deliberately runs here rather than in a package "postpack" script: under
+  # pnpm >=11, `pnpm pack` re-stats every "files" entry after postpack runs and fails with ENOENT if postpack
+  # deleted one. Running it here, after all packs complete, avoids that while still cleaning the working tree.
+  flub exec --no-private --releaseGroup $RELEASE_GROUP -- "pnpm run --if-present clean:manifest"
```
CraigMacomber, line 31: Due to the issue noted above, this was a bit nasty to root cause, but this fix seems reasonable.

I think pnpm 11 checking for odd changes during pack is good to help ensure packaged content reflects what it looks like was supposed to be in the package, which is nice for auditability/security against attacks which do strange things during publish (like the xz utils attack famously), so I don't think their policy change is a bug, just improved strictness which we violated and needed this change to satisfy.

tools/pipelines/templates/include-install-build-tools.yml
```diff
@@ -10,6 +10,8 @@ parameters:
 - name: buildDirectory
   type: string
 
+# The version of the build tools to install, or "repo" to install the one from the repository.
+# If "repo" is selected, this includes a global install of what ever version of pnpm build-tools uses.
```
CraigMacomber: This isn't new, but it caused some complications in this change as this now installs pnpm 11, which broke tests-tools by overriding its pnpm 10 install.

PR #27515, opened 2026-06-08
feat(tree): Add vTextExperimental codec with specialized node shape encoding

brrichards:
## Description

Wires up an encoder for the specialized node shape (`{f: {base, fields, value?}}`) introduced in #27200, behind a new experimental field-batch format version, `vTextExperimental: "text"`. The bulk of this diff is snapshot files (~3,950 lines across 54 snapshots); the actual code changes are ~1,520 lines across 9 source/test files.

The format is registered with `minVersionForCollab: undefined`, so it is never selected for writing unless explicitly requested via `writeVersionOverrides` (with `allowPossiblyIncompatibleWriteVersionOverrides`). New clients can always decode it. 

### What is a cohort?

Example:

```
CharacterFormat (object)
    bold:   Value -> leaf.boolean
    font:   Value -> leaf.string
    italic: Value -> leaf.boolean
    size:   Value -> leaf.number
```

All these fields are required (`kind: "Value"`) single-leaf fields, so all four are specializable. A cohort is the set of `CharacterFormat` instances in a batch that share identical values for the fields under consideration, e.g. `(false, "Arial", false, 12)`. When a cohort is worth folding, one specialized `f` shape bakes those values into the shape table and every member emits zero stream tokens for them.

Excluded from specialization: identifier-kind fields (they must keep the `SpecialField.Identifier` id-compressor encoding), fields the incremental-encoding policy routes out-of-band, and polymorphic fields.

### Two passes per field batch

Pass 1 — counting. Before any encoding, the batch is walked in full (every node at every depth), grouping each candidate node into a whole-tuple cohort and capturing the tuple's values on first sight. Incremental fields are skipped. Their sub-chunks get their own counting pass when encoded recursively, so counting them here would inflate the outer batch's totals. This pass produces no output.

Fold decision (lazy, once per node type per batch). Folding is decided by estimated byte gain. A cohort's gain is its member count × the inline bytes its pinned values would no longer occupy, minus the one-time cost of its shape-table entry. Folding also has a batch-wide cost: once a node type resolves to more than one shape, *every* instance pays a per-node dispatch token (`AnyShape`), so the summed gains must also beat that tax. A single cohort covering every instance stays monomorphic and pays no dispatch at all.

Pinning lets mixed data participate rather than all-or-nothing on the full tuple, the decision greedily selects a subset of fields to fold. Fully-unique fields are dropped up front, and while nothing folds profitably, the remaining field with the most distinct values is dropped and the decision retried. A record type with a unique `id` field can still fold its repetitive formatting fields.

Pass 2 — encoding. The normal `compressedEncode`, which consults the finalized decision to declare, per node type, either the base shape, a single specialized shape (monomorphic), or `AnyShape` dispatch with each instance resolving to its cohort's shape or the base.

### Batch scoping

`EncoderContext` gains `beginBatch`/`endBatch` hooks (no-ops for V1/V2). The VText context keeps a stack of per-batch specialization states, so the recursive `compressedEncode` calls made for incremental sub-chunks each get independently scoped counts and fold decisions. A sub-chunk's cohorts never leak into the outer batch's decisions or vice versa.

## Reviewer Guidance


brrichards:
> compressedEncode calls context.beginBatch(fieldBatch) (which pushes a VTextBatchState onto the VText batch stack) but does not use try/finally to guarantee context.endBatch() (which pops it) is called if an exception is thrown during encoding. If anyFieldEncoder.encodeField or updateShapesAndIdentifiersEncoding throws, the batch stack retains an extra element. For VText incremental sub-chunk encoding, compressedEncode is called recursively with the same EncoderContext instance (see incrementalFieldEncoder which passes context to the recursive call). If any implementation of IncrementalEncoder.encodeIncrementalField catches the exception thrown by its encode callback and then allows the outer compressedEncode to continue, currentBatch() will return the stale inner batch instead of the outer batch. All subsequent cohort lookups and specialization decisions in the remainder of the outer encode would use the wrong VTextBatchState, producing wrong shapes and potentially corrupted output.

Maybe I'm thinking about this wrong, but if an error happens during encoding I imagine that is a very serious error that we don't want to just have happen and then continue with a potential continuation of using the tree/data. I also think that an error is always thrown in the case of this happening. I don't think a try/finally is doing anything here other than just catching an error that would already propogate up and then updating our batch stack that we dont want to use anymore?

brrichards:
> cohortKey allocates a fresh string[] (const parts: string[] = []) on every invocation. It is called once per VTextObjectNodeEncoder node during the encoding pass (O(N)) and once per node per count-pass iteration (up to 10 × O(N)). For a large collaborative document with 1M rich-text nodes this produces ~10M short-lived array allocations per summary encode, generating significant GC pressure. For documents that are encoded frequently (e.g. on every container attach or summary upload under incremental-encode paths) the heap churn accumulates.

1 million rich text-nodes.... The up to 10 is just the cap of iterations, but for formatted text nodes it will converge in around 2-3, I believe v8 handles these short-lived array well enough that this shouldn't be bad on performance. Following this PR will be heavy performance testing/optimizations for this new codec. If this does end up being heavy then it can be optimized in the following PR's

packages/dds/tree/src/feature-libraries/chunked-forest/codec/format/versions.ts
```diff
@@ -33,7 +33,7 @@ export const FieldBatchFormatVersion = strictEnum("FieldBatchFormatVersion", {
 	/**
 	 * Experimental codec with optimizations for text.
 	 */
-	vTextExperimental: "text",
+	vTextExperimental: 3,
```
CraigMacomber: I think we want to hold off on doing this until we have confirmed this new format is accomplishing our goals, and we don't need/want to make any adjustments to it (once it gets a  number, we should never tweak the format and should support loading it forever)

packages/dds/tree/src/feature-libraries/chunked-forest/codec/compressedEncode.ts
```diff
@@ -540,8 +543,42 @@ export class EncoderContext implements NodeEncodeBuilder, FieldEncodeBuilder {
...
+		 * call (including the recursive sub-chunk calls made by {@link incrementalFieldEncoder}).
+		 * Lets an encoder policy set up state scoped to a single batch. The VText encoder
+		 * pushes a fresh per-batch specialization state and runs its counting pass. Paired with
+		 * {@link onEndBatch}.
+		 * @remarks
+		 * The per-batch state is owned and typed by the policy, not by this generic context;
+		 * keeping it off the encoder instances is what lets recursive sub-chunk encodes get their
+		 * own state with no snapshot/restore.
+		 */
+		private readonly onBeginBatch:
```
CraigMacomber: This pattern seems odd to me. Can you just run any before and after logic you need inside schemaCompressedEncodeVTextExperimental before and after doing the actual encode? Also, when 
schemaCompressedEncode is called, it has 
everything thats needed to run this when constructing the context, which is only used for that encode, so I don't get why we need to defer this using a callback.

You could even have it call compressedEncode directly so it could provide a customized or even subclassed EncoderContext.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +107,551 @@ export function schemaCompressedEncodeV2(
...
+}
+
+/**
+ * Test-only variant of {@link schemaCompressedEncodeVTextExperimental} that accepts a custom
+ * `minOccurrencesForSpecialization` threshold so small test inputs can exercise the
+ * specialization heuristic. Production callers must use
+ * {@link schemaCompressedEncodeVTextExperimental}, which hard-codes
+ * {@link defaultMinOccurrencesForSpecialization}.
+ */
+export function schemaCompressedEncodeVTextExperimentalForTests(
```
CraigMacomber: if this only exits for tests, it should be in the test code, not the production code files, unless applications using FF need it for writing their own tests (which I don't think is the case here as its not package exported)

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +107,551 @@ export function schemaCompressedEncodeV2(
...
+ * Encodes ObjectNodes using {@link SpecializedNodeShapeEncoder} shapes that constant-fold
+ * required single-valued fields whose contents are predictable across a batch.
+ *
+ * @remarks
+ * Uses a two-pass encode: pass 1 ({@link countNode}) counts each field-value tuple's
+ * occurrences, pass 2 ({@link encodeNode}) specializes tuples that cross
+ * {@link defaultMinOccurrencesForSpecialization}. See {@link SpecializableField} for
+ * the two kinds of field specialization supported.
+ */
+class VTextObjectNodeEncoder implements NodeEncoder {
```
CraigMacomber, line 552: I haven't figured out the actual algorithm you are using yet, but it sounds like it might be a bit expensive. You should probably have some performance tests to measure how slow it is compared to the existing approach, as well as some opt size tests to compare how much of a win this gives in various cases, and see if any of our existing test cases regress.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +107,551 @@ export function schemaCompressedEncodeV2(
...
+	public constructor(
+		private readonly base: NodeShapeBasedEncoder,
+		private readonly specializableFields: readonly SpecializableField[],
+		private readonly minOccurrencesForSpecialization: number,
+		private readonly nodeBuilder: NodeEncodeBuilder,
+		private readonly currentBatch: () => VTextBatchState,
+	) {}
+
+	public get shape(): AnyShape {
+		return AnyShape.instance;
```
CraigMacomber: This seems like the encoding is forcing an indirection, putting a reference to the actual shape you are specializing in the data array unconditionally. That seems like a lot of overhead, and something we should avoid.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +107,551 @@ export function schemaCompressedEncodeV2(
...
+				state.resolveCounts.set(key, merged);
+			}
+			state.counts = new Map();
+		}
+		return changed;
+	}
+}
+
+/**
+ * The per-encoder slice of {@link VTextBatchState}.
```
CraigMacomber: I'm finding this new code hard to read. I can't understand VTextBatchState without knowing what cohorts are, but CohortState is defined in terms of VTextBatchState.

Its the same kind of issue I called out in https://github.com/microsoft/FluidFramework/pull/27467/files/d75042feb21c542606a64903c494098914c757c0#r3345015961

Perhaps we can define a cohort. Then maybe define a cohort state in terms of that, then build up from there with the actual algorithm?

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +107,551 @@ export function schemaCompressedEncodeV2(
...
+ */
+interface CohortState {
+	/** Cohort key occurrence counts for the current count-pass iteration. */
+	counts: Map<string, number>;
+	/** Finalized counts from previous iterations, used for threshold decisions. */
+	resolveCounts: Map<string, number>;
+	/** Cached specialized shapes, keyed by cohort key. */
+	specializedEncoders: Map<string, SpecializedNodeShapeEncoder>;
```
CraigMacomber: This is hard to understand: its hard to tell how these are used because the types are not very specifcic.

Do these fields need to be reassignable? If not, make them readonly so its clear the maps are only modified not reassigned.

Nothing explicitly states what the keys are here: are these three parallel maps with the same keys?
If so, generally thats an anti patten as having a single map with one copy of each key and an inner object with the relevant properties often makes more sense (there are reasons for either approach, but right now all I can really say is more clearity is needed: if these are parakllel maps you need to document/clarify that, and say why, while using a single map is more self documenting)

CraigMacomber: If those strings are some special keying scheme, it might be good to have a type alias for that so you can document the syntax and semantics in one place which is easily found. Using a stronger branded typw could do the same bug with added type safty and might be a good here.

CraigMacomber:
I don't follow the encoding algorithm, but I don't see how this addresses the fact that for even a simple type with N leaf fields which are Booleans, there, there are 2^N different sets of fields you can specialize, and for each of those 2^N cases, there are  a ton of ways that specialization could be made out of other specialized shapes or not.

It sounds like you are counting occurrences of different shapes, but there are ~2^N possible shapes to count of every node with N children, plus the conditional count of the shapes those shapes could be made out of.

Anyway, the description in the PR and the comments I had time to go through in the code mostly left me wondering what a "cohort" is and how "cohort keys" work, and what approximate/heuristic this is intending to implement I think an optimal approach would be computationally intractable.

I could imagine computing a frequency table for every leaf value in the schema, then use some heuristics based on that to come up with shapes. Any field with exactly one value can be trivially inlined into the shape, but going beyond that requires knowing correlations between the fields, which gets messy and likely impractical.

You could decide that specializations for a given type always form a tree, where you specialize fields in the order:
1. Fields with exactly one value (specialized in root)
2. Fields sorted by how much data might be saved (for each value of the key, length of encoded value * number of occurrences - 1)

Then lazily populate that shape tree when encoding?

That's just an idea: not saying the approach you have is bad, or different, just that I'd expect something in the top level descript addressing high level algorithm concerns like how this decided which fields we should specialize and in what order, and I could get that from the existing description.

justus-camp-microsoft:
The "cohort" thing is something left-over from my handover branch I sent to Brennan and was one of those things where Claude decides on some terminology and just runs away with it. I meant to clean it up but never got the time to do it. I agree with what you're saying about this algorithm being expensive and that this is overcomplicating what could be a bit simpler. 

On your suggestions, I believe I started with the two cases you're describing and didn't see the kind of compression you were expecting without doing what this branch ended up as. Either way, simplifying it, adding some benchmarks that are focused on encoding cost instead of compression ratio, and maybe starting with something that has a little worse compression but is more reviewable/iteratable are all things I agree with.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: I restructured the comments to hopefully address this

packages/dds/tree/src/feature-libraries/chunked-forest/codec/compressedEncode.ts
brrichards: Went with your suggestion to do a customized encoderContext.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/format/versions.ts
brrichards: Ah, that makes sense. I've changed it back to "text"

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: This function was used so we could export it without haveing to import all the internals that are a part of this into the test file itself while still allowing us to modify the min occurrence for using the specialized shape. I've removed this function and updated the non test one to allow us to pass in a min occurrence. That function is now the exported one used in testing as well. (I like that better because we are testing the actual function now instead of the exported copy)

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: You're right and this was causing a small unnecessary of regression from V2. Since I've updated the algorithm used in the first pass to something simpler this shouldn't be a problem anymore

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: I updated the description to try and give a decent high level explanation of the algorithm. The current algorithm only does a single "counting" pass now instead of multiple which is still worse performance wise compared to v2, but theres no regression in the testing, and formatted text encoding sees 50-80% decrease in size (This is without a few of the optimization Justus was working on already)

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: I updated this to hopefully be more clear. With only a single count pass, it's now two maps.

The first map, `readonly counts: Map<SpecializationKey, number>`, has a key of `SpecializationKey` which is a string that is the stringified tuple of the specializable child fields of the current node (only the required, single-valued boolean/string/number leaf fields, not all children). The value is the count of nodes that share this identical key. That shared key is what defines a cohort: the set of nodes that all stringify to the same key.

Once an entry in the first map reaches the threshold (`minOccurrencesForSpecialization`, default 8, the cohort is `>= 8`), we add an entry to the second map under that same `SpecializationKey`, with the value being a `SpecializedNodeShapeEncoder` which emits the specialized (`'f'`) shape. So the second map is a sparse subset of the first: only cohorts that crossed the threshold appear in it.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/compressedEncode.ts
```diff
@@ -143,7 +146,11 @@ export function asFieldEncoder(encoder: NodeEncoder): FieldEncoder {
 		): void {
 			forEachNode(cursor, () => encoder.encodeNode(cursor, context, outputBuffer));
 		},
-		shape: encoder.shape,
+		// Read lazily: some node encoders (e.g. the VText cohort encoder) only know their
+		// shape once per-batch state has been resolved, after this wrapper is constructed.
+		get shape(): Shape {
+			return encoder.shape;
+		},
```
Josmithr, line 159: Presumably this is guaranteed to be resolved by the time it could be accessed?

packages/dds/tree/src/feature-libraries/chunked-forest/codec/codecs.ts
```diff
@@ -259,4 +268,14 @@ export const fieldBatchCodecBuilder = VersionDispatchingCodecBuilder.build("Fiel
 			EncodedFieldBatchV2,
 		),
 	},
+	{
+		minVersionForCollab: undefined,
```
Josmithr, line 368: I don't have a complete picture of the implications here, but should this be set to the current client version?

packages/dds/tree/src/feature-libraries/chunked-forest/codec/compressedEncode.ts
```diff
@@ -542,6 +560,28 @@ export class EncoderContext implements NodeEncodeBuilder, FieldEncodeBuilder {
...
+		// No per-batch state in the base encoder.
+	}
+
+	/**
+	 * Invoked when a {@link compressedEncode} call completes, to tear down per-batch state. The
+	 * teardown counterpart to {@link beginBatch}.
+	 */
+	public endBatch(): void {
+		// No per-batch state in the base encoder.
+	}
```
Josmithr, line 611: What are the implications if code calls `beginBatch` but fails to call `endBatch`?

CraigMacomber:
> All these fields are kind: "Value" pointing at a single leaf type. So all five are specializable fields.

This sounds like its expressing a limitation of our specialization format.

Our format can specialize non leaf values: it can specialize any field. It can even specialize a field which contains a large tree where some arbitrary subset of it is constant, which could includes arrays, maps, unions, etc.

Thus, instead I think this is a heuristic (or limitation) of the specialization algorithm.

> All these fields are kind: "Value" pointing at a single leaf type. So all five are specializable fields. A node's specialization key is the exact tuple of all those fields values, e.g. (false, "Arial", false, 12, etc...)
>
> So a cohort is all CharacterFormat instances in the batch who share identical tuples of their fields If a cohort is >= 8 then one specialized f shape bakes all the fields in the tuple into the shape table, and every node in that cohort emits zero stream tokens for those fields.

You are mixing the general definition with the example too much here, and the use of "etc..." in the example makes it extra confusing. You should a general definition which can be applied to things outside of the example. The optional example can go before or after the general definition to help with understanding how its applied, and I'd recommend an example which is complete and simple (no ...etc in it which may or may not contain non leaf fields, optional fields, unions of leaves which complicate things)

The only definition of cohort you give specific to CharacterFormat, and its unclear what has to be greater than 8 (the number of fields? I think in this case its the number of nodes in the field batch, but maybe it is (or should be) the number of nodes in this incremental chunk of the field batch?

Also if I have a schema:
```
ExampleText:
   Text(string)
   Bold(bool)
```

It sounds like this algorithm would never specialize just the "Bold" field, but always include the "Text" field as well, and only specialize if we have at least 8 instances with the same text?

To me this algorithm is starting to sound like a simple deduplication of identical subtrees. While that is useful, "EncodedSpecializedNodeShape" is much more general than that, and also not very efficient at that. If we just wanted to deduplicate identical subtrees:

1. I think you could explain that is a much clearer way. For example, "We check a subset (all trees whose children are all leaves) of sub-trees to see how many duplicates of them there are. If there are 8 or more, we generate and use a EncodedSpecializedNodeShape which bakes in the constant values."
2. EncodedSpecializedNodeShape is a bad way to encode this. An alternative EncodedFieldShape which is a shape plus data array would be much more compact.

Also general definitions about how an algorithm works, or what a concept used in the code is should generally be documented in the code not the PR description: you can refer to those definitions in the code from the PR description when helpful, and optionally summarize them, but the canonical detailed documentation should be in the code where future users can read it.

What goes in the PR description is content about the change which does not belong in the code itself. For example details about how this fixes problems with the code from before the change (details which are no longer relevant after the change is merged), details about how work is being staged, planned and/or prioritized, and enough contexts for reviewers before the change to help them select if they should review and know how to start reviewing: risk/scope is a great thing to include as part of that so reviewers know what to watch out for. Its also good to include enough details so that after its merged if there is a regression to help attribute or understand it (usually the risk/scope information should cover that).

Also you say "node's specialization key" and make no mention if the actual type `SpecializationKey`: its great you have that type now to centralize those docs: the PR description is one of the places that should leverage that by referring to it.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +111,566 @@ export function schemaCompressedEncodeV2(
...
+	assert(
+		valueType === "string" || valueType === "number" || valueType === "boolean",
+		"valueKey only supports primitive leaf values",
+	);
+	return `${valueType}:${value as string | number | boolean}`;
+}
+
+/**
+ * Identifies a "cohort": the set of nodes that share identical {@link SpecializableField} leaf
+ * values and so can encode through one {@link SpecializedNodeShapeEncoder}.
```
CraigMacomber: This definition isn't clear. I think it should require that the type of the nodes match, and we should be clear if it requires all fields be specialized, or just some. Also note that its possible to specialize a field without fully specializing it: this code might not take advantage of that but SpecializedNodeShapeEncoder can.


Also if the definition of "SpecializationKey" is "A key which identifies a cohort", then I think it should be named "CohortKey"

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +111,566 @@ export function schemaCompressedEncodeV2(
...
+	 * Build the {@link SpecializationKey} for the node at the cursor's current position from its
+	 * {@link SpecializableField} leaf values. See {@link SpecializationKey} for the key syntax.
+	 *
+	 * @remarks
+	 * Runs once per node in the count pass (and again per node in the encode pass for polymorphic
+	 * cohorts), so the key is built by concatenation rather than a `JSON.stringify` per node.
+	 */
+	private specializationKey(cursor: ITreeCursorSynchronous): SpecializationKey {
+		let key = "";
+		for (const field of this.specializableFields) {
```
CraigMacomber, line 782: Seems like the order of fields is important for this to be stable: maybe document the dependence on a stable field order, or sort them?

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +111,566 @@ export function schemaCompressedEncodeV2(
...
+			cursor.enterField(brand(field.key));
+			const hasNode = cursor.firstNode();
+			assert(hasNode, "required specializable field must contain a node");
```
CraigMacomber: Does this mean we cannot specialize over an optional field being missing?

If we do such a requirment, it might be better to use enterNode(0), which is only valid if the node exists and thus doesn't require the assert (though is undefined behavior which in some cases might not error if the node is missing): asserting the length is 1 might be better here anyway.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +111,566 @@ export function schemaCompressedEncodeV2(
...
+
+	/**
+	 * Build the {@link SpecializationKey} for the node at the cursor's current position from its
+	 * {@link SpecializableField} leaf values. See {@link SpecializationKey} for the key syntax.
+	 *
+	 * @remarks
+	 * Runs once per node in the count pass (and again per node in the encode pass for polymorphic
+	 * cohorts), so the key is built by concatenation rather than a `JSON.stringify` per node.
+	 */
+	private specializationKey(cursor: ITreeCursorSynchronous): SpecializationKey {
```
CraigMacomber: Making all these string keys can be expensive. You might want to consider keeping them in an object form.

I assume these are turned into strings to be usable in maps? If you can reuse the sub-objects efficiently, keeping them in object form might save significant memory, and could be used as keys in our BTreeMap type with a custom compare function. That's just something to consider: I'm not sure it would be a good idea here.

CraigMacomber:
> For each candidate node (an ObjectNode with required single-valued boolean/string/number leaf fields. See example above), the pass builds a specialization key: the tuple of that node's specializable leaf values. Nodes sharing a key form a cohort. The pass tallies how many times each cohort occurs.

I think it might make more sense to evaluate this on a per field level than a per node level. If a given field is "true" 50 times it might be worth specializing that field. If we only include leave values and `undefined` as possible to specialize for now, that avoids having to make string or aggregate keys for that step so it can likely be quite efficient.

Of course that doesn't solve the problem of how to construct the different specializations of multiple fields: that's harder.

There are also cases where specializing is a net loss, beyond just he case you called out from the overhead of creating the new specialized shape. For example if we have a node which wraps a boolean, creating a specialized true and a specalized false version of it is not useful as that replaces a boolean (which we can encode as 1 or 0) with a shape reference which is at best the same size, but likely larger.


All this complexity and the massive limitation of the current approach is making me think we might want to take a different approach. Might be better to discuss that in a meeting.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/codecs.ts
CraigMacomber: No. We do not want to stabilize/release this format. We want to avoid the stability/support comment until we are more confident its working well.

Josmithr: I see, so setting to `undefined` effectively means "unsupported"?

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: updated this so that the order of the fields no longer matter

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: updated the naming

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: yeah it only looks at required fields to specialize. I'll still add the assert

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +117,722 @@ export function schemaCompressedEncodeV2(
...
+
+/**
+ * Per-encoder bookkeeping for one batch. The count pass fills {@link CohortState.cohorts}; the fold
+ * {@link CohortState.decision} is then derived from it lazily, once, on first access.
+ */
+interface CohortState {
+	/** Every whole-node cohort observed in the batch, keyed by its all-fields {@link CohortKey}. */
+	readonly cohorts: Map<CohortKey, Cohort>;
+	/** The fold decision, computed lazily once on first access. */
+	decision?: FoldDecision;
```
CraigMacomber, line 504: I think this is a relativly poor design.

Generally you want to orginize your types so they make the correct use clear, and by construction, don't enable invalid use.

With this setup, nothing stops you from making more updates to cohorts after setting the FoldDecision.

Addationally this design needlessly couples thing more than necessary. There is no need for the counting phase to have types that depend on how the counts are "folded". THis unnecessary coupling could cause some real annoyances if you wanted to for example have a different folding approach later, or a rerefence folding to use in tests.

This approach also is less testable in isolated parts because of this coupling.

I think it would make much more sense to first collect data into some type. Then pass that type to some function which computes the decisions you need. This gives you two seperate operations you can test independently,
and makes it clear that the counting code can only count things (has no access to the FoldDecision), and allows the decision code to take in readonly counts making it clear it is a pure function producing output.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +117,722 @@ export function schemaCompressedEncodeV2(
...
+	count: number;
+	/** The pinned-field leaf values, in pinned-field order. */
+	readonly values: readonly Value[];
+}
+
+/**
+ * Per-encoder bookkeeping for one batch. The count pass fills {@link CohortState.cohorts}; the fold
+ * {@link CohortState.decision} is then derived from it lazily, once, on first access.
+ */
+interface CohortState {
```
CraigMacomber, line 500: I think this name is bad. This is count information for a batch, which internally happens to be agregreated by cohort. I feel like the name of this should be something like "BatchCounts". "CohortState" sounds like something that is per cohort.

I'd also rename `Cohort` to something matching, like `CohortCounts` so its clear it represents data about that cohort, and is not an actual encoding of all the data in the cohort.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +117,722 @@ export function schemaCompressedEncodeV2(
...
+	count: number;
+	/** Every specializable-field leaf value, in the encoder's (sorted) field order. */
+	readonly values: readonly Value[];
+}
+
+/**
+ * A cohort over the chosen pinned fields only: nodes that agree on every pinned field (their
+ * non-pinned fields may differ). This is the unit that actually folds into one specialized shape.
+ */
+interface PinnedCohort {
```
CraigMacomber, line 490: typesceript interfaces get structual typing, so a PinnedCohort and a Cohort are esentially the same and can be assigned to eachother since they have the same fields with the same types. I don't think this is intended.

Maybe rename the values field to avoid this collision (to "specalisedValues" maybe)? Of it a subtype relationship is desired.

Also if this one is only made later, perhaps its count should be readonly?

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +117,722 @@ export function schemaCompressedEncodeV2(
...
+ * decision can choose which fields to pin and build the shapes without re-walking a cursor).
+ */
+interface Cohort {
+	count: number;
+	/** Every specializable-field leaf value, in the encoder's (sorted) field order. */
+	readonly values: readonly Value[];
+}
+
+/**
+ * A cohort over the chosen pinned fields only: nodes that agree on every pinned field (their
```
CraigMacomber, line 487: We have a lot of terms here (cohort, pinned, specialized, fold) which are not really defined.

I don't think "cohort" here refers to the "Cohort" type, so I think the Cohort type should be renamed to CohortCounts or something (as I noted elsewhere) so that we avoid the confusion. That however just fixes that its easy to get the wrong definition for chort, we still need somewhere to provide the correct one, which the various types related to it should link.

Perhaps we can avoid introducing the term "pinned" here as well and just use sepcalized since that term alread is in use here with a relevent meaning.

I'm not clear on the use of "fold" but I don't think its https://en.wikipedia.org/wiki/Fold_(higher-order_function): If you had a function which took in counts and returned encoding decisions, what ever its called, linking it here could help.

Also this type doesn't make a lot of sense on its own, as it doesn't say which fields are pinned/specalized or in what order. I think this must only be useful inside of some computation which also has that data: I'd suggest at least including a remark pointing out what uses this and why, or move this to be more local to it.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -94,6 +117,722 @@ export function schemaCompressedEncodeV2(
...
+	}
+
+	/**
+	 * Folds the worthwhile cohorts and returns them with the declared shape (or an empty set + the
+	 * base shape if none is worth it). Folding makes the type polymorphic, so every instance then
+	 * pays a per-instance dispatch token; the decision weighs saved bytes against shape cost and that
+	 * tax. A single cohort covering every instance stays monomorphic. Many low-value cohorts can't
+	 * beat the tax, so nothing folds.
+	 */
+	private foldPinnedCohorts(
```
CraigMacomber, line 680: Generally its good to factor code with complex flow control out into unit testable pure functions working with value types. Its much easier to understand, modify and test that way.

This code looks quite complex with lots of conditional flow control. Having it tried up in a priuvate class method for a class that does lots of other things seems like a bad pattern.

In this case Id expect something like:

1. A "collectCounts" function with takes in a batch, and returns an immutable counts object in some format specifi to the cohort based stratagy.
2. A "selectEncodingStratagies" function which takes in the immutable counts object, and returns some object which describes how to encode anything that could occure in that batch in a transparent value type style way specific to the cohort based stratagy.
3. A function which takes in the cohort based decisions, and abstracts that into a general prupose set of encoding decisions. Likley this can use our existing types. Maybe a field encoder which can handle all the roots in the batch?
4. A "encodeBatch" function which takes in a batch and the object from above and returns an encoded batch.

All 4 can be pure functions which we can test indepependently. This also makes it easy to test/add alternative versions of some or all of them to try out different approaches.
I think the complex login in here would end up in step 2 using the above archetecture.

CraigMacomber: I have some notes on value types with references in this old page on my site http://www.craigm.info/thoughts:design . The boundaries talk that it links ( https://www.destroyallsoftware.com/talks/boundaries ) is particularly relevant, and I feel quite excellent.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: I'm going to rename things and also give clearer definitions to terms being used. I think my comments needed a lot of work in this PR and hopefully with the coming round of commits everthing should be structured and defined clearly. I've been using cohort, but I don't really like that name and changed it to nodeGroup which is essentially the group of nodes that share specialized fields

packages/dds/tree/src/feature-libraries/chunked-forest/codec/compressedEncode.ts
brrichards: Yes. Only the encode pass reads shape. The encode pass runs after beginBatch. beginBatch runs the count pass and pushes the batch state first. If code reads shape outside a batch, VTextEncoderContext.currentBatch() asserts. It should not return an old shape.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/compressedEncode.ts
brrichards: With some of the other changes I did i was able to drop the stack that this batch was one so beginBatch and endBatch were removed

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: This will be fixed with the renaming I'm doing.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: You're right about the design parts. I've gone back and changed it so

1. collectBatchCounts does the count pass. It returns read-only BatchCounts. It only counts. It cannot access a decision.
2. decideBatchSpecializations gives the counts of each encoder to chooseSpecialization. It returns read-only BatchSpecializations.
3. The encode pass reads only the decisions.

Some of the methods up top may look different since I've renamed fold to specializations, since you mentioned in another comment that folding isn't really what's happening here. 

also the batch state does not keep the counts after step 2 so code can no longer change counts after the decision. chooseSpecialization is now exported as well. New unit tests call it directly with counts that the test makes.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: Changing to using the BTreeMap like you suggested. Will write a small benchmark test to verify memory usage of strings vs objects

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: renamed to avoid collision

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
brrichards: I've gone ahead and redesigned things to align with your suggestions. 

Now `collectBatchCounts` takes a batch and returns read-only `BatchCounts`. 

Then `decideBatchSpecializations` changes the counts into `BatchSpecializations`. For each node type it calls `chooseSpecialization`. This is now its own function that is unit tested as well. 

Then, `buildBatchContext` changes the decisions into a custom `EncoderContext`. That deals with the new specialized shape, base shape, and handles a polymorphic type. This step also uses the existing encoder types.

Finally, `compressedEncode` encodes the batch with that context. 

Because of how context is now being handled I was able to remove the batch stack that I previously had.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/compressedEncode.ts
```diff
@@ -475,7 +475,7 @@ export const incrementalFieldEncoder: FieldEncoder = {
 
 		const chunkReferenceIds = context.incrementalEncoder.encodeIncrementalField(
 			cursor,
-			(chunk: TreeChunk) => compressedEncode([chunk.cursor()], context),
+			(chunk: TreeChunk) => context.encodeIncrementalChunk([chunk.cursor()]),
```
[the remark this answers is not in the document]
brrichards: The type was not able to be identified, but the incremental chunks were still supported by text. Added a union of version 2 and version text so that it can now be typed correctly. version 1 doesnt have inc summary so I don't think it need to be in the union

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -254,3 +274,846 @@ function valueShapeFromSchema(schema: ValueSchema | undefined): undefined | Enco
...
+			options.idCompressor,
+			options.incrementalEncoder,
+			brand(FieldBatchFormatVersion.vTextExperimental),
+			options.isSummary,
+		);
+	}
+
+	public override encodeIncrementalChunk(fieldBatch: FieldBatch): EncodedFieldBatchV1OrV2 {
+		return encodeBatchVText(fieldBatch, this.options);
+	}
```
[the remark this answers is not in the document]
brrichards: I believe this is related to the comment above and the fix for that will address this as well

packages/dds/tree/src/feature-libraries/chunked-forest/codec/codecs.ts
```diff
@@ -331,4 +341,14 @@ export const fieldBatchCodecBuilder = VersionDispatchingCodecBuilder.build("Fiel
...
+		formatVersion: FieldBatchFormatVersion.vTextExperimental,
+		codec: makeFieldBatchCodecForVersion(
+			FieldBatchFormatVersion.vTextExperimental,
+			uncompressedEncodeVTextExperimental,
+			schemaCompressedEncodeVTextExperimental,
+			EncodedFieldBatchVTextExperimental,
```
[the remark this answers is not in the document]
brrichards: This is the same typing problem as the first previous review, I'll update the union to be all of them and then apply the broader fix.

packages/dds/tree/src/feature-libraries/chunked-forest/codec/schemaBasedEncode.ts
```diff
@@ -254,3 +275,851 @@ function valueShapeFromSchema(schema: ValueSchema | undefined): undefined | Enco
...
+ * An incremental sub-chunk does the four steps again, with its own counts. See
+ * {@link EncoderContext.encodeIncrementalChunk}.
+ */
+function encodeBatchVText(
+	fieldBatch: FieldBatch,
+	options: VTextEncodeOptions,
+): EncodedFieldBatchV1OrV2 {
+	const counts = collectBatchCounts(fieldBatch, options);
+	const decisions = decideBatchSpecializations(counts, options.specializableFieldsOf);
+	return compressedEncode(fieldBatch, buildBatchContext(decisions, options));
```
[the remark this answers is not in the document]
brrichards: related to the other comments in this review and should be solved with recent round of fixes

packages/dds/tree/src/feature-libraries/chunked-forest/codec/uncompressedEncode.ts
```diff
@@ -38,6 +39,16 @@ export function uncompressedEncodeV2(batch: FieldBatch): EncodedFieldBatchV2 {
...
 }
 
+/**
+ * Encode data from `cursor` in the simplest way supported by `EncodedChunk` using {@link FieldBatchFormatVersion.vTextExperimental}.
+ * @remarks See {@link uncompressedEncode} for more details.
+ */
+export function uncompressedEncodeVTextExperimental(
+	batch: FieldBatch,
+): EncodedFieldBatchVTextExperimental {
+	return uncompressedEncode(batch, brand(FieldBatchFormatVersion.vTextExperimental));
```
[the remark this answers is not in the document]
brrichards: see above comment

packages/dds/tree/src/test/feature-libraries/chunked-forest/codec/compressedEncode.spec.ts
```diff
@@ -114,10 +112,30 @@ function makeFieldBatchCodec(
...
+	const codec = {
+		encode: (
+			data: FieldBatch,
+			context: FieldBatchEncodingContext,
+		): EncodedFieldBatchV1OrV2 => {
+			return compressedEncode(data, encoderContext);
+		},
+		decode: (
+			data: EncodedFieldBatchV1OrV2,
+			fieldBatchContext: FieldBatchDecodingContext,
```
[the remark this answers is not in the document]
brrichards: Currently copilot is picking up on wrong typing during any form of encode/decode. It's correctly getting supported from the experimental codec it's just being typed as v2 and not . The final fix for this was putting a never f shape on v2 similar to e on v1. This shows these errors during compile now. I've updated the remaining places and hopefully the wrong typing will be a non-issue.

PR #25120, opened 2025-07-30
(compat) Move Driver layer compat validation to Container creation time

agarwal-navin:
#24462 and #24855 added layer compat validation logic across the Loader / Driver layer. In the driver layer, it added the `ILayerCompatDetails` on document service and the Loader layer validates that when Container creates a document service.
However, we should be able to detect layer incompatibility much sooner and fail fast. This PR does that by moving the `ILayerCompatDetails` from document service to document service factory. The validation in Loader layer now happens in the Container constructor.

[AB#45406](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/45406)

PR #27516, opened 2026-06-08
feat(driver): validate Loader compatibility with Driver across the boundary

agarwal-navin:
## Description

This adds **bidirectional** layer-compatibility validation across the Driver / Loader boundary.

Previously the Loader only validated that the Driver was compatible with it (`validateDriverCompatibility`). The reverse — that the Loader is compatible with the Driver — was never checked, because the Driver holds no reference to the Loader and so cannot validate it.

To close that gap without inverting the layer hierarchy, the Driver now **publishes the requirements the Loader must meet** via a new `ILayerCompatSupportRequirements` `FluidObject` exposed on each driver's `IDocumentServiceFactory`. The Loader reads those requirements and validates itself against them **on the Driver's behalf**, via the new `validateLoaderCompatibilityWithDriver` function. Both directions now run from the `Container` constructor.

## The gap this addresses

The Driver does not hold reference to Loader and hence cannot check what features it supports or what it's generation is. So, if it wants to make a breaking change (say change in behavior of one of the APIs), it cannot make that change conditionally based on whether the Loader is compatible with the new behavior. Instead, it must ship the change dark, wait for saturation and then enable the new behavior. Say it does this in version N.

The gap this leaves is that if there is a combination of new Driver (post enabling new behavior) and old Loader (pre version N), the Loader will crash at the failure point of the new behavior which can be misleading and is exactly the kind of problems the layer compat checks are supposed to solve.

With this change, in the above scenario, with new Driver (post enabling new behavior) and old Loader (pre version N), the old Loader will fail the compatibility check with a clear error.

> Note that we still cannot conditionaly turn the new behavior in the Driver on based on Loader version. That gap cannot be closed until we pass in a Loader object to the Driver.

### Loaders before this change

In the above scenarios, if the old Loader is pre this change, it will crash at the failure point of the new behavior. This is not ideal but there is not much we can do about this. Moreover, this is fine because the new policy has been documented and communicated to customer way in advance.

## Also included in this change

- **Bug fix:** the Driver-boundary check was reusing `loaderCompatDetailsForRuntime`; it now uses a dedicated `loaderCompatDetailsForDriver`.
- A separate `IProvideLayerCompatSupportRequirements` provider (rather than embedding requirements in `ILayerCompatDetails`), keeping "what I am" and "what I require of you" distinct.
- Reverse-direction unit tests in `container-loader`, and the reverse (`driver -> loader`) direction added to the end-to-end `layerCompat` suite (runs per driver type across both create and load flows).

The new `ILayerCompatSupportRequirements` property on `LocalDocumentServiceFactory` and `OdspDocumentServiceFactoryCore` mirrors the existing `ILayerCompatDetails` property (added in #25120). A changeset is included.

[AB#58900](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/58900)

### DIFF
--- a/.changeset/few-nights-juggle.md
+++ b/.changeset/few-nights-juggle.md
@@ -1,0 +1,10 @@
+---
+"@fluidframework/local-driver": minor
+"@fluidframework/odsp-driver": minor
+"__section": feature
+---
+New ILayerCompatSupportRequirements property on LocalDocumentServiceFactory and OdspDocumentServiceFactoryCore
+
+A new optional property, `ILayerCompatSupportRequirements`, has been added to `LocalDocumentServiceFactory` and `OdspDocumentServiceFactoryCore`.
+
+The Driver layer uses this property to publish the requirements that the Loader layer must meet to be compatible with it. Because the Driver has no reference to the Loader, it cannot validate the Loader directly; instead the Loader reads these requirements and validates itself against them on the Driver's behalf. This enables the Loader / Driver compatibility check to run in both directions.
--- a/.changeset/few-nights-juggle.md
+++ b/.changeset/few-nights-juggle.md
@@ -5,6 +5,6 @@
 ---
 New ILayerCompatSupportRequirements property on LocalDocumentServiceFactory and OdspDocumentServiceFactoryCore
 
-A new optional property, `ILayerCompatSupportRequirements`, has been added to `LocalDocumentServiceFactory` and `OdspDocumentServiceFactoryCore`.
+A new optional property, `ILayerCompatSupportRequirements`, has been added to [`LocalDocumentServiceFactory`](https://fluidframework.com/docs/api/local-driver/localdocumentservicefactory-class) and [`OdspDocumentServiceFactoryCore`](https://fluidframework.com/docs/api/odsp-driver/odspdocumentservicefactorycore-class).
 
 The Driver layer uses this property to publish the requirements that the Loader layer must meet to be compatible with it. Because the Driver has no reference to the Loader, it cannot validate the Loader directly; instead the Loader reads these requirements and validates itself against them on the Driver's behalf. This enables the Loader / Driver compatibility check to run in both directions.
--- a/LayerCompatibility.md
+++ b/LayerCompatibility.md
@@ -43,6 +43,7 @@ flowchart
 
     %% Validation relationships
     Loader -.->|validates| Driver
+    Driver -.->|validates| Loader
     Loader -.->|validates| Runtime
     Runtime -.->|validates| Loader
     Runtime -.->|validates| DataStore
@@ -58,9 +59,11 @@ flowchart
 
 ### Loader ↔ Driver Boundary
 
-- **Validator:** Loader validates Driver
+- **Validator:** Both directions
+    - Loader validates Driver during container creation
+    - Driver validates Loader during container creation
 - **When:** During container creation (before connection established)
-> Note: Driver doesn't validate Loader because there are no interacts from Driver to Loader. So, there isn't a need to maintain compatibility in that direction.
+> Note: Unlike the other boundaries, both validations are performed by the Loader. The Driver has no reference to the Loader, so it cannot run the validation itself. Instead, the Driver publishes the requirements the Loader must meet (via `ILayerCompatSupportRequirements`, exposed on its `IDocumentServiceFactory`), and the Loader validates itself against them on the Driver's behalf.
 
 ### Loader ↔ Runtime Boundary
 
@@ -87,6 +90,8 @@ The incompatibilities between the layers are detected as follows:
 - **Loader ↔ Driver** and **Loader ↔ Runtime**: Incompatibility is detected during container create / load. If incompatibility is detected, a `FluidErrorTypes.layerIncompatibilityError` error is thrown, failing the container creation / load.
 - **Runtime ↔ DataStore**: Incompatibility is detected during data store create / load. If incompatibility is detected, a `FluidErrorTypes.layerIncompatibilityError` error is thrown, failing the data store creation / load.
   If the data store is being created / loaded during container create / load, that will also throw a `FluidErrorTypes.layerIncompatibilityError` error, failing the container creation / load.
+
+> Note: Even with these compatibility policies and windows in place, there may be clients running layers older than the versions in which compatibility validation was introduced. Such a layer does not provide compatibility details, so validation treats it leniently and does not fail it on that basis, preserving backward compatibility. If that old layer is paired with a newer layer in a combination that is in fact incompatible — for example, where a layer compat feature has since been enabled by default and the corresponding back-compat code removed — the interaction may still fail. In that case it does not surface as a `FluidErrorTypes.layerIncompatibilityError`; instead it fails in that feature's own failure mode (whatever error the specific feature surfaces).
 
 #### Telemetry
 
@@ -139,6 +144,7 @@ These support windows are enforced through the compatibility validation system d
 - Stream ops from ordering service
 - Implement service-specific protocols
 
+**Validates:** Loader layer (performed by the Loader on the Driver's behalf)
 **Validated By:** Loader layer
 
 ### 2. Loader Layer
@@ -158,7 +164,7 @@ These support windows are enforced through the compatibility validation system d
 - Coordinate between Driver and Runtime layers
 
 **Validates:** Driver and Runtime layers
-**Validated By:** Runtime layer
+**Validated By:** Runtime and Driver layers
 
 ### 3. Runtime Layer
 
--- a/LayerCompatibility.md
+++ b/LayerCompatibility.md
@@ -63,7 +63,10 @@ flowchart
     - Loader validates Driver during container creation
     - Driver validates Loader during container creation
 - **When:** During container creation (before connection established)
-> Note: Unlike the other boundaries, both validations are performed by the Loader. The Driver has no reference to the Loader, so it cannot run the validation itself. Instead, the Driver publishes the requirements the Loader must meet (via `ILayerCompatSupportRequirements`, exposed on its `IDocumentServiceFactory`), and the Loader validates itself against them on the Driver's behalf.
+
+> Note: Unlike the other boundaries, both validations are performed by the Loader.
+> The Driver has no reference to the Loader, so it cannot run the validation itself.
+> Instead, the Driver publishes the requirements the Loader must meet (via `ILayerCompatSupportRequirements`, exposed on its `IDocumentServiceFactory`), and the Loader validates itself against them on the Driver's behalf.
 
 ### Loader ↔ Runtime Boundary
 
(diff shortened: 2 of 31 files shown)


markfields:
> The Loader reads those requirements and validates itself against them on the Driver's behalf

Can we write out the implications of the fact that for years, old loaders (before this change) will not be validating this compatibility?  Related - would be helpful to explicitly state in the PR description and existing layer compat MD files why we want old loaders to give up if the driver is too much newer.  You mention that drivers don't hold a reference to the Loader, so explaining that even so, this matters, would be good.

packages/loader/container-loader/src/loaderLayerCompatState.ts
```diff
@@ -117,10 +130,48 @@ export function validateDriverCompatibility(
 	validateLayerCompatibility(
 		"loader",
 		"driver",
-		loaderCompatDetailsForRuntime,
+		loaderCompatDetailsForDriver,
```
markfields: Any hard thinking we need to do on back compat here?  Like, for loaders before this fix, the driver compatibility check is basically going to arbitrarily be right or wrong.  We should probably enumerate (for our own sake) the set of loader-driver-runtime version combos that will be false positives or false negatives for this validation.

packages/loader/container-loader/src/loaderLayerCompatState.ts
```diff
@@ -117,10 +130,48 @@ export function validateDriverCompatibility(
...
+	disposeFn: (error?: ICriticalContainerError) => void,
+	mc: MonitoringContext,
+): void {
+	if (maybeDriverCompatRequirements === undefined) {
+		return;
+	}
+	validateLayerCompatibility(
+		"driver",
+		"loader",
+		maybeDriverCompatDetails ?? defaultLayerCompatDetails,
```
markfields: Side note: This always surprises me seeing "layer 1"'s compat details passed in.

Maybe would be more self-documenting if its signature were something like this:

```
function validateLayerCompatibility(
	layer1: {
		layer: FluidLayer;
		versionInfo: Pick<ILayerCompatDetails, "pkgVersion" | "generation">;
		compatSupportRequirements: ILayerCompatSupportRequirements;
	},
	layer2: {
		layer: FluidLayer;
		compatDetails: ILayerCompatDetails | undefined;
		strictCompatibilityCheck?: boolean;
	},
	context: {
		disposeFn: (error?: IErrorBase) => void;
		mc: MonitoringContext;
	},
): void
```

packages/common/client-utils/src/layerCompat.ts
```diff
@@ -119,11 +119,35 @@ export const defaultLayerCompatDetails: ILayerCompatDetails = {
...
+	readonly ILayerCompatSupportRequirements: ILayerCompatSupportRequirements;
+}
+
 /**
  * The requirements that a layer needs another layer to support for them to be compatible.
+ *
+ * @remarks
+ * A layer can also publish its requirements across a boundary as a `FluidObject` (via
+ * {@link IProvideLayerCompatSupportRequirements}) for the case where the layer on the other side of the boundary
+ * cannot run the compatibility validation itself. Most boundaries are validated by both layers because each holds
```
markfields: Sounds backwards.  Shouldn't this say "for the case where this layer cannot validate the other layer itself, and needs the other layer to check itself against the published requirements" or something?

packages/drivers/local-driver/src/localLayerCompatState.ts
```diff
@@ -25,3 +30,26 @@ export const localDriverCompatDetailsForLoader: ILayerCompatDetails = {
...
+export const localDriverCompatRequirementsForLoader: ILayerCompatSupportRequirements = {
+	/**
+	 * Minimum generation that Loader must be at to be compatible with this Driver. This is calculated based on the
+	 * LayerCompatibilityPolicyWindowMonths.DriverLoader value which defines how many months old can the Loader layer
+	 * be compared to the Driver layer for them to still be considered compatible.
+	 * The minimum valid generation value is 0.
+	 */
+	minSupportedGeneration: Math.max(
+		0,
+		generation - LayerCompatibilityPolicyWindowMonths.DriverLoader,
```
markfields, line 49: Follow-up:  I'd name these more explicitly, like forNewDriverOldLoader or something

markfields (approved):

packages/loader/container-loader/src/loaderLayerCompatState.ts
agarwal-navin: Good suggestion. Applied it.

packages/common/client-utils/src/layerCompat.ts
agarwal-navin: You are right. Fixed it.

packages/loader/container-loader/src/loaderLayerCompatState.ts
agarwal-navin: I don't think we need to do anything here. the `loaderCompatDetailsForRuntime` was empty and so is the new `loaderCompatDetailsForRuntime`. They are identical so we got lucky. If `loaderCompatDetailsForRuntime` had some supported features then we would have to think about back-compat.

agarwal-navin:
> > The Loader reads those requirements and validates itself against them on the Driver's behalf
> 
> Can we write out the implications of the fact that for years, old loaders (before this change) will not be validating this compatibility? Related - would be helpful to explicitly state in the PR description and existing layer compat MD files why we want old loaders to give up if the driver is too much newer. You mention that drivers don't hold a reference to the Loader, so explaining that even so, this matters, would be good.

I updated the PR description with a new section explaining this. Can you check and see if that answers your questions. Also, do you think any of this should go in the official documentation?

packages/drivers/local-driver/src/localLayerCompatState.ts
markfields: Thumbs up meaning you'll consider it in a follow-up PR?

packages/loader/container-loader/src/loaderLayerCompatState.ts
```diff
@@ -95,13 +108,19 @@ export function validateRuntimeCompatibility(
...
-		"runtime",
-		loaderCompatDetailsForRuntime,
-		runtimeSupportRequirementsForLoader,
-		maybeRuntimeCompatDetails,
-		() => {} /* disposeFn - no op. This will be handled by the caller */,
-		mc,
+		/* validatingLayer */ {
+			layer: "loader",
+			compatDetails: loaderCompatDetailsForRuntime,
+			compatSupportRequirements: runtimeSupportRequirementsForLoader,
```
markfields, line 114: Nit: `compatDetails` reads a little strangely here. As far as I can tell, we only need the validating layer's package information, so could we rename this property to reflect that it's layer info rather than compatibility details?

Also, since this validation runs in the opposite direction—the runtime is proving compatibility with the loader—could we pass `loaderCoreCompatDetails` here instead of `loaderCompatDetailsForRuntime`?

markfields (approved):

packages/drivers/local-driver/src/localLayerCompatState.ts
agarwal-navin: yes

packages/loader/container-loader/src/loaderLayerCompatState.ts
agarwal-navin: Good catch. FIxed it.

.changeset/few-nights-juggle.md
```diff
@@ -0,0 +1,10 @@
+---
+"@fluidframework/local-driver": minor
+"@fluidframework/odsp-driver": minor
+"__section": feature
+---
+New ILayerCompatSupportRequirements property on LocalDocumentServiceFactory and OdspDocumentServiceFactoryCore
+
+A new optional property, `ILayerCompatSupportRequirements`, has been added to `LocalDocumentServiceFactory` and `OdspDocumentServiceFactoryCore`.
```
Josmithr: Nit: link to API docs?

packages/common/client-utils/src/layerCompat.ts
```diff
@@ -119,11 +119,36 @@ export const defaultLayerCompatDetails: ILayerCompatDetails = {
...
 };
 
+/**
+ * @internal
+ */
+export const ILayerCompatSupportRequirements: keyof IProvideLayerCompatSupportRequirements =
+	"ILayerCompatSupportRequirements";
+
+/**
+ * @internal
```
Josmithr, line 132: Docs please 🙂

(I'm surprised the linter didn't complain about this)

LayerCompatibility.md
```diff
@@ -58,9 +59,11 @@ flowchart
 
 ### Loader ↔ Driver Boundary
 
-- **Validator:** Loader validates Driver
+- **Validator:** Both directions
+    - Loader validates Driver during container creation
+    - Driver validates Loader during container creation
 - **When:** During container creation (before connection established)
-> Note: Driver doesn't validate Loader because there are no interacts from Driver to Loader. So, there isn't a need to maintain compatibility in that direction.
+> Note: Unlike the other boundaries, both validations are performed by the Loader. The Driver has no reference to the Loader, so it cannot run the validation itself. Instead, the Driver publishes the requirements the Loader must meet (via `ILayerCompatSupportRequirements`, exposed on its `IDocumentServiceFactory`), and the Loader validates itself against them on the Driver's behalf.
```
Josmithr: Nit (per our markdown best practices):
```suggestion

> Note: Unlike the other boundaries, both validations are performed by the Loader.
> The Driver has no reference to the Loader, so it cannot run the validation itself.
> Instead, the Driver publishes the requirements the Loader must meet (via `ILayerCompatSupportRequirements`, exposed on its `IDocumentServiceFactory`), and the Loader validates itself against them on the Driver's behalf.
```

LayerCompatibility.md
```diff
@@ -88,6 +91,8 @@ The incompatibilities between the layers are detected as follows:
 - **Runtime ↔ DataStore**: Incompatibility is detected during data store create / load. If incompatibility is detected, a `FluidErrorTypes.layerIncompatibilityError` error is thrown, failing the data store creation / load.
   If the data store is being created / loaded during container create / load, that will also throw a `FluidErrorTypes.layerIncompatibilityError` error, failing the container creation / load.
 
+> Note: Even with these compatibility policies and windows in place, there may be clients running layers older than the versions in which compatibility validation was introduced. Such a layer does not provide compatibility details, so validation treats it leniently and does not fail it on that basis, preserving backward compatibility. If that old layer is paired with a newer layer in a combination that is in fact incompatible — for example, where a layer compat feature has since been enabled by default and the corresponding back-compat code removed — the interaction may still fail. In that case it does not surface as a `FluidErrorTypes.layerIncompatibilityError`; instead it fails in that feature's own failure mode (whatever error the specific feature surfaces).
```
Josmithr: Nit (per our Markdown best practices):
```suggestion
> Note: Even with these compatibility policies and windows in place, there may be clients running layers older than the versions in which compatibility validation was introduced.
> Such a layer does not provide compatibility details, so validation treats it leniently and does not fail it on that basis, preserving backward compatibility.
> If that old layer is paired with a newer layer in a combination that is in fact incompatible — for example, where a layer compat feature has since been enabled by default and the corresponding back-compat code removed — the interaction may still fail.
> In that case it does not surface as a `FluidErrorTypes.layerIncompatibilityError`; instead it fails in that feature's own failure mode (whatever error the specific feature surfaces).
```

Josmithr (approved):

PR #27520, opened 2026-06-09
fix(tree): prevent revert of commits that were rolled back

yann-achard-MS:
## Description

A commit can be rolled back through `SharedTreeCore.rollback`. Once that is done, the revertible for that commit becomes invalid. Before this PR, we allowed applications to attempt a revert, which could either crash or perform undesirable edits. This PR ensures that the relevant revertible is disposed.

## Breaking Changes

None


packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
```diff
@@ -1899,6 +1899,27 @@ describe("sharedTreeView", () => {
...
+
+		const { undoStack, unsubscribe } = createTestUndoRedoStacks(view.events);
+		view.root.insertAtEnd({ id: "A" });
+		view.root[0].id = "B";
+		assert.equal(undoStack.length, 2);
+
+		view.checkout.mainBranch.removeAfter(init);
+		assert.equal(view.root.length, 0);
+		assert.equal(undoStack.length, 0);
+		unsubscribe();
```
jason-ha: nit: It is good to separate lines into Setup, Act, Verify, and Cleanup, when they exist. I see lines 1904 thru 1916 as setup. 1918 (`view.checkout.mainBranch.removeAfter(init)`) is the Act(ion). And 1920 (`assert.equal(undoStack.length, 0)`) appears to be Verification of the "spec".
1919 (`assert.equal(view.root.length, 0)`) is a consistency check?
I don't know about 1921 `unsubscribe()`.

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
```diff
@@ -1899,6 +1899,27 @@ describe("sharedTreeView", () => {
...
+		const init = view.checkout.mainBranch.getHead();
+
+		const { undoStack, unsubscribe } = createTestUndoRedoStacks(view.events);
+		view.root.insertAtEnd({ id: "A" });
+		view.root[0].id = "B";
+		assert.equal(undoStack.length, 2);
+
+		view.checkout.mainBranch.removeAfter(init);
+		assert.equal(view.root.length, 0);
+		assert.equal(undoStack.length, 0);
```
jason-ha: Should there be a capture in here of `undoStack[0]` and/or `[1]` before `removeAfter` and an attempt to call `revert` on either or both?

Otherwise, this looks like it is testing that `createTestUndoRedoStacks`'s returned `undoStack` is well-maintained, which is not the production code/logic.

jason-ha (approved):
"undoB" and "undoC" names look odd to me in test case but functionally looks great.

yann-achard-MS:
> "undoB" and "undoC" names look odd to me in test case but functionally looks great.

Ah, darn, I was playing with expanding the scope of the test for a while, and decided against it, but I forgot to reset the names. Fixed.

jason-ha (approved):

PR #27521, opened 2026-06-09
Fix version utils to get the versions it tries from the set it has installed

CraigMacomber:
## Description

Avoid trying to test client package versions looked up in the npm registry: instead consult the manifest of versions we have installed locally.

This prevents the act of publishing a new package from breaking tests which test against recent versions (using the N-1 style conventions).

## Reviewer Guidance


packages/test/test-version-utils/scripts/updateCompatVersions.ts
```diff
@@ -26,7 +26,7 @@
  * Commit all files produced by this script, including the updated `pnpm-lock.yaml`
  * and `CompatibilityCheckpoints.md`.
  *
- * MACHINE-MAINTAINED (do not edit by hand — regenerated by this script):
```
CraigMacomber, line 29: I found this phrasing ambiguous: it could mean this file is MACHINE-MAINTAINED, possibly by the listed scripts. This reads a bit like a warning that should be in those files. I think I have made it a little clearer.

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -349,57 +353,111 @@ export function calculateRequestedRange(
...
+	/**
+	 * If `baseVersion` is a Fluid internal version, this controls whether the public or internal
+	 * version is adjusted by `requested`.
+	 */
+	adjustPublicMajor?: boolean;
+	/**
+	 * When `true`, resolve against the npm registry. Defaults to `false`, which resolves against the
+	 * committed compat versions manifest.
+	 * @privateRemarks
+	 * Likely a better design for expressing what set of versions to consider would be to decouple the discovery of the set of versions from the resolution logic.
```
CraigMacomber, line 377: As this break is blocking CI, and this is also not an area of the codebase I normally work on, I simply made as simple of a fix as I reasonably could. As noted here, I think there is a better design which could be used that someone might wany to employ here.

packages/test/test-version-utils/src/test/versionUtils.spec.ts
```diff
@@ -28,14 +35,19 @@ const checkRequestedVersionSatisfies = (
 	expectedVersion,
 ): void => {
 	try {
-		const version = getRequestedVersion(baseVersion, requested, adjustPublicMajor);
+		const version = getRequestedVersion({
+			baseVersion,
+			requested,
+			adjustPublicMajor,
+			useOnlineRegistry: true,
```
jason-ha, line 42: Anything using online registry under unit test seems wrong/fraught unless service is mocked (which doesn't appear to be the case). Is there no current option to avoid online use during unit test?

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -128,7 +128,11 @@ export const resolveRangeViaRegistry = cached((rangeSpec: string): string => {
  * Throws if the currently installed compat workspace does not include any version that matches the spec.
  * @param rangeSpec - A valid (as per [semver](https://www.npmjs.com/package/semver)) range specification
  */
-function resolveRangeViaManifest(rangeSpec: string): string {
+export function resolveRangeViaManifest(rangeSpec: string): string {
+	if (semver.valid(rangeSpec)) {
```
Josmithr, line 132: I'm assuming `valid` checks if the string is a *version* rather than a *range*? Nit: might be worth a quick comment.

packages/test/test-version-utils/src/test/versionUtils.spec.ts
CraigMacomber: I would like to avoid this: its clearly bad but not a regression (its just more clear that its bad now). Its testing a bunch of cases which don't exist in our manifest, and if they did, would break in future changes to the manifest. I fixed the cases breaking CI, and made it clear that this test has network dependence: I consider that progress, and enough for now.

The approach I suggest in https://github.com/microsoft/FluidFramework/pull/27521/changes#r3382359766 would solve this problem much better, but require more work: something someone else should probably do after the build is un broken, but not something that seems in scope as release engineer work, not as part of my other work.

packages/test/test-version-utils/src/versionUtils.ts
CraigMacomber: I believe that checks if its a valid semver expression ( a superset of the range cases validated below), as documented by the linked API docs from this functions documentation a few lines above.

This is the exact same check used in the registry variant of this function, with the exact same documentation. We could deduplicate this logic if we did https://github.com/microsoft/FluidFramework/pull/27521/changes#r3382359766.


As we redirected some codepaths from the other function to this manifest based one, making them have the same error messages and checking was needed to make the tests and error quality not degrade.

packages/test/test-version-utils/src/index.ts
```diff
@@ -64,4 +64,8 @@ export {
 	SkippedErrorExpectingTestWithDriverType,
 	SkippedErrorExpectingTestWithDriverBaseType,
 } from "./itSkipsOnFailure.js";
-export { getRequestedVersion, versionToComparisonNumber } from "./versionUtils.js";
+export {
+	getRequestedVersion,
+	type GetRequestedVersionOptions,
```
jason-ha, line 69: `getRequestedVersion` doesn't appear to be used outside of this package. Instead of exporting more, let's just not export `getRequestedVersion`.

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -128,7 +128,11 @@ export const resolveRangeViaRegistry = cached((rangeSpec: string): string => {
...
+		return rangeSpec;
```
jason-ha, line 133: This check is subtle and deserves comment. The check here looks to see if rangeSpec is a specific version (not a range).
There are no comments as to why that is checked nor why is should always be acceptable to return.

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -349,57 +353,111 @@ export function calculateRequestedRange(
...
+ * @internal
+ */
+export interface GetRequestedVersionOptions {
+	/** The base version to move from (eg. "2.60.0"). */
+	baseVersion: string;
+	/**
+	 * If the value is a negative number, the base version will be adjusted down.
+	 * If the value is a string then it will be returned as-is. Throws on positive number.
+	 */
+	requested?: number | string;
```
jason-ha, line 367: At least some callers are violating exactOptionalPropertyTypes currently. All callers except one test case specify property; so, remove the optional and allow undefined.
```suggestion
	 * If the value is a string then it will be returned as-is.
	 * If the value is a negative number, the base version will be adjusted down.
	 * If the value is 0 or `undefined`, return the base version.
	 * Otherwise (the value is a positive number), throw.
	 */
	requested: number | string | undefined;
```

packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -349,57 +353,111 @@ export function calculateRequestedRange(
...
+	// Explicit version strings are already concrete requests, so return them directly.
+	if (typeof requested === "string") {
+		return requested;
+	}
+
 	const calculatedRange = calculateRequestedRange(baseVersion, requested, adjustPublicMajor);
+	const cacheKey = JSON.stringify({ calculatedRange, useOnlineRegistry });
+	const cachedResolution = resolutionCache.get(cacheKey);
+	if (cachedResolution !== undefined) {
+		return cachedResolution;
```
jason-ha, line 433: Unused before, eh?
The PR description could you updated content and this would be reasonable callout.

jason-ha (approved):
I think we should make some additional changes, but none have to happen before merging.
... except more descriptive PR description would be appreciated.

packages/test/test-version-utils/src/versionUtils.ts
jason-ha: The other code, of course, should also be commented. 
I think the logic of that case is that given a specific version, that is the only possible resolution, and if given an unavailable version, then you'll have trouble elsewhere (reasonably fast).
In the manifest case, it is more important to check what is in the manifest. This new check subverts that without the gains.

PR #27522, opened 2026-06-09
Bump eslint-config-fluid to 14.0

TommyBrosman:
- v12 was published from the v11 branch (which was cut partway through v12), so the v12 release only included a subset of the changes.
- v13 should be published from a commit before the v14 bump. As v12 was not released, v13 will include the v12 changes.

See #27490 for context.


CraigMacomber:
> v13 should be published from a commit after the v13 bump

Not quite. I think you mean:

> v13 should be published from a commit **before** the **v14** bump

CraigMacomber:
Also:

> As v12 was not released, v13 will include the v12 changes.

The problem here was that v12 was released, it just contained a subset of the v12 changes (In this case likely an empty subset) since it released from a branch (the 11 branch) which was cut part way through 12, not at the end of it.

CraigMacomber:
> v13 should be published from a commit before the v14 bump. As v12 was not released, v13 will include the v12 changes.

Its more specific than that. It should be:

> v13 should be published from **the** commit before the v14 bump. As v12 was not released, v13 will include the v12 changes.


(Yes, I missed that detail in my previous comment). Anyway, code looks good, I'll approve.

CraigMacomber (approved):

PR #27523, opened 2026-06-09
docs: add Concepts documentation tree (testing + summarization branches)

agarwal-navin:
## Description

Introduces a tree-style engineering documentation structure for the repo. A new root-level `CONCEPTS.md` acts as a single, navigable entry point that gives a high-level tour of how the framework works and links out to deeper docs that live next to the code they describe.

Each package can in turn have its own CONCEPTS.md doc if it chooses to do so. For example, tree DDS has a lot of interesting concepts some with docs and some without - it can tie it all with this style.

Another advantage of this is it helps agents find information easily.

`CONCEPTS.md` covers, at a glance. These are just the initial set of concepts to show as examples - More will be added in the future.

- **Programming models** — encapsulated vs. declarative
- **Layers** — the four layers (Driver / Loader / Runtime / Datastore) and how the container runtime, data store runtime, and DDSes nest within them
- **Compatibility** — API, layer, cross-client, and data-at-rest
- **Op processing**, **Summarization**, and **Garbage collection** — linking to the existing module READMEs under `container-runtime`
- **Testing**

This PR fully wires in the first two branches as worked examples:

- **Testing** — new `TESTING.md` is the cross-cutting hub. It makes explicit that tests live in two places: per-package unit/fuzz tests under each package's `src/test`, and the cross-cutting suites (e2e, compat, snapshot, load) under `packages/test`. A new `packages/test/README.md` describes that directory's shared/cross-cutting test infrastructure.
- **Summarization** — wired through the existing, high-quality `container-runtime/src/summary/README.md` rather than duplicating it.

The root `README.md` gets a callout near the top linking to `CONCEPTS.md` so the tree is discoverable from the repo landing page.

Docs-only change: no code, public API, or package behavior is affected.

This is an initial proposal for the documentation style — feedback welcome on:

- The overall tree shape and section ordering in `CONCEPTS.md`.
- The split between root-level overview docs vs. docs that live next to code.
- Which branches to flesh out next (op processing and GC already have module READMEs ready to wire deeper; Layers could use a dedicated overview doc).

Marking as draft while the style is being discussed.

### DIFF
--- a/CONCEPTS.md
+++ b/CONCEPTS.md
@@ -1,0 +1,171 @@
+# Fluid Framework — Concepts
+
+This is the **root of the engineering documentation tree**. It gives a high‑level,
+conceptual tour of how the Fluid Framework is put together and links out to deeper
+documents that live next to the code they describe.
+
+The goal is a single, navigable starting point: read a short overview of a concept here,
+then follow the link into the package/folder where that concept lives for the full detail.
+
+> **How this tree works**
+> Each top‑level concept below has a short "what & why" summary and a **Learn more** list of
+> links. Overview docs live here at the root; detailed docs live in the relevant sub‑folder
+> (e.g. summarization lives in `container-runtime`, end‑to‑end testing lives in `packages/test`).
+> When you add a new doc deep in the repo, link it back up into the right section here so it
+> stays reachable.
+
+---
+
+## Table of contents
+
+- [Programming models](#programming-models) — encapsulated vs. declarative
+- [Layers](#layers) — the four layers: Driver / Loader / Runtime / Datastore
+- [Compatibility](#compatibility) — API, layer, cross‑client, and data‑at‑rest
+- [Op processing](#op-processing) — batching, compression, chunking
+- [Summarization](#summarization) — snapshots, the summarizer client, incremental summaries
+- [Garbage collection](#garbage-collection) — mark / sweep of unreferenced objects
+- [Testing](#testing) — unit, end‑to‑end, cross‑version, fuzz, and load testing
+
+---
+
+## Programming models
+
+Fluid exposes two ways to build an application on top of the same runtime:
+
+- **Encapsulated model** — your code packages data and logic into `DataObject`s that the
+  container loads by code. Maximum flexibility and dynamic loading; more boilerplate.
+- **Declarative model** — you describe a fixed schema of shared objects up front (e.g. via
+  `SharedTree` / the `fluid-framework` container schema) and the framework wires it up. Simpler
+  to consume; the shape is known at build time.
+
+Understanding which model an application uses drives a lot of downstream behavior, including how
+cross‑client compatibility is reasoned about.
+
+**Learn more**
+
+- [Application models](./ApplicationModels.md) — encapsulated vs. declarative, in depth
+
+---
+
+## Layers
+
+A Fluid client is a stack of **four layers**, each with a defined contract to the layer above
+and below:
+
+| Layer | Responsibility | Made up of |
+| ----- | -------------- | ---------- |
+| **Driver** | Talks to a specific service (ODSP, Routerlicious/AFR, local). Sends/receives ops & snapshots. | the service drivers |
+| **Loader** | Resolves and loads a container, manages the connection lifecycle. | the container loader |
+| **Runtime** | Orchestrates data stores, op processing, summarization, and GC. | the **container runtime** |
+| **Datastore** | The code+data units within a container and the shared data types that hold user data. | the **data store runtime** and the **DDSes** (SharedTree, SharedMap, sequences, …) |
+
+Note the nesting: the **container runtime** belongs to the **Runtime** layer, while both the
+**data store runtime** and the **DDSes** belong to the **Datastore** layer.
+
+Layers are versioned and shipped independently, which is why **layer compatibility** is a
+first‑class concern (see [Compatibility](#compatibility)).
+
+**Learn more**
+
+- [Layer compatibility](./LayerCompatibility.md) — how layer boundaries are validated
+- [Layer compatibility dev guide](./LayerCompatibilityDevGuide.md)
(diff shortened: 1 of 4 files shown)


TESTING.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Running tests
+
+See the [root README](./README.md#testing) for the canonical commands:
+
+- `pnpm test` — all tests from the repo root, or a scoped set when run inside a package.
+- `pnpm build-and-test <name-regex>` — incremental build + test.
+- `ci:test` / `ci:test:coverage` — mirror the CI pipeline.
+- Add `.only` / `.skip` to focus or exclude individual tests; some tests need the Git LFS
+  submodule (`FluidFrameworkTestData`) fetched.
```
alexvy86, line 71: They seem like separate concerns to me.
```suggestion
- Add `.only` / `.skip` to focus or exclude individual tests.
- Snapshot tests need the Git LFS submodule (`FluidFrameworkTestData`) fetched.
```

alexvy86 (commented):
The testing part might want some mention of packages/service-clients/end-to-end-tests, and _maybe_ about the fact that examples/ has several packages with a pattern where tests/ lives next to src/ instead of inside it.

TESTING.md
```diff
@@ -0,0 +1,74 @@
+# Testing
```
Josmithr, line 1: We also have a `DEV.md` file at the root that has instructions for various things like running tests in the repo. It might be worth adding a link to this from there if/where appropriate.

packages/test/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+`packages/test` holds the tests and utilities that **don't belong to any single package** —
+because they exercise the whole client stack, run against real services, or run across multiple
+versions. Per‑package unit and fuzz tests live next to their own code, not here.
+
+## Suites
+
+| Package | What it is |
+| ------- | ---------- |
+| [`test-end-to-end-tests`](./test-end-to-end-tests/README.md) | The main e2e suite: a real container loop, parameterized over drivers and version combinations. |
```
Josmithr, line 16: Nit (since we haven't used the abbreviation "e2e" yet in the document):
```suggestion
| [`test-end-to-end-tests`](./test-end-to-end-tests/README.md) | The main end-to-end (e2e) suite: a real container loop, parameterized over drivers and version combinations. |
```

packages/test/README.md
```diff
@@ -0,0 +1,43 @@
...
+| [`snapshots`](./snapshots/README.md) | Validates that current code can load older snapshots/summaries (corpus partly in the `FluidFrameworkTestData` submodule). |
+| [`test-service-load`](./test-service-load/README.md) | Load/stress tool: many simulated clients and high op rates against a real or local service. |
+| [`local-server-tests`](./local-server-tests/) | Scenarios that need direct local‑server control (disconnect/nack). |
```
Josmithr, line 19: Nit
```suggestion
| [`local-server-tests`](./local-server-tests/) | Scenarios that need direct local‑server control (e.g. disconnect/nack). |
```

CONCEPTS.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Programming models
+
+Fluid exposes two ways to build an application on top of the same runtime:
+
+- **Encapsulated model** — your code packages data and logic into `DataObject`s that the
+  container loads by code. Maximum flexibility and dynamic loading; more boilerplate.
+- **Declarative model** — you describe a fixed schema of shared objects up front (e.g. via
+  `SharedTree` / the `fluid-framework` container schema) and the framework wires it up. Simpler
+  to consume; the shape is known at build time.
```
Josmithr, line 39: I don't think the SharedTree schema is relevant to the point here. I think just the container schema is relevant in this context.

TESTING.md
```diff
@@ -0,0 +1,74 @@
...
+time**. No single kind of test covers all of that, so the repo layers several complementary
+styles. Crucially, **tests live in two places**:
+
+- **In each package** — every package has its own unit tests (and, where it makes sense, fuzz
+  tests) alongside the code they exercise.
+- **In [`packages/test`](./packages/test/README.md)** — the cross‑cutting suites and shared
+  infrastructure that don't belong to any one package (end‑to‑end, cross‑version/compat, drivers,
+  snapshots, load).
+
+## Test styles at a glance
```
Josmithr, line 17: We also have performance benchmarks. I don't know what guidance we have documented for this, but would be worth calling out (we could always leave TODOs where links should be to indicate documentation that still needs to be written).

TESTING.md
```diff
@@ -0,0 +1,74 @@
...
+| **Unit tests** | Logic inside a single package, in isolation. | `src/test` in **every** package |
+| **Fuzz / stochastic** | Randomized op sequences against a DDS to find convergence/edge‑case bugs. | `src/test` of the relevant package, built on [`stochastic-test-utils`](./packages/test/stochastic-test-utils/README.md) |
+| **End‑to‑end (e2e)** | A real container loop (create → change → summarize → load) run against each driver. | [`packages/test/test-end-to-end-tests`](./packages/test/test-end-to-end-tests/README.md) |
+| **Cross‑version / compat** | The same e2e scenarios with mixed‑version layers, to validate [compatibility](./CONCEPTS.md#compatibility). | [`packages/test/test-version-utils`](./packages/test/test-version-utils/README.md) |
+| **Snapshot compatibility** | New code can still load summaries/snapshots written by old code. | [`packages/test/snapshots`](./packages/test/snapshots/README.md) |
+| **Load / stress** | Behavior under many clients and high op rates. | [`packages/test/test-service-load`](./packages/test/test-service-load/README.md), [`local-server-stress-tests`](./packages/test/local-server-stress-tests/) |
+
+## Tests that live in each package
```
Josmithr, line 28: I think the SharedTree package has some in-package testing best practices documentation as well that we may or may not want to hoist up to this level.

Josmithr: (Not necessarily for this PR)

PR #11082, opened 2022-07-13
Fix version utils

sonalivdeshpande:
Similar to https://github.com/microsoft/FluidFramework/pull/10820

[AB#784](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/784)

PR #27526, opened 2026-06-09
fix(client-test-version-utils): restore resolveRangeViaManifest behavior

jason-ha:
Cleanup from #27521 CI repair:
- restore behavior of `resolveRangeViaManifest` to only allow version that is within manifest
- update unit tests to check expectations
- remove `getRequestedVersion` from package exports
- require `requested` be specified when calling `getRequestedVersion`
- add/correct various comments

Cleanup from #11082:
- correct `getRequestedVersion` throwing test cases (that weren't checking error message content)
- remove extraneous try-catch and wrap in `calculateRequestedRange`

### DIFF
--- a/packages/test/test-version-utils/src/versionUtils.ts
+++ b/packages/test/test-version-utils/src/versionUtils.ts
@@ -93,6 +93,12 @@ function validateRangeSpec(rangeSpec: string): void {
 /**
  * Resolves a semver dependency spec to the single highest version matching that spec which is published in the npm registry.
  * @param rangeSpec - A valid (as per [semver](https://www.npmjs.com/package/semver)) range specification
+ * @remarks
+ * When rangeSpec is an exact version, registry check is skipped.
+ * @privateRemarks
+ * This function can be costly time-wise and results are trivially cached.
+ * Higher-order caching is handled by `getRequestedVersion` but this is kept to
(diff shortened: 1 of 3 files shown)


packages/test/test-version-utils/src/versionUtils.ts
```diff
@@ -354,20 +351,20 @@ export function calculateRequestedRange(
...
  */
 export interface GetRequestedVersionOptions {
 	/** The base version to move from (eg. "2.60.0"). */
 	baseVersion: string;
 	/**
-	 * If the value is a negative number, the base version will be adjusted down.
-	 * If the value is a string then it will be returned as-is. Throws on positive number.
+	 * If the value is a string, then it will be returned as-is.
+	 * If the value is a negative number, the base version will be adjusted down and returned.
+	 * If the value is 0 or `undefined`, the base version will be returned as-is.
```
CraigMacomber, line 361: Since `-0` is a thing, and is a negative number, and usually considered equal to 0, and also might make sense in this code (to try and test against the most recent release, since -1 means previous beta break), we might need to clarify or rephrase this to be unambiguous on how -0 behaves.

CraigMacomber (approved):

scottn12 (approved):

sonalideshpandemsft (approved):

PR #27528, opened 2026-06-10
tree: Improve asymptotic performance of RangeMap.union

alex-pardes:
## Description

Reimplemented `RangeMap.union` to use an algorithm which takes advantage of efficient b-tree union to find the union of two range maps in time linearithmic in the number of interleaved ranges.


packages/dds/tree/src/util/rangeMap.ts
```diff
@@ -215,19 +217,116 @@ export class RangeMap<K, V> {
...
+			(key, length, value) => merged.tree.set(key, { value, length }),
+		);
+
+		return merged;
+	}
+
+	/**
+	 * Calls provided handlers on intersecting portions of `mapA` and `mapB`.
+	 * @param intersectionCallback - called once for each key which has an entry in both `mapA` and `mapB`.
+	 * @param residualCallback - called for each key which only has an entry in one of the input maps,
```
yann-achard-MS:
```suggestion
	 * @param intersectionCallback - called once for each key range which has an entry in both `mapA` and `mapB`.
	 * @param residualCallback - called for each key range which only has an entry in one of the input maps,
```

yann-achard-MS (approved):

PR #27530, opened 2026-06-10
Updated the client release group to use eslint-config-fluid v13.0

TommyBrosman:
## Description

Updates the client release group to use `@fluidframework/eslint-config-fluid` v13 (from v9) and fixes all lint errors the new config surfaces.

Because the client and examples release groups share a single pnpm catalog entry for `eslint-config-fluid`, bumping the catalog applies v13 to **both** groups. This PR therefore also fixes lint errors in the examples, not just the client release group.

### Lint fixes

- **`no-only-tests`** — added targeted suppressions where `.only` is selected by test data/options rather than committed accidentally (`merge-tree`, `container-runtime`, `local-server-stress-tests`, `map`).
- **`no-restricted-syntax`** — import the TypeBox `Type` namespace via `import * as Type from "@sinclair/typebox"` instead of the named value export, which defeats tree-shaking (`tree-cli-app`).
- **`@eslint-react/no-create-ref`** — use `useRef(null)` instead of `createRef()` in a function component (`triangle-view`). The original allocated a new ref object every render.
- **`@eslint-react/no-useless-fragment`** — return `null` instead of an empty fragment (`data-object-grid`).
- **`@eslint-react/dom/no-render`** — migrate `ReactDOM.render` to `createRoot().render()` (`shared-tree-demo`, `app-insights-logger`, `view-framework-sampler`, `devtools-test-app`).
- Repaired the `text-editor` and `inventory-app` eslint configs, which referenced the removed classic `react` plugin (v13 replaced it with `@eslint-react`), re-expressing the intended rules via `@eslint-react` / `react-hooks`.

### Notable behavioral details

- **Render-callback timing:** `ReactDOM.render`'s third callback argument has no equivalent on `createRoot().render()`. Where the callback mattered, the side effect was moved into a `useLayoutEffect` (runs after DOM commit, before paint) rather than firing before the commit. `flushSync` is not an option (`@eslint-react/dom/no-flush-sync` is an error).
- **`devtools-test-app` StrictMode removal:** this app creates a singleton Devtools instance and disposes it in an effect cleanup. The original wrapped the app in `<StrictMode>` while rendering via the legacy `ReactDOM.render` API. React 18's StrictMode setup→cleanup→setup effect double-invoke only runs under `createRoot`; under legacy `ReactDOM.render` the tree stays in React-17 mode, so that check never fired and the wrapper was inert for effects. Migrating to `createRoot` activated the double-invoke, which disposed the singleton on the simulated unmount and left later renders using a disposed instance (`The devtools instance has been disposed`). The app is not resilient to that remount cycle, so `<StrictMode>` was removed to preserve the app's actual prior behavior; making it StrictMode-safe would be a larger refactor out of scope here.

## Reviewer Guidance

- This is a tooling/lint-config change. No production/public API surface is modified — changes are limited to example apps, test files, and eslint configs.
- All affected example apps were run manually and verified working: `data-object-grid`, `tree-cli-app`, `triangle-view` (via the multiview container), `app-insights-logger`, `shared-tree-demo`, `view-framework-sampler`, `devtools-test-app`, `text-editor`, `inventory-app`.
  - Caveat: `shared-tree-demo` was verified without a real `SPE_CLIENT_ID` (the changed create-new render path does not depend on SPE).
- The `text-editor` / `inventory-app` config changes drop three classic stylistic `react/*` rules (`jsx-boolean-value`, `jsx-fragments`, `self-closing-comp`) that have no `@eslint-react` equivalent. Flag if you'd prefer those preserved another way.

### DIFF
--- a/examples/apps/data-object-grid/eslint.config.mts
+++ b/examples/apps/data-object-grid/eslint.config.mts
@@ -16,7 +16,7 @@ const config: Linter.Config[] = [
 			"@typescript-eslint/unbound-method": "off",
 			"import-x/no-unassigned-import": "off",
 
-			// In React components, it is convention to use null to represent the absence of render output.
+			// In React components it is convention to use null to represent the absence of render output.
 			"unicorn/no-null": "off",
 		},
 	},
--- a/examples/apps/data-object-grid/eslint.config.mts
+++ b/examples/apps/data-object-grid/eslint.config.mts
@@ -15,6 +15,9 @@ const config: Linter.Config[] = [
 			"@typescript-eslint/strict-boolean-expressions": "off",
 			"@typescript-eslint/unbound-method": "off",
 			"import-x/no-unassigned-import": "off",
+
+			// In React components, it is convention to use null to represent the absence of render output.
+			"unicorn/no-null": "off",
 		},
 	},
 ];
--- a/examples/apps/data-object-grid/eslint.config.mts
+++ b/examples/apps/data-object-grid/eslint.config.mts
@@ -16,6 +16,7 @@ const config: Linter.Config[] = [
 			"@typescript-eslint/unbound-method": "off",
 			"import-x/no-unassigned-import": "off",
 
+			// TODO: AB#75619: Exclude React code from this rule in the base eslint config.
 			// In React components it is convention to use null to represent the absence of render output.
 			"unicorn/no-null": "off",
 		},
--- a/examples/apps/data-object-grid/src/dataObjectGridView.tsx
+++ b/examples/apps/data-object-grid/src/dataObjectGridView.tsx
@@ -102,7 +102,8 @@ const DataObjectGridView: FC<IDataObjectGridViewProps> = (
 
 	// Render nothing if there are no items
 	if (model.getItems().length === 0) {
-		return <></>;
+		// eslint-disable-next-line unicorn/no-null -- React renders nothing for a null return
+		return null;
 	}
 
 	const onGridChangeEvent = (
--- a/examples/apps/data-object-grid/src/dataObjectGridView.tsx
+++ b/examples/apps/data-object-grid/src/dataObjectGridView.tsx
@@ -102,7 +102,6 @@ const DataObjectGridView: FC<IDataObjectGridViewProps> = (
 
 	// Render nothing if there are no items
 	if (model.getItems().length === 0) {
-		// eslint-disable-next-line unicorn/no-null -- Prefer returning null for components rendering no content
 		return null;
 	}
 
@@ -155,7 +154,6 @@ const DataObjectGridView: FC<IDataObjectGridViewProps> = (
 			cols={36}
 			rowHeight={50}
 			width={1800}
-			// eslint-disable-next-line unicorn/no-null -- null is required for the GridLayout
 			compactType={null}
 			isDroppable={editable}
 			isDraggable={editable}
--- a/examples/apps/data-object-grid/src/dataObjectGridView.tsx
+++ b/examples/apps/data-object-grid/src/dataObjectGridView.tsx
@@ -102,7 +102,7 @@ const DataObjectGridView: FC<IDataObjectGridViewProps> = (
 
 	// Render nothing if there are no items
 	if (model.getItems().length === 0) {
-		// eslint-disable-next-line unicorn/no-null -- React renders nothing for a null return
+		// eslint-disable-next-line unicorn/no-null -- Prefer returning null for components rendering no content
 		return null;
 	}
 
--- a/examples/data-objects/inventory-app/eslint.config.mts
+++ b/examples/data-objects/inventory-app/eslint.config.mts
@@ -13,35 +13,14 @@ const config: Linter.Config[] = [
 	{
 		files: ["**/*.jsx", "**/*.tsx"],
 		rules: {
-			"react/no-deprecated": "off",
 			"react-hooks/exhaustive-deps": ["error"],
 			"react-hooks/rules-of-hooks": "error",
-			"react/jsx-key": [
-				"error",
-				{
-					"checkFragmentShorthand": true,
-					"checkKeyMustBeforeSpread": true,
-					"warnOnDuplicates": true,
-				},
-			],
-			"react/jsx-boolean-value": ["error", "always"],
-			"react/jsx-fragments": "error",
-			"react/no-string-refs": "error",
-			"react/no-unstable-nested-components": [
-				"error",
-				{
-					"allowAsProps": true,
-				},
-			],
-			"react/self-closing-comp": "error",
-			"react/jsx-no-target-blank": "error",
-			"react/jsx-no-useless-fragment": [
-				"error",
-				{
-					"allowExpressions": true,
-				},
-			],
-			"react/prop-types": "off",
+			"@eslint-react/no-missing-key": "error",
+			"@eslint-react/jsx-key-before-spread": "error",
+			"@eslint-react/no-string-refs": "error",
+			"@eslint-react/no-nested-component-definitions": "error",
+			"@eslint-react/dom/no-unsafe-target-blank": "error",
+			"@eslint-react/no-useless-fragment": ["error", { allowExpressions: true }],
 		},
 	},
 ];
--- a/examples/data-objects/inventory-app/eslint.config.mts
+++ b/examples/data-objects/inventory-app/eslint.config.mts
@@ -15,12 +15,7 @@ const config: Linter.Config[] = [
 		rules: {
 			"react-hooks/exhaustive-deps": ["error"],
 			"react-hooks/rules-of-hooks": "error",
-			"@eslint-react/no-missing-key": "error",
 			"@eslint-react/jsx-key-before-spread": "error",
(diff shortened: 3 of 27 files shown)


alexvy86 (approved):
Approving for devtools, but LGTM overall

examples/apps/data-object-grid/src/dataObjectGridView.tsx
```diff
@@ -102,7 +102,8 @@ const DataObjectGridView: FC<IDataObjectGridViewProps> = (
 
 	// Render nothing if there are no items
 	if (model.getItems().length === 0) {
-		return <></>;
+		// eslint-disable-next-line unicorn/no-null -- React renders nothing for a null return
+		return null;
```
CraigMacomber: Buy why use null? React functional components can also return undefined: does that result in something being rendered? That would avoid violating our no null rule.

Justification comments for using null should be about how there isn't a better option than null, not that null happens to work. If full is really the best option, the comment should instead say something like "Null is the preferred way to render nothing in a react component"

CraigMacomber (commented):
I'm not an expert on the react code being impacted, but everything else looks good, and the react stuff seems potentially reasonable, but I'll leave its review to others.

examples/apps/data-object-grid/src/dataObjectGridView.tsx
TommyBrosman: I can change the wording. Rationale for why this matters:
- Returning `</>` causes React to instantiate an empty fragment node. This is not really an issue in an example app, but incurs an unnecessary performance hit for apps with larger component trees.
- Returning `undefined` works, but is discouraged due to ambiguity. I believe some component authors use it for cases where the state required by the component is not available. There are also cases where allowing an undefined return type masks an early return bug. Allowing `undefined` returns from components at all is a relatively recent thing: https://github.com/reactwg/react-18/discussions/75

TommyBrosman: Fixed. Another approach would be just removing the rule for React examples.

examples/apps/data-object-grid/src/dataObjectGridView.tsx
```diff
@@ -102,7 +102,8 @@ const DataObjectGridView: FC<IDataObjectGridViewProps> = (
 
 	// Render nothing if there are no items
 	if (model.getItems().length === 0) {
-		return <></>;
+		// eslint-disable-next-line unicorn/no-null -- Prefer returning null for components rendering no content
```
Josmithr: I assume we now have two conflicting rules here now? We may want to consider disabling the `no-null` rule in React contexts (for a future separate PR).

examples/data-objects/inventory-app/eslint.config.mts
```diff
@@ -13,35 +13,14 @@ const config: Linter.Config[] = [
...
-					"allowExpressions": true,
-				},
-			],
-			"react/prop-types": "off",
+			"@eslint-react/no-missing-key": "error",
+			"@eslint-react/jsx-key-before-spread": "error",
+			"@eslint-react/no-string-refs": "error",
+			"@eslint-react/no-nested-component-definitions": "error",
+			"@eslint-react/dom/no-unsafe-target-blank": "error",
+			"@eslint-react/no-useless-fragment": ["error", { allowExpressions: true }],
```
Josmithr: I assume these aren't already covered by the base config?

Josmithr (approved):

examples/apps/data-object-grid/src/dataObjectGridView.tsx
TommyBrosman: Yeah, that was what I suggested in the thread with @CraigMacomber. I don't think a single line is worth a suppression necessarily, but perhaps we need to add this setting to reactRecommendedOverride or something. I'm not actually sure if that's the right place to put it. I'll leave it out of this PR.

examples/data-objects/inventory-app/eslint.config.mts
TommyBrosman: I checked, a lot of these overrides are the same as the base configuration (also true for text-editor). I'll fix it in the next commit.

examples/apps/data-object-grid/src/dataObjectGridView.tsx
TommyBrosman: Actually, it might make sense to put it in the package's eslint config now. It will be easier to find in one place if we ever introduce this rule for all tsx files (which we should).

packages/tools/devtools/devtools-view/src/components/graphs/DynamicComposedChart.tsx
```diff
@@ -234,45 +299,6 @@ export function DynamicComposedChart(props: DynamicComposedChartProps): ReactEle
 		);
 	};
 
-	/**
-	 * Renders a custom view for the X Axis displayed on the Rechart chart
-	 * @remarks Recharts doesn't have a type for the arguments passed to this function
-	 */
-	const CustomizedXAxisTick = (xAxisProps: any): ReactElement => {
```
alexvy86, line 241: Just curious: what's the rationale for making this into functions? Avoid the closure over `graphColorPalette`? If so, is it for a particular reason?

alexvy86 (approved):
Reapproving for devtools

packages/tools/devtools/devtools-view/src/components/graphs/DynamicComposedChart.tsx
TommyBrosman: I missed that, I'll switch it back to using closures.

examples/external-data/eslint.config.mts
```diff
@@ -49,7 +49,11 @@ const config: Linter.Config[] = [
 			"import-x/no-internal-modules": [
 				"error",
 				{
-					"allow": [...importInternalModulesAllowedForTest, "**/src/*/*.js"],
+					"allow": [
+						...importInternalModulesAllowedForTest,
+						"**/src/*/*.js",
+						"react-dom/client",
```
Josmithr, line 57: Nit: probably worth a quick comment

packages/tools/devtools/devtools-view/src/components/graphs/DynamicComposedChart.tsx
```diff
@@ -166,6 +165,72 @@ const createGraphColorPalette = (
 	}
 };
 
+/**
+ * Renders a custom view for the X Axis displayed on the Rechart chart
+ * @remarks Recharts doesn't have a type for the arguments passed to this function
+ */
```
Josmithr: This is intended for the component, and not the props interface, right? Same below,

examples/apps/data-object-grid/eslint.config.mts
```diff
@@ -15,6 +15,9 @@ const config: Linter.Config[] = [
 			"@typescript-eslint/strict-boolean-expressions": "off",
 			"@typescript-eslint/unbound-method": "off",
 			"import-x/no-unassigned-import": "off",
+
+			// In React components it is convention to use null to represent the absence of render output.
```
Josmithr, line 20: Nit: we should probably leave a TODO that this should be disabled for React code in the base config, and this override can be removed once that's been updated.

pnpm-lock.yaml
```diff
@@ -33523,6 +33803,17 @@ snapshots:
     transitivePeerDependencies:
       - supports-color
 
+  '@typescript-eslint/utils@8.61.0(eslint@9.39.1(jiti@2.6.1))(typescript@5.4.5)':
```
jason-ha, line 33806: This is now the 4th version of `@typescript-eslint/utils` and similar packages. Why aren't any older versions going away?
I guess ^ references aren't good for most linting packages and none of this can be deduplicated?

examples/apps/data-object-grid/eslint.config.mts
TommyBrosman: Done.

Josmithr (approved):

examples/view-integration/view-framework-sampler/src/views/reactView.tsx
```diff
@@ -4,7 +4,8 @@
  */
 
 import { type FC, type JSX, useEffect, useState } from "react";
-import ReactDOM from "react-dom";
+// eslint-disable-next-line import-x/no-internal-modules -- This is the pattern prescribed by React
+import { createRoot } from "react-dom/client";
```
jason-ha, line 8: Seems like general rules we'd like in the general config or at least in examples/.eslintrc.data.cjs.

TommyBrosman: Ah, I meant to put this in the package eslint config. I agree though: in general we should just add an exception for all React source. I'll file a task.

TommyBrosman: AB#75625 tracks the cleanup.

pnpm-lock.yaml
TommyBrosman: (Discussed offline. Likely not in scope for this PR, but still worth looking at.)

anthony-murphy (approved):

PR #25398, opened 2025-09-04
Update client to ES2022

CraigMacomber:
## Description

Update client to ES2022. 

ES2022 saves over 2% (7309 bytes) on bundle size (measured using SharedTree bundles which includes all of the runtime shared tree uses as well).

Also building targeting ES2022 provides a better debug experience for JS private properties by not polyfilling them to weak map lookups.

Note that Fluid Framework has no validation that our dependencies don't already have dependencies on newer things, and no compatibility testing for older or less supported runtimes.

Tracked by [AB#34388](https://dev.azure.com/fluidframework/internal/_workitems/edit/34388).

## Breaking Changes

See changeset

## Reviewer Guidance

The review process is outlined on [this wiki page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

This needs special review based on potential impact on customers and our current lack of process for how to handle such updates.

PR #27531, opened 2026-06-10
build(tree): enable importHelpers to dedupe TypeScript runtime helpers

TommyBrosman:
## Description

Enable `importHelpers` (with a `tslib` dependency) in `@fluidframework/tree`. With this on, the TypeScript compiler emits `import { ... } from "tslib"` instead of inlining the runtime helpers (e.g. `__assign`, `__awaiter`, `__spreadArray`) into every module that uses them. Webpack can then dedupe those helpers down to a single shared copy, shrinking the bundle.

Measured impact (parsed bytes), comparing against the parent revision:

| Entrypoint | Base | Current | Diff | % |
|---|--:|--:|--:|--:|
| `sharedTree.js` | 401,086 | 392,105 | **−8,981** | **-2.2%** |
| `fluidFrameworkAll.js` | 1,326,045 | 1,317,064 | **-8,981** | **-0.7%** |

Gzipped: `sharedTree.js` -1,948 B, `fluidFrameworkAll.js` -1,949 B.

The mechanism is visible in the per-package breakdown: inlined helpers drop out of `@fluidframework/fluid-framework` (-12,499 B) while a single shared `tslib` module is added (+3,902 B), for a net ~-8,981 B per affected entrypoint. All other entrypoints are unchanged.

Changes:
- Add `"importHelpers": true` to `packages/dds/tree/tsconfig.json`.
- Add `tslib` as a dependency in `packages/dds/tree/package.json` (and corresponding lockfile update).

## Reviewer Guidance

- `tslib` is added as a runtime dependency. This is not a widely-used dependency in our repo (client-logger/fluid-telemetry uses it, so does the devtools browser extension). Are there any potential issues if tree depends on it?
- The bundle numbers above were produced with `flub bundle collect-and-compare` against the commit's parent revision. See #27151 for the version of the tools used.


packages/dds/tree/tsconfig.json
```diff
@@ -5,6 +5,11 @@
 	"compilerOptions": {
 		"rootDir": "./src",
 		"outDir": "./lib",
+		// Avoid emitting __classPrivateFieldGet/Set, __esDecorate, __runInitializers, etc.
```
CraigMacomber, line 8: For at least __classPrivateFieldGet, I think it would be much better to just target es version we said we required when we released FF 2.0, ES2022, so we use the build in JS private fields, instead of emitting legacy adapter stuff with size and time overhead. We have known for a long time doing so would give a significant bundle size and likely performance win, and also better devX. I have an old PR for that in https://github.com/microsoft/FluidFramework/pull/25398

If we decide not to go that route, or it doesn't address all the cases, I think whatever choice we make should apply to all of FF Client, and not just tree.

TommyBrosman: I think setting a unified policy for client packages makes sense. Only a subset will see noticeable size reductions though:

- dds/tree
- runtime/container-runtime
- loader/container-loader
- dds/sequence
- dds/shared-object-base
- framework/aqueduct

The ES2022 change makes sense to me, and I hope we get it prioritized. I can update my comment to mention that this __classPrivateFieldGet will go away with the ES bump. I think there is value to landing the tree changes first: we get an idea as to whether using tslib works well in client packages.

Josmithr (approved):

PR #27532, opened 2026-06-10
improvement(tree): Misc. TableSchema optimizations

Josmithr:
1. Avoid unnecessary ID->node->ID roundtrips in `TableSchema.getCell`
2. Batch group column/row deletions into range deletions when possible (also reduces overall op size)
3. Misc. property / array access pattern improvements


packages/dds/tree/src/tableSchema.ts
```diff
@@ -823,33 +823,45 @@ export namespace System_TableSchema {
...
 									delete (row as RowValueInternalType).cells[columnToRemove.id];
 								}
+							}
 
-								// We have already validated that all of the columns exist above, so this is safe.
-								this.table.columns.removeAt(this.table.columns.indexOf(columnToRemove));
+							// Remove column nodes. Ranges are iterated in reverse so each `removeRange`
+							// call doesn't shift the indices of yet-to-remove ranges at lower positions.
+							for (let r = ranges.length - 1; r >= 0; r--) {
+								const { start, end } = ranges[r] as { start: number; end: number };
```
daesunp: nit: Would it be better to use a `for of` loop here to remove this cast? or import `IndexRange` and cast it as that instead?

packages/dds/tree/src/tableSchema.ts
```diff
@@ -823,33 +823,45 @@ export namespace System_TableSchema {
 					}
 
 					// Resolve any IDs to actual nodes.
-					// This validates that all of the rows exist before starting transaction.
+					// This validates that all of the columns exist before starting transaction.
 					// This improves user-facing error experience.
-					const columnsToRemove: ColumnValueType[] = [];
+					// Set insertion order is preserved on iteration, which matches the caller-supplied order
+					// expected by both the cell-deletion loop and the returned array.
+					const columnsToRemove = new Set<ColumnValueType>();
```
daesunp, line 832: Nit: Now that we are using a set, duplicate removes should no longer throw (for example `removeColumns([col[0], col[0]])`). Should we add a unit test to verify that this doesn't throw?

daesunp (approved):

packages/dds/tree/src/tableSchema.ts
Josmithr: We're intentionally iterating backwards here, but we should definitely use the type.

Josmithr: Actually, the cast here was to work around the fact that we have TypeScript configured to require explicit bounds checking (and so the result could be undefined). Updated to use our `oob` helper.

packages/dds/tree/src/tableSchema.ts
Josmithr: Good call. Added!

daesunp (approved):

PR #27533, opened 2026-06-12
fix(runtime): Preserve errorType etc through IResponseException roundtrip

markfields:
Fixes [AB#75494](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/75494)

## Description

If we are under a "request"-based API like FluidDataStoreRuntime.request, when we throw a DataProcessingError with tons of telemetry props plus all LoggingError's goodness, all that gets stripped off as it temporarily is converted to an IResponseException (which only has message and stack and a few response-related props) and back to Error.

Instead we should stash the original error on there and use it (with the extra response-related props) when going back to Error shape.

This is obscuring important errors that propagate up to host error logging.

## Breaking Changes

None, this preserves the behavior needed by consumers of the implicit `IResponseException` contract (despite it being not exported)


markfields:
Leaving as a draft because I anticipate this could have unintended side effects in 1P partner repo we tightly integrate with, due to strongly-typed errors newly propagating where they should have been the whole time.

So I'm going to wait until after the 2.110 release to merge this.

packages/runtime/runtime-utils/src/dataStoreHelpers.ts
```diff
@@ -82,21 +156,23 @@ export function exceptionToResponse(error: unknown): IResponse {
  * @internal
  */
 export function responseToException(response: IResponse, request: IRequest): Error {
```
markfields, line 158: Comment from Copilot when considering whether this PR would fix a real-world problem where error context was lost during propagation:

> responseToException  also marks the preserved error as an  IResponseException . When Scheduler initialization then crosses its outer boundary, the PR’s  exceptionToResponse  takes the existing- IResponseException  branch, which currently copies status/message/headers but does not carry  originalError  forward. The outer  responseToException  therefore reconstructs a plain  Error  again.
Fluid PR #27533 needs to preserve the original  LoggingError  through repeated/nested response conversions for this exact path. Once it does that, no Tasks change should be required.

PR #27535, opened 2026-06-12
Cleanup pack-packages for better error propagation and pnpm 11 support

CraigMacomber:
## Description

Prevent swallowing of errors, and fix pnpm 11 support.

This use of && seemed harmful to me, as it obscured the error (this script did not fail). Claude agreed:

> These are separate statements (not chained with `&&`) so that `set -e` aborts the script on the first
> failure. Under `set -e`, a failing command that is not the last link of an `&&` chain is exempt from
> exit-on-error, so chaining would silently swallow a failed `pnpm pack` and let the script continue.

The && below is also removed for the same reason.

## Reviewer Guidance

### DIFF
--- a/build-tools/packages/build-cli/package.json
+++ b/build-tools/packages/build-cli/package.json
@@ -50,7 +50,6 @@
 		"format:biome": "biome check --write .",
 		"lint": "npm run eslint",
 		"lint:fix": "npm run eslint:fix",
-		"postpack": "npm run clean:manifest",
 		"test": "npm run test:mocha",
 		"test:coverage": "c8 npm run test",
 		"test:mocha": "mocha --forbid-only",
--- a/scripts/pack-packages.sh
+++ b/scripts/pack-packages.sh
@@ -19,15 +19,23 @@ if [ -f ".releaseGroup" ]; then
   if [ "$PACKAGE_MANAGER" == "pnpm" ]; then
     flub exec --no-private --concurrency=1 --releaseGroup $RELEASE_GROUP -- "pnpm --if-present pack:tests"
   fi
-  flub exec --no-private --concurrency=1 --releaseGroup $RELEASE_GROUP -- "$PACKAGE_MANAGER pack" && \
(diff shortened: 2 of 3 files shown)


build-tools/packages/build-cli/package.json
```diff
@@ -50,7 +50,6 @@
 		"format:biome": "biome check --write .",
 		"lint": "npm run eslint",
 		"lint:fix": "npm run eslint:fix",
-		"postpack": "npm run clean:manifest",
```
tylerbutler, line 53: This _might_ cause weirdness with the released packages. Are you sure it's not needed?

scripts/pack-packages.sh
```diff
@@ -19,15 +19,23 @@ if [ -f ".releaseGroup" ]; then
...
+  flub exec --no-private --concurrency=1 --releaseGroup $RELEASE_GROUP -- "$PACKAGE_MANAGER pack"
+  flub exec --no-private --concurrency=1 --releaseGroup $RELEASE_GROUP -- "mv -t $STAGING_PATH/pack/tarballs/ ./*.tgz"
   flub exec --no-private --releaseGroup $RELEASE_GROUP -- "[ ! -f ./*test-files.tar ] || (echo 'test files found' && mv -t $STAGING_PATH/test-files/ ./*test-files.tar)"
 
+  # Clean up generated files that are listed in a package's "files" array (e.g. oclif.manifest.json) after packing.
+  # These are produced during build and intentionally shipped in the published tarball, but should not linger in
+  # the working tree. This cleanup deliberately runs here rather than in a package "postpack" script: under
+  # pnpm >=11, `pnpm pack` re-stats every "files" entry after postpack runs and fails with ENOENT if postpack
+  # deleted one. Running it here, after all packs complete, avoids that while still cleaning the working tree.
+  flub exec --no-private --releaseGroup $RELEASE_GROUP -- "pnpm run --if-present clean:manifest"
```
tylerbutler, line 31: Oh I see now. Carry on.

tylerbutler (approved):
I don't think this code is well tested so be ready for something to break unexpectedly when you try to release later.

PR #27539, opened 2026-06-12
improvement(tree): Update table row insertion/removal constraints to only specify columns with intersecting cells.

Josmithr:
Update `insertRows` and `removeRows` to only specify columns for which there are existing cells as constraints. 

For sparse tables, this is a pretty big win in terms of both constraint performance and usability.

Note that `setCell` and `removeCell` also have corresponding column "nodeInDocument" constraints, so the code is still robust to cell-level changes.


packages/dds/tree/src/tableSchema.ts
```diff
@@ -1072,6 +1074,55 @@ export namespace System_TableSchema {
...
 
+			/**
+			 * Builds an array of `nodeInDocument` constraints over each column referenced by a cell
+			 * in any of `rows`.
+			 *
+			 * @remarks
+			 * The orphan-cell invariant only cares about columns that have cells in the affected rows;
+			 * columns absent from those rows' cells cannot be orphaned by this operation, so they need
+			 * not be constrained. This keeps the constraint set proportional to the cells touched
+			 * rather than the total column count, which matters for sparse tables.
```
yann-achard-MS: The current comment makes it sound as though it's the columns that cannot be orphaned.
```suggestion
			 * columns absent from those rows' cells have no cell to orphan, so they need not be
			 * constrained. This keeps the constraint set proportional to the cells touched rather than
			 * the total column count, which matters for sparse tables.
```

packages/dds/tree/src/tableSchema.ts
```diff
@@ -1072,6 +1074,55 @@ export namespace System_TableSchema {
...
+					return undefined;
+				}
+
+				const columnCache = this.#getColumnCache();
+				const constraints: TransactionConstraintAlpha[] = [];
+				for (const id of referencedColumnIds) {
+					const column = columnCache.get(id) ?? fail("Column ID not found in cache");
+					constraints.push({ type: "nodeInDocument", node: column });
+				}
+				return constraints.length > 0 ? constraints : undefined;
```
yann-achard-MS: Nit: seems a bit redundant to have both this check and the `if (referencedColumnIds.size === 0)` check.

yann-achard-MS (approved):

yann-achard-MS (approved):

PR #27542, opened 2026-06-12
feat(tree): runtime schema upgrade mechanism

jenn-le:
## Description

Adds runtime support for enabling selected staged schema upgrades in SharedTree.

This updates `TreeView.initialize` and `TreeView.upgradeSchema` to accept an optional `upgrades` property bag. Applications can use this to control when staged schema upgrades are included in stored schema, which makes it easier to separate code rollout from schema rollout and to test future document shapes before enabling them broadly.

This also adds test coverage for enabling staged allowed type upgrades and staged optional field upgrades through the new API parameter.

## Reviewer Guidance

The main area to review is the runtime schema upgrade path in SharedTree, especially the handling of explicit upgrade targets when the default schema compatibility result is already equivalent.


.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,44 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+
```
Josmithr, line 7:
```suggestion
```

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,44 @@
...
+Enable staged schema upgrades at runtime
+
+SharedTree now supports enabling selected staged schema upgrades when initializing a document or upgrading its stored schema.
+This lets applications deploy code that understands a schema change before enabling that change in documents, making it easier to separate code rollout from feature rollout.
+
+Pass an application-owned `upgrades` property bag to the alpha `TreeViewAlpha.initialize` or `TreeViewAlpha.upgradeSchema` APIs to include the corresponding staged schema upgrades in the generated stored schema.
```
Josmithr: Nit: can we link to the API docs for the `initialize`, since that already exists?

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,44 @@
...
+The property names are chosen by the application, which makes them convenient to wire to feature flags or other rollout controls.
+
+The same API also makes staged schema upgrades easier to test before production rollout.
+Tests and validation tools can create or upgrade documents with a specific set of staged upgrades enabled, then verify that older and newer application versions behave correctly against the resulting stored schema.
```
Josmithr: Nit: "against" doesn't sound right to me. Would "with" make more sense?
```suggestion
Tests and validation tools can create or upgrade documents with a specific set of staged upgrades enabled, then verify that older and newer application versions behave correctly with the resulting stored schema.
```

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,44 @@
...
+const upgrades = enableFooUpgrade ? { enableFooUpgrade: fooSchemaUpgrade } : undefined;
+
+const view = asAlpha(tree.viewWith(new TreeViewConfiguration({ schema: AppSchema })));
+
+if (view.compatibility.canInitialize) {
+	view.initialize(initialContent, upgrades);
+} else if (view.compatibility.canUpgrade) {
+	view.upgradeSchema(upgrades);
+}
+```
```
Josmithr, line 121: Could we make this into a more "real world" kind of example? As someone unfamiliar with the design, it isn't immediately clear to me what is being accomplished here. Inline comments would also be useful.

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,44 @@
...
+
+When `upgrades` is omitted or empty, staged schema upgrades remain disabled and SharedTree preserves its existing restrictive stored-schema behavior.
```
Josmithr: > preserves its existing restrictive stored-schema behavior.

Might be worth elaborating on this - it isn't immediately clear to me what this is referring to.

.changeset/enable-staged-schema-upgrades.md
Josmithr: We can also pre-fill a link to the API docs for `upgradeSchema`, since the URL pattern is predictable. Since those docs will be generated at roughly the same time we release the code, this is generally okay to do.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -450,7 +450,33 @@ export interface TreeViewAlpha<
...
+	 * @remarks
+	 * This will update the {@link TreeView.compatibility}, allowing access to `root`.
+	 * Beware that this may impact other clients' ability to view the document: see {@link SchemaCompatibilityStatus.canView} for more information.
+	 *
+	 * It is an error to call this when {@link SchemaCompatibilityStatus.canUpgrade} is false.
+	 * {@link SchemaCompatibilityStatus.canUpgrade} being true does not mean that an upgrade is required, nor that an upgrade will have any effect.
+	 *
+	 * @param upgrades - Staged schema upgrades to enable while generating the upgraded stored schema.
+	 */
+	upgradeSchema(upgrades?: Readonly<Record<string, SchemaUpgrade>>): void;
```
Josmithr: I think it would be worth adding an `@example` comment block or two to this. Ideally, a real-world-ish example that illustrates the use-case. We could probably re-use the same example for the changeset doc.

A further thought: when I added `TableSchema`, I created some examples as tests, and used those same examples in the docs. I left comments in both places indicating that they need to be kept in-sync with one another. Not quite as good as if we could automatically embed living code examples in our API docs, but maybe better than nothing. Might be applicable here?

packages/dds/tree/src/simple-tree/toStoredSchema.ts
```diff
@@ -97,20 +97,39 @@ export const permissiveStoredSchemaGenerationOptions: StoredFromViewSchemaGenera
 	includeStagedOptional: () => true,
 };
 
+function storedSchemaGenerationOptionsForUpgrades(
```
Josmithr: Nit: docs?

packages/dds/tree/src/simple-tree/api/tree.ts
jenn-le: I wonder if it would be better to directly link to the tests on github so we wouldn't have to deal with keeping it in sync? Would require not breaking the link of course so depends on what we think is less annoying

jenn-le: added some shorter examples and also linked to the test file

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,97 @@
...
+const enableChecklistItems = featureFlags.enableChecklistItems;
+
+const upgrades = enableChecklistItems
+	? {
+			// The property name is application-owned, so it can match the feature flag.
+			enableChecklistItems: checklistItemSchemaUpgrade,
+		}
+	: undefined;
+
+const view = asAlpha(tree.viewWith(new TreeViewConfiguration({ schema: AppSchema })));
```
Josmithr: I think including `AppSchema` and `checklistItemsSchemaUpgrade` in the example here would be really helpful. Without it, it's hard to form a complete picture of what the scenario looks like.

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,97 @@
+---
```
Josmithr, line 1: Not necessarily for this PR, but I think it would be really useful for us to have some scenario-driven walkthrough documentation for how a user would stage and then eventually roll out new features leveraging our APIs.

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
```diff
@@ -243,22 +248,30 @@ export class SchematizingSimpleTreeView<
...
 
-		if (!compatibility.canUpgrade) {
+		// `compatibility` is computed for the default no-upgrades schema target.
+		// Explicit upgrades can produce a different target, so validate that schema directly.
+		const newSchema = toUpgradeSchema(this.viewSchema.viewSchema.root, upgrades);
+		const storedSchema = this.checkout.storedSchema.clone();
+		if (!allowsRepoSuperset(defaultSchemaPolicy, storedSchema, newSchema)) {
 			throw new UsageError(
-				"Existing stored schema cannot be upgraded (see TreeView.compatibility.canUpgrade).",
+				"Existing stored schema cannot be upgraded to the requested schema.",
```
Josmithr: Nit: is the removed `(see TreeView.compatibility.canUpgrade)` postfix no longer useful?

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -450,7 +450,68 @@ export interface TreeViewAlpha<
...
+	 * When {@link SchemaCompatibilityStatus.canUpgrade} is true,
+	 * this can be used to modify the stored schema to make it match the view schema.
+	 * @remarks
+	 * This will update the {@link TreeView.compatibility}, allowing access to `root`.
+	 * Beware that this may impact other clients' ability to view the document: see {@link SchemaCompatibilityStatus.canView} for more information.
+	 *
+	 * It is an error to call this when {@link SchemaCompatibilityStatus.canUpgrade} is false.
+	 * {@link SchemaCompatibilityStatus.canUpgrade} being true does not mean that an upgrade is required, nor that an upgrade will have any effect.
+	 *
+	 * @example Enabling a staged allowed type for documents selected by a feature flag:
```
Josmithr: Nit
```suggestion
	 * @example Enabling a staged allowed type for documents, selected by a feature flag:
```

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -450,7 +450,68 @@ export interface TreeViewAlpha<
...
+	 * ```typescript
+	 * const stagedChecklistItem = SchemaFactoryBeta.staged(ChecklistItem);
+	 * const checklistItemSchemaUpgrade = stagedChecklistItem.metadata.stagedSchemaUpgrade;
```
Josmithr: Is `metadata.stagedSchemaUpgrade` a part of our API? Or is this just a possible pattern apps can use? If the latter, might be worth a quick note so that readers don't look for an API that doesn't exist.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -450,7 +450,68 @@ export interface TreeViewAlpha<
...
+	 * assert(checklistItemSchemaUpgrade !== undefined);
+	 *
+	 * const view = asAlpha(
+	 * 	tree.viewWith(new TreeViewConfiguration({ schema: AppSchemaWithStagedChecklist })),
+	 * );
+	 *
+	 * if (featureFlags.enableChecklistItems && view.compatibility.canUpgrade) {
+	 * 	view.upgradeSchema({ enableChecklistItems: checklistItemSchemaUpgrade });
+	 * 	addChecklistItem(view.root, { text: "Review rollout" });
```
Josmithr: Nit: an inline comment explaining what this line is doing might be helpful. I'm assuming this is actually making an edit to the tree post-upgrade?

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -450,7 +450,68 @@ export interface TreeViewAlpha<
...
+	 * const dueDateSchemaUpgrade = stagedDueDate.isStagedOptional;
+	 * assert(dueDateSchemaUpgrade !== false);
+	 *
+	 * const view = asAlpha(
+	 * 	tree.viewWith(new TreeViewConfiguration({ schema: AppSchemaWithStagedDueDate })),
+	 * );
+	 *
+	 * if (featureFlags.enableOptionalDueDate && view.compatibility.canUpgrade) {
+	 * 	view.upgradeSchema({ enableOptionalDueDate: dueDateSchemaUpgrade });
+	 * 	view.root.dueDate = undefined;
```
Josmithr: I'm assuming this isn't necessary? The default value of `dueDate` will be undefined already since it's an optional field?

packages/dds/tree/src/simple-tree/api/tree.ts
Josmithr: Longer term, I really want to invest in better tooling for automatically embedding examples. It would be great to just decorate our docs with some specifier for what code to embed and have that be automatic. But for now, none of our options are particularly great.

Thanks for the examples! Those help a lot IMO.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -450,7 +450,68 @@ export interface TreeViewAlpha<
...
+	 * {@link https://github.com/microsoft/FluidFramework/blob/main/packages/dds/tree/src/test/simple-tree/api/stagedSchemaUpgrade.spec.ts | staged schema upgrade tests}.
+	 *
+	 * @param upgrades - Staged schema upgrades to enable while generating the upgraded stored schema.
+	 */
+	upgradeSchema(upgrades?: Readonly<Record<string, SchemaUpgrade>>): void;
+
+	/**
+	 * Initialize the tree, setting the stored schema to match this view's schema and setting the tree content.
+	 *
+	 * Only valid to call when this view's {@link SchemaCompatibilityStatus.canInitialize} is true.
```
Josmithr, line 612: Nit
```suggestion
	 * @remarks
	 * Only valid to call when this view's {@link SchemaCompatibilityStatus.canInitialize} is true.
```

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -450,7 +450,68 @@ export interface TreeViewAlpha<
...
+	 *
+	 * Applications should typically call this function before attaching a `SharedTree`.
+	 * @param content - The content to initialize the tree with.
+	 * @param upgrades - Staged schema upgrades to enable while generating the initial stored schema.
```
Josmithr: It still isn't totally clear to me why the upgrades need to be passed to `initialize`. What are the invariants here. If a user doesn't specify them here, but then calls `upgrade`, what happens? Or if they specify different sets of upgrades between the calls?

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
jenn-le: It's no longer accurate since we're now explicitly checking that the new schema is a superset of the old one rather than just checking the canUpgrade property

packages/dds/tree/src/simple-tree/api/tree.ts
jenn-le: It is part of our API!

packages/dds/tree/src/simple-tree/api/tree.ts
jenn-le: Staged optional fields are fields that were required so this is just demonstrating that enabling the upgrade allows us to remove the previous value, I'll add a comment about that

packages/dds/tree/src/simple-tree/api/tree.ts
jenn-le: This is so that enabling an upgrade for a particular client through some feature flag enables the upgrade for all documents the client interacts with. Meaning any document they create (initialize) or open (upgrade) gets the enabled upgrade. Enabling an upgrade is a permanent change to a document. So even if the feature flag is turned off for this client and therefore excluded from the set of upgrades, any document that has already been upgraded will retain the upgrade. Currently, if an already enabled upgrade is disabled and then upgrade is called, it will error. The following work is to add an api to check if an upgrade has already been enabled though and part of that work will be making it so that this operation is a no op rather than erroring.

jenn-le: I added a test showing it currently errors and a todo for making it a noop

jenn-le: Actually, I'll just add that API to this PR so that there's no point where this would cause an error

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
jenn-le: I actually just realized that I need to do some more work to make this property work properly in this case

packages/dds/tree/src/simple-tree/api/tree.ts
jenn-le: This API will be a bit involved actually so I'll keep my original plan of working on it next. In the meantime, I added more documentation calling out this limitation

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -450,7 +450,71 @@ export interface TreeViewAlpha<
...
+	 * 	view.root.dueDate = undefined;
+	 * }
+	 * ```
+	 *
+	 * Full end-to-end staged schema upgrade examples can be found in the
+	 * {@link https://github.com/microsoft/FluidFramework/blob/main/packages/dds/tree/src/test/simple-tree/api/stagedSchemaUpgrade.spec.ts | staged schema upgrade tests}.
+	 *
+	 * @param upgrades - Staged schema upgrades to enable while generating the upgraded stored schema.
+	 */
+	upgradeSchema(upgrades?: Readonly<Record<string, SchemaUpgrade>>): void;
```
CraigMacomber: A few issues I have with this API:
1. Nothing in the examples or documentation I can see explains why each upgrade is associated with a string or how that string is used. If the string is not used, then we should not require a unique string key for every upgrade, and instead use something other than a record (like a set, array, iterable etc). If the string is used for better error messages, I think it should be provided when creating the SchemaUpgrade object, not in every location which uses it.
2. I find the name of the upgrades to make this API misleading. One might expect upgradeSchema to be a no-op if the set of upgrades passed into it is empty, but its not. I think it should be called something more specific like "addationalUpgrades" or "optionalUpgrades" or "stagedUpgrades".
3. We are duplicating documentation between here and initialize. I recommend having one link the other, or both share an explicitly named type which has the documentation.
4. Docs don't say what the behavior is when this parameter isn't provided. Is the default to automatically opt into any which already exist in the current stored schema? DO we still do that opt in if we explicitly provide this parameter or must the caller do so?
5. If someone wants to just test with all staged upgrades enabled, it's quite hard to do (this is something integration testing might want to do). This could be addressed by taking something in something like our existing StoredFromViewSchemaGenerationOptions type. Doing so would also future proof things as we could add more options to that object in the future if desired. Might be nice to have the view expose the same type  (StoredFromViewSchemaGenerationOptions or something like it) for what is currently in the stored schema as well so you can test if a given upgrade already exists, and note that is the default thats used if you don't provide a custom one, and the caller can make one which wraps it.

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,203 @@
...
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+Enable select staged schema upgrades at runtime via view configuration
+
+SharedTree now supports enabling selected staged schema upgrades when initializing a document or upgrading its stored schema.
+This lets applications deploy code that understands a schema change before enabling that change in documents,
+making it easier to separate code rollout from feature rollout.
```
CraigMacomber: rollout is a real word, you can add it to our vale dictionary to shut this up.

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,203 @@
...
+
+### API
+
+Select the schema upgrades to enable during view creation by passing an `enabledUpgrades` list in the config object used with [`ITreeAlpha.viewWith`](https://fluidframework.com/docs/api/tree/viewabletree-interface#viewwith-methodsignature).
```
CraigMacomber: vale is right you should use the full word not a short form here.

```suggestion
Select the schema upgrades to enable during view creation by passing an `enabledUpgrades` list in the configuration object used with [`ITreeAlpha.viewWith`](https://fluidframework.com/docs/api/tree/viewabletree-interface#viewwith-methodsignature).
```

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,203 @@
...
+
+This also means partial rollback is currently difficult in practice: callers usually cannot target only documents that have not already been upgraded, so disabling the flag after some upgrades have happened can still lead to `UsageError` if `upgradeSchema` is called from views configured without previously enabled tokens.
+
+### Testing
+
+The same API also makes staged schema upgrades easier to test before production rollout.
+Tests and validation tools can create or upgrade documents with a specific set of staged upgrades enabled in view construction options, then verify that older and newer application versions behave correctly with the resulting stored schema.
+This gives applications a direct way to exercise future document shapes without permanently enabling those upgrades for all new documents.
+
+For example, a compatibility test can start with an existing document, verify that the staged shape is not writable yet, then explicitly enable the staged schema upgrade and validate the upgraded document shape:
```
CraigMacomber: I think you triggered vale's ling sentences warning 7 times in this file. I think its worth trying to adjust the phrasing as it suggests. In this case you might be able to use something much shorter like:


```suggestion
These APIs can be used to test scenerios emulating when future application versions create documents which allow or contain the staged types.
Without such testing, it is often impractical to ensure staging accomplished its goal of preparing the current version of the application (and not just the schema) to be able to handle documents with the new types.
```

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,200 @@
...
+const view = alphaTree.viewWith(
+	new TreeViewConfigurationAlpha({
+		schema: AppSchema,
+		enabledUpgrades,
+	}),
+);
+```
+
+When configured `enabledUpgrades` is omitted or empty, staged schema upgrades remain disabled.
+The document's stored schema continues to allow only the schema that has already been enabled, so clients can understand the staged schema in code but cannot write data that depends on it yet.
```
Josmithr: Nit: I find this phrasing to be a bit confusing. It isn't obvious how the first clause implies the second.

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,200 @@
...
+const enabledUpgrades = [schemaUpgrade];
+
+const alphaTree = asAlpha(tree);
+const view = alphaTree.viewWith(
+	new TreeViewConfigurationAlpha({
+		schema: AppSchema,
+		enabledUpgrades,
+	}),
+);
+```
```
Josmithr, line 45: I like this. From an API perspective, this is simple and easy to emulate.

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,200 @@
...
+const schemaUpgrade = stagedType.metadata.stagedSchemaUpgrade;
+assert(schemaUpgrade !== undefined);
+
+const enabledUpgrades = [schemaUpgrade];
+
+const alphaTree = asAlpha(tree);
+const view = alphaTree.viewWith(
+	new TreeViewConfigurationAlpha({
+		schema: AppSchema,
+		enabledUpgrades,
```
Josmithr: Nit: is there any possibility of future ambiguity here? E.g. should this be "enabledSchemaUpgrades"?

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,200 @@
...
+			includeStaged: (upgrade) => enabledFeatures.has(upgrade),
+			includeStagedOptional: (upgrade) => enabledFeatures.has(upgrade),
+		},
+	}),
+);
+```
+
+This approach is useful for scenarios such as:
+- Fine-grained rollout control based on feature sets rather than individual feature flags
+- A/B testing different schema configurations
```
Josmithr: How would A/B testing work for schema? Wouldn't that require migrating the actual stored schema to the lowest common denominator of the tested formats?

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,200 @@
...
+"fluid-framework": minor
+"__section": tree
+---
+Enable select staged schema upgrades at runtime via view configuration
+
+SharedTree now supports enabling selected staged schema upgrades when initializing a document or upgrading its stored schema.
+This lets applications deploy code that understands a schema change before enabling that change in documents,
+making it easier to separate code rollout from feature rollout.
+
+### API
```
Josmithr: Heading level minimum for changesets is 4, FYI.

https://github.com/microsoft/FluidFramework/blob/main/.changeset/README.md#formatting

.changeset/enable-staged-schema-upgrades.md
jenn-le: There's a bit of ambiguity with "enabledSchemaUpgrades" too since that kind of implies that if you don't pass anything, no schema upgrades are enabled even if you might be doing a direct upgrade. I was thinking about calling it "enableStagedSchemaUpgrades" or "enableStagedUpgrades" but wasn't quite sure so ended up just going with the shortest option and letting the api doc do the heavy lifting. Do you have a preference?

.changeset/enable-staged-schema-upgrades.md
jenn-le: I just removed this sentence, it's unecessary and kind of specific to staged allowed types

.changeset/enable-staged-schema-upgrades.md
jenn-le: Yup, that is correct, I removed this line

.changeset/enable-staged-schema-upgrades.md
Josmithr: Hrm. I personally like being verbose and explicit, but I'm frequently the outlier there. I wonder if some scoping here might make things clearer without creating long property names. E.g.
```typescript
schemaOptions?: {
    enabledStagedUpgrades: ...
    storedSchemaGenerationOptions: ...
}
```

Still kind of verbose, but at least breaks things up somewhat?

But all of this is nitpick. So long as the docs call out the important details, I think this is fine as is.

packages/dds/tree/src/entrypoints/alpha.ts
```diff
@@ -10,314 +10,316 @@
...
-	TreeViewEvents, 
-	Unenforced, 
-	Unhydrated, 
-	UnionToIntersection, 
-	ValidateRecursiveSchema, 
-	ValidateRecursiveSchemaTemplate, 
-	ViewableTree, 
-	WithType, 
-	rollback, 
+	AllowedTypes,
```
Josmithr: Looks like you need to re-gen this file. FYI

packages/dds/tree/src/simple-tree/api/configuration.ts
```diff
@@ -155,6 +158,38 @@ export interface ITreeViewConfiguration<
...
+	 * Explicit policy for generating stored schema from the view schema.
+	 * @remarks
+	 * This policy is fixed at view construction time and cannot be changed afterwards.
+	 *
+	 * If provided, this policy is used directly for compatibility checks and for
+	 * `initialize` / `upgradeSchema` schema generation.
+	 *
+	 * This option is mutually exclusive with {@link ITreeViewConfigurationAlpha.enabledUpgrades}.
+	 */
+	readonly storedSchemaGenerationOptions?: StoredFromViewSchemaGenerationOptions;
```
Josmithr: We should be able to enforce this mutual exclusivity at the type level, no? E.g.
```typescript
export type ITreeViewConfigurationAlpha<
	TSchema extends ImplicitFieldSchema = ImplicitFieldSchema,
> = ITreeViewConfiguration<TSchema>  & ({
    readonly enabledUpgrades: Iterable<SchemaUpgrade>;
} | {
    readonly storedSchemaGenerationOptions: StoredFromViewSchemaGenerationOptions;
} | {
    // Neither
});
```

.changeset/enable-staged-schema-upgrades.md
jenn-le: I decided to rename it to enabledStagedUpgrades without the extra layer, it felt a lot more verbose without much benefit

packages/dds/tree/src/simple-tree/api/schemaCompatibilityTester.ts
```diff
@@ -12,58 +13,49 @@ import { getDiscrepanciesInAllowedContent } from "./discrepancies.js";
 import type { SchemaCompatibilityStatus } from "./tree.js";
 
 /**
- * A collection of View information for schema
- * @remarks
- * This contains everything needed to determine compatibility with a given stored schema.
+ * Determines the compatibility of a stored document (based on its stored schema) with a viewer (based on its view schema).
```
Josmithr, line 16: Nit: could this be a separate PR? This seems like a nice cleanup that we could land while the rest of the PR gets reviewed.

packages/dds/tree/src/simple-tree/api/configuration.ts
```diff
@@ -155,6 +158,47 @@ export interface ITreeViewConfiguration<
...
+				 *
+				 * This option is mutually exclusive with {@link ITreeViewConfigurationAlpha.enabledStagedUpgrades}.
+				 */
+				readonly storedSchemaGenerationOptions: StoredFromViewSchemaGenerationOptions;
+		  }
+
+		// Neither enabledStagedUpgrades nor storedSchemaGenerationOptions provided
+		// Staged schema members remain disabled and initialize/upgradeSchema
+		// generate the most restrictive stored schema compatible with the view schema.
+		| {}
```
CraigMacomber: Expressing this as a union which is `&`ed with ITreeViewConfiguration isn't particularly nice for a few reasons:
1. This kind of typing (as compared to an interface with fields which are unions) looks messy in IntelliSense and errors, can make discoverability harder.
2. Is harder to document. We can't put documents of the 3 different object formats here, and the only place we can document them is at the top level of ITreeViewConfigurationAlpha or on the fields (which does not work well for the empty one)
3. In this specific case, I don't think the way these are separated makes a lot of sense. What we have are three different ways to specify "policy for generating stored schema from the view schema.", but its not clear thats what setting enabledStagedUpgrades or the empty object does.

I think it would be cleaner to have ITreeViewConfigurationAlpha be an interface with a field, `storedSchemaGenerationOptions`. If we want to allow all three options for it, we can make the field optional and allow undefined, `Iterable<SchemaUpgrade>` or StoredFromViewSchemaGenerationOptions, and document that undefined defaults to `restrictiveStoredSchemaGenerationOptions`.

The downside to this (which your approach actually does a good job of solving) is that would make it a bit unclear on the semantics of passing in `Iterable<SchemaUpgrade>` since it would no longer be in a field named `enabledStagedUpgrades`. I see two good ways to solve this in my proposed single `storedSchemaGenerationOptions` field approach:
1. Use your existing `{readonly enabledStagedUpgrades: Iterable<SchemaUpgrade>; }` type in the type union (with undefined and StoredFromViewSchemaGenerationOptions) so its more clear.
2. (I prefer this option, as its more modular/composable, but I think both are ok) Export a function which generates a StoredFromViewSchemaGenerationOptions from an iterable of upgrades (You have one internally I believe, we simple expose it, and then only take in undefined or StoredFromViewSchemaGenerationOptions. I think calling the function `StoredFromViewSchemaGenerationOptions.enabledStagedUpgrades(...iterableHere)` would be nice (create a const object named StoredFromViewSchemaGenerationOptions with that method). Then a caller could go add to their config `storedSchemaGenerationOptions: StoredFromViewSchemaGenerationOptions.enabledStagedUpgrades(myUpgrade)`. Not quite as concise in that case, but makes the degrees of freedom of the API more discoverable, and is quite clear. It leaves clear room in the API for things like `.disabledStagedUpgrades`. We can also export `permissive` and `restrictive` as properties of this object to avoid cluttering the top level namespace too much.

packages/dds/tree/src/simple-tree/core/toStored.ts
```diff
@@ -6,18 +6,19 @@
 import type { SchemaUpgrade } from "./allowedTypes.js";
 
 /**
- * Options for transforming a view simple-schema to a stored simple-schema (See {@link TreeStoredSchema}).
+ * Options for transforming a view simple-schema to a stored simple-schema.
+ * @alpha
```
CraigMacomber, line 11:
```suggestion
 * @input
 * @alpha
```

I think this interface is really just intended as input and apps don't need to read from it, so we fan tag it accordingly to allow more changes in the future.

packages/dds/tree/src/simple-tree/api/configuration.ts
CraigMacomber: I see this was marked as resolved, with no response and I don't see any code change. Unresolving.

jenn-le: none of my commits had been pushed, just fixed :)

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -422,8 +422,7 @@ export interface TreeView<in out TSchema extends ImplicitFieldSchema> extends ID
 	/**
 	 * Description of the current compatibility status between the view schema and stored schema.
 	 * @remarks
-	 * {@link TreeViewEvents.schemaChanged} is fired when the document's stored schema changes,
-	 * which may change this compatibility status.
+	 * {@link TreeViewEvents.schemaChanged} is fired when the compatibility status changes.
```
Josmithr: Nit: the previous wording offered a bit more context, which I think was useful. Might be better to expand this to something like: "is fired when the compatibility status of the document's stored schema changes"

packages/dds/tree/src/simple-tree/core/toStored.ts
```diff
@@ -26,6 +27,92 @@ export interface StoredFromViewSchemaGenerationOptions {
 	includeStagedOptional(upgrade: SchemaUpgrade): boolean;
 }
 
+/**
+ * Provides factory methods for creating {@link (StagedSchemaUpgradePolicy:interface)} instances.
+ * @alpha
+ */
+export interface StagedSchemaUpgradePolicyFactory {
```
Josmithr, line 58: We presumably don't expect users of the package to implement this interface / create their own variables of this type, right? If not, we should probably mark this as `@sealed`.

packages/dds/tree/src/simple-tree/toStoredSchema.ts
```diff
@@ -64,129 +64,77 @@ import { createTreeSchema } from "./treeSchema.js";
...
+ * Excludes all staged schema members.
+ * Prefer using `StagedSchemaUpgradePolicy.restrictive`.
  * @alpha
  */
-export const restrictiveStoredSchemaGenerationOptions: StoredFromViewSchemaGenerationOptions =
-	{
-		includeStaged: () => false,
-		includeStagedOptional: () => false,
-	};
+export const restrictiveStagedUpgradePolicy: StagedSchemaUpgradePolicy =
```
Josmithr: Do we need this separate variable anymore, since we have `StagedSchemaUpgradePolicy.restrictive`?

Josmithr: Same with `permissive` below.

packages/dds/tree/src/simple-tree/core/toStored.ts
```diff
@@ -26,6 +27,92 @@ export interface StoredFromViewSchemaGenerationOptions {
...
+ *
+ * // Use restrictive (default, no staged upgrades):
+ * const options = StagedSchemaUpgradePolicy.restrictive;
+ *
+ * // Use permissive (all staged upgrades, useful for testing):
+ * const options = StagedSchemaUpgradePolicy.permissive;
+ * ```
+ *
+ * @alpha
+ */
```
Josmithr, line 93: Elsewhere (e.g. `TreeNodeApi`), we've taken the approach of putting the documentation on the base interface, and just leveraging `{@inheritDoc}` for the singleton. That way we don't need to duplicate anything, and the docs show up in both places.

packages/dds/tree/src/simple-tree/api/configuration.ts
```diff
@@ -155,6 +157,36 @@ export interface ITreeViewConfiguration<
...
+ * @alpha
+ */
+export interface ITreeViewConfigurationAlpha<
+	TSchema extends ImplicitFieldSchema = ImplicitFieldSchema,
+> extends ITreeViewConfiguration<TSchema> {
+	/**
+	 * Policy for generating stored schema from the view schema.
+	 *
+	 * @remarks
+	 * This policy is fixed at view construction time and cannot be changed afterwards.
```
CraigMacomber: Its fix when creating the ITreeViewConfigurationAlpha, which is clear from it being a readonly field here, so I'm not sure this line adds value. Its not wrong those, so if you think its helpful you can keep it.

.changeset/enable-staged-schema-upgrades.md
```diff
@@ -5,41 +5,51 @@
...
+[`ITreeAlpha.viewWith`](https://fluidframework.com/docs/api/tree/viewabletree-interface#viewwith-methodsignature)
+to select which schema upgrades to enable at runtime.
 
-Use `StagedSchemaUpgradePolicy.enabledStagedUpgrades(...)` with `SchemaUpgrade` objects obtained from schema
-factory APIs such as [`SchemaFactoryBeta.staged`](https://fluidframework.com/docs/api/tree/schemastaticsbeta-interface#staged-propertysignature)
+Use `StagedSchemaUpgradePolicy.enabledStagedUpgrades(...)` with `SchemaUpgrade` objects from
+[`SchemaFactoryBeta.staged`](https://fluidframework.com/docs/api/tree/schemastaticsbeta-interface#staged-propertysignature)
 or [`SchemaFactoryAlpha.stagedOptional`](https://fluidframework.com/docs/api/tree/schemafactoryalpha-class#stagedoptional-property):
 
 ```typescript
```
Josmithr, line 24: Nit: I think a *brief* explanation of what the example is demonstrating would be useful before the code.

packages/dds/tree/src/simple-tree/api/configuration.ts
```diff
@@ -155,6 +157,36 @@ export interface ITreeViewConfiguration<
...
+	 *
+	 * If provided, this policy is used directly for compatibility checks and for
+	 * `initialize` / `upgradeSchema` schema generation.
+	 *
+	 * If omitted or `undefined`, defaults to {@link StagedSchemaUpgradePolicyFactory.restrictive}
+	 * which does not enable any staged schema upgrades.
+	 *
```
CraigMacomber: I think swapping the order here would be better. This way we can imply that when not provided the default is used in these locations. Also I think linking snapshotSchemaCompatibility when ever reasonable is a good idea since its important people discover it when doing anything with schema evolution.


```suggestion
	 * If omitted or `undefined`, defaults to {@link StagedSchemaUpgradePolicyFactory.restrictive}
	 * which does not enable any staged schema upgrades.
	 *
	 * Used when generated stored schema to include in documents in `initialize` / `upgradeSchema`
	 * as well as {@link snapshotSchemaCompatibility} to validate the compatibility of such documents.
```

packages/dds/tree/src/simple-tree/core/toStored.ts
```diff
@@ -29,6 +29,30 @@ export interface StagedSchemaUpgradePolicy {
...
+ *
+ * @remarks
+ * Use the properties and methods on this object to obtain staged-schema generation options
+ * for different scenarios:
+ *
+ * - `StagedSchemaUpgradePolicy.restrictive` — no staged upgrades (default)
+ *
+ * - `StagedSchemaUpgradePolicy.permissive` — all staged upgrades enabled
+ *
+ * - `StagedSchemaUpgradePolicy.enabledStagedUpgrades(...)` — only specific upgrades enabled
```
Josmithr: Nit: any reason not to use `{@link}`s for these? Those will be more likely to be auto-fixed if/when something gets renamed. And the tooling will complain if not.

packages/dds/tree/src/simple-tree/api/configuration.ts
```diff
@@ -252,14 +284,22 @@ export class TreeViewConfigurationAlpha<
 		SimpleNodeSchema<SchemaType.View> & TreeNodeSchema
 	>;
 
-	public constructor(props: ITreeViewConfiguration<TSchema>) {
+	/**
+	 * Stored-schema generation policy for this view, fixed at construction time.
```
CraigMacomber: You should either inherit-doc or link between this and the interface one. Having two versions of this comment with differeing details and no linking doesn't seem ideal.

packages/dds/tree/src/simple-tree/api/configuration.ts
```diff
@@ -252,14 +284,22 @@ export class TreeViewConfigurationAlpha<
...
+	public readonly stagedUpgradePolicy: StagedSchemaUpgradePolicy;
+
+	public constructor(props: ITreeViewConfigurationAlpha<TSchema>) {
 		super(props);
 		const treeSchema = createTreeSchema(this.schema);
 		this.root = treeSchema.root;
 		this.definitions = treeSchema.definitions;
 
+		this.stagedUpgradePolicy =
+			props.stagedUpgradePolicy ?? resolveStoredSchemaGenerationOptions(undefined);
```
CraigMacomber: Your API docs listed a specific default, think I think is prabably equlivelent to this, but I think this should prabably use the default exactly as documents incase this or it change at all.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -437,6 +437,44 @@ export interface TreeView<in out TSchema extends ImplicitFieldSchema> extends ID
...
 	 * It is an error to call this when {@link SchemaCompatibilityStatus.canUpgrade} is false.
 	 * {@link SchemaCompatibilityStatus.canUpgrade} being true does not mean that an upgrade is required, nor that an upgrade will have any effect.
+	 *
+	 * When using {@link TreeViewConfigurationAlpha} with a {@link ITreeViewConfigurationAlpha.stagedUpgradePolicy},
+	 * staged schema upgrades matching the configured policy are included in the target stored schema.
+	 * Once a staged schema upgrade has been enabled in a document's stored schema, loading that document
+	 * with a view that does not include equivalent staged members in its construction-time policy will cause
+	 * `upgradeSchema` to throw a `UsageError` because the requested target would narrow the stored schema.
+	 *
+	 * @example Enabling a staged allowed type for documents, selected by a feature flag:
```
Josmithr: Nit: theoretically (per TSDoc spec) text on the same line as the example tag is intended to be treated as the example "title". So I think it's reasonable to omit the trailing `:` here.
```suggestion
	 * @example Enabling a staged allowed type for documents, selected by a feature flag
```

packages/dds/tree/src/test/simple-tree/api/stagedSchemaUpgrade.spec.ts
```diff
@@ -18,30 +18,82 @@ import {
...
 
 // Some documentation links to this file on GitHub: renaming it may break those links.
 
+describe("runtime schema upgrade API", () => {
+	const stagedString = SchemaFactoryAlpha.staged(SchemaFactoryAlpha.string);
+	const stringUpgrade = stagedString.metadata.stagedSchemaUpgrade;
+	assert(stringUpgrade !== undefined);
+
+	// Schema B: number or string (string is staged)
+	const schemaB = SchemaFactoryAlpha.optional(
```
Josmithr: Nit: "schemaB" isn't very descriptive. Would it be worth being a bit more verbose here? E.g. "schemaWithStagedType"?

Josmithr: Oh, I see below we have some tests with `schemaA` defined. If we wanted to express these in terms of "before" and "after", that might make sense (but in that case, we should probably define the "before" case here next to the "after" case).

Alternatively, "withStaged" and "withoutStaged" might make sense.

Josmithr: But this is all nitpick. Take it or leave it 🙂 (especially since renaming would complicate the associated tree variable naming and such)

Josmithr (approved):
Left a handful more minor suggestions. But otherwise this looks great to me. Psyched to have this functionality available!

packages/dds/tree/src/simple-tree/api/configuration.ts
jenn-le: Nice catch, this was left over from before the api change :)

Josmithr (approved):

PR #27546, opened 2026-06-13
improvement(client): fix `exactOptionalPropertyTypes` for telemetry-utils

jason-ha:
`ChildLogger` (which is almost always the logger anything accesses) supports `minLogLevel` with a getter that may return `undefined`. Update `ITelemetryBaseLogger` to allow `minLogLevel` to be `undefined`.

This is a type breaking change for anyone not using `exactOptionalPropertyTypes:false` as recommended in package READMEs, but not a runtime breaking change.

`telemetry-utils` corrections:
- use `...(cond ? {} { prop: value })` to only define properties with defined values.
- make optional private properties required allowing `| undefined`.
- in logger helper functions allow explicit `undefined`
  - internal `ChildLogger.create` clarified to expect to arguments though `undefined` is allowed.

### DIFF
--- a/.changeset/yummy-moons-train.md
+++ b/.changeset/yummy-moons-train.md
@@ -6,5 +6,5 @@
 ITelemetryBaseLogger.minLogLevel may be undefined
 
 Typing for `ITelemetryBaseLogger.minLogLevel` is updated to reflect that in some implementations `minLogLevel` is present but evaluates to `undefined`.
-When building with `excactOptionalPropertyTypes:false` as suggested in compatibility requirements, there is no apparent type change.
+When building with `excactOptionalPropertyTypes:false` as suggested in [compatibility requirements](https://github.com/microsoft/FluidFramework/blob/68732d93a6cc8be2df966b9bb40f58bdd9fad69b/packages/common/core-interfaces/README.md#supported-tools), there is no apparent type change.
 If a type error is experienced, make sure to check for `undefined` or use `?? LogLevel.info` when reading.
--- a/.changeset/yummy-moons-train.md
+++ b/.changeset/yummy-moons-train.md
@@ -1,0 +1,9 @@
+---
+"@fluidframework/core-interfaces": minor
+"__section": breaking
+---
+ITelemetryBaseLogger.minLogLevel may be undefined
+
+Typing for `ITelemetryBaseLogger.minLogLevel` is updated to reflect that in some implementations `minLogLevel` is present but evaluates to `undefined`.
+When building with `excactOptionalPropertyTypes:false` as specified in compatibility requirements, there is not apparent type change.
+If a type error is experienced, make sure to check for `undefined` or use `?? LogLevel.info` when reading.
--- a/.changeset/yummy-moons-train.md
+++ b/.changeset/yummy-moons-train.md
@@ -1,9 +1,10 @@
 ---
+"@fluidframework/azure-client": minor
 "@fluidframework/core-interfaces": minor
 "__section": breaking
 ---
 ITelemetryBaseLogger.minLogLevel may be undefined
 
 Typing for `ITelemetryBaseLogger.minLogLevel` is updated to reflect that in some implementations `minLogLevel` is present but evaluates to `undefined`.
-When building with `excactOptionalPropertyTypes:false` as specified in compatibility requirements, there is not apparent type change.
+When building with `excactOptionalPropertyTypes:false` as suggested in compatibility requirements, there is no apparent type change.
 If a type error is experienced, make sure to check for `undefined` or use `?? LogLevel.info` when reading.
--- a/packages/utils/telemetry-utils/src/logger.ts
+++ b/packages/utils/telemetry-utils/src/logger.ts
@@ -807,7 +807,7 @@ export class PerformanceEvent {
 
 	private event: ITelemetryGenericEventExt | undefined;
 	private readonly startTime = performanceNow();
-	private startMark?: string;
+	private startMark: string | undefined;
 
 	private constructor(
 		private readonly logger: TelemetryLoggerExt,
@@ -862,7 +862,7 @@ export class PerformanceEvent {
 			const endMark = `${this.event.eventName}-end`;
 			window.performance.mark(endMark);
 			window.performance.measure(`${this.event.eventName}`, this.startMark, endMark);
-			delete this.startMark;
+			this.startMark = undefined;
 		}
 	}
 
--- a/packages/utils/telemetry-utils/src/logger.ts
+++ b/packages/utils/telemetry-utils/src/logger.ts
@@ -805,7 +805,7 @@ export class PerformanceEvent {
 		return performanceNow() - this.startTime;
 	}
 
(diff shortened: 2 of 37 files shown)


.changeset/yummy-moons-train.md
```diff
@@ -0,0 +1,9 @@
+---
+"@fluidframework/core-interfaces": minor
+"__section": breaking
+---
+ITelemetryBaseLogger.minLogLevel may be undefined
+
+Typing for `ITelemetryBaseLogger.minLogLevel` is updated to reflect that in some implementations `minLogLevel` is present but evaluates to `undefined`.
+When building with `excactOptionalPropertyTypes:false` as specified in compatibility requirements, there is not apparent type change.
```
Josmithr: Nit: shall we link to the compat requirements documentation?

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -846,23 +846,23 @@ export class PerformanceEvent {
 		this.performanceEndMark();
 
 		// To prevent the event from being reported again later
-		this.event = undefined;
+		delete this.event;
```
ChumpChief: I'd typically avoid property deletion unless existence of the property is semantic separately from `undefined` (which it doesn't seem to be here).  And in most cases I'd recommend against existence being semantic anyway.

As a similarly quick solution, maybe instead consider making the event member non-optional but nullable to narrow the typing.

```ts
private event: ITelemetryGenericEventExt | undefined;
```

Probably the preferable solution would be to make the event prop immutable and instead have a boolean `alreadyReported` flag, but not necessary in the short term.

.changeset/yummy-moons-train.md
```diff
@@ -0,0 +1,9 @@
+---
+"@fluidframework/core-interfaces": minor
```
Josmithr, line 3: Should we also list the other packages that re-export `ITelemetryBaseLogger`?

Josmithr: Which I guess is just azure-client

.changeset/yummy-moons-train.md
jason-ha: I don't think so. This space is weird. I am really hoping we clean more of this up before 3.0 and can make a stronger recommendation to consumers. As it is, I think they can really do either. Currently, we imply `excactOptionalPropertyTypes:false` and but note how to work with FF when specifying `true`.
If we want to change ClientRequirements text, I would do that separately from this PR.

packages/utils/telemetry-utils/src/logger.ts
jason-ha: For an internal private member, I don't think either choice matters so much. I think in a pure performance optimized world, I would guess deletion is better.
For this `exactOptionalPropertyTypes` setting I think the best solve for us is to compile every production package both way and see that there are no errors. That means we handle input from consumers not specifying true and passing an unfortunate `undefined` as well as never hand anything out ourselves that sets `undefined` for something optional.
The only time existence should be semantic, IMO, is when working with a record.

packages/utils/telemetry-utils/src/logger.ts
```diff
@@ -862,7 +862,7 @@ export class PerformanceEvent {
 			const endMark = `${this.event.eventName}-end`;
 			window.performance.mark(endMark);
 			window.performance.measure(`${this.event.eventName}`, this.startMark, endMark);
-			this.startMark = undefined;
+			delete this.startMark;
```
ChumpChief: Similar recommendation here EDIT: actually I don't think there's any point in deleting or unsetting here, might as well just remove this line.

jason-ha: I'm not super keen on potential behavior differences for these exactOptionalPropertyType changes. The one difference I see from not clearing is debuggability to distinquish cancel from normal end. I don't know if that is interesting.
If we were to do cleanup, then I think there are some other things that could be done.
So here, I'll restore `= undefined`.

.changeset/yummy-moons-train.md
Josmithr: In that case, maybe best to omit the "as specified in compatibility requirements" part? If a developer reads that, and it's relevant to them, they will likely want to look for that documentation. If we don't have something for them, probably best not to imply we do.

Josmithr (approved):
Docs and devtools changes look good, with the exception of one remaining nitpick.

.changeset/yummy-moons-train.md
jason-ha: It doesn't show in this thread, but I slightly altered the wording. It now reads "as _suggested_ in compatibility requirements". The actual compatibility text says:
>     -   `exactOptionalPropertyTypes` is currently not fully supported.
>        If used, narrowing members of Fluid Framework types types using `in`, `Reflect.has`, `Object.hasOwn` or `Object.prototype.hasOwnProperty` should be avoided as they may incorrectly exclude `undefined` from the possible values in some cases.

Is there still a nit?

Josmithr: My point is that we are referring to our "compat requirements", but you don't want to link to them. If we don't want to link to them, then presumably we don't really want folks looking at them? And if that's the case, I would argue that we shouldn't refer to them at all. I.e. just remove the "as suggested in compatibility requirements" part.

ChumpChief (approved):
Thank you for the updates!

.changeset/yummy-moons-train.md
jason-ha: I don't know how my brain ended up with the wrong idea of the request. I thought I read should we *edit* the compat requirements. Not should we *link* them. Yeah, we can link them. :)

.changeset/yummy-moons-train.md
```diff
@@ -0,0 +1,10 @@
+---
+"@fluidframework/azure-client": minor
+"@fluidframework/core-interfaces": minor
+"__section": breaking
+---
+ITelemetryBaseLogger.minLogLevel may be undefined
+
+Typing for `ITelemetryBaseLogger.minLogLevel` is updated to reflect that in some implementations `minLogLevel` is present but evaluates to `undefined`.
+When building with `excactOptionalPropertyTypes:false` as suggested in [compatibility requirements](https://github.com/microsoft/FluidFramework/blob/68732d93a6cc8be2df966b9bb40f58bdd9fad69b/packages/common/core-interfaces/README.md#supported-tools), there is no apparent type change.
```
Josmithr, line 9: Should this be
```suggestion
When building with `excactOptionalPropertyTypes:false` as suggested in [compatibility requirements](https://github.com/microsoft/FluidFramework/tree/main/packages/common/core-interfaces/README.md#supported-tools), there is no apparent type change.
```
?

jason-ha: I don't want to use `tree/main` because this will change. I hope it changes for 3.0. So, using SHA for durability.

jzaffiro (approved):

.changeset/yummy-moons-train.md
Josmithr: Fair

PR #27547, opened 2026-06-15
test: Cross-client test matrix follow ups

scottn12:
## Description

Follow-up to the review on #27233. Aligns the cross-client compat test matrix and checkpoint docs with the review feedback:

- **Cross-client matrix:** `genCrossClientCompatConfig` now also pairs the current build against the lower bound of its *own* compatibility checkpoint (both directions), so two in-window clients that share the current checkpoint are covered. The pair is skipped when the build already resolves to that lower-bound version (self-test). `updateCompatVersions.ts` installs that version so it is available to the matrix.
- **Docs/comments:** the update-script trigger note now covers *any* change to `checkpoints.ts` (not just designations); the two duplicate "designate a checkpoint" steps in the dev guide are collapsed into one; and all references use `pnpm -r --filter @fluid-private/test-version-utils run update-compat-versions` so the command runs from any directory.


CrossClientCompatibility.md
```diff
@@ -30,9 +30,9 @@ This document explains:
 <!-- prettier-ignore -->
 | Term | Definition |
 | --- | --- |
-| **Compatibility Checkpoint Release** | The first Fluid release in a checkpoint range (e.g., `2.100.0` for CC-4). |
-| **Compatibility Checkpoint Range** | The semver range of Fluid releases that are part of a checkpoint (e.g., `>=2.100.0 <2.130.0` for CC-4). All releases in the range share the same cross-client compatibility guarantees as the first release of the range. |
-| **Checkpoint N** | `CC-N`; any individual compatibility checkpoint range. `Checkpoint N-1` `CC-(N-1)`, the compatibility checkpoint range before `N`, and so on for `N-2`, `N-3`, etc. |
+| **Compatibility Checkpoint Release** | The first Fluid release in a checkpoint range (e.g., `2.100.0` for CC#4). |
+| **Compatibility Checkpoint Range** | The semver range of Fluid releases that are part of a checkpoint (e.g., `>=2.100.0 <2.130.0` for CC#4). All releases in the range share the same cross-client compatibility guarantees as the first release of the range. |
```
jason-ha: `2.100.0` and `>=2.100.0 <2.130.0` are out of date. Not the first/range of CC#4.
May also want to put CC#4 in inside backticks. At least in github diff it is underlining them like a hyperlink.

CompatibilityCheckpoints.md
```diff
@@ -24,26 +24,26 @@ checkpoints are designated.
...
-| CC-5 (TBD) | `>=3.0.0 <4.0.0`(estimated) | ~2026-07-06 | CC-2, CC-3, CC-4, CC-5, CC-6, CC-7 | `>=2.0.0 <6.0.0`(estimated) |
-| CC-6 (TBD) | `>=4.0.0 <5.0.0`(estimated) | ~2027-01-06 | CC-3, CC-4, CC-5, CC-6, CC-7 | `>=2.40.0 <6.0.0`(estimated) |
-| CC-7 (TBD) | `>=5.0.0 <6.0.0`(estimated) | ~2027-07-06 | CC-4, CC-5, CC-6, CC-7 | `>=2.80.0 <6.0.0`(estimated) |
+| CC#1 | `>=1.4.0 <2.0.0 \| 2.0.0-internal* \| 2.0.0-rc*` | 2024-04-09 | CC#1, CC#2, CC#3, CC#4 | `>=1.4.0 <2.101.0`` \| 2.0.0-internal* \| 2.0.0-rc*` |
+| CC#2 | `>=2.0.0 <2.40.0` | 2024-06-26 | CC#1, CC#2, CC#3, CC#4, CC#5 | `>=1.4.0 <4.0.0`(estimated)` \| 2.0.0-internal* \| 2.0.0-rc*` |
+| CC#3 | `>=2.40.0 <2.80.0` | 2025-05-12 | CC#1, CC#2, CC#3, CC#4, CC#5, CC#6 | `>=1.4.0 <5.0.0`(estimated)` \| 2.0.0-internal* \| 2.0.0-rc*` |
+| CC#4 | `>=2.80.0 <2.101.0` | 2026-01-06 | CC#1, CC#2, CC#3, CC#4, CC#5, CC#6, CC#7 | `>=1.4.0 <6.0.0`(estimated)` \| 2.0.0-internal* \| 2.0.0-rc*` |
+| CC#5 (TBD) | `>=3.0.0 <4.0.0`(estimated) | ~2026-07-06 | CC#2, CC#3, CC#4, CC#5, CC#6, CC#7 | `>=2.0.0 <6.0.0`(estimated) |
```
jason-ha: Predates the automatic generation, but also odd that it wasn't updated.
Why does CC#4 include 2.100 but not any of the more recent versions?
Seems like it should have upper-bound of <3.0.0 (the first of future) and `(estimated)` qualifier. Or at least it should be able to include all known current releases (seems harder).

packages/test/test-version-utils/src/compatConfig.ts
```diff
@@ -341,6 +341,15 @@ export const genCrossClientCompatConfig = (): CompatConfig[] => {
...
 	const deltaVersions: Map<string, string> = new Map();
 	const current = getCurrentCheckpoint(pkgVersion);
+
+	// Pair the current build against the lower bound of its own checkpoint, so two
+	// in-window clients that share the current checkpoint are covered. Skipped when the
+	// build already resolves to that lower-bound version (that combo would be a self-test).
+	const currentCheckpointVersion = resolveRangeViaManifest(checkpointResolutionRange(current));
+	if (currentCheckpointVersion !== currentVersion) {
+		deltaVersions.set(currentCheckpointVersion, current.name);
+	}
```
jason-ha, line 351: What should/does happen with a patch release for the first version in a checkpoint?
Should it test against the prior patch? Should it explicitly not?

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -113,21 +113,21 @@ export const checkpoints: readonly Checkpoint[] = [
  */
 const futureCheckpoints: readonly DocumentedCheckpoint[] = [
 	{
-		name: "CC-5",
+		name: "CC#5",
 		index: 5,
 		lowerBoundVersion: "3.0.0",
 		startDate: "2026-07-06",
```
jason-ha: Let's fix this date. Estimated release date is 2026-08-24.
And I am guessing that next ones are 2027-04-19 and 2027-12-06. (Those are 6-9 months after prior that align to multiple of 10 minor.)

packages/test/test-version-utils/src/checkpoints.ts
```diff
@@ -248,14 +248,14 @@ const tableEndMarker = "<!-- GENERATED-TABLE-END -->";
...
-	`<!-- To modify this table, edit \`${designatedSourceRelativePath}\` then run \`pnpm --filter @fluid-private/test-version-utils run update-compat-versions\` -->`,
+	`<!-- To modify this table, edit \`${designatedSourceRelativePath}\` then run \`pnpm -r --filter @fluid-private/test-version-utils run update-compat-versions\` -->`,
 ].join("\n");
 
 /**
  * Explicit upper bounds for designated checkpoints whose range does not end at
  * the next checkpoint's `lowerBoundVersion`.
  */
-const designatedClosingVersions: Readonly<Record<string, string>> = { "CC-4": "2.101.0" };
+const designatedClosingVersions: Readonly<Record<string, string>> = { "CC#4": "2.101.0" };
```
jason-ha: Commented where I saw the funny result in the table. Why do we want this? It means there will be a gap in the table. If this is kept it seems like it will require manual maintenance that I don't see documented.
Also, it looks like the type for `closingVersion` property is incorrect or logic needs updated. Type should be `closingVersion?: string | undefined` OR line 265 needs updated:
```typescript
const documentedCheckpoints: readonly DocumentedCheckpoint[] = [
	...checkpoints.map(
		(c): DocumentedCheckpoint => ({
			...c,
			status: "designated",
			...(designatedClosingVersions[c.name] === undefined ? {} : { closingVersion: designatedClosingVersions[c.name] }),
		}),
```

CrossClientCompatibility.md
scottn12: Updated in latest

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Updated in latest

packages/test/test-version-utils/src/checkpoints.ts
scottn12: Updated in latest

CompatibilityCheckpoints.md
scottn12: Yeah good call - updated in latest to make the upper bound extend to the next estimated CC's lower bound.

packages/test/test-version-utils/src/compatConfig.ts
scottn12: Each checkpoint resolves the lower bound to the latest published patch of its first minor, so we always test against the newest patch. I think that is correct but let me know if you think differently

jason-ha (approved):

packages/test/test-version-utils/src/compatConfig.ts
jason-ha: I think that is fine. I don't know that we have any process in place to refresh checked versions when a patch is released. Probably it is fine to test against any of the patches. But latest is best for security.

PR #27527, opened 2026-06-10
Add PR bundle-size comments workflow

ChumpChief:
## Description

Adds a new GitHub Actions workflow (`pr-bundle-size-comments.yml`) that listens for ADO `check_run` events on the bundle-publishing pipelines, runs a per-PR bundle-size comparison via flub, and (eventually) posts the result as a sticky PR comment.

The workflow ships in **bake-in mode**: the four sticky-comment post steps are commented out. After merge, the workflow runs end-to-end on every PR-build event, logs the would-be comment body in the action log, and writes nothing back to the PR. Turning on actual comment posting is a follow-up, tracked in [AB#75349](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/75349).

Supporting changes:

- New flub command `flub report comparePipelineBundleArtifacts` that drives the actual comparison: fetches both sides' bundle artifacts from ADO, runs `compareJsonReportsByPackage`, prints JSON or a human-readable summary.
- `BuildMatch` refactor in `library/azureDevops/getArtifactForCommit.ts` so callers explicitly state whether they're looking up a build by `commit` (main-side) or `prHead` (PR-side, via `triggerInfo['pr.sourceSha']` since `sourceVersion` is the GitHub-generated test-merge SHA).
- DRY: introduce `fluidframeworkAdoOrgUrl` in `library/azureDevops/constants.ts` and use it from both bundleSize commands and `codeCoverage.ts` (which previously had the literal URL inlined twice).
- Opportunistic cleanup of stale `# release notes` comments in other `.github/workflows/*` files. The comments claimed versions that no longer matched the pinned SHAs, which had been bumped without the URLs being updated.

Resolves [AB#74457](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/74457).

## Reviewer Guidance

The review process is outlined on [this wiki page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

Start at `.github/workflows/pr-bundle-size-comments.yml`. The header comment explains the trigger model (check_run created/completed lifecycle), the concurrency strategy (within-SHA dedupe + cross-SHA staleness handled separately in the identify-* jobs), and the bake-in posture.

PR #27549, opened 2026-06-15
Polish pr-bundle-size-comments workflow with structured failure stickies

ChumpChief:
[AB#75772](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/75772)

## Description

Adds structured failure-mode dispatch to the `pr-bundle-size-comments` workflow that landed in #27527, and fixes several issues observed in production: a concurrency-cancellation bug that prevented `acknowledge-build` from ever running to completion, a 100-build search horizon that lost old PRs' merge-base baselines, stickies left stale forever when a build subsequently failed or was cancelled, and the workflow firing for `internal` ADO project check_runs we can't process. Plus diagnostic polish.

### Concurrency fix

The concurrency group key was just the check_run head SHA, so every ADO sub-check that fires on the same commit (`Build - client packages (Build Stage Build)`, `Build - client packages (🔒 SDLSources ...)`, etc.) was cancelling any in-flight workflow run for that SHA via `cancel-in-progress` — even though those events are noise we filter out at the job `if:`. Net effect: `acknowledge-build` never ran to completion in practice. Including the check name in the group key isolates each event stream while preserving the intended completed-cancels-pending semantics for matching events.

### Search horizon bumped

`recentBuildsToFetch` 100 → 500. ADO has no query-by-SHA API; bumped the client-side filter window from ~3 weeks to ~3 months of main-branch CI at the observed ~5 builds/day.

### Structured failure-mode dispatch

`findBuildId` and `getArtifactForCommit` now return discriminated unions — `{ kind: "completed" | "no-build" | "in-progress" | "all-failed" | "no-id" }` — instead of throwing string-formatted Errors that downstream consumers would have to grep. `flub report comparePipelineBundleArtifacts --json` emits a matching `{ kind, side: "base" | "head" }` payload for failures (with `no-analyzer-jsons` added at the command level on top of the library kinds).

The workflow's render step dispatches on `kind`:
- `completed` → the per-package comparison body, including 0-delta lines for full inventory.
- `in-progress` → a matter-of-fact "Pending — …" line that mirrors `acknowledge-build`'s pending sticky. Normal "wait for the build" path, not an error.
- everything else → "⚠️ Comparison unavailable." plus a friendly per-(side, kind) paragraph telling the user what's happening and whether they need to act (merge a recent main commit, fix the PR build, …).

### Deferred-baseline / failed-build auto-resolution

Both `identify-*` jobs no longer gate on `conclusion == 'success'`. Any completion re-triggers the appropriate path: flub's structured kind dispatches the right body and the sticky updates from "Pending — …" to (e.g.) "The baseline CI build for the merge-base commit failed — likely a flaky producer build. Try merging a recent main commit into this PR, or re-queue the baseline pipeline manually."

### Project scoping

Both the `public` and `internal` ADO projects publish check_runs with the same names (e.g. `Build - client packages`). The original `if:` filters matched on name only, so the workflow was also firing for internal-project events — which we can't process (no anonymous read access in the workflow context) and which don't produce the artifacts we consume. Add `contains(check_run.details_url, '<public-project-guid>')` to all three triggering jobs so we only react to public-project check_runs. There's no structured `project` field in the check_run payload, so a details_url GUID match is the only stable distinguisher.

### Diagnostic improvements

- **`Affected PRs (N):` summary at the end of both identify-* jobs** so the action log says at a glance whether the matrix was empty or which PRs it pointed at.
- **`Matched PR #X head=... base=... mergeBase=...` summary in `acknowledge-build`'s find_pr step** for symmetry — the happy path was previously silent.
- **Cleaner matrix display name on the `compare` job**: `compare (PR #27546)` instead of the auto-generated `compare (27546, 0412ca768f..., main, fb992c35c4...)`.
- **Collapsible log groups for the JSON output and rendered markdown bodies** — the bulk content is folded by default; the "would post" header line stays visible above the fold so it's still clear what's there.

## Reviewer Guidance

The structured-failure refactor is the main load-bearing piece — start at `library/azureDevops/getArtifactForCommit.ts` (where the discriminated union and `describeArtifactFailure` helper live) and follow it through `commands/report/comparePipelineBundleArtifacts.ts` (the `fetchSide`/`handleFailure` dispatch) and `.github/workflows/pr-bundle-size-comments.yml` (the `renderComparison`/`renderFailure` split + the `identify-*` filter widening).

### DIFF
--- a/.github/workflows/pr-bundle-size-comments.yml
+++ b/.github/workflows/pr-bundle-size-comments.yml
@@ -41,13 +41,23 @@ permissions:
   pull-requests: read
 
 jobs:
+  # NOTE on the `contains(... '10f9b53e-c7fd-4538-9fff-13cd088a436c')` check that appears in each of the
+  # triggering jobs below: that GUID is the `public` ADO project under `dev.azure.com/fluidframework`. The
+  # pipelines whose check_runs we react to all live there. We need the scope because both the `public` and
+  # `internal` projects publish check_runs with the same name (e.g. `Build - client packages`); without it
+  # the workflow also fires on `internal`'s check_runs, which we can't process (no anonymous read access)
+  # and don't produce the artifacts we consume. The GUID appears in `check_run.details_url`; there's no
+  # structured `project` field in the event payload, so `contains(details_url, '<guid>')` is the only
+  # stable distinguisher. The project lives at https://dev.azure.com/fluidframework/public.
+
   # Posts the initial "build pending" sticky comment when ADO queues the `Build - client packages` check
   # for a PR commit. We only measure bundle size on client packages today, so PRs that don't trigger that
   # check (server-only, docs-only) receive no sticky.
   acknowledge-build:
     if: >-
       github.event.action == 'created' &&
-      github.event.check_run.name == 'Build - client packages'
+      github.event.check_run.name == 'Build - client packages' &&
+      contains(github.event.check_run.details_url, '10f9b53e-c7fd-4538-9fff-13cd088a436c')
     runs-on: ubuntu-latest
     # Uncomment to enable comment posting after bake-in (tracked: AB#75349).
     # permissions:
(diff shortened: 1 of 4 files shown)


MarioJGMsoft (approved):
LGTM! I'm a big fan of the `side`, `kind` system being introduced for error logging

MarioJGMsoft (approved):

PR #27551, opened 2026-06-16
build(client-drivers): enable `exactOptionalPropertyTypes` in most packages

jason-ha:
`odsp-driver` is notably left with `exactOptionalPropertyTypes` disabled as 50 errors are flagged when enabled.

common corrections:
- use `...(cond ? {} { prop: value })` to only define properties with defined values.
- delete optional properties instead of setting to `undefined`.

special cases:
- `packages/drivers/driver-web-cache`: disable back compat for `FluidCacheConfig` as `ITelemetryBaseLogger.minLogLevel` recently changed and that change is not detecatble via `exactOptionalPropertyTypes` enablement.
- `packages/drivers/odsp-driver/src/zipItDataRepresentationUtils.ts`: explicit `false` instead of possibly `undefined`.
- `packages/drivers/odsp-driver/src/contractsPublic.ts`: updated to explicitly allow `undefined` for optional properties as is done by `decodeOdspFluidDataStoreLocator`.
- `packages/drivers/routerlicious-driver/src/documentServiceFactory.ts`: capture optional property instead of getter that might have to evaluate to `undefined`.
- `packages/test/test-drivers`: only set optional members when defined by caller

### DIFF
--- a/packages/drivers/odsp-driver/src/test/odspDriverUrlResolverForShareLink.spec.ts
+++ b/packages/drivers/odsp-driver/src/test/odspDriverUrlResolverForShareLink.spec.ts
@@ -125,12 +125,8 @@ describe("Tests for OdspDriverUrlResolverForShareLink resolver", () => {
 					...(resolvedUrl1.fileVersion === undefined
 						? {}
 						: { fileVersion: resolvedUrl1.fileVersion }),
-					...(resolvedUrl1.context === undefined
-						? {}
-						: { context: resolvedUrl1.context }),
-					...(resolvedUrl1.appName === undefined
-						? {}
-						: { appName: resolvedUrl1.appName }),
+					...(resolvedUrl1.context === undefined ? {} : { context: resolvedUrl1.context }),
+					...(resolvedUrl1.appName === undefined ? {} : { appName: resolvedUrl1.appName }),
 				});
 				const resolvedUrl2 = await resolver.resolve({ url });
 				assert.strictEqual(resolvedUrl2.driveId, driveId, "Drive id should be equal");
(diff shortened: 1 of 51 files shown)


packages/drivers/odsp-driver/src/test/odspDriverUrlResolverForShareLink.spec.ts
```diff
@@ -117,7 +117,17 @@ describe("Tests for OdspDriverUrlResolverForShareLink resolver", () => {
 		it(`resolve - Should resolve odsp driver url correctly, hasVersion: ${urlWithNav.hasVersion}, hasContext: ${urlWithNav.hasContext}`, async () => {
 			const runTest = async (resolver: OdspDriverUrlResolverForShareLink): Promise<void> => {
 				const resolvedUrl1 = await resolver.resolve({ url: urlWithNav.url });
-				const url: string = createOdspUrl({ ...resolvedUrl1, dataStorePath });
+				const url: string = createOdspUrl({
+					siteUrl: resolvedUrl1.siteUrl,
```
Josmithr: Should we restore the leading `...resolvedUrl1` here to ensure we continue to spread other properties as before?

Josmithr (approved):

packages/drivers/odsp-driver/src/test/odspDriverUrlResolverForShareLink.spec.ts
jason-ha: The problem is the original spread. In the source type property `fileVersion` is allowed to exist with `undefined`. So, the naive spread would create that situation. We need a per property build like this to account for the difference in the types.
Alternative would be to update `OdspFluidDataStoreLocator` to explicitly allow `undefined` `fileVersion`. I am pretty sure we are emitting `undefined` for `OdspFluidDataStoreLocator.fileVersion` today; so maybe that is a better change.

Josmithr (approved):

PR #27552, opened 2026-06-17
Enable bundle-size sticky comments and add notable-changes summary

ChumpChief:
## Description

Flips the `pr-bundle-size-comments` workflow out of bake-in mode and enriches the rendered sticky body. After #27549 the workflow had been logging would-be sticky content into the action log for observation; this PR enables the actual posting (`marocchino/sticky-pull-request-comment`) on both `acknowledge-build` (the "Pending — …" body) and `compare` (the results / failure body). Resolves [AB#75349](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/75349) and [AB#56981](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/56981).

### Notable changes summary

Above the per-bundle inventory, the comment now leads with a "Notable changes" section that lifts only the bundles a reviewer would actually want to see at a glance:

- ➕ for added bundles
- ➖ for removed bundles
- 🔴 for changed bundles whose parsed delta is ≥ 500 bytes
- 🟢 for changed bundles whose parsed delta is ≤ −500 bytes

Each bundle's full line is rendered once and reused in both the notable summary and the full inventory, so the two never diverge. When nothing qualifies, the section reads "No bundles changed by ≥ 500 bytes parsed." (kept so reviewers know the analysis ran successfully, just had nothing dramatic to lift).

### Collapsible per-bundle inventory

The full per-package inventory (every bundle, including 0-delta lines) is wrapped in a `<details>`/`<summary>` block. The top of the sticky stays compact (just the header, base/head SHAs, and the notable summary); reviewers click "Per-bundle deltas" to expand the rest.

### Display-name fix on the `compare` job

The previous `name: compare (PR #${{ matrix.pr.number }})` was unquoted, so YAML treated the space-then-`#` as a comment and the stored job name was the truncated `compare (PR`. Quote the value and wrap in a conditional so the matrix-expanded case renders `compare (PR #27548)` and the empty-matrix case (acknowledge-build no-op) falls back to plain `compare`. Bug was visible in pre-merge runs of #27549.

## Reviewer Guidance

The shape of the sticky body is defined by `renderComparison` / `renderFailure` in `.github/workflows/pr-bundle-size-comments.yml`. The structured failure-mode dispatch from #27549 means flipping comment posting on is fully wired up — the failure-mode sticky bodies are now what real users will see, not just lines in the action log. End-to-end failure-rendering was validated on the canary PR (#27548): the deliberate-build-break commit there produced the all-failed body via the workflow's render path, observable in run 27657241881.

### DIFF
--- a/.github/workflows/pr-bundle-size-comments.yml
+++ b/.github/workflows/pr-bundle-size-comments.yml
@@ -395,30 +395,53 @@ jobs:
             const fs = require("fs");
             const data = JSON.parse(fs.readFileSync("bundle-comparison.json", "utf8"));
 
-            // Render the happy-path comparison body — per-package sections, or
-            // "No bundle-size changes." if everything is unchanged.
+            // Render the happy-path body: a "Notable changes" summary (added,
+            // removed, or changed-with-parsed-delta ≥ NOTABLE_THRESHOLD) above
+            // the collapsed full per-bundle inventory.
             function renderComparison(comparison) {
+              const NOTABLE_THRESHOLD = 500;
+
               const fmtDelta = d => (d > 0 ? `+${d}` : `${d}`);
 
-              // Render one bundle's diff line — always emitted, even for
-              // unchanged bundles, so the body is a complete inventory.
-              function renderBundleLine(bundle, cmp) {
+              // Indicator emoji and notability for one bundle. ➕/➖ for added/removed,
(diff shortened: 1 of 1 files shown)


MarioJGMsoft (approved):
LGTM

PR #27554, opened 2026-06-17
Bump concurrently to 10.0.3 to resolve shell-quote vulnerability

frankmueller-msft:
## Summary

Bumps `concurrently` from `^9.2.1` to `^10.0.3` across all workspaces. `concurrently@10.0.3` depends on `shell-quote@1.8.4`, which resolves the command injection vulnerability in `shell-quote@1.8.3`.

## Approach

Per feedback from @Abe27342 and @alexvy86, this bumps to `concurrently` 10.x rather than patching 9.x. The breaking changes in v10 are minimal and safe for us:
- Dropped support for Node <22 (we already require Node 22)
- Converted to ESM (we only use `concurrently` as a CLI tool)

Changes:
- Updated `concurrently` specifier from `^9.2.1` to `^10.0.3` in all package.json files
- Removed the `shell-quote: ^1.8.4` override from root `pnpm-workspace.yaml` (no longer needed)
- Regenerated all 11 lockfiles

## CG Alert

Resolves: https://dev.azure.com/fluidframework/internal/_componentGovernance/17385/alert/14301083


alexvy86 (commented):
I see some inconsistencies (not all lockfiles list the override) but that might be moot because now that the new patch version is updated in the lockfiles, it's very probable the overrides in all the pnpm-workspace.yaml files are now unnecessary. I'd be a bit surprised if something pins its shell-quote dep specifically to 1.8.3 instead of a range that will still be happy with 1.8.4. I'd prefer to keep the list of overrides clean, with only the ones that we *must* keep so the lockfile doesn't revert to a bad version.

frankmueller-msft:
Good call — I verified that `concurrently@9.2.1` declares an **exact** dependency on `shell-quote`: `"1.8.3"` (no caret/tilde). Without the override, `pnpm install` will always resolve to the vulnerable version. Since `concurrently` is present in all 11 workspaces, the override is unfortunately required everywhere.

I've updated the comments in each `pnpm-workspace.yaml` to make this clear, and added a section to the PR description explaining it.

alexvy86:
> concurrently@9.2.1 declares an exact dependency on shell-quote: "1.8.3"

Ah, unfortunately true. Then my concern about the lockfiles is still valid, it seems like maybe the changes in some lockfiles were just pattern-matched and applied without actually running `pnpm install` because they don't list `shell-quote` in their overrides section. The one for `build-tools` doesn't have it, for example (I didn't look at the rest). So I'd suggest ensuring that we've actually run `pnpm install` in all workspaces and the lockfiles are fully in sync with what the pnpm-workspace files say.

frankmueller-msft:
Good catch — the pnpm-10 workspaces had their overrides in `pnpm-workspace.yaml`, but those are ignored when `pnpm.overrides` already exists in `package.json`. Fixed by:

- Moving the override to `package.json` under `pnpm.overrides` for all pnpm-10 workspaces (matching the existing pattern there)
- Keeping it in `pnpm-workspace.yaml` for pnpm-11 workspaces (root, website)
- **All lockfiles regenerated via `pnpm install`** — no more pattern-matched edits

alexvy86 (approved):
Not quite sure what's up with the build failures, but the change seems correct to me

ChumpChief:
Without having looked in detail here, probably need to reconcile with #27556 that already merged

build-tools/pnpm-lock.yaml
```diff
@@ -35,6 +35,7 @@ overrides:
   serialize-javascript@>=6 <7: ^7.0.4
   picomatch@>=2 <3: ^2.3.2
   picomatch@>=4 <5: ^4.0.4
+  shell-quote: ^1.8.4
```
ChumpChief: Given that there aren't actual package versions changing in many of these lock files, is the override needed (here and elsewhere)?  We should generally keep the override list small for maintenance reasons unless they're needed.

frankmueller-msft:
@alexvy86 please re-approve after I fixed the build failures

Abe27342:
Looks like `concurrently` published security patches to its 9.x branch that also bump shell-quote. Let's just update concurrently instead. I'd actually advocate we go ahead and bump to latest (10.0.3) since the breaking change list on v10 does not look particularly scary (dropped support for node <22, converted to ESM but we only use it via CLI). That should avoid us needing to add any overrides.

alexvy86:
I agree with Abram, now that we're on Node22, just bumping to `concurrently` 10.x seems like the best option here.

alexvy86:
> Without having looked in detail here, probably need to reconcile with #27556 that already merged

I'm a bit surprised that PR didn't have weird results once merged... if `concurrently` pins its dep to 1.8.3, I would expect that just updating the lockfiles across the repo to 1.8.4 without a corresponding override would cause a revert to 1.8.3 during the next install.

frankmueller-msft:
Updated the PR to use the approach suggested by @tylerbutler — bumped `concurrently` from 9.2.1 to 9.2.3 directly, which includes `shell-quote@1.8.4`. Removed the `shell-quote` override from `pnpm-workspace.yaml`. All 11 lockfiles now resolve `concurrently@9.2.3` with `shell-quote@1.8.4` and no `shell-quote@1.8.3` remains anywhere.

frankmueller-msft:
Updated to bump `concurrently` to `^10.0.3` per @Abe27342 and @alexvy86's suggestion. The v10 breaking changes (Node <22 dropped, ESM-only) are safe for us since we're on Node 22 and only use concurrently via CLI. Removed the `shell-quote` override as well — no longer needed with concurrently 10.x.

alexvy86 (approved):

jason-ha:
@frankmueller-msft, I don't know what happened with the merge message, but the title here "Bump concurrently to 10.0.3 to resolve shell-quote vulnerability" is much better than what was committed "Override shell-quote to ^1.8.4".

frankmueller-msft:
I agree @jason-ha. I should have updated the merge message.

PR #27555, opened 2026-06-17
Take an optional CommandLogger in pickFreshestRemote

ChumpChief:
## Description

Resolves [AB#73300](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73300).

Follow-up to review feedback on #27298: `pickFreshestRemote` was emitting its candidate table and per-remote failure diagnostics via raw `console.log` / `console.error`, which bypasses the oclif command's logger and leaks into `--json` stdout.

This routes those diagnostics through an optional `CommandLogger` parameter:
- The 4 `console.log` calls (eligible-remotes header, skipped entries, per-candidate rows) → `logger?.info(...)`
- The 2 `console.error` calls (per-remote unexpected-failure fallbacks) → `logger?.warning(...)`

The parameter is optional so non-command callers (or callers that want silence) can omit it.

`commands/check/bundleSize.ts` is updated to pass `this.logger` at the one current call site.

## Reviewer Guidance

Verified locally by running `flub check bundleSize` against this branch:
- **Plain mode**: the "Eligible remotes:" table still prints (now with the standard `INFO:` prefix oclif's logger adds — small visual change vs. raw `console.log`, semantically correct).
- **`--json` mode**: stdout is clean JSON, and stderr is empty too — oclif's `--json` mode suppresses `info` calls entirely. The library function doesn't need to know about `--json` mode; the command-level logger handles the suppression.

### DIFF
--- a/build-tools/packages/build-cli/src/library/git/pickFreshestRemote.ts
+++ b/build-tools/packages/build-cli/src/library/git/pickFreshestRemote.ts
@@ -5,6 +5,8 @@
 
 import { execFileSync } from "node:child_process";
 
+import type { CommandLogger } from "../../logging.js";
+
 /**
  * A remote ref paired with its locally-resolved tip commit.
  */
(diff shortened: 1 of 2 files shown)


build-tools/packages/build-cli/src/library/git/pickFreshestRemote.ts
```diff
@@ -138,12 +140,15 @@ function pickFreshest(candidates: RemoteCandidate[]): RemoteCandidate[] {
...
  *
+ * @param logger - Optional sink for the candidate-table, skipped-entry, and
+ * per-remote-failure diagnostics. Omit to run silently (e.g. for `--json` callers).
  * @returns The selected remote's name, or `undefined` if no remote matches
  * `filter` or none have a locally-resolvable `<name>/<branch>`.
  */
 export function pickFreshestRemote(
 	branch: string,
 	filter: (url: string) => boolean,
+	logger?: CommandLogger,
```
MarioJGMsoft, line 151: Since it's going to be optional, what happens when it's not there?  Doesn't that mean that all of the `logger?.` function calls will not do anything?

ChumpChief: That's right - it's basically a contract of "if you want logging, pass a logger - if you want silence, omit the logger"

MarioJGMsoft (approved):

PR #27560, opened 2026-06-18
Improve pr-bundle-size-comments discoverability and handle path-filtered builds

ChumpChief:
## Description

[AB#75913](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/75913).

Six small workflow-level changes split across three concerns: making runs and comments easier to correlate, tightening the workflow's reaction to ADO check_run completions so we don't post spurious or stale stickies, and keeping the sticky visible on chatty PRs.

### Discoverability

- **Add a dynamic `run-name:`** including the workflow, check name, event subtype, and head SHA. Actions list rows now look like `PR Bundle Size Comments | Build - client packages | completed | dbc8bdbdad...` instead of just the workflow name on every row.
- **Embed `<!-- pr-bundle-size-comments: run_id=... attempt=... -->`** in both the acknowledge-build and compare sticky bodies. Invisible in the rendered markdown but visible when viewing source — lets us trace a sticky back to the run that last wrote it.
- **Simplify the `compare` job's `name:`** to drop the conditional fallback. Skipped runs render the raw expression text regardless of how the expression is written, so the fallback was pure tax.

### Path-filtered build handling

ADO's `Build - client packages` pipeline fires `check_run.completed` even when its path filter skips the build (reporting `conclusion: neutral`). Without filtering, we triggered a compare that found no artifact and posted a spurious "⚠️ Comparison unavailable" sticky.

- **Narrow `identify-from-pr-build` / `identify-from-baseline-build`** to fire only on `success` / `failure` conclusions, preserving the documented intent that a real build failure still updates the sticky to a per-kind body.
- **Add `delete-sticky-on-neutral`**, a small companion job that handles the case where a PR previously *did* touch bundle-relevant paths (and got a real sticky) and a later push removed those changes. On `neutral`, it looks up the PR for the head SHA and deletes the sticky via `marocchino/sticky-pull-request-comment` with `delete: true` (a no-op when no sticky exists).

### Sticky placement

- **Pass `recreate: true` to `marocchino/sticky-pull-request-comment`** on both sticky-write steps so the comment moves to the bottom of the timeline on each update, matching the existing `changeset-reporter` / `linkcheck-reporter` convention for build-result stickies in this repo.

## Reviewer Guidance

Verifying end-to-end requires real check_run triggers. Once this lands, the bundle-size canary PR (#27548) will exercise the discoverability changes, the success-path identify flow, and `recreate: true` on its next build. The neutral / delete-sticky path is harder to exercise directly without a PR that toggles bundle-relevant paths between commits.

A follow-up to think about: `acknowledge-build` runs on `action == 'created'` before the conclusion is known, so it can't be filtered the same way. On a path-skipped PR where acknowledge-build wins the race against the `completed` event, its "Pending" sticky would stick around. Tracking separately.


.github/workflows/pr-bundle-size-comments.yml
```diff
@@ -154,17 +163,70 @@ jobs:
...
+                   nodes {
+                     ... on PullRequest { number }
+                   }
+                 }
+               }`,
+              { q: `type:pr is:open repo:${owner}/${repo} head-sha:${sha}` },
+            );
+            const pr = search.nodes[0];
+            if (!pr) {
+              core.info(`No matching open PR for SHA ${sha}; nothing to delete.`);
```
TommyBrosman, line 204: When can this happen? Does this script ever run for non-PRs?

ChumpChief: Technically could but the more common scenario would be the PR head changing - someone pushed another commit so the search for PRs with the given head-sha doesn't find it.  In that case we expect a new set of runs on the fresher commit that will supersede runs against the stale head-sha.

MarioJGMsoft (approved):
LGTM

PR #27561, opened 2026-06-18
Remove @fluid-internal/platform-dependent package

ChumpChief:
## Description

Removes the `@fluid-internal/platform-dependent` package, replacing its `generateRandomUInt32Array` utility with the web-standard `crypto.getRandomValues()` API.

This package originally existed to work around the absence of the Web Crypto API in Node.js. Node added `crypto.getRandomValues()` support in v17.4.0, and the `crypto` global object was added in v19.0.0 ([release notes](https://nodejs.org/en/blog/release/v19.0.0)). Since we now require Node 19+, we can use the standard API directly and eliminate the separate browser/server code paths.

## Changes

- Replaced the `generateRandomUInt32Array` import in `guidUtils.ts` with direct use of `crypto.getRandomValues(new Uint32Array(4))`
- Removed the `@fluid-internal/platform-dependent` package directory
- Regenerated `PACKAGES.md`
- Updated `pnpm-lock.yaml`

[AB#7451](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/7451)


experimental/PropertyDDS/packages/property-common/platform-dependent/browser.js
```diff
@@ -1,21 +0,0 @@
...
-
-/**
- * Create an array with random uint32 values
- *
- * @param {number} length - size of a new array
- * @return {number[]} - an array with random values
- */
-function generateRandomUInt32Array(length) {
-	const array = new Uint32Array(length);
-	const crypto = window.crypto || window.msCrypto; // IE 11 support
```
CraigMacomber, line 14: Ya, we don't need IE 11 support!

experimental/PropertyDDS/packages/property-common/src/guidUtils.ts
```diff
@@ -50,7 +49,8 @@ const guidRNG = {
 			this.isInitialized = true;
 
 			if (in_seed === undefined) {
-				const randomValues = generateRandomUInt32Array(4);
+				const randomValues = new Uint32Array(4);
+				crypto.getRandomValues(randomValues);
```
CraigMacomber, line 53: Looks like this is supported back as far as nodejs 15, and works in web workers and all modern browsers for years. Seems good to me. ( https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues )

CraigMacomber (approved):

CraigMacomber (approved):

experimental/PropertyDDS/packages/property-common/src/guidUtils.ts
ChumpChief: Kind of - the blocker the last time I looked at it was that node didn't support the `globalThis.crypto` global which only came in for Node 19.

Node 15: `require("crypto").webcrypto.getRandomValues()`
Node 17.4: `require("crypto").getRandomValues()`
Node 19: `crypto.getRandomValues()`

PR #27564, opened 2026-06-18
feat(build-tools): support TypeScript 6 tsbuildinfo format

jason-ha:
TypeScript 6 changed the .tsbuildinfo format by removing the `program` wrapper — all keys (fileNames, fileInfos, options, etc.) are now at the top level. This caused fluid-build to always consider TS6 packages as not up-to-date since it couldn't parse the new format.

Changes:
- Add `normalizeTsBuildInfo()` that detects TS5 (nested) vs TS6 (flat) format and normalizes to a unified shape
- Update `readTsBuildInfo()` to use the new normalization function
- Add `typescript-6.0` (npm:typescript@~6.0.3) as a type-only dependency
- Extend type unions to include TS6 types in tscTask.ts, tscUtils.ts, and tsCompile.ts
- Update `ITsBuildInfo.fileInfos` type to include optional `impliedFormat` field used by TS6
- Add unit tests for normalizeTsBuildInfo covering both formats

### DIFF
--- a/build-tools/packages/build-tools/src/test/tscTask.test.ts
+++ b/build-tools/packages/build-tools/src/test/tscTask.test.ts
@@ -1,0 +1,138 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { strict as assert } from "node:assert/strict";
+import { describe, it } from "mocha";
+import { normalizeTsBuildInfo } from "../fluidBuild/tasks/leaf/tscTask.js";
+
+describe("normalizeTsBuildInfo", () => {
+	it("parses TS5.x format with program wrapper", () => {
+		const ts5BuildInfo = {
+			program: {
+				fileNames: ["../lib.d.ts", "./src/index.ts"],
+				fileInfos: ["abc123", { version: "def456", affectsGlobalScope: true }],
+				options: {
+					target: 1,
+					module: 1,
+					strict: true,
+				},
+				changeFileSet: [1],
+				affectedFilesPendingEmit: undefined,
+				emitDiagnosticsPerFile: undefined,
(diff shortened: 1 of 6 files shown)


build-tools/packages/build-tools/src/test/tscTask.test.ts
```diff
@@ -0,0 +1,138 @@
...
+			composite: true,
+			declaration: true,
+			module: 100,
+			target: 8,
+			tsBuildInfoFile: "./tsconfig.tsbuildinfo",
+		});
+		assert.deepEqual(result.program.affectedFilesPendingEmit, [2]);
+		assert.equal(result.version, "6.0.3");
+	});
+
```
MarioJGMsoft, line 84: nit: couldn't all of the undefined test cases be under the same describe block? I feel like that might be cleaner. The way it currently is know is okay though.

MarioJGMsoft (approved):
Overall looks good, just left a small nit.

build-tools/packages/build-tools/src/test/tscTask.test.ts
jason-ha: agree would be preferred if the cases grow - three seems like the limit to ungrouped

build-tools/packages/build-tools/src/test/tscTask.test.ts
```diff
@@ -0,0 +1,138 @@
...
+			fileInfos: ["abc123"],
+			options: { strict: true },
+			semanticDiagnosticsPerFile: [[1, [{ code: 2322, message: "Type error" }]]],
+			version: "6.0.3",
+		};
+
+		const result = normalizeTsBuildInfo(ts6WithErrors);
+
+		assert.notEqual(result, undefined, "Expected a defined result");
+		assert.ok(result);
```
jason-ha, line 131: These could reduced to just .ok check, I think.

PR #27567, opened 2026-06-19
Add optional telemetry logger to Breakable and independent tree views

jzaffiro:
Plumb an optional logger through the SharedTree `Breakable` object to support telemetry through all levels of the tree.
Additionally, support logging on independent tree views by adding a logger to the options passed through the independent tree.

To make the new `logger` option easy to consume, this PR also re-exports the following telemetry types from `@fluidframework/core-interfaces` through the `fluid-framework` package:

- `ITelemetryBaseEvent`
- `ITelemetryBaseLogger`
- `LogLevel`
- `LogLevelConst`

Consumers of `fluid-framework` can now import these types directly without taking a separate dependency on `@fluidframework/core-interfaces`.

[AB#73616](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73616)

## Reviewer Guidance

Specifically, is there a better place to plumb telemetry through for an independent tree?


packages/dds/tree/src/shared-tree/independentView.ts
```diff
@@ -54,6 +58,36 @@ import {
...
+/**
+ * Minimal telemetry logger interface accepted by the independent tree view APIs.
+ *
+ * @remarks
+ * Structurally compatible with `ITelemetryBaseLogger` from `@fluidframework/core-interfaces`,
+ * so any standard Fluid telemetry logger can be passed directly.
+ *
+ * @privateRemarks
+ * This interface is declared locally rather than re-exporting `ITelemetryBaseLogger` so these
+ * `@alpha` APIs do not pull `ITelemetryBaseLogger` (and its transitive `LogLevel` /
```
CraigMacomber: These are stable public APIs: it seems odd we don't already include them in fluidframework, but I guess none of our stable APIs use them? Might be good to run it by API reviewers and/or someone who knows more about our telemetry and decide if we should simply add those types to fluid-framework.

packages/dds/tree/src/shared-tree/independentView.ts
```diff
@@ -54,6 +58,36 @@ import {
...
+	 * Log a telemetry event.
+	 * @param event - The event to log. Must include a `category` and an `eventName`.
+	 * @param logLevel - Optional numeric log level for the event.
+	 */
+	send(event: { category: string; eventName: string }, logLevel?: number): void;
+
+	/**
+	 * Minimum log level to emit. Events with a lower level than this are dropped.
+	 */
+	minLogLevel?: number | undefined;
```
CraigMacomber: If IndependentViewLogger is just suppose to be the minimal interface needed for independent views to log into (which is what would make sense for an input type here in my opinion), this seems a little odd in a few minor ways:

- This is mutable: I don't think we expect an independent view to want to modify the logging levels
- send should handle omitting of messages below the logging threshold, so we shouldn't need to read this field either.
- having it be optional and allow undefined makes it unclear what happens in the default/unset case. Is nothing logged? Is everything logged? Are only messages with logLevel undefined logged?

Also I think all events we generate should have a log level, so I think we should make send's log level required.

packages/dds/tree/src/shared-tree/independentView.ts
```diff
@@ -204,7 +254,15 @@ export function createIndependentTreeBeta<const TSchema extends ImplicitFieldSch
 export function createIndependentTreeAlpha<const TSchema extends ImplicitFieldSchema>(
 	options?: CreateIndependentTreeAlphaOptions,
 ): ViewableTree & Pick<ITreeAlpha, "exportVerbose" | "exportSimpleSchema"> {
-	const breaker = new Breakable("independentView");
+	// The public-facing `IndependentViewLogger` is intentionally a structural subset of
+	// `ITelemetryBaseLogger`; cast at this boundary so we can hand it to the standard telemetry helpers.
+	const logger = extractTelemetryLoggerExt(
+		createChildLogger({
+			logger: options?.logger as ITelemetryBaseLogger | undefined,
```
CraigMacomber: This seems questionable. If the typing really did make this save, a cast would not be needed. There are also safety issues here not captured at the type level.

Right now, there is no guarantee that our logger uses the same log level conventions for example.

Also this is a maintainability landmine. We declare our interface as `@input` and `@alpha` which allows us to make a lot of changes to it which are disallowed by the public interface. If we make any such changes, this type cast here will likely still compile, while it becomes incorrect introducing a bug.

packages/dds/tree/src/shared-tree/independentView.ts
```diff
@@ -54,6 +58,36 @@ import {
...
+ *
+ * @alpha @input
+ */
+export interface IndependentViewLogger {
+	/**
+	 * Log a telemetry event.
+	 * @param event - The event to log. Must include a `category` and an `eventName`.
+	 * @param logLevel - Optional numeric log level for the event.
+	 */
+	send(event: { category: string; eventName: string }, logLevel?: number): void;
```
CraigMacomber: input object don't need to be modifiable right, so we could make category and eventName readonly. Of course thatss incompatible with the fluid logger (except for the fact that TypeScript will let you assign away readonlyness).

I guess I don't really like our existing stable telemetry APIs in a bunch of different ways. I suppose aligning with them makes sense though.

.changeset/free-peas-lose.md
```diff
@@ -0,0 +1,29 @@
+---
+"@fluidframework/tree": minor
+"__section": tree
+---
+
```
Josmithr:
```suggestion
```

packages/dds/tree/src/shared-tree/independentView.ts
Josmithr: @jason-ha is it expected that the fluid-framework package doesn't export our logging interfaces? This doesn't seem right...

Josmithr: I would generally prefer we just re-use the existing types if we can, rather than creating our own here. If we only need/want a narrow subset of those types, we can express our types in terms of those.

jason-ha: I don't know about telemetry here and we should probably have a little discussion.
I *think* that fluid-framework doesn't need export the basic telemetry interfaces because they are only applicable with service clients. One basic logger given very early is threaded throughout the FF stack.

So, if this stands apart from needing to use service client, then there is a case to request them here. Otherwise, are we missing threading from existing sources?

Josmithr: > I think that fluid-framework doesn't need export the basic telemetry interfaces because they are only applicable with service clients. 

Ah, yeah that would make sense. 

> Otherwise, are we missing threading from existing sources?

This is a good point/question. Certainly, for attached/hydrated trees, it seems like we would want to be able to get the logger from the runtime. But it's also presumably useful for us to potentially log telemetry in unattached/unhydrated contexts as well, in which case we don't have access to an existing logger instance.

I'd be interested to learn what other DDSs do, assuming any of them do telemetry logging already.

packages/dds/tree/src/shared-tree/independentView.ts
jzaffiro: I'm a bit unclear on the guidance here then - if it's better to align with existing patterns, should I leave this signature unchanged?

packages/dds/tree/src/shared-tree/independentView.ts
jzaffiro: Fixed by using public types.

packages/dds/tree/src/shared-tree/independentView.ts
jzaffiro: Fixed by using public types.

packages/dds/tree/src/shared-tree/independentView.ts
jzaffiro: Fixed by using public types.

packages/dds/tree/src/shared-tree/independentView.ts
jzaffiro: Existing logging types are now re-exported through fluid-framework.

packages/dds/tree/src/shared-tree/independentView.ts
```diff
@@ -64,14 +68,29 @@ export interface IndependentViewOptions extends ForestOptions, Partial<CodecWrit
...
 	 */
 	idCompressor?: IIdCompressor | undefined;
+
+	/**
+	 * Optional logger for telemetry.
+	 * @remarks
+	 * Events emitted by the independent tree are tagged with the `independentView` namespace.
+	 * If not provided, telemetry events are dropped.
+	 */
+	logger?: ITelemetryBaseLogger | undefined;
```
Josmithr: Since this is an input type, this should probably be `readonly`

packages/framework/fluid-framework/api-report/fluid-framework.legacy.public.api.md
```diff
@@ -524,6 +524,20 @@ export type IsListener<TListener> = TListener extends (...args: any[]) => void ?
 // @public @system
 export type IsUnion<T, T2 = T> = T extends unknown ? [T2] extends [T] ? false : true : "error";
 
+// @public
+export interface ITelemetryBaseEvent extends ITelemetryBaseProperties {
```
Josmithr, line 528: Nit: I think it would probably be best to add these new `fluid-framework` re-exports in a separate PR with its own changeset.

Otherwise, we should at least add another changeset for these exports, and update the PR description to call them out.

packages/dds/tree/src/shared-tree/independentView.ts
```diff
@@ -122,7 +141,8 @@ export function independentView<const TSchema extends ImplicitFieldSchema>(
  */
 export function independentInitializedView<const TSchema extends ImplicitFieldSchema>(
 	config: TreeViewConfiguration<TSchema>,
-	options: ForestOptions & ICodecOptions,
+	options: ForestOptions &
+		ICodecOptions & { readonly logger?: ITelemetryBaseLogger | undefined },
```
Josmithr: Given that we have the same logger addition in a handful of places, I wonder if we should add a `IndependentViewTelemetryOptions` (or something) input type that we can re-use across these few places. Would give us a single place to document the property.

packages/framework/fluid-framework/api-report/fluid-framework.legacy.public.api.md
jzaffiro: API reports weren't generated correctly in a separate PR, so left them in this one and added a changeset.

Josmithr: Was the other PR already merged? It shouldn't be possible to merge the API changes without updating the reports.

jzaffiro: No, I meant locally - just adding the types to the `index.ts` didn't trigger API report generation.

.changeset/free-peas-lose.md
```diff
@@ -0,0 +1,28 @@
+---
+"@fluidframework/tree": minor
```
Josmithr, line 2: We should add `fluid-framework` here as well, since these APIs are surfaced there too.

Josmithr (approved):

PR #27568, opened 2026-06-19
build(build-tools): enable exactOptionalPropertyTypes

jason-ha:
Update build-tools packages to use exactOptionalPropertyTypes:true in normal build as well as check exactOptionalPropertyTypes:false successfully type checks (no emit, skipLibCheck).

### Fixes
build-tools:
 - **`common/biomeConfigUtils.ts`** — Conditionally spread `stopAt` only when defined, avoiding passing `undefined` to the `find-up` library's `Options.stopAt` which expects `string`.
 - **`fluidBuild/commonOptions.ts`** — Changed `defaultRoot` and `root` from optional properties (`?: string`) to required properties with explicit `undefined` in the union (`string | undefined`), allowing direct assignment from `process.env`.
 - **`fluidBuild/fluidRepo.ts`** — Removed explicit `ignoredDirs: undefined` assignment; conditionally spread `ignoredDirs` only when the mapped value is defined.
 - **`fluidBuild/tasks/workers/apiExtractorWorker.ts`** — Split return into two distinct object literals (success without `error`, failure with `error`) instead of assigning `undefined` to  the optional `error` property.
 - **`fluidBuild/tsCompile.ts`** — Conditionally spread `projectReferences` only when `commandLine.projectReferences` is defined, avoiding passing `undefined` to TypeScript API's  `CreateProgramOptions`.

version-tools:
 - **`commands/version.ts`** — Conditionally spread `bumpType` only when defined, avoiding `undefined` assignment to an optional property.

### DIFF
--- a/build-tools/packages/build-cli/package.json
+++ b/build-tools/packages/build-cli/package.json
@@ -40,6 +40,7 @@
 		"build:test": "tsc --project ./src/test/tsconfig.json",
 		"check:biome": "biome check .",
 		"check:format": "npm run check:biome",
+		"check:inexactOptionalPropertyTypes": "tsc --project ./tsconfig.inexactOptionalPropertyTypes.json",
 		"ci:build:docs": "api-extractor run",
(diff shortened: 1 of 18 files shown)


MarioJGMsoft (approved):
LGTM

PR #27569, opened 2026-06-20
chore(claims): rename @fluidframework/claims to @fluid-internal/claims

kian-thompson:
## Description

Renames the recently-added claims DDS package from `@fluidframework/claims` to `@fluid-internal/claims` to align its scope with its intended use: an implementation detail for other Fluid packages, with no public API surface. Every export in the package is already tagged `@internal`, so the public/beta/alpha API reports are empty — the `@fluid-internal` scope makes this explicit and matches the convention used by `@fluid-internal/client-utils`, `@fluid-internal/presence-runtime`, and similar packages.

The change also simplifies the package's `exports` map. Because the entire package is internal, the standard `.` vs `./internal` split is meaningless here. This PR:

- Removes the redundant `./internal` export subpath (the root `.` entry point already exposes the full internal API).
- Removes the `internal.d.ts` types-resolution shim that existed only to support the `./internal` subpath under `node16`/`bundler` module resolution.
- Keeps the test-only entry point under `./internal/test` to match `@fluid-internal/presence-runtime` and stay within the `npm-package-exports-apis-linted` policy exemption for `./internal/*` subpaths.

Updates the only in-repo consumer (`@fluid-internal/local-server-stress-tests`) to reference the new package name and updated subpath. Lockfile regenerated. No functional changes.

## Reviewer Guidance

- This is a metadata-only rename of a package that was just added in #27389 (and a follow-up lockfile fix in #27565). It has not yet shipped in a release, so there is no migration burden for external consumers.
- The empty public/beta/alpha API reports are intentional and unchanged in content (only the package-name header line differs).


packages/dds/claims/package.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "@fluidframework/claims",
+	"name": "@fluid-internal/claims",
 	"version": "2.110.0",
```
ChumpChief, line 3: Maybe also mark this `private: true` until we're ready to publish

ChumpChief (approved):

MarioJGMsoft (approved):
LGTM!

PR #27571, opened 2026-06-22
Usage error for transaction edits within event callback

daesunp:
## Description

Follow-up to [PR#27285](https://github.com/microsoft/FluidFramework/pull/27285). `runTransaction`/`runTransactionAsync` from inside a change-event callback (`nodeChanged`, `treeChanged`, or `changed`) now hits the same edit-lock guard as every other edit. `runTransaction` throws a `UsageError` and `runTransactionAsync` rejects with one ("Running a transaction is forbidden during a change
event callback").


.changeset/quick-pillows-hide.md
```diff
@@ -0,0 +1,17 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+Starting a transaction during a change-event callback now throws
+
+Building on the change-event edit lock, starting a transaction from inside one of a `SharedTree`'s change-event callbacks is now forbidden, the same as the direct edits that were already blocked.
```
Josmithr: Have we released the changes from your previous PR yet? If not, it would probably make sense to just expand on the existing changeset you added in the previous PR.

Josmithr (approved):
Left a changeset suggestion. Otherwise looks good to me!

Josmithr (approved):

PR #27572, opened 2026-06-22
extract reusable functions from plainTextView

daesunp:
## Description

Extracts reusable functions from plainTextView for shared use in our example app, and boards.


packages/framework/react/src/text/plain/usePlainTextInput.ts
```diff
@@ -0,0 +1,254 @@
...
+export interface PlainTextInputBinding {
+	/** Spread onto an uncontrolled `<input>` / `<textarea>`. See {@link PlainTextInputProps}. */
+	readonly inputProps: PlainTextInputProps;
+	/** Focus the bound element (e.g. for a click-anywhere-to-focus container). */
+	readonly focus: () => void;
+}
+
+/**
+ * React hook that two-way binds an (uncontrolled) text input to a
+ * {@link @fluidframework/tree#TextAsTree.Tree}, returning spreadable input props.
```
Josmithr: This isn't really what I'd been imagining. While the 2-way binding could be convenient for some cases, I'm not sure how universally applicable it will be. The caveats called out in the remarks below seem to support this concern.

My assumption is that we would just have a simple 1-way sync from tree -> string. This is simple, and I would think pretty generally applicable. I.e., something like:

Usage:
```typescript
const { text, selection } = useTreeSynchronizedString(tree, initialSelection);
```

Implementation:
```typescript
function useTreeSynchronizedString(tree: TextAsTree.Tree, initialSelection?: SelectionRange): {
	text: string,
	selection: SelectionRange | undefined
} : {
	// useEffect block that binds character-change listener and recomputes `text` and `selection` any time the text tree changes
	// I think this mostly just amounts to calling `applyTextOps` on any change
}
```

The app author is still responsible for providing the mapping in the other direction.

I think this would be a lot less complex. And an application author that wants to do something like what you've built below could easily use our code as reference / a starting place to create such a utility. But I worry that this code is simultaneously complex and may or may not be widely applicable.

What do you think? I am curious what led you to this larger scope - I could easily be missing something here.

daesunp: I started with a smaller scope, but kept expanding it to optimize for "ease of use" for the boards prototype, and ended up basically re-wrapping our existing example without noticing. Your suggestion makes a lot more sense, and I updated with your suggestions :)

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+ * The value returned by {@link useTreeSynchronizedString}.
+ * @internal
+ */
+export interface SynchronizedString {
+	/** The tree's current text. */
+	readonly text: string;
+	/**
+	 * A selection range tracked across edits.
+	 * @remarks
+	 * Seeded from `initialSelection` and adjusted by {@link applyTextOps} as the text changes. It
```
Josmithr: Nit: we should specify that this is the `initialSelection` passed into `useTreeSynchronizedString`, to be explicit.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+	 * follows the same logical position across edits, but the hook does not observe the user's live
+	 * caret, so a consumer that needs an accurate caret should track the element's selection itself.
```
Josmithr: I'm not sure I follow this remarks comment. Is this just saying that this should not be used to track cursor position?

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+	 */
+	readonly selection: TextSelection | undefined;
+}
+
+/**
+ * React hook that provides a one-way sync from a {@link @fluidframework/tree#TextAsTree.Tree} to a
+ * string: it returns the tree's current text (and a tracked selection), recomputed whenever the
+ * tree's characters change.
+ * @remarks
+ * This is the small, broadly-applicable primitive for reading a text tree in React. It owns only the
```
Josmithr: Nit: I think we can remove this first sentence - I don't think it offers much value to consumers of the API.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+ * subscription and the incremental {@link applyTextOps} apply (with a full re-read fallback). It
```
Josmithr: Nit: This second sentence also feels more like we're documenting implementation details. We probably don't really need this either.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+	selectionRef.current = selection;
+
+	useEffect(() => {
+		// Re-seed when the bound node changes.
+		const full = tree.fullString();
+		textRef.current = full;
+		setText(full);
+
+		return tree.onCharactersChanged((ops) => {
+			if (ops === undefined) {
```
Josmithr, line 82: When does this happen? I vaguely remember there being an edge case here, but I can't remember what it was.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * ```tsx
+ * const { text } = useTreeSynchronizedString(tree);
+ * return (
+ *   <textarea
+ *     value={text}
+ *     onChange={(e) => applyTextEdit(tree, e.target.value)}
+ *   />
+ * );
+ * ```
```
Josmithr: Nit: I'm wondering if we may want to give a bit more prescriptive guidance here. For simple APIs like `textarea`, where the events only return the fully updated string, then recommending `applyTextEdit` probably makes sense. But I think we should probably recommend doing more properly incremental edits when possible. 

We can probably get this across by noting that different text APIs offer different formats for change deltas, and so we don't offer a one-size-fits all solution for mapping those edits. But for simple APIs like `textarea`, we offer this naive implementation that can be used.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+		const full = tree.fullString();
+		textRef.current = full;
+		setText(full);
+
+		return tree.onCharactersChanged((ops) => {
+			if (ops === undefined) {
+				// No incremental delta available — re-read the whole string (selection unchanged).
+				const reread = tree.fullString();
+				textRef.current = reread;
+				setText(reread);
```
Josmithr, line 90: What happens to the selection in this case?

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+ * return (
+ *   <textarea
+ *     value={text}
+ *     onChange={(e) => applyTextEdit(tree, e.target.value)}
+ *   />
+ * );
+ * ```
+ * @internal
+ */
+export function useTreeSynchronizedString(
```
Josmithr, line 62: We should definitely add a suite of unit tests for this.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+	tree: TextAsTree.Tree,
+	initialSelection?: TextSelection,
```
Josmithr, line 64: Probably not for this PR, since it seems like we don't have a use-case for it yet. But I wonder if we should expand this to track 0 or more selections.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+			if (ops === undefined) {
+				// No incremental delta available — re-read the whole string (selection unchanged).
+				const reread = tree.fullString();
+				textRef.current = reread;
+				setText(reread);
+				return;
+			}
+
+			const oldText = textRef.current;
+			const oldSelection = selectionRef.current ?? {
```
Josmithr: Shouldn't the default here be no selection?

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,95 @@
...
+
+	// Mirror the latest text/selection in refs so the change listener reads current values without
+	// having to re-subscribe (which would re-run the effect) on every update.
+	const textRef = useRef(text);
+	textRef.current = text;
+	const selectionRef = useRef(selection);
+	selectionRef.current = selection;
+
+	useEffect(() => {
+		// Re-seed when the bound node changes.
```
Josmithr: Nit: I'm not sure this comment is necessary. We declare `tree` in the effect's dependencies array, which is probably sufficiently self-documenting IMO.

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -3,7 +3,163 @@
...
+ *
+ * This function is intentionally DOM-free so consumers maintaining their own UI (a custom React
+ * component, a non-`textarea` element, or a non-React renderer) can reuse it: read the current value
+ * and selection from your view, call this, then write the result back.
+ *
+ * `onCharactersChanged` delivers `undefined` (rather than an op list) when an incremental delta is
+ * unavailable; in that case skip this function and re-read the whole text via `root.fullString()`.
+ *
+ * Writing the result back must not re-enter the tree — see {@link applyTextEdit} for the re-entrancy pattern.
+ * @internal
```
Josmithr: Nit: do we need to package-export this? Presumably `applyTextEdit` is sufficient for our current use-cases?

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -3,7 +3,163 @@
...
+	root: TextAsTree.Tree,
+	newText: string,
+	label: unknown = root,
+): void {
+	const context = TreeAlpha.context(root);
+	if (context.isBranch()) {
+		context.runTransaction(() => syncTextToTree(root, newText), { label });
+	} else {
+		syncTextToTree(root, newText);
+	}
```
Josmithr: Is this branching needed? `context.runTransaction` should be callable regardless of whether or not we're on a branch. If we're not, I believe the implementation does what your fallback case here does anyways.

And if this is the case, I'm not sure we need this `applyTextEdit` helper at all. Users can reasonably grab the node's context and call `runTransaction` and `syncTextToTree` on their own.

Josmithr (commented):
Left a few more comments. This one is likely the most critical: https://github.com/microsoft/FluidFramework/pull/27572/changes#r3516458250

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: reworded with suggestion

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Updated remarks clarifying that if they want the real caret position should get it from the rendered element

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Updated with suggestion

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: removed sentence

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Updated the remarks to note that different text apis report their edits in different formats

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Updated so that tracked selection is clamped into [0, reread.length]

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Updated test suite to include useTreeSynchronizedString test

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Updated so selection stays undefined (if no selection was provided)

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Removed unnecessary comment

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: removed exports

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: removed if statement and updated to use context.runTransaction

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Added comment for clarification. Undefined happens when we have a field modified by multiple batches, and the field gets invalidated (and we just return undefined) in those cases, we unfortunately have to fallback on the full read

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -3,7 +3,161 @@
...
+ *   try {
+ *     applyTextEdit(root, newText);
+ *   } finally {
+ *     updating = false;
+ *   }
+ * }
+ * ```
+ * @internal
+ */
+export function applyTextEdit(
```
Josmithr: I don't think this helper function is needed. I think we can just let callers combine `runTransaction` and `syncTextToTree` as needed.

packages/framework/react/src/test/text/textEditor.test.tsx
```diff
@@ -139,15 +133,60 @@ describe("Plain TextArea view", () => {
...
 
-					rendered.rerender(content);
-					assert.match(rendered.baseElement.textContent ?? "", /👋🌍!🎉/);
+					assert.equal(getTextareaValue(rendered.container), "👋🌍!🎉");
 				});
 			});
 		}
 	});
 
+	// The hook is a one-way tree → string sync; exercise it directly via renderHook.
```
Josmithr: These tests should live in the test module that corresponds with the source code module. For now, that would be `useTreeSynchronizedString.spec.ts`. That said, it might make sense to merge that function into `plainUtils.ts`, in which case these would move to `plainUtilts.test.ts`

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,112 @@
...
+				// delta — e.g. the field was modified across multiple batches within one flush (such
+				// as an interleaved schema change) — or when the tree is out of sync with the delta.
+				// Any tracked selection is clamped into the new text, since a shrinking edit could
+				// otherwise leave it out of bounds.
+				const reread = tree.fullString();
+				textRef.current = reread;
+				setText(reread);
+				if (selectionRef.current !== undefined) {
+					const clamp = (offset: number): number =>
+						Math.min(Math.max(offset, 0), reread.length);
```
Josmithr: We've defined `clamp` in 2 different places in this PR. My guess is that we have a `clamp` utility already somewhere that we should use. But if not, we should add this to `arrayUtils` (or someplace) so we can deduplicate.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,112 @@
...
+			if (ops === undefined) {
+				// No incremental delta is available for this change, so re-read the whole string.
+				// This happens when the character field's marks couldn't be composed into a single
+				// delta — e.g. the field was modified across multiple batches within one flush (such
+				// as an interleaved schema change) — or when the tree is out of sync with the delta.
+				// Any tracked selection is clamped into the new text, since a shrinking edit could
+				// otherwise leave it out of bounds.
+				const reread = tree.fullString();
+				textRef.current = reread;
+				setText(reread);
```
Josmithr, line 90: Is this correct for all cases? I can imagine a case where a batched set of changes include deleting all characters, then inserting other characters. In a normal case, this would (presumably) result in the selection being effectively removed. But in this case, it would result in unrelated characters being selected.

I haven't thought through all of the relevant cases here, but I have a suspicion that this isn't the policy we want here.

I would recommend thinking through the cases here and ensuring we have them covered by tests. Since this is a helper utility, rather than built-in functionality, it's probably okay for there to be edge cases we don't handle well. But we should make sure these cases and associated policy are clearly documented, and that the policy we go with is the one that will cause least disruption.

packages/framework/react/src/test/text/textEditor.test.tsx
```diff
@@ -139,15 +133,60 @@ describe("Plain TextArea view", () => {
...
+	describe("useTreeSynchronizedString", () => {
```
Josmithr: I suspect we should probably have many more test cases than this. The code has a number of branches and edge cases we should make sure we are covering. E.g. cases where the selection gets clamped, cases where the selection is reduced to 0 characters, etc.

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: removed helper, and inlined code.

packages/framework/react/src/test/text/textEditor.test.tsx
daesunp: moved tests into a new `useTreeSynchronizedString.spec.ts` files

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Added `clamp` utility

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: This bug unfortunately wasn't caught because it lived in the ops === undefined fallback branch, and our test edited using `insertAt/removeRange`, which emits a proper delta, and never reached the fallback. 

Created helper `collapseSelectionOnReread` so that we can easily add unit tests for this, and changed the policy to collapse the selection to a single caret at the clamped previous  start.

packages/framework/react/src/test/text/textEditor.test.tsx
daesunp: Added `collapseSelectionOnReread` (with unit tests), unit tests for applyTextOps cases, and hook tests.

packages/framework/react/src/test/text/plainUtils.test.ts
```diff
@@ -68,49 +73,100 @@ describe("plainUtils", () => {
 			]);
 			assert.deepEqual(result, { value: "lo", selection: { start: 2, end: 2 } });
 		});
+
+		it("collapses a selection to a caret when its whole range is removed", () => {
+			// "abcdef" with selection "cd" (2-4); remove "bcde" (1-5), which fully contains the selection.
+			// Both offsets pull back to the removal start, leaving a collapsed caret.
```
Josmithr: Nit: I don't think "collapsed caret" is the right terminology for us to use. A caret is semantically different from a selection range, which may or may not be empty. I think "empty selection" or "empty selection range" is probably better. (for here and elsewhere).

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
Josmithr: I'm not sure a collapsed selection is necessarily right either. The position of the collapsed range may also be completely arbitrary. I wonder if it might be better to just fully drop the selection in this case instead.

Alternatively, maybe we can do something a bit cleverer. In our string -> tree sync utility, we make a best effort to compute some set of edits that could represent the diff from before -> after. I wonder if we could leverage something similar here for the selection. The results may not be perfect in some cases, but they're likely to be right more often than not (or at least, that would be my guess).

packages/framework/react/src/text/plain/plainTextView.tsx
```diff
@@ -50,126 +47,41 @@ export const MainView: FC<MainViewProps> = ({ root, undoRedo, editLabel }) => {
...
+			return;
+		}
+		const textarea = textareaRef.current;
+		if (textarea !== null && selection !== undefined) {
+			textarea.setSelectionRange(selection.start, selection.end);
+		}
+	}, [text, selection]);
+
+	// String → tree: write the user's edit back into the tree, wrapped in a transaction tagged with
+	// `effectiveLabel` so the remove + insert pair is applied atomically and undone/redone as a unit.
```
Josmithr: I hadn't previously realized that `syncTextToTree` didn't already wrap its edits in a transaction. That seems like a bug. Since we export that function on its own, we should also update it to directly wrap its 2 edits in a transaction. We still want to keep this outer transaction here for labeling purposes though.

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -3,14 +3,180 @@
...
+import { type TextAsTree, utf16LengthForCodePoints } from "@fluidframework/tree/internal";
+
+import { clamp } from "../../utilities.js";
+
+/**
+ * A text selection or cursor range expressed as UTF-16 code-unit offsets.
+ * @remarks
+ * These match the `selectionStart` / `selectionEnd` properties of an HTML `<input>` / `<textarea>`,
+ * and the offsets used to index into a JavaScript string. A collapsed cursor has `start === end`.
+ * @internal
```
Josmithr: If we aren't package-exporting these types, then we should remove the `@internal` release tags from them.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+import type { TextAsTree } from "@fluidframework/tree/internal";
+import { useEffect, useRef, useState } from "react";
+
+import { applyTextOps, collapseSelectionOnReread, type TextSelection } from "./plainUtils.js";
+
+/**
+ * The value returned by {@link useTreeSynchronizedString}.
+ * @internal
```
Josmithr: Since the plan is to begin using this new API in our prototypes, we should tag this type and the react hook as `@alpha`.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Updated to use your second suggestion. Added `remapSelectionOnReread` where it gets the contiguous edit from the longest shared prefix and remaps the selection.

falls back to dropping the selection range when the either the start or end of the selection falls within the changed range.

packages/framework/react/src/test/text/plainUtils.test.ts
daesunp: Updated terminology with suggestion

packages/framework/react/src/text/plain/plainTextView.tsx
daesunp: Updated to wrap edits in transaction in `syncTextToTree`

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: removed release tags

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: Tagged as `@alpha`

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -3,25 +3,254 @@
...
+} from "@fluidframework/tree/internal";
+
+import { clamp } from "../../utilities.js";
+
+/**
+ * A text selection or cursor range expressed as UTF-16 code-unit offsets.
+ * @remarks
+ * These match the `selectionStart` / `selectionEnd` properties of an HTML `<input>` / `<textarea>`,
+ * and the offsets used to index into a JavaScript string. A collapsed cursor has `start === end`.
+ * @alpha
```
Josmithr, line 20: Since this type represents data that the user passes to us, we should probably tag it as `@input`

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -3,25 +3,254 @@
...
+
+/**
+ * Best-effort remap of a tracked selection across a text change that arrived without an incremental
+ * delta (so the whole string had to be re-read).
+ * @remarks
+ * When no delta is available (see {@link @fluidframework/tree#TextAsTree.Members.onCharactersChanged}),
+ * the ops that transformed `oldText` into `newText` are unknown, so a selection range cannot be
+ * moved faithfully. This makes a best effort by inferring a single contiguous edit the same way
+ * {@link computeSync} does — the longest shared prefix/suffix between `oldText` and `newText`, with
+ * everything in between treated as one replaced span:
```
Josmithr, line 160: Nit: any possibility of us sharing code with `computeSync`?

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+
+import type { TextAsTree } from "@fluidframework/tree/internal";
+import { useEffect, useRef, useState } from "react";
+
+import { applyTextOps, remapSelectionOnReread, type TextSelection } from "./plainUtils.js";
+
+/**
+ * The value returned by {@link useTreeSynchronizedString}.
+ * @alpha
```
Josmithr, line 14: Since this interface represents data specifically used as output for the user to read (and not data they write to / provide to us as input), we should probably tag this as `@sealed`.

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -3,25 +3,254 @@
...
+ * internally, so a single call applies (and undoes/redoes) as one atomic unit.
+ *
+ * To attach an undo/redo label, nest this inside your own transaction, e.g.
+ * `TreeAlpha.context(root).runTransaction(() => syncTextToTree(root, newText), { label })`. The
+ * `label` correlates the edit for undo/redo grouping where the context supports it (see
+ * {@link @fluidframework/tree#TreeContextAlpha.runTransaction}).
+ *
+ * Mutating the tree synchronously fires
+ * {@link @fluidframework/tree#TextAsTree.Members.onCharactersChanged}. If you also subscribe to that
+ * event to apply remote edits to your view (e.g. via `applyTextOps`), set a re-entrancy flag
```
Josmithr: We don't export `applyTextOps` anymore.

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -3,25 +3,254 @@
...
- * @internal
+ * and replacing only the middle span. The resulting remove + insert pair is wrapped in a transaction
+ * internally, so a single call applies (and undoes/redoes) as one atomic unit.
+ *
+ * To attach an undo/redo label, nest this inside your own transaction, e.g.
+ * `TreeAlpha.context(root).runTransaction(() => syncTextToTree(root, newText), { label })`. The
+ * `label` correlates the edit for undo/redo grouping where the context supports it (see
+ * {@link @fluidframework/tree#TreeContextAlpha.runTransaction}).
+ *
+ * Mutating the tree synchronously fires
```
Josmithr, line 227: Nit: this warning / example might be a good usecase for an `@example` block. The first line of these blocks serves as the "title", so you could start this with something like:

```ts
 * @example Avoiding re-entrant edits
 * ...
```

packages/framework/react/src/text/plain/plainUtils.ts
```diff
@@ -3,25 +3,254 @@
...
  * Sync `newText` into the provided `root` tree by applying the minimal remove + insert pair
  * needed to transform the tree's current content into `newText`.
  * @remarks
  * The diff is computed by finding the longest shared prefix/suffix between current and new content
- * and replacing only the middle span.
- * @internal
+ * and replacing only the middle span. The resulting remove + insert pair is wrapped in a transaction
+ * internally, so a single call applies (and undoes/redoes) as one atomic unit.
+ *
+ * To attach an undo/redo label, nest this inside your own transaction, e.g.
```
Josmithr: Nit: I'm not sure this note about undo/redo labeling is necessary. If a user is leveraging our label-aware UndoRedo manager type, then they should already be aware of how to do this.

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
```diff
@@ -0,0 +1,119 @@
...
+	readonly text: string;
+	/**
+	 * A selection range tracked across edits, or `undefined` when no selection is being tracked.
+	 * @remarks
+	 * Seeded from the `initialSelection` passed to {@link useTreeSynchronizedString} and adjusted as
+	 * the tree's characters change so it follows the same logical position across edits.
+	 *
+	 * When an incremental delta is unavailable and the hook must re-read the whole string, the range
+	 * cannot be mapped faithfully; the hook makes a best effort to remap it from the old/new text and
+	 * drops the selection (becomes `undefined`) when it can't place an endpoint reliably.
```
Josmithr: I think we should avoid surfacing this to the user like this. It might be best to just leave a short note that this is best-effort, and there are rare edge cases where selections may be dropped.

daesunp: Updated to a short note

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: removed

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: Added `@example` tag for that block

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: removed `applyTextOps` reference

packages/framework/react/src/text/plain/useTreeSynchronizedString.ts
daesunp: updated with suggested tags

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: factored out shareable code into `commonAffixLengths` helper

packages/framework/react/src/text/plain/plainUtils.ts
daesunp: updated tag

Josmithr (approved):
Looks good! 🚀

PR #27573, opened 2026-06-22
feat(claims-example): add example app demonstrating the Claims DDS

kian-thompson:
## Description

Adds a new `@fluid-example/claims-example` app that demonstrates the **Claims DDS** (`@fluid-internal/claims`) running inside a real Fluid container, and makes a small supporting change to the Claims DDS API itself.

The Claims DDS is an internal building block for first-writer-wins ownership: clients compete to bind a key to a value, and the ordering service picks a single winner that every client converges on. This example wires it up by hand inside a custom `ClaimsDataObject`, since the Claims DDS has no public consumption path yet.

**What the example shows**

- A `ClaimsDataObject` owns a single Claims DDS (stored by handle on the root) and exposes a narrow `trySetClaim` / `getOwner` surface, so the view never touches the Claims DDS directly.
- Each browser tab gets its own `claimant` identity. Opening the same container URL in a second tab gives you a competing client.
- There is a small, fixed set of known keys (`ClaimKey1`, `ClaimKey2`). Claiming a key creates a fresh `SharedDirectory` that records the owner and binds its handle as the claim value, so every client resolves the winning handle to the same shared object — no key enumeration or side structure required.
- **First-writer-wins:** the first client to claim a key wins, and a losing client is switched to reflect the winner. Connected claims come back `Pending` and settle after the op roundtrip; detached claims resolve synchronously.
- The view reads each key's owner live from the resolved backing directory and re-renders on the Claims DDS `claimed` event, so ownership stays in sync both locally and remotely.

The app runs against the shared `@fluid-example/example-driver`, so the backing service (local/t9s, ODSP, …) is selected at build time rather than hard-coded.

**Claims DDS change**

Removes `currentValue` from the Claims DDS result types (`ClaimConfirmation` / `ClaimResult`), which also makes them non-generic. Callers that need the current owner read it back via `claims.get(key)` — which is fresher and, for write-once keys, immutable. This narrows the `@internal` API surface with no change to claim/consensus behavior.

## Reviewer Guidance

The example deliberately uses `/internal` entry points because the Claims DDS has no public consumption path, and we don't intend to add one.

[AB#74216](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/74216)


examples/apps/claims-example/src/index.ts
```diff
@@ -0,0 +1,322 @@
...
+		console.log(`  ✓ CAS succeeded. "config" now resolves to: "${updated.get("name")}"`);
+	} else {
+		console.log("  ✗ CAS failed — a concurrent write was detected.");
+	}
+	console.log();
+}
+
+// ─── Main ─────────────────────────────────────────────────────────────────────
+
+async function main(): Promise<void> {
```
ChumpChief: This is a test, rather than an example.  The example needs to align with the structure of our other examples, as the point of this is to demonstrate and explore (a) what realistic usage patterns of the feature look like in practice and (b) what integration of the feature with an application looks like in practice.

I'd recommend forking from either external-views (if you want to experiment with a non-aqueduct approach) or diceroller (if you want to experiment with an aqueduct approach).  Since a followup step will be experimenting with where the aliasing feature will integrate (data store runtime, aqueduct, etc.) that might factor into your choice.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
```diff
@@ -0,0 +1,224 @@
...
+		this.claims = await claimsHandle.get();
+		this.claimedKeysDir = this.root.getSubDirectory(claimedKeysDirName);
+
+		// Switch to (resolve) a key's winning directory whenever it is claimed, locally or
+		// remotely. This is also how a losing client ends up pointing at the winner.
+		this.claimsOrThrow.events.on("claimed", (key) => this.onClaimed(key));
+
+		// A late-joining client (or a reload of an existing document) discovers
+		// already-claimed keys from the mirror and resolves each of their winners.
+		for (const key of this.claimedKeysDirOrThrow.keys()) {
```
ChumpChief: This is kind of interesting.  Claims doesn't currently have an iterator/iteration methods, should it?

Alternatively/also: maybe we think in normal scenarios it's less likely the user will want to iterate a claims object?  If so, maybe there's a scenario to explore where the claims is used to negotiate a winner but then the winning handle is transferred to a more-standard location like the root directory?  The Pages scenario might look closer to that?

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
```diff
@@ -0,0 +1,224 @@
...
+			this.resolve(key);
+		}
+	}
+
+	public readonly trySetClaim = async (key: string): Promise<boolean> => {
```
ChumpChief, line 118: This is also interesting - the claims interface supports synchronous return but here it gets wrapped in an async method that immediately awaits the `result.promise`.  Would it be easier for a customer to use `trySetClaim` if it dropped the synchronous return of a `ClaimResult` and instead always returned a `Promise<ClaimConfirmation>` (and a smaller interface)?

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
```diff
@@ -0,0 +1,224 @@
...
+
+		const result = this.claimsOrThrow.trySetClaim(key, directory.handle as DirectoryHandle);
+
+		// Connected claims come back "Pending"; await the op roundtrip for the real outcome.
+		const outcome = result.status === "Pending" ? await result.promise : result;
+
+		if (outcome.status === "Accepted") {
+			// The "claimed" event resolves the winner, but resolve eagerly so the caller sees an
+			// accurate result even in detached mode (where no op — and so no event — is produced).
+			this.recordClaimedKey(key);
```
ChumpChief: This is the only path (claim winner) that writes the claim into the sidekick subdirectory DDS via recordClaimedKey.  If the claim winner dies before seeing their own success or otherwise doesn't complete this flow, the sidekick DDS will remain out of sync, preventing that claim's enumeration.

This could heal by having other contestants update the sidekick in the AlreadyClaimed path, but it's probably still not as nice as built-in enumeration if we decide this is a scenario customers will care about (e.g. would probably result in non-winner clients racing to set the value - not functionally wrong but chatty protocol).

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
```diff
@@ -0,0 +1,224 @@
...
+		// already-claimed keys from the mirror and resolves each of their winners.
+		for (const key of this.claimedKeysDirOrThrow.keys()) {
+			this.resolve(key);
+		}
+	}
+
+	public readonly trySetClaim = async (key: string): Promise<boolean> => {
+		// Each attempt creates a fresh backing directory recording this client as the owner.
+		// If the claim wins, this directory becomes the shared winner for the key.
+		const directory = SharedDirectory.create(this.runtime);
```
ChumpChief, line 129: Maybe consider if you want to early-exit when `claims.has(key)` (or the subdirectory DDS, or the resolvedDirectories).  Not that creating a SharedDirectory is particularly expensive throw-away work but I imagine the normal usage flow would be to first check if the work is necessary.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
```diff
@@ -0,0 +1,224 @@
...
+	/**
+	 * Subdirectory mirroring the claimed keys (see {@link claimedKeysDirName}). Set during
+	 * initialization.
+	 */
+	private claimedKeysDir: IDirectory | undefined;
+
+	/**
+	 * Backing SharedDirectories for resolved winning claims, keyed by claim key.
+	 */
+	private readonly resolvedDirectories = new Map<string, ISharedDirectory>();
```
ChumpChief, line 74: I don't think I fully understand the purpose of a separate `resolvedDirectories` from `claimedKeysDir`.  Is it just to avoid the async `handle.get()`?

It feels a little bad to have triple-tracking of each claim, if there's any way to simplify.

ChumpChief (commented):
Very cool, thank you for doing this :)

Probably the most interesting things to focus on are the questions that might impact the claims DDS API surface.  And maybe also consider if there's a way to shape the example so it looks how Pages will implement aliasing?

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: I'd like to keep the synchronous `ClaimResult` for flexibility. In detached mode there's no op roundtrip, so `trySetClaim` can resolve `Accepted`/`AlreadyClaimed` synchronously — letting callers act without an `await` (and its microtask tick), which matters for paths that resolve eagerly during creation where no op or `claimed` event is produced.

Always returning `Promise<ClaimConfirmation>` is a smaller surface, but it forces an async hop even when the answer is already known locally. So, I'd keep the sync return and let callers opt into async depending on their use case.

Three use cases I imagine:
1. Result is known locally, so it's sync
2. Result is not known locally, so return pending. I decide I don't want to wait and will set up some handler to change things over when the result is known. I imagine this might be useful for offline/staging mode in future?
3. Result is not known locally, so return pending. I wait for the claim to resolve.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: I simplified this a lot in the example as it isn't necessary to iterate. Ideally, I want the claims DDS to store the handle to the other DataObject, not simply use it to elect a winner. If future consumers of this DDS would like to iterate through the keys, then they can add the functionality.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: Added

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: Simplified this code a lot, now it's a "resolvedOwners" that just stores claim owner information. Purely UI related for the example, and I don't plan on using this type of mechanism for the partner team request.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: Fixed by the reshape. The sidekick mirror is gone, so there's nothing left to drift.

examples/apps/claims-example/src/view.tsx
```diff
@@ -5,17 +5,21 @@
...
+const claimKeys = [claimKey1, claimKey2];
+
 export const ClaimsView: FC<IClaimsViewProps> = ({ claimsDataObject }: IClaimsViewProps) => {
 	// Re-render whenever a claim is made or an owner changes, locally or remotely.
 	const [, forceUpdate] = useState({});
-	const [keyToClaim, setKeyToClaim] = useState<string>("");
 	const [status, setStatus] = useState<string>("");
+	// Keys with a claim currently in flight. The Claims DDS throws if a second claim is started
+	// for a key while the first is still pending, so we disable the button until it settles.
+	const [claiming, setClaiming] = useState<ReadonlySet<string>>(() => new Set());
```
ChumpChief: Nit, shouldn't need a function here
```suggestion
	const [claiming, setClaiming] = useState<ReadonlySet<string>>(new Set());
```

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
ChumpChief: I think it's fine to keep if you believe there's real customer value, but my feedback was primarily about ease of use (fewer unique outcomes for the customer to handle), surface area is a secondary concern.  I personally prefer the sync implementation since as we discussed previously I'd probably always rely on the events rather than the return value, but just thinking about what someone using the return value would need to do to handle it fully.

FWIW, microtask ticks are not usually a compelling argument - if the system could result in an async outcome in some branches it needs to be robust to microtask ticks anyway.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
```diff
@@ -62,103 +60,90 @@ export class ClaimsDataObject
...
 		}
 
 		if (outcome.status === "AlreadyClaimed") {
-			// We lost the race: switch to the winner's directory instead of the one we created.
-			const winner = await outcome.currentValue?.get();
-			if (winner !== undefined) {
-				this.adoptDirectory(key, winner);
-			}
+			// We lost the race: resolve the winner's owner instead of the directory we created.
+			this.resolve(key);
```
ChumpChief, line 146: In this latest iteration, `outcome.currentValue` isn't used anymore.

Arguably it was already duplicative (customer could just `claims.get(key)` to get the current value).  Is there value in keeping it, or maybe would be preferable for `ClaimResult` to just be status + maybe-promise?

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
```diff
@@ -62,103 +60,90 @@ export class ClaimsDataObject
...
 
 		if (outcome.status === "Accepted") {
 			// The "claimed" event resolves the winner, but resolve eagerly so the caller sees an
 			// accurate result even in detached mode (where no op — and so no event — is produced).
-			this.recordClaimedKey(key);
 			this.resolve(key);
 			return true;
 		}
 
 		if (outcome.status === "AlreadyClaimed") {
```
ChumpChief, line 144: Optional for this PR, but it might be interesting to show a trivial data-merge in the case that there is further non-overlapping data from each racing client.  E.g. maybe in addition to the "owner" key, each client wants to write their ID as a standalone key so the merged data after a 2-client race would look like

```json
[
  ["owner", "clientA"],
  ["clientA", "clientA"],
  ["clientB", "clientB"]
]
```

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
ChumpChief: Same question though, I think the purpose of `resolvedOwners` is to avoid the async `handle.get()` - if so then keeping this as `resolvedDirectories` is probably actually correct (more likely a customer will retain a reference to a DDS than a reference/copy of a specific contained value of the DDS).  And then `getOwner` just consults the resolved directory synchronously.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
```diff
@@ -62,103 +60,90 @@ export class ClaimsDataObject
...
+			return false;
+		}
+
 		// Each attempt creates a fresh backing directory recording this client as the owner.
 		// If the claim wins, this directory becomes the shared winner for the key.
 		const directory = SharedDirectory.create(this.runtime);
 		directory.set(ownerKey, this.claimant);
 
-		const result = this.claimsOrThrow.trySetClaim(key, directory.handle as DirectoryHandle);
+		const result = this.claims.trySetClaim(key, directory.handle as DirectoryHandle);
```
ChumpChief, line 132: Not needed for this PR but maybe worth logging for consideration as a followup -- I'm realizing that the `UsageError` in `Claims.trySetClaim` is maybe tough for a customer to avoid in some scenarios since there's no means to detect whether a local claim is in-flight.  They'd either need to ensure single-writer for the Claims (such that there can't be local competition) or else build their own pending tracking if they allow multiple-writers.  Or they could try to catch the error and set up event handling to wait for the eventual local winner but that also feels tough.

I wonder if there's a nice way to fold that scenario into an `AlreadyClaimed`-style response (returned once the winner -- local or otherwise -- is determined).  Yes we already know locally that they can't be the winner, but it's not like they can do anything else in the meantime until the winner is actually determined anyway.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
```diff
@@ -62,103 +60,90 @@ export class ClaimsDataObject
...
+	 * Resolved owner of each winning claim, keyed by claim key.
+	 *
+	 * @remarks
+	 * Claim values are handles, which resolve asynchronously, but the view reads owners
+	 * synchronously while rendering. A winning directory's owner is written once and never
+	 * changes, so this map memoizes the resolved owner string — letting {@link getOwner} stay
+	 * synchronous without holding onto (or subscribing to) the backing directory.
 	 */
-	private claimedKeysDir: IDirectory | undefined;
+	private readonly resolvedOwners = new Map<string, string>();
```
ChumpChief: I think this is better than the prior sidekick DDS implementation, but there probably is still something interesting to explore in terms of what we expect a customer wants for aliasing.

In particular, do we think customers (EDIT: "data object authors") will be happy having two long-term stores of "root-level" data (`this.root` and a separate `claims` DDS)?  I can imagine at least three possible options that each have their pros/cons (and maybe more than one is valid, depending on a customer's specific scenario?):
1. Data object retains some data in `this.root` and other data in `claims` long-term, they don't attempt to reconcile with one another.
2. Use the `claims` to get consensus on a handle and then copy it into `this.root` (or wherever it should be stored).  Anyone can do the copy even if the winner died, but it's a little chatty if they race.
3. Migrate completely to the `claims` as the source of truth, replacing `this.root` as the de-facto place to store handles.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
```diff
@@ -62,103 +60,90 @@ export class ClaimsDataObject
...
-		this.root.createSubDirectory(claimedKeysDirName);
 	}
 
 	/**
-	 * Called every time the data object is initialized. Resolves the Claims DDS and mirror
-	 * subdirectory and wires up listeners so the local view stays in sync as claims are made.
+	 * Called every time the data object is initialized. Resolves the Claims DDS and wires up
+	 * listeners so the local view stays in sync as claims are made.
 	 */
 	protected override async hasInitialized(): Promise<void> {
```
ChumpChief, line 100: One more scenario to consider as a followup - it's probably pretty likely that customers will want to trigger an alias attempt on load for schema changes to existing data objects ("if DDS foobar doesn't exist when I load, then try to create it and alias it").  It might be interesting to explore what that looks like - e.g. Do they make that call here (emitting the op) in `hasInitialized`?  Do they await the result here (delaying the load but ensuring a complete data object after initialization)?

ChumpChief (commented):
Mostly just more thoughts on what usage will look like for existing customers.  Would like your thoughts on these in case they inspire any changes now but most of them are probably fine to consider in followups if you prefer.

Another avenue to help evaluate the API would be to issue a prerelease build with the Claims DDS and attempt an integration into whatever we think the best first candidate is.  Or, try to replicate patterns you find from that candidate into this example.  Point being "can we tell the whole story of the expected integration", and ideally iterate on that using prereleases or examples since they're faster than waiting for our official releases and don't incur compat burden/delays.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: Makes sense. I brought back the `resolvedDirectories` as a sort of "resolved cache" to read synchronously.

examples/apps/claims-example/src/view.tsx
kian-thompson: Fixed

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: Fair point on the microtask ticks. The  "Pending"  path is async anyway, so callers have to be robust to that regardless. The one case I'd still keep sync for is a purely-synchronous caller (detached/init flows that can't  await  at all). Narrow, but real. Since you're fine either way, I'll keep the sync return for now and revisit if the tri-state proves annoying in practice.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: You're correct, it is a little redundant. Removed `currentValue` from the API surface

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: Worth showcasing at some point, but I'll keep it out of this PR

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: Good callout, I'll keep the `UsageError` for now. I lean toward seeing this as a usage-pattern issue (single-writer, or app-level pending tracking) rather than something the DDS must absorb. If we later decide it's worth smoothing over, we could revisit by adding a dedicated state to represent "a local claim is already in-flight" rather than throwing.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: The intent is that the consumer only interacts with this data object's API (`trySetClaim` / `getOwner`), never the `claims` DDS or `this.root` directly. So where the data lives is an encapsulated detail, not something they reconcile.
Consolidating internally (option 2/3) is worth exploring later, but it doesn't change what the customer sees.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
kian-thompson: Agreed claims and aliasing are related in shape, both first-writer-wins consensus. But they differ in scope and discovery: aliasing works at the container level over data stores and gives a root-discoverable name with no pre-existing reference, whereas claims are scoped to keys within a single DDS and require an existing handle path to reach.
So for schema migration where you discover-or-create a data store on load, aliasing is really the right tool. Folding it into this DDS-level example would conflate the two. Worth exploring on its own.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
ChumpChief: Fine to defer for now, I'd probably lean away from a new dedicated state though since that gives another case that callers need to handle (and the behavior isn't truly that unique, other states already represent "you tried but failed")

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
ChumpChief: Our (Fluid's) customers write the data object though, the internal implementation of the data object is exactly their concern.  Small edit to the comment above to specify that I'm talking about "data object authors" as the customer - not "view layer authors".  This is still worth considering, but OK to defer for now.  Probably make it a priority to have a really solid story/pattern there that you can present to Pages.

examples/apps/claims-example/src/container/claimsDataObject/claimsDataObject.ts
ChumpChief: I'm referring to aliasing as a concept, not just the specific ContainerRuntime implementation at the container level.

The primary motivating scenario for Claims is aliasing at a granular level where the existing container-level feature isn't a fit (as demonstrated in this example, and likely performed on load too - not just in response to user input), it's absolutely a scenario that needs an answer.

ChumpChief (approved):
Thank you for making the adjustments, should be good to merge.

There are a lot of followups here exploring more-realistic usage patterns.  For this feature I think the next priority should be analyzing the potential Pages usage to ensure we have a realistic picture of real usage (and incorporate any findings into this example along the way to prove them out) - e.g.:
* How do they safely add a Claims instance to data objects in existing documents
* When/how do they make new claims (on load, negotiated via OldestClientObserver, etc.)
* Will they want to move the singleton handle reference into the `root` vs. keep referencing it in the Claims long-term
* Is one Claims per-DataObject sufficient, or e.g. does it become cumbersome if they have deeply nested DDSs, high multiplicity of DDSs, etc.
* Will they have a need for enumeration (maybe related to the question of moving references into an already-enumerable final location)
    * If not enumerable, does this make it hard to find and "clean up" stale members that should be GC'd?
* Will they have a single-writer to the Claims (they own gatekeeping/proxying in the case that there are multiple places they want to invoke setting a claim), or will we need to facilitate that scenario (which may imply the need for additional API surface for safe usage e.g. `isClaimPending(key)`)

PR #27574, opened 2026-06-22
chore: remove deprecated ILoaderOptions.enableOfflineLoad

dannimad:
Removes the deprecated `enableOfflineLoad` field from `ILoaderOptions` in `@fluidframework/container-definitions` and the legacy `Fluid.Container.enableOfflineLoad` config-provider feature-gate read from the container loader. Offline load now defaults to enabled for interactive clients, overridable via `Fluid.Container.enableOfflineFull`.


ChumpChief (approved):

markfields:
@dannimad - doesn't this PR need a changeset?

packages/loader/container-loader/src/container.ts
```diff
@@ -963,9 +963,7 @@ export class Container
 
 		const offlineLoadEnabled =
 			this.isInteractiveClient &&
-			(this.mc.config.getBoolean("Fluid.Container.enableOfflineLoad") ??
-				this.mc.config.getBoolean("Fluid.Container.enableOfflineFull") ??
-				options.enableOfflineLoad !== false);
+			(this.mc.config.getBoolean("Fluid.Container.enableOfflineFull") ?? true);
```
markfields, line 966: It's a little curious that this defaults to `true` for the Loader layer but `false` for the runtime layer.  I think that given it's only expected to be used for testing, it's fine.  Also good to default to `true` for the slower-to-update loader layer.

Might be helpful to write down which FF release first had it default to true.

markfields (approved):

markfields (approved):

alexvy86 (approved):
Approving for docs

PR #27580, opened 2026-06-22
fix(tree): no auto-dispose of forks made during onChanged event

yann-achard-MS:
## Description

The "changed" event is emitted from a `TreeBranch` when a change is made to the branch.
Previously, when this event was fired due to a transaction being committed, it was possible to fork the branch in response to the "changed" event, but such a fork would be automatically disposed immediately after the event callback.
This was a bug. Such forks are no longer disposed automatically.

This PR also removes the [recently introduced](https://github.com/microsoft/FluidFramework/pull/27285) restriction that prevents forking in event callbacks. Forking is not an edit and therefore should not be prevented.

## Breaking Changes

An application that relied on the old behavior may now experience a memory leak if it does not explicitly dispose the forks.

### DIFF
--- a/packages/dds/tree/src/shared-tree/treeCheckout.ts
+++ b/packages/dds/tree/src/shared-tree/treeCheckout.ts
@@ -1202,7 +1202,6 @@ export class TreeCheckout implements ITreeCheckout {
 			throw new UsageError("A view cannot be forked while it has a pending transaction.");
 		}
 
-		this.editLock.checkUnlocked("Branching");
 		const branch = this.#transaction.activeBranch.fork();
 		const storedSchema = this.storedSchema.clone();
 		const forkBreaker = new Breakable("TreeCheckout");
--- a/packages/dds/tree/src/shared-tree/treeCheckout.ts
+++ b/packages/dds/tree/src/shared-tree/treeCheckout.ts
@@ -306,7 +306,6 @@ export function createTreeCheckout(
 		chunkCompressionStrategy?: TreeCompressionStrategy;
 		logger?: TelemetryLoggerExt;
 		breaker?: Breakable;
-		disposeForksAfterTransaction?: boolean;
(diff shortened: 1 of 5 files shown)


packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -658,9 +657,6 @@ export class TreeCheckout implements ITreeCheckout {
 		branch: SharedTreeBranch<SharedTreeEditBuilder, SharedTreeChange>,
 	): SquashingTransactionStack<SharedTreeEditBuilder, SharedTreeChange> {
 		return new SquashingTransactionStack(branch, this.mintRevisionTag, () => {
-			const disposeForks = this.disposeForksAfterTransaction
```
Josmithr, line 661: Should we add a regression test to ensure forks are not disposed in this scenario going forward?

Josmithr (approved):
Left 1 suggestion. Otherwise, looks good to me

packages/dds/tree/src/shared-tree/treeCheckout.ts
yann-achard-MS: I did add a test in treeCheckout.spec.ts

daesunp (commented):
Nit: There still seem to be uses of `this.disposeForksAfterTransaction` (for example in the constructor of TreeCheckout). Should these be removed as well?

yann-achard-MS:
> Nit: There still seem to be uses of `this.disposeForksAfterTransaction` (for example in the constructor of TreeCheckout). Should these be removed as well?

Thanks. I pushed new changes to remove it.

packages/dds/tree/src/shared-tree/treeCheckout.ts
Josmithr: 🤦‍♂️ I completely misread the diff there.

daesunp (approved):

PR #27583, opened 2026-06-22
Feat(tree) `TreeBranchAlpha.isMissingEditsFrom`

yann-achard-MS:
## Description

Adds a new method (`isMissingEditsFrom(branch: TreeBranch): boolean`)  to `TreeBranchAlpha`.
`isMissingEditsFrom` can be used to determine whether there are edits on the given `branch` that have not yet been merged into this branch.

## Breaking Changes

None

### DIFF
--- a/packages/dds/tree/src/simple-tree/api/tree.ts
+++ b/packages/dds/tree/src/simple-tree/api/tree.ts
@@ -342,6 +342,15 @@ export interface TreeBranchAlpha extends TreeBranch, TreeContextAlpha {
 	 * Update the tests and docs to match when that is done.
 	 */
 	applyChange(change: JsonCompatibleReadOnly): void;
+
+	/**
+	 * Returns true iff there are changes on the given branch that are not present on this branch.
+	 * @param branch - The branch to compare to.
+	 *
+	 * The new edits, if any, can be applied to this branch by {@link TreeBranch.rebaseOnto | rebasing this branch onto the given branch}
+	 * or by {@link TreeBranch.merge | merging the given branch into this branch}.
+	 */
+	isMissingEditsFrom(branch: TreeBranch): boolean;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -342,6 +342,15 @@ export interface TreeBranchAlpha extends TreeBranch, TreeContextAlpha {
 	 * Update the tests and docs to match when that is done.
 	 */
 	applyChange(change: JsonCompatibleReadOnly): void;
+
+	/**
+	 * Returns true iff there are changes on the given branch that are not present on this branch.
```
Josmithr: Nit: My preference for user-facing docs is to avoid mechanical statements like "Returns" in the summary component of a comment like this. I also think we should avoid abbreviations like "iff". My recommendation would be something like:

> Determines if there are changes on the given branch that are not present on this branch.

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -342,6 +342,15 @@ export interface TreeBranchAlpha extends TreeBranch, TreeContextAlpha {
...
+	 * @param branch - The branch to compare to.
+	 *
+	 * The new edits, if any, can be applied to this branch by {@link TreeBranch.rebaseOnto | rebasing this branch onto the given branch}
+	 * or by {@link TreeBranch.merge | merging the given branch into this branch}.
+	 */
+	isMissingEditsFrom(branch: TreeBranch): boolean;
```
Josmithr, line 355: Should we document that this is expected to throw an error if the two branches don't share a common ancestor?

Josmithr (approved):

Josmithr (approved):

Josmithr (approved):

PR #27584, opened 2026-06-23
perf(examples): Add devtools toggle button to text example 

brrichards:
## Description

The devtools SharedTree visualizer subscribes to each container's SharedTree "op" event and re-serializes the entire tree (exportVerbose → visualizeVerboseNodeFields) on every edit. On a per-character formatted document that work dominates CPU. This was done without eargerly without devtools being called. A button has been added to the top left of the text-editor app that enables/disables devtools (default disabled). 

Screenshot 2026-06-23 at 9 58 00 AM

Manually tested with trace testing. When disabled exportVerbose is not called and text edits at large doc sizes drop from 500ms+ to 9 ms
## Reviewer Guidance

### DIFF
--- a/examples/data-objects/text-editor/src/app.tsx
+++ b/examples/data-objects/text-editor/src/app.tsx
@@ -107,7 +107,7 @@ interface DualUserViews {
 	 * Properties for (re)initializing Devtools. Held so the UI can toggle Devtools on and off at runtime
 	 * (see {@link DevtoolsToggle}).
 	 */
-	devtoolsProps: DevtoolsProps;
+	devtoolsProps?: DevtoolsProps;
 }
 
 async function createAndAttachNewContainer(client: AzureClient): Promise<{
@@ -430,13 +430,16 @@ const UserPanel: FC<{
  * Button that enables/disables Fluid Devtools at runtime.
  */
 const DevtoolsToggle: FC<{
-	devtoolsProps: DevtoolsProps;
+	devtoolsProps: DevtoolsProps | undefined;
 }> = ({ devtoolsProps }) => {
 	// Devtools defaults to off
 	const [enabled, setEnabled] = useState(false);
 
 	// Handles initialization and cleanup of devtools instance
 	useEffect(() => {
+		if (devtoolsProps === undefined) {
+			return;
+		}
 		if (enabled) {
 			const instance = initializeDevtools(devtoolsProps);
 			return () => {
@@ -454,7 +457,7 @@ const DevtoolsToggle: FC<{
 			onClick={() => setEnabled((value) => !value)}
 			title={
 				enabled
-					? "Disable Fluid Devtools (recommended before capturing a performance trace)"
+					? "Disable Fluid Devtools (recommended before capturing a performance trace. Devtools visualizes every node on every edit)"
 					: "Enable Fluid Devtools"
 			}
 			style={{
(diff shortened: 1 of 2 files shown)


examples/data-objects/text-editor/src/app.tsx
```diff
@@ -415,6 +426,53 @@ const UserPanel: FC<{
...
+	}, [enabled, devtoolsProps]);
+
+	return (
+		<button
+			type="button"
+			onClick={() => setEnabled((value) => !value)}
+			title={
+				enabled
+					? "Disable Fluid Devtools (recommended before capturing a performance trace)"
+					: "Enable Fluid Devtools"
```
Josmithr: Nit: might be worth adding a warning about the perf implications 😅

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -99,6 +103,11 @@ interface DualUserViews {
 	user1: TreeViewAlpha<typeof TextEditorRoot>;
 	user2: TreeViewAlpha<typeof TextEditorRoot>;
 	containerId: string;
+	/**
+	 * Properties for (re)initializing Devtools. Held so the UI can toggle Devtools on and off at runtime
+	 * (see {@link DevtoolsToggle}).
+	 */
+	devtoolsProps: DevtoolsProps;
```
Josmithr: Nit: we could make this optional and not allow devtools toggling when not provided. Might simplify things a bit. E.g., tests could just not specify this.

Josmithr (approved):

examples/data-objects/text-editor/src/app.tsx
brrichards: Yeah good call! Updated explaining why. Let me know if you thought of putting it somewhere else

examples/data-objects/text-editor/src/app.tsx
brrichards: Made it optional. It did remove the need for the test file change

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -210,8 +219,9 @@ async function initFluid(): Promise<DualUserViews> {
 
 	console.log(`User 2 connected to document: ${containerId}`);
 
-	// Initialize Devtools
-	initializeDevtools({
+	//* Build the Devtools initialization props. Devtools starts disabled and is toggled on/off at runtime
```
Josmithr: Nit
```suggestion
	// Build the Devtools initialization props. Devtools starts disabled and is toggled on/off at runtime
```

examples/data-objects/text-editor/src/app.tsx
Josmithr: Oh, I was actually thinking of adding a `()` note on the enable button noting the implications. But either way is probably fine.

Josmithr (approved):

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -415,6 +426,56 @@ const UserPanel: FC<{
...
+			style={{
+				padding: "6px 12px",
+				borderRadius: "4px",
+				border: "1px solid #ccc",
+				background: enabled ? "#e6f4ea" : "#f5f5f5",
+				color: "#333",
+				fontSize: "13px",
+				fontWeight: 600,
+				cursor: "pointer",
+			}}
```
CraigMacomber, line 472: I don't think this button should need custom styling. What ever defaults we get from our CSS should probably be fine, and if not we should probably fix it there.

That said, this is how a lot of stuff in this example works, so I'll approve, but we should probably clean this up at some point.

CraigMacomber (approved):

examples/data-objects/text-editor/src/app.tsx
brrichards: That makes sense. If we do add the add/remove user views for side by side. It would be nice to clean this up before then (since that will add at least 2 more buttons)

PR #27588, opened 2026-06-23
Opt plain-text character array into incremental summarization

brrichards:
## Description

Enables Incremental Summariation for plain text. Plain text is contained in large uniform chunks, and will benefit from incremental summarization. Formatted is currently not stored in a way where incremental summarization will provide a benefit compared to the overhead cost of incremental summarization. 

## Testing
Single test in textDomain.spec.ts that confirms the plain text character array is opted into incremental summarization.

## Reviewer Guidance


packages/dds/tree/src/text/textDomain.ts
Josmithr, line 261: Now that users can leverage incremental summarization effectively here, it would probably be worth adding some docs in a `@remarks` block here explaining the benefits and how to enable it. Linking to the relevant APIs would be good too.

Josmithr (commented):
Left a suggestion for documentation we should add. It would also probably be worth adding a changeset document to this PR, since we're surfacing improved functionality that requires user opt-in. We can likely re-use much of the same documentation for both purposes.

.changeset/slick-carpets-kiss.md
```diff
@@ -0,0 +1,36 @@
...
+	FluidClientVersion,
+} from "@fluidframework/tree/alpha";
+
+const config = new TreeViewConfigurationAlpha({ schema: TextAsTree.Tree });
+
+const SharedTreeWithIncrementalSummaries = configuredSharedTreeAlpha({
+	forest: ForestTypeOptimized,
+	treeEncodeType: TreeCompressionStrategy.CompressedIncremental,
+	shouldEncodeIncrementally: incrementalEncodingPolicyForAllowedTypes(config),
+	minVersionForCollab: FluidClientVersion.v2_74,
```
Josmithr: Let's add inline comments here indicating which of these properties are required, and any more nuanced requirements. E.g., if there is a minimum `minVersionForCollab` that needs to be specified, let's note that the value can be set to that or higher.

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -259,6 +266,16 @@ export function processCharactersChangedDelta(
...
 /**
  * A collection of text related types, schema and utilities for working with text beyond the basic {@link SchemaStatics.string}.
+ * @remarks
+ * The character content of a {@link TextAsTree.(Tree:type)} is opted in to incremental summarization:
+ * its schema carries the {@link incrementalSummaryHint} metadata.
+ * When incremental summarization is enabled on the tree, character content which does not change between
+ * summaries is not re-encoded or re-uploaded to the service; instead the unchanged chunks are referenced by handle.
+ * For large texts which are edited in only a few places at a time, this can substantially reduce summary upload size
+ * and the CPU cost of producing each summary.
+ *
```
Josmithr: Nit: I think adding a quick, explicit note that this can be opted into via `configuredSharedTree` (with an API link) would be useful.

Josmithr (approved):

.changeset/slick-carpets-kiss.md
brrichards: updated the comments

packages/dds/tree/src/text/textDomain.ts
brrichards: Good call, I added a sentence with a link to `configuredSharedTree`

packages/dds/tree/src/text/textDomain.ts
```diff
@@ -180,7 +181,13 @@ export function charactersFromString(value: string): Iterable<string> {
 	return value;
 }
 
-class StringArray extends sf.array("StringArray", SchemaFactory.string) {
+class StringArray extends sf.arrayAlpha(
+	"StringArray",
+	// Opt the character content into incremental summary optimization.
+	sf.types([SchemaFactory.string], {
+		custom: { [incrementalSummaryHint]: true },
```
CraigMacomber: I was thinking we would just set this in the text editor schema above the TextNodes (both plain and formatted) so we would avoid reencoding the whole plain text when changing the formatted text for example.

Incremental summary chunks are pretty expensive, so even with our larger chunk sizes now, I'm not sure its a win (each one gets included in a top level list of all chunks in the document, complicates container level GC, and has to be referenced by a handle which has a non negligible size not to mention each incremental chunk itself has a lot of boilerplate wrapping it, including another copy of the needed shapes (not too bad for strings I guess). I also think we might not do a great job of encoding sequences at the root of chunks efficiently (I think they might all get the any shape encoder, adding per character overhead in this case).

Thus I think putting the opt in here is a good way to stress test incremental summary, but probably not a performance win in most cases.

CraigMacomber (approved):

CraigMacomber (approved):

PR #27590, opened 2026-06-23
fix(build-tools): Allow safe tsc project reuse in tsc-project-single-use policy

jason-ha:
Previously the tsc-project-single-use policy failed whenever a tsconfig project file was referenced by more than one script, even when the uses could not collide. This blocked valid patterns such as reusing a project with different command line options (e.g. a noEmit type-check that writes a distinct tsbuildinfo file).

The policy now resolves each tsc invocation's effective options and only fails when two scripts would write the same incremental build info (tsbuildinfo) file, which is the actual collision that corrupts parallel builds.

The build info and tsconfig resolution logic was extracted from the fluid-build tsc task into shared, exported functions in tscUtils (getResolvedTsConfig, getTsBuildInfoFileFromConfig, getTsBuildInfoFullPath, remapOutFile) so the policy check and the build task share a single implementation. Unit tests cover the extracted functions, including extends-based option merging and command line overrides.

This policy change will allow secondary tsconfig project file use for alternate checks, like verifying that both exactOptionalPropertyType true and false settings work for production code, without authoring a host of additional files. And example build entry could be:
```json
"check:inexactOptionalPropertyTypes": "tsc --project ./tsconfig.json --noEmit --exactOptionalPropertyTypes false --skipLibCheck --emitDeclarationOnly false --tsBuildInfoFile ./tsconfig.inexactOptionalPropertyTypes.tsbuildinfo",
```

### DIFF
--- a/build-tools/packages/build-cli/src/library/repoPolicyCheck/fluidBuildTasks.ts
+++ b/build-tools/packages/build-cli/src/library/repoPolicyCheck/fluidBuildTasks.ts
@@ -865,20 +865,20 @@ export const handlers: Handler[] = [
 	},
 	{
 		/**
-		 * Checks that all tsc project files (tsconfig.json), are only used once as the main
-		 * configuration among scripts.
-		 * Multiple uses may indicate a collision during build.
+		 * Checks that tsc project files (tsconfig.json) used by multiple scripts do not collide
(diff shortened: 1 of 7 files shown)


MarioJGMsoft (approved):
LGTM!

PR #27591, opened 2026-06-23
feat(devtools): Register event subscriptions on-demand

Josmithr:
Previously, devtools would (upon initialization) immediately register on-op event listeners on all provided root DDSs and immediately begin broadcasting data visualization messages on the window (regardless of whether or not the browser extension was running and listening).

This PR updates devtools to register these listeners lazily on demand. This leverages the existing laziness that was implemented for *nested* DDSs, such that we only register event listeners when we receive the `GET_DATA_VISAUALIZATION` message from the browser extension.

A corresponding `CLOSE_DATA_VISUALIZATION` message has been added, which the extension now uses to indicate when it is done needing data from the client. The client leverages this alongside simple ref-counting to unsubscribe DDS listeners when there is no longer interest in data.

- Note: this PR does not address the potential listener leak that can occur if a consumer requests data events but fails to send the corresponding close message. But since the code previously never de-registered listeners once initially registered, this is still a strict improvement. We can follow up on this in the future as needed.

Note that there are no cross-layer compat issues here between the devtools client library and the extension.
| client ↓ \ extension → | Old extension (doesn't send `CLOSE` message) | New extension (sends `CLOSE` message) |
| --- | --- | --- |
| **Old client** (eager broadcast, no `CLOSE` handler) | **Baseline.** Eager broadcasting of all reachable roots; subscriptions never released. Works as before. | **Compatible.** Old core ignores the unrecognized `CLOSE_DATA_VISUALIZATION`. Still eager; view filters incoming `DATA_VISUALIZATION` by `fluidObjectId` and detaches handlers on unmount, so extra/stale broadcasts are harmless. No regression. |
| **New client** (on-demand, ref-counted, `CLOSE` handler) | **Compatible, strict improvement.** Nothing broadcasts until the view sends `GET_DATA_VISUALIZATION`. Old view never sends `CLOSE`, so each subscribed node stays subscribed after collapse — **subscription leak**, but behavior is no worse than the Old-core baseline (which monitored everything unconditionally). | **Fully intended behavior.** On-demand subscribe on expand/mount, balanced unsubscribe on collapse/unmount. Monitoring (and broadcasting) occurs only while a view is actively displaying an object. No leak (assuming proper view teardown). |


packages/tools/devtools/devtools-core/src/data-visualization/DataVisualization.ts
```diff
@@ -341,8 +381,11 @@ export class DataVisualizerGraph
  *
  * A visual representation can be requested via {@link VisualizerNode.render}.
  *
- * Additionally, whenever the associated `ISharedObject` is updated (i.e. whenever its "op" event is emitted),
- * an updated visual tree will be emitted via this object's {@link SharedObjectListenerEvents | "update" event}.
+ * Additionally, while at least one consumer is {@link VisualizerNode.subscribe | subscribed}, whenever the associated
```
jikim-msft, line 384: NIT: Should we repalce "whenever" to "when subscribed" mirroring the class-level JSDOC from above?

jikim-msft:
NIT: While renaming existing types to `DataVisualizerEvents`, there's a dangling reference to `{@link SharedObjectListenerEvents | "update"} event` at Line 511 of `DataVisualization.ts`. Might as well fix it here for consistency

@Josmithr

packages/tools/devtools/devtools-core/src/BaseDevtools.ts
```diff
@@ -515,40 +522,23 @@ export abstract class BaseDevtools<TContainer extends DecomposedContainer>
 		return this._disposed;
 	}
 
-	/**
-	 * Initialize data visualization monitoring immediately to ensure event listeners are set up
-	 * and console logs appear even when devtools UI is not open.
-	 */
-	private async initializeDataVisualizationMonitoring(): Promise<void> {
```
jikim-msft, line 522: Removing the startup monitoring here — together with moving the `op` listener out of `VisualizerNode`'s constructor and into `subscribe()` — means the client now only attaches an `op` listener (and therefore only broadcasts `DataVisualization` updates) for objects a consumer has explicitly subscribed to via `GetDataVisualization`.

I think that regresses the sidebar "container has changes" indicator. The only code path that emits `reason: DataChanged` is `dataUpdateHandler`, which fires off the `DataVisualizerGraph` "update" event — and after this change that event only fires while a `VisualizerNode` has an active subscription. But `Menu.tsx` drives the indicator purely from inbound `DataVisualization` messages with `reason === DataChanged`, while only ever sending `GetContainerState`, never `GetDataVisualization`:

https://github.com/microsoft/FluidFramework/blob/a072ce262a4667fec982b5f5c5370c393f6aae6c/packages/tools/devtools/devtools-view/src/components/Menu.tsx#L765-L788

Since `FluidHandleView` (the detail panel) is the only sender of `GetDataVisualization`, no `op` listener exists for a container's objects unless the user has that specific object's detail view open — so the "has changes" indicator will no longer light up for anything the user hasn't drilled into. Previously this method rendered every root "to set up event listeners on all shared objects" at startup, so the indicator fired regardless.

The new `DataVisualizerGraph` test "Broadcasts updates while subscribed and stops after unsubscribe" actually captures the mechanism: an op emitted before `subscribe()` yields `updateCount === 0`. For the objects the menu cares about (which are never subscribed), that zero-broadcast path is exactly the regression.

Is this intended / handled in a follow-up? I didn't see the indicator called out in the compatibility matrix. If it's meant to keep working, it likely needs an always-on signal independent of the detail-panel subscription — e.g. the menu subscribing per container, or a lightweight "changed" notification decoupled from full visualization.

Josmithr: Interesting. I'd forgotten about that indicator - thanks for the callout. I wish it was easier to write integration tests for these scenarios 😟

So, I think this is a regression we probably want to take, though it is definitely worth considering how we should update the change indicator in response. My main reasons for this are:

1. Performance - the fact that apps using devtools immediately pay a non-trivial cost in listening for DDS changes and broadcasting messages on the window (even when no one is listening) is a deterrent for using them at all. We were forced to disable the devtools (by default) in our text demo because the performance hit was so substantial.

2. This issue already exists for any nested DDSs - in our visualizer graph (currently), we eagerly bind listeners and post updates upon devtools initialization, **but only for the root DDSs declared in the input set**. Any other DDSs referenced by handle already don't publish updates until the view (extension) requests them.
Consider the following setup: devtools is configured with a single SharedDirectory which contains a single child item: a handle to a SharedTree. The corresponding application exclusively uses the SharedTree, and never directly changes its directory contents. Our change indicator would already never fire in this case.

Question for you: was the change indicator a requirement for anything? Accessibility maybe? Or was it just a nice-to-have UX improvement? If it wasn't a hard requirement, I personally think the performance considerations here are more valuable. To be clear: I really like this feature; I just personally think the performance implications are more important 😕

jikim-msft: Thanks for the detailed write up. I also agree with the PRs overall direction to emphasize more on the performance side (Paying for `op`  listeners + window broadcasts on every root DDS regardless of whether anyone's listening is a real cost). 

To your question: **no**, it wasn't a hard requirement. I believe it was purely a nice-to-have UX cue. The indicator is just a visual treatment on the menu item (yellow background + glow), with no accessibility hook or downstream dependency tied to it, so nothing breaks functionally by losing it.

Overall I'm on board with taking the regression in this PR. I think we should just make sure we don't silently drop it so maybe leave a doc or create a follow up item for the clean up in the future.

Thanks!

jikim-msft (approved):

packages/tools/devtools/devtools-core/src/BaseDevtools.ts
Josmithr: Okay, thank you. Given that, I'll take a stab at cleaning up the indicator logic - I'd rather remove it than leave it in an only half-working state.

Josmithr: Updated.

jikim-msft (approved):

PR #27594, opened 2026-06-24
feat(website): disable cookies being imported from libraries

MarioJGMsoft:
## Description

There is an ongoing effort to remove non-essential cookies from the Fluid Framework website.

There were some cookies being created from libraries used for the application insights. This change ensures that we only create cookies that we actually need.

The cookies that are being disabled are `ai_session` and `ai_user`. The existing `userId` cookie used for analytics will remain unaffected.

Fixes [AB#73908](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73908)

## Reviewer Guidance

As far as I'm aware we don't actively use the cookies that are being removed, but if they are actually necessary, please let me know.

### DIFF
--- a/website/src/appInsights.ts
+++ b/website/src/appInsights.ts
@@ -21,6 +21,7 @@ if (typeof window !== "undefined" && window.location.hostname !== "localhost") {
 		appInsights = new ApplicationInsights({
 			config: {
 				connectionString: `InstrumentationKey=${instrumentationKey}`,
+				disableCookieUsage: true,
 				enableAutoRouteTracking: true,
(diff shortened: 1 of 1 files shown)


WayneFerrao (approved):
LGTM! Afaik, AppInsights only uses userID for the telemetry dashboard.

steffenloesch (approved):

MarioJGMsoft:
The cookies were actually identified to be essential, that means we have no problems keeping them.

PR #27598, opened 2026-06-25
Fix(tree): no auto-dispose of forks made during onChanged event

yann-achard-MS:
This is a port of the changes made in #27580.

## Description

The "changed" event is emitted from a `TreeBranch` when a change is made to the branch.
Previously, when this event was fired due to a transaction being committed, it was possible to fork the branch in response to the "changed" event, but such a fork would be automatically disposed immediately after the event callback.
This was a bug. Such forks are no longer disposed automatically.

This PR also removes the [recently introduced](https://github.com/microsoft/FluidFramework/pull/27285) restriction that prevents forking in event callbacks. Forking is not an edit and therefore should not be prevented.

## Breaking Changes

An application that relied on the old behavior may now experience a memory leak if it does not explicitly dispose the forks.

### DIFF
--- a/.changeset/quick-mangos-hide.md
+++ b/.changeset/quick-mangos-hide.md
@@ -1,0 +1,10 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": fix
+---
+Forks created on "changed" event are no longer auto-disposed
+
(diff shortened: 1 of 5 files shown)


frankmueller-msft (approved):

jason-ha:
Changes are available in 2.110 which has been integrated into Pages' repo

PR #27604, opened 2026-06-25
test(build-cli): Fix tests so they pass even without a full git checkout

CraigMacomber:
## Description

Fix the one test which fails if a shallow git clone with no tags is done.

This makes the test much less fragile, allowing it to work the same in more environments.
This is also a prerequisite for optimizing CI to do smaller checkouts.

Change split out from https://github.com/microsoft/FluidFramework/pull/27342

## Reviewer Guidance


tylerbutler (approved):
LGTM, though my reviews shouldn't count for much these days. Thanks for cleaning this up. A much younger tylerbu apologizes for the terrible testing. 😄

PR #27607, opened 2026-06-26
Safer typing in translateRequestFields

CraigMacomber:
## Description

Our code scanner has flagged this as a Critical issue for over 3 years. It seems to be a false positive. This change adjusts the code so it can have more robust typing, avoid the need for a cast, and hopefully make it clear to the scanner that this is actuall safe.

## Reviewer Guidance

### DIFF
--- a/server/routerlicious/packages/services-shared/src/restLessServer.ts
+++ b/server/routerlicious/packages/services-shared/src/restLessServer.ts
@@ -121,9 +121,7 @@ export class RestLessServer {
 		(request as any)._body = true;
 		const body: string | undefined = bodyField instanceof Array ? bodyField[0] : bodyField;
 		request.body = body;
-		request.headers["content-length"] = body
-			? `${body.length}`
-			: undefined;
+		request.headers["content-length"] = body ? `${body.length}` : undefined;
 	}
 
 	private parseRequestBody(request: IncomingMessageEx): void {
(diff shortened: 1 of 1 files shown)


server/routerlicious/packages/services-shared/src/restLessServer.ts
```diff
@@ -119,9 +119,10 @@ export class RestLessServer {
 		const bodyField: RequestField = fields[RestLessFieldNames.Body];
 		// Tell body-parser middleware not to parse the body
 		(request as any)._body = true;
-		request.body = bodyField instanceof Array ? bodyField[0] : bodyField;
-		request.headers["content-length"] = request.body
-			? `${(request.body as string).length}`
+		const body: string | undefined = bodyField instanceof Array ? bodyField[0] : bodyField;
```
jason-ha, line 122: `body` needs `: string | undefined` because insufficient typing to know those are the only possibilities?

jason-ha (approved):

jason-ha (approved):

server/routerlicious/packages/services-shared/src/restLessServer.ts
CraigMacomber: That can be inferred. I just included it explicitly for additional clarity. What I suspect matters here is simplifying things by introducing the new variable so I can remove the cast, but I figured as I'm trying to fix a false positive static analysis, more explicitness might help.

jason-ha: It is okay. I have been finding more code with maintenance burdens from explicit types that don't matter when something is updated. Either the explicit now incorrect type needs updated or removed. Removed is easier.
If a human needs told that those are the two possibilities, then it is good to have the type and take such maintenance burdens.

PR #27610, opened 2026-06-26
feat(client-tree): transaction post-processing option infrastructure

jason-ha:
`RunTransactionParams` now accepts an optional `postProcessor` (used by `runTransaction` and `runTransactionAsync`). When specified, the post-processor is invoked when the transaction is committed, allowing for additional processing of the transaction's edits.

The option is supplied when the transaction starts and is opt-in: when it is omitted the existing behavior is preserved.

Note: a minimization post-processor will be added in a future change.

New test cases:
```text
  sharedTreeView
    Transactions
      ✔ converts a post-processor and injects it as the post-processor at transaction start
      ✔ injects no post-processor when no params are provided to runTransaction
      ✔ injects the post-processor at start for the outermost transaction only
      ✔ converts and injects a post-processor through runTransactionAsync
      ✔ does not commit (and so does not apply the post-processor) when rolled back

  SquashingTransactionStacks
    transaction post-processing
      ✔ is invoked with the squashed change when started with a post-processor
      ✔ is not invoked when the transaction is empty
      ✔ is not invoked when the transaction is aborted
      ✔ invokes an "outermost" post-processor once for the outermost transaction started
      ✔ invokes the "outermost" post-processor for an inner transaction when the outer has no post-processor
      ✔ invokes the "outermost" post-processor for an inner transaction when the outer has different post-processor
      ✔ invokes an "always" post-processor at every transaction commit that supplied it
      ✔ invokes an "always" post-processor only where it was supplied
      ✔ invokes an "outermost" post-processor once per sibling nested transaction that supplied it
      ✔ invokes an "outermost" post-processor once when the outer and both sibling nested transactions supply it
      ✔ invokes a mix of "outermost" and "always" post-processors in the expected order
      ✔ applies the change returned by the post-processor to the branch
```


.changeset/strong-pets-marry.md
```diff
@@ -0,0 +1,11 @@
+---
+"@fluidframework/tree": minor
+"__section": tree
+---
+Add an opt-in postProcessor option when running a transaction
```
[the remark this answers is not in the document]
jason-ha: Yes, I did. I'd put the term in backticks, but changeset style guidelines prefer no backticks in titles. :/

packages/dds/tree/src/simple-tree/api/transactionTypes.ts
```diff
@@ -125,6 +127,21 @@ export type TransactionResult =
...
+ * A type-erased function that post-processes the change produced when a transaction is committed.
+ * @remarks
+ * Supply one via {@link RunTransactionParams.postProcessor} to "minimize" the change that a transaction produces, so that
+ * the resulting squashed change contains no extraneous information (e.g. data for nodes that were both created and removed
+ * within the transaction, or changes whose effects cancel out to nothing).
+ *
+ * This type intentionally hides its internal representation: its concrete shape is an implementation detail of
+ * `@fluidframework/tree`. Obtain a value of this type from a `@fluidframework/tree` helper rather than constructing one
+ * directly.
+ * @system @sealed @alpha
```
yann-achard-MS: >@system [](http://example.com/codeflow?start=3&length=7)

What does `@system` do? #Closed

packages/dds/tree/src/simple-tree/api/transactionTypes.ts
```diff
@@ -125,6 +127,21 @@ export type TransactionResult =
 	| Omit<TransactionResultSuccess<unknown>, "value">
 	| Omit<TransactionResultFailed<unknown>, "value">;
 
+/**
+ * A type-erased function that post-processes the change produced when a transaction is committed.
+ * @remarks
+ * Supply one via {@link RunTransactionParams.postProcessor} to "minimize" the change that a transaction produces, so that
+ * the resulting squashed change contains no extraneous information (e.g. data for nodes that were both created and removed
+ * within the transaction, or changes whose effects cancel out to nothing).
```
yann-achard-MS: Nit: I thought the minimization was only one possible usage/effect for such a interface. Should this be rephrased as an exampled instead? #Closed

packages/dds/tree/src/shared-tree/transactionPostProcessor.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import type { SharedTreeChange } from "./sharedTreeChangeTypes.js";
+
+/**
+ * The internal representation of a {@link TransactionPostProcessor}: a {@link ChangeProcessor} over
+ * {@link SharedTreeChange}.
+ * @remarks This is the (non-type-erased) form used internally to apply a transaction's post-processor.
+ */
+export type TransactionChangeProcessor = ChangeProcessor<SharedTreeChange>;
+
```
yann-achard-MS, line 17: The naming scheme feel chaotic: instantiating the `ChangeProcessor` type with a `SharedTreeChange` type parameter adds the "Transaction" as prefix despite the fact that `ChangeProcessor` was already about transactions.

How about:
`TransactionPostProcessor` stays the same
`ChangeProcessor` -> `TransactionPostProcessorInternal`
`TransactionChangeProcessor` -> `SharedTreeTransactionPostProcessorInternal`

OR

`TransactionPostProcessor` -> `OpaqueTransactionPostProcessor`
`ChangeProcessor` -> `TransactionPostProcessor`
`TransactionChangeProcessor` -> `SharedTreeTransactionPostProcessor` #Closed

packages/dds/tree/src/shared-tree-core/transaction.ts
```diff
@@ -216,6 +224,60 @@ export type OnPopWithViewUpdate<TChange> = (
...
 
+/**
+ * Specifies when a {@link ChangeProcessor} should be invoked relative to nested transactions that supply it.
+ */
+export enum ChangeProcessorApplicability {
+	/**
+	 * Invoke the processor only once, when the outermost transaction that supplied it is committed.
+	 * @remarks Supplying the same processor on a transaction already enclosed by one that supplied it has no
+	 * additional effect. Note that two sibling (sequential) nested transactions are each "outermost" within their own
+	 * subtree, so each will invoke the processor.
```
yann-achard-MS: >the processor. [](http://example.com/codeflow?start=33&length=14)

"the" makes it sound as they would invoke the same one. How about "their supplied processor"? #Closed

packages/dds/tree/src/shared-tree-core/transaction.ts
```diff
@@ -216,6 +224,60 @@ export type OnPopWithViewUpdate<TChange> = (
 	viewUpdate: TChange | undefined,
 ) => void;
 
+/**
+ * Specifies when a {@link ChangeProcessor} should be invoked relative to nested transactions that supply it.
+ */
```
yann-achard-MS, line 231: I think the docs on this enum and its members are technically correct but very hard to correctly understand. I think this is because the parts of it that jump out/are easiest to understand lead the reader towards a plausible but invalid interpretation of what is being said. For example, the "when" in "Specifies when a ChangeProcessor should be invoked" starts building the (misguided) expectation that a single ChangeProcessorApplicability value could have implications for an unbounded number of transactions frames. The parts that are meant to rescue the reader from this misunderstanding (e.g. "...relative to nested transactions that supply it") fall short because they are hard to grasp until you actually understand the whole thing (I had to go look at the implementation). I think the reason they are so hard to grasp is because they strive to explain the semantics of single values by relating it to a pattern of transactions and what was passed to them. It doesn't help that each such transaction in the pattern may itself be passed some applicability value (which doesn't matter here, but it's not clear at the outset that it doesn't. IOW, the current explanations go top-down in terms of emergence, when they could instead go bottom-up.

Here's what I propose:
1. In the top-level enum doc, spell out that the applicability defines whether the processor should be invoked for the specific (sub)transaction that supplies it. It does not affect the applicability of other (sub-)transactions be they parent, children, sibling, or other, and no matter what processor was supplied to them. Leave out anything about relativity to other transactions.
2. Rename Outermost to IfOutermost.
3. Rephrase the docs for the two options in terms of what was supplied to wrapping/above transactions:
- IfOutermost: Invoke the processor only if no other transaction above it has supplied the same processor.
- Always: Invoke the processor regardless of whether other transactions above it have supplied the same processor.
 #Closed

packages/dds/tree/src/simple-tree/api/transactionTypes.ts
```diff
@@ -146,4 +163,18 @@ export interface RunTransactionParams {
...
+	/**
+	 * An optional {@link TransactionPostProcessor | post-processor} applied to the change produced when this transaction is
+	 * committed, in order to "minimize" it so that the resulting squashed change contains no extraneous information.
+	 * @remarks
+	 * When omitted, the transaction's edits are squashed without any such processing (the existing behavior).
+	 *
+	 * The post-processor is applied once, when the outermost transaction that was started with one is committed. Supplying
+	 * a post-processor on a nested transaction that is already enclosed by a transaction started with one has no additional
+	 * effect.
+	 *
```
yann-achard-MS, line 183: This doesn't seem to match the implementation: you could pass a different processor, or pass a processor whose applicability was `Always` and it would have an effect.

I would leave that out. Each individual processor we export should come with its own documentation about when it makes sense to nest it. #Closed

packages/dds/tree/src/test/shared-tree-core/transaction.spec.ts
```diff
@@ -362,6 +364,389 @@ describe("SquashingTransactionStacks", () => {
...
+			editBranch(transaction.activeBranch, "A");
+			transaction.start({ postProcessor: innerOutermost }); // inner transaction with a post-processor
+			editBranch(transaction.activeBranch, "B");
+
+			// Act
+			transaction.commit(); // inner commit
+
+			// Verify
+			// The inner transaction is the outermost one that supplied a post-processor, so it is post-processed when committed.
+			assert.equal(received.length, 1);
```
yann-achard-MS, line 536: This verify comment and the one below both need updating #Closed

yann-achard-MS (changes requested):
:clock1:

packages/dds/tree/src/simple-tree/api/transactionTypes.ts
jason-ha: Thanks - comment is stale. (I evolved this from a change set that was minimize only.) Will update.

packages/dds/tree/src/simple-tree/api/transactionTypes.ts
jason-ha: https://fluidframework.com/docs/build/releases-and-apitags#api-support-levels
It basically means that customer should avoid direct use - really it means don't look at it and really don't depend on it staying exactly as it is. We might change it at any time.
Here we just need it to be a self-consistent type.

packages/dds/tree/src/shared-tree/transactionPostProcessor.ts
jason-ha: I was thinking that `ChangeProcessor` was more generic concept we are expressing internally.
But `TransactionChangeProcessor` is a bit funky.
In general we should have descriptive names for what things are/do rather than how they are used.
`TransactionChangeProcessor` is a bit of an exception as it is an alias for a specific use. So we could push it further toward intended use and rename to `TransactionPostProcessorInternal` OR walk back to what it is `SharedTreeChangeProcessor`.

`SharedTreeChangeProcessor` doesn't seem name-wise an improvement over `ChangeProcessor<SharedTreeChange>` but does have the advantage of not requiring two imports.
How about `TransactionChangeProcessor` -> `TransactionPostProcessorInternal` and others remain as they are?

packages/dds/tree/src/simple-tree/api/transactionTypes.ts
jason-ha: (I guess I needed to triple check the agent when I told it to get rid of the stale comments about X.)

packages/dds/tree/src/shared-tree-core/transaction.ts
jason-ha: I think I will try to rephrase with less transaction focus and more of a reusable concept.

packages/dds/tree/src/test/shared-tree-core/transaction.spec.ts
```diff
@@ -362,6 +364,391 @@ describe("SquashingTransactionStacks", () => {
...
+			editBranch(transaction.activeBranch, "A");
+			transaction.start({ postProcessor: innerOutermost }); // inner transaction with a post-processor
+			editBranch(transaction.activeBranch, "B");
+
+			// Act
+			transaction.commit(); // inner commit
+
+			// Verify
+			// The inner transaction is the outermost one that supplied specific
+			// post-processor, so it is post-processed when committed.
```
yann-achard-MS: "that supplied that specific post-processor"? #Closed

packages/dds/tree/src/test/shared-tree-core/transaction.spec.ts
```diff
@@ -362,6 +364,391 @@ describe("SquashingTransactionStacks", () => {
...
+				["inner"],
+			);
+
+			// Act
+			transaction.commit(); // outer commit
+
+			// Verify
+			// The outermost transaction supplied unique post-processor, so
+			// committing outer does invoke post-processor.
+			assert.equal(received.length, 2);
```
yann-achard-MS, line 548: Grammar (missing some articles). #Closed

packages/dds/tree/src/shared-tree/transactionPostProcessor.ts
yann-achard-MS: Deal!

yann-achard-MS (approved):
Approved with comments.

.changeset/strong-pets-marry.md
```diff
@@ -0,0 +1,11 @@
+---
+"@fluidframework/tree": minor
```
CraigMacomber, line 2: As the tree APIs are reexported from fluid-framework, which is where we would like new apps to consume them, it should be included in the list if impacted packages.

```suggestion
"@fluidframework/tree": minor
"fluid-framework": minor
```

CraigMacomber (approved):
Changeset looks good for docs.

PR #27612, opened 2026-06-29
ci(docs): drop legacy-account uploads, write docs storage to Torus only

RishhiB:
## What

Follow-up to #27503. Removes the `(legacy)` dual-write upload tasks so the docs storage uploads go **only** to the new Torus account (`fluid-docs-torus` / `$(STORAGE_ACCOUNT_NEW)` = `fluidframeworkcdn`).

- `publish-api-model-artifact.yml`: 6 upload tasks → 3 (drop the 3 legacy ones)
- `templates/include-upload-release-reports.yml`: 2 upload tasks → 1 (drop the legacy one; keep the `always()` cleanup step)

Remaining tasks renamed back to their original names (the `(Torus)` qualifier is redundant now).

## Why

The AFD origin behind `storage.fluidframework.com` was cut over to the `fluidframeworkcdn` (Torus) account via managed identity on **2026-06-26** and verified in production (both the FF "Website validation" and office-bohemia "External Partner to Loop Integration" consumer pipelines are green; endpoint serves 200s). The old MSIT `fluidframework` account is no longer the CDN origin, so dual-writing to it is dead weight. This completes SFI task **#47248** (MI auth, no SAS).

## Note

The old account is now write-free from these pipelines and can be decommissioned separately (it still holds non-served archives: `$web`, historical `api-extractor` SHA tarballs, `storybook`).

### DIFF
--- a/tools/pipelines/publish-api-model-artifact.yml
+++ b/tools/pipelines/publish-api-model-artifact.yml
@@ -208,22 +208,8 @@ stages:
                   verbose: true # Optional
                   #quiet: # Optional
 
-              # During the storage migration we dual-write each artifact to the legacy account
-              # (fluid-docs / STORAGE_ACCOUNT, served by the current AFD origin) and the new Torus
-              # account (fluid-docs-torus / STORAGE_ACCOUNT_NEW). Once the AFD origin is repointed and
-              # verified, the legacy tasks can be removed.
               - task: AzureCLI@2
-                displayName: 'Upload JSON (legacy)'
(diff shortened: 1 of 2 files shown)


Abe27342 (approved):

Abe27342:
would be nice to do one last cleanup and get the variable name back to just STORAGE_ACCOUNT again (you'll have to add that var to ADO, then commit the PR, then after next release or so you can delete the old one)

Abe27342 (approved):

PR #27614, opened 2026-06-29
More more settings to workspace files

CraigMacomber:
## Description

More settings to pnpm-workspace where possible. Reduces diff needed to move to pnpm 11.

Changes split out from https://github.com/microsoft/FluidFramework/pull/27510

## Reviewer Guidance

The review process is outlined on [this wiki page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

PR #27618, opened 2026-06-29
Update how we do pnpm config in tools

CraigMacomber:
## Description

Continuation of https://github.com/microsoft/FluidFramework/pull/27614 to the tools directory.

## Reviewer Guidance

### DIFF
--- a/tools/api-markdown-documenter/pnpm-workspace.yaml
+++ b/tools/api-markdown-documenter/pnpm-workspace.yaml
@@ -7,13 +7,6 @@
 packages:
   - "."
 
-# These entries must be duplicated from package.json due to a pnpm bug where onlyBuiltDependencies
-# in package.json is not respected when the workspace has its own lockfile.
-# See: https://github.com/pnpm/pnpm/issues/9082
-onlyBuiltDependencies:
-  - '@biomejs/biome'
-  - unrs-resolver
-
 # Supply chain security settings - see /DEV.md for documentation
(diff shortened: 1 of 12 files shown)


tools/api-markdown-documenter/pnpm-workspace.yaml
```diff
@@ -46,3 +39,39 @@ trustPolicyExclude:
   # repo); provenance attestation was lost on a subsequent 6.x release.
   - "undici-types@6.21.0"
 strictDepBuilds: true
+
+engineStrict: true
+frozenLockfile: true
+strictPeerDependencies: true
+updateNotifier: false
```
MarioJGMsoft, line 46: Is there a reason why you didn't keep the `# Disable pnpm update notifications since we use corepack to install package managers` from the .npmrc file?

MarioJGMsoft (approved):
Overall looks good, just curious about the comment I left.

tools/api-markdown-documenter/pnpm-workspace.yaml
CraigMacomber: Mostly it just got lost in translation, but didn't seem high enough value to restore, especially given we have like 20 copies of this same setting which would need the docs, and you can just look up what it means at https://pnpm.io/settings#updatenotifier (same for all the other options here). Interestingly this setting doesn't even work due to https://github.com/pnpm/pnpm/issues/12731 though.

PR #27620, opened 2026-06-30
git ignore .npmrc

CraigMacomber:
## Description

git ignore .npmrc

## Reviewer Guidance


packages/test/test-version-utils/compat-workspaces/full/.npmrc
```diff
@@ -1 +0,0 @@
-node-linker=isolated
```
CraigMacomber, line 1: This is the default for this setting (see https://pnpm.io/settings#nodelinker ) so it does nothing. Thus can safely be removed, removing the last versioned npmrc file.

Abe27342 (approved):

PR #27627, opened 2026-06-30
Support for webpack scenarios and fixes for bundleAnalysisRepo (--webpack-dir / fetch-on-reuse)

TommyBrosman:
This change consists of two improvements to the `bundleAnalysisRepo` tooling in `build-cli`, one of which enables webpack "scenarios."

## Description

Webpack scenarios are a feature that enables the analysis of webpack configurations and specific sets of imports that are designed to match real use cases. This differs from our typical approach of using a single webpack config for repo-wide analysis. New scenarios are defined under /examples/utils/bundle-size-tests/scenarios.

1. **`--webpack-dir` flag**: Decouples the directory that webpack runs in (and where `analyzer.json` / `compareBundlesOutput` are collected) from `--package-dir` (where `build:compile` runs). This lets a scenario subdirectory that only contains a webpack config be built and collected correctly. Previously `npm run webpack` in a config-less subdir walked up to the parent `package.json` and built the wrong config. The flag is added to `generate bundleAnalysisRepo`, `generate bundleAnalysisReposWithComparison`, and `check bundleAnalysisReposComparison` (defaulting to `--package-dir`, so existing usage is unchanged).

2. **Fetch-on-reuse for the inner base-repo clone**: When an inner `base-repo` clone already exists, `ensureInnerRepoAtRevision` now runs `git fetch origin --no-tags --prune` before checkout, bringing a reused clone to parity with a fresh one. This fixes the stale-clone `fatal: unable to read tree <sha>` failure for merge-base / revision SHAs created after the original clone, without forcing a full re-clone.

Also adds an `encapsulated-no-tree` bundle scenario under `examples/utils/bundle-size-tests` used to exercise and validate both features end-to-end.

## Reviewer Guidance

- What kind of documentation would you like to see around defining scenarios or individual scenarios?

## Todo

- [x] Ensure the webpack configs in scenarios can be linted.
  - I ended up going the CJS route. webpack-bundle-analyzer has no types and making the config a CJS seemed better than enabling skipLibCheck for the whole package.


build-tools/packages/build-cli/src/commands/check/bundleAnalysisReposComparison.ts
```diff
@@ -41,12 +41,19 @@ export default class CheckBundleAnalysisReposComparison extends BaseCommand<
 				"'current_<epoch>').",
 			default: "current",
 		}),
+		"webpack-dir": Flags.string({
+			description:
+				"Directory whose compareBundlesOutput holds the bundle stats to compare. Matches the " +
```
alexvy86: Nit
```suggestion
				"Directory whose compareBundlesOutput subfolder holds the bundle stats to compare. Matches the " +
```

examples/utils/bundle-size-tests/scenarios/encapsulated-no-tree/webpack.config.cjs
```diff
@@ -0,0 +1,81 @@
...
+				},
+			}),
+		],
+		usedExports: true,
+	},
+	output: {
+		filename: bundleName,
+		library: {
+			name: "encapsulatedNoTree",
+			type: "jsonp",
```
alexvy86: Deliberate `jsonp`? Most of our webpack configs don't specify a type; I don't know enough about webpack to say what impact this has on the output bundle or if it matters for the analysis we're doing on them.

build-tools/packages/build-cli/src/library/bundleSize/collectBundle.ts
```diff
@@ -210,15 +218,27 @@ function cleanWorkspace(repoRoot: string): void {
...
+	if (resolve(webpackRoot) === resolve(packageRoot)) {
+		// Same directory: run the package's own `webpack` script.
+		run("npm run webpack", webpackRoot);
+	} else {
+		// The webpack directory differs from the package directory (e.g. a scenario with its own
+		// webpack config but no package.json). `npm run webpack` would walk up to the package's
+		// package.json and run with cwd at the package root, building the wrong config. Invoke
+		// webpack directly with cwd = webpackRoot so its own (auto-detected) webpack.config is used
+		// and process.cwd()-relative outputs land here.
+		run("npx webpack", webpackRoot);
```
alexvy86: Would it be possible to pass a flag instead? I don't recall if npx will reuse the webpack executable that's already on disk or if it would dynamically download the latest version, which we probably want to avoid.

alexvy86 (commented):
Overall idea makes sense to me. A couple of things:

- I don't know enough of the legacy API surface to guarantee that the scenario config covered it all. Could suggest some form of programmatic validation of that as an eventual follow-up (and maybe note that in the documentation of that file).
- A few comments below. Not an expert in webpack so I only called out the bit that stood out to me in the new scenario config; at a glance the config made sense but I could have missed something else there.

build-tools/packages/build-cli/src/library/bundleSize/collectBundle.ts
TommyBrosman: I realized there's a bigger issue: if I am not running this command in a package scope where webpack is available, it will fail. bundle-size-tools has an explicit dev dependency on webpack. I need to assert that webpack is installed before running anything. I'm going to make the script fail if webpack is not installed since it is not supposed to change source code.

TommyBrosman: Added the check for both the case where the webpack config is in the package root and when it is in a subpath.

examples/utils/bundle-size-tests/scenarios/encapsulated-no-tree/webpack.config.cjs
TommyBrosman: Fair, all the others use var ouput. I'll switch to that.

alexvy86 (approved):
LGTM :)

alexvy86 (approved):

PR #27628, opened 2026-07-01
feat(tree): add alpha Component utilities for open-polymorphic schema

CraigMacomber:
## Description

Promotes the `Component` composition namespace out of the `openPolymorphism.integration.ts` test file and into `@fluidframework/tree` as a new `@alpha` API (also re-exported from `fluid-framework`). This pattern was previously validated only by the open-polymorphism examples; it's now a supported (alpha) part of the package because an app depends on it.

`Component` provides utilities for composing independently authored application "components" that contribute to a shared configuration — useful for "open polymorphism" schema patterns where the set of allowed types for a field or collection can be extended by separate libraries. Each component is a `Component.Factory` that receives a lazy reference to the composed configuration and returns the content it contributes; `Component.composeComponents` combines them into a `Component.ComposedComponents`.

Changes:
- New `packages/dds/tree/src/simple-tree/api/componentApi.ts` with the `@alpha`-tagged, fully documented `Component` namespace, exporting only what's needed (`Factory`, `LazyArray`, `Configurable`, `composeComponents`, `ComposedComponents`); the `Config` implementation class stays internal.
- Exported from `@fluidframework/tree` and (via the alpha entrypoint) `fluid-framework`; API reports updated.
- The integration test now imports `Component` instead of defining it locally.
- New focused unit tests in `componentApi.spec.ts` supplementing the existing example/integration tests.
- Fixed a latent caching bug in `getConfigured` (it previously ran `configure` on every call; now cached via `getOrCreate`).
- Changeset added.

## Reviewer Guidance

- The `Component` namespace is generic (not tied to tree types) but lives under `simple-tree/api` alongside other schema-composition utilities. Feedback welcome on placement and on the generic `Component` name for a top-level `@alpha` export.
- `Configurable` / `getConfigured` are part of the proposed API but aren't exercised by the integration examples (they're now covered by unit tests). Let me know if they should be trimmed to keep the surface minimal.


packages/dds/tree/src/componentApi.ts
```diff
@@ -0,0 +1,256 @@
...
+ * other components, as long as nothing evaluates the lazy references until composition has completed.
+ * Use {@link Component.composeComponents} to combine a set of components into a {@link Component.ComposedComponents}.
+ *
+ * @privateRemarks
+ * The examples and integration tests for this pattern live in `openPolymorphism.integration.ts`.
+ * Those tests and the docs here should be kept in sync.
+ *
+ * @alpha
+ */
+export namespace Component {
```
Josmithr, line 43: I think the docs here would greatly benefit from an `@example` or two illustrating how to use it.

packages/dds/tree/src/componentApi.ts
```diff
@@ -0,0 +1,316 @@
...
+import { UsageError } from "@fluidframework/telemetry-utils/internal";
+
+import { getOrCreate } from "./util/index.js";
+
+/**
+ * Utilities for composing application "components" which contribute to a shared configuration.
+ *
+ * @remarks
+ * Nothing in this namespace is specific to tree schema,
+ * however it is designed to be able to handle needs to components which woth with {@link TreeSchema}.
```
Josmithr:
```suggestion
 * however it is designed to be able to handle needs to components which work with {@link TreeSchema}.
```

packages/dds/tree/src/test/openPolymorphism.integration.ts
```diff
@@ -108,9 +109,10 @@ class TextItem
 }
 
 describe("Open Polymorphism design pattern examples and tests for them", () => {
+	// A simple pattern for doing open polymorphism with a mutable static registry.
 	// Currently, allowed type arrays are processed eagerly, making this pattern no longer work.
-	describe.skip("mutable static registry", () => {
-		it("without customizeSchemaTyping", () => {
+	describe("mutable static registry", () => {
+		it.skip("without customizeSchemaTyping", () => {
```
Josmithr, line 116: Can we add comments to the skipped tests explaining why they are skipped and what needs to be done to fix them?

CraigMacomber: Comment on line 113 still explains this. The change here was simply to skip the tests instead of the whole describe block so we could leave the error case test enabled.

CraigMacomber: I have now added a comment on this test directly and the other skipped one referring to the other comment

packages/dds/tree/src/test/openPolymorphism.integration.ts
```diff
@@ -411,134 +394,382 @@ describe("Open Polymorphism design pattern examples and tests for them", () => {
...
 			}
-		}
-		const colorsComponent: MyAppComponent = () => ({
-			backgrounds: () => [() => Color],
+
+			type Composed = Component.ComposedComponents<Composed, MyAppComponentContent>;
+			type MyAppComponent = Component.Factory<Composed, MyAppComponentContent>;
+
+			// A simple item type, with the required statics.
+			class MyItem extends sf.object("MyItem", {}) {
```
Josmithr, line 564: Should this implement `MyStatics`?

packages/dds/tree/src/componentApi.ts
```diff
@@ -0,0 +1,316 @@
+/*!
```
Josmithr, line 1: Obviously not for this PR, but I wonder how easy it would be to expose a more restrictive, but easier to understand model on top of this that is explicitly Tree schema focused for the common use cases. We're trying to offer a lot of flexibility below, but I imagine the vast majority of users who would want to leverage a pattern like this for SharedTree would:
a) not need a lot of the flexibility
b) find leveraging them complex / potentially confusing.

If we offered a `TreeComponent` API layer above this that was more prescriptive, it might give the average user an easier onboarding experience, and then they could move to the lower level alternatives as needed to address complex scenarios.

Just a thought.

Josmithr (approved):
This seems reasonable to expose as alpha. I think we'd really want to get a lot of user feedback before promoting further though.

Long term, this also doesn't seem like it really belongs in SharedTree. Maybe a SharedTree-specific wrapper (like I suggested [here](https://github.com/microsoft/FluidFramework/pull/27628/changes#r3547776872)) would want to live in this package long term, but this seems general enough to live somewhere else.

packages/dds/tree/src/test/openPolymorphism.integration.ts
CraigMacomber: TypeScript has no mechanism for requiring a class statically implement an interface as part of the class declaration, you simply have to use it as the interface afterwards.

packages/dds/tree/src/componentApi.ts
CraigMacomber: Since you want to let the users provide arbitrary types for the statics so they can require factories and such, putting type constraints on those types to force them to be schema (with all the type parameters and complexity related to that) makes the typing and API more complex, not less.

Limiting the system to something simpler, like a the minimal examples we have in this file ended up being so useless, that even tour brainstorm demo app couldn't use it: the features we added were to address its needs which I don't think will be uncommon in practice.

And if you don't need recursion, and just have a single allowed types collection, you can make due without any of this code at all and just write the schema normally, optionally type erasing the schema with erase schema details if you want.

I suspect there might be one level of indirection we can strip out, and can have an alternative compose API that doesn't let you construct a custom context type, and still have something useful. I'll look into it.

PR #27631, opened 2026-07-01
fix(routerlicious-driver): retry transient network errors in restWrapper

arafat-java:

## Description

Port of a fix validated in a downstream fork (concurrent-editing, commit f71f3c4), adapted to this repo's current `restWrapper.ts`/`restWrapper.spec.ts` conventions (which have diverged from the vendored copy — different imports, sinon-based test mocking instead of nock).

The routerlicious driver can run in environments (e.g. AWS Lambda) where the execution context is frozen between invocations. A keep-alive socket pooled by the underlying fetch transport can be closed by the peer (e.g. an ALB) during the freeze; the next request reuses the dead socket and rejects with "socket hang up" / ECONNRESET / EPIPE. `request()` now retries such transient network errors on a fresh socket, up to 3 attempts (1 initial + 2 retries, 250ms apart), before surfacing the error. A self-signed-certificate failure is excluded from retry since it's permanent, not transient.

## Reviewer Guidance

The retry bound (3 attempts) and delay (250ms) were chosen empirically in the downstream fork; happy to discuss if a different value/backoff strategy is preferred. This was authored in an environment without network access to build/test locally — CI will be the first real verification.


packages/drivers/routerlicious-driver/src/restWrapper.ts
```diff
@@ -169,9 +180,17 @@ class RouterliciousRestWrapper extends RestWrapper {
...
+			// this process's execution context was frozen) on a fresh socket before surfacing the error.
+			for (let attempt = 1; ; attempt++) {
+				const perfStart = performanceNow();
+				try {
+					const result = await fetch(completeRequestUrl, fetchRequestConfig);
+					return {
+						response: result,
+						duration: performanceNow() - perfStart,
+					};
+				} catch (error: any) {
```
CraigMacomber: for improved type safety, type errors as `unknown` not `any`

packages/drivers/routerlicious-driver/src/restWrapper.ts
```diff
@@ -117,6 +120,14 @@ class RouterliciousRestWrapper extends RestWrapper {
 	 */
 	private readonly retryCounter = new Map<string, number>();
 
+	/**
+	 * Maximum number of attempts (1 initial + retries) for a single fetch that fails with a transient network error.
+	 */
+	private static readonly maxNetworkErrorAttempts = 3;
+
+	/** Delay between transient network-error retries. */
+	private static readonly networkErrorRetryDelayMs = 250;
```
CraigMacomber, line 129: In networking, I generally expect https://en.wikipedia.org/wiki/Exponential_backoff unless there is some known category of issue this timeout is selected to be relevant to. Might be good to document why this specific duration makes sense.

Might also be good to document the consequences of setting this or the retry count higher (I assume it can result in something taking longer before failing, or maybe hiding flakey service issues?).

Anyway, this isn't an area I work in, so I'm no expert here, those are just my thoughts.

arafat-java: Good point, and exponential backoff is the right default for anything with more than a couple of retries. I kept it flat here mainly because the scope is deliberately narrow: this only covers one dead pooled socket left over from an execution-context freeze (e.g. Lambda), not a degrading or overloaded service — so there's no real risk of piling on load, and with only 2 retries total (3 attempts), backoff wouldn't meaningfully change the behavior over a flat delay anyway.

packages/drivers/routerlicious-driver/src/restWrapper.ts
arafat-java: fixed in b546660

PR #27633, opened 2026-07-01
test(client-tree): unimplemented "minimize" post-processor

jason-ha:
Creates a "minimize" placeholder for transaction post-process functionality.

Unit tests are added to cover the intended new functionality with unsupported tests disabled. (Since implementation is not expected to change observable result tests for observable result can be enabled with the simple do-nothing implementation that is the placeholder.)

```text
  transaction minimize post-processor
    ✔ can be supplied as a transaction post-processor without error
    ✔ throws when edits are made before and after a schema change
    ✔ preserves the observable result across an async transaction and new content appears in change
    ✔ temporary schema change throws restoring schema
    self-tests - no minimization applicable
      ✔ embeds surviving markers but not transient marker for a purely additive scenario
      ✔ result carries no build when pre-existing content is only rearranged
      ✔ result carries no build when pre-existing content is only removed
      ✔ reflects the order of only-rearranged inserted nodes and keeps every build
    preserves the observable result and new content appears in change
      ✔ keeps inserted nodes
      ✔ nets a create-then-remove to no change
      ✔ keeps only the persisted node when a transient node is also created
      ✔ reflects only the final value of a node replaced within the transaction
      ✔ reflects only the surviving node when inserted content is relocated then removed
      ✔ keeps the surrounding nodes when a node in the middle of an inserted run is removed
      ✔ keeps the surrounding nodes when an inserted node is moved then removed
      ✔ keeps only the trailing node when a moved node and its successor from leading node are removed
      ✔ keeps only the leading node when a moved node and its insertion companion are removed
      ✔ leaves pre-existing content unchanged when a transient node is inserted then removed
      ✔ keeps pre-existing content and the surviving inserted node
      ✔ reflects only the final value of a root object field set multiple times
      ✔ reflects only the final undefined root when only item's value of a field is set and then the item is removed
      ✔ reflects only the final value of a field of newly inserted root object when it is replaced
      ✔ reflects only the final value of a nested field set multiple times
      ✔ reflects an undefined root when a nested field is set and then the root object is removed
      ✔ reflects an empty root object when a nested field is set and then the nested field is removed
      ✔ reflects an empty root when a nested object with a value is added and then the root object is removed
      ✔ reflects only the final value of a field of a newly inserted nested object when nested field value is replaced
      ✔ reflects only the final value of a newly inserted nested object when nested object is replaced
      ✔ reflects only the final value of a newly inserted nested object whose value was set before it was replaced
      ✔ reflects the surviving object when a newly inserted object's nested object is removed
      ✔ reflects edits made before a schema change
      ✔ reflects edits made after a schema change
    produces a consistent view and the same observable result as not minimized
      ✔ for A inserted
      ✔ for A then B inserted
      ✔ for A added then removed
      ✔ for A kept and B transient
      ✔ for A replaced by B
      ✔ for B inserted before A then A removed
      ✔ for A,B,C inserted then B removed
      ✔ for A then B,C inserted then rearranged
      ✔ for A,B,C inserted then B moved then removed
      ✔ for A,B,C inserted then B moved then B and A removed
      ✔ for A,B,C inserted then B moved then C and B removed
      ✔ for preexisting content and transient insert
      ✔ for preexisting content and surviving insert
      ✔ for preexisting content rearranged
      ✔ for preexisting content removed
      ✔ for root Box value set twice
      ✔ for nested Box value set twice
      ✔ for root Box value set then root Box removed
      ✔ for nested Box value set then root Box removed
      ✔ for nested Box value set then nested Box removed
      ✔ for nest Box with value then root Box removed
      ✔ for add root Box then replace value
      ✔ for add nested Box then replace value
      ✔ for add nested Box then replace nested Box
      ✔ for add nested Box set value then replace nested Box
      ✔ for add Box with nested Box then remove nested Box
      ✔ for edit before schema change
      ✔ for edit after schema change
    removes extraneous data from the squashed changes (expected to fail until minimize is implemented)
      - drops the build and destroy for a create-then-remove
      - keeps only the persisted node's build when a transient node is also created
      - keeps only the final node's build when a node is replaced
      - keeps only the surviving node's build when inserted content is relocated then removed
      - keeps the surrounding builds when a node in the middle of an inserted run is removed
      - drops the build for an inserted node that is moved then removed
      - keeps only the trailing node's [modified] build when a moved node and its successor from leading node build are removed
      - keeps only the leading node's build when a moved node and its insertion companion are removed
      - carries no build for a transient insert over pre-existing content
      - keeps only the surviving inserted node's build over pre-existing content
      - keeps only the final value's build when a field is set multiple times
      - carries no build when root's value of a field is set and then the root is removed
      - keeps only the final value's builds when a field of newly inserted object is replaced
      - keeps only the final value's build when a nested field is set multiple times
      - carries no build when a nested field is set and then the root object is removed
      - keeps one build when a nested field is set and then the nested object is removed
      - carries no build when a nested object with a value is added and then the root object is removed
      - keeps only the final value's builds when a field of a newly inserted nested object is replaced
      - keeps only the final value's builds when a newly inserted nested object is replaced
      - keeps only the final value's builds when a newly inserted nested object's value is set before the nested object is replaced
      - keeps only the surviving object's build when a newly inserted object's nested object is removed
      - keeps only edits' surviving builds made before a schema change
      - keeps only edits' surviving builds made after a schema change
```

Additionally bumps `tree` test lib to ES2022 so that `String.replaceAll` is available.


packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
```diff
@@ -0,0 +1,23 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import type { ModularChangeset } from "./modularChangeTypes.js";
+
+/**
+ * "Minimizes" a {@link ModularChangeset} so that it contains no extraneous information, i.e. no data that has no net
+ * effect on the document.
```
yann-achard-MS:
```suggestion
 * "Minimizes" a {@link ModularChangeset} so that it contains no extraneous content,
 * i.e. no content that was both created and removed within the same transaction.
```
Here's my current view on key terms:
- "Information" could be anything in the changeset (content or not).
- "Content" is data that could exist in the document.
- "Data", without further qualifiers, is ambiguous to me: is it like information or is it like content?
- "Document", without qualifiers, includes all detached content. (I might say "document tree" to indicate just the attached tree.)

So, when read with that dictionary...
* "no extraneous information" is too strong of a statement. For example, changesets include IDs that link the detach and attach portions of a move. These IDs have zero impact on the document, but we're not going to strip that from the changeset.
* "no data that has no net effect on the document", if taken to mean "no content that has no net effect on the attached document tree", is also too strong of a statement. For example, we plan to leave moves that affect detached content.
* "no data that has no net effect on the document", if taken to mean "no content that has no net effect on the whole document, including detached content", is too weak of a statement: it is correct, but it doesn't really describe the impact of the minimizer.

packages/dds/tree/src/shared-tree/transactionMinimize.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @remarks
+ * This iterates over the change's constituent {@link ModularChangeset}s,
+ * replacing each with its {@link minimizeModularChangeset | minimized} form.
+ * Schema changes are left unchanged.
+ */
+function minimizeSharedTreeChange(change: SharedTreeChange): SharedTreeChange {
+	return {
+		changes: change.changes.map(
+			(inner): SharedTreeInnerChange =>
```
yann-achard-MS: You can reuse `mapDataChanges` from src\shared-tree\sharedTreeChangeFamily.ts

packages/dds/tree/src/shared-tree/transactionMinimize.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * "Minimizes" a {@link SharedTreeChange} so that it contains no extraneous
+ * information, i.e. no data that has no net effect on the document.
+ *
+ * @remarks
+ * This iterates over the change's constituent {@link ModularChangeset}s,
+ * replacing each with its {@link minimizeModularChangeset | minimized} form.
+ * Schema changes are left unchanged.
+ */
```
yann-achard-MS, line 28: Note that, in scenarios where there is more than one data change, we won't be able to minimize the overall change as much: if a data change creates content that a latter data change removes, that content won't be minimized away.

This means we have a choice to make: 
Option 1: we throw a usage error if the number of data changes is > 1.
Option 2: we keep this code as is (no error) but we update the docs to weaken the contract.

I don't have a strong opinion.

yann-achard-MS: The advantage of throwing is that the contract remains simple, and there's no risk that an app author would end up with a solution that is less secure than they thought.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,788 @@
...
+ * Counts the total number of detached-node `builds` carried by the data changes within a {@link SharedTreeChange}.
+ * @remarks A `build` is retained for every node created during the transaction. After minimization, a `build` should
+ * only remain for nodes that are still present in the document once the transaction is squashed.
+ */
+function countBuilds(change: SharedTreeChange): number {
+	let total = 0;
+	for (const inner of change.changes) {
+		if (inner.type === "data") {
+			total += inner.innerChange.builds?.size ?? 0;
+		}
```
yann-achard-MS, line 71: Note that each build may be building 1+ nodes. I'm not sure if the current tests care about that, but we probably should have some tests that do, so we'll want to sum the top-level length of the chunk on each build entry.

packages/dds/tree/src/shared-tree/transactionMinimize.ts
jason-ha: ?Option 3? (that I don't think we must do now) Track ids for all new content over the entire set and react to removed added content.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
yann-achard-MS: The fact that the tests actually look at the JSON for the changeset make this less of an issue. I'm fine leaving it as is. In fact, do we even care about the number of builds and destroys?

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,788 @@
...
+
+	// These tests only assert the observable end state of the document. Minimization must never change the
+	// observable result of a transaction, so these are expected to PASS regardless of whether minimization is
+	// actually implemented.
+	describe("preserves the observable result", () => {
+		it("keeps inserted nodes", () => {
+			const { view, stringifiedChange } = runStringArrayScenario(scenarioAThenBInserted);
+			assert.deepEqual([...view.root], ["A❤️", "B❤️"]);
+			assert.match(stringifiedChange, someSurvivingMarkerRegex);
+		});
```
yann-achard-MS, line 1088: These tests are looking at the stringified changed. This is unexpected given how they whole describe block is advertised.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,788 @@
...
+			// "A❤️", "B❤️", and "C❤️" all survive (only reordered), so both builds (A and B-C) should remain.
+			assert.equal(countBuilds(change), 2);
+		});
+	});
+
+	// These tests only assert the observable end state of the document. Minimization must never change the
+	// observable result of a transaction, so these are expected to PASS regardless of whether minimization is
+	// actually implemented.
+	describe("preserves the observable result", () => {
+		it("keeps inserted nodes", () => {
```
yann-achard-MS, line 1084: Rather than having a test for each scenario and asserting a specific end state, consider putting all the scenarios in an array (or two arrays) and running them both with and without minimization, then comparing the resulting document tree state.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,788 @@
...
+	});
+
+	// post-processor infrastructure is agnotic to the transation being async or sync, so this test is just for "good measure".
+	it("preserves the observable result across an async transaction", async () => {
+		const { view, stringifiedChange } =
+			await runStringArrayScenarioAsync(scenarioAReplacedByB);
+		assert.deepEqual([...view.root], ["B❤️"]);
+		assert.match(stringifiedChange, someSurvivingMarkerRegex);
+	});
+
```
yann-achard-MS, line 1293: I wouldn't bother with this, but I'm not against it.
If we keep such a test, it would make sense to make it a test where minification makes a difference.

Also, typo: "agnotic" -> "agnostic".

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,788 @@
...
+	// minimization algorithm is implemented. (`minimize` is currently a no-op.)
+	describe.skip("removes extraneous data from the squashed change (expected to fail until minimize is implemented)", () => {
+		it("drops the build and destroy for a create-then-remove", () => {
+			const { view, stringifiedChange } = runStringArrayScenario(scenarioAAddedThenRemoved);
+			assert.doesNotMatch(stringifiedChange, transientMarkerRegex);
+			const change = getHeadChange(view);
+			// The created node is not present in the final document, so its build/destroy should be removed.
+			assert.equal(countBuilds(change), 0);
+			assert.equal(countDestroys(change), 0);
+		});
```
yann-achard-MS, line 1438: If we stop worrying about the number of builds and destroys (see comment at the top), then we could put all the scenarios in one (or two) arrays and have a single test implementation that asserts there's no transient markers in the stringified change.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,788 @@
...
+	view: BoxArrayView;
+	stringifiedChange: JsonString<unknown>;
+} {
+	const view = createBoxArrayView(scenario.initialContent);
+	const stringifiedChange = serializeScenarioChange(view, scenario);
+	return { view, stringifiedChange };
+}
+
+// #region Scenario definitions
+// Each scenario declares the initial document content and the edits applied to the strongly-typed root node
```
yann-achard-MS, line 319: We need a systematic way to think about what scenarios are needed. Was the current set of scenarios based on such a thing?

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: small history aside: when I developed these, I did so from design that cares about builds and that was the known inspection point to me. Only later did I find path to inspect the serialized changes.
1) And neither of the two checks are full-proof. The builds are implementation subjective beyond some point. The serialized change could use an encoding and ☠️ might not show anyway - the serialization logic is far beyond the immediate implementation's working area.
2) More tests are generally better than a large one. We want to test small elements to aid in our development - easier to follow and easier to diagnose trouble. We also want (especially with agentic AI) to cover multiple small variations to avoid result hacking. One or two large complicated cases are desirable two.
So I keep 'em.
That said overnight I was thinking that we can use a single schema and there are more cases that only doable with more complicated schema.

I think once we can enable these cases, we could combine most or all of them and not need to double execute scenarios.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
yann-achard-MS, line 1: Another set of test we want to run is checking that the end state of all peers is consistent when one peer minimizes a transaction. You can test that by having a test for each scenario where the test is applied on one peer and the peers are then synchronized. You can use `validateViewConsistency` to validate that both peers get the same state.

I expect these tests will start failing once you implement minimization. That's because we're going to need to roll back the changes applied to the local branch and apply the minimized change. I forgot about that when I reviewed  https://github.com/microsoft/FluidFramework/pull/27610.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: Unfortunately inspecting JSON with hopes of not finding content (our goal) is unreliable. Negative tests are unreliable.
That is mitigated with some positive tests even if those are fussy.
Using the "sum the top-level length of the chunk on each build entry" seems like it should make things less fussy. Will attempt.

packages/dds/tree/src/shared-tree/transactionMinimize.ts
jason-ha: Option 1 implemented (and tested)

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: updated describe description

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: both seem best. The current pattern helps ensure tests are good and the comparison give breadth and depth. What is the good way to compare two?

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: fixed typo

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: The collection should cover all fundamental cases where some content is added while under transaction but ultimately is not going to be exhaustive.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: You'll have to tell me more. Follow-up PR?

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: now using comparison of exportVerbose results

packages/dds/tree/src/shared-tree/transactionMinimize.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @remarks
+ * Supply this via {@link RunTransactionParamsAlpha.postProcessor} when {@link RunTransaction | running a transaction}.
+ * "Extraneous information" includes, for example, data for nodes that were both
+ * created and removed within the transaction, or changes whose effects cancel
+ * out to nothing. Minimizing the change reduces the size of the edit that is
+ * submitted to (and stored by) the service without altering the observable
+ * effect of the transaction.
+ *
+ * The current implementation is limited and is unable to guarentee that the
```
yann-achard-MS:
```suggestion
 * The current implementation is limited and is unable to guarantee that the
```

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
yann-achard-MS: I was initially thinking that we didn't want to be looking in the changeset (because looking at the end state should be sufficient), but having these checks does have an advantage: it makes the negative checks (where we check that we don't find stuff) more trustworthy.

Resolving comment.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: I think the revised Object scenarios have good coverage now. Each scenario has a classification now.
A glaring omission from object tests are movements.
But I think these are good to get started with.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: #27673 has fixes for prior problem.
I have also added some view consistency checks into these cases.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: I've added top-level length checking as well.
No too interesting at the moment, but could be with more complicated cases.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,1568 @@
...
+}
+
+/**
+ * Counts the total number of detached-node `destroys` carried by the data changes within a {@link SharedTreeChange}.
+ */
+function countDestroys(change: SharedTreeChange): number {
+	let total = 0;
+	for (const inner of change.changes) {
+		if (inner.type === "data") {
+			total += inner.innerChange.destroys?.size ?? 0;
```
yann-achard-MS: FWIW, destroys also affect a range of nodes (which is specified by the value in the BTree) so you could count tops here as well.
This may not make a difference for now, but it might in the future.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,1568 @@
...
+}) {}
+{
+	type _check = ValidateRecursiveSchema<typeof Box>;
+}
+const OptionalBox = sf.optional(Box);
+const BoxArray = sf.array("BoxArray", Box);
+
+// A second schema factory is used to avoid collisions with the first factory's
+// schema names and altering the schema to check upgrading.
+const sf2 = new SchemaFactory("transaction-minimize");
```
yann-achard-MS, line 91: I'm surprised that using a separate factory with the same exact scope somehow helps avoid collisions.
I think the expected pattern would be to use a single factory and then `.scope` it separately for the string vs. string + box. Or just use different names for "RootArray".
Feel free to leave as-is though since the current code works.

jason-ha: Different factory is what I found in other test code.
I don't know about .scope.
The two named RootArray need to be the same name from what I understand. Schema conversion expects that RootArray is in the upgraded schema. From what I could gather, the old document will say node is of type RootArray and to edit that without replacing nodes of that type, RootArray itself needs to be updated.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: Hmm. I have outstanding changes to check the baseline builds and destroys. While the builds show as expected, the system tells me that none of the test cases have any destroys. I guess they aren't what I thought they were. Let's discuss over Teams.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,1568 @@
...
+}
+
+/**
+ * Runs a {@link TransactionScenario} in a single minimized transaction.
+ *
+ * @returns The resulting scenario view and the persisted (serialized) change as a JSON string.
+ *
+ * @remarks
+ * The persisted change is the operation SharedTree writes for document storage. It is obtained via the alpha
+ * `getChange` API surfaced on the local {@link https://fluidframework.com | "changed"} event. Unlike the in-memory
```
yann-achard-MS: That's a weird link to be giving here.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,1568 @@
...
+ * {@link SharedTreeChange} (whose inserted node contents live in tree chunks that a naive `JSON.stringify` does not
+ * traverse), the serialized change fully encodes inserted node values, so tests can assert that transient content
+ * (tagged with ☠️) was stripped by inspecting the JSON text.
+ *
+ * Additional work needs to be done to inspect any modifications to detached node content.
+ * Scenarios may need to define other edits to perform that will reattach detached nodes
```
yann-achard-MS:
```suggestion
 * Scenarios may need to define other edits that will reattach detached nodes
```

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,1568 @@
...
+			// None were inserted; should always pass.
+			assert.doesNotMatch(stringifiedChange, transientMarkerRegex);
+			assert.match(stringifiedChange, someSurvivingMarkerRegex);
+			const change = getHeadChange(view);
+			// "A❤️", "B❤️", and "C❤️" all survive (only reordered), so both builds (A and B-C) should remain.
+			assert.deepEqual(countBuilds(change), { builds: 2, tops: 3 });
+		});
+	});
+
+	// These tests only assert the observable end state of the document. Minimization must never change the
```
yann-achard-MS, line 1080: This comment is out of date. I think that sentence is unnecessary TBH.

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -0,0 +1,1568 @@
...
+			() =>
+				// This transaction is expected to throw because edits are made
+				// before and after a schema change, which is not allowed by
+				// the current minimization implementation.
+				runScenario(schemaUpgradeScenarios.edit_before_and_after_schema_change),
+			/At most one edit group can be minimized, but 2 were found/,
+		);
+	});
+
+	// post-processor infrastructure is agnostic to the transation being async or sync, so this test is just for "good measure".
```
yann-achard-MS:
```suggestion
	// post-processor infrastructure is agnostic to the transaction being async or sync, so this test is just for "good measure".
```

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: Seems appropriate still. It is odd to be able to have any passing tests for an implementation that doesn't provide value.

yann-achard-MS: I agree the explanation ("Minimization must never change...") is good. I just meant the first sentence which seems redundant with the describe block title (plus it's not hard to see the pattern in the tests themselves). Re being out of date, I meant that the test do more than assert the observable end state of the document.

yann-achard-MS (approved):

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
jason-ha: Hmm. I think that is really all this describe block currently does. The current view is checked (to have specific content/structure) and the op is partially checked. Other describe blocks do more.

PR #27637, opened 2026-07-02
fix(container-loader): prevent 0x3eb assert when read-only is forced from a connected handler

kian-thompson:
## Description

Fixes a container crash (assert `0x3eb`, "catchUpMonitor should be gone") that can occur when an application forces read-only mode synchronously from a container `"connected"` event handler.

**Root cause:** When a read connection is already caught up at the moment it reaches `Connected`, `CatchUpMonitor` fires its caught-up listener *synchronously from its constructor* — before `ConnectionStateCatchup` assigns `this.catchUpMonitor = new CatchUpMonitor(...)`. That listener raises the container `"connected"` event. If an app handler reacts by disconnecting (e.g. `forceReadonly(true)`), the `Disconnected` case runs `this.catchUpMonitor?.dispose(); this.catchUpMonitor = undefined`, which **no-ops against the not-yet-assigned field**. As the stack unwinds, the stale monitor gets attached, and the next reconnect's transition to `Connected` trips `assert(this.catchUpMonitor === undefined, 0x3eb)`.

**Fix:** Split the synchronous caught-up check out of the `CatchUpMonitor` constructor into an explicit `start()` method. `ConnectionStateCatchup` now assigns `this.catchUpMonitor` **before** calling `start()`, so any re-entrant disconnect during the connection transition observes a fully-assigned field and clears the monitor correctly. This defends against re-entrancy from any source (app handlers, runtime, etc.).

### Reproduction / verification

A regression test in `connectionStateHandler.spec.ts` reproduces the exact conditions:
1. connection already caught up at connect (`lastSequenceNumber === lastKnownSeqNumber`), and
2. a synchronous re-entrant `Disconnected` from within the `"connected"` notification.

Against the previous code it threw `0x3eb`; with the fix the reconnect proceeds cleanly to `Connected`. `catchUpMonitor.spec.ts` was updated for the constructor/`start()` split. Full suite: 287 passing.

## Reviewer Guidance

- `CatchUpMonitor` / `ICatchUpMonitor` are internal (not exported), so there is no public API surface change and no API report/type-test updates.
- Worth confirming: the `start()`-after-assignment ordering is the intended invariant, and no other caller constructs `CatchUpMonitor` expecting the constructor to fire synchronously.

### DIFF
--- a/packages/loader/container-loader/src/catchUpMonitor.ts
+++ b/packages/loader/container-loader/src/catchUpMonitor.ts
@@ -68,6 +68,7 @@ export class CatchUpMonitor implements ICatchUpMonitor {
 
 	/**
 	 * Evaluate whether the container is already caught up, synchronously notifying the listener if so.
+	 * @remarks
 	 * Must be called by the owner after storing its reference to this monitor, so that any re-entrancy
 	 * triggered by the listener observes a fully-constructed, assigned monitor.
 	 */
--- a/packages/loader/container-loader/src/catchUpMonitor.ts
+++ b/packages/loader/container-loader/src/catchUpMonitor.ts
@@ -16,7 +16,13 @@ type CaughtUpListener = () => void;
 /**
  * Monitor that emits an event when a Container has caught up to a given point in the op stream
  */
-export type ICatchUpMonitor = IDisposable;
+export interface ICatchUpMonitor extends IDisposable {
+	/**
+	 * Evaluate whether the container is already caught up, synchronously notifying the listener if so.
+	 * Must be called after the owner has stored its reference to the monitor.
+	 */
+	start(): void;
+}
 
 /**
  * Monitors a Container's DeltaManager, notifying listeners when all ops have been processed
@@ -37,6 +43,14 @@ export class CatchUpMonitor implements ICatchUpMonitor {
 
 	/**
 	 * Create the CatchUpMonitor, setting the target sequence number to wait for based on DeltaManager's current state.
+	 *
+	 * @remarks
+	 * The constructor only wires up the op listener. Call {@link CatchUpMonitor.start} once the monitor
+	 * reference has been stored to evaluate whether the container is already caught up. This ordering is
+	 * important: `start` can synchronously invoke the listener (when already caught up), which may re-enter
+	 * the owner and observe the monitor. If that synchronous check ran from the constructor, the owner's
+	 * field would not yet be assigned, so re-entrant cleanup (e.g. a disconnect) would silently no-op and
+	 * leave a stale monitor behind (see assert 0x3eb).
 	 */
 	constructor(
 		private readonly deltaManager: IDeltaManager<unknown, unknown>,
@@ -50,7 +64,14 @@ export class CatchUpMonitor implements ICatchUpMonitor {
 		);
 
 		this.deltaManager.on("op", this.opHandler);
+	}
 
+	/**
+	 * Evaluate whether the container is already caught up, synchronously notifying the listener if so.
+	 * Must be called by the owner after storing its reference to this monitor, so that any re-entrancy
+	 * triggered by the listener observes a fully-constructed, assigned monitor.
+	 */
+	public start(): void {
 		// Simulate the last processed op to set caughtUp in case we already are
 		this.opHandler({ sequenceNumber: this.deltaManager.lastSequenceNumber });
 	}
--- a/packages/loader/container-loader/src/catchUpMonitor.ts
+++ b/packages/loader/container-loader/src/catchUpMonitor.ts
@@ -19,7 +19,7 @@ type CaughtUpListener = () => void;
 export interface ICatchUpMonitor extends IDisposable {
 	/**
 	 * Evaluate whether the container is already caught up, synchronously notifying the listener if so.
-	 * Must be called after the owner has stored its reference to the monitor.
+	 * @remarks Must be called after the owner has stored its reference to the monitor.
 	 */
 	start(): void;
 }
--- a/packages/loader/container-loader/src/test/connectionStateHandler.spec.ts
+++ b/packages/loader/container-loader/src/test/connectionStateHandler.spec.ts
@@ -313,6 +313,68 @@ describe("ConnectionStateHandler Tests", () => {
 		);
 	});
 
+	// Regression test for assert 0x3eb ("catchUpMonitor should be gone").
+	//
+	// Repro conditions (both are required to exercise the original bug):
+	//   1. The connection is already caught up at the moment of the Connected transition, so
+	//      CatchUpMonitor fires its listener *synchronously* (from `start()`).
+	//   2. Something re-enters with a synchronous Disconnected from within that "connected"
+	//      notification (e.g. an app/runtime handler reacting to the "connected" event by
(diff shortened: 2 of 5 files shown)


packages/loader/container-loader/src/test/connectionStateHandler.spec.ts
```diff
@@ -313,6 +313,68 @@ describe("ConnectionStateHandler Tests", () => {
...
+	//      notification (e.g. an app/runtime handler reacting to the "connected" event by
+	//      disconnecting - such as a container forced into readonly mode).
+	//
+	// Before the fix, the CatchUpMonitor's synchronous caught-up check ran from its constructor,
+	// before `this.catchUpMonitor` was assigned, so the re-entrant Disconnected handler's
+	// `this.catchUpMonitor = undefined` no-oped and left a stale monitor attached - the next
+	// reconnect's transition to Connected then tripped assert 0x3eb. The fix assigns the field
+	// before calling `start()`, so the re-entrant disconnect disposes and clears the monitor
+	// correctly and the reconnect succeeds.
+	it("read client: re-entrant disconnect from a synchronous 'connected' notification reconnects cleanly (0x3eb regression)", () => {
```
Abe27342, line 331: Not sure if you feel this test coverage is adequate, but if you think it's worth it this e2e test seems to reproduce the issue:

```
/*!
 * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
 */

import { strict as assert } from "assert";

import { LoaderHeader, type IContainer } from "@fluidframework/container-definitions/internal";
import { ConnectionState } from "@fluidframework/container-loader";
import {
	createDetachedContainer,
	loadExistingContainer,
	type ILoadExistingContainerProps,
} from "@fluidframework/container-loader/internal";
import type { IErrorBase } from "@fluidframework/core-interfaces";
import { LocalDeltaConnectionServer } from "@fluidframework/server-local-server";
import { timeoutPromise } from "@fluidframework/test-utils/internal";

import { createLoader } from "./utils.js";

/**
 * Creates and attaches a container to establish the document, then returns the props needed to load
 * a fresh instance of it. The created container is disposed, but the underlying
 * {@link LocalDeltaConnectionServer} (held by the returned document service factory) stays alive.
 */
async function createContainerAndGetLoadProps(): Promise<ILoadExistingContainerProps> {
	const deltaConnectionServer = LocalDeltaConnectionServer.create();

	const { loaderProps, codeDetails, urlResolver } = createLoader({
		deltaConnectionServer,
	});

	const container = await createDetachedContainer({ ...loaderProps, codeDetails });
	await container.getEntryPoint();

	await container.attach(urlResolver.createCreateNewRequest("test"));
	const url = await container.getAbsoluteUrl("");
	assert(url !== undefined, "container must have url");
	container.dispose();
	return { ...loaderProps, request: { url } };
}

describe("forceReadonly during connect", () => {
	/**
	 * Regression test for the 0x3eb assert ("catchUpMonitor should be gone").
	 *
	 * Scenario: an application forces the container into read-only mode from within its "connected"
	 * event handler (a supported IContainer API + a supported time to call it). When the container
	 * reaches Connected while already caught up on the op stream, the "connected" event is raised
	 * synchronously from inside the CatchUpMonitor constructor - i.e. before
	 * ConnectionStateCatchup assigns the newly-created monitor to its field. forceReadonly(true)
	 * then synchronously disconnects and reconnects in read mode. The disconnect clears the (not yet
	 * assigned) monitor field, and once the stack unwinds the constructor's return value resurrects a
	 * stale monitor. When the read-mode reconnect subsequently reaches Connected,
	 * `assert(this.catchUpMonitor === undefined, 0x3eb)` fires and the container closes.
	 *
	 * The container should instead settle into a connected, read-only state without closing.
	 */
	it("does not assert (0x3eb) when the app forces read-only from the connected event", async () => {
		const loadProps = await createContainerAndGetLoadProps();

		// Load without connecting so we can install the "connected" listener before the container
		// establishes its delta connection.
		const container: IContainer = await loadExistingContainer({
			...loadProps,
			request: {
				url: loadProps.request.url,
				headers: {
					...loadProps.request.headers,
					[LoaderHeader.loadMode]: { deltaConnection: "none" },
				},
			},
		});

		let closeError: IErrorBase | undefined;
		container.once("closed", (error?: IErrorBase) => {
			closeError = error;
		});

		// Simulate the application forcing read-only the moment the container connects.
		// Registered before the settle listeners below so it runs first for the initial "connected" event.
		container.once("connected", () => {
			container.forceReadonly?.(true);
		});

		// Wait until the container settles into a connected read-only state, or closes (the bug).
		const outcomeP = timeoutPromise<"connected" | "closed">(
			(resolve) => {
				container.once("closed", () => resolve("closed"));
				container.on("connected", () => {
					if (
						container.connectionState === ConnectionState.Connected &&
						container.readOnlyInfo.readonly === true
					) {
						resolve("connected");
					}
				});
			},
			{
				durationMs: 5000,
				errorMsg: "container neither reached connected read-only state nor closed",
			},
		);

		container.connect();

		const outcome = await outcomeP;

		assert.strictEqual(
			outcome,
			"connected",
			`container closed instead of connecting read-only: ${closeError?.message}`,
		);
		assert.strictEqual(
			container.connectionState,
			ConnectionState.Connected,
			"container should be connected",
		);
		assert.strictEqual(
			container.readOnlyInfo.readonly,
			true,
			"container should be read-only",
		);

		container.dispose();
	});
});
```

Abe27342: eh, seems potentially unnecessary after looking at the diff. The test you have pretty accurately captures behavior (doesn't rely on a bunch of mocks)

Abe27342 (approved):

packages/loader/container-loader/src/catchUpMonitor.ts
```diff
@@ -16,7 +16,13 @@ type CaughtUpListener = () => void;
 /**
  * Monitor that emits an event when a Container has caught up to a given point in the op stream
  */
-export type ICatchUpMonitor = IDisposable;
+export interface ICatchUpMonitor extends IDisposable {
+	/**
+	 * Evaluate whether the container is already caught up, synchronously notifying the listener if so.
+	 * Must be called after the owner has stored its reference to the monitor.
```
Josmithr: Nit
```suggestion
	 * @remarks Must be called after the owner has stored its reference to the monitor.
```

packages/loader/container-loader/src/catchUpMonitor.ts
```diff
@@ -50,7 +64,14 @@ export class CatchUpMonitor implements ICatchUpMonitor {
 		);
 
 		this.deltaManager.on("op", this.opHandler);
+	}
 
+	/**
+	 * Evaluate whether the container is already caught up, synchronously notifying the listener if so.
+	 * Must be called by the owner after storing its reference to this monitor, so that any re-entrancy
```
Josmithr, line 72: Nit:
```suggestion
	 * @remarks
	 * Must be called by the owner after storing its reference to this monitor, so that any re-entrancy
```

Josmithr (approved):
Approving for docs

Abe27342 (approved):

Josmithr (approved):

PR #27638, opened 2026-07-06
feat(examples): add/remove views in text example

brrichards:
## Description

Adds the ability to add/remove views(users) in the text-editor application. When trace testing multiple users in a single browser seem to run on the same thread causing local ops to get tangled with remote ops in the profiling when they come in. This PR aims to allow only a single view while still being able to have multiple for other testing/demo purposes. 

Changes in this PR:
- Add/remove views
- Devtools is attached to each view
- Small cleanup on the devtools button to remove custom css since the basic css is good enough, and will align with the base css of the add/remove buttons 

## UI Changes

Devtools and add view button:
image

User Views with remove button at top right corner of a view:
image

## Reviewer Guidance


examples/data-objects/text-editor/src/app.tsx
```diff
@@ -119,15 +128,42 @@ function getConnectionConfig(userId: string): AzureLocalConnectionConfig {
...
+// Panel colors are one base hex offset by the panel's left-to-right position, so
+// they're deterministic and identical across browsers (no stored palette).
+const colorForIndex = (index: number): string =>
+	`#${((0x4a90d9 + index * 0x2a1f3c) % 0x1000000).toString(16).padStart(6, "0")}`;
+
+const initialUserCount = 2;
+
+/** One connected user's container + view. `id` is just a stable React key. */
+export interface UserView {
+	id: number;
```
CraigMacomber: Given id is "stable" I suspect it, and likely the others, should be readonly.

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -119,15 +128,42 @@ function getConnectionConfig(userId: string): AzureLocalConnectionConfig {
...
+	container: IFluidContainer<typeof containerSchema>;
+	treeView: TreeViewAlpha<typeof TextEditorRoot>;
+}
+
+const makeUserId = (id: number): string =>
```
CraigMacomber: I think its needlessly confusing to have "userId" as used here and "UserView.id" above be unrelated, and also undocumented in what they do.

Could you make this into a function instead of a constant (for consistency with how our code normally works), then document both what this function does, and what its parameter does? It could also use a line comment inside the function explaining why it does what it does, since I think just using a random UUID would also work, I presume there is some benefit to taking in an id, and doing what ever that is with the random.

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -119,15 +128,42 @@ function getConnectionConfig(userId: string): AzureLocalConnectionConfig {
...
-	 */
-	devtoolsProps?: DevtoolsProps;
+// Panel colors are one base hex offset by the panel's left-to-right position, so
+// they're deterministic and identical across browsers (no stored palette).
+const colorForIndex = (index: number): string =>
+	`#${((0x4a90d9 + index * 0x2a1f3c) % 0x1000000).toString(16).padStart(6, "0")}`;
+
+const initialUserCount = 2;
+
+/** One connected user's container + view. `id` is just a stable React key. */
```
CraigMacomber: Put docs for properties in the doc comment on the property, not as part of the doc on the interface.

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -119,15 +128,42 @@ function getConnectionConfig(userId: string): AzureLocalConnectionConfig {
...
-	user2: TreeViewAlpha<typeof TextEditorRoot>;
-	containerId: string;
-	/**
-	 * Properties for (re)initializing Devtools. Held so the UI can toggle Devtools on and off at runtime
-	 * (see {@link DevtoolsToggle}).
-	 */
-	devtoolsProps?: DevtoolsProps;
+// Panel colors are one base hex offset by the panel's left-to-right position, so
+// they're deterministic and identical across browsers (no stored palette).
+const colorForIndex = (index: number): string =>
```
CraigMacomber: this could use a doc comment. Alos should prababyl be a function.

Maybe this generates a collection of suitable light background colors, unique based on the index? Doc comment should say something like that if thats what it does.

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -173,26 +203,35 @@ async function loadExistingContainer(
...
-	const devtoolsLogger = createDevtoolsLogger();
-
-	const client1 = new AzureClient({
-		connection: getConnectionConfig(user1Id),
+/**
+ * Connects one user (its own Fluid client) to an existing document.
+ * Shared by the initial load and the "Add user" button.
+ */
+async function connectUser(
+	id: number,
```
CraigMacomber: there are a lot of id numbers going around. If they are all the same, maybe make a type alias for them and put the docs there?

examples/data-objects/text-editor/src/app.tsx
CraigMacomber: Might be simpler to just have one id kind, and use this string for the react key, and remove the parameter here.

examples/data-objects/text-editor/src/test/app.test.tsx
```diff
@@ -5,44 +5,70 @@
...
-		}),
-	);
+	treeView.initialize(createInitialRoot(initialValue));
 	return treeView;
 }
 
+/**
+ * Creates a {@link UserView}
+ */
+function createUserView(id: number, initialValue: string): UserView {
```
CraigMacomber: if could use a more specific type, and/or a doc.

Also, doesn't this create a container with a view, not create a view? (I'm assuming that based on taking in an initial value anyway), so it likely should be renamed. Maybe createTestView (and the doc comment notes it creates a dummy tree to hook that user view up to?).

CraigMacomber: The current doc comment is only what the type system already documents: same for the function name. It would be good to provide more specifics rather than just state the same thing 3 ways. For example, why might someone use this instead of the other ways to create a view? Does this do anything other than create a UserView? THose are the kinds of things that would be good to imply in the name or doc comment.

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -173,26 +203,35 @@ async function loadExistingContainer(
...
+	containerId: string,
```
CraigMacomber, line 235: When we have multiple ids in something, a doc comment on each is pretty important to help people know what they all do. I think the name on this one is rather clear, but the above one is not: still comments would be good, or more specific types.

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -204,63 +243,25 @@ async function initFluid(): Promise<DualUserViews> {
...
-				container: user1Container,
-				containerKey: "User 1 Container",
-			},
-			{
-				container: user2Container,
-				containerKey: "User 2 Container",
-			},
-		],
-	};
+	for (let id = initialUsers.length + 1; id <= initialUserCount; id++) {
```
CraigMacomber: Might be worth a comment,

```suggestion
	// Number each initial user, starting at 1
	for (let id = initialUsers.length + 1; id <= initialUserCount; id++) {
```

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -119,15 +128,42 @@ function getConnectionConfig(userId: string): AzureLocalConnectionConfig {
...
+// they're deterministic and identical across browsers (no stored palette).
+const colorForIndex = (index: number): string =>
+	`#${((0x4a90d9 + index * 0x2a1f3c) % 0x1000000).toString(16).padStart(6, "0")}`;
+
+const initialUserCount = 2;
+
+/** One connected user's container + view. `id` is just a stable React key. */
+export interface UserView {
+	id: number;
+	container: IFluidContainer<typeof containerSchema>;
```
CraigMacomber: I think we might just need this for dev tools. Might be nice to document that, OR remove the treeView as its kinda off to pass in the container (which includes the tree) and also the tree view derived from the tree.

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -119,15 +120,72 @@ function getConnectionConfig(userId: string): AzureLocalConnectionConfig {
...
+}
+
+const initialUserCount = 2;
+
+/**
+ * Identifies one user in this app.
+ *
+ * Serves as the React key for the user's panel, the key for its Devtools registration,
+ * and the user id reported to the Fluid service's audience. Ids are randomly generated
+ * and never reused within a page, so a removed user's id is not given to a later-added
```
Josmithr: Nit
```suggestion
 * and the user ID reported to the Fluid service's audience. IDs are randomly generated
 * and never reused within a page, so a removed user's ID is not given to a later-added
```

examples/data-objects/text-editor/src/app.tsx
```diff
@@ -448,60 +502,88 @@ const UserPanel: FC<{
...
+	 * How "Add user" connects a new user to the document. Defaults to
+	 * {@link connectUser}; tests inject a fake to avoid a real service connection.
+	 */
+	connectUser?: typeof connectUser;
+}> = ({ containerId, devtoolsLogger, initialUsers, connectUser: connect = connectUser }) => {
+	const [users, setUsers] = useState<UserView[]>(initialUsers);
+
+	// Devtools defaults to off and is toggled at runtime (see DevtoolsToggle).
+	const [devtoolsEnabled, setDevtoolsEnabled] = useState(false);
+	// (Re)initializes Devtools with the current users' containers whenever it's enabled
```
Josmithr, line 542: Nit
```suggestion

	// (Re)initializes Devtools with the current users' containers whenever it's enabled
```

Josmithr (approved):
Left a few nitpicks, and Craig left a couple comments as well. Otherwise, looks good to me. I would recommend adding a screenshot or two to the PR description for visual context.

examples/data-objects/text-editor/src/app.tsx
brrichards: updated

examples/data-objects/text-editor/src/app.tsx
brrichards: updated to be a function as well as documenting the use of index

examples/data-objects/text-editor/src/app.tsx
brrichards: updated to reflect this

examples/data-objects/text-editor/src/test/app.test.tsx
brrichards: updated the docs to be more specific about the reasoning behind using this instead of other ways to create views

examples/data-objects/text-editor/src/app.tsx
brrichards: There is only the containerId now and the userId which is now type aliased

examples/data-objects/text-editor/src/app.tsx
brrichards: updated

examples/data-objects/text-editor/src/app.tsx
brrichards: updated the doc comments above to describe what the containerID is and links back to userID

examples/data-objects/text-editor/src/app.tsx
brrichards: good call. updated!

examples/data-objects/text-editor/src/app.tsx
brrichards: After changing the id structure, the id's are no longer numbered like this, but left a new comment on the for loop when initializing the 2 starting users

Josmithr (approved):

PR #27647, opened 2026-07-06
test(client-tree): enable one 8915 case

jason-ha:
At some point upgraded views started reflecting new schema, allowing edits to use it.

Under a transaction, original views must have been disposed. Otherwise "Cannot create a second tree view from the same checkout" is thrown. Also `initialize` needs called before `runTransation` to ensure tree is within schema compliance.

### DIFF
--- a/packages/dds/tree/src/test/shared-tree/sharedTree.spec.ts
+++ b/packages/dds/tree/src/test/shared-tree/sharedTree.spec.ts
@@ -2319,8 +2319,7 @@ describe("SharedTree", () => {
 				new TreeViewConfiguration({ schema: JsonAsTree.Array, enableSchemaValidation }),
 			);
 			view1Json.upgradeSchema();
-			// TODO:#8915: This should be able to insert the _number_ 44, not the string, but currently cannot - see bug #8915
-			view1Json.root.insertAtEnd("44");
+			view1Json.root.insertAtEnd(44);
 
 			tree1.containerRuntime.connected = true;
 
@@ -2467,10 +2466,14 @@ describe("SharedTree", () => {
 			const view = tree.viewWith(
 				new TreeViewConfiguration({ schema, enableSchemaValidation }),
 			);
+			view.initialize({ foo: [] });
 			Tree.runTransaction(view, () => {
-				view.initialize({ foo: [] });
 				view.root.foo.insertAtStart("a");
 				view.root.foo.insertAtStart("b");
+				// This dispose does not seem to be enough to prevent
+				//   Error: Cannot create a second tree view from the same checkout
+				// when creating view2 below.
+				view.dispose();
 
 				// Update schema which now allows all primitives under field "foo".
 				const view2 = tree.viewWith(


packages/dds/tree/src/test/shared-tree/sharedTree.spec.ts
```diff
@@ -2467,10 +2466,14 @@ describe("SharedTree", () => {
...
 			);
+			view.initialize({ foo: [] });
 			Tree.runTransaction(view, () => {
-				view.initialize({ foo: [] });
 				view.root.foo.insertAtStart("a");
 				view.root.foo.insertAtStart("b");
+				// This dispose does not seem to be enough to prevent
+				//   Error: Cannot create a second tree view from the same checkout
+				// when creating view2 below.
+				view.dispose();
```
jenn-le, line 2476: I think view2 also needs to be disposed since there's a call to viewWith outside the transaction

packages/dds/tree/src/test/shared-tree/sharedTree.spec.ts
CraigMacomber, line 2328: now that this is inserting the right thing, we should adjust this as well to validate it:


```suggestion
			assert.deepEqual([...view1Json.root], ["42", "43", 44]);
```

CraigMacomber: Looks like git hub isn't lining up the diff correctly (this should replace the assert on that line, not replace nothing), but I suspect you get the idea

packages/dds/tree/src/test/shared-tree/sharedTree.spec.ts
```diff
@@ -2467,10 +2466,14 @@ describe("SharedTree", () => {
 			const view = tree.viewWith(
 				new TreeViewConfiguration({ schema, enableSchemaValidation }),
 			);
+			view.initialize({ foo: [] });
 			Tree.runTransaction(view, () => {
```
CraigMacomber, line 2470: Since transactions apply to a given view, what is being done here is rather odd.

I think it actually should work (maybe) since internally the transaction is really on the checkout, which I think lives through view dispose and is reused by the second view, but maybe it might make more sense to remove the transaction wrapper here unless its important for what is being tested.

packages/dds/tree/src/test/shared-tree/sharedTree.spec.ts
CraigMacomber: I though this comment was saying view2 can't be created. Disposing view2 can't be a requirement for creating view2.

packages/dds/tree/src/test/shared-tree/sharedTree.spec.ts
CraigMacomber: Actually, I forgot that transactions do odd things with branches. I wouldn't not be surprised if disposing a view inside of a traction on it doesn't work correctly, and disposed the hidden branch for the transaction leaving stuff behind which causes the issue you see below.

If removing the transaction fixes this, then we have an additional bug, which we might want to handle simply by usage error if disposing a view inside of a transaction on it.

jason-ha: I don't know what this test is really going for but I think the transaction is relevant. I found it when targeting schema change under a transaction.
The related task was Closed when I found this code and I thought reenables were accidentally left behind. What actually happened was the task was closed as stale. I've since reactivated.
As I was attempting to use this pattern, I found there were changes along the way that weren't accounted for. So, these changes fixup what I know of, but test still fails -- didn't investigate deeply.
I was able to get my related tests running with schema changes under transaction. See `schemaUpgradeScenarios` in https://github.com/microsoft/FluidFramework/blob/fefd4b6dcef0ce1d4cdc726fe188442a51e436b3/packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts which is part of pending PR #27633.

packages/dds/tree/src/test/shared-tree/sharedTree.spec.ts
jenn-le: I think this comment is wrong, running the test shows me the error is happening at line 2488

PR #27648, opened 2026-07-07
feat: Expose hasStagedChanges/hasStagedChangesChanged on IContainerRuntimeBase

dannimad:
## Description

`IContainerRuntimeBase` now exposes `hasStagedChanges`, a boolean indicating whether there are any changes submitted while in Staging Mode (via `enterStagingMode`) that have not yet been discarded or committed. A new `hasStagedChangesChanged` event is emitted whenever this value changes.

This is distinct from `isDirty`/`dirty`: a container runtime can be dirty due to ordinary unacknowledged local changes without having any staged changes, and vice versa. Per review feedback, StagingMode-related concepts stay on `IContainerRuntimeBase` for now rather than being exposed on `IContainer`.

```typescript
containerRuntime.on("hasStagedChangesChanged", (hasStagedChanges) => {
        // update UI to reflect whether there are staged changes awaiting commit/discard
});

const hasStagedChanges = containerRuntime.hasStagedChanges;
```

Test coverage mirrors the existing `isDirty`/`dirty` test suites:
- `PendingStateManager` unit tests for `hasStagedChanges()`
- `ContainerRuntime` baseline flag tests across Attached/Attaching/Detached states
- `ContainerRuntime` staging-mode lifecycle tests (`commitChanges`/`discardChanges`)
- `ContainerRuntime` `hasStagedChangesChanged` event tests
- End-to-end lifecycle tests in `local-server-tests` (`documentStagedChanges.spec.ts`)

While adding tests, found and fixed a real bug: `hasStagedChanges`/`hasStagedChangesChanged` were not refreshed after `flush()` alone, so the cached value could remain stale (`false`) even with staged ops pending in the `PendingStateManager`. Fixed by calling `updateHasStagedChangesState()` at the end of `ContainerRuntime.flush()`.

## Reviewer Guidance

- The `flush()` fix is the only behavioral change outside of tests — please double check the reasoning around staleness in `updateHasStagedChangesState()`.


packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4871,13 +4902,31 @@ export class ContainerRuntime
...
 
-		if (this.lastEmittedDirty === dirty) {
+		if (this.lastEmittedDirty !== dirty) {
+			this.lastEmittedDirty = dirty;
+			if (this.emitDirtyDocumentEvent) {
+				this.emit(dirty ? "dirty" : "saved");
+			}
+		}
+
+		this.updateHasStagedChangesState();
```
markfields: This feels excessive — `updateDocumentDirtyState` is called from ~6 places (process, resubmit, connect, rollback, etc.) and most of those can't change staged-changes state. The only transition this catches that the other two call sites (`flush()` line 3728, `commitChanges` line 3922) don't is `discardChanges` (line 3910 calls `updateDocumentDirtyState` but not `updateHasStagedChangesState`).

Consider dropping this from `updateDocumentDirtyState` and adding an explicit `this.updateHasStagedChangesState()` call in the `discardChanges` lambda (after the `updateDocumentDirtyState()` at line 3910). More surgical, avoids running the check on every unrelated dirty-state update.

packages/runtime/container-runtime/src/pendingStateManager.ts
```diff
@@ -333,6 +333,24 @@ export class PendingStateManager implements IDisposable {
...
 
+	/**
+	 * Checks the pending messages to see if any of them are staged (submitted while in Staging Mode) and
+	 * represent user changes (aka "dirtyable" messages).
+	 */
+	public hasStagedChanges(): boolean {
+		for (let i = 0; i < this.pendingMessages.length; i++) {
+			const element = this.pendingMessages.get(i);
+			if (
+				element?.batchInfo.staged === true &&
```
markfields: Why check `hasTypicalRuntimeOp` and `isContainerMessageDirtyable` here? `batchInfo.staged === true` should be sufficient — if something was submitted during staging mode, it's a staged change, full stop. The dirtyable/typicalRuntimeOp filters are about whether something counts as a "user change" for dirty tracking, but that's a different concept.

Even if the only staged change were "non-dirtyable," we'd still want to report that staged changes exist (the user entered staging mode and submitted something — they need to commit or discard it).

If you want to keep the extra conditions for some reason, then at minimum rename this to `hasStagedUserChanges` so the semantics are clear and don't mislead consumers into thinking `false` means "nothing staged."

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -4707,6 +4772,77 @@ describe("Runtime", () => {
...
+					"Should not have staged changes immediately after entering staging mode (no ops yet)",
+				);
+
+				submitDataStoreOp(
+					containerRuntime,
+					"1",
+					genTestDataStoreMessage("staged-op"),
+					"LOCAL_OP_METADATA",
+				);
+				containerRuntime.flush();
```
markfields, line 4804: I believe hasStagedChanges should be true as soon as the changes are submitted, right? So move this assert above flush.

And then you'll have to change the code to make the test pass.  See my other comment in the local-server-tests file for more about why I think this

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -5477,6 +5613,173 @@ describe("Runtime", () => {
...
+						1,
+						"Should not emit again while hasStagedChanges stays true",
+					);
+				});
+
+				it("fires during orderSequentially's internal staging mode usage (EnableRollback=true), unlike stagingModeChanged", async () => {
+					// orderSequentially with EnableRollback calls enterStagingModeCore(silent=true) internally,
+					// which suppresses stagingModeChanged. hasStagedChangesChanged is driven by
+					// updateDocumentDirtyState/updateHasStagedChangesState instead, which are not gated by
+					// the `silent` flag, so it CAN fire in this scenario.
```
markfields, line 5801: Why is this the preferred design?  Or are we ambivalent and just documenting what it does (which is the simplest approach)

packages/runtime/runtime-definitions/src/dataStoreContext.ts
```diff
@@ -351,6 +361,15 @@ export interface IContainerRuntimeBase extends IEventProvider<IContainerRuntimeB
 	 * @see {@link IContainerRuntimeBase.enterStagingMode}
 	 */
 	readonly inStagingMode: boolean;
+
+	/**
+	 * Returns true if there are any staged changes, i.e. changes submitted while in Staging Mode that have
+	 * not yet been discarded or committed.
+	 *
+	 * @remarks This is distinct from {@link @fluidframework/container-runtime-definitions#IContainerRuntime.isDirty}: a container may be dirty due to
+	 * ordinary unacknowledged local changes without having any staged changes, and vice versa.
```
markfields: > and vice versa

This is highly unlikely, usually any staged changes will render the container dirty.  Probably not worth commenting on at all, just remove those 3 words?

packages/test/local-server-tests/src/test/documentStagedChanges.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+			[[mapId, SharedMap.getFactory()]],
+			"default",
+		);
+
+		const runtimeFactory = new ContainerRuntimeFactoryWithDefaultDataStore({
+			defaultFactory,
+			registryEntries: [[defaultFactory.type, Promise.resolve(defaultFactory)]],
+			// Use a low threshold so staged ops are auto-flushed to the PendingStateManager (which is what
+			// hasStagedChanges checks) right away instead of waiting for the default (1000 ops) batch size.
+			runtimeOptions: { stagingModeAutoFlushThreshold: 1 },
```
markfields: Ok this raises an interesting question.  I had forgotten about this autoflush feature.  I wonder if hasStagedChanges should also return true if the outbox is nonempty and we're in staging mode.  I think the difference between outbox and PSM is irrelevant to a consumer using Staging Mode.

packages/test/local-server-tests/src/test/documentStagedChanges.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+
+		sharedMap.set("key2", "value2");
+		await Promise.resolve(); // Let the scheduled flush (turn-based) run
+		checkStagedChangesState("after second value set", true, 1, 0);
+
+		stageControls.discardChanges();
+		checkStagedChangesState("after discardChanges", false, 1, 1);
+	});
+
+	it("is independent from the ordinary isDirty/dirty flag and event", async () => {
```
markfields, line 274: This test covers `isDirty` independence via the *discard* path, but the *commit* path is arguably more interesting: immediately after `commitChanges()`, the previously-staged ops become in-flight (unacked), so `isDirty` should remain `true` while `hasStagedChanges` flips to `false`. Consider adding a variant that asserts that — it's a subtle but important distinction for consumers.

packages/test/local-server-tests/src/test/documentStagedChanges.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+			"Should still be dirty due to the original ordinary edit",
+		);
+		checkStagedChangesState("after discardChanges", false, 1, 1);
+
+		await loaderContainerTracker.ensureSynchronized();
+
+		assert.equal(containerRuntime.isDirty, false, "Should be clean after processing");
+		checkStagedChangesState("after processing", false, 1, 1);
+	});
+});
```
markfields, line 367: Consider adding a test that verifies `hasStagedChanges` is `false` on a freshly-loaded container (loaded from snapshot). Staged state is never persisted today, but one day we may support hydrating a container with staged changes — having this as an explicit baseline will surface it when that changes.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -5477,6 +5613,173 @@ describe("Runtime", () => {
 					);
 				});
 			});
+
+			describe("hasStagedChangesChanged event", () => {
```
markfields, line 5664: Consider adding a unit test here that enters staging mode, submits no ops, then exits (both commit and discard paths) — asserting `hasStagedChangesChanged` is never emitted (event count stays 0) and `hasStagedChanges` remains `false` throughout. The e2e test checks the property at line 171 but doesn't assert the event isn't fired, and this is the right place for that kind of negative assertion as a unit test.

markfields (commented):
Looks pretty good, but I think there's a mismatch with what expected behavior probably is for a consumer - they'll expect `hasStagedChanges` to be true as soon as the change happens, regardless of when the batch flushes from the Outbox.

packages/runtime/container-runtime/src/containerRuntime.ts
dannimad: Agreed — confirmed 6 call sites, only discardChanges/commitChanges/flush/submitMessage can actually affect staged state. Removed the blanket call from updateDocumentDirtyState and added explicit this.updateHasStagedChangesState() calls only where staged state can truly change: discardChanges, and submitMessage (new — see next item).

packages/test/local-server-tests/src/test/documentStagedChanges.spec.ts
dannimad: This was a real bug: while in Staging Mode, a submitted op sits in the Outbox until the next flush, and during that window hasStagedChanges incorrectly reported false. Fixed computeCurrentHasStagedChanges() to also check this.inStagingMode && this.outbox.containsUserChanges(). Updated submitMessage to call updateHasStagedChangesState() explicitly, and updated/added tests in both containerRuntime.spec.ts and documentStagedChanges.spec.ts to assert the value is true immediately on submit, before flush.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
dannimad: Missing test coverage (3 separate suggestions)
Added all three:

Unit tests for entering/discarding and entering/committing staging mode with no ops — assert hasStagedChangesChanged never fires and hasStagedChanges stays false.
documentStagedChanges.spec.ts: new test asserting the commit path — isDirty stays true (in-flight/unacked) right after commitChanges() while hasStagedChanges flips to false.
documentStagedChanges.spec.ts: new test loading a second, freshly-loaded container from the same document while the first has uncommitted staged changes — asserts the fresh container reports hasStagedChanges: false (added a loadContainer helper using loadExistingContainer for this).

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
dannimad: Just documenting. I think this could be a scenario where we could ask clarification and fix in a follow up

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4036,6 +4049,38 @@ export class ContainerRuntime
 		);
 	}
 
+	/**
+	 * Returns true if there are any staged changes, i.e. changes submitted while in Staging Mode
+	 * (see {@link @fluidframework/runtime-definitions#IContainerRuntimeBaseInternal.enterStagingMode})
+	 * that have not yet been discarded or committed.
+	 *
+	 * @remarks This is distinct from {@link ContainerRuntime.isDirty}: a container may be dirty due to
```
markfields, line 4057: The "and vice versa" should be kept here after all! Or explain the case specifically - The new unit test (`hasStagedChanges is true for a non-dirtyable op sitting in the Outbox while staging`) explicitly demonstrates that `hasStagedChanges` can be `true` while `isDirty` is `false` — that's the "vice versa" case.

packages/runtime/runtime-definitions/src/dataStoreContext.ts
```diff
@@ -351,6 +361,15 @@ export interface IContainerRuntimeBase extends IEventProvider<IContainerRuntimeB
 	 * @see {@link IContainerRuntimeBase.enterStagingMode}
 	 */
 	readonly inStagingMode: boolean;
+
+	/**
+	 * Returns true if there are any staged changes, i.e. changes submitted while in Staging Mode that have
+	 * not yet been discarded or committed.
+	 *
+	 * @remarks This is distinct from {@link @fluidframework/container-runtime-definitions#IContainerRuntime.isDirty}: a container may be dirty due to
+	 * ordinary unacknowledged local changes without having any staged changes.
```
markfields, line 370: Same here — "and vice versa" should be kept. A non-dirtyable op submitted during staging mode results in `hasStagedChanges=true` but `isDirty=false` (as the new unit test demonstrates).

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -5477,6 +5660,232 @@ describe("Runtime", () => {
...
+				it("emits with true after submitting a staged op (before flush), and hasStagedChanges is true when handler runs", () => {
+					stubChannelCollection(containerRuntime);
+					const events: boolean[] = [];
+					const hasStagedChangesAtEventTime: boolean[] = [];
+
+					containerRuntime.on("hasStagedChangesChanged", (hasStagedChanges) => {
+						events.push(hasStagedChanges);
+						hasStagedChangesAtEventTime.push(containerRuntime.hasStagedChanges);
```
markfields, line 5672: nit: just assert here that `containerRuntime.hasStagedChanges === hasStagedChanges`? And remove `hasStagedChangesAtEventTime`

packages/runtime/container-runtime/src/test/pendingStateManager.spec.ts
```diff
@@ -1281,6 +1281,146 @@ describe("Pending State Manager", () => {
...
+			];
+			const psm = createPendingStateManager(pendingMessages as IPendingMessage[]);
+			assert.strictEqual(
+				psm.hasStagedChanges(),
+				true,
+				"Should be true with staged non-dirtyable op: anything staged counts, regardless of dirtyability",
+			);
+		});
+
+		it("returns true if a staged message is an empty batch (runtimeOp undefined)", () => {
```
markfields, line 1374: Side note - I don't think this is possible anyway.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -4871,13 +4916,31 @@ export class ContainerRuntime
 	private updateDocumentDirtyState(): void {
 		const dirty: boolean = this.computeCurrentDirtyState();
 
-		if (this.lastEmittedDirty === dirty) {
+		if (this.lastEmittedDirty !== dirty) {
```
markfields, line 4919: Why is this being inverted?  Especially considering that `updateHasStagedChangesState` follows the other form!

It's a nit, but I'd revert this to reduce the diff and keep them symmetrical.

packages/test/local-server-tests/src/test/documentStagedChanges.spec.ts
```diff
@@ -0,0 +1,367 @@
...
+
+		sharedMap.set("key", "value");
+
+		await Promise.resolve(); // Let the scheduled flush (turn-based) run
+
+		checkStagedChangesState("after value set in staging mode", true, 1, 0);
+
+		stageControls.commitChanges();
+
+		checkStagedChangesState("after commitChanges", false, 1, 1);
```
markfields, line 210: Also assert isDirty is true here then false after ensureSynchronized?

[edit] - Oh, there is that test below.  Are they redundant otherwise?

markfields (approved):

CraigMacomber (approved):
The changeset seems fine from a docs perspective.

.changeset/five-icons-build.md
```diff
@@ -0,0 +1,18 @@
+---
+"@fluidframework/runtime-definitions": minor
+"@fluidframework/container-runtime-definitions": minor
+"__section": feature
+---
+Expose whether a container runtime has staged changes
+
+`IContainerRuntimeBase` now exposes `hasStagedChanges`, a flag indicating whether there are any changes submitted while in Staging Mode (via `enterStagingMode`) that have not yet been discarded or committed. A new `hasStagedChangesChanged` event is emitted whenever this value changes.
```
Josmithr, line 8: Nit: it would probably be worth noting that this is a beta API. That way consumers of our public APIs can know to ignore these change notes.

I also like to recommend linking to our API docs here when easy. E.g.
```suggestion
[`IContainerRuntimeBase`](https://fluidframework.com/docs/api/runtime-definitions/icontainerruntimebase-interface) now exposes `hasStagedChanges`, a flag indicating whether there are any changes submitted while in Staging Mode (via `enterStagingMode`) that have not yet been discarded or committed. A new `hasStagedChangesChanged` event is emitted whenever this value changes.
```

Josmithr (approved):

PR #27654, opened 2026-07-08
Update pinned dependency versions across workspaces

frankmueller-msft:
## Description

Updates a set of pinned transitive dependency versions across all pnpm workspaces in the repository, using the established `overrides:` mechanism in each `pnpm-workspace.yaml`, and regenerates the affected lockfiles.

### Packages updated

| Package | Notes |
| --- | --- |
| `axios` | Pre-1.0 line pinned to `^0.32.0`; 1.x line pinned to `1.16.0` |
| `ws` | 8.x line pinned to `^8.21.0`; 7.x line pinned to `^7.5.11` where present |
| `fast-uri` | `^3.1.2` |
| `tmp` | `^0.2.6` (no release exists in the 0.0.x/0.1.x lines) |
| `form-data` | 4.x line pinned to `^4.0.6` |
| `langsmith` | `^0.6.0` |
| `systeminformation` | `^5.31.6` |
| `simple-git` | `^3.36.0` |
| `@nevware21/ts-utils` | `^0.14.0` |
| `@github/copilot` | `^1.0.43` |

### Scope

The overrides are applied per-workspace (each release group maintains its own `overrides:` block and lockfile). Only the workspaces that actually resolved an affected version were touched. Intentionally version-pinned cross-compat test workspaces and test fixtures were left unchanged.

The direct `simple-git` devDependency declarations were also bumped to `^3.36.0` to keep them consistent with the override, satisfying the dependency-version consistency check.

### Validation

- Regenerated every affected lockfile and confirmed the intended versions resolve.
- `flub check policy` passes.
- Lockfile supply-chain policy checks pass.

These are override + lockfile changes plus matching version-range alignment, so no changesets are required.


Abe27342 (approved):
looks fine, but double-checking: have you confirmed overrides are actually necessary? as in, cannot update root dependencies such that transitive deps are updated to meet the patched versions?

frankmueller-msft:
@Abe27342 Good question — I checked this per-package against the dependency graph rather than assuming. Short version: overrides are necessary for the bulk of these, and are the cleaner lever for the rest.

**Most of these are transitive-only** — they aren't declared as direct dependencies anywhere in the repo, so there is no root/direct dependency to bump:

| Package | Direct declarations | Pulled transitively by (examples) |
| --- | --- | --- |
| `fast-uri` | 0 | `ajv` |
| `tmp` | 0 | `fengari` |
| `form-data` | 0 | various HTTP clients |
| `systeminformation` | 0 | `pm2-sysmonit` |
| `@nevware21/ts-utils` | 0 | app-insights chain |
| `@github/copilot` | 0 | tooling chain |

For these, the intermediate parents' semver ranges **already permit** the patched version (e.g. `ajv` depends on `fast-uri ^3.0.0`, which allows `3.1.x`). So there is no newer parent to adopt — the patched version is in-range, but a plain install can leave the older resolution pinned in the lockfile. The override forces the in-range floor and, importantly, keeps it pinned so a future lockfile regeneration can't drift back.

**The two that *are* direct deps still need the override to cover other lines:**
- `axios` — declared directly as `^1.15.1` (server/examples). That caret covers the 1.x line, but `axios` is *also* pulled transitively on the pre-1.0 line (`0.30.x`/`0.31.x`) via older Fluid packages, which a direct 1.x bump can't reach. The two `axios@...` overrides cover both lines.
- `ws` — declared directly as `^7.5.10` (routerlicious). That caret covers 7.x, but `ws` is also pulled transitively on the 8.x line via `socket.io`/`engine.io`/`jsdom`/`puppeteer-core`/`openai`. The `ws@>=8 <9` + `ws@>=7 <8` overrides cover every path with one lever.

Finally, this is the repo's existing convention — `pnpm-workspace.yaml` already carried overrides for `axios`, `simple-git`, `systeminformation`, `tar`, `qs`, `express`, `serialize-javascript`, etc. for exactly this reason. I extended that established pattern rather than introduce a new mechanism, and kept the change to the workspaces that actually resolved an affected version.

Abe27342:
> @Abe27342 Good question — I checked this per-package against the dependency graph rather than assuming. Short version: overrides are necessary for the bulk of these, and are the cleaner lever for the rest.
> 
> **Most of these are transitive-only** — they aren't declared as direct dependencies anywhere in the repo, so there is no root/direct dependency to bump:
> 
> Package	Direct declarations	Pulled transitively by (examples)
> `fast-uri`	0	`ajv`
> `tmp`	0	`fengari`
> `form-data`	0	various HTTP clients
> `systeminformation`	0	`pm2-sysmonit`
> `@nevware21/ts-utils`	0	app-insights chain
> `@github/copilot`	0	tooling chain
> For these, the intermediate parents' semver ranges **already permit** the patched version (e.g. `ajv` depends on `fast-uri ^3.0.0`, which allows `3.1.x`). So there is no newer parent to adopt — the patched version is in-range, but a plain install can leave the older resolution pinned in the lockfile. The override forces the in-range floor and, importantly, keeps it pinned so a future lockfile regeneration can't drift back.
> 
> **The two that _are_ direct deps still need the override to cover other lines:**
> 
> * `axios` — declared directly as `^1.15.1` (server/examples). That caret covers the 1.x line, but `axios` is _also_ pulled transitively on the pre-1.0 line (`0.30.x`/`0.31.x`) via older Fluid packages, which a direct 1.x bump can't reach. The two `axios@...` overrides cover both lines.
> * `ws` — declared directly as `^7.5.10` (routerlicious). That caret covers 7.x, but `ws` is also pulled transitively on the 8.x line via `socket.io`/`engine.io`/`jsdom`/`puppeteer-core`/`openai`. The `ws@>=8 <9` + `ws@>=7 <8` overrides cover every path with one lever.
> 
> Finally, this is the repo's existing convention — `pnpm-workspace.yaml` already carried overrides for `axios`, `simple-git`, `systeminformation`, `tar`, `qs`, `express`, `serialize-javascript`, etc. for exactly this reason. I extended that established pattern rather than introduce a new mechanism, and kept the change to the workspaces that actually resolved an affected version.

Sounds like in a lot of these cases we should be using scoped overrides then. Historically we have advocated for using overrides to bump package versions, then not committing the override changes. But this does rely on package manager install behavior to not downgrade such versions on future installs.

Main danger of using unscoped overrides is it can cause headache in the future if the override actually forces a package to downgrade its major version.

PR #27659, opened 2026-07-08
fix(container-runtime): prevent invalid handle reuse after parent skip recursion

jzaffiro:
## Description

Fixes a summarizer bug that caused ODSP `404 fluidElementNotFound` errors during incremental summarization, most visibly for containers using SharedTree as the root DDS:

```
ODSP fetch error [404] ... fluidElementNotFound
Cannot locate node with path '.app/.channels/<dataStore>/.channels/root' under '<snapshotId>'
```

### Root cause

When a `SummarizerNode` emits a `SummaryType.Handle` (skips recursion), its subtree's content is not directly stored in the resulting summary version — the child paths are only reachable transitively through the parent's handle reference. However, `completeSummaryCore` was still promoting each descendant's `_lastSummaryReferenceSequenceNumber` (via `refreshLatestSummaryFromPending`) as if their content had been written.

On a later summary, if the parent became dirty and re-summarized while a descendant had not changed, the descendant would emit a handle pointing back to the version where its content only lived behind the parent's handle. The ODSP snapshot upload path (`odspSummaryUploadManager.writeSummaryTree`) then constructed an id like `${parentHandle}/${rootNodeName}${handlePath}` that the server cannot resolve, because it does not follow nested handle references, yielding `fluidElementNotFound`.

### Why this is disproportionately visible on SharedTree

SharedTree's `ForestIncrementalSummaryBuilder` emits many nested chunk-level handles inside the DDS itself — a pattern unique among current DDSes. Workloads with large trees and frequent GC state churn on the containing data store made the bad handle emission a near-certainty. Other DDSes only reuse handles at the DDS-root boundary and largely dodge the bug.

### Fix

In `SummarizerNode.completeSummaryCore`, when `parentSkipRecursion` is true, clear `_lastSummaryReferenceSequenceNumber` on the entire descendant subtree (not just newly-created nodes, which was the prior guard). This forces those descendants to emit a full `SummaryType.Tree` the next time they are summarized instead of an unresolvable handle. Once they've been directly summarized and acked, normal handle reuse resumes.

The clearing also indirectly disables SharedTree's chunk-level incremental handles for that next summary, since `summarizerNode.ts` passes `incrementalSummaryContext: undefined` when there is no tracked reference sequence number.

### Cost

Bounded and rare: it only triggers on the summary immediately after a parent used a handle *and* then became dirty. Cost is one extra full serialization of the affected subtree, one time per cycle. The prior behavior produced broken summaries and forced full retries, which is strictly worse.

### Repro / verification

Two regression tests were added in `summarizerNode.spec.ts` under `Handle Reuse After Parent Skip Recursion`:

1. After a parent emits a handle and later becomes dirty, an unchanged descendant must emit a full tree (not a handle).
2. After that forced re-summarization is acked, normal handle reuse resumes on the following summary.

All 998 `container-runtime` tests pass locally. Linting and formatting are clean.

## Reviewer Guidance

- Please sanity-check the interaction with `SummarizerNodeWithGc.completeSummaryCore`. With the fix, when the parent skips recursion, `super.completeSummaryCore` does not add the descendant to `pendingSummaries`; the GC subclass's lookup then gracefully no-ops, and the 15 GC-focused tests still pass.
- I considered narrowing the fix to only clear descendants that had actually been loaded from a previous summary (rather than "all descendants"), but the extra bookkeeping did not seem worth it given how rarely this path fires and how surgical the fallout is (one extra tree serialization).

### DIFF
--- a/packages/runtime/container-runtime/src/summary/summarizerNode/summarizerNode.ts
+++ b/packages/runtime/container-runtime/src/summary/summarizerNode/summarizerNode.ts
@@ -345,13 +345,17 @@ export class SummarizerNode implements IRootSummarizerNode {
 			this.wipReferenceSequenceNumber !== undefined,
 			0x1a4 /* "Not tracking a summary" */,
 		);
-		if (parentSkipRecursion && this._lastSummaryReferenceSequenceNumber === undefined) {
-			// This case the child is added after the latest non-failure summary.
-			// This node and all children should consider themselves as still not
-			// having a successful summary yet.
-			// We cannot "reuse" this node if unchanged since that summary, because
-			// handles will be unable to point to that node. It never made it to the
-			// tree itself, and only exists as an attach op in the _outstandingOps.
+		if (parentSkipRecursion) {
+			// When the parent skipped recursion (emitted a handle instead of a full subtree),
+			// this node's content is not directly stored in the new summary version. It is only
+			// transitively accessible via the parent's handle reference. We must clear
+			// _lastSummaryReferenceSequenceNumber so that this node does not attempt to emit a
+			// handle in future summaries pointing to this version, because the storage service
+			// cannot resolve nested paths through handle references.
+			this._lastSummaryReferenceSequenceNumber = undefined;
+			for (const child of this.children.values()) {
+				child.completeSummaryCore(proposalHandle, true);
+			}
 			this.clearSummary();
 			return;
 		}
--- a/.changeset/summarizer-handle-after-parent-skip.md
+++ b/.changeset/summarizer-handle-after-parent-skip.md
@@ -1,0 +1,16 @@
+---
+"@fluidframework/container-runtime": minor
+"__section": fix
+---
+Fix summarizer emitting invalid nested handles after a parent skipped recursion
+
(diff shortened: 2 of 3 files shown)


agarwal-navin:
I don't understand what the bug is. What's described in `Root-cause` seems like expected behavior to me and that should not be a bug. Then there is the `Why this is disproportionately visible on SharedTree` which kinds tells me the bug is actually in `ForestIncrementalSummaryBuilder`? But it isn't clear to me what the bug is there or how it's introduced.
@jzaffiro @markfields FYI

packages/runtime/container-runtime/src/summary/summarizerNode/summarizerNode.ts
```diff
@@ -345,13 +345,17 @@ export class SummarizerNode implements IRootSummarizerNode {
...
-			// handles will be unable to point to that node. It never made it to the
-			// tree itself, and only exists as an attach op in the _outstandingOps.
+		if (parentSkipRecursion) {
+			// When the parent skipped recursion (emitted a handle instead of a full subtree),
+			// this node's content is not directly stored in the new summary version. It is only
+			// transitively accessible via the parent's handle reference. We must clear
+			// _lastSummaryReferenceSequenceNumber so that this node does not attempt to emit a
+			// handle in future summaries pointing to this version, because the storage service
+			// cannot resolve nested paths through handle references.
+			this._lastSummaryReferenceSequenceNumber = undefined;
```
agarwal-navin, line 355: This is not right. This breaks incremental summary - basically whenever a parent hasn't changed, in the next summary, all of it's children will re-summarize.

The comment is also incorrect / unclear. Handles are a client / runtime concept - when the service sees handles, it basically gets the summary tree from the previous summary and replaces the handle with it. So, in the next summary, if a child doesn't change and uses handle to point to this summary, the server should find the contents just fine.

PR #27660, opened 2026-07-08
comparison forest

daesunp:
## Description

Adds `ComparisonForest` which wraps a main forest and a reference forest. Every delta is applied to both forests, and their contents are asserted to be equal after each delta.

`ForestTypeExpensiveDebug` now uses `ComparisonForest`, and the top-level, move, and undoRedo fuzz tests also run it again it.


packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
```diff
@@ -0,0 +1,258 @@
...
+	public deregisterAnnouncedVisitor(visitor: () => AnnouncedVisitor): void {
+		this.main.deregisterAnnouncedVisitor(visitor);
+	}
+
+	public acquireVisitor(): DeltaVisitor {
+		const main = this.main;
+		const reference = this.reference;
+		const mainVisitor = main.acquireVisitor();
+		const referenceVisitor = reference.acquireVisitor();
+		const visitor: DeltaVisitor = {
```
CraigMacomber: Can you implement this using our existing `combineVisitors(visitors: readonly CombinableVisitor[]): CombinedVisitor` util?

CraigMacomber: maybe combine those two visitors with a third that just has the assertForestsEqual on free?

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
```diff
@@ -0,0 +1,258 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
```
CraigMacomber, line 2: this file should be in our test code not in the production code, and should use existing tree comparison utils (I think we have an assertTreeEqual or some exiting way to compare forests) it can use.

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * @remarks
+ * Object keys (for example field keys) are compared as an unordered set, so this is independent of the
+ * order in which different forest implementations happen to enumerate fields. Array indices are just
+ * numeric keys, so array elements are still compared position-by-position.
+ *
+ * This is implemented iteratively (using an explicit work stack rather than recursion) so that comparing
+ * deeply nested trees does not risk exhausting the call stack.
+ */
+function deepEqual(a: unknown, b: unknown): boolean {
```
CraigMacomber: Avoid trying to deep compare "unknown". Its always going to be problematic (cycles, identify of functions etc) and thus only work for more specific cases, which we should simply describe using the types.

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
CraigMacomber: Actually, I guess hooking thus up via a exported Forest type (expensive debug) means this stays out of the bundle for people who don't use it. Maybe this is fine? That does leave us without access to our nice tree comparison utilities with their existing logic for making nice errors.

Maybe moving those into here and reusing them would be a better solution? I would like to avoid having an extra copy of tree compare with nice human facing error message logic.

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
```diff
@@ -0,0 +1,258 @@
...
+ * @throws an Error describing the divergence if the forests differ.
+ */
+export function assertForestsEqual(
+	main: IForestSubscription,
+	reference: IForestSubscription,
+): void {
+	const mainContent = detachedFieldsContent(main);
+	const referenceContent = detachedFieldsContent(reference);
+	if (!deepEqual(mainContent, referenceContent)) {
+		fail(
```
CraigMacomber: this is invalid, and CI should detect it. You cannot use fail with a non constant string for the first parameter.

You can however use the API I added where you can have a debug build only second parameter which computes a nicer message

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
```diff
@@ -0,0 +1,258 @@
...
+			this.reference.clone(schema, breaker),
+		);
+	}
+
+	public chunkField(cursor: ITreeCursorSynchronous): TreeChunk[] {
+		return this.main.chunkField(cursor);
+	}
+
+	public allocateCursor(source?: string): ITreeSubscriptionCursor {
+		return this.main.allocateCursor(source);
```
CraigMacomber, line 91: I imagine it would be a bit of a pain to implement, but this could validate the cursors match as well, by pairing together both options. Lets not do that yet, but note it as possible:


```suggestion
		// Currently we just use the cursor from "main", but we could use "reference" or a custom cursor ensuring both match in behavior.
		// TODO: Add a custom combined cursor type which validates two cursors match, and use it here.
		return this.main.allocateCursor(source);
```

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
daesunp: Replaced with:
```
const comparisonVisitor = createAnnouncedVisitor({ free: () => assertForestsEqual(main, reference) });
return combineVisitors([mainVisitor, referenceVisitor, comparisonVisitor]);
```

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
daesunp: Updated to:
```
function forestContentEquals(
	main: ReadonlyMap<FieldKey, JsonableTree[]>,
	reference: ReadonlyMap<FieldKey, JsonableTree[]>,
): boolean
```

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
daesunp: fail now takes a constant string first `"Forests are not equal"` and diff moved into debug-only builder function.

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
daesunp: Updated with suggestion

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
daesunp: moved comparison utilities and kept in production code

packages/dds/tree/src/feature-libraries/treeTextCursor.ts
```diff
@@ -132,3 +133,120 @@ export function jsonableTreeFromForest(forest: IForestSubscription): JsonableTre
...
+
+/**
+ * Best-effort human readable serialization of forest content for error messages.
+ * Falls back to a placeholder if the content is too deeply nested to serialize.
+ */
+function describeContent(content: ReadonlyMap<FieldKey, JsonableTree[]>): string {
+	try {
+		return JSON.stringify(Object.fromEntries(content));
+	} catch {
+		return "<content too large to serialize>";
```
CraigMacomber: Does thisd really fail because content it too large, or due to the fact that it doesn't support Handles, and cursucrives foerever when hitting one?

In general catching an exception, and making no effort to determin waht that exception is other than assuming something is a bad pattern as it can be wrong and that being wrong can be hard to notice.

packages/dds/tree/src/feature-libraries/treeTextCursor.ts
```diff
@@ -132,3 +133,120 @@ export function jsonableTreeFromForest(forest: IForestSubscription): JsonableTre
...
+}
+
+/**
+ * Structural equality for the {@link JsonableTree} content of two forests, keyed by detached field.
+ *
+ * @remarks
+ * Fields are compared as an unordered set of keys, so this is independent of the order in which different
+ * forest implementations enumerate fields. The nodes within each field are compared in order.
+ *
+ * Implemented iteratively (using an explicit work stack rather than recursion) so that comparing deeply
```
CraigMacomber: We already have plenty of code which walks trees recursivly: the extra complicty to do this is not necessary, especially given this code is only for an expensive debug option

packages/dds/tree/src/feature-libraries/treeTextCursor.ts
```diff
@@ -132,3 +133,120 @@ export function jsonableTreeFromForest(forest: IForestSubscription): JsonableTre
 	readCursor.free();
 	return jsonable;
 }
+
+/**
+ * Extracts the full contents of a forest (every detached field, keyed by field key) as {@link JsonableTree}s.
+ */
+function detachedFieldsContent(forest: IForestSubscription): Map<FieldKey, JsonableTree[]> {
+	const cursor = forest.getCursorAboveDetachedFields();
+	const content = new Map<FieldKey, JsonableTree[]>();
```
CraigMacomber: This is a rather odd choice. Genreally we copy content into JsonableTree or map trees for comparison in tests so that node's deepEqual can print a nice diff for us. If we arn't going to rely on that, doing the comparison using the cursors directly seems like it would make more sense.

packages/dds/tree/src/feature-libraries/treeTextCursor.ts
```diff
@@ -132,3 +133,120 @@ export function jsonableTreeFromForest(forest: IForestSubscription): JsonableTre
...
+	for (let hasField = cursor.firstField(); hasField; hasField = cursor.nextField()) {
```
CraigMacomber: There is no need to reimplement mapCursorFields and jsonableTreeFromFieldCursor, yuo can just use them `new Map(mapCursorFields(cursor, () => [cursor.getFieldKey(), jsonableTreeFromFieldCursor(cursor)]))`

But you don't even need to do that since that is stil just logically reimplementing jsonableTreeFromCursor. If you do:
```
const content = jsonableTreeFromCursor(cuirsor);
```
Then you don't have this odd special map case at the root, and forestContentEquals can become just a tree compare and take in the root JsonableTree node and be simplified significantly.

packages/dds/tree/src/feature-libraries/treeTextCursor.ts
```diff
@@ -132,3 +132,68 @@ export function jsonableTreeFromForest(forest: IForestSubscription): JsonableTre
...
+	}
+	for (const key of aKeys) {
+		const aField = getGenericTreeField(a, key, false);
+		const bField = getGenericTreeField(b, key, false);
+		if (aField.length !== bField.length) {
+			return false;
+		}
+		for (const [index, aChild] of aField.entries()) {
+			const bChild = bField[index];
+			if (bChild === undefined || !treesEqual(aChild, bChild)) {
```
CraigMacomber: Since you checked the lengths, it should never be undefined. Since bChild is a JsonableTree and thus can't be null, we can check it like this:
```suggestion
			const bChild = bField[index] ?? oob();
			if (!treesEqual(aChild, bChild)) {
```
That will indicate a bug in our code if we get undefined (which is good as that would require a bug), rather than returning false.

CraigMacomber: Since you checked the lengths, it should never be undefined. Since bChild is a JsonableTree and thus can't be null, we can check it like this:
```suggestion
			const bChild = bField[index] ?? oob();
			if (!treesEqual(aChild, bChild)) {
```
That will indicate a bug in our code if we get undefined (which is good as that would require a bug), rather than returning false.

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	public acquireVisitor(): DeltaVisitor {
+		const main = this.main;
+		const reference = this.reference;
+		const mainVisitor = main.acquireVisitor();
+		const referenceVisitor = reference.acquireVisitor();
+		// A visitor which does nothing except assert the two forests match once the delta has been fully applied to both.
+		const comparisonVisitor = createAnnouncedVisitor({
+			free: () => assertForestsEqual(main, reference),
```
CraigMacomber: Lets put assertForestsEqual(main, reference) in a methods (maybe assertValid) on this class so its easier to call in other places.

Then lets add a call to it in the constructor, and also clone (so we can see if the corruption happens before the clone, or from closing), and acquireVisitor (to see if the issue was there before the edit)

packages/dds/tree/src/test/feature-libraries/comparison-forest/comparisonForest.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+	makeTreeChunker,
+} from "../../../feature-libraries/index.js";
+import { Breakable } from "../../../util/index.js";
+import { testForest } from "../../forestTestSuite.js";
+import { initializeForest } from "../../feature-libraries/index.js";
+import { fieldJsonCursor } from "../../json/index.js";
+import { jsonSequenceRootSchema } from "../../sequenceRootUtils.js";
+import { testIdCompressor, testRevisionTagCodec } from "../../utils.js";
+
+describe("ComparisonForest", () => {
```
CraigMacomber, line 26: We should have a test case where we introduce a difference in the forest, then assert it catches it.

ComparisonForest exposes the underlying forests, so you should be able to use that to make a change to one of them, then do an operation of the comparison forest and assert it throws.

packages/dds/tree/src/feature-libraries/treeTextCursor.ts
daesunp: simplified to just `JsonableTree` per forest

packages/dds/tree/src/feature-libraries/treeTextCursor.ts
daesunp: updated with suggestion

packages/dds/tree/src/feature-libraries/treeTextCursor.ts
daesunp: replaced iterative solution with recursive `treesEqual` function

packages/dds/tree/src/feature-libraries/treeTextCursor.ts
daesunp: Updated with suggestion

packages/dds/tree/src/feature-libraries/comparison-forest/comparisonForest.ts
daesunp: Updated with suggestion

packages/dds/tree/src/test/feature-libraries/comparison-forest/comparisonForest.spec.ts
daesunp: Added test which initializes only the main forest, and verifies that `forest.aquireVisitor()` throws

packages/dds/tree/src/feature-libraries/treeTextCursor.ts
daesunp: Updated with suggestion

CraigMacomber (approved):

CraigMacomber (approved):

PR #27663, opened 2026-07-09
feat: updated registry in pnpm-workspace.yaml

MarioJGMsoft:
## Description

There was an update that motivates us to update the repo so that it points to the public registry feed that we use in our PR builds.

## Reviewer Guidance

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -55,8 +55,6 @@ minimumReleaseAgeExclude: [
   # Temporary exceptions go here as needed.
 ]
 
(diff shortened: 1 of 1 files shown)


Abe27342 (approved):

ChumpChief (approved):
Seems good - I'm guessing we'll also need to do the same for the other repos we maintain (FluidExamples, FluidHelloWorld, maybe some internal ones?)

Not blocking: @Abe27342 any thoughts on using our public registry vs. the https://packagefeedproxy.microsoft.io/npm/ mentioned in the announcement (I don't know anything about it)

PR #27664, opened 2026-07-09
Add simpler way to use Component

CraigMacomber:
## Description

Add overload and default parameters to easy simple case use of Component API

## Reviewer Guidance


packages/dds/tree/src/componentApi.ts
```diff
@@ -234,30 +256,50 @@ export namespace Component {
...
-	): ComposedComponents<TConfig, TComponent> {
+	export function composeComponents<TComponent, TConfig>(
+		allComponents: readonly Factory<TComponent, TConfig>[],
+		lazyConfiguration: (composed: ComposedComponents<TComponent, TConfig>) => TConfig,
+	): ComposedComponents<TComponent, TConfig>;
+	export function composeComponents<TComponent, TConfig>(
+		allComponents: readonly Factory<TComponent, TConfig>[],
+		lazyConfiguration: (composed: ComposedComponents<TComponent, TConfig>) => TConfig = (
+			composed,
+		) => composed as unknown as TConfig,
```
CraigMacomber: I generally don't like overloads as they add this kind of non-type safe stuff.

In this case the other option is two differently named functions (since generic inference won't let you get away with one generic function for this)

I'm thinking maybe simply "compose" and "composeCustom". These seem fine especially given the functions are nested in a namespace already, which avoids the issue of having a top level too generic sounding compose, and makes the "custom" one not have too long of a name.

Thoughts on which is better?

Josmithr: I personally like the user-experience of overloads for discoverability, but I agree that in this case (especially with the proposed naming) it matters less since intellisense will surface both options for autocomplete and such.

So, I guess I have a slight preference for the overloads, given that we only need the one type-cast, but I'd be fine with either.

CraigMacomber: Kept the overload, but did the renames to the shorter name.

Josmithr (approved):
This is definitely a notable simplification, in my opinion. I still worry about how approachable this API is overall, but this is definitely better.

CraigMacomber:
@microsoft-github-policy-service rerun

PR #27649, opened 2026-07-07
test(tree): sandboxing demo

yann-achard-MS:
## Description

Adds an example architecture for exposing a SharedTree view in a sandbox.

## Breaking Changes

None

## Reviewer Guidance

See the `src/test/shared-tree/sanboxing/sandboxing.md` file for more details.

PR #27665, opened 2026-07-09
feat(tree): no-commit option on `applyChange`

yann-achard-MS:
## Description

Adds an optional parameter to `TreeBranchAlpha.applyChange` that makes it possible not to generate a commit for the applied change. This makes sense when the change being applied does not correspond to a real commit but instead correspond to rebase.

This is used in https://github.com/microsoft/FluidFramework/pull/27649.

## Breaking Changes

None

### DIFF
--- a/packages/dds/tree/src/simple-tree/api/tree.ts
+++ b/packages/dds/tree/src/simple-tree/api/tree.ts
@@ -356,6 +356,8 @@ export interface TreeBranchAlpha extends TreeBranch, TreeContextAlpha {
 	/**
 	 * Apply a serialized change to this branch.
 	 * @param change - the change to apply.
+	 * @param generateCommit - whether to generate a commit for this change. Defaults to true.
+	 * Typically set to false when applying the net change corresponding to a rebase.
 	 * Changes are acquired via `getChange` in a branch's {@link TreeBranchEvents.changed | "changed"} event.
 	 * @remarks Changes may only be applied to a SharedTree with the same IdCompressor instance and branch state from which they were generated.
 	 * They may be created by one branch and applied to another, but only if both branches share the same history at the time of creation and application.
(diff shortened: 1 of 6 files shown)


packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -364,7 +366,7 @@ export interface TreeBranchAlpha extends TreeBranch, TreeContextAlpha {
 	 * TODO: This method will support applying changes from different IdCompressor instances as long as they have the same local session ID.
 	 * Update the tests and docs to match when that is done.
 	 */
-	applyChange(change: JsonCompatibleReadOnly): void;
+	applyChange(change: JsonCompatibleReadOnly, generateCommit?: boolean): void;
```
Josmithr, line 369: Can this be simplified to 
```suggestion
	applyChange(change: JsonCompatibleReadOnly, generateCommit: boolean = true): void;
```

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -356,6 +356,8 @@ export interface TreeBranchAlpha extends TreeBranch, TreeContextAlpha {
 	/**
 	 * Apply a serialized change to this branch.
 	 * @param change - the change to apply.
+	 * @param generateCommit - whether to generate a commit for this change. Defaults to true.
+	 * Typically set to false when applying the net change corresponding to a rebase.
```
Josmithr, line 360: What do we mean by "generate a commit" here? This isn't immediately clear to me. Might be worth expanding on that here and/or expanding on the rebase example.

Josmithr: Based on the tests below, this seems to be whether or not we emit change events in response to the applied change? If so, I think we should clarify that here.

packages/dds/tree/src/simple-tree/api/tree.ts
yann-achard-MS: It cannot. Initializers are only allowed in function implementations.

Josmithr: Ugh. Right. 😢

PR #27666, opened 2026-07-09
Migrate website search to Pagefind

MarioJGMsoft:
## Description

Replace the website's hosted search integration with a static Pagefind index so search works without deployment credentials and can be validated from a local production build.

The website build now generates the Pagefind index after Docusaurus, and a local `SearchBar` theme override renders Pagefind's accessible component UI. The Pagefind module loads after React hydration to ensure the live navbar trigger controls the active modal. Search results label documentation as v1 or v2 and favor current v2 content while preserving normal relevance ranking. This also removes obsolete search configuration and deployment variables, updates search styling and documentation, and adds end-to-end coverage for written docs, generated API docs, version labeling and ranking, and search after client-side navigation.

For a before and after view of the search in the website, see linked item.

## Reviewer Guidance

The post-hydration script loading in `SearchBar` is intentional; loading the custom elements before hydration can leave the visible trigger connected to a detached modal.

### Known regressions / follow-up items

- **Tune conceptual versus API relevance.** Queries such as `SharedTree` still prioritize generated API references over the current quick-start content. Boost conceptual documentation relative to generated API pages without removing relevant API matches.
- **Reduce noisy and duplicate matches.** `Fluid Framework` returns 1,198 matches instead of 133 and exposes many heading-level links. Scope indexing to primary content and tune which versions and headings are indexed.
- **Restore a shareable all-results view.** The current search links to `/search?q=...`; the Pagefind modal has no URL-backed equivalent. Add a dedicated search route with query state.
- **Restore recent-search history.** The current search persists and displays recent queries; the Pagefind modal does not. Add local query history or explicitly document that this feature will not be retained.

Validated with:

- `npm run lint:fix`
- `npm run build`
- `validateHashes.sh`
- `npm run test` (6 passed)

Fixes [AB#77189](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/77189)


WayneFerrao:
Thanks for doing this! This is a huge step towards maintaining search while reducing overhead 🎉


A few thoughts:

Can you document details in the PR description on any regressions that this migration might add?
Does pagefind generate results from v1 and v2? Or do we lose versioning and are ok with that?
Does this search cover written docs as well as api docs? If so, add tests to cover all cases.

MarioJGMsoft:
> Thanks for doing this! This is a huge step towards maintaining search while reducing overhead 🎉
> 
> A few thoughts:
> 
> Can you document details in the PR description on any regressions that this migration might add? 
> Does pagefind generate results from v1 and v2? Or do we lose versioning and are ok with that? 
> Does this search cover written docs as well as api docs? If so, add tests to cover all cases.

I updated the PR description with the regressions that we can do in follow up work.
The pagefind results are from both v1 and v2 docs. The issue here is that the v1 docs are shown first before the v2.
Edit: I was able to add logic to prioritize more v2 docs, making them higher priority (they show up first when you search something.)
It does link both, API and Written docs, I can add the tests to confirm this.

website/src/css/notFound.css
```diff
@@ -29,7 +29,7 @@
 
 .ffcom-not-found-description {
 	font-size: 18px;
-	color: var(--docsearch-muted-color);
+	color: #666c77;
```
CraigMacomber: Doesn't this force the same color in dark and non-dark themes? Is there a reason to remove the use of what I assume is a theme-able variable here? The choice to delete the themed variable and replace it with a hard coded value from one of the themes seems odd.

MarioJGMsoft: It doesn't because if you look in the file, it has this code: `[data-theme="dark"] .ffcom-not-found-description {color: #798a9e;}` which from my understanding, overrides the color if the notFound page is in dark theme.

website/src/theme/DocItem/Metadata/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useDoc } from "@docusaurus/plugin-content-docs/client";
+import OriginalDocItemMetadata from "@theme-original/DocItem/Metadata";
+import type { ReactElement } from "react";
+
+/**
+ * Adds the documentation version to Pagefind result metadata.
+ */
+export default function DocItemMetadata(): ReactElement {
+	const { metadata } = useDoc();
+	const version = metadata.version === "current" ? "v2" : `v${metadata.version}`;
```
WayneFerrao: `v2` here is hardcoded to be the same as `current`. The next time we release (v3), this will break. Docs will still render v2 while content is at v3.
Derive the label from the actual current versiona number rather than a constant.

Make the version config the single source of truth instead of a constant. Read the configured label via useDocsVersion():

```typescript
import { useDocsVersion } from "@docusaurus/plugin-content-docs/client";

const { label } = useDocsVersion();
// <meta data-pagefind-meta={`version:${label}`} />
```

and set the labels in docusaurus.config.ts under the docs preset's versions block (current: { label: "v2" }, "1": { label: "v1" }). Then a release is a one-line config bump — no theme code changes.

WayneFerrao (changes requested):

website/src/theme/DocItem/Metadata/index.tsx
MarioJGMsoft: Thanks for the feedback. Applied change

WayneFerrao (approved):
Thanks for doing this! Excited for a new search experience without the maintenance overhead. I look forward to the follow up tasks adding more value and usability as well 🚀

WayneFerrao (approved):

PR #27668, opened 2026-07-10
Throw DataCorruptionError for meaningful duplicate batch detections

dannimad:
## Description

Previously, duplicate-batch detection only logged a `DuplicateBatch` telemetry event and never threw the `DataCorruptionError` — this was a temporary mitigation for a live service bug that could redeliver batches.

This PR re-enables throwing, but only when either the incoming batch or the previously-recorded batch has an **explicit** `batchId` (i.e., it came from a resubmit via `PendingStateManager.replayPendingStates`), which indicates a genuine duplicate-batch scenario (e.g. container forking) rather than the known service-outage artifact. Duplicates where neither batch has an explicit `batchId` continue to be log-only.

Updated the `Duplicate Batch Detection` tests in `containerRuntime.spec.ts` to assert the new throwing behavior where applicable, and added a new test covering the still-log-only case (no explicit `batchId` on either side).

## Reviewer Guidance

### DIFF
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -3244,13 +3244,14 @@ export class ContainerRuntime
 						"Duplicate batch - The same batch was sequenced twice",
 						{ batchId: batchStart.batchId },
 					);
-
+					const batchIdExplicit = batchStart.batchId !== undefined;
+					const otherBatchIdExplicit = result.otherBatchInfo?.batchIdExplicit ?? false;
 					this.mc.logger.sendTelemetryEvent(
 						{
 							eventName: "DuplicateBatch",
 							details: {
 								batchId: batchStart.batchId,
-								batchIdExplicit: batchStart.batchId !== undefined,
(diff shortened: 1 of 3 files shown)


packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3244,13 +3244,14 @@ export class ContainerRuntime
 						"Duplicate batch - The same batch was sequenced twice",
 						{ batchId: batchStart.batchId },
 					);
-
+					const batchIdExplicit = batchStart.batchId !== undefined;
+					const otherBatchIdExplicit = result.otherBatchInfo?.batchIdExplicit ?? false;
```
markfields, line 3248: Consider a comment reminding that otherBatchInfo may not be known (if loaded from a snapshot) and in this corner case we treat it as false to avoid false positives on the error

markfields (approved):
LGTM!

alexvy86 (approved):
Approving for docs. I don't have a lot of context about the functionality, but the docs seem correct to me from an outsider perspective :) .

PR #19593, opened 2024-02-11
Implement getSnapshot() in odsp driver

jatgarg:
## Description

1.) This api will be used to fetch snapshot by specifying group ids in the trees latest request. 
2.) Also add tests to test the new flow.
3.) Change the logic to locally cache snapshot or not based on the user specified option or whether it is first snapshot fetch call or not.

This is part of data virtualization effort which is documented here: [Design Doc](https://microsoft.sharepoint.com/:w:/t/FluidFramework/ESJQvukxffJNiTZzbNCWky4B891mpOpWD7BHhbOznXsMZg?e=gAUCrx)

PR #16152, opened 2023-06-27
[Data Migration] Add APIs to load a paused container at a specified sequence number

scottn12:
## Description

This PR adds/modifies APIs to enable users to load a container to a specific sequence number, as well as pause the container at that sequence number indefinitely. Please see the [design document (MSFT internal)](https://microsoft.sharepoint.com/:w:/t/FluidFramework/EeMkiaBIum9CuXxIGIQmpdQBuo3IP4NTllFtSHO5IEnCfA?e=32Wa29) to learn more about the purpose, scope, alternatives, and considerations.

## Overview of Changes

The following is an overview of the new/changed functionality when loading a container.  Please note, the mechanisms to load a container to a specified sequence number and to freeze the container are **independent** from each other. However, when used together we can achieve the "Load up to X sequence number and stay there indefinitely" behavior.

### Pausing the container
This PR adds a new API to the existing `IContainerLoadMode` interface. When the new load mode, `pauseAfterLoad`, is set to true, the container will halt processing incoming/outgoing ops indefinitely.

### Loading to a specified sequence number

This PR reworks the relationship between two existing load options: `LoaderHeader.sequenceNumber` and `IContainerLoadMode.opsBeforeReturn`. In the current state of the code there is possible collision between these two options. For example, if `IContainerLoadMode.opsBeforeReturn` is set to `"none"`, but `LoaderHeader.sequenceNumber` is set to a sequence number after the latest snapshot (but a valid/existing sequence number), then `LoaderHeader.sequenceNumber` will essentially "win" and the container will be loaded to the specified sequence number.

In this PR, we formalize the relationship between the two options by requiring that `opsBeforeReturn` be set to the newly added `"sequenceNumber"` option if `LoaderHeader.sequenceNumber` is defined, and vice versa.

## Breaking Changes

This PR will break existing users of `LoaderHeader.sequenceNumber`, since it will require `opsBeforeReturn` to be set to `"sequenceNumber"`.

## Testing

Since `Loader` and `Container` do not have unit tests setup, I added testing for this functionality in the e2e test suite. Specifically, I expanded the "No Delta stream loading mode testing" which tries to test all load mode combinations. This was done in a somewhat hacky way, so I would greatly appreciate feedback/suggestions for testing.

## Reviewer Guidance

I would appreciate feedback on (but not limited to) the following items:

[...]

PR #27669, opened 2026-07-10
Add point-in-time container loading by sequence number

sonalideshpandemsft:
This PR threads a point-in-time load target through the loader stack using an alpha API:

- Adds `LoaderHeader.sequenceNumber` support for point-in-time load targets.
- Adds alpha `ILoadContainerToSequenceNumberProps` and `loadContainerToSequenceNumber`.
- Keeps `loadExistingContainer` as the normal/latest-load helper and does not add historical-load props to its beta API.
- Passes the target through `Loader.resolve`, `Container.load`, `SerializedStateManager.fetchSnapshot`, and `ISnapshotFetchOptionsAlpha`.
- Uses an internal `"sequenceNumber"` load mode so the container can replay to the requested target, pause inbound/outbound queues, and return a read-only historical view.
- Rejects clearly when the requested target cannot be reached because the required trailing ops are unavailable, rather than waiting indefinitely.
- Preserves the existing `loadContainerPaused` path as a separate internal helper; it does not use the new sequence-number load mode.
- Adds alpha availability APIs for checking whether a point in document history appears materializable before attempting a load.

For ODSP, this adds point-in-time base snapshot selection. When `loadToSequenceNumber` is supplied, ODSP skips the latest snapshot path, lists recent versions without cache, reads candidate snapshot attributes, and selects the closest usable base snapshot at or before the requested target. If no usable base snapshot exists, ODSP fails clearly instead of returning latest.

The availability probe is exposed through the standalone alpha `IPointInTimeMaterializationStorageService` capability. ODSP uses it to report whether a point appears `materializable`, is missing a base version, is missing required replay ops, is inaccessible due to permissions, or cannot be determined.


anthony-murphy:
`packages/loader/container-loader/src/createAndLoadContainerUtils.ts:310`

Deep Review: On the generic `loadExistingContainer` path, the returned container never stops at `loadToSequenceNumber` — it loads from an older snapshot and then catches up to *latest*, which is not a point-in-time view.

`loadExistingContainer` injects `headers[LoaderHeader.sequenceNumber] = alphaProps.loadToSequenceNumber` (createAndLoadContainerUtils.ts:310-312) but sets no `loadMode`/`opsBeforeReturn`. In `Container.load`, `loadToSequenceNumber` is consumed *only* by `serializedStateManager.fetchSnapshot` for base-snapshot selection (container.ts:1644-1651); the op-processing `switch (loadMode.opsBeforeReturn)` still branches only on `undefined | "cached" | "all"` — there is no branch that halts at `loadToSequenceNumber`. Only `loadContainerPaused` installs the `opHandler`/`pauseContainer` logic that stops at the target (loadPaused.ts:114-120). So a host on the documented generic path gets an older base snapshot plus catch-up to latest, while the changeset promises the loader will "materialize before returning the container."

PR #16152 formalized requiring `opsBeforeReturn: "sequenceNumber"` whenever `LoaderHeader.sequenceNumber` is defined precisely because otherwise the header "will essentially 'win'". This PR re-threads the header without that coupling or any validation.

**Question for the author:** is the returned container on the generic path intended to stop *at* the target, or is exact point-in-time materialization only supported via `loadContainerPaused`? The answer determines the fix:
- (a) Restore an explicit wait-to-target load mode so `Container.load` halts op processing once `loadToSequenceNumber` is reached (with validation of the `loadMode`/target combination, as #16152 did), plus an end-to-end test asserting the returned container is *at* the target; or
- (b) Narrow the API/docs so only `loadContainerPaused` promises exact materialization and `ILoadExistingContainerPropsAlpha` does not expose `loadToSequenceNumber` on the generic surface.

Update the changeset headline example in `.changeset/add-historical-load-target.md` to match whichever contract is chosen.

packages/loader/container-loader/src/loadPaused.ts
```diff
@@ -109,11 +112,7 @@ export async function loadContainerPaused(
 
 		// We need to setup a listener to stop op processing once we reach the desired sequence number (if specified).
 		opHandler = (): void => {
```
anthony-murphy, line 111: Deep Review: The central new behavior of this PR has no test coverage. The added tests exercise only ODSP base-snapshot selection (fetchSnapshot.spec.ts:345-468) and option plumbing (serializedStateManager.spec.ts:321-366).

There is no test for `loadContainerPaused` replaying from a historical base snapshot and pausing exactly at `loadToSequenceNumber` (loadPaused.ts:114-120), nor for `Container.canMaterializePointInTime` delegating through the storage-adapter chain (`ContainerStorageAdapter` → `RetriableDocumentStorageService`/`ProtocolTreeStorageService` → driver, container.ts:678-682). This is exactly the untested path where the generic-contract gap in the other thread lives — a target-assertion test would have caught it.

Please add:
- A test driving `loadContainerPaused` with a mocked delta stream asserting the container pauses *at* the target, including the "already at target" and "snapshot newer than target" branches.
- A test asserting container-level `canMaterializePointInTime` returns `unknownUnavailable` when the underlying driver lacks the method.

(The `opHandler` dropped the `loadToSequenceNumber !== undefined` guard at loadPaused.ts:114-119, but the early return at loadPaused.ts:86-93 keeps that safe — the removed guard is not itself a live bug; the untested-path concern stands independently.)

sonalideshpandemsft:
> `packages/loader/container-loader/src/createAndLoadContainerUtils.ts:310`
> 
> Deep Review: On the generic `loadExistingContainer` path, the returned container never stops at `loadToSequenceNumber` — it loads from an older snapshot and then catches up to _latest_, which is not a point-in-time view.
> 
> `loadExistingContainer` injects `headers[LoaderHeader.sequenceNumber] = alphaProps.loadToSequenceNumber` (createAndLoadContainerUtils.ts:310-312) but sets no `loadMode`/`opsBeforeReturn`. In `Container.load`, `loadToSequenceNumber` is consumed _only_ by `serializedStateManager.fetchSnapshot` for base-snapshot selection (container.ts:1644-1651); the op-processing `switch (loadMode.opsBeforeReturn)` still branches only on `undefined | "cached" | "all"` — there is no branch that halts at `loadToSequenceNumber`. Only `loadContainerPaused` installs the `opHandler`/`pauseContainer` logic that stops at the target (loadPaused.ts:114-120). So a host on the documented generic path gets an older base snapshot plus catch-up to latest, while the changeset promises the loader will "materialize before returning the container."
> 
> PR #16152 formalized requiring `opsBeforeReturn: "sequenceNumber"` whenever `LoaderHeader.sequenceNumber` is defined precisely because otherwise the header "will essentially 'win'". This PR re-threads the header without that coupling or any validation.
> 
> **Question for the author:** is the returned container on the generic path intended to stop _at_ the target, or is exact point-in-time materialization only supported via `loadContainerPaused`? The answer determines the fix:
> 
> * (a) Restore an explicit wait-to-target load mode so `Container.load` halts op processing once `loadToSequenceNumber` is reached (with validation of the `loadMode`/target combination, as [[Data Migration] Add APIs to load a paused container at a specified sequence number #16152](https://github.com/microsoft/FluidFramework/pull/16152) did), plus an end-to-end test asserting the returned container is _at_ the target; or
> * (b) Narrow the API/docs so only `loadContainerPaused` promises exact materialization and `ILoadExistingContainerPropsAlpha` does not expose `loadToSequenceNumber` on the generic surface.
> 
> Update the changeset headline example in `.changeset/add-historical-load-target.md` to match whichever contract is chosen.

This change is for "generic path intended to stop at the target". By generic path, I mean the generic alpha historical API path:
`loadExisitingContainer({ loadToSequenceNumber })`, not `loadExisitingContainer(...)`
The intended contract is:
- `loadExisitingContainer(...)` with no `loadToSequenceNumber` behaves normally and loads the latest
- `loadExisitingContainer({ loadToSequenceNumber })` is an explicit point-in-time request, so it must stop at the target sequence number. 
- `loadContainerPaused(..., loadToSequenceNumber)` also stops at the target

packages/loader/container-loader/src/container.ts
```diff
@@ -2630,6 +2697,17 @@ export class Container
...
  */
 export interface ContainerAlpha extends IContainer {
+	/**
+	 * Checks whether a point in document history can currently be materialized.
+	 * @remarks
+	 * This optional probe is intended for hosts that want to explain historical load failures before attempting a load.
+	 * If the underlying driver cannot answer, implementations should return `notAvailable` rather than claiming
+	 * the point is unavailable for a specific reason.
+	 */
+	canMaterializePointInTime(
```
anthony-murphy: i think this would be better as a free function, rather than being built into the IContainer interface if possible

packages/loader/container-loader/src/index.ts
```diff
@@ -19,6 +19,7 @@ export {
 	type IContainerLoadDriverProps,
 	type ICreateAndLoadContainerProps,
 	type ICreateDetachedContainerProps,
+	type ILoadExistingContainerPropsAlpha,
```
anthony-murphy: i do not feel like this should just be added to the existing load flow. it is fundamentally different from normal loading, and i think should have its own independent path with its own types. pieces can be reused where it makes sense.

also is the plan for these version to always be readonly? like do with local versions?

packages/common/driver-definitions/src/storage.ts
```diff
@@ -533,3 +585,31 @@ export interface ISnapshotFetchOptions {
...
+	 * Sequence number the loader should materialize before returning the container.
+	 */
+	loadToSequenceNumber?: number;
+}
+
+/**
+ * Alpha storage service extension for point-in-time materialization probes.
+ * @legacy @alpha
+ */
+export interface IDocumentStorageServiceAlpha extends IDocumentStorageService {
```
anthony-murphy: i do no think this should be combined with IDocumentStorageService. it should be a separate interface, as the load path and usage is fundamentally different. it can still be alpha, it can still be in driver definitions, it should just be separate from the existing interfaces

anthony-murphy: this is partially due to the fact that we do not expect other drivers to support this flow generally

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -300,10 +311,21 @@ export async function loadExistingContainer(
 	loadExistingContainerProps: ILoadExistingContainerProps,
 ): Promise<IContainer> {
 	const loader = new Loader(loadExistingContainerProps);
-	return loader.resolve(
-		loadExistingContainerProps.request,
-		loadExistingContainerProps.pendingLocalState,
-	);
+	const alphaProps = loadExistingContainerProps as ILoadExistingContainerPropsAlpha;
+	const headers = { ...loadExistingContainerProps.request.headers };
+	if (alphaProps.loadToSequenceNumber !== undefined) {
```
anthony-murphy: this is probably the wrong mechanism to do this. i think your want to be more like frozen load

anthony-murphy:
Deep Review: Thanks for confirming — `loadExistingContainer({ loadToSequenceNumber })` is meant to stop at the target. That sharpens rather than closes the earlier concern: because the path is *required* to stop at the target, it now hangs uncancellably when the target is unreachable (target in the future, ops not yet flushed, or a sustained 429), and the `connect()` fallback in `loadContainerPaused` becomes unreachable dead code. I've posted the full mechanism and a recommended fix on a new inline thread on `container.ts`, anchored at the new `case "sequenceNumber"` wait. The decision that determines the fix: when the requested sequence number isn't yet available, should the load reject with a clear error, enforce the `canMaterializePointInTime` preflight, or abort/timeout into the wait?

packages/loader/container-loader/src/container.ts
```diff
@@ -1652,6 +1680,43 @@ export class Container
 				);
 				break;
 			}
+			case "sequenceNumber": {
```
anthony-murphy, line 1685: Deep Review: This targeted-load path can hang uncancellably when the requested `loadToSequenceNumber` is unreachable, and the `connect()` fallback in `loadContainerPaused` is now dead code.

The paused/targeted load forces `opsBeforeReturn: "sequenceNumber"` with `deltaConnection: "none"` (loadPaused.ts:63-66). The new `case "sequenceNumber"` builds `opsBeforeReturnP = Promise.all([attachDeltaManagerOpHandler(..., "sequenceNumber", ...), targetReachedP])`, and `targetReachedP` resolves **only** when an op with `sequenceNumber >= loadToSequenceNumber` is processed. Because `deltaConnection` is `"none"`, `DeltaManager` drains delta storage bounded to `loadToSequenceNumber + 1` and stops without opening the WebSocket. If storage cannot supply an op reaching the target — target in the future, ops not yet flushed, or a sustained 429 — `targetReached()` is never called and `await opsBeforeReturnP` blocks forever. No `AbortSignal` is threaded into `Container.load`; the `signal` param in `loadContainerPaused` (loadPaused.ts:126) is wired only into the `opHandler`/`connect()` block.

Separately, after `loadExistingContainer(...)` returns, `Container.load` has already paused at the target, so `lastProcessedSequenceNumber === loadToSequenceNumber` and the "already there" early return (loadPaused.ts:88-96) always fires — making the `opHandler` + `container.connect()` block (loadPaused.ts:107-158), including the comment explaining why `connect()` is needed to fetch ops, unreachable dead code.

Your clarification that `loadExistingContainer({ loadToSequenceNumber })` is intended to stop at the target confirms this is a robustness gap, not an ambiguity: a path that *must* stop at the target hangs when the target is unreachable.

Recommendation: decide and implement the unreachable-target behavior — (a) detect delta-storage exhaustion (the targeted fetch drains without reaching `loadToSequenceNumber`) and reject with a clear "requested sequence number is not yet available" error; and/or (b) thread the `AbortSignal`/a timeout into the `"sequenceNumber"` wait; and/or (c) gate the load behind the new `canMaterializePointInTime` availability probe. Then remove the now-unreachable `opHandler` + `connect()` block (loadPaused.ts:107-158) and its stale comment, or restore a path that keeps it reachable. Add an end-to-end test asserting the targeted load *fails clearly* rather than hangs when the target exceeds available ops.

packages/loader/container-loader/src/index.ts
sonalideshpandemsft: I figured it's best to keep it readonly (similar to local versions) rather than open this old point and keep editing from there. My thinking was, for NITL, copilot changes are auto-applied, and this change lets us inspect the document exactly at the sequence number produced by that copilot change, even if more edits happen afterward.

example:
- seq 10: copilot auto applies change A
- seq 11-20: other users make more edits

Historical load reconstructs state at seq 10.
If it writes, the new op is not inserted at seq 11.
It gets appended at seq 21, after edits 11-20.

sonalideshpandemsft: Also, edits from a old point would be a larger change? but I'm open to thoughts around this

packages/loader/container-loader/src/container.ts
anthony-murphy: Deep Review: Related to the hang above, the same `"sequenceNumber"` / `deltaConnection: "none"` path leaves the read-only historical contract enforced only by convention. `pauseContainer()` pauses the queues, but `handleDeltaConnectionArg("none")` is a bare `break` that never clears `inboundQueuePausedFromInit` — unlike the `undefined`/`"delayed"` cases, which set it `false`. A later `Container.connect()` → `resumeInternal()` sees `inboundQueuePausedFromInit === true`, resumes `inbound`/`inboundSignal`, and advances the container past `loadToSequenceNumber` — so a host can silently walk the "historical" container forward off the target. The predecessor `loadContainerPaused` deliberately calls `container.forceReadonly?.(true)`; this path never does, and the new tests assert only paused queues, not readonly state.

You confirmed [readonly is the intended contract](https://github.com/microsoft/FluidFramework/pull/27669#discussion_r3561494522) (like local versions — a later write appends after latest rather than at the historical point). Convention-only pausing doesn't enforce that intent. Suggest failing closed on the historical path: call `forceReadonly(true)` on the returned container mirroring `loadContainerPaused`, and reject `connect()`/`getPendingLocalState()` on historical containers with a typed `UsageError`, while still allowing local read/UI interaction. Add a test asserting readonly state and that `connect()` does not advance the sequence number.

packages/loader/container-loader/src/container.ts
```diff
@@ -1723,6 +1817,8 @@ export class Container
 					async () => this._deltaManager.inbound.waitTillProcessingDone(),
 				);
 
+				this.forceReadonly(true);
```
anthony-murphy: Deep Review: `this.forceReadonly(true)` (line 1820) sits inside the shared `if (opsBeforeReturnP !== undefined) {` post-load block, not on the historical path. The `opsBeforeReturn` switch sets `opsBeforeReturnP` for `case "sequenceNumber"` **and** for `case "cached"`/`case "all"`, and `Container.load` defaults `loadMode` to `{ opsBeforeReturn: "cached" }` — so ordinary non-historical loads hit this block too. `forceReadonly` sets a sticky `_forceReadonly` override via `connectionManager.forceReadonly(readonly)`, and nothing ever calls `forceReadonly(false)`. Result: a host loading with the default `"cached"` or explicit `"all"` mode gets a container that can never submit ops or reconnect for writes.

This is new code fixing the prior [read-only-by-convention gap](https://github.com/microsoft/FluidFramework/pull/27669#discussion_r3562544991) — the fix is correct in intent but over-scoped in placement. `loadContainerPaused` already scopes its own `forceReadonly(true)` to `pauseAfterLoad`; that's the reference for the correct scoping here.

The `"cached"`/`"all"` paths have no readonly regression test, so CI won't catch this. Add a test asserting default, `"cached"`, and `"all"` loads return a writable container (can submit ops / reconnect for writes) — that's the gap that let the regression escape.

Gate the transition to the historical path:

```suggestion
			if (opsBeforeReturn === "sequenceNumber") {
				this.forceReadonly(true);
			}
```

anthony-murphy: Deep Review: Resolved in `b324507`. `this.forceReadonly(true)` is now scoped to `if (opsBeforeReturn === "sequenceNumber")`, so default/`"cached"`/`"all"` loads are no longer forced permanently read-only, and a new regression test in `loadPaused.spec.ts` asserts normal loads stay writable and can connect. This was the sole gating blocker from the last review — safe to resolve this thread.

packages/drivers/odsp-driver/src/odspDocumentStorageManager.ts
```diff
@@ -260,6 +295,187 @@ export class OdspDocumentStorageService extends OdspDocumentStorageServiceBase {
...
+	private async areReplayOpsAvailable(
+		baseSnapshotSequenceNumber: number,
+		targetSequenceNumber: number,
+		scenarioName: string | undefined,
+	): Promise<boolean> {
+		if (baseSnapshotSequenceNumber >= targetSequenceNumber) {
+			return true;
+		}
+
+		const service = new OdspDeltaStorageService(
```
anthony-murphy, line 385: Deep Review: `canMaterializePointInTime` is presented as a feasibility probe, but `areReplayOpsAvailable` verifies op availability by *actually fetching* every op between `baseSnapshotSequenceNumber + 1` and `targetSequenceNumber` — it loops `while (from <= targetSequenceNumber)` calling `service.get(from, to, ...)` in `opsBatchSize` (default 5000) batches, then inspects only `messages.length`/`message.sequenceNumber` before discarding the payloads. It spins up a fresh `OdspDeltaStorageService` here rather than reusing existing delta-fetch plumbing, and the subsequent real historical load re-downloads the same ops.

The fetch is bounded to the `[base, target]` window in pages, and the availability contract (`storage.ts:160-164`) never promises a cheap metadata-only probe — so this is an optimization on a rare, opt-in alpha path, not a correctness bug. On a large base→target gap, though, the same op range is downloaded twice.

Options: cache the probe's fetched ops so the subsequent real load reuses them, bound the probe range, or provide a metadata-only availability check if ODSP can supply one. If the double-fetch is an accepted alpha cost, record that in `DEV.md`.

packages/loader/container-loader/src/retriableDocumentStorageService.ts
```diff
@@ -79,6 +82,25 @@ export class RetriableDocumentStorageService implements IDocumentStorageService,
 		);
```
anthony-murphy, line 82: Deep Review: `RetriableDocumentStorageService.canMaterializePointInTime` and `RetryErrorsStorageAdapter.canMaterializePointInTime` both wrap the delegate in `runWithRetry(...)`, but the ODSP implementation (`odspDocumentStorageManager.ts:349-396`) wraps its whole body in `try { ... } catch (error) { ... }` and converts *every* error into a terminal status (`authorizationError`/`fileNotFoundOrAccessDeniedError` → `permissionOrAccessDenied`, everything else → `notAvailable`). Because the inner method never throws, `runWithRetry` can never fire — a transient 429/network blip during the probe is reported as terminal `notAvailable` rather than retried.

The probe is advisory and does not itself fail the real load, so this is a Tier-3 polish item. Either drop the dead `runWithRetry` wrappers (they add no resilience), or let genuinely-retryable ODSP errors escape before being normalized to a status so `runWithRetry` can act on them — and document whichever intent is chosen.

packages/drivers/odsp-driver/src/odspDocumentStorageManager.ts
```diff
@@ -260,6 +295,187 @@ export class OdspDocumentStorageService extends OdspDocumentStorageServiceBase {
...
+			this.odspResolvedUrl.endpoints.deltaStorageUrl,
+			this.getAuthHeader,
+			this.epochTracker,
+			this.logger,
+		);
+		const batchSize = this.hostPolicy.opsBatchSize ?? 5000;
+		let from = baseSnapshotSequenceNumber + 1;
```
anthony-murphy, line 392: Deep Review: `fetchHistoricalSnapshot` constructs `const snapshot: ISnapshot = { blobContents: new Map(), ... }` (empty map) before `initializeFromSnapshot`. Unlike the normal `getSnapshot` path (fully-populated blob map), every protocol/app blob read during boot on the historical path becomes an individual `readBlob` round-trip.

The candidate tree was selected via a version/attributes scan (not a full snapshot download), so the blob bodies were never in hand to populate for free — this is a deliberate tradeoff, and the boot-latency cost is on a rare path.

If historical-load boot latency matters, batch-prefetch the candidate tree's blobs into `blobContents` instead of relying on N sequential on-demand `readBlob` calls. Otherwise document the on-demand tradeoff as intentional for the alpha path.

packages/loader/container-loader/src/serializedStateManager.ts
```diff
@@ -257,6 +258,7 @@ export class SerializedStateManager implements IDisposable {
 	public async fetchSnapshot(
 		specifiedVersion: string | undefined,
 		pendingLocalState: IPendingContainerState | undefined,
+		loadToSequenceNumber?: number,
```
anthony-murphy: Deep Review: `SerializedStateManager.fetchSnapshot/getSnapshot/fetchISnapshot` (serializedStateManager.ts:261,567,593), `attachDeltaManagerOpHandler` (deltaManager.ts:581), and `attachOpHandler` (container.ts:2204) declare `loadToSequenceNumber?: number` (optional). In #16152 ChumpChief left an explicit convention on this exact parameter: "Both this and `pendingLocalState` params would be better typed with an `| undefined` rather than an optional." The PR honors it in `Container.load` (`loadToSequenceNumber: number | undefined`) but not in these sibling helpers, so adherence is inconsistent within the same PR.

Type these internal `loadToSequenceNumber` parameters as `number | undefined` rather than optional `?`, matching `Container.load` and the #16152 convention.

packages/loader/container-loader/src/deltaManager.ts
```diff
@@ -612,7 +614,26 @@ export class DeltaManager<TConnectionManager extends IConnectionManager>
...
+				fetchToSequenceNumber === undefined ? undefined : fetchToSequenceNumber + 1,
+			);
+			if (
+				prefetchType === "sequenceNumber" &&
+				fetchToSequenceNumber !== undefined &&
+				this.lastQueuedSequenceNumber < fetchToSequenceNumber
+			) {
+				throw new GenericError(
+					"Cannot satisfy request to load the container at the specified sequence number. Requested sequence number is not available.",
+				);
```
anthony-murphy, line 635: Deep Review: The blocking `"sequenceNumber"` prefetch masks real storage/retry failures as a generic retention error.

This path awaits `fetchMissingDeltasCore(...)` and then, as its *only* failure signal, throws `GenericError("Cannot satisfy request to load the container at the specified sequence number. Requested sequence number is not available.")` from a single guard: `this.lastQueuedSequenceNumber < fetchToSequenceNumber`. But `fetchMissingDeltasCore` catches every fetch error, logs `GetDeltas_Exception`, closes the delta manager, and does **not** rethrow (deltaManager.ts:1233-1264; see #5773, which documents that it "does not throw as fetchMissingDeltasCore catches all exceptions and reports it"). So a storage/auth/network failure during this blocking load never propagates — it is silently converted into the misleading "not available" retention message rather than surfacing the root cause. A transient 429 or auth blip and a genuine retention gap become indistinguishable to the caller.

This is distinct from the earlier [uncancellable-hang discussion](https://github.com/microsoft/FluidFramework/pull/27669#issuecomment-4939241614), which was resolved by adding the fail-fast reject — this is about the *fidelity* of that reject's error.

Recommendation: add a propagation path for the blocking prefetch — return a fetch status or rethrow the normalized fetch error for this awaited call — and reserve the "Requested sequence number is not available" message for the genuine short-read case (`lastQueuedSequenceNumber < fetchToSequenceNumber` with no underlying fetch error). Add a test asserting a transient fetch failure surfaces the underlying error rather than the retention message.

Before applying, confirm against the diff that the new blocking path's only throw is the `lastQueuedSequenceNumber < fetchToSequenceNumber` guard and that `fetchMissingDeltasCore` still swallows — a two-line verification.

anthony-murphy:
`packages/drivers/odsp-driver/src/odspVersionManager.ts:611`

Deep Review: Historical version listing uses the wrong API root and version for ODC/OneDrive origins.

`getOdspFileVersions()` hand-builds the versions URL:

```ts
const siteOrigin = new URL(siteUrl).origin;
const url = `${siteOrigin}/_api/v2.0/drives/${driveId}/items/${itemId}/versions`;
```

But this same module already imports and uses the purpose-built helper `getApiRoot` for its version-scoped URLs (`` `${getApiRoot(new URL(siteUrl))}/drives/${driveId}/items/${itemId}/${version}` ``). `getApiRoot` (`odspUrlHelper.ts:33-40`) returns `` `${url.origin}/${prefix}v2.1` `` and sets `prefix = ""` for ODC/OneDrive origins — so the correct root is `origin/v2.1/...` for ODC and `origin/_api/v2.1/...` for SharePoint. The hand-built string targets `_api/v2.0` for *all* origins: wrong root for ODC (extra `_api/`) and wrong API version (`v2.0` vs `v2.1`) everywhere.

Version enumeration is the entry point of the whole point-in-time flow, so this defeats historical load and the `canMaterializePointInTime` probe for the entire ODC tenant class.

Recommendation: build the versions URL from the existing helper instead of hand-constructing it, e.g. `` const url = `${getApiRoot(new URL(siteUrl))}/drives/${driveId}/items/${itemId}/versions`; `` and remove the now-unused `siteOrigin` local. Update the `_api/v2.0/.../versions` references in `DEV.md` to match, and add a test covering an ODC-origin resolved URL so the root/version selection is pinned.

(Question for the author: was the `_api/v2.0` versions endpoint tested against a OneDrive/ODC-origin document, or only SharePoint? That confirms the tenant-class impact and whether a regression test is needed.)

anthony-murphy:
## Deep Review

*Reviewed commit `a227a02` on 2026-07-13.*

Not ready for sign-off. Three contained correctness bugs on the ODSP historical-load path gate merge. New this round and most serious: the `"odspVersion"` branch of `fetchHistoricalSnapshot` returns the snapshot without hoisting the app tree to the root, so the container loads with no application data and no error — silent data loss (author-conceded). The two siblings: the availability probe checks the live delta feed instead of the selected version's feed, and the availability wrapper collapses retry-exhausted/transient failures into `notAvailable`. All three are localized fixes; the architecture itself remains well-validated.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Answer the open review-thread questions so each ODSP item can be fixed or documented as intended: is the `"odspVersion"` branch intentionally left without `combineProtocolAndAppSnapshotTree` (and if so, where is the app-tree hoist performed before the runtime consumes it); should `areReplayOpsAvailable` probe the selected version's historical delta URL rather than the hard-coded live `deltaStorageUrl`; is collapsing retry-exhausted/transient (and non-ODSP-storage) failures into `notAvailable` intended; and does `getOdspFileVersions` rely on a single service page (no continuation paging) by design with an ordering guarantee that permits an early break.

### Context for Reviewers

- Builds on #16152 — the standalone `loadContainerToSequenceNumber` helper answers vladsud's objection to baking point-in-time semantics into Container/Loader options; `"sequenceNumber"` stays out of the beta `IContainerLoadMode` as internal plumbing, not a reversal of that decision.
- ChumpChief's #16152 off-by-one (`lastProcessedSequenceNumber === loadToSequenceNumber`) is now handled — reviewers should confirm `loadPaused.spec.ts` covers the "target already reached at snapshot" boundary.
- Core historical-load architecture is independently validated: three blind design proposals with no knowledge of the implementation converged on the same shape (name the point by global sequence number, a preflight availability probe returning a typed result, select the closest snapshot ≤ target and bound replay, fail fast on an unreachable target, force read-only, additive alpha-gated surface). Review effort should concentrate on the S1–S3 correctness bugs, not the design shape.
- Historical selection sits on the cache-sensitive `getSnapshot` path; worst case ~50 uncached `getVersions`/`getSnapshotTree` round-trips plus a full op-range walk, with `noCache` on `getVersions` only and a hardcoded 50-version ceiling — **jatgarg** (#19593 `getSnapshot` owner) should confirm cache policy.
- Suggested reviewers: **vladsud** and **ChumpChief** on the `Container`/load-mode design (#16152); **jatgarg** on the ODSP historical-snapshot selection, cost, and cache; **dannimad** on the `serializedStateManager` signature change (#20191); **scottn12** on the original #16152 feature.

- **Needs human judgment — #16152 layering.** The PR bakes point-in-time op-processing into `Container.load` via an internal `"sequenceNumber"` mode and preserves `loadContainerPaused` alongside the new `loadContainerToSequenceNumber` — two coexisting pause/replay paths. Whether the internal mechanism is the right layering, and whether the two paths should be unified, is an owner call — tag **vladsud**, **ChumpChief**, **scottn12**.
- **Needs human judgment — probe placement & interface separation.** `canMaterializePointInTime` reaches through Container's private `storageAdapter` via an `as unknown as` cast, and the alpha surface is modeled as a standalone `IPointInTimeMaterializationStorageService` with structural detection — a driver-definitions owner call. Related: whether to promote the option-extension to a distinct driver method before stabilization.
- **Needs human judgment — header semantics.** Reusing `LoaderHeader.sequenceNumber` ("at least N") as the transport for an "exactly N and stop" target is a design tradeoff: a dedicated header vs. a documented compatibility contract.
- **Needs human judgment — ODSP cost & 50-version bound.** Worst-case ~50 uncached `getVersions`/`getSnapshotTree` round-trips plus a full op-range walk, with `noCache` on `getVersions` only and a hardcoded 50-version ceiling. **jatgarg** should confirm the cost is acceptable and decide the ceiling / cache policy.
- **Cannot be assessed by the pipeline.** Whether ODSP delta storage returns short successful ranges in production (affects `areReplayOpsAvailable`'s strict `messages.length` check); real op-flush/429 transience windows; the probe's real-world cost at large replay distances; cross-fork determinism — all need runtime/service knowledge from ODSP owners (**jatgarg**, **vladsud**).

</details>

- `c0c0e72` 2026-07-13 · **5/10** — `getApiRoot` URL bug on ODC/OneDrive version listing flagged; architecture validated
- `178bc1c` 2026-07-13 · **6/10** — error-fidelity gap on the blocking prefetch flagged; architecture validated
- `b324507` 2026-07-12 · **8/10** — forceReadonly blocker fixed; no correctness or design gate remained
- `e6da50d` 2026-07-12 · **5/10** — forceReadonly fix over-scoped, forcing every default/`"cached"`/`"all"` load read-only
- `36c41c5` 2026-07-10 · **5/10** — core architecture validated; three contained correctness gaps flagged inline
- `a963920` 2026-07-10 · **2/10** — targeted stop-at-target path hangs uncancellably when the target is unreachable
- `40c63cc` 2026-07-10 · **2/10** — generic path never stops at target; core replay/pause path untested

</details>

packages/drivers/odsp-driver/src/odspDocumentStorageManager.ts
```diff
@@ -260,6 +298,242 @@ export class OdspDocumentStorageService extends OdspDocumentStorageServiceBase {
...
+			this.historicalFileVersionId = candidate.fileVersionId;
+			const snapshot = await this.getOdspVersionManager().fetchSnapshotFromVersion(
+				candidate.fileVersionId,
+				true,
+				snapshotFetchOptions.scenarioName,
+			);
+			snapshot.latestSequenceNumber = candidate.sequenceNumber;
+			snapshot.sequenceNumber = candidate.sequenceNumber;
+			this.initializeFromSnapshot(snapshot, false, true);
+			return snapshot;
```
anthony-murphy, line 516: Deep Review: The two branches of `fetchHistoricalSnapshot` are asymmetric, and the `"odspVersion"` branch drops the protocol/app tree combine — a silent data-loss defect (author-conceded).

The `"odspVersion"` branch does `const snapshot = await this.getOdspVersionManager().fetchSnapshotFromVersion(...); ... this.initializeFromSnapshot(snapshot, false, true); return snapshot;` with **no** combine step, while the `"latest"` branch builds `snapshotTree: this.combineProtocolAndAppSnapshotTree(candidate.snapshot.snapshotTree)`. `fetchSnapshotFromVersion` → `fetchSnapshot` → `convertOdspSnapshotToSnapshotTreeAndBlobs` → `buildHierarchy(...)` yields a raw tree whose root contains only `".app"` and `".protocol"` subtrees. `combineProtocolAndAppSnapshotTree` normally hoists the app tree's `blobs`/`trees` to the root. Without it, `getDocumentAttributes` still succeeds (it reads `tree.trees[".protocol"]`), so load does not fail loudly — but the data-store channels remain under `.app` and the runtime sees an empty root. The container loads with no application data and no error.

The existing historical test (`fetchSnapshot.spec.ts:378-390`) asserts only the selected snapshot id/sequence and uses an empty `.app` tree, so it cannot catch the missing hoist.

Recommendation: apply `this.combineProtocolAndAppSnapshotTree(snapshot.snapshotTree)` to the `odspVersion` snapshot before `initializeFromSnapshot`/return, exactly as the `latest` branch does. Add a test that asserts a non-empty `.app` tree is hoisted to the root on the historical path (the current test's empty `.app` tree cannot catch this).

(Question for the author: is the returned snapshot intentionally left in raw `.app`/`.protocol` shape here, and if so, where is the app-tree hoist performed before the runtime consumes it?)

packages/drivers/odsp-driver/src/odspDocumentStorageManager.ts
```diff
@@ -260,6 +298,242 @@ export class OdspDocumentStorageService extends OdspDocumentStorageServiceBase {
...
+	private async areReplayOpsAvailable(
+		baseSnapshotSequenceNumber: number,
+		targetSequenceNumber: number,
+		scenarioName: string | undefined,
+	): Promise<boolean> {
+		if (baseSnapshotSequenceNumber >= targetSequenceNumber) {
+			return true;
+		}
+
+		const service = new OdspDeltaStorageService(
```
anthony-murphy, line 385: Deep Review: The availability probe checks the live delta feed instead of the selected version's feed, so its verdict doesn't match the op source the load will actually use. (Distinct from the double-fetch cost on the [existing thread here](https://github.com/microsoft/FluidFramework/pull/27669#discussion_r3567967816).)

`areReplayOpsAvailable()` always constructs `new OdspDeltaStorageService(this.odspResolvedUrl.endpoints.deltaStorageUrl, ...)` — the **live** delta feed. It is invoked both from the probe (`canMaterializePointInTime`) and the actual load's candidate loop before accepting an `odspVersion` candidate. But this PR also adds `getHistoricalDeltaStorageUrl()` (exposed only after a version is selected), and `OdspDocumentService.connectToDeltaStorage()` uses the selected historical version's `opStream` URL when `historicalFileVersionId` is set. `DEV.md:250-254` states the historical boundary should not mix a historical snapshot from version `V` with ops read from the tip file.

Because the probe uses the live feed, a `materializable` verdict does not guarantee the load's real op source can serve the range, and valid historical targets are falsely rejected once the live feed ages out old ops.

Recommendation: probe the same op source the load will use — for `odspVersion` candidates, probe the candidate/version-bound delta URL (`getDeltaStorageUrlForVersion`/`getHistoricalDeltaStorageUrl`) rather than the hard-coded live `deltaStorageUrl`, falling back to the live feed only when appropriate. Add a test covering a target reachable only via the version-bound stream.

(Question for the author: should `areReplayOpsAvailable()` probe the selected version's `opStream`/historical delta URL rather than the live `deltaStorageUrl`?)

packages/loader/container-loader/src/retriableDocumentStorageService.ts
anthony-murphy: Deep Review: Elevating this beyond the `runWithRetry` polish point above — there's a correctness/diagnosability bug here, so this is Tier 2, not just dead retry wiring.

`RetriableDocumentStorageService.canMaterializePointInTime` wraps `this.runWithRetry(...)` in a broad `catch (error: unknown)` and returns `{ status: "notAvailable", ... }` for **any** exhausted error. The ODSP pre-classification does narrow the blast radius — auth/permission errors become `permissionOrAccessDenied` and retryable errors are rethrown (`odspDocumentStorageManager.ts:353-372`, verified by `fetchSnapshot.spec.ts:496-531`) — but retry-exhausted/transient and non-ODSP-storage failures are still collapsed into `notAvailable`. Hosts then cannot distinguish "point unavailable" from "transient/auth failure." The rest of the storage-wrapper stack (`getSnapshotTree`, `getSnapshot`, `readBlob`, `getVersions`, and the ODSP `RetryErrorsStorageAdapter`) returns `runWithRetry(...)` directly without this swallowing catch. PR #4427's review discussed throwing when retry should not continue rather than hiding failures.

Recommendation: narrow the catch so retry-exhausted/transient and non-ODSP failures propagate (or surface as a distinguishable status); reserve `notAvailable` for the explicit "capability not implemented" case, matching the ODSP `RetryErrorsStorageAdapter`.

(Question for the author: is collapsing retry-exhausted/transient and non-ODSP-storage errors into `notAvailable` intended?)

PR #27673, opened 2026-07-10
fix(client-tree): view consistency for post-processed transactions

jason-ha:
When a transaction is post-processed roll back current changes so that the view gets any changes ensuring consistency.

Add coverage using a simple test post-processor that eliminates all changes.

Additionally optimize inner transaction handling to skip roll back and reapply when post-processor makes no changes.


packages/dds/tree/src/shared-tree-core/transaction.ts
```diff
@@ -464,6 +464,17 @@ export class SquashingTransactionStack<
 								const change =
 									postProcessor === undefined ? squash : postProcessor.processChange(squash);
 
+								if (change !== squash) {
+									// The post-processor produced a change that differs from the
+									// one that was applied to the view as the transaction's edits
+									// were made. Roll back the transaction's changes on the transaction
+									// branch (which rolls back the view) and apply the post-processed
+									// change in their place so that the view fully reflects modified
```
yann-achard-MS:
```suggestion
									// change in their place so that the view fully reflects the modified
```

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
```diff
@@ -995,6 +996,97 @@ describe("sharedTreeView", () => {
...
+			const config = new TreeViewConfiguration({ schema: rootArray, enableSchemaValidation });
+			const view = provider.trees[0].kernel.viewWith(config);
+			view.initialize([]);
+
+			// Act
+			view.runTransaction(() => {
+				view.root.insertAtEnd("A", "B");
+				view.runTransaction(
+					() => {
+						view.root.insertAtStart("C", "D");
```
yann-achard-MS: For the love of all that is holy, please make the scenario consistent with the alphabetic order of the letters!
```suggestion
						view.root.insertAtEnd("C", "D");
```

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
```diff
@@ -995,6 +996,97 @@ describe("sharedTreeView", () => {
...
+				// remove "B" since only the original elements remain as the post
+				// processor erased the other changes.
+				view.root.removeRange(1);
+			});
+
+			// Verify
+			assert.deepEqual(view.root, ["A"]);
+			provider.synchronizeMessages();
+			const otherView = provider.trees[1].kernel.viewWith(config);
+			assert.deepEqual(otherView.root, ["A"]);
```
yann-achard-MS: Nit: I would have left this line out since `validateViewConsistency` takes care of that, but I don't mind either way. Feel free to resolve as is.

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
```diff
@@ -995,6 +996,97 @@ describe("sharedTreeView", () => {
...
 		});
 
+		// A eat-everything post-processor: it erases all changes, demonstrating
+		// a processor that modifies the change set.
+		const eraseChangesPostProcessor = createTransactionPostProcessor({
+			applicability: ChangeProcessorApplicability.Always,
+			processChange: () => ({ changes: [] }),
+		});
+
+		it("transacted view is consistent with inner change processor that makes changes (erases changes) and outer edits", () => {
```
yann-achard-MS: I find the names of these new tests to be confusing: they sound as though the consistency check is between the view and the change processor.
Proposal: have a describe block that talks about consistency between the view on which the transaction is made and the view on a peer. Then have each test title dive into the particulars.

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
jason-ha: Yes. Extra explicit check against literal. Will keep.

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
jason-ha: It is consistent. The consistency is the order in which the letters are introduced.
The insert at start is intended to have effect that current elements are shifted.
And then the remove overlaps both builds to have transaction effect change in more complicated manner.

I'll make it slightly more complicated and add comments.

yann-achard-MS: I see... well, I still prefer the ordering to be that of the elements in the sequence, but I see the sense in using insertion order, so it comes down to personal taste and the tests are simple enough. Resolving.

yann-achard-MS (approved):

jason-ha:
@microsoft-github-policy-service rerun

jason-ha:
@microsoft-github-policy-service rerun

PR #27677, opened 2026-07-13
Improve error messages for token failures

shlevari:
Improves error message for token failures.

### DIFF
--- a/packages/utils/tool-utils/src/odspTokenManager.ts
+++ b/packages/utils/tool-utils/src/odspTokenManager.ts
@@ -12,6 +12,7 @@ import { Mutex } from "async-mutex";
 import { debug } from "./debug.js";
 import type { IAsyncCache, IResources } from "./fluidToolRc.js";
 import { loadRC, lockRC, saveRC } from "./fluidToolRc.js";
+import { redactTokens } from "./redactTokens.js";
 
 // TODO: Add documentation
 // eslint-disable-next-line jsdoc/require-description
@@ -191,7 +192,8 @@ export class OdspTokenManager {
 						tokens = this.ficTokenToIOdspTokens(newTokenData, isPush);
 						await this.updateTokensCacheWithoutLock(cacheKey, tokens);
 					} catch (error) {
-						debug(`${cacheKeyToString(cacheKey)}: Error in refreshing token. ${error}`);
+						const message = (error as Error)?.message ?? "unknown";
+						debug(`${cacheKeyToString(cacheKey)}: Error refreshing token: ${message}`);
 					}
 				} else {
 					tokens = tokensFromCache;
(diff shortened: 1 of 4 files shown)


packages/utils/tool-utils/src/redactTokens.ts
```diff
@@ -0,0 +1,27 @@
...
+const describeSecret = (secret: string | undefined): string =>
+	secret === undefined ? "absent" : `present(len=${secret.length})`;
+
+/**
+ * Produce a log-safe, non-secret description of an {@link IOdspTokens} object.
+ *
+ * Emits only the presence, length, and timing of the token fields — never the token values
+ * themselves — so it is safe to embed in error messages and logs (which flow verbatim to telemetry).
+ */
+export function redactTokens(tokens: Partial<IOdspTokens> | undefined): string {
```
ChumpChief, line 17: At the callsite the typesystem should know `tokens` is defined if you want to drop the undefined handling here, up to you.  Not a blocker.

packages/utils/tool-utils/src/odspTokenManager.ts
```diff
@@ -191,7 +192,8 @@ export class OdspTokenManager {
 						tokens = this.ficTokenToIOdspTokens(newTokenData, isPush);
 						await this.updateTokensCacheWithoutLock(cacheKey, tokens);
 					} catch (error) {
-						debug(`${cacheKeyToString(cacheKey)}: Error in refreshing token. ${error}`);
+						const message = (error as Error)?.message ?? "unknown";
+						debug(`${cacheKeyToString(cacheKey)}: Error refreshing token: ${message}`);
```
ChumpChief: This loses the stack probably, is it necessary?  Is there a more scoped option (e.g. modify offending fields)?  Not a blocker.

ChumpChief (approved):

packages/utils/tool-utils/src/odspTokenManager.ts
shlevari: It's a good point. Also, how does "debug" actually compile/send telemetry? I think this actually might be a spot I was overzealous to clean up and can probably remain the same

shlevari: I reverted this to what it was before. Let me know if you disagree

ChumpChief: Probably fine to revert - I'd expect if anything held offending content in the stringified error it would have been the message anyway

> Also, how does "debug" actually compile/send telemetry?

AFAICT it's just a wrapper on console logging but TBH I'm not super familiar (I also see it has some localStorage integration so not sure if we utilize that and/or put its contents in telemetry).

ChumpChief (approved):

PR #27679, opened 2026-07-13
pnpm dedupe

CraigMacomber:
## Description

`pnpm dedupe` in the root.

## Reviewer Guidance


CraigMacomber:
I have confirmed that we get a large diff (likely undoing most of this) on the next actual lock file change which doesn't dedupe, so probably not a good idea to land this.

PR #27683, opened 2026-07-14
Add ILocalFluidHandleEvents.payloadUploaded event and disposal-safe upload wait helpers

vladsud:
## Description

Local Fluid handles created with a pending payload (for example, blob handles created via `uploadBlob` when `createBlobPayloadPending` is enabled) previously exposed only a `payloadShared` milestone. Sharing a payload is really two steps: (1) uploading the payload to storage, and (2) sending the `BlobAttach` op that makes it available to remote collaborators — and step 2 requires a connection. There was no signal for step 1 on its own, so a local client could not tell when an upload had finished without also waiting to be connected and for the payload to be shared.

This PR adds that missing signal and the ergonomics to consume it safely:

- **New `payloadUploaded` event** on `ILocalFluidHandleEvents` (`@fluidframework/core-interfaces`), emitted once a locally-created pending-payload handle finishes uploading to storage, before the payload is shared to remote collaborators. `BlobManager` now emits an internal `blobUploaded` event when the upload completes (both in the normal upload flow and when loading from pending state with an already-uploaded-but-not-attached blob) and forwards it to the handle as `payloadUploaded`.

- **`waitForPayloadUploaded(handle, abortSignal?)`** (`@fluidframework/runtime-utils`) — resolves once a locally-created pending-payload handle has uploaded its payload (immediately if already uploaded/shared), and rejects if sharing fails or the optional `abortSignal` aborts.

- **`withDisposalAbort(source, operation)`** (`@fluidframework/runtime-utils`) — runs an async operation with an `AbortSignal` that aborts when the given source (e.g. an `IContainer`) is disposed, and removes the disposal listener once the operation settles (on success *and* failure) so it never leaks. Compose it with `waitForPayloadUploaded` so the wait can never hang past container disposal:

  ```ts
  await withDisposalAbort(container, (signal) => waitForPayloadUploaded(handle, signal));
  ```

- **`waitForEvent(listenable, resolveOn, options?)`** (`@fluidframework/runtime-utils`) — a general-purpose helper that resolves the next time one of the given events fires on any `Listenable`, with optional `abortSignal` cancellation and `rejectOn` failure events; it cleans up all subscriptions before settling. `waitForPayloadUploaded` is built on it.

This unlocks the intended workflow: load a container from pending state (paused), upload new blobs, `await` their uploads via `waitForPayloadUploaded` (guarded by `withDisposalAbort`), store the handles in DDSs, then `connect()` — guaranteeing blobs are uploaded before any ops are sent, and that their `BlobAttach` ops are sequenced ahead of the DDS ops that reference them.

### Documentation clarifications

The payload-pending type cluster in `@fluidframework/core-interfaces` (`PayloadState`, `IFluidHandleEvents.payloadShared`, `ILocalFluidHandleEvents.payloadUploaded`/`payloadShareFailed`, and `isFluidHandlePayloadPending`) received substantial doc-only clarifications so an uneducated reader can understand:

- When these events can actually fire, and that they are only meaningful when `createBlobPayloadPending` is enabled (otherwise the caller already receives a handle past the `payloadUploaded`/`payloadShared` points).
- What `get()` does on a not-yet-shared handle: the creating client resolves immediately from the in-memory bytes; another holder *waits* (potentially indefinitely) for the payload to become shared rather than throwing.
- That `isFluidHandlePayloadPending` is a capability check (returns `true` for both `"pending"` and `"shared"` states), not a current-state check, with the narrowing pattern to test the current state.
- The op-ordering guarantee: if an upload starts and finishes while disconnected, the `BlobAttach` op is flushed ahead of any DDS ops produced during that disconnected period, so remote collaborators never observe the handle in a pre-`payloadShared` state.

## Reviewer Guidance

- The new `@fluidframework/runtime-utils` exports (`waitForEvent`, `waitForPayloadUploaded`, `withDisposalAbort`, `WaitForEventOptions`, `IDisposalEventSource`) are `@legacy @beta`; API reports are regenerated accordingly. These are additive, so no type-validation baseline changes were required.
- `waitForPayloadUploaded` observes the *next* upload completion (there is no queryable "uploaded" state), so it must be called before the payload can finish uploading — i.e. immediately after creating the handle and before storing it in a DDS. This late-subscription consideration is documented on the API; feedback welcome on whether a queryable state is worth adding later.
- New tests: `packages/runtime/runtime-utils/src/test/waitForEvent.spec.ts` (19 tests covering `waitForEvent`, `waitForPayloadUploaded`, and `withDisposalAbort`, including a compile-time assertion that a real `IContainer` satisfies `IDisposalEventSource`) and additions to `blobManager.spec.ts` exercising the `payloadUploaded` emission.


ChumpChief:
Checking with @vladsud to understand the scenario better.

packages/runtime/container-runtime/src/blobManager/blobManager.ts
```diff
@@ -703,6 +718,12 @@ export class BlobManager {
 					this.internalEvents.on("blobExpired", onBlobExpired);
 					signal?.addEventListener("abort", onSignalAbort);
 					this.sendBlobAttachMessage(localId, localBlobRecord.storageId);
+					// The blob is now uploaded and its BlobAttach op has been enqueued into the outbox.
+					// We raise the "uploaded" milestone here (rather than when the upload to storage completed)
+					// so that if a listener responds by submitting an op (e.g. storing this handle in a DDS),
+					// that op is enqueued after the BlobAttach op. Note the BlobAttach op is additionally
+					// guaranteed to be flushed ahead of any main-batch op regardless of enqueue order.
+					this.internalEvents.emit("blobUploaded", localId);
```
ChumpChief, line 726: Since this is now positioned to signal "The blobAttach op has been queued" the event should be named accurately to reflect that.

```suggestion
					this.internalEvents.emit("blobAttachStarted", localId);
```

packages/common/core-interfaces/src/handles.ts
```diff
@@ -108,29 +108,83 @@ export interface IFluidHandleInternalPayloadPending<
 }
 
 /**
- * The state of the handle's payload.
- * - "pending" - The payload is not shared to all collaborators
- * - "shared" - The payload is available to both the local client and remote collaborators
+ * The sharing state of a handle's payload.
```
ChumpChief, line 111: The comments in this file got a lot more verbose, to the point that now they're hard to understand.  AFAICT they aren't actually adding any new info, just repeating themselves.  Recommend reverting.

If you want to add a new README at packages/runtime/container-runtime/src/blobManager/README.md could be an appropriate place for internal-facing system documentation, but it doesn't need to be repeated for every public-facing property and event.

packages/runtime/runtime-utils/src/waitForEvent.ts
ChumpChief, line 1: Seems this might live better in client-utils next to createEmitter?  It's not runtime-specific but it is createEmitter-specific.

packages/common/client-utils/src/events

alexvy86:
Local review summary (Deep mode): **Request changes**

The PR is close, but there is one correctness issue that should be fixed before merge.

| Sev | Area | File | Finding | Suggested fix |
|---|---|---|---|---|
| **CRITICAL** | Correctness | `packages/runtime/container-runtime/src/blobManager/blobManager.ts:524-527,726` | `internalEvents.emit("blobUploaded")` synchronously triggers `blobHandle.notifyUploaded()`. If a consumer `payloadUploaded` listener throws, that exception can bubble into `uploadAndAttach`/`tryAttach` after the attach op is already enqueued, causing a false failure path (`payloadShareFailed`) and inconsistent local handle state. | Isolate customer-listener exceptions from core attach flow (guard emit/notify path, report telemetry, continue attach lifecycle), and ensure listener cleanup always runs. |
| **MEDIUM** | API Quality | `packages/runtime/runtime-utils/src/handles.ts:99-103,123-133` | `waitForPayloadUploaded` accepts `IFluidHandle` but returns immediately for non-local handles; docs also imply “necessarily uploaded,” which can mislead for remote pending handles. | Narrow API name/contract to local handles (or document explicit no-op semantics for non-local pending handles without the “necessarily uploaded” claim). |
| **MEDIUM** | API Quality | `packages/common/core-interfaces/src/handles.ts` | New TSDoc is very verbose/repetitive across related interfaces/events, making key contract points harder to find. | Condense repeated narrative and keep only contractual guarantees in API docs; move deep rationale to package-level docs/README. |

### Thoughts on open comments
1. **Event naming (`payloadUploaded` vs `blobAttachStarted`)**: Concern is valid; `blobAttachStarted` is more precise for the emitted milestone, though `payloadUploaded` can work if docs are tightened.
2. **Doc verbosity**: Agree; this is useful feedback.
3. **`waitForEvent` location**: Partially agree. It is generic enough for broader utility, but `runtime-utils` is still defensible; likely non-blocking unless package-layering rules require `client-utils`.

ChumpChief (changes requested):
Marking "request changes" to better communicate status to partners who are getting confused by this PR.  We may or may not add API surface as indicated in my latest round of feedback, but that is TBD.  @alexvy86 may be picking up this work from @vladsud, also TBD.

PR #27685, opened 2026-07-14
Generalize ContainerRuntime.loadRuntime to allow custom Registry

CraigMacomber:
## Description

This reworks the `@internal` static `ContainerRuntime.loadRuntime` to accept a caller-provided `IFluidDataStoreRegistry` instead of raw `registryEntries`, converging it with the existing `ContainerRuntime.loadRuntime2` (which already takes an `IFluidDataStoreRegistry`).

Previously `loadRuntime` constructed a `FluidDataStoreRegistry` from `registryEntries` internally and then delegated to `loadRuntime2`. This change moves that construction up into the public `loadContainerRuntime` wrapper, so the two static entry points now share a single registry model.

What changed:

- `ContainerRuntime.loadRuntime` now takes `Omit<LoadContainerRuntimeParams, "registryEntries" | "runtimeOptions">` plus `registry: IFluidDataStoreRegistry` (and internal `runtimeOptions`), and simply forwards to `loadRuntime2`. The `new FluidDataStoreRegistry(...)` construction moves up into `loadContainerRuntime`.
- The public `loadContainerRuntime` / `loadContainerRuntimeAlpha` entry points are **unchanged** (they still take `registryEntries` and build the registry for you). Only the internal static API changed, so there are no API report changes and no changeset is included.
- Adds a `ContainerRuntime.loadRuntimeAPIVersion` marker so that `@fluidframework/test-utils` — which calls this internal API across multiple package versions and can't special-case on package version — can detect which `loadRuntime` signature a given `ContainerRuntime` exposes.
- Updates the call sites that wrap/override `loadRuntime` to match the new signature: `mixinAttributor` (`@fluid-experimental/attributor`) now wraps the incoming `IFluidDataStoreRegistry` (intercepting `get()` to inject the attributor factory) rather than appending to `registryEntries`; `createTestContainerRuntimeFactory` (`@fluidframework/test-utils`) and `replayFluidFactories` (`@fluidframework/replay-tool`) are updated as well, along with the affected unit and layer-compat tests.

### Motivation

This is a self-contained cleanup factored out of #27078 ("Introduce Unified API for Fluid Client"). 

What this change buys is convergence: it moves the older `loadRuntime` — and everything built on it (`mixinAttributor`, `test-utils`, `replay-tool`) — onto the same `IFluidDataStoreRegistry` model that `loadRuntime2` and the new `ServiceClient` already use. Without it, #27078 would leave behind two parallel `loadRuntime` variants with different registry models. As the code comment notes, once these callers are aligned the `registryEntries`-based `loadRuntime` can eventually be removed and `loadRuntime2` renamed back to `loadRuntime`.

## Reviewer Guidance

- This change only affects the `@internal` static `ContainerRuntime.loadRuntime`; the public `loadContainerRuntime` / `loadContainerRuntimeAlpha` signatures and all API reports are unchanged, so no changeset is included — please confirm you agree this needs no changeset.
- The `loadRuntimeAPIVersion` marker plus the `@fluidframework/test-utils` update handle the cross-version compatibility this internal, layer-compat-relevant API requires; feedback on that approach is welcome.


packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -837,6 +840,10 @@ export async function loadContainerRuntime(
  * @param params - An object which specifies all required and optional params necessary to instantiate a runtime.
  * @returns An object containing the runtime.
  *
+ * @privateRemarks
+ * By using loadRuntime2 instead of loadRuntime, this prevents mixinAttributor's overriding of loadRuntime from affecting this new API:
```
CraigMacomber, line 850: This is not a change in implementation, but rather documenting a potentially confusing consequence of how this was implemented. I'm unclear if this was desired/intended, but I figured making it more clear would be good.

packages/test/test-utils/src/testContainerRuntimeFactory.ts
```diff
@@ -62,6 +65,26 @@ interface backCompat_ContainerRuntime {
...
+		registryEntries: NamedFluidDataStoreRegistryEntries;
+		existing: boolean;
+		requestHandler?: (request: IRequest, runtime: IContainerRuntime) => Promise<IResponse>;
+		provideEntryPoint: (containerRuntime: IContainerRuntime) => Promise<FluidObject>;
+		runtimeOptions?: IContainerRuntimeOptionsInternal;
+		containerScope?: FluidObject;
+		// Some versions of this API exist which expect this to be provided under the name `minVersionForCollab` (before https://github.com/microsoft/FluidFramework/pull/27806),
+		// so this fails to actually set minVersionForCollab in some old cases, but it was optional then so does not error.
+		// That predates the versioning of this API, so we mitigate by setting both.
+		oldestSupportedClient: OldestSupportedClientVersion | undefined;
```
CraigMacomber, line 83: This fixes a regression from https://github.com/microsoft/FluidFramework/pull/27806/files

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -934,21 +943,21 @@ export class ContainerRuntime
 	 * `loadRuntime` could be removed (replaced by `loadRuntime2` which could be renamed back to `loadRuntime`).
 	 */
 	public static async loadRuntime(
-		params: LoadContainerRuntimeParams & {
-			/**
-			 * Constructor to use to create the ContainerRuntime instance.
-			 * @remarks
-			 * Defaults to {@link ContainerRuntime}.
-			 */
+		params: Omit<LoadContainerRuntimeParams, "registryEntries" | "runtimeOptions"> & {
```
jason-ha, line 946: Instead of omitting "runtimeOptions", how about leaving it and letting intersection upgrade it?
Then if there is conflict with "base" runtimeOptions it should be more apparent. 
If it weren't such a hassle to maintain, I'd say there should be a base common type for these two interfaces.

packages/framework/attributor/src/mixinAttributor.ts
```diff
@@ -63,10 +63,18 @@ export const mixinAttributor = (
...
 			const factory = new RuntimeAttributorFactory();
-			const registryEntriesCopy: NamedFluidDataStoreRegistryEntries = [
-				...registryEntries,
-				[RuntimeAttributorFactory.type, Promise.resolve(factory)],
-			];
+			// Wrap the registry to also include the attributor factory
+			const registryWithAttributor: IFluidDataStoreRegistry = {
+				get IFluidDataStoreRegistry(): IFluidDataStoreRegistry {
+					return registryWithAttributor;
+				},
```
jason-ha, line 70: Why does this need to be a getter?
Because you want no chance of writing it?
(getters that can be simple props trigger me :) )

packages/test/test-utils/src/testContainerRuntimeFactory.ts
```diff
@@ -171,20 +194,47 @@ export const createTestContainerRuntimeFactory = (
...
-				runtimeOptions: this.runtimeOptions,
-				containerScope: context.scope,
-				existing,
-				oldestSupportedClient: this.minVersionForCollab,
-			});
+			const registryEntries: NamedFluidDataStoreRegistryEntries = [
+				["default", Promise.resolve(this.dataStoreFactory)],
+				[this.type, Promise.resolve(this.dataStoreFactory)],
+			];
+			// eslint-disable-next-line import-x/no-deprecated
```
jason-ha, line 201: disables should be commented

jason-ha (commented):
Looks pretty good to me

packages/test/test-utils/src/testContainerRuntimeFactory.ts
jason-ha: + @WillieHabi ^^
If we don't have this in this minor (today's release), what is the fallout for 2.x maintenance?

packages/framework/attributor/src/mixinAttributor.ts
CraigMacomber: Its returning itself: make that a property and its used before initilized.

packages/test/test-utils/src/testContainerRuntimeFactory.ts
jason-ha: From Teams thread it seems like we'd be fine since testing with code that is flexible.
This will need addressed when making further updates.
(I actually would like the code in this area to reference the previous package types and mix together in some rationale, traceable way. So instead of adding this we'd import prior type info explicitly.)

packages/runtime/container-runtime/src/containerRuntime.ts
CraigMacomber: This matches the existing loadRuntime2, which it is wrapping.

If I remove the omit and keep the `&` it also generates a different type, giving ` Property 'registryEntries' is missing in type...` errors at the call sites since we don't want to require that property, and losing the omit makes it required.

packages/test/test-utils/src/testContainerRuntimeFactory.ts
CraigMacomber: This was preexisting (line 180) Diff is a little messy, it was there before my change.

Deprecation is:
`@deprecated — Will be removed once Loader LTS version is "2.0.0-internal.7.0.0". Migrate all usage of IFluidRouter to the "entryPoint" pattern. Refer to Removing-IFluidRouter.md`

SO maybe we need to keep using this in our back compat testing until we no longer support 1.x LTS? But as its preexisting and I'm not confident, I'd prefer to leave it as is.

packages/runtime/container-runtime/src/containerRuntime.ts
jason-ha: I mean to keep omit for "registryEntries" and just drop the "runtimeOptions" removal. But it was like that so, okay!

jason-ha (approved):

PR #27686, opened 2026-07-14
feat(tree): Alpha ArrayNode APIs 

brrichards:
## Description

Adds/exposes new alpha methods for ArrayNode. Some of these api's are aliases for already implemented methods, but might be benficial for LLMs that are familiar with the common array method names.

Added alpha methods:
- `at(index)` - Already implementted at runtime, but now officially declared and documented in the API
- `unshift(...items)` - Alias for `insertAtStart`. Unlike `Array.prototype.unshift`, it does not return the new length 
- `pop()` - Removes and returns the last item in the array or returns `undefined` if it is empty
- `shift()` - removes and returns the first item in the array, or returns `undefined` if it is empty

## Testing
Moved helper method `buildAlphaArray` one level up to be used in multiple new tests covering all 4 API's

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
+++ b/packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
@@ -131,6 +131,8 @@ export interface TreeArrayNode<
 	 * Inserts new item(s) at the end of the array.
 	 *
 	 * @remarks
+	 * Unlike `Array.prototype.push`, this method does not return the new length of the array.
+	 *
 	 * The order of the inserted items relative to other concurrently inserted items at the same location is only partially specified:
 	 * Concurrently inserting `[A, B]` and `[X, Y]` at the same location may yield
 	 * either `[A, B, X, Y]` or `[X, Y, A, B]`, regardless of the order in which those edits are sequenced.
@@ -460,6 +462,41 @@ export interface TreeArrayNodeAlpha<
 		? InsertableTreeNodeFromImplicitAllowedTypes<TAllowedTypes>
 		: InsertableTreeNodeFromImplicitAllowedTypes<ImplicitAllowedTypes>,
 > extends TreeArrayNode<TAllowedTypes, T, TNew> {
+	/**
+	 * Returns the item located at the specified index.
+	 * @param index - The zero-based index of the item to retrieve.
+	 * Negative indices count back from the last item in the array: for `index < 0`, `index + array.length` is used.
+	 * @returns The item at the specified index, or `undefined` if the index is out of bounds.
+	 */
+	at(index: number): T | undefined;
+
+	/**
+	 * Removes the last item from the array and returns it.
+	 * @returns The removed item, or `undefined` if the array is empty (in which case the array is not modified).
+	 */
+	pop(): T | undefined;
+
+	/**
+	 * Removes the first item from the array and returns it.
+	 * @returns The removed item, or `undefined` if the array is empty (in which case the array is not modified).
+	 */
+	shift(): T | undefined;
+
+	/**
+	 * Inserts new item(s) at the start of the array.
+	 *
+	 * @param value - The content to insert.
+	 * @remarks
+	 * Unlike `Array.prototype.unshift`, this method does not return the new length of the array.
+	 *
+	 * The order of the inserted items relative to other concurrently inserted items at the same location is only partially specified:
+	 * Concurrently inserting `[A, B]` and `[X, Y]` at the same location may yield
+	 * either `[A, B, X, Y]` or `[X, Y, A, B]`, regardless of the order in which those edits are sequenced.
+	 * No other interleavings are possible. (e.g. `[A, X, B, Y]` is not possible.)
+	 *
+	 */
+	unshift(...value: readonly (TNew | IterableTreeArrayContent<TNew>)[]): void;
+
 	/**
 	 * Removes existing item(s) and/or adds new item(s).
 	 * @param start - The index at which to start changing the array. If negative, it is treated as `array.length + start`.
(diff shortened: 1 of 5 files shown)


packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -460,6 +462,41 @@ export interface TreeArrayNodeAlpha<
...
+	 * @param value - The content to insert.
+	 * @remarks
+	 * Unlike `Array.prototype.unshift`, this method does not return the new length of the array.
+	 *
+	 * The order of the inserted items relative to other concurrently inserted items at the same location is only partially specified:
+	 * Concurrently inserting `[A, B]` and `[X, Y]` at the same location may yield
+	 * either `[A, B, X, Y]` or `[X, Y, A, B]`, regardless of the order in which those edits are sequenced.
+	 * No other interleavings are possible. (e.g. `[A, X, B, Y]` is not possible.)
+	 *
+	 */
```
Josmithr, line 506: Nit: this documentation should presumably be the same as the docs on `insertAtStart`. For consistency and to avoid duplication, I would recommend moving these docs to `insertAtStart`, then update the docs here to be `{@inheritDoc TreeArrayNode.insertAtStart}`.

Josmithr: We could do the same for `push` and `insertAtEnd`

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -161,6 +161,13 @@ describe("ArrayNode", () => {
 		schemaType: typeof PojoEmulationNumberArray | typeof CustomizableNumberArray,
 	): void {
 		describeHydration(title, (init) => {
+			function buildAlphaArray(
+				initial: number[],
```
Josmithr: Nit: readonly array?

Josmithr (approved):
Left a couple nitpicks, but otherwise this looks great to me! Thanks!

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
brrichards: Good call. Push is where this comment is from (since it's a single insertAt call like the others methods you mentioned). I'll go ahead and deduplicate and link to insertAt

CraigMacomber (approved):

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
brrichards: changed to readonly

Josmithr (approved):

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -460,6 +474,38 @@ export interface TreeArrayNodeAlpha<
...
+	/**
+	 * Returns the item located at the specified index.
+	 * @param index - The zero-based index of the item to retrieve.
+	 * Negative indices count back from the last item in the array: for `index < 0`, `index + array.length` is used.
+	 * @returns The item at the specified index, or `undefined` if the index is out of bounds.
+	 */
+	at(index: number): T | undefined;
+
+	/**
+	 * Removes the last item from the array and returns it.
```
yann-achard-MS, line 486: It's not clear from the docs which of the three possible merge semantics this has:
1. removes the item that is last at the time the edit is authored
2. removes the item that is last at the time the edit is sequenced
3. purposefully unspecified (meaning either 1 or 2 could happen)

The current implementation does option 1. The question is do we want to make it possible to change that in the future? If so, we would document this as having the merge semantics of option 3.

I think the answer is "no" because pop returns the removed item as a `T` which is good but makes it impossible to have an indirection for something that would have "this will be your removed item once things settle" semantics. It's also unlikely anyone would want to use that or do so correctly.

So it's fine to document this as having the semantics of option 1. We can say something like "Note that once this edit is sequenced, the node being detached by this operation may no longer be last in the array and may in fact not be inside this array at all. This operation will remove that node from wherever it then resides."

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -460,6 +474,38 @@ export interface TreeArrayNodeAlpha<
...
+	 * @returns The removed item, or `undefined` if the array is empty (in which case the array is not modified).
+	 */
+	pop(): T | undefined;
+
+	/**
+	 * Removes the first item from the array and returns it.
```
yann-achard-MS, line 492: Same idea as for `pop`.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -460,6 +474,38 @@ export interface TreeArrayNodeAlpha<
...
+	/**
+	 * Inserts new item(s) at the start of the array.
+	 *
+	 * @param value - The content to insert.
+	 * @remarks
+	 * Unlike `Array.prototype.unshift`, this method does not return the new length of the array.
+	 *
+	 * Equivalent to `insertAt(0, ...value)`:
+	 * see {@link (TreeArrayNode:interface).insertAt} for details, including merge semantics with concurrent edits.
+	 */
```
yann-achard-MS, line 506: Actually for `unshift` I think it makes sense to narrow the contract to what actually happens in practice because it is what people will expect.

Maybe something like this:
```suggestion
	/**
	 * Inserts new item(s) at the start of the array.
	 *
	 * @param value - The content to insert.
	 * @remarks
	 * Unlike `Array.prototype.unshift`, this method does not return the new length of the array.
	 *
	 * All items inserted by a single call to this method are inserted consecutively.
	 * The order of the inserted items relative to other concurrently inserted items is fully specified:
	 * Concurrently unshifting `[A, B]` and `[X, Y]` will yield `[A, B, X, Y]` if the edit unshifting `[A, B]` is sequenced after the edit unshifting `[X, Y]`.
	 */
```

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
yann-achard-MS, line 515: Would be good to link to the docs for `insertAt` here so the merge semantics are called out.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
yann-achard-MS: Actually, it would be better to put such documentation on `removeAt` or `removeRange` and link to that from here and from `shift`.

PR #27687, opened 2026-07-14
tree: Edit filtering

alex-pardes:
## Description

This adds functionality to field change handlers for filtering out specific edits from changesets. This will be used in future work on transaction minimization. The `mute` method on `FieldChangeHandler` has been reimplemented by using `filterEdits` to remove all edits.


packages/dds/tree/src/feature-libraries/modular-schema/index.ts
```diff
@@ -44,6 +44,7 @@ export {
 	NodeAttachState,
 	type FieldChangeEncodingContext,
 	type NestedChangesIndices,
+	EditFilterStatus,
```
jason-ha: nit: these were all sorted until the last three additions
slot this new one alphabetically
preferred: also sort the last three. (tooling will add things in the right place if things are already sorted in some fashion when it gets there)

packages/dds/tree/src/feature-libraries/optional-field/optionalField.ts
```diff
@@ -814,3 +814,49 @@ export const optional: Optional = new FlexFieldKind(
...
+			}
+		} else if (!preserveOtherEdits) {
+			delete filtered.valueReplace;
+		}
+	}
+
+	if (!preserveOtherEdits) {
+		filtered.moves = [];
+	}
+	return filtered;
```
jason-ha, line 855: Any value in returning the original if there were no modifications?

packages/dds/tree/src/feature-libraries/optional-field/optionalField.ts
```diff
@@ -814,3 +814,49 @@ export const optional: Optional = new FlexFieldKind(
 		]),
 	},
 );
+
+function filterEdits(
```
jason-ha, line 818: How about a few words?

packages/dds/tree/src/feature-libraries/modular-schema/fieldChangeHandler.ts
```diff
@@ -167,10 +168,37 @@ export interface FieldChangeRebaser<TChangeset> {
 	replaceRevisions(change: TChangeset, replacer: RevisionReplacer): TChangeset;
 
 	/**
-	 * Returns a copy of the given changeset with the same declarations (e.g., new cells) but no actual changes.
-	 * This is a kludge. TODO: remove once AB#46104 is completed.
+	 * Returns a copy of the given changeset with edits removed as specified.
+	 * @param change - The change to filter edits from
+	 * @param filterDetach - This should be called for each range of detaches in the changeset,
+	 * and the attach should be preserved, removed, or converted to a non-move detach as specified.
```
yann-achard-MS:
```suggestion
	 * and the detach should be preserved, removed, or converted to a non-move detach as specified.
```

packages/dds/tree/src/feature-libraries/modular-schema/fieldChangeHandler.ts
```diff
@@ -167,10 +168,37 @@ export interface FieldChangeRebaser<TChangeset> {
...
+	 * should be preserved or removed.
 	 */
-	mute(change: TChangeset): TChangeset;
+	filterEdits(
+		change: TChangeset,
+		filterDetach: (
+			id: ChangeAtomId,
+			count: number,
+			endpoint?: ChangeAtomId,
+		) => RangeQueryResult<EditFilterStatus>,
```
yann-achard-MS: Might be worth defining the delegate as a type separately so that you can reuse it (and document it).

packages/dds/tree/src/feature-libraries/sequence-field/markQueue.ts
```diff
@@ -36,11 +33,14 @@ export class MarkQueue {
...
-		for (let i = splitMarks.length - 1; i > 0; i--) {
-			this.stack.push(splitMarks[i] ?? oob());
+		const splitLength = this.getLengthForSplit(mark);
+		if (splitLength < mark.count) {
+			const [part1, part2] = splitMark(mark, splitLength);
+			this.stack.push(part2);
+			return part1;
+		} else {
+			return mark;
 		}
```
jason-ha: nit: drop the `else {` and leave return at higher scope to make it slightly easier to be confident there is always a return from all cases.

packages/dds/tree/src/feature-libraries/sequence-field/sequenceFieldChangeRebaser.ts
```diff
@@ -21,7 +21,5 @@ export const sequenceFieldChangeRebaser = {
 	rebase,
 	prune,
 	replaceRevisions,
-	mute: (change: Changeset): Changeset => {
-		return change.map((mark) => omitMarkEffect(mark));
-	},
+	filterEdits,
 };
```
jason-ha: pre-existing: this should probably be `as const`. Not supposed to be mutable, right?

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
```diff
@@ -0,0 +1,180 @@
...
+	return { length: count, value: EditFilterStatus.Preserve };
+}
+
+function removeAll(
+	id: ChangeAtomId,
+	count: number,
+	endpoint?: ChangeAtomId,
+): RangeQueryResult<EditFilterStatus> {
+	return { length: count, value: EditFilterStatus.Remove };
+}
```
jason-ha: Is nice for reader to have these before use.

jason-ha (commented):
I've gotten partially through - yet to review filterEdits.ts and new tests

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
```diff
@@ -0,0 +1,180 @@
...
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+// eslint-disable-next-line import-x/no-internal-modules
+import { filterEdits } from "../../../feature-libraries/sequence-field/filterEdits.js";
+// eslint-disable-next-line import-x/no-internal-modules
+import { EditFilterStatus } from "../../../feature-libraries/modular-schema/index.js";
+import {
+	areEqualChangeAtomIdOpts,
```
yann-achard-MS: lint: areEqualChangeAtomIdOpts not used

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
```diff
@@ -0,0 +1,180 @@
...
+const tag1 = mintRevisionTag();
+const tag2 = mintRevisionTag();
+const tag3 = mintRevisionTag();
+const id1: ChangeAtomId = { revision: tag1, localId: brand(0) };
+const id2: ChangeAtomId = { revision: tag1, localId: brand(1) };
+const id3: ChangeAtomId = { revision: tag2, localId: brand(0) };
+const id4: ChangeAtomId = { revision: tag2, localId: brand(1) };
+const id5: ChangeAtomId = { revision: tag3, localId: brand(0) };
+const id6: ChangeAtomId = { revision: tag3, localId: brand(1) };
+const id7: ChangeAtomId = { revision: tag3, localId: brand(1) };
```
yann-achard-MS: id7 is structurally equal to id6, was that intended?

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
```diff
@@ -0,0 +1,180 @@
...
+				false,
+			);
+
+			assertChangesetsEqual(filtered, [
+				MarkMaker.tomb(id1.revision, id1.localId, 1),
+				MarkMaker.tomb(tag1, brand(5), 2),
+			]);
+		});
+
+		it("Can remove convert move-out to remove", () => {
```
yann-achard-MS:
```suggestion
		it("Can convert move-out to remove", () => {
```

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
```diff
@@ -0,0 +1,180 @@
...
+			const moveInCell = offsetChangeAtomId(id2, 1);
+			const expected = [
+				MarkMaker.remove(1, id1),
+				MarkMaker.tomb(moveInCell.revision, moveInCell.localId),
+			];
+
+			assertChangesetsEqual(filtered, expected);
+		});
+
+		it("Can remove convert move-in to insert", () => {
```
yann-achard-MS:
```suggestion
		it("Can convert move-in to insert", () => {
```

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
alex-pardes: I generally have the opposite preference, but I don't mind making this change.

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
jason-ha, line 26: FieldChangeDelta and NestedChangeIndices remain type imports. I am surprised linting doesn't complain.
We haven't rolled it out, but the preferred import pattern here would be:
```suggestion
import type {
	FieldChangeDelta,
	NestedChangesIndices,
// eslint-disable-next-line import-x/no-internal-modules
} from "../../../feature-libraries/modular-schema/fieldChangeHandler.js";
// eslint-disable-next-line import-x/no-internal-modules
import { EditFilterStatus } from "../../../feature-libraries/modular-schema/fieldChangeHandler.js";
```
(I am not sure why (given tree's structure) `import-x/no-internal-modules` hasn't been altered for tree tests.)

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
```diff
@@ -1000,4 +1006,109 @@ describe("optionalField", () => {
...
+				Change.clear("self", id2),
+				Change.move(id3, "self"),
+			);
+
+			const filtered = optionalChangeRebaser.filterEdits(
+				change,
+				preserveAll,
+				preserveAll,
+				true,
+			);
```
jason-ha: From the call site it is very hard to see what is what argument-wise.
I am a strong advocate for named parameters when there might be any confusion. For this it would mean the signature could be either
```typescript
	filterEdits(
		change: TChangeset,
		options: {
			filterAttach: EditFilterFunc;
			filterDetach: EditFilterFunc;
			preserveOtherEdits: boolean;
		},
	): TChangeset;
```
or
```typescript
	filterEdits(
		args: {
			change: TChangeset,
			filterAttach: EditFilterFunc;
			filterDetach: EditFilterFunc;
			preserveOtherEdits: boolean;
		},
	): TChangeset;
```
then the call sites are crystal clear.

Without structured arguments it is best to comment the arguments that might be confused. So, we'd want
```typescript
			const filtered = optionalChangeRebaser.filterEdits(
				change,
				/* filterDetach */ preserveAll,
				/* filterAttach */ preserveAll,
				/* preserveOtherEdits */ true,
			);
```

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
```diff
@@ -1000,4 +1006,109 @@ describe("optionalField", () => {
...
+			const change = Change.atOnce(changeWithoutMove, Change.move(id0, id1));
+
+			const filtered = optionalChangeRebaser.filterEdits(
+				change,
+				preserveAll,
+				preserveAll,
+				false,
+			);
+
+			assertEqual(filtered, changeWithoutMove);
```
jason-ha: This is confusing for uninformed reader.
`changeWithoutMove` certainly has a change named `move` in it (`Change.move(id3, "self")`). But I guess if `"self"` is involved then it isn't actually a `move`?
Assuming there is no other expression to get the "write" that the `Change.move` actually is, then would be good to comment.
From what I think I've read that `Change.clear` could be a move, but it isn't per `"self"` target.

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
```diff
@@ -1000,4 +1006,109 @@ describe("optionalField", () => {
...
+			endpoint?: ChangeAtomId,
+		): RangeQueryResult<EditFilterStatus> {
+			return { length: count, value: EditFilterStatus.Remove };
+		}
+
+		it("can preserve all", () => {
+			const id0: ChangeAtomId = { revision: tag, localId: brand(0) };
+			const id1: ChangeAtomId = { revision: tag, localId: brand(1) };
+			const id2: ChangeAtomId = { revision: tag, localId: brand(2) };
+			const id3: ChangeAtomId = { revision: tag, localId: brand(3) };
```
jason-ha: These need to be mutable and can't be common constants like `nodeId1` and others? No change required. Mostly want to know.

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
```diff
@@ -1000,4 +1006,109 @@ describe("optionalField", () => {
...
+
+		it("can remove edits", () => {
+			const id0: ChangeAtomId = { revision: tag, localId: brand(0) };
+			const id1: ChangeAtomId = { revision: tag, localId: brand(1) };
+			const id2: ChangeAtomId = { revision: tag, localId: brand(2) };
+			const id3: ChangeAtomId = { revision: tag, localId: brand(3) };
+
+			const changeWithoutReplace = Change.atOnce(
+				Change.childAt(id0, nodeId1),
+				Change.move(id0, id1),
```
jason-ha: For Jason's knowledge: Is there a glossary for terms like "edits"? A "move" isn't an "edit"?

jason-ha (commented):
Thoughts from reviewing `optionalField.spec.ts`.

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
jason-ha: I'd love to know more. Because you don't want details?
I also generally don't want details, but I want to know about a thing. FWIW a compiler does too. Here it has to parse everything to know that those helpers exist and come back. For variables there is a rule that they exist before use. All up, applying ordering to functions / globals seems good.

That said, I think ordering is different when it comes to implementing an interface. There publics are usually good to have first.

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
```diff
@@ -1000,4 +1006,109 @@ describe("optionalField", () => {
 			assert.deepEqual(actual, expected);
 		});
 	});
+
+	describe("Filter edits", () => {
```
jason-ha, line 1015: I don't see any `PreserveWithoutMove` in here. Does it never apply to these?

packages/dds/tree/src/feature-libraries/sequence-field/filterEdits.ts
```diff
@@ -0,0 +1,148 @@
...
+				case EditFilterStatus.Preserve: {
+					filtered = mark;
+					break;
+				}
+				case EditFilterStatus.PreserveWithoutMove: {
+					// KLUDGE: PreserveWithoutMove requires us to return a mark which attaches the node
+					// using the detach ID of the move (the endpoint ID).
+					// Insert marks use the cell ID as the attach ID, so we must change the cell ID here.
+					// This could be a problem if there were references to the old cell ID outside this changeset,
+					// but PreserveWithoutMove is only used for transaction squashing, where that is not a problem.
```
jason-ha: squashing -> minimizing
(squashing term is already used in all transactions)

packages/dds/tree/src/feature-libraries/sequence-field/filterEdits.ts
```diff
@@ -0,0 +1,148 @@
...
+					const newCellId = endpoint ?? { revision: mark.revision, localId: mark.id };
+					filtered = {
+						type: "Insert",
+						count: mark.count,
```
jason-ha: If I understand correctly, this count is ignored. If it can't be omitted, then use `result.length` that is the final value, right?

packages/dds/tree/src/feature-libraries/sequence-field/filterEdits.ts
```diff
@@ -0,0 +1,148 @@
...
+						id: mark.id,
+					};
+
+					if (mark.changes !== undefined) {
+						filtered.changes = mark.changes;
+					}
+					break;
+				}
+				case EditFilterStatus.Remove: {
+					filtered = omitMarkEffect(mark);
```
jason-ha, line 87: `omitMarkEffect` uses `NoopMark` in its typing, but I noticed later in `MarkListFactory.pushContent` there is check for `isNoopMark` and this result isn't that. The `mark.type` is `undefined`.
This is the right result from filter? `mark.type` being `undefined` means...?

jason-ha: I figured this out -- `NoopMarkType` is stealthily `undefined`.
Though this did lead me to see some ~funny code in `MarkListFactory.pushContent`. It has
`if (isNoopMark(mark) && mark.changes === undefined && !isTombstone(mark)) {`
and the definition of `isTombstone` is
`mark.type === NoopMarkType && mark.cellId !== undefined && mark.changes === undefined`
.
So `!isTombstone` use has some redundancy. Whole if could be `isNoopMark(mark) && mark.changes === undefined && mark.cellId === undefined` aka isNonTombstoneNoopMark.

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
```diff
@@ -0,0 +1,179 @@
...
+				MarkMaker.insert(1, offsetChangeAtomId(id1, 2)),
+				MarkMaker.remove(1, id3),
+				MarkMaker.skip(1),
+				MarkMaker.remove(1, offsetChangeAtomId(id3, 2)),
+			];
+
+			assertChangesetsEqual(filtered, expected);
+		});
+
+		it("Preserves tombstones", () => {
```
jason-ha, line 126: This is just slightly more than preserving tombstones. It also will replace a filtered out `Insert` with a tombstone.

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
```diff
@@ -0,0 +1,179 @@
...
+function removeAll(
+	id: ChangeAtomId,
+	count: number,
+	endpoint?: ChangeAtomId,
+): RangeQueryResult<EditFilterStatus> {
+	return { length: count, value: EditFilterStatus.Remove };
+}
+
+export function testFilterEdits(): void {
+	describe("Filter edits", () => {
```
jason-ha, line 48: It is probably okay, but there aren't any tests that remove more than one count but less than all. (Nothing like filter out 2 of 3.)

packages/dds/tree/src/feature-libraries/modular-schema/fieldChangeHandler.ts
```diff
@@ -167,10 +168,35 @@ export interface FieldChangeRebaser<TChangeset> {
...
+export type EditFilterFunc = (
+	id: ChangeAtomId,
+	count: number,
+	endpoint?: ChangeAtomId,
+) => RangeQueryResult<EditFilterStatus>;
+
+export enum EditFilterStatus {
+	Remove,
+	Preserve,
+	PreserveWithoutMove,
```
jason-ha, line 217: Docs requested (PreserveWithoutMove really deserves explanation)

packages/dds/tree/src/feature-libraries/sequence-field/filterEdits.ts
```diff
@@ -0,0 +1,148 @@
...
+						type: "Insert",
+						count: mark.count,
+						cellId: newCellId,
+						revision: mark.revision,
+						id: mark.id,
+					};
+
+					if (mark.changes !== undefined) {
+						filtered.changes = mark.changes;
+					}
```
jason-ha: There don't appear to be any tests that hit this case or the similar Remove/MoveOut.

jason-ha (commented):
Looks good - some comments and maybe additional test cases would make it even better

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
alex-pardes: I've renamed the test to use the term "replace" instead of "edit."

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
alex-pardes: I've made them common constants.

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
alex-pardes: Agreed that the terminology is unclear. I've updated this to use the term "rename" instead of "move," for the case where the move only changes the ID of a detached node. In any case, this will all be simplified once my attach/detach refactor goes in.

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
alex-pardes: I'm not sure I understand the question. Are you asking if there should be a test which preserves only non-move edits?

packages/dds/tree/src/feature-libraries/optional-field/optionalField.ts
alex-pardes: Doesn't seem worth special casing to me.

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
alex-pardes: Yes. I don't want to see details first, and for me that includes declarations like this. Are you suggesting that this pattern is bad due to the TS compiler having to extra work, or just pointing out how compiling works?

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
jason-ha: There are no test filters here that return `PreserveWithoutMove`.
The implementation (`filterEdits` of optionalFields.ts) also doesn't look for that result. Though implementation also doesn't look for `Preserve` either - only checks against `Remove`.
My assumption is that `PreserveWithoutMove` is meant to be treated the same as `Preserve` and "WithoutMove" is just irrelevant in this context.
But if the implementation were changed to check for `!== Preserve` but filters might still produce `PreserveWithoutMove` and tests were kept as they are, then all of the tests would pass. Yet as soon as there is a filter returning `PreserveWithoutMove` we'd have a bug.

packages/dds/tree/src/feature-libraries/optional-field/optionalField.ts
```diff
@@ -814,3 +814,41 @@ export const optional: Optional = new FlexFieldKind(
...
+			if (detachResult === EditFilterStatus.Remove) {
+				assert(
+					attachId === undefined || attachResult === EditFilterStatus.Remove,
+					"Cannot remove detach without also removing attach",
+				);
+
+				delete filtered.valueReplace;
+			} else if (attachResult === EditFilterStatus.Remove) {
+				delete filtered.valueReplace.src;
+			}
```
jason-ha, line 846: FYI
I haven't soused out the root of why AI wants to make a change here. I merged this branch with my builds minimization and asked it to hook up edit minification.
It wanted to change these lines to:
```suggestion
			const next: Mutable<Replace> = { ...filtered.valueReplace };
			if (detachResult === EditFilterStatus.Remove) {
				// The input content being detached is gone (transient, or trimmed from its build),
				// so the field is effectively empty at input and detaches nothing. Any child changes
				// to that (now absent) input content are likewise dropped.
				next.isEmpty = true;
				filtered.childChanges = filtered.childChanges.filter(([id]) => id !== "self");
			}
			if (attachResult === EditFilterStatus.Remove) {
				// The content being attached is transient, so nothing is attached.
				delete next.src;
			}
			// Drop the replace entirely once it no longer has any observable effect, so no extraneous
			// effect (or detached-node reference) remains.
			if (isReplaceEffectful(next)) {
				filtered.valueReplace = next;
			} else {
				delete filtered.valueReplace;
			}
```
Here is some context on the change. (I had told it to try to limit changes to minimizeModularChange.ts.):
>  optionalField.ts  (necessary supporting change): reworked its  filterEdits  so a removed detach makes the replace  isEmpty  (keeping any attach) and drops the now-absent input content's  "self"  child changes, and the whole  valueReplace  is dropped once non-effectful. This is what makes the "insert-then-replace" / nested-object cases correct under the filtering technique.
>
> Note on scope: The task suggested "likely only  minimizeModularChange.ts " would need change. The optional-field  filterEdits  change was required for the filtering technique to correctly handle the "trimmed input detach" case (transient content built inline inside a surviving node). It's a small, well-contained fix and keeps all 4 existing  filterEdits  unit tests passing. Let me know if you'd prefer a different split.

I don't know if it should be crafting different filter or is correct about a change like this and unit test should have more cases.
NO CHANGES NEEDED now.

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
alex-pardes: Updated

packages/dds/tree/src/feature-libraries/optional-field/optionalField.ts
alex-pardes: As far as I can understand what it's saying, the AI seems confused about what edit filtering is supposed to be doing.

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
alex-pardes: I understand now. The "move" in `PreserveWithoutMove` refers to a move of an attached node to another location in the tree. This currently cannot happen in optional fields. It is confusing that optional field also uses the term "move" with a different meaning. Once my attach/detach refactor goes in, the existing optional field sense of move will be gone, and `PreserveWithMove` will also not be needed, as moves will not be handled specially.

packages/dds/tree/src/test/feature-libraries/sequence-field/filterEdits.test.ts
jason-ha: I think the tooling can handle it - extra work but fine. Not true for all languages.
Our when I use my brain as compiler this is a slight bump. It would probably be less of an issue with better tooling. For example, if from web view I could hover and see the declaration.
Probably if the code base was just very consistent things would be okay.

packages/dds/tree/src/test/feature-libraries/optional-field/optionalField.spec.ts
jason-ha: Love it - though I would suggest extending into `filterMark`.
You *could* create an options type considering how often it is repeated. Nice place to document things too. :)
Okay to Resolve as is.

jason-ha: Also you can destructure in impl when you don't need options as an object.
```typescript
interface FilterEditOptions {
	filterAttach: EditFilterFunc;
	filterDetach: EditFilterFunc;
	preserveOtherEdits: boolean;
}
```

```typescript
function filterMark(
	mark: Mark,
	{ filterDetach, filterAttach, preserveOtherEdits }: FilterEditOptions
): Mark {
```

jason-ha (approved):
I have some requests (see prior), but they can all be in a follow-up.

packages/dds/tree/src/feature-libraries/sequence-field/filterEdits.ts
alex-pardes: Updated "Can convert move-out to remove" to include child changes. MoveIn marks cannot have child changes, so I removed that case.

packages/dds/tree/src/feature-libraries/modular-schema/fieldChangeHandler.ts
alex-pardes: Updated

packages/dds/tree/src/feature-libraries/sequence-field/filterEdits.ts
alex-pardes: Correct. I've updated this.

packages/dds/tree/src/feature-libraries/sequence-field/filterEdits.ts
```diff
@@ -0,0 +1,143 @@
...
+					filtered = {
+						type: "Remove",
+						count: mark.count,
+						revision: outputCellId.revision,
+						id: outputCellId.localId,
+					};
+
+					if (mark.changes !== undefined) {
+						filtered.changes = mark.changes;
+					}
```
jason-ha, line 120: These were removed from Insert/MoveIn, are these used? Tests now reach here?

jason-ha (approved):

PR #27688, opened 2026-07-14
Migrate website search to Pagefind

MarioJGMsoft:
## Description

Replace the website's hosted search integration with a static Pagefind index so search works without deployment credentials and can be validated from a local production build.

The website build now generates the Pagefind index after Docusaurus, and a local `SearchBar` theme override renders Pagefind's accessible component UI. The Pagefind module loads after React hydration to ensure the live navbar trigger controls the active modal. Search results label documentation as v1 or v2 and favor current v2 content while preserving normal relevance ranking. This also removes the previous Typesense search configuration and deployment variables, updates search styling and documentation, and adds end-to-end coverage for written docs, generated API docs, version labeling and ranking, and search after client-side navigation.

This revision fixes the visual regression that caused the Pagefind search control to sit above neighboring navbar controls. The search trigger now uses a full-height flex wrapper so it remains vertically centered without changing the dark-mode toggle's position, with Playwright coverage comparing both controls' vertical centers.

For videos comparing a before and after of the website UX, please see the comments of the linked work item.

Fixes [AB#77189](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/77189)

## Reviewer Guidance

The post-hydration script loading in `SearchBar` is intentional; loading the custom elements before hydration can leave the visible trigger connected to a detached modal.

### Known regressions / follow-up items

- **Tune conceptual versus API relevance.** Queries such as `SharedTree` still prioritize generated API references over the current quick-start content.
- **Reduce noisy and duplicate matches.** Scope indexing to primary content and tune which versions and headings are indexed.
- **Restore a shareable all-results view.** Add a dedicated search route with query state.
- **Restore recent-search history.** Add local query history or explicitly document that this feature will not be retained.


WayneFerrao (approved):

website/test/site-tests/Nav.spec.ts
```diff
@@ -8,6 +8,7 @@ import { test, expect } from "@playwright/test";
 test.describe("Nav", () => {
```
Josmithr, line 8: Thank you for adding tests!!!!!

website/src/css/custom.scss
```diff
@@ -37,18 +31,60 @@
...
-
-	// Workaround for <https://github.com/easyops-cn/docusaurus-search-local/issues/336>
-	--ifm-navbar-search-input-icon: url("@site/static/assets/search-icon-dark.svg");
+	--docsearch-muted-color: #748497;
 
 	//Text color for classes in code blocks
 	--ifm-attr-name-color: #00b3e6;
 }
 
+html[data-theme="dark"] {
```
Josmithr, line 40: Nit: any time we add css for the plugins we use, we should leave a comment explaining what plugin they're for, and link to that plugin's theming documentation (if possible).

PR #27689, opened 2026-07-14
Add an internal ODSP file-version manager for point-in-time load

lindsnguyen:
## Description

First component of point-in-time document load in `@fluidframework/odsp-driver`: an ODSP **file-version manager** that, given a target Fluid sequence number, selects the closest ODSP file version at or before it — the base from which a document can be loaded or replayed to that point in time.

Two pieces:

- `OdspVersionManager` — the selection logic (closest version ≤ target, dedup, tip handling, caching/refresh). It depends on an injected `IOdspFileVersionFetcher`, so the logic is unit-tested without ODSP.
- `createOdspFileVersionFetcher` — the real implementation: enumerates the file's versions (following `@odata.nextLink` paging) and resolves each version's sequence number from the version-scoped snapshot endpoint, reusing the driver's existing auth/epoch/fetch and snapshot parser.

These are **internal to the package** (not exported from its public entry points) and **not yet wired into a call site**, so there is no consumer-facing change. This is an incremental prototype; follow-ups will add the load/replay driver and the loader hookup.

Design notes and rationale live alongside the code in `packages/drivers/odsp-driver/src/odspVersionManager/DEV.md` — a "catechism" that mirrors the test suite (each documented behavior maps to a test via a stable ID).

Work item: https://dev.azure.com/fluidframework/internal/_workitems/edit/77292

## Reviewer Guidance

This is an incremental prototype posted for early feedback:

- The selection logic is fully unit-tested (14 tests, fetcher mocked), including caching and refresh — with a concurrent-refresh case that proves an in-flight fetch cannot repopulate a cleared cache.
- The fetcher has stubbed-`fetch` integration tests (10 tests) covering version enumeration with `@odata.nextLink` paging (and the empty-response fallback), sequence-number resolution from both JSON and `application/ms-fluid` (binary) snapshots, the missing-sequence-number error, non-success responses, and token-refresh retry.
- Feedback welcome on the `IOdspFileVersionFetcher` seam and the DEV.md catechism approach.

### DIFF
--- a/.changeset/odsp-file-version-manager.md
+++ b/.changeset/odsp-file-version-manager.md
@@ -1,0 +1,9 @@
+---
+"@fluidframework/odsp-driver": minor
+"__section": other
+"__includeInReleaseNotes": false
+---
+
+Add an internal ODSP file-version manager
+
+Adds `OdspVersionManager` and `IOdspFileVersionFetcher` (with `createOdspFileVersionFetcher`) to `@fluidframework/odsp-driver`. Given a target sequence number, the manager selects the closest ODSP file version at or before it — the base for loading or replaying a document at a point in time. These are internal to the package (not exported from its public entry points) and not yet wired into a call site, so there is no consumer-facing change.
--- a/.changeset/odsp-file-version-manager.md
+++ b/.changeset/odsp-file-version-manager.md
@@ -1,9 +1,0 @@
----
-"@fluidframework/odsp-driver": minor
-"__section": other
-"__includeInReleaseNotes": false
----
-
-Add an internal ODSP file-version manager
-
-Adds `OdspVersionManager` and `IOdspFileVersionFetcher` (with `createOdspFileVersionFetcher`) to `@fluidframework/odsp-driver`. Given a target sequence number, the manager selects the closest ODSP file version at or before it — the base for loading or replaying a document at a point in time. These are internal to the package (not exported from its public entry points) and not yet wired into a call site, so there is no consumer-facing change.
--- a/packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
+++ b/packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
@@ -1,0 +1,179 @@
+# ODSP Version History Load — Design Catechism
+
+## How to read this document
+
+This spec is a **catechism**: a hierarchical series of questions and answers that is the prose
+mirror of the test suite. Each topic-level question (`###`) corresponds to a `describe` block; each
+leaf question maps to one test, linked by a stable **code ID** at the end of the answer. The same ID
+appears as a `// @q <id>` tag on the matching test.
+
+ID format: `<area>-<topic>-<nn>`. Area is `M` (the version manager — [Part II](#part-ii--the-version-manager))
+or `F` (the file-version fetcher — [Part III](#part-iii--the-file-version-fetcher)). Topic is a short
+mnemonic (e.g. `SELECT`, `RESOLVE`). `nn` is a zero-padded counter. IDs are append-only: existing IDs
+never renumber, so links stay stable across edits.
+
+The contract:
+
+- **Every leaf Q&A has a test, and every test has a leaf Q&A**, joined by its code ID. A change to
+  behavior is a change to both.
+- **This document reflects the code as it is now.** Future/aspirational work lives in
+  [Part IV — Directional](#part-iv--directional), written as questions that cannot yet be answered
+  "yes".
+
+### Terminology (legend)
+
+This code is one piece of a larger capability — letting a document be viewed or recovered at an
+earlier point in time — which spans two repositories:
+
+- **Part 1 — point-in-time load** (this repository): given a target sequence number, produce a
+  read-only view of the document at that number.
+- **Part 2 — capture & marker** (a separate host repository): record the markers that name the points
+  worth returning to. Out of scope for this document.
+
+Part 1 is built in three components:
+
+- **Component A — the version manager**: choose which file version to load or replay from. **This
+  folder is Component A**, and this document is mostly about it.
+- **Component B — the recomposed driver**: load the chosen version and replay ops forward to the exact
+  target. Not built yet.
+- **Component C — the loader hookup**: expose Component B through the container loader. Not built yet.
+
+## Part I — Foundations
+
+These are conceptual answers with no single test; they frame everything below.
+
+### What problem does this code solve?
+
+Given a target Fluid **sequence number** (Fluid numbers every change to a document: op 1, 2, 3, …),
+we want to materialize the document as it was at that number. The first step is choosing a **base**:
+the most recent saved version whose state is at or before the target, from which the remaining ops
+can be replayed. This code finds that base.
+
+### What is an ODSP file version, and how is it different from a Fluid snapshot?
+
+Two different things are both called a "version":
+
+- A **file version** is an entry in the file's version history — a recoverable saved state of the whole
+  file, addressed by a label such as `"42.0"`. These are what a user could restore to.
+- A **Fluid snapshot** is an internal checkpoint the runtime writes; the driver's snapshot list
+  (`getVersions`) enumerates these, not the file versions.
+
+Selecting a base uses the **file version history**, enumerated by the driveItem `/versions` API — not
+the driver's snapshot list.
+
+### Why the closest version at or before the target, rather than any earlier one?
+
+Any base at or before the target can be replayed forward to the target and yields the same state, so
+the choice is not about correctness. The **closest** one minimizes how many ops must be replayed, and
+minimizes the chance that the needed ops have been trimmed from retention. So selection returns the
+greatest version sequence number that is still less than or equal to the target.
+
+### How is a version's sequence number obtained?
+
+By fetching that version's snapshot from the **version-scoped snapshot endpoint**
+(`.../versions/{label}/opStream/snapshots/trees/latest?blobs=2`), which returns the snapshot in the
+driver's normal (`application/json` or `application/ms-fluid`) framing. The driver's existing snapshot
+parser reads it, and the sequence number is `trees[0].sequenceNumber`. `blobs=2` inlines blob contents
+so the parser has everything it needs.
+
+### What is `minimumSequenceNumber`, and is it used for selection?
+
+It is the collaboration-window floor at a version (the point below which ops may be trimmed). It is
+carried on `ResolvedVersion` when read, but it is **not** an input to base selection — a version is a
+single point at its `sequenceNumber`, not a range.
+
+### What is deliberately not built here?
+
+Loading the base and replaying ops to the exact target (Component B), the loader hookup (Component C),
+and any test against a live ODSP file. See [Part IV](#part-iv--directional).
+
+## Part II — The Version Manager
+
+`OdspVersionManager` selects the base version. It depends on an injected `IOdspFileVersionFetcher`, so
+these behaviors are tested with an in-memory fake.
+
+### Which version does `findBaseForSeq` pick for a target sequence number?
+
+The list is newest-first, and the tip (index 0, the live document) is not a base candidate. Among the
+remaining versions, the answer is the greatest sequence number less than or equal to the target.
+
+- **Target between two versions?** The closer, older one. `M-SELECT-01`
+- **Target equal to a version?** That version, an exact match (zero ops to replay). `M-SELECT-02`
+- **Target newer than every version?** The newest recoverable version. `M-SELECT-03`
+- **Target older than every version?** `noBaseVersion`, reporting the oldest sequence number seen.
+  `M-SELECT-04`
+
+### How does it handle duplicate versions and the tip?
+
+- **Two versions share a sequence number?** Return the newest label (a metadata-only re-save leaves the
+  sequence number unchanged; the newest is closest to the head). `M-DEDUP-01`
+- **The tip (index 0)?** Never treated as a base; its sequence number is never even resolved.
+  `M-TIP-01`
+- **Only the tip exists?** `noBaseVersion`. `M-TIP-02`
+- **No versions at all?** `noBaseVersion`. `M-EMPTY-01`
+
+### What work does it avoid?
+
+- **Resolving more versions than needed?** It stops at the first version at or before the target and
+  does not resolve older ones. `M-STOP-01`
+- **Re-fetching across calls?** The version list and each resolved sequence number are cached.
+  `M-CACHE-01`
+- **Stale caches?** `refresh()` drops them so the next query re-enumerates. `M-CACHE-02`
+
+### What happens when a version cannot be resolved?
+
+The failure propagates; it is never swallowed into a wrong base. `M-ERR-01`
+
+### What does `listVersions` return?
+
+Every version with its resolved sequence number, newest-first. `M-LIST-01`
+
+## Part III — The File-Version Fetcher
+
+`createOdspFileVersionFetcher` is the real `IOdspFileVersionFetcher`, talking to ODSP. Its behaviors
+are tested against a stubbed `fetch` that returns canned responses through the real request,
+authentication, and snapshot-parsing code.
+
+### How does it enumerate versions?
+
+It calls the driveItem versions URL — built from the same API root as the snapshot call — and maps
+the `value` array to versions (newest-first). `F-LIST-01`
+
+### How does it resolve a version's sequence number?
+
+- **A well-formed snapshot?** It calls the version-scoped snapshot URL (`.../versions/{label}/opStream/snapshots/trees/latest?blobs=2`),
+  parses the response, and returns `trees[0].sequenceNumber`. `F-RESOLVE-01`
+- **A snapshot with no sequence number?** It throws, naming the version, rather than returning a wrong
+  value. `F-RESOLVE-02`
+
+## Part IV — Directional
+
+Aspirational behaviors, written as questions that cannot yet be answered "yes".
+
+### Should sequence-number resolution be lazy or binary-search, rather than eager?
+
+Resolving each version costs one snapshot fetch. With up to ~50 versions, an eager newest-to-oldest
+scan can fetch more than necessary. The public contract (`findBaseForSeq`) already hides the strategy,
+so a binary search over versions could replace it without changing callers.
+
+### Should there be a component that loads the base and replays ops to the exact target? (Component B)
+
+The manager only chooses the base. Materializing the document at an arbitrary target requires loading
+that version read-only and replaying the ops between the base and the target — sourcing later ops from
+the live op stream when the historical range has been trimmed.
+
+### Should this be exposed through the container loader? (Component C)
+
+Once Component B exists, a loader-facing entry point would let callers request "load at sequence
+number N" directly.
+
+### Should there be an end-to-end test against a real ODSP file?
+
+The fetcher is covered by stubbed-`fetch` integration tests, but not against a live file (which needs
+tenant credentials). An end-to-end test would exercise the real endpoints.
+
+### Should the raw driveItem `/content` download be a supported fallback?
+
+The `/content` download also contains a version's snapshot, but wrapped in a container framing the
+snapshot parser does not read directly. If the version-scoped snapshot endpoint is ever unavailable,
+unwrapping `/content` could be a fallback path.
--- a/packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
+++ b/packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
@@ -162,10 +162,23 @@ The manager only chooses the base. Materializing the document at an arbitrary ta
 that version read-only and replaying the ops between the base and the target — sourcing later ops from
 the live op stream when the historical range has been trimmed.
 
+Where would it live? In **this package**. Loading a historical file version is a storage-layer concern:
+it needs the version-scoped snapshot fetch, the epoch tracker, and authentication — all internal to this
+driver — and it consumes the version manager directly. A generic wrapping driver that only replays ops
+over an inner document service (the pattern `@fluidframework/replay-driver` uses) cannot reach the
+version-scoped base fetch, so the recomposition belongs beside `OdspDocumentService` /
+`OdspDocumentServiceFactory` rather than in a separate package. Because it consumes the version manager
+in-package, the version manager itself needs no exported surface; only the recomposed factory is exposed
+(as a legacy/alpha entry point) for the loader hookup to construct.
+
+A base is only needed when the live document's own snapshot no longer covers the target; when it does,
+loading paused at the target from the live snapshot suffices, and no historical version is loaded.
+
 ### Should this be exposed through the container loader? (Component C)
 
-Once Component B exists, a loader-facing entry point would let callers request "load at sequence
-number N" directly.
+Once Component B exists, a thin combining layer would construct the recomposed factory together with a
+standard loader, letting callers request "load at sequence number N" directly. It depends only on
+Component B's exposed factory, never on the version manager.
 
 ### Should there be an end-to-end test against a real ODSP file?
 
--- a/packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
+++ b/packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
@@ -65,8 +65,11 @@ the driver's snapshot list.
 
 Any base at or before the target can be replayed forward to the target and yields the same state, so
 the choice is not about correctness. The **closest** one minimizes how many ops must be replayed, and
-minimizes the chance that the needed ops have been trimmed from retention. So selection returns the
-greatest version sequence number that is still less than or equal to the target.
+minimizes the chance that the needed ops have been trimmed from retention. Selection therefore aims for
+the greatest version sequence number at or before the target. Because versions are enumerated
+newest-first and version order is expected to track sequence order, an early-stop scan finds it; if that
+ordering is ever violated, a valid but not-strictly-closest base may be chosen (still correct, just less
+optimal) — see [Part IV](#part-iv--directional) for the planned order-tolerant search.
 
 ### How is a version's sequence number obtained?
 
@@ -75,12 +78,6 @@ By fetching that version's snapshot from the **version-scoped snapshot endpoint*
 driver's normal (`application/json` or `application/ms-fluid`) framing. The driver's existing snapshot
 parser reads it, and the sequence number is `trees[0].sequenceNumber`. `blobs=2` inlines blob contents
 so the parser has everything it needs.
-
-### What is `minimumSequenceNumber`, and is it used for selection?
-
-It is the collaboration-window floor at a version (the point below which ops may be trimmed). It is
-carried on `ResolvedVersion` when read, but it is **not** an input to base selection — a version is a
-single point at its `sequenceNumber`, not a range.
 
 ### What is deliberately not built here?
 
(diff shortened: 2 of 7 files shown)


packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -0,0 +1,161 @@
...
+ */
+export interface ResolvedVersion extends OdspFileVersionRef {
+	/**
+	 * The Fluid sequence number the version's snapshot represents.
+	 */
+	readonly sequenceNumber: number;
+	/**
+	 * The collaboration-window floor at this version, if read. Not used for selection.
+	 */
+	readonly minimumSequenceNumber?: number;
```
sonalideshpandemsft: why is this required?

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
```diff
@@ -0,0 +1,127 @@
...
+				versionId: version.id,
+				lastModifiedDateTime: version.lastModifiedDateTime,
+				sizeBytes: version.size,
+			}));
+		});
+
+	const resolveSequenceNumber = async (versionId: string): Promise<number> =>
+		getWithRetryForTokenRefresh(async (options) => {
+			// Fetch the version's snapshot from the version-scoped snapshot endpoint, which returns the
+			// driver's native json / ms-fluid framing. `blobs=2` inlines blob contents (including the
```
sonalideshpandemsft: what's does ms-fluid framing mean?

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
```diff
@@ -0,0 +1,127 @@
...
+export interface OdspFileVersionFetcherProps {
+	readonly urlParts: IOdspUrlParts;
+	readonly getAuthHeader: InstrumentedStorageTokenFetcher;
+	readonly epochTracker: EpochTracker;
+	readonly logger: TelemetryLoggerExt;
+}
+
+/**
+ * Create an {@link IOdspFileVersionFetcher} that talks to a specific ODSP file.
+ */
```
sonalideshpandemsft, line 58: all the methods should be marked legacy alpha

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
```diff
@@ -0,0 +1,127 @@
...
+		});
+
+	return { listFileVersions, resolveSequenceNumber };
+}
+
+/**
+ * A version's snapshot must carry a sequence number; a missing one is surfaced as an error rather
+ * than returning a wrong value.
+ */
+function requireSequenceNumber(sequenceNumber: number | undefined, versionId: string): number {
```
sonalideshpandemsft: why does take in versionId?

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
lindsnguyen: oh good point, I originally had this in the code when i noticed it as a field in the attributes as well but realized later it wasn't something we actually needed. accidently left this after removing the population portion will remove

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -0,0 +1,161 @@
...
+	private readonly seqByVersion = new Map<string, number>();
+
+	public constructor(private readonly fetcher: IOdspFileVersionFetcher) {}
+
+	public refresh(): void {
+		this.versionsCache = undefined;
+		this.seqByVersion.clear();
+	}
+
+	public async findBaseForSeq(target: number): Promise<BaseForSeq> {
```
sonalideshpandemsft, line 107: I'm not clear how this method op trimming? Is the plan to implement that next?

Version | S
-- | --
V4 | 250
V3 | 170
V2 | 100
V1 | 50

Target: T = 200.

findBaseForSeq(200) scans:
V4 (S=250) -- skipped, 250 > 200.
V3 (S=170) -- 170 ≤ 200 → returned as { kind: "found", base: V3 }.

That's correct: V3 is the closest version at or before T.

But, suppose ops 171–175 were trimmed. Replaying V3 forward to T=200 is not possible now. How is this handled? or is this not possible?

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -0,0 +1,161 @@
...
+		const versions = await this.getVersions();
+		const resolved: ResolvedVersion[] = [];
+		for (const version of versions) {
+			resolved.push({ ...version, sequenceNumber: await this.resolveSeq(version.versionId) });
+		}
+		return resolved;
+	}
+
+	private async getVersions(): Promise<OdspFileVersionRef[]> {
+		this.versionsCache ??= await this.fetcher.listFileVersions();
```
anthony-murphy: Deep Review: Cache race re-populates stale data and defeats `refresh()`.

Line 148 caches the awaited *result*, not the pending promise: `this.versionsCache ??= await this.fetcher.listFileVersions();`. `??=` evaluates the LHS (`undefined`), then awaits the RHS, then assigns. If `refresh()` (lines 113-116, which sets `this.versionsCache = undefined` and clears `seqByVersion`) runs during that `await`, the in-flight fetch completes afterward and writes stale data back into `versionsCache` — silently undoing the refresh.

The same evaluate-LHS-before-await hazard applies to `resolveSeq()` (lines 152-159): a pending `this.seqByVersion.set(...)` on line 158 repopulates the map `refresh()` just cleared. Concurrent first-callers also each launch their own `listFileVersions()` because the promise isn't shared, defeating request coalescing.

Fix: cache the pending `Promise` instead of the resolved value, e.g. `private versionsCacheP: Promise<OdspFileVersionRef[]> | undefined;` then `this.versionsCacheP ??= this.fetcher.listFileVersions(); return this.versionsCacheP;`. This coalesces concurrent callers and prevents a pending fetch from writing back after `refresh()` drops the reference. For `resolveSeq`, cache the pending promise in the map or guard the `set` against a refresh generation counter. Add a test covering `refresh()` during an in-flight fetch — the current spec only tests sequential refresh.

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
```diff
@@ -0,0 +1,127 @@
...
+	props: OdspFileVersionFetcherProps,
+): IOdspFileVersionFetcher {
+	const { urlParts, getAuthHeader, epochTracker, logger } = props;
+	const { siteUrl, driveId, itemId } = urlParts;
+
+	const listFileVersions = async (): Promise<OdspFileVersionRef[]> =>
+		getWithRetryForTokenRefresh(async (options) => {
+			// The file's version history (distinct from the driver's snapshot list), from the same API
+			// root as the snapshot call so consumer (ODC) and enterprise (SPO) hosts are handled alike.
+			const url = `${getApiRoot(new URL(siteUrl))}/drives/${driveId}/items/${itemId}/versions`;
```
anthony-murphy: Deep Review: Version history is truncated to the first `/versions` page — no pagination.

`listFileVersions` issues exactly one GET to `.../items/${itemId}/versions` (line 63) and maps only `response.content.value` (lines 76-80) — no `@odata.nextLink` follow-up and no `?top=` bound. The sibling snapshot-`/versions` call in the same package bounds its request (`odspDocumentStorageManager.ts:505`, `${snapshotUrl}/versions?top=${count}`).

DEV.md Part IV assumes "up to ~50 versions", but the driveItem `/versions` REST API pages large histories via `@odata.nextLink`. For a document whose base version lives past the first page, `findBaseForSeq` scans a truncated list and returns `noBaseVersion` (or a too-new base) instead of the correct older base — the exact "wrong answer instead of an error" the DEV.md catechism says the design avoids, and precisely the older-target case point-in-time load exists to serve.

Fix: teach `listFileVersions` to follow `@odata.nextLink` (appending each page's `value`) until a valid base is found or history is exhausted, or bound with `?top=` and page as needed. Note the driveItem `/versions` endpoint differs from the sibling snapshot `/versions` path, so the `?top=` precedent is a guide, not a drop-in. Add a test proving a base past page 1 is still selected and that a truncated page cannot yield a wrong `noBaseVersion`. Gating this before the feature is wired to a call site is required for correctness on real documents.

packages/drivers/odsp-driver/src/odspVersionManager/index.ts
```diff
@@ -0,0 +1,17 @@
...
+	type BaseForSeq,
+	type IOdspVersionManager,
+	type OdspFileVersionRef,
+	type IOdspFileVersionFetcher,
+	type ResolvedVersion,
+} from "./odspVersionManager.js";
+export {
+	createOdspFileVersionFetcher,
+	type OdspFileVersionFetcherProps,
+} from "./odspFileVersionFetcher.js";
```
anthony-murphy: i think you are over exporting. i don't think anything is odspFileVersionFetcher needs to be exported from the package. minimizing exports helps bundle size, as anything not exported can be minimized. even exporting as internal will prevent minimization

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
```diff
@@ -0,0 +1,127 @@
...
+import type { OdspFileVersionRef, IOdspFileVersionFetcher } from "./odspVersionManager.js";
+
+/**
+ * Raw shape of a OneDrive/SharePoint driveItem version (an entry in the `/versions` response).
+ * @see https://learn.microsoft.com/en-us/onedrive/developer/rest-api/resources/driveitemversion
+ */
+interface IDriveItemVersion {
+	/** The version's label, e.g. "42.0". */
+	readonly id: string;
+	readonly lastModifiedDateTime: string;
```
sonalideshpandemsft, line 36: why is lastModifiedDateTime needed?

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -0,0 +1,161 @@
...
+	 */
+	refresh(): void;
+}
+
+/**
+ * Default {@link IOdspVersionManager}. Caches the version list and resolved sequence numbers. The
+ * resolution strategy (eager, newest-to-oldest, stopping at the first usable base) is hidden behind
+ * {@link findBaseForSeq} and can change without affecting callers.
+ */
+export class OdspVersionManager implements IOdspVersionManager {
```
anthony-murphy, line 96: rather than export the class. you should export a function which return the interface and internally constructs the class. this keeps the implementation off our exposed api surface

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Selects the file version to use as the base for loading or replaying to a target sequence number.
+ */
+export interface IOdspVersionManager {
+	/**
+	 * Given a target sequence number, return the closest version at or before it (`found`), or
+	 * `noBaseVersion` if the target predates the oldest retained version.
+	 */
+	findBaseForSeq(target: number): Promise<BaseForSeq>;
```
anthony-murphy, line 88: do we need any more api than this exposed? i would go pretty minimal here to start with

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -0,0 +1,161 @@
...
+ * resolution strategy (eager, newest-to-oldest, stopping at the first usable base) is hidden behind
+ * {@link findBaseForSeq} and can change without affecting callers.
+ */
+export class OdspVersionManager implements IOdspVersionManager {
+	private versionsCache: OdspFileVersionRef[] | undefined;
+	private readonly seqByVersion = new Map<string, number>();
+
+	public constructor(private readonly fetcher: IOdspFileVersionFetcher) {}
+
+	public refresh(): void {
```
anthony-murphy, line 102: its not necessary to do it now, but i was thinking we want to embed a number of heuristics throughout this to ensure we can efficiently manage things like refresh, and i wouldn't expose it to callers.

my rough thinking:

- each version gets a TTL (5min?) 
- we have a nearness const, lets say 1000 ops from the version's seq, as we don't need to get the optimal version for any seq, just one close enough that we can load to quickly.

when we get a request for base seq we look for any version which we've loaded who'd fetch time is less than the ttl, and is near enough to the requested base seq. if we find one, return it.

if we don't find one we get the version list, remove any versions not in the fresh list from seqByVersion, but keep those that do exist, as a versions seq will never change. then we start looking for a near enough version by loading seqs. versions won't necessarily be ordered, but we could do a fuzzy binary search to find a near enough version, which worst case is O(N) calls, but in practice is probably close to (N log N).

these are just rough ideas, we don't have to do any of it now, as they are really about optimizing calls to odsp. we can figure out a good plan latter, but its useful to think about how we could make it better as it might help with the api.

.changeset/odsp-file-version-manager.md
```diff
@@ -0,0 +1,9 @@
+---
```
anthony-murphy: this change should not have a changeset. we should mark all the apis internal. no consumer needs to know this is being built yet.

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
anthony-murphy: i think this can all be internal. i don't think consumer will directly use any of this. it will be used by whatever odsp version loader we ship.

at this phase i would suggest defaulting to internal for all work in this area, and only once we have a working system via e2e test do we focus on api and export the final shape.

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
lindsnguyen: ms-fluid is ODSP's compact binary snapshot format (content-type can be `application/ms-fluid` vs
`application/json`) The server can return either, so we accept both and
pin the version like the driver's snapshot fetch does; json is parsed by
`convertOdspSnapshotToSnapshotTreeAndBlobs`, binary by `parseCompactSnapshotResponse`.

honestly only learning about this though copilot so I'll update the comment to make it more clear

packages/drivers/odsp-driver/src/odspVersionManager/index.ts
lindsnguyen: Trimmed to just createOdspVersionManager + its return types. The fetcher, the class, props, and the fetcher interface aren't exported anymore.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
lindsnguyen: nope, you're right, only this needed. updated.

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
lindsnguyen: per our chat conversation, nothing's exported from the index now, so not adding any tags. If that changes post-prototype (before release) we'll tag them accordingly.

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
lindsnguyen: it was passed in to be used by the error text, but I realized this helper isn't really needed so removing and just throwing the error in the original method, checking if the seq number result from the parsed methods is undefined.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
lindsnguyen: yeah this PR doesn't do anything for op trimming. `findBaseForSeq` only selects the base version. The actual replay to T, and the trimmed-ops handling (jump to a newer snapshot that already absorbed them, else report not-materializable) will be done with this workitem: https://dev.azure.com/fluidframework/internal/_workitems/edit/77293

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
lindsnguyen: It's not neccessarily needed but I thought it might be useful if we want to use that information somehow to locate the correct version instead of just looping through one by one, or doing a binary serach of sorts until we find the correct version that has a seq# before the target seq#. Since we're already getting that info from the /versions list. Not tested or anything, but. I wanted the code to reflect what info is there in case it's useful for future changes. if we end up not using it we can remove it then.

Similarly though, size was another field given back, except that we're not using it for anything and I don't see why we'd need it so will remove that.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
lindsnguyen: Yeah one thought I had as well was leveraging the `lastModifiedDateTime` that came back from the `/versions` list api to somehow make it easier to find the version closest to our given seq number. Specifically, we'd have to implement inside the marker code to take a timestamp, but then we could use that as the starting point of the binary search rather than just starting from the midpoint? Haven't thought about it too deeply, but definately soemthing we can think about once we're no longer in the prototyping phase.

I've noted your ideas and mine in the DEV.md file so that we can do a followup once we're done with prototyping.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
lindsnguyen: switched both getVersions and resolveSeq to cache the pending promise instead of the awaited value, so a refresh() during an in-flight fetch can't be overwritten when the fetch settles, and concurrent callers coalesce onto one fetch. added a test (M-CACHE-03) covering refresh() while a version-list fetch is in flight.

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
lindsnguyen: listFileVersions now follows `@odata.nextLink`, fetching and concatenating each page until there's no nextLink, so a base version past the first page is found instead of a wrong noBaseVersion. Added a test (F-LIST-02) that returns a nextLink on page 1 and asserts versions from both pages are included and the nextLink is followed.

anthony-murphy:
## Deep Review

*Reviewed commit `0984768` on 2026-07-15.*

Not ready for sign-off, but moving well. The two blockers from the last review — the `refresh()` cache race and unbounded `/versions` pagination — are both fixed here with new tests (`M-CACHE-03`, `F-LIST-02`) and their threads are resolved. Two contained correctness concerns remain, flagged inline: rejected fetch promises are cached forever with no public recovery path, and `findBaseForSeq` unconditionally drops the newest version via `slice(1)` while the real fetcher emits no synthetic tip.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- Incremental prototype: the manager and fetcher are internal to the package, not exported from `index.ts`, and not wired into any call site. This is the load-bearing review lens — several convention divergences are guidance for the pending implementation, not shipped defects.
- Prior review blockers are fixed in this commit: the cache race now caches the pending promise (`M-CACHE-03`) and pagination now follows `@odata.nextLink` (`F-LIST-02`); both threads plus the over-export, changeset, and release-tag threads are resolved.
- Two different endpoints share the `/versions` suffix. The driver already owns a `/versions` call enumerating **Fluid snapshots** (`odspDocumentStorageManager.ts:505`, bounded with `?top=`); this PR adds a *different* one enumerating **driveItem file versions**. They look alike in logs and code search but are semantically unrelated — the new fetch reuses the same `"versions"` telemetry label, which could confuse telemetry.
- Greenfield directory: all 6 files are new (1147 additions, 0 deletions). Judge against the package's living conventions and the DEV.md catechism, not prior PRs.

- **Tip / ordering contract (needs domain knowledge)** — both early-stop correctness and the `slice(1)` concern hinge on what driveItem `/versions` returns: strictly newest-first, sequence order tracking list order, and whether index 0 is the non-recoverable live document or a recoverable version. Whether these are guaranteed API contracts or observed behavior needs server-side ODSP knowledge.
- **Fetcher placement (design tradeoff)** — `createOdspFileVersionFetcher` re-composes existing driver primitives (auth/epoch/retry/parsers) instead of living on `OdspDocumentStorageService`. Whether the standalone seam is the long-term home or a foldable prototype is an owner call.
- **Fail-closed policy (product decision)** — the scan intentionally fails the whole load when one version's sequence number is unresolvable. Skip-and-continue-with-telemetry is a defensible alternative but a product-policy choice to confirm.
- **Release tags** — `sonalideshpandemsft` [noted inline](https://github.com/microsoft/FluidFramework/pull/27689#discussion_r3582657976) that the methods should carry release tags; the author deferred tagging until post-prototype since nothing is exported ([reply](https://github.com/microsoft/FluidFramework/pull/27689#discussion_r3583324453)). Confirm the intended decoration when the surface is finalized.
- **Routing** — route to `odsp-driver` storage-area owners (maintainers of `odspDocumentStorageManager.ts` and the existing `/versions` path); they hold the validation, telemetry, and bounded-paging conventions this fetcher is benchmarked against.

</details>

- `00dd5bf` 2026-07-14 · **5/10** — async cache race and unbounded `/versions` pagination flagged

</details>

anthony-murphy:
`packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts:598`

Deep Review: Rejected fetch promises are cached permanently, with no public recovery path.

Both caches store the *promise*, not the resolved value. `getVersions()` does `this.versionsCache ??= this.fetcher.listFileVersions();` (line 598) and `resolveSeq()` stores `this.fetcher.resolveSequenceNumber(versionId)` in `seqByVersion` (lines 605-608) with no eviction on rejection. `getWithRetryForTokenRefresh` only retries auth/transient-token errors and rethrows everything else, so a single transient failure (a 500, or a 429 throttle after its one retry — both common against ODSP) leaves a rejected promise memoized forever; every later `findBaseForSeq`/`listVersions` replays that same rejection without re-attempting the fetch.

The only recovery, `refresh()` (line 553), is not on the public `IOdspVersionManager` interface (lines 534-540 expose only `findBaseForSeq`), and `OdspVersionManager` is not exported from `index.ts` — a consumer holding the `createOdspVersionManager()` result cannot recover except by recreating the manager.

This is coupled to the fix for the earlier race: caching the pending promise (which correctly closes the `refresh()` race) is exactly what makes a *rejected* promise sticky.

Fix: on rejection, evict the cached promise — attach a `.catch` that clears `this.versionsCache` / deletes the `seqByVersion` entry, guarding that it is still the same pending promise, so a later call retries while in-flight coalescing is preserved for successful requests. Add a "fails once, then succeeds on retry" test — the current error tests only assert propagation/coalescing on the first call. Optionally expose `refresh()` on `IOdspVersionManager` as a secondary recovery path.

anthony-murphy:
`packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts:561`

Deep Review: `findBaseForSeq` drops the newest version via `slice(1)`, but the production fetcher emits no synthetic tip.

`findBaseForSeq` does `const candidates = versions.slice(1);` (line 561) with the comment "index 0 ≈ the live document," and the `BaseForSeq` `@remarks` state that when the target is at/after the newest recoverable version, that newest version IS the correct `found` result (lines 497-500). But the real `listFileVersions` enumerates the driveItem `/versions` API and pushes each returned entry directly — it does not prepend a synthetic non-recoverable "tip." The `F-LIST-01` integration test proves the real fetcher returns exactly `["42.0","40.0"]` for a two-entry server response. Every manager unit test instead injects a fake `ref("tip")` at index 0 (`[ref("tip"), ref("42.0"), ref("40.0")]`) — an input shape production never emits, which masks the mismatch.

If driveItem `/versions[0]` is itself a recoverable saved version (DEV.md defines a file version as "a recoverable saved state"), then for history `[42.0(seq 448), 40.0(seq 418)]` and target `500`, the correct base `42.0` is silently skipped; with a single retained version, `findBaseForSeq` returns `noBaseVersion`.

The tier hinges on one question: **what does driveItem `/versions[0]` represent?** If index 0 is the live/in-progress document (non-recoverable), `slice(1)` is correct and this reduces to a test-fidelity gap. If index 0 is the newest recoverable version, `slice(1)` silently discards it. The code comment says "≈ the live document" (note the approximation); the `@remarks` say the newest recoverable version should be returned as `found`. Please confirm the API contract and align code, docs, and tests.

Fix: (a) confirm `/versions[0]` is the non-recoverable live document, make the fetcher/type say so explicitly, and drive at least one manager test from the *real* fetcher output shape (no injected synthetic tip); or (b) if `/versions[0]` is recoverable, drop the unconditional `slice(1)` and select over all entries. Add a test against the real `/versions` shape either way.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
anthony-murphy: Deep Review: Confirmed in this commit — caching the pending promise closes the `refresh()` race, and `M-CACHE-03` exercises `refresh()` while a version-list fetch is in flight. One coupled follow-up: a *rejected* promise is now cached permanently with no public recovery path; I've opened a separate inline thread on `getVersions`/`resolveSeq` for that.

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
anthony-murphy: Deep Review: Confirmed in this commit — `listFileVersions` now follows `@odata.nextLink` and concatenates pages, and `F-LIST-02` asserts a base past page 1 is selected. Resolved.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
sonalideshpandemsft: can we open a work item for this?

sonalideshpandemsft (approved):

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
anthony-murphy: every op has a timestamp, so if you can get a seq, you can also get the timestamp of that seq, which will be a server side timestamp, so likely more accurate than a client side timestamp, but there will still be time skew, so it won't be perfect. but  i like the idea as another heuristic to speed up the version search

lindsnguyen: filed: https://dev.azure.com/fluidframework/internal/_workitems/edit/77401

packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
```diff
@@ -0,0 +1,277 @@
...
+- **Target equal to a version?** That version, an exact match (zero ops to replay). `M-SELECT-02`
+- **Target newer than every version?** The newest recoverable version. `M-SELECT-03`
+- **Target older than every version?** `noBaseVersion`, reporting the oldest sequence number seen.
+  `M-SELECT-04`
+
+### How does it handle duplicate versions and the tip?
+
+- **Two versions share a sequence number?** Return the newest label (a metadata-only re-save leaves the
+  sequence number unchanged; the newest is closest to the head). `M-DEDUP-01`
+- **The tip (index 0)?** Never treated as a base; its sequence number is never even resolved.
```
markfields, line 109: Why is there this special case for the latest version?

PR #27656, opened 2026-07-08
Add minDocumentRuntimeVersion compatibility API

WillieHabi:
## Description

Adds `minDocumentRuntimeVersion` as the preferred name for configuring the minimum Fluid runtime version required to open or process documents created or loaded by a container runtime.

The existing `minVersionForCollab` parameter remains supported as a deprecated compatibility alias for now, with removal planned for 3.0.0 as part of https://github.com/microsoft/FluidFramework/issues/27180. Callers should specify only one of `minDocumentRuntimeVersion` or `minVersionForCollab`.

This also makes `MinDocumentRuntimeVersion` the preferred exported type for new API surfaces. Existing APIs that referenced `MinimumVersionForCollab` have been updated to reference `MinDocumentRuntimeVersion`. `MinimumVersionForCollab` remains supported as a deprecated alias.

Additional compatibility-related surfaces now use the preferred name where safe:

- `CodecWriteOptionsBeta` now supports `minDocumentRuntimeVersion` and keeps `minVersionForCollab` as a deprecated compatibility alias.
- Runtime/datastore context surfaces now expose `minDocumentRuntimeVersion` while keeping `minVersionForCollab` as a deprecated compatibility alias where needed.
- Service client method parameter names now use `minDocumentRuntimeVersion`.

Persisted schema/wire-format fields intentionally remain named `minVersionForCollab`

PR #27693, opened 2026-07-14
Introduce Unified ServiceClient API for Fluid Client

CraigMacomber:
## Description

This add a simple API as a cleaner alternative to fluid-static and aquaduct.

The implementation details are well encapsulated, making this a viable alternative to the `fluid-static` public API and a viable successor to the existing much larger legacy API surface we do not want to stabilize.

This addresses several issues with our existing APIs:

1. non-legacy (fluid-static declarative model) API:
    1. No stable way to use local service for testing (Internally or for customers). Forces tests to use alternative patterns (like test-utils, mocks etc. or a separate azure local service process).
    2. No way to avoid bundling in SharedDirectory.
    3. Defining and using DataStores is not well supported, especially for the root.
    4. Dependency layering is such that its disallowed for DDS packages to use the public API surfaces to do collab tests.
    5. No abstraction for Services / ServiceClient makes logic portable over services harder to write.
    6. Doesn't support lazy loading DDS and DataStore code (at least not clearly/cleanly).
    7. No migration path from legacy APIs to this
3. legacy (encapsulated model, including aquaduct):
    1. Leak way too many details with too many non-sealed interfaces presenting tons of theoretical extension points we don't want to stabilize.
    2. Encourages use of DataObject and its subclassing based extension approach which is limiting (Like how it makes switching to root trees) hard, and makes changing the implementation without introducing breaking changes very difficult.

Because this new "unified" API offers value for our own internal testing (for example in tree), it is useful, even as alpha (or internal).

Exposing this new API as alpha should be low risk, as it makes no long-term commitments.

Since we have first party in-repo use cases, it should be practical to refine and iterate on this API with use in test and examples before we consider promoting it to beta.

Requirements for this new API:

1. Testing:
    2. Can easily be used for tests including collab and reopening documents.
    3. Can be used by DDS (like shared-tree) tests replacing the need for many of our mocks and test utils
    4. Can be used by customers (apps and libraries requiring stable APIs when promoted past alpha). When adopted, customers should not have need of any of our legacy test-utils and mocks (so we can migrate our legacy apps off of our test utils and eventually make them internal only).
5. Will be possible to stabilize to public (or at least beta) without depending on or stabilizing legacy APIs.
6. Is easy to use for tests, including collab.
7. Is easy to use for real production apps (should work the same as with tests).
8. Provides a clear and practical incremental migration path from both existing APIs: there can be gaps we need to close for some use-cases but those gaps need to be practical to close.

## Reviewer Guidance

The example in the changeset is a good place to get started, followed by the examples/tests in the tree package.
Next I'd recomend reviewing the actual types for the API followed by its implementation.
See https://github.com/microsoft/FluidFramework/pull/27078 for many more examples of how this could be implemented in more examples and for other services.


.changeset/service-client-alpha.md
```diff
@@ -0,0 +1,39 @@
...
+Apart from the `@fluidframework/local-driver` helpers (which come from `@fluidframework/local-driver/alpha`), these APIs are also re-exported from `fluid-framework`. None reference any `@legacy` types.
+
+Example:
+
+```typescript
+import { createEphemeralServiceClient } from "@fluidframework/local-driver/alpha";
+import { treeDataStoreKind } from "fluid-framework/alpha";
+
+const client = createEphemeralServiceClient();
+const root = treeDataStoreKind({ type: "my-app-root", config, initializer });
```
Josmithr: Would it be valuable to show how `config` / `initializer` are initialized and/or acquired?

packages/common/driver-definitions/src/serviceClient.ts
```diff
@@ -0,0 +1,361 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import type { ErasedBaseType } from "@fluidframework/core-interfaces/internal";
+
+/*
+ * This file defines the public API for the ServiceClient and related types.
```
Josmithr: Nit
```suggestion
 * This file defines the public API for the {@link ServiceClient} and related types.
```

packages/common/driver-definitions/src/serviceClient.ts
```diff
@@ -0,0 +1,361 @@
...
+ *
+ * Package policy (intentional, @alpha):
+ * All code interacting through this API within a single client must use one copy — a single instance at a single version — of each Fluid Framework client package.
+ * Mixing multiple copies or versions of a package is not supported. This mirrors the @public "declarative model" APIs and is a deliberate simplification.
+ * It is enforced best-effort only: @sealed nominal erased types catch many mismatches at compile time, and factory identity checks throw a UsageError ("Conflicting ... with same type") at run time, but the checking is not exhaustive.
+ * See `LayerCompatibilityUnified.md` for the full policy, rationale, and failure signatures.
```
Josmithr, line 24: Can we use a `{@link}` here to point to the file? Since module comments don't get picked up by API extractor (with the exception of the `@packageDocumentation` comment in the barrel file), we shouldn't need to worry about generating unusable links. And I would think intellisense would support this (would need to be verified).

Josmithr (commented):
(I haven't finished reviewing yet, but wanted to leave this suggestion before I forget)

I think it would be valuable to add a simple example app (e.g. dice roller) that illustrates how a user would use these APIs to create such an app. This would also be useful for myself and others to review the user-facing implications here.

.changeset/service-client-alpha.md
CraigMacomber: DOne.

packages/common/driver-definitions/src/serviceClient.ts
CraigMacomber: There is no Intelisense for comments like this: it doesn't parse tags or do any syntax hilighting, and there is nothing you can hover over to see a formatted version of it.

If we do file or region level linter suppressions for two different rules and also format this as a doc comment by adding the extra "*", then a link tag here gets syntax hilighted as a link, but there is no way to click it. Markdown links are not even syntax hilighted. file:// links work but require absolute paths which are machine specific.

So, no, I don't think we can put a link here unless we do something like link the URL on github this file will end up at if we merged this PR.

packages/common/driver-definitions/src/serviceClient.ts
```diff
@@ -0,0 +1,360 @@
...
+ * Strips patch and prerelease from a SemVer string, returning only the major and minor version.
+ * @remarks
+ * This formats a version in the same style used by {@link MinimumVersionForCollaboration}, specifying only the major and minor versions,
+ * which are the portions used for feature selection.
+ * @privateRemarks
+ * This fills a similar role as cleanedPackageVersion in `@fluidframework/runtime-utils`.
+ * It can be used to workaround our generated pkgVersion values being invalid `MinimumVersionForCollaboration` on CI (due to prerelease) or patched release branches.
+ * @alpha
+ */
+export function featureVersion<major extends `${bigint}`, minor extends `${bigint}`>(
```
Josmithr, line 157: Nit: does the semver library not offer something we can leverage for this?

packages/common/driver-definitions/src/serviceClient.ts
```diff
@@ -0,0 +1,360 @@
...
+ * A context which has a registry and can create data stores using it.
+ * @sealed
+ * @alpha
+ */
+export interface DataStoreCreator {
+	/**
+	 * Create a new detached datastore `T` which can be attached to the {@link FluidContainer}.
+	 * by adding a handle to a DataStore or SharedObject which is already attached to the {@link FluidContainer}.
+	 * @remarks
+	 * `kind` must be included in the registry used to create or load this {@link DataStoreCreator}.
```
Josmithr: Probably worth documenting any errors that are expected to be thrown by this (e.g. if kind isn't in the registry)

packages/common/driver-definitions/src/serviceClient.ts
```diff
@@ -0,0 +1,360 @@
...
+	 */
+	close(): void;
+}
+
+/**
+ * A Fluid container with an associated {@link ServiceClient} it can attach to.
+ * @sealed
+ * @alpha
+ */
+export interface FluidContainerWithService<T = unknown> extends FluidContainer<T> {
```
Josmithr: Should `T` be `TData` to match `FluidContainer`? (Same feedback for the below interfaces as well)

packages/common/driver-definitions/src/serviceClient.ts
```diff
@@ -0,0 +1,360 @@
...
+ * Since it implements {@link DataStoreKey}, a `DataStoreKind` can also be used directly as the key to look
+ * itself up in a {@link Registry}.
+ * @privateRemarks
+ * TODO:
+ * SharedObjects should be usable as these (though putting shared objects directly in the container might need special logic).
+ * Type erased {@link IFluidDataStoreFactory}.
+ * @sealed
+ * @alpha
+ */
+export interface DataStoreKind<out T = unknown>
```
Josmithr, line 325: We should document `T` (same feedback for below types)

packages/common/driver-definitions/src/serviceClient.ts
```diff
@@ -0,0 +1,360 @@
...
+
+/**
+ * A connection to a Fluid storage service.
+ * @sealed
+ * @alpha
+ */
+export interface ServiceClient {
+	/**
+	 * Creates a detached container associated with this service client.
+	 * @privateRemarks
```
Josmithr, line 355: Nit: we should probably add a `@param` comment for `root` (we have one in the other overload)

packages/common/driver-definitions/src/serviceClient.ts
```diff
@@ -0,0 +1,360 @@
...
+		root: DataStoreKey<T>,
+		registry: DataStoreRegistry,
+	): Promise<FluidContainerWithService<T>>;
+
+	/**
+	 * Loads an existing container from the service.
+	 * @param id - The unique identifier of the container to load.
+	 * @param root - The {@link DataStoreKind} for the root, or a registry which will be used to look up the root based on its type.
+	 *
+	 * @throws a UsageError if the DataStoreKind's type (either the root directly or looked up from the registry) does not match the type of the root data store in the container.
```
Josmithr: Nit: `{@link}` to `UsageError`?

packages/dds/shared-object-base/src/dataStoreKind.ts
```diff
@@ -0,0 +1,272 @@
...
+ * @privateRemarks
+ * TODO: The framework provided SharedObjects should be exposed in a way indistinguishable from custom Sub-DataStores.
+ * This can be done by unifying the DataStoreKind and SharedObjectKindAlpha types.
+ * For now, this would mean having DataStoreKind extend SharedObjectKindAlpha, since we can allow a DataStore in all places SharedObjects are allowed,
+ * but do not allow SharedObjects at the root.
+ * Fixing this, and allowing shared objects at the root (maybe use a trivial wrapper DataStore) could simplifying things, allowing DataStores and Containers to share some types (like how they create detached contents, have registries, have a root etc).
+ *
+ * Part of this unification could be to relax the output from the factories / registries. Allowing the output to be an arbitrary type, which might be a promise, and might not be one could help.
+ * Removal of the IFluidLoadable requirement, and allowing the returned type to expose handles to itself how ever it wants (or not at all) might be viable and simplify typing And allow for strongly typed handles at creation time at least).
```
Josmithr: Nit
```suggestion
 * Removal of the IFluidLoadable requirement, and allowing the returned type to expose handles to itself however it wants (or not at all) might be viable and simplify typing And allow for strongly typed handles at creation time at least).
```

packages/dds/shared-object-base/src/dataStoreKind.ts
```diff
@@ -0,0 +1,272 @@
...
+ * Maybe when Registry(type) gives a promise, it could instead give a factory which outputs a promise wrapped type? The check that the provided creation key is valid for that factory can be deferred until the promise resolves.
+ *
+ * Idea: creation key can have an interface that subsets the factory / SharedObjectKindAlpha / DataStoreKind so they can be used, or some branded key (string, and/or object with stronger identity what knows the type string) can be used.
+ * Have Key interface contain validation function to check that the factory used (or maybe the value produced from it) is valid for that key.
+ * During load, get with validation that simply checks the factory's type string matches the key's type string.
```
Josmithr: Typo

packages/dds/shared-object-base/src/dataStoreKind.ts
```diff
@@ -0,0 +1,272 @@
...
+	 * @param context - A {@link DataStoreContext} that can be used to create additional shared objects.
+	 */
+	view(root: TRoot, context: DataStoreContext): Promise<TOutput>;
+}
+
+/**
+ * Creates a {@link @fluidframework/driver-definitions#DataStoreKind} from {@link DataStoreOptions}.
+ * @alpha
+ */
+export function dataStoreKind<T, TRoot extends IFluidLoadable>(
```
Josmithr: Nit: our guidelines call for functions to be verb phrases when practical. E.g. `createDataStoreKind`?

packages/dds/shared-object-base/src/dataStoreKind.ts
```diff
@@ -0,0 +1,272 @@
...
+): ISharedObjectRegistry {
+	return {
+		get: (type: string) => {
+			const entry = lookup(type);
+			return asSharedObjectKind(entry).getFactory();
+		},
+	};
+}
+
+const rootSharedObjectId = "root";
```
Josmithr, line 203: This could use documentation, especially since it impacts the persisted data representation.

packages/dds/shared-object-base/src/dataStoreKind.ts
```diff
@@ -0,0 +1,272 @@
...
+	context: IFluidDataStoreContext,
+	existing: boolean,
+	options: DataStoreOptions<TRoot, T>,
+): Promise<IFluidDataStoreChannel> {
+	const sharedObjectRegistry = await options.registry();
+	const runtime: FluidDataStoreRuntime = new FluidDataStoreRuntime(
+		context,
+		convertRegistry(sharedObjectRegistry),
+		existing,
+		async (rt: IFluidDataStoreRuntime) => {
```
Josmithr: Nit: can we avoid the "rt" abbreviation here?

packages/drivers/local-driver/src/localService.ts
```diff
@@ -0,0 +1,318 @@
...
+ *
+ * @remarks
+ * This can be used to cleanup lingering timers from containers created using a client from {@link createEphemeralServiceClient}.
+ * Such timers are a common case of hangs on exist (often worked around using Mocha's `--exit` flag).
+ * @alpha
+ */
+export async function closeEphemeralContainers(): Promise<void> {
+	const toClose = containers;
+	containers = [];
+	for (const c of toClose) {
```
Josmithr: Nit: `container` instead of `c`

packages/drivers/local-driver/src/ephemeralService.ts
```diff
@@ -0,0 +1,318 @@
...
+ * @param documentId - the existing document to open.
+ * @privateRemarks
+ * Like createLocalResolverCreateNewRequest, but without the option to create a new document.
+ * TODO: At some point we should avoid specifying the URL in so many places, but the current APIs don't accommodate it yet.
+ */
+const createLoadExistingRequest = (documentId: string): IRequest => {
+	return { url: `http://localhost:3000/${documentId}` };
+};
+
+let documentIdCounter = 0;
```
Josmithr, line 445: Docs?

CraigMacomber:
> (I haven't finished reviewing yet, but wanted to leave this suggestion before I forget)
> 
> I think it would be valuable to add a simple example app (e.g. dice roller) that illustrates how a user would use these APIs to create such an app. This would also be useful for myself and others to review the user-facing implications here.

As noted in the PR description, see https://github.com/microsoft/FluidFramework/pull/27078 which has several apps ported to the new APIs.

packages/runtime/runtime-utils/src/serviceClientBase.ts
```diff
@@ -0,0 +1,249 @@
...
+	IContainerRuntimeBase,
+	IFluidDataStoreChannel,
+	IFluidDataStoreContext,
+	IFluidDataStoreFactory,
+} from "@fluidframework/runtime-definitions/internal";
+import { UsageError } from "@fluidframework/telemetry-utils/internal";
+
+/*
+ * This file provides common implementation logic for ServiceClient implementations.
+ *
```
Josmithr: Nit
```suggestion
```

packages/runtime/runtime-utils/src/serviceClientBase.ts
```diff
@@ -0,0 +1,249 @@
...
+	public static narrowGeneric<T>(
+		value: T,
+	): asserts value is DataStoreKindImplementation<
+		T extends DataStoreKind<infer T2> ? T2 : never
+	> &
+		T {
+		if (!DataStoreKindImplementation.guard(value)) {
+			throw new Error("Invalid DataStoreKindImplementation");
+		}
+	}
```
Josmithr, line 60: Nit: any reason these can't be free functions instead?

packages/runtime/runtime-utils/src/serviceClientBase.ts
```diff
@@ -0,0 +1,249 @@
...
+
+	/**
+	 * Creates the service-specific request used to attach this container.
+	 * @remarks
+	 * Called by {@link ServiceContainerBase.attach} after validating that no attach is already in progress.
+	 */
+	protected abstract createAttachRequest(): IRequest;
+
+	/**
+	 * Extracts the container id from {@link ServiceContainerBase.container}'s resolved URL after attachment.
```
Josmithr: Nit
```suggestion
	 * Extracts the container ID from {@link ServiceContainerBase.container}'s resolved URL after attachment.
```

packages/runtime/runtime-utils/src/serviceClientUtils.ts
```diff
@@ -0,0 +1,243 @@
...
+	): Promise<FluidContainerWithService<T>>;
+
+	load<T>(
+		registry: DataStoreRegistry<T>,
+		options: TOptions,
+		id: string,
+	): Promise<FluidContainerAttached<T>>;
+}
+
+class ServiceClientImpl<TOptions> implements ServiceClient {
```
Josmithr: Nit: `ServiceClientImplementation`?

packages/runtime/runtime-utils/src/serviceClientUtils.ts
```diff
@@ -0,0 +1,243 @@
...
+		return this.statics.load(normalizeRegistry(root), this.options, id);
+	}
+}
+
+/**
+ * Creates a {@link @fluidframework/driver-definitions#ServiceClient} that delegates container
+ * creation and loading to the supplied container class statics.
+ * @internal
+ */
+export function makeServiceClientImpl<TOptions>(
```
Josmithr: Should this be just "makeServiceClient"?

packages/framework/fluid-framework/api-report/fluid-framework.alpha.api.md
```diff
@@ -1307,6 +1360,9 @@ export type MemberChangedListener<M extends IMember> = (clientId: string, member
 // @alpha @deprecated
 export const minimize: TransactionPostProcessor;
 
+// @alpha @input
+export type MinimumVersionForCollaboration = `2.${bigint}.0`;
```
Josmithr, line 1376: @CraigMacomber it looks like we're migrating away from the "minimum version for collaboration" naming elsewhere. FYI.
https://github.com/microsoft/FluidFramework/pull/27656

packages/common/driver-definitions/src/serviceClient.ts
CraigMacomber: Added a comment to the code about this.

packages/common/driver-definitions/src/serviceClient.ts
CraigMacomber: This is actually full user controlled by the registry they provide. I have updated the docs here to indicate that.

packages/runtime/runtime-utils/src/serviceClientBase.ts
CraigMacomber: They could be. I think it makes a for a much nicer API for them to be statics. This puts them right next to (API wise) the inherited static like `narrow` `upCast` and `[hasInstance]`, and avoids cluttering the top level namespace with a bunch of utilities for working with this specific class. As they compile town to almost nothing, I don't think tree shaking should be a concern (which would be the main reason to make statics on an internal class into free functions)

packages/framework/fluid-framework/api-report/fluid-framework.alpha.api.md
CraigMacomber: I'm going to leave this as open an unaddressed for now, waiting until either this is close to ready or that PR goes in to do a rename.

.changeset/service-client-alpha.md
```diff
@@ -0,0 +1,69 @@
...
+---
+Add new @alpha ServiceClient API for creating and loading Fluid containers
+
+This introduces an experimental (`@alpha`), service-agnostic API for working with Fluid containers whose root is an arbitrary data store, along with an in-memory implementation for testing.
+
+The new surface is made up of:
+
+- `ServiceClient` (`@fluidframework/driver-definitions`): the entry point for creating and loading containers. Along with it come the supporting container types (`FluidContainer`, `FluidContainerWithService`, `FluidContainerAttached`), the data store model (`DataStoreKind`, `DataStoreKey`, `DataStoreRegistry`, `DataStoreCreator`), and the generic registry primitives (`Registry`, `RegistryKey`, `registryLookup`, `basicKey`).
+- `createDataStoreKind` and `sharedObjectRegistryFromIterable` (`@fluidframework/shared-object-base`): build a `DataStoreKind` from a root shared object and a registry of shared object kinds.
+- `treeDataStoreKind` and `instantiateTreeFirstTime` (`@fluidframework/tree`): a SharedTree-specific convenience wrapper that produces a `DataStoreKind` backed by a `TreeView`.
```
Josmithr: I don't find these function names to be very intuitive / user-friendly. The "DataStoreKind" nomenclature is weird (in my opinion), and doesn't reflect how the APIs are _used_ (e.g. `const numberStore = treeDataStoreKind(...)`).

`treeDataStoreKind` also doesn't follow our guidelines that call for functions to be named as verb phrases.

Would something like "defineDataStore" be a better alternative to `createDataStoreKind`? And maybe "defineTreeDataStore" for the tree case?

packages/common/driver-definitions/src/serviceClient.ts
```diff
@@ -0,0 +1,395 @@
...
+): TOut {
+	return key.adapt(registry(key.type));
+}
+
+/**
+ * Creates a simple {@link RegistryKey} which does no type conversion.
+ * @typeParam T - The type of the registry entry, which is returned unchanged by the key.
+ * @alpha
+ */
+export function basicKey<T>(type: string): RegistryKey<T, T> {
```
Josmithr: Nit: "createBasicKey"? Or "createBasicRegistryKey"? (to follow our verb phrase function-naming conventions).

Josmithr: (`registryLookup` above is also suspect in this regard, but "lookUpInRegistry" feels clunkier than necessary to me 🤷‍♂️)

CraigMacomber: Renamed. Also did lookupInRegistry

.changeset/service-client-alpha.md
CraigMacomber: I like the use of define and omission of kind here. Did both renames, and updates to docs to use the phrasing as well.

packages/framework/fluid-framework/api-report/fluid-framework.alpha.api.md
CraigMacomber: That PR has been superseded by https://github.com/microsoft/FluidFramework/pull/27726, which limits the scope of the rename. There is still some ongoing discussion about the new name, so I'm not going to assume it will happen and will be leaving this as is for now unless its settled before this this gets approved. Its an alpha API, we we can do the rename later if needed.

Josmithr (approved):
This looks good to me. And I'd love to see us unblocked on using this to clean various things up. 🚀

Josmithr (approved):

packages/dds/tree/src/entrypoints/alpha.ts
```diff
@@ -321,12 +323,14 @@ export {
...
+	instantiateTreeFirstTime, 
 	minimize, 
 	normalizeAllowedTypes, 
 	persistedToSimpleSchema, 
 	replaceConciseTreeHandles, 
 	replaceHandles, 
 	replaceVerboseTreeHandles, 
-	trackDirtyNodes
+	trackDirtyNodes, 
+	defineTreeDataStore
```
jason-ha, line 334: This does not appear sorted - didn't use `generate:entrypoint-sources`?
(I just ran it and get a diff here as tool sorts it.)

CraigMacomber: This branch had tons of rebases, refactors and renames and merges, and originally predated these entry point files I think, so I suspect at some point yes, this did not get regenerated. My mistake.

I really wish this was like our other generated code versioned code (lock file, type tests, package versions, API reports etc), and got generated by our build and failed CI if out of sync.

PR #27695, opened 2026-07-15
fix(odsp-driver): harden version-manager caching and fetch validation

lindsnguyen:
## Description

Follow-up to #27689 hardening the internal ODSP file-version manager in `@fluidframework/odsp-driver`.
Everything is internal to the package (not exported, not wired into a call site), so there is no
consumer-facing change and no changeset.

- **Caching.** Resolved sequence numbers and version epochs are memoized via `PromiseCache` (this
  converts the `epochByVersion` map added in #27748), so each is resolved at most once per manager
  instance and a failed resolution is retried rather than cached. The version list itself is re-fetched
  each call, since it changes as new versions are cut. The manager is short-lived — the point-in-time
  document service factory creates one per load, calls `findBaseForSeq` once, then drops it — so the
  caches need no eviction.
- **Fetcher hardening.** `resolveSequenceNumber` handles `application/json` and `application/ms-fluid`
  explicitly and throws a typed `incorrectServerResponse` on anything else (e.g. an HTML error page)
  rather than mis-parsing it; it also rejects a missing or non-integer sequence number instead of
  feeding a wrong value into base selection. Version labels are percent-encoded into the snapshot URL.
- **Docs & tests.** `DEV.md` and the manager/fetcher specs are updated to match.

## Reviewer Guidance

- Internal to `@fluidframework/odsp-driver` (no public API, no changeset).
- The version manager is created per load and used for a single `findBaseForSeq`, so its caches are plain
  memoization with no eviction or single-flight machinery.

### DIFF
--- a/packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
+++ b/packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
@@ -9,8 +9,9 @@ appears as a `// @q <id>` tag on the matching test.
 
 ID format: `<area>-<topic>-<nn>`. Area is `M` (the version manager — [Part II](#part-ii--the-version-manager))
 or `F` (the file-version fetcher — [Part III](#part-iii--the-file-version-fetcher)). Topic is a short
-mnemonic (e.g. `SELECT`, `RESOLVE`). `nn` is a zero-padded counter. IDs are append-only: existing IDs
-never renumber, so links stay stable across edits.
+mnemonic (e.g. `SELECT`, `RESOLVE`). `nn` is a zero-padded counter assigned the lowest number available
+within its `<area>-<topic>` group. When a scenario is removed, the remaining IDs in that group are
+renumbered to close the gap, so the numbering stays compact with no reserved or retired IDs.
 
 The contract:
 
@@ -40,6 +41,38 @@ Part 1 is built in three components:
 - **Component C — the loader hookup**: expose Component B through the container loader. **Built** in
   `@fluidframework/container-loader` (`loadContainerToSequenceNumber`) — see
   [Part V](#part-v--components-b--c-as-built).
+
+### Vocabulary (the overloaded terms)
+
+Several words describe "the saved state of a document" at different layers, which is a frequent source of
+confusion. Everything is built from two primitives — **ops** and the **state** they produce — and the rest
+is either a bundle of ops or a saved snapshot of state.
+
+- **Op**: a single change, globally ordered by a **sequence number**. The full history *is* the ordered
+  ops; the document is those ops replayed.
+- **Summary**: the Fluid **runtime**'s serialized tree of current state. A summarizer client *produces* a
+  summary and proposes it via a `summarize` → `summaryAck` op handshake. ("Summary" = the act/content the
+  client generates.)
+- **Snapshot**: the stored, fetchable form of a summary (tree + blobs, carrying a `sequenceNumber`). To
+  load, you fetch a snapshot and replay ops forward. Note two different snapshot lists exist: the driver's
+  *Fluid* snapshot list (`getVersions`) is **not** the ODSP file-version history — a different address
+  space.
+- **ODSP file version**: an entry in the file's version history (driveItem `/versions`, labels like
+  `"42.0"`). ODSP surfaces stored snapshots as recoverable version rows, with a retention cap and dedup.
+  **This is what Component A selects over.**
+- **op stream / `opStream`**: overloaded. As a concept, the ordered op log. As a URL segment, a path prefix
+  under which *both* snapshots (`.../opStream/snapshots/...`) and raw ops (`.../opStream?filter=...`) live.
+- **delta storage**: the durable REST op log (`OdspDeltaStorageService`), which fetches raw ops by
+  sequence-number range and is retention-limited. Distinct from —
+- **trailing / bundled ops**: a small tail of ops baked *inside* a snapshot (up to `latestSequenceNumber`),
+  returned with the snapshot for free — not a separate fetch.
+- **sequence number** = an op's global order index (a snapshot's is the op its tree is current through);
+  **latestSequenceNumber** = the last bundled trailing op; **minimumSequenceNumber (MSN)** = the
+  collaboration-window floor (seq all connected clients have acked) — **not** a retention/trimming signal.
+
+Chain of custody for one saved state: a **summarizer** writes a **summary** → stored as a **snapshot** →
+surfaced by ODSP as a **file version**. Same state, three names because three layers own it.
+
 
 ## Part I — Foundations
 
@@ -108,45 +141,60 @@ these behaviors are tested with an in-memory fake.
 
 ### Which version does `findBaseForSeq` pick for a target sequence number?
 
-The list is newest-first, and the tip (index 0, the live document) is not a base candidate. Among the
-remaining versions, the answer is the closest one at or before the target — the greatest sequence number
-at or before the target when version order tracks sequence order, which an early-stop newest-first scan
-finds.
+The list is newest-first, and the tip (index 0, the newest version) is excluded: it is the one version
+whose sequence number is not yet static, so it cannot be a stable base. Among the remaining (sealed)
+versions the answer is the closest one at or before the target — the greatest sequence number at or before
+the target when version order tracks sequence order, which an early-stop newest-first scan finds.
 
 - **Target between two versions?** The closer, older one. `M-SELECT-01`
 - **Target equal to a version?** That version, an exact match (zero ops to replay). `M-SELECT-02`
-- **Target newer than every version?** The newest recoverable version. `M-SELECT-03`
+- **Target newer than every sealed version?** The newest sealed version. `M-SELECT-03`
 - **Target older than every version?** `noBaseVersion`, reporting the oldest sequence number seen.
   `M-SELECT-04`
 
-### How does it handle duplicate versions and the tip?
+### How does it handle the tip, duplicate versions, and empty history?
 
+- **The tip (index 0)?** Never treated as a base; its sequence number is never even resolved. `M-TIP-01`
+- **Only the tip exists?** `noBaseVersion` — the sole version is excluded as a base. The wired consumer
+  (the point-in-time document service factory) surfaces this as a `UsageError`; loading the live document
+  for a near-head target is a possible future consumer choice, not current behavior. `M-TIP-02`
 - **Two versions share a sequence number?** Return the newest label (a metadata-only re-save leaves the
   sequence number unchanged; the newest is closest to the head). `M-DEDUP-01`
-- **The tip (index 0)?** Never treated as a base; its sequence number is never even resolved.
-  `M-TIP-01`
-- **Only the tip exists?** `noBaseVersion`. `M-TIP-02`
 - **No versions at all?** `noBaseVersion`. `M-EMPTY-01`
 
-### What work does it avoid?
+### What work does it avoid when scanning?
 
-- **Resolving more versions than needed?** It stops at the first version at or before the target and
-  does not resolve older ones. `M-STOP-01`
-- **Re-fetching across calls?** The version list and each resolved sequence number are cached.
-  `M-CACHE-01`
-- **Stale caches?** `refresh()` drops both the version list and the resolved sequence numbers, so the
-  next query re-enumerates and re-resolves. `M-CACHE-02`
-- **A `refresh()` while a fetch is still in flight?** The cache holds the pending fetch rather than its
-  eventual value, so a fetch that started before the refresh cannot write its now-stale result back over
-  the cleared cache; the next query re-fetches. `M-CACHE-03`
+- **Resolving more versions than needed?** It stops at the first version at or before the target and does
+  not resolve older ones. `M-STOP-01`
+
+### What is cached, and what is re-fetched?
+
+Resolved sequence numbers are memoized; the version list is not. A sealed version's sequence number never
+changes, so once resolved it is reused for the manager's lifetime rather than re-fetched. The version
+list, by contrast, changes as new versions are cut, so it is **re-enumerated on every query**. This
+mirrors how Page History (`host-page-history`) works: its `PageHistoryVersionManager` pulls the ODSP
+version list fresh on each navigation (`refreshVersions()`), caching only the expensive loaded *content* —
+not the list. Page History's `#getOdspVersions` likewise dedups and drops the tip (`slice(1)`), the same
+two rules applied here.
+
+The manager is short-lived — the point-in-time document service factory creates one per load and calls
+`findBaseForSeq` once, then drops it — so the memoization caches never grow large and need no eviction.
+
+- **Re-resolving a sequence number across calls?** Each version's number is memoized, so a later query
+  reuses it rather than re-fetching. `M-CACHE-01`
+- **A version-list fetch that fails?** It propagates rather than being read as empty; the list is
+  re-enumerated on the next call. `M-CACHE-02`
 
 ### What happens when a version cannot be resolved?
 
-The failure propagates; it is never swallowed into a wrong base. `M-ERR-01`
+The failure propagates; it is never swallowed into a wrong base. `M-ERR-01` A failed resolution is not
+cached, so a later call re-attempts it rather than replaying the cached rejection. `M-ERR-02`
 
 ### What does `listVersions` return?
 
-Every version with its resolved sequence number, newest-first. `M-LIST-01`
+Every version with its resolved sequence number, newest-first. `M-LIST-01` The tip's number is resolved
+fresh on every call (never cached, since it is not yet stable), while sealed versions are served from the
+cache. `M-LIST-02`
 
 ### How does it verify a chosen base can be replayed to the target?
 
@@ -169,10 +217,9 @@ them.
 - **Either epoch unknown?** Fails closed — without both epochs the shared-lineage claim cannot be proven.
   `M-VALIDATE-03`
 
-A numbered version's snapshot is immutable, so its epoch is read once and cached per versionId; the
+A numbered version's snapshot is immutable, so its epoch is read once and memoized per versionId; the
 live document's epoch can change (a restore or download-and-reupload bumps it) and is therefore read
-fresh on every lineage check, never cached. `refresh()` clears the cached version epochs alongside the
-version list and resolved sequence numbers. `M-VALIDATE-CACHE-01`, `M-VALIDATE-CACHE-02`
+fresh on every lineage check, never cached. `M-VALIDATE-CACHE-01`
 
 **Op availability.** This is _not_ re-checked up front, and Component B adds no check of its own. Op
 retention trims a contiguous _prefix_ from the oldest end of the stream, and op sequence numbers are
@@ -206,8 +253,14 @@ field yields an empty list rather than an error. `F-LIST-03`
   parses the response, and returns `trees[0].sequenceNumber`. `F-RESOLVE-01`
 - **A snapshot with no sequence number?** It throws, naming the version, rather than returning a wrong
   value. `F-RESOLVE-02`
+- **A snapshot whose sequence number is present but not a valid non-negative integer?** It throws rather
+  than coercing a wrong value into base selection. `F-RESOLVE-06`
 - **A binary (`application/ms-fluid`) snapshot?** It reads it with the driver's compact-snapshot parser
   and returns the same sequence number the JSON path would. `F-RESOLVE-03`
+- **An unexpected content-type (e.g. an HTML error page)?** It throws rather than mis-parsing the body as
+  a compact snapshot. `F-RESOLVE-04`
+- **A version label with characters that need escaping?** The label is percent-encoded into the snapshot
+  URL. `F-RESOLVE-05`
 
 ### How does it handle request failures?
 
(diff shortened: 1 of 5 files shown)


packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -96,27 +96,28 @@ export interface IOdspVersionManager {
 export class OdspVersionManager implements IOdspVersionManager {
 	private versionsCache: Promise<OdspFileVersionRef[]> | undefined;
 	private readonly seqByVersion = new Map<string, Promise<number>>();
+	// Bumped by refresh(); lets a rejecting in-flight fetch tell whether it still owns its cache entry.
+	private generation = 0;
```
anthony-murphy: Deep Review: This is the load-bearing change and its correctness rests on an unverified contract. The PR deletes the tip-exclusion #27689 deliberately introduced — `const candidates = versions.slice(1);` with the comment `// Recoverable base candidates = every version except the tip (index 0 ≈ the live document).` — and now scans all versions (`for (const version of versions)`). It also deletes the two tests that pinned the old contract, `M-TIP-01` ("never treats the tip as a recoverable base") and `M-TIP-02` ("noBaseVersion when only the tip exists"), replacing them with cases that assume index 0 is recoverable.

The whole reversal rests on one claim in the description: index 0 being recoverable "came from a live console check against the API," and the PR's own guidance requests "a reviewer with ODSP domain knowledge confirming that contract." No second party has confirmed it, and the prior code embedded the opposite assumption. If driveItem `/versions[0]` is in fact the live/uncommitted document, `findBaseForSeq` will return an unloadable base for any target at or after the head.

Two things close this:
1. Record the live console-check evidence on-record (work item 77401 or the PR description) so the contract isn't justified only by an unrepeatable one-off.
2. Get ODSP domain confirmation (anthony-murphy or an ODSP owner) that driveItem `/versions[0]` resolves to a normal recoverable version — not a synthetic "tip"/live document — before this selection logic is wired into a load path.

If the contract cannot be confirmed, reinstate the tip-exclusion and its `M-TIP-01`/`M-TIP-02` tests.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -96,27 +96,28 @@ export interface IOdspVersionManager {
...
 	public refresh(): void {
 		this.versionsCache = undefined;
 		this.seqByVersion.clear();
+		this.generation++;
 	}
 
 	public async findBaseForSeq(target: number): Promise<BaseForSeq> {
-		// Recoverable base candidates = every version except the tip (index 0 ≈ the live document).
 		const versions = await this.getVersions();
-		const candidates = versions.slice(1);
```
anthony-murphy, line 100: the problem with including the tip version is that it is the only version that isn't static, so it seq can change, so if we use it, we need special handling for it.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -96,27 +96,28 @@ export interface IOdspVersionManager {
...
 	private readonly seqByVersion = new Map<string, Promise<number>>();
+	// Bumped by refresh(); lets a rejecting in-flight fetch tell whether it still owns its cache entry.
+	private generation = 0;
 
 	public constructor(private readonly fetcher: IOdspFileVersionFetcher) {}
 
 	public refresh(): void {
 		this.versionsCache = undefined;
 		this.seqByVersion.clear();
+		this.generation++;
```
anthony-murphy: i'm not sure i really like this approach. can't we just store the fetch promise, and ensure only a single fetch promise is every active. i also think we need to stop clearing seqByVersion, most for the time they will be reusable, so we can only delete entries where the version number no longer exists in the latest list

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -142,18 +143,43 @@ export class OdspVersionManager implements IOdspVersionManager {
...
+			// Cache the pending promise (a version's sequence number never changes) so concurrent callers
+			// coalesce and a refresh() is not clobbered by a fetch already in flight. A rejected resolution
+			// is evicted so a later call retries, unless a refresh() has since advanced the generation.
+			const generation = this.generation;
+			pending = (async (): Promise<number> => {
+				try {
+					return await this.fetcher.resolveSequenceNumber(versionId);
+				} catch (error) {
+					if (this.generation === generation) {
+						this.seqByVersion.delete(versionId);
```
anthony-murphy: this seems wrong. there isn't every a reason to fetch a seq for version we already have a seq of, we should only be fetching seqs when we don't have them

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
lindsnguyen: reversed changes and restored the tests. recoverable does not mean stable,  ran the api via the console which showed the tip is loadable however we can't confirm either way that the seq is static, so it can't be a stable base.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
lindsnguyen: chagned the code to directly use PromiseCache inside core-utils, also no longer caching the version list, just the seq num, aligned with how page history current pulls these apis

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
lindsnguyen: seqCache.addOrGet now so we should only get if we don't already have

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
anthony-murphy: Deep Review: Resolved. You reverted the tip-drop and restored `M-TIP-01`/`M-TIP-02`, and ODSP owner anthony-murphy confirmed on [the tip thread](https://github.com/microsoft/FluidFramework/pull/27695#discussion_r3590126689) that the tip's sequence number isn't static so it can't serve as a stable base — tip-exclusion is the intended, domain-endorsed behavior. One follow-up, tracked in the review summary: the PR description still narrates the removed tip-drop and cites a non-existent `M-SELECT-05`; update it to match the shipped code (paste-ready text is in the summary's Suggested description).

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -131,32 +130,43 @@ export class OdspVersionManager implements IOdspVersionManager {
 
 	public async listVersions(): Promise<ResolvedVersion[]> {
 		const versions = await this.getVersions();
-		// Resolution order does not matter here, so resolve concurrently; the newest-first array order is
+		// Resolution order does not matter, so resolve concurrently; the newest-first array order is
```
anthony-murphy, line 169: Deep Review: The rewrite drops the concurrent-fetch coalescing #27689 added. `findBaseForSeq`/`listVersions` now call `this.fetcher.listFileVersions()` unconditionally, so N concurrent callers issue N overlapping `/versions` enumerations and interleave mutations of the shared `lastKnownVersionIds`/`seqCache`. This is not a wrong-base bug — each caller acts on the concrete list it fetched, and failed seq resolutions auto-evict via `PromiseCache` — but it is extra round-trips plus an untested interleaving edge on the `lastKnownVersionIds` prune/repopulate path.

Related resolved thread: [r3590146089](https://github.com/microsoft/FluidFramework/pull/27695#discussion_r3590146089) — anthony-murphy asked to "store the fetch promise, ensure only a single fetch promise is ever active"; the seq cache moved to `PromiseCache`, but the list fetch itself is no longer coalesced.

Wrap `listFileVersions()` in a single-flight promise (e.g. `this.pendingFetch`) cleared on settle so simultaneous callers join one in-flight enumeration before `lastKnownVersionIds` is updated — preserving per-query freshness for sequential calls while restoring coalescing for concurrent bursts. If concurrent-query volume is provably low, add an explicit "concurrent callers not expected" note in DEV.md instead. Are concurrent `findBaseForSeq`/`listVersions` calls on a single instance expected in practice?

packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
```diff
@@ -91,45 +123,58 @@ these behaviors are tested with an in-memory fake.
...
+
+Resolved sequence numbers are cached; the version list is not. A sealed version's sequence number never
+changes, so once resolved it is kept indefinitely and never re-fetched. The version list, by contrast,
+changes as new versions are cut, so it is **re-enumerated on every query** rather than held in a cache that
+could go stale. This mirrors how Page History (`host-page-history`) works: its `PageHistoryVersionManager`
+pulls the ODSP version list fresh on each navigation (`refreshVersions()`), caching only the expensive
+loaded *content* — not the list. Page History's `#getOdspVersions` likewise dedups and drops the tip
+(`slice(1)`), the same two rules applied here.
+
+- **Re-resolving a sequence number across calls?** Each version's number is cached, so a later query
```
anthony-murphy: Deep Review: `M-CACHE-03` is repurposed here to document the new "version leaves the list" behavior and `M-CACHE-02` is removed, but DEV.md's own convention (lines 7-13) states scenario IDs are append-only and are never renumbered — reusing or removing IDs breaks traceability between the doc and the tests that reference these IDs.

Keep the old IDs (`M-CACHE-02`, `M-CACHE-03`) reserved/deprecated rather than renumbering, and assign new, unused IDs for the added cache behaviors so existing references stay valid — or drop the append-only clause if it is no longer intended.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -89,34 +91,31 @@ export interface IOdspVersionManager {
 }
 
 /**
- * Default {@link IOdspVersionManager}. Caches the version list and resolved sequence numbers. The
+ * Default {@link IOdspVersionManager}. Caches resolved sequence numbers (which never change); the version
+ * list is re-enumerated on each query rather than cached, since new versions are cut over time. The
  * resolution strategy (eager, newest-to-oldest, stopping at the first usable base) is hidden behind
  * {@link findBaseForSeq} and can change without affecting callers.
  */
```
anthony-murphy, line 90: Deep Review: In #27689 anthony-murphy asked to export a factory rather than the class — "you should export a function which returns the interface and internally constructs the class. this keeps the implementation off our exposed api surface." The follow-up added `createOdspVersionManager` returning `IOdspVersionManager`, but `export class OdspVersionManager` remains — because tests construct it directly with a mock fetcher (`new OdspVersionManager(fetcher)` at test/odspVersionManager.spec.ts:59, 214, 249), so a naive "just drop `export`" would break the suite. The class never reaches the folder barrel (`odspVersionManager/index.ts`) nor the package public index, so this is narrow, not a public-API leak.

Add a test-injection factory overload accepting `IOdspFileVersionFetcher` so tests inject the mock without importing the class, then drop `export`; or add a one-line comment documenting that the class export is retained deliberately for test injection.

anthony-murphy:
`packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts:220`

Deep Review: The new `else` branch throws a plain `new Error(...)` naming the version and content-type. The sibling snapshot path records `contentType`/`accept` in `propsToLog` (fetchSnapshot.ts:372-380) then throws `new NonRetryableError("Unknown snapshot content type", OdspErrorTypes.incorrectServerResponse, propsToLog)` (fetchSnapshot.ts:502-507), and `odspUtils.ts:103-119` treats `incorrectServerResponse` specially in the token-refresh retry wrapper — so typed classification has observable behavior beyond the message text. A generic `Error` won't be classified as non-retryable and carries no structured telemetry.

Swap the plain `Error` for a `NonRetryableError` with `OdspErrorTypes.incorrectServerResponse` and `contentType`/`accept` telemetry, mirroring `fetchSnapshot.ts`.

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
```diff
@@ -121,15 +124,29 @@ export function createOdspFileVersionFetcher(
 				const snapshotJson = (await response.content.json()) as IOdspSnapshot;
 				sequenceNumber =
 					convertOdspSnapshotToSnapshotTreeAndBlobs(snapshotJson).sequenceNumber;
-			} else {
+			} else if (contentType.includes("application/ms-fluid")) {
 				// ms-fluid framing: the compact binary form; read it with the driver's compact-snapshot parser.
 				const bytes = new Uint8Array(await response.content.arrayBuffer());
 				sequenceNumber = parseCompactSnapshotResponse(bytes, logger).sequenceNumber;
+			} else {
```
anthony-murphy, line 169: Deep Review: The comment justifying the `NonRetryableError` throw states the wrapper "classifies it as incorrectServerResponse ... rather than a generic Error, which would be treated as retryable." That inverts the actual behavior. `getWithRetryForTokenRefresh` (odspUtils.ts:115-118) *retries* `incorrectServerResponse` once — `case OdspErrorTypes.incorrectServerResponse: return get(options);` — while its `default:` branch does `throw error;`, so a generic `Error` is NOT retried.

The PR's own tests confirm this: odspFileVersionFetcher.spec.ts:430/451 note the typed error "is retried once by getWithRetryForTokenRefresh, so two responses are stubbed" — directly contradicting the production comment.

Reword the comment to say the typed `incorrectServerResponse` classification triggers a single retry (to absorb a transient wire glitch) and carries structured telemetry — not that it makes the error non-retryable, and not that a generic `Error` would be retried. Apply the same correction to the missing-`sequenceNumber` throw comment at lines 144-149.

anthony-murphy:
`packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts:88`

Deep Review: `findBaseForSeq` returns a single `{ kind: "noBaseVersion", oldestResolvedSeq? }` for two outcomes that demand opposite consumer responses: (a) the file has no sealed version (only the tip → consumer should load the live file) and (b) sealed versions exist but all are newer than the target (predates retained history → consumer should error).

The two doc comments describe `noBaseVersion` inconsistently: the `BaseForSeq` `@remarks` says "if the file's only version is the tip, the result is `noBaseVersion` and a consumer loads the live file instead," while this `IOdspVersionManager.findBaseForSeq` doc says it means "the target predates the oldest retained version."

The discriminant already exists at runtime — `oldestResolvedSeq` is `undefined` iff no sealed candidate resolved (case a) and defined otherwise (case b), and both branches are tested (odspVersionManager.spec.ts:99-104 asserts `oldestResolvedSeq === 418`; 127-132 covers the only-tip case) — but it is never documented as the discriminant.

Document `oldestResolvedSeq` as the explicit discriminant (`undefined` ⇒ no sealed version ⇒ load-live; defined ⇒ predates retained history) and resolve the contradiction between the two `@remarks`. No code change is required unless maintainers confirm live fallback is the mandated design.

packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
```diff
@@ -41,6 +42,38 @@ Part 1 is built in three components:
...
+  returned with the snapshot for free — not a separate fetch.
+- **sequence number** = an op's global order index (a snapshot's is the op its tree is current through);
+  **latestSequenceNumber** = the last bundled trailing op; **minimumSequenceNumber (MSN)** = the
+  collaboration-window floor (seq all connected clients have acked) — **not** a retention/trimming signal.
+
+Chain of custody for one saved state: a **summarizer** writes a **summary** → stored as a **snapshot** →
+surfaced by ODSP as a **file version**. Same state, three names because three layers own it.
+
+
 ## Part I — Foundations
```
anthony-murphy, line 78: Deep Review: `M-TIP-02` and the `BaseForSeq` `@remarks` assert a present-tense "loads the live document instead" fallback, but the only merged consumer (#27703, `odspPointInTimeDocumentServiceFactory.ts:82-89`) throws `UsageError` on `noBaseVersion`. DEV.md states it "reflects the code as it is now" (DEV.md:16-22), so this present-tense fallback contradicts the shipped consumer.

Align `M-TIP-02` and the `BaseForSeq` remark with current behavior — describe the `UsageError` the wired consumer raises and mark live-file fallback as directional/future, not present-tense. If maintainers confirm live fallback is the mandated design, change `odspPointInTimeDocumentServiceFactory.ts:82` instead.

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
anthony-murphy: Deep Review: The inverted production comment is corrected — thanks. One residual polish item on the same code: `NonRetryableError` is still retried once. `getWithRetryForTokenRefresh` (odspUtils.ts:115-119) routes `incorrectServerResponse` into the retry-once branch, and tests `F-RESOLVE-04`/`F-RESOLVE-02` stub two responses to confirm it — so a class named "NonRetryable" is in fact retried, and for a deterministic failure (a snapshot structurally missing `sequenceNumber`, or a persistent HTML error page) the retry doubles the fetch/parse work.

Keep the retry — it matches driver-wide `incorrectServerResponse` semantics — but add a short comment at both throw sites (odspFileVersionFetcher.ts:137-141 and :147-151) noting the typed `incorrectServerResponse` classification intentionally triggers a single retry to absorb a transient wire glitch. That resolves the name/behavior contradiction for a future maintainer editing the retry switch.

packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
```diff
@@ -97,45 +130,62 @@ these behaviors are tested with an in-memory fake.
...
+  (single-flight), so a burst issues one fetch and one prune rather than overlapping fetches; a later
+  sequential call re-enumerates. `M-CACHE-06`
 
 ### What happens when a version cannot be resolved?
 
-The failure propagates; it is never swallowed into a wrong base. `M-ERR-01`
+The failure propagates; it is never swallowed into a wrong base. `M-ERR-01` A failed resolution is not
+cached, so a later call re-attempts it rather than replaying the cached rejection. `M-ERR-02`
 
 ### What does `listVersions` return?
```
anthony-murphy, line 193: Deep Review: This line adds `Ops in the standalone op log are retained for a window (time-based, believed ~30 days — unverified, confirm with ODSP)` — a self-labeled placeholder in developer docs. DEV.md's own contract (DEV.md:16-22) states the document reflects the current code and that future/aspirational content belongs in Part IV, so an unverified operational fact shouldn't sit inline.

Confirm the retention window with ODSP and cite the source, or drop the specific `~30 days` figure and describe the retention behavior qualitatively until confirmed.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -131,32 +137,57 @@ export class OdspVersionManager implements IOdspVersionManager {
...
+		const versions = await this.fetcher.listFileVersions();
+		const live = new Set(versions.map((version) => version.versionId));
+		for (const versionId of this.lastKnownVersionIds) {
+			if (!live.has(versionId)) {
+				this.seqCache.remove(versionId);
+			}
 		}
-		return pending;
+		this.lastKnownVersionIds = versions.map((version) => version.versionId);
+		return versions;
```
anthony-murphy: Deep Review: `reconcileVersions()` prunes departed version IDs from `seqCache` after re-enumeration — for each ID in `lastKnownVersionIds` no longer present in the fresh list, it calls `this.seqCache.remove(versionId)`. That prune is only load-bearing if a version ID can leave the list and later re-appear bound to a *different* sequence number; the PR's own "a sealed version's number is fixed" comments suggest that can't happen for sealed ODSP versions. Reviewers split on whether this is necessary memory hygiene or over-engineering for an internal, unwired component.

Add a one-line comment stating the intent: if it's memory hygiene (bounding cache growth as versions age out), say so; if it defends against a version ID re-appearing with a different seq, say that — otherwise the next reader infers a stronger invariant than intended. If seq-reappearance is impossible for sealed versions, consider dropping the prune. Related resolved thread on the same query path: [r3616571973](https://github.com/microsoft/FluidFramework/pull/27695#discussion_r3616571973).

anthony-murphy: Deep Review: One further angle on this same prune, independent of whether it stays for memory hygiene. `reconcileVersions()` only evicts ids present in `this.lastKnownVersionIds` that are no longer live, then overwrites `lastKnownVersionIds`. `findBaseForSeq` scans candidates sequentially with `await this.resolveSeq(...)` per version, repopulating `seqCache` as it goes. Consider: query A enumerates `[tip,43,42,40]` and is mid-scan; query B re-enumerates `[tip,43,42]` (40 aged out), evicts `40`, and sets `lastKnownVersionIds=[tip,43,42]`; query A then reaches `40` (still in its captured list) and re-populates `seqCache` for `40`. Because `lastKnownVersionIds` no longer contains `40`, no future `reconcileVersions` will ever evict it, and `PromiseCache` (default `{ policy: "indefinite" }`) exposes `remove(key)` but no key enumeration — so that entry becomes permanently unreachable to eviction.

No correctness impact: the cached value is the correct immutable sealed seq. The only cost is the leaked entry. If concurrent `findBaseForSeq`/`listVersions` calls on a single instance are serialized upstream, this race can't occur and it closes with no code change — worth confirming and noting. Otherwise, prune `seqCache` against the current live set for all reachable keys, or switch to a bounded/expiring cache policy.

packages/drivers/odsp-driver/src/odspVersionManager/DEV.md
anthony-murphy: Deep Review: Resolved — the `~30 days` placeholder is now documented as `~7 days` (confirmed with the ODSP storage team) in DEV.md, so the unverified operational figure no longer sits inline. This thread can be marked resolved.

anthony-murphy:
## Deep Review

*Reviewed commit `f0d5753` on 2026-08-03.*

Not ready for sign-off, but close. This internal-only follow-up to #27689 is in strong shape — no blocking correctness issues survive. What remains is contained documentation polish: a stale PR-description premise, the DEV.md ID-renumbering decision, and one no-correctness cache-eviction race flagged inline.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Update the PR description: drop "not wired into a call site / no consumer-facing change" and note that merged consumer #27703 (`odspPointInTimeDocumentServiceFactory.ts`) already calls `findBaseForSeq` and surfaces `noBaseVersion` as a `UsageError` — so the cache/error-handling changes are observable to that point-in-time-load consumer. Paste-ready text below.

### Context for Reviewers

- Rejected-resolution caching now uses `PromiseCache` (evicts rejected entries by default), closing anthony-murphy's #27689 permanently-cached-rejection follow-up — pinned by `M-ERR-02` / `M-CACHE-04`.
- Version epochs moved from the `epochByVersion` map to `PromiseCache`, delivering lindsnguyen's committed #27748 follow-up; the live-document epoch stays uncached to preserve the immutable-vs-mutable distinction.
- `refresh()` is intentionally removed: the mutable version list is re-enumerated per call and the live epoch re-read each validation, so there is no stale state to invalidate.
- Tip-drop (`versions.slice(1)`, index-1 scan start) is reviewer-mandated and owner-confirmed on [the tip thread](https://github.com/microsoft/FluidFramework/pull/27695#discussion_r3590126689): the tip is the one version whose sequence number isn't yet static, so it can't be a stable base. An earlier iteration removed it and was reverted — do not re-propose scanning index 0.
- New, single-author, low-churn prototype area. Suggested reviewers: **anthony-murphy** (tip-exclusion, caching, API-surface direction), **sonalideshpandemsft** (authored consumer #27703; can confirm the `noBaseVersion`→`UsageError` outcome), **lindsnguyen** (design owner; authored the reversed DEV.md ID contract).

- **Needs human judgment** — whether the `reconcileVersions()` departed-version `seqCache` prune should remain (memory hygiene / ID-reappearance defense) or be simplified; a solution-space evaluator rated the leaner design stronger. Base-selection owner (anthony-murphy) should ratify. Flagged inline at [r3692804141](https://github.com/microsoft/FluidFramework/pull/27695#discussion_r3692804141) and [r3706587281](https://github.com/microsoft/FluidFramework/pull/27695#discussion_r3706587281) (a concurrent-scan race can leak one unreachable-to-eviction `seqCache` entry — no correctness impact).
- **Needs human judgment** — whether `noBaseVersion` should carry an explicit `kind`/discriminant now or defer until the consumer is fully wired (anthony-murphy and sonalideshpandemsft, author of #27703). The runtime `oldestResolvedSeq` discriminant (`undefined` ⇒ no sealed version ⇒ load-live; defined ⇒ predates history) already distinguishes the two cases.
- **Needs human judgment** — whether per-query version-list re-enumeration (no TTL) is acceptable network cost once the loader is fully wired. anthony-murphy owns the earlier #27689 TTL guidance.
- **Cannot be assessed by the pipeline** — whether the content-type allowlist (`application/json`, `application/ms-fluid`) is exhaustive for what ODSP returns in production, and whether `/versions[0]` on deep histories always resolves as the tip. Verify against the live service.

</details>

### Suggested description

```markdown
## Description

Follow-up to #27689 addressing review feedback on the internal ODSP file-version manager in
`@fluidframework/odsp-driver`. The manager is not exported from the package `index.ts`. It is,
however, already consumed by merged PR #27703 (`odspPointInTimeDocumentServiceFactory.ts`), which
calls `findBaseForSeq` and surfaces a `noBaseVersion` result as a `UsageError` — so the behavioral
changes below are observable to that point-in-time-load consumer, even though nothing new is exported
from this package.

**Base selection — tip exclusion retained.** An earlier iteration removed the tip-drop
(`versions.slice(1)`) and scanned all versions. Per review feedback (anthony-murphy, owner-confirmed),
that was reverted: the tip (newest version, index 0) is the one version whose sequence number is not
yet static, so it cannot serve as a stable base. `findBaseForSeq` scans the sealed versions (index 1+)
newest-first and returns the closest one with `seq <= target`. A file whose only version is the tip
returns `noBaseVersion`; the wired consumer (#27703) surfaces that as a `UsageError`
(`M-TIP-01`/`M-TIP-02`). The two `noBaseVersion` sub-cases are distinguished at runtime by
`oldestResolvedSeq` (`undefined` ⇒ no sealed version ⇒ load-live; defined ⇒ target predates retained
history).

**Robustness — failed fetches.** `refresh()` is removed. The version list is re-enumerated on every
query (it changes as new versions are cut), pruning cached sequence numbers for versions that have left
the list (`lastKnownVersionIds`). Resolved sequence numbers and per-version epochs are cached via
`PromiseCache`, which evicts rejected entries by default, so a transient 5xx/429 no longer memoizes a
permanent failure (`M-ERR-02`, `M-CACHE-04`).

**Robustness — content-type.** `resolveSequenceNumber` now handles `application/json` and
`application/ms-fluid` explicitly and throws a typed `NonRetryableError`
(`incorrectServerResponse`, retried once by the shared token-refresh wrapper) on anything else, rather
than mis-parsing (`F-RESOLVE-04`). It also rejects a missing/non-integer/negative sequence number.

**Docs & tests.** Version labels are percent-encoded into the snapshot URL (`F-RESOLVE-05`). `DEV.md`
and the manager/fetcher specs are updated to match. Note: the DEV.md scenario-ID convention is changed
to gap-free renumbering (removed IDs are not reserved) — historical references to specific `M-*`/`F-*`
IDs are no longer guaranteed stable.

No changeset (no public API change).
```

- `c9e8090` 2026-08-03 · **8/10** — op-retention window confirmed (~7 days); stale description, DEV.md ID decision, and cache-eviction race open
- `7e0ad41` 2026-07-31 · **8/10** — strong shape, no blocking issues; stale description, op-retention figure, and reconcileVersions design call open
- `e2faf66` 2026-07-31 · **8/10** — every prior inline thread resolved; three documentation-polish items remained
- `0a416e9` 2026-07-30 · **8/10** — internal-only follow-up in strong shape; documentation-level items flagged inline
- `0eece2d` 2026-07-20 · **8/10** — base-selection settled and owner-confirmed; four contained polish items flagged inline
- `4bd31d9` 2026-07-17 · **6/10** — base-selection concern resolved; PR description still narrated the abandoned iteration
- `7003041` 2026-07-15 · **5/10** — base-selection reversal contract unconfirmed; rejected-promise eviction and content-type hardening resolved

</details>

sonalideshpandemsft (approved):

PR #27696, opened 2026-07-15
feat(tree): option to retain history

yann-achard-MS:
## Description

Adds a new `retainHistory` flag to `SharedTreeOptions` (defaults to `false`).
Setting `retainHistory` to `true` will prevent SharedTree from garbage-collecting historical data about old changes.

Note that this will cause unbounded growth both in memory on the client and in summaries/snapshots (the at-rest data representing a Fluid document). For these reasons, this option is only intended for debugging and experimentation.

This option is needed by an FF customer that is experimenting with history-management features.

## Breaking Changes

None


.changeset/swift-buses-hang.md
```diff
@@ -0,0 +1,11 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": tree
+---
+Retain history option
+
+Adds a new `retainHistory` flag to `SharedTreeOptions` (defaults to `false`).
```
Josmithr: Nit
```suggestion
Adds a new `retainHistory` flag to [`SharedTreeOptions`](https://fluidframework.com/docs/api/tree/sharedtreeoptions-interface) (defaults to `false`).
```

packages/dds/tree/src/shared-tree-core/sharedTreeCore.ts
```diff
@@ -78,6 +78,11 @@ export interface SharedTreeCoreOptionsInternal extends CodecWriteOptions {
 	 * See {@link SharedTreeOptionsBeta.healUnresolvableIdentifiersOnDecode}.
 	 */
 	readonly healUnresolvableIdentifiersOnDecode?: boolean;
+
+	/**
+	 * See {@link SharedTreeOptions.retainHistory}.
```
Josmithr: Nit: would this be better formatted as 
```suggestion
	 * {@inheritDoc SharedTreeOptions.retainHistory}.
```
?

packages/dds/tree/src/shared-tree/sharedTree.ts
```diff
@@ -639,6 +639,20 @@ export interface SharedTreeOptions
...
+
+	/**
+	 * When `true`, prevents trunk commits from being trimmed/evicted, even after they fall outside
+	 * the collaboration window.
+	 *
+	 * @defaultValue `false`
+	 *
+	 * @remarks
+	 * By default, SharedTree evicts trunk commits once all peers have acknowledged them (i.e. once they
+	 * are outside the collaboration window) and they are not otherwise retained (e.g. by revertibles or
```
Josmithr: Nit
```suggestion
	 * are outside the collaboration window), and they are not otherwise retained (e.g. by revertibles or
```

Josmithr (approved):

PR #27697, opened 2026-07-15
arrayNode anchor tracking using deltas

daesunp:
## Description

Adds anchor tracking to arrayNodes using its deltas. This can be used in our collab text apis utilities to leverage this


packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -1507,9 +1536,28 @@ export interface ArrayPlaceAnchor {
...
+ * place. This keeps the anchor pinned to the gap between children even when the child originally at its index is
+ * removed, which the previous child-tracking implementation could not do (it jumped to the end of the array).
+ *
+ * The best-effort behavior noted in the public remarks is the no-delta path: when the composed delta is unavailable
+ * (see {@link NodeChangedDataDelta.delta}), the exact shift is unknown so the index is only clamped back into range.
+ * This is intentionally weak rather than reintroducing a per-edit O(n) snapshot of the array to diff old-vs-new
+ * (which would defeat the point of delta-based tracking) for a case that is rare and, for this alpha API, acceptable
+ * to degrade. A stronger fallback could instead let a consumer that already maintains the array contents re-seed the
+ * anchor's position on this path, so it lives with the consumer that has the data instead of being duplicated here.
+ *
```
daesunp, line 1720: I wasn't sure if updating the fallback pattern of the collab text apis was the right call here (for the undefined deltas case), but I can update this if we decide that that's the better pattern

packages/common/core-utils/src/test/math.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { strict as assert } from "node:assert";
+
+import { clamp } from "@fluidframework/core-utils/internal";
+
+describe("clamp()", () => {
+	const tests: [value: number, min: number, max: number, expected: number][] = [
```
Josmithr: Instead of using an array-based tuple for the value, min, etc., it would probably be better to use an object layout with real property keys. Will make the tests much easier to read.

packages/common/core-utils/src/math.ts
```diff
@@ -0,0 +1,13 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Clamps `value` to the inclusive range [`min`, `max`].
+ * @remarks Assumes `min <= max`.
```
Josmithr: We should probably assert this.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -1485,11 +1485,28 @@ export interface ArrayPlaceAnchor {
...
 	 * This value is updated as the array is edited in a way that depends on the specific anchor implementation.
-	 * This index may take on a value from 0 to the length of the array (inclusive).
+	 * While the anchor is being tracked (before {@link ArrayPlaceAnchor.dispose} is called), this index is a value
+	 * from 0 to the current length of the array (inclusive).
 	 * If used as the index to insert content into the array, this means it can point to any location in the array,
 	 * including just after the last child.
+	 *
+	 * After {@link ArrayPlaceAnchor.dispose} has been called the anchor stops updating and this returns the last
+	 * tracked value, which may no longer be within the array's bounds if the array has since shrunk. Do not use the
+	 * index of a disposed anchor as an insertion index without first re-validating it against the current length.
```
Josmithr: Is there any reason to support users calling `index` after the anchor has been disposed? We could just throw an exception in this case, or leverage our `breakable` infrastructure to accomplish this.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -1485,11 +1485,28 @@ export interface ArrayPlaceAnchor {
...
 	 */
 	get index(): number;
+
+	/**
+	 * Stop tracking this anchor and release any resources (such as event subscriptions) it holds.
+	 * @remarks
+	 * An anchor subscribes to changes on the array node in order to keep its {@link ArrayPlaceAnchor.index}
```
Josmithr: Nit: I don't think we need to note that the anchor leverages our eventing system. That's an implementation detail, and not really something users need to know about.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -1507,9 +1524,28 @@ export interface ArrayPlaceAnchor {
...
+ * it keeps reporting a valid in-bounds index, but that index may no longer correspond to the same logical position
+ * as before the change (this can happen when an incremental delta is unavailable for a change, such as during a
+ * schema change). Consumers that require an accurate position across such changes should re-derive it from their
+ * own state rather than relying solely on this anchor.
  * @privateRemarks
+ * The index is maintained incrementally from the shallow (insert/remove) delta delivered by the array node's
+ * `childrenChangedAfterBatch` event rather than by re-reading a tracked child's position: inserts and removes
+ * before the anchor point shift it, while edits after it (or a removal of the span it sits between) leave it in
+ * place. This keeps the anchor pinned to the gap between children even when the child originally at its index is
+ * removed, which the previous child-tracking implementation could not do (it jumped to the end of the array).
```
Josmithr: Let's avoid notes about any previous implementations. Those aren't relevant to anyone reading the code.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
Josmithr: This seems fine to me for now. But we should probably add an explicit TODO about fixing the underlying edge-cases that cause this. We should really have similar TODOs in the other places that face limitations based on this as well. We're going to need to fix that eventually, and it would be useful to know just how many places are limited by it, so we can better prioritize.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -1519,17 +1555,79 @@ export function createArrayInsertionAnchor(
 	node: TreeArrayNode,
 	currentIndex: number,
 ): ArrayPlaceAnchor {
-	const field = getInnerNode(node).getBoxed(EmptyKey);
-	const child = field.boxedAt(currentIndex);
+	// Clamp the starting index into the valid inclusive range so the anchor always reports a location
+	// which is valid to insert content at, even if a caller provides an out of range index.
+	let trackedIndex = clamp(currentIndex, 0, getInnerNode(node).getBoxed(EmptyKey).length);
```
Josmithr: We shouldn't modify the user's input position from the outset. We should instead throw a `UsageError` if the initial index is invalid (which I believe `boxedAt` did for us previously).

Josmithr (commented):
I left a few comments. Overall direction looks good!

Question: can we update our selection-tracking utilities in the react package to leverage this? Seems like we should be able to remove some code over there.

packages/common/core-utils/src/test/math.spec.ts
daesunp: Updated with suggestions

packages/common/core-utils/src/math.ts
daesunp: Updated with assert

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
daesunp: Updated with usage error when anchor is disposed

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
daesunp: Removed eventing/subscription wording from the public docs

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
daesunp: removed references to old behavior

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
daesunp: replaced with `validateIndex` which throws usage error on out of range inputs

daesunp:
I made some unpushed changes that updated our selection tracking utilities, and it cleans up nicely :) Did you want me to include those changes in this PR, or in a follow up?

Josmithr:
> I made some unpushed changes that updated our selection tracking utilities, and it cleans up nicely :) Did you want me to include those changes in this PR, or in a follow up?

Either way is fine by me. Whatever is easiest.

.changeset/red-paws-cut.md
```diff
@@ -0,0 +1,18 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+Array insertion anchors now track their index from change deltas
+
+The `@alpha` `ArrayPlaceAnchor` returned by `createArrayInsertionAnchor` now maintains its `index` incrementally from the array node's change delta instead of re-deriving it from the child that happened to sit at the anchor point when it was created. Inserts and removes before the anchor shift it, while edits after it leave it in place.
```
Josmithr: Nit: let's link to the API docs where appropriate here, since these APIs already exist.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -1485,11 +1485,20 @@ export interface ArrayPlaceAnchor {
...
 	 * including just after the last child.
+	 * @throws A {@link @fluidframework/telemetry-utils#UsageError} if the anchor has been {@link ArrayPlaceAnchor.dispose | disposed}.
 	 */
 	get index(): number;
+
+	/**
+	 * Stop tracking this anchor and release any resources it holds.
+	 * @remarks
+	 * Call this when the anchor is no longer needed (for example when a tracked cursor position is discarded).
+	 * Reading {@link ArrayPlaceAnchor.index} after disposal throws. Calling `dispose` more than once has no effect.
```
Josmithr: To make this a bit more future-proof, it might be better to say something like "Interacting with an anchor (including reading its properties) after it has been disposed is invalid and will throw."

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -1519,17 +1548,84 @@ export function createArrayInsertionAnchor(
 	node: TreeArrayNode,
 	currentIndex: number,
 ): ArrayPlaceAnchor {
+	// Validate the caller-provided index rather than silently correcting it: an out-of-range or non-integer
```
Josmithr: Nit: I don't think we need to note the alternative option of silently correcting. This is another example of (subtly) referencing previous implementation details in a way that is more likely to be confusing than helpful, in my opinion.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -1519,17 +1548,84 @@ export function createArrayInsertionAnchor(
...
+ * Compute the new index of an {@link ArrayPlaceAnchor} after applying a shallow array delta.
+ * @remarks
+ * The marks cover the array's sequence field in order. `readPosition` walks the pre-edit array as the marks are
+ * consumed while `index` stays fixed at the anchor's pre-edit location; comparing the two (both in pre-edit
+ * coordinates) shifts `newIndex` to keep the anchor in the same gap between children as content is inserted or
+ * removed around it. Each mark is a removal (`detach`), an insertion (`attach`), or a retain (neither); a mark may
+ * carry both `detach` and `attach`, which is handled as a removal followed by an insertion. The per-mark behavior is
+ * commented inline below.
+ *
+ * This mirrors the cursor-tracking accounting used for text selections.
```
Josmithr: And presumably the latter will go away (and just leverage this code) in a future PR? If so, might be best to just remove this line (or be sure to remember to remove it in the follow-up PR).

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -1146,13 +1146,71 @@ describe("ArrayNode", () => {
 			assert.equal(anchor.index, 2);
 		});
 
-		// This case sticks to the end of the array, which is not ideal, and will need to be fixed with a more sophisticated anchor implementation.
+		// With delta-based tracking the anchor stays in the gap the removed item occupied rather than
```
Josmithr: This comment appears to be referencing the old (unwanted) behavior. Let's just remove it.

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -1146,13 +1146,71 @@ describe("ArrayNode", () => {
...
+		// jumping to the end of the array.
 		it("removed item", () => {
 			const array = init(CustomizableNumberArray, [1, 2, 3]);
 			const anchor = createArrayInsertionAnchor(array, 1);
 			array.removeAt(1);
-			// It's good to test that this still gives a valid index and does not crash, but ideally this would anchor to the range between items rather than jumping to the end.
-			assert.equal(anchor.index, 2);
+			// The item originally at the anchor point is gone; the anchor rests between the surviving
+			// neighbors (now [1, 3]) instead of jumping to the end.
```
Josmithr: Let's remove the "instead of jumping to the end" - this is a reference to previous behavior, which is more likely to be confusing than helpful.

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -1146,13 +1146,71 @@ describe("ArrayNode", () => {
...
+
+		it("shifts left when a multi-element range before the anchor is removed", () => {
+			const array = init(CustomizableNumberArray, [10, 20, 30, 40, 50]);
+			const anchor = createArrayInsertionAnchor(array, 4); // gap between 40 and 50
+			array.removeRange(0, 2); // remove 10, 20 -> [30, 40, 50]
+			assert.equal(anchor.index, 2); // still in the gap between 40 and 50
+			anchor.dispose();
+		});
+
+		it("collapses to the start when a removed range straddles the anchor", () => {
```
Josmithr: Nit: it isn't immediately clear to me what "straddles the anchor" means. Might be worth updating the wording here to be a bit more explicit.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
Josmithr: Did you take a look at our `breakable` infra? I think this is the sort of use case it was built for.

dannimad (commented):

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
daesunp: removed the "rather than silently correcting it" phrasing from the comment

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
daesunp: Yes, will update it in a follow up PR, and removed the "This mirror the cursor-tracking..." comment

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
daesunp: renamed the test to "collapses to the removal start when the removed range contains the anchor's position"

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
daesunp: removed comment

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -1146,13 +1146,70 @@ describe("ArrayNode", () => {
...
+			anchor.dispose();
+		});
+
+		it("throws on an out-of-range initial index", () => {
+			const array = init(CustomizableNumberArray, [1, 2, 3]);
+			// One past the end is valid (an insertion point after the last child)...
+			const anchor = createArrayInsertionAnchor(array, 3);
+			assert.equal(anchor.index, 3);
+			anchor.dispose();
+			// ...but beyond that, or negative, is a usage error rather than being silently corrected.
```
Josmithr:
```suggestion
			// Anything beyond that, or negative, is a usage error.
```

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
Josmithr: @daesunp Ping

dannimad:

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
daesunp: Sorry, forgot to update this thread!

I tried using the `breakable` infra, but `breakable.break(Error)` would actually throw when called, so `dispose()` called twice would also throw (instead of having no effect). We could swallow the errors from `breakable.break()` by wrapping it in a try catch statement, but the boolean/usage error pattern seemed a lot cleaner.

`SchematizingSimpleTreeView` also seems to use a similar pattern for disposal where they keep track of a `disposed` flag, and in `ensureUndisposed()` they have:
```
if(this.disposed){
   this.failDisposed()
}
```
where `failDisposed()` just throws a usage error.

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
daesunp: Updated with suggestion

Josmithr (approved):

PR #27699, opened 2026-07-15
fix(ci): configure pnpm to use authenticated registry

MarioJGMsoft:
## Description

The client pipeline already configures an artifct for npm. However, pnpm 11 does not automatically adopt that configuration through `NPM_CONFIG_USERCONFIG`.

Set `PNPM_CONFIG_USERCONFIG` so pnpm 11 adopts the pipeline's existing configuration. This ensures pnpm registry commands, including registry-backed version utility tests, use the appropriate configuration
## Reviewer Guidance

Fixes [AB#73158](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73158)


tools/pipelines/templates/include-install-pnpm.yml
```diff
@@ -97,6 +97,7 @@ steps:
       set -eu -o pipefail
       echo "Using node $(node --version)"
       echo "##vso[task.setvariable variable=NPM_CONFIG_USERCONFIG]$NPM_CONFIG_USERCONFIG"
+      echo "##vso[task.setvariable variable=PNPM_CONFIG_USERCONFIG]$NPM_CONFIG_USERCONFIG"
```
CraigMacomber, line 100: I don't think this belongs here.

I see two reasons:

1. This part of this pipeline is setting up npm so we can use npm to install pnpm. See the "Install and configure pnpm" section for where it would go.
2. I'm pretty sure pnpm is picking up the existing npm setting already, since it wasn't, our `pnpm install` steps would already be failing to run under networks isolation. Maybe I'm missing something, but I think we have had such enforcement for a long time now, and this has been working as is.

CraigMacomber: Another concerning thing looking at this change is we also set NPM_CONFIG_USERCONFIG in tools/pipelines/templates/include-setup-npmrc-for-download.yml : we should not be doing this in multiple places. If a change is need to this copy of the logic, one is likely needed there as well: we should deduplicate them first.

Also note there is additional logic to plumb NPM_CONFIG_USERCONFIG into our docker images, see its use in NPM_CONFIG_USERCONFIG 

If there is some other special value we need for this for pnpm reasons, it likely would have to do something inside the docker image to handle it as well.

CraigMacomber, line 100: Where is PNPM_CONFIG_USERCONFIG documented to be a thing pnpm looks at? I don't see it in its docs: https://pnpm.io/npmrc#auth-file-locations

CraigMacomber: Those docs do state `The npm-style NPM_CONFIG_USERCONFIG variable is also honored as a fallback.` which implies  NPM_CONFIG_USERCONFIG  should just work (like think it has been), but nothing in there implies PNPM_CONFIG_USERCONFIG is a thing that has any impact.

PNPM_CONFIG_USERCONFIG is mentioned in the implementation of pnpm, just not in their docs (which is annoying): 

https://github.com/search?q=repo%3Apnpm%2Fpnpm%20PNPM_CONFIG_USERCONFIG%20&type=code


The docs in https://github.com/pnpm/pnpm/blob/fb4639f8d2885da64b79cf97d385af97004b65b4/pnpm/crates/config/src/lib.rs#L1125-L1126 imply `PNPM_CONFIG_NPMRC_AUTH_FILE` / `PNPM_CONFIG_USERCONFIG` / `npm_config_userconfig` all do the same thing though thats from a future version of pnpm.

Anyway, I conclude I don't have enough information to know what's going on.

MarioJGMsoft:
The fix wasn't right.

PR #27702, opened 2026-07-16
fix(property-changeset): don't throw on a duplicate identical insert into an indexed collection

arafat-java:
## Description

`ChangeSetIndexedCollectionFunctions`'s merge logic throws `CS-003: Internal error: Added an already existing entry: <key>` whenever an incoming insert's key already has an insert recorded in the base ChangeSet — even when the two inserts are identical (e.g. from a duplicated or replayed op). Since this runs on the normal op-processing path (`SharedPropertyTree._applyRemoteChangeSet` → `processMessage`), a duplicate op can put a document's summarizer into a crash-loop.

This PR treats an insert that exactly matches the already-recorded insert for the same key as a safe no-op, for both primitive and typed/polymorphic collections. A mismatched value for the same key is still treated as a genuine conflict and continues to throw.

To reproduce: apply a ChangeSet insert for a key to a base ChangeSet that already has an identical insert for that same key (see the new test cases in `indexedCollection.spec.ts` for a minimal repro). Before this change that throws; after, it's a no-op.

## Reviewer Guidance

I haven't been able to run this package's own test suite locally yet (environment setup issue unrelated to this change — Azure DevOps Artifacts feed auth). The two new test cases closely follow this file's existing patterns, and the equivalent fix (applied to the compiled/patched form of this exact function) has been running against a real local Fluid server with 113 passing tests in a downstream consumer — but I'd like CI/a reviewer to confirm this package's own suite passes before merging. Opening as a draft until I can confirm locally.

### DIFF
--- a/.changeset/fix-duplicate-insert-indexed-collection.md
+++ b/.changeset/fix-duplicate-insert-indexed-collection.md
@@ -1,0 +1,10 @@
+---
+"@fluid-experimental/property-changeset": patch
+"__section": fix
+---
+
+Crash-looping from duplicate inserts in indexed collection ChangeSets has been fixed
+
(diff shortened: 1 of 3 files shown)


dannimad (changes requested):
Do not merge until the release process is complete

arafat-java:
> Do not merge until the release process is complete

@dannimad When is the release process complete date
If it is gone past that then can we get this approved and merged

PR #24855, opened 2025-06-16
(compat) Added supported features and generation across Loader / Driver boundary - Part 2

agarwal-navin:
## Reviewer Guidance

Follow up to #24462 which added the logic to local driver and the validation login in the Loader layer. This PR extends the same logic to other drivers as well.

Also, note that there are a couple of test-only document service implementations to which this logic is not added - `FileDocumentService` and `StaticStorageDocumentService`. These drivers should not care about layer compatibility for now. If they do need this, it can be added later on a case-by-case basic. Keeping the logic on a need basis to keep things simple.

## Description

Added logic to support layer compatibility validation at the Loader / Driver layer boundary. Note that the validation only happens in the Loader layer and not in the Driver layer. This is because Driver layer doesn't call into any objects in the Loader layer. So, it doesn't need to support compat in that direction.

For context, please look at #22877 which added the same support to the Runtime / Loader boundary. This logic added by this PR is the same as in #22877.

### Supported features
- The `Driver` layer adds a `supportedFeatures` property to the `document service` that includes a set of features it supports. This is advertised to the Loader layer at the layer boundary.
- The `Loader` layer has a `requiredFeatures` array that includes a set of features that it requires the driver layer to support in order to be compatible. Note that this is internal and is not advertised.
- At runtime, the `Loader` layer validates that all the entries in `requiredFeatures` is present in the `supportedFeatures` of the `Driver`. If not, it throws a usage error that says layers are not compatible.
- Any change that adds new behavior or changes existing behavior across this layer boundary must add an entry to the supported features set of the `Driver`. For example, changes such as "does the driver support this new summary format foo" should add an entry to the supported features set of the driver.

### Generation
- In addition to `supported features`, a `generation` is also added to the `Driver` layer. This is advertised to the `Loader` layer at the layer boundary.
- `Generation` is an integer which will be incremented every month. It can be used to validate compatibility more strictly (time-based) than `supported features` where layers are incompatible only if there are unsupported features.
_Note: The logic to update the generation will be added in a later PR. See [AB#27054](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/27054) for a proposed solution._
- One key reason for adding generation is to have a clear cut off point where we officially stop supporting layer combinations. Our compat tests will test combinations up to this cut off point. The idea is that we test what we support and whatever we don't test should not be supported. This will help us achieve that.
[...]

PR #20793, opened 2024-04-23
Cleanup Container loading code

vladsud:
There is a bunch of logic in Container.load() that can be easily implemented as a layer on top of core layer, not as part of core layer.

This change removes such concepts as 
- IContainerLoadMode.opsBeforeReturn ===  "sequenceNumber"
- IContainerLoadMode.pauseAfterLoad
- LoaderHeader.sequenceNumber

**Improvements:**
- There is no more overlap with pending state options - it was possible (before) to ask to load from pending state, and up to sequence number - that would not work correctly.
- It was possible before to ask for non-paused, but loaded up to some sequence number Container - it would happily return container loaded from snapshot that has higher sequence number than the one where we needed to stop. This combo is no longer possible.
- Our E2E tests were covering some combination that do not matter and that should not have been existed.
- No need for a bunch of checks in Loader layer around non-valid combos or types.
- There is reference to https://dev.azure.com/fluidframework/internal/_workitems/edit/5055 in the code. That issue should be resolved with this PR, as we set op handler before we process any ops (including cached ops) with this PR.
- This logic could hang if container gets disposed while processing ops - adding checks for that.
- Logic around checking container.deltaManager.inbound.length looks fishy - removed
- Added ability to provide AbortSignal, as this process can hang (as explained in comments).

PR #10692, opened 2022-06-15
Allow local processing of ODSP Fluid files

kian-thompson:
# [AB#262](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/262)

For more information about how to contribute to this repo, visit this [page](https://github.com/microsoft/FluidFramework/blob/main/CONTRIBUTING.md)

## Description

For eDiscovery / search indexing work, we need capability to process Fluid files in a new service we are building.

## PR Checklist

-   [ ] I have updated the documentation accordingly.
-   [X] I have added tests to cover my changes.
-   [X] All new and existing tests passed.
-   [X] My code follows the code style of this project.
-   [X] I ran the lint checks which produced no new errors nor warnings for my changes.
-   [X] I have checked to ensure there aren't other open Pull Requests for the same update/change.

## Does this introduce a breaking change?

-   [ ] Yes
-   [X] No

PR #27703, opened 2026-07-16
Add ODSP point-in-time container loading by sequence number

sonalideshpandemsft:
This change adds a point-in-time container loading: materializing a read-only Fluid container as it existed at a requested sequence number, by reading a historical ODSP snapshot and replaying the remaining ops on top of it. 

### container-loader 
Exposes a new entry point:
- `loadContainerToSequenceNumber` -- loads a container in read-only, paused mode and returns it once ops have been replayed up to the requested sequence number. 
- `ILoadContainertoSequenceNumberProps` 

The caller passes their OdspDocumentServiceFactory. The loader detects a point-in-time capability on the factory and adapts it internally (pointInTimeService.ts not exported) so the container's normal createDocumentService(resolvedUrl) call is routed to the driver's point-in-time service with the target threaded through.
Loading reuses the existing frozen (storageOnly) mechanism

### odsp-driver
Adds `OdspDocumentServiceFactoryCore.createPointInTimeDocumentService` which:
- Uses the version manager to select a closest file version at or before the target sequence number
- Creats a point-in-time document service (snapshot stoage at the base sequence) and a live document service 
- Returns an OdspPointInTimeDocumentService


packages/loader/container-loader/src/loadContainerToSequenceNumber.ts
```diff
@@ -0,0 +1,97 @@
...
+ * The supplied {@link IContainerDriverServices.documentServiceFactory} must support point-in-time
+ * loading: it must be able to serve a snapshot at or before
+ * {@link ILoadContainerToSequenceNumberProps.loadToSequenceNumber} and replay the document forward
+ * through that sequence number. For ODSP, pass an `OdspDocumentServiceFactory` (from
+ * `@fluidframework/odsp-driver`) directly - the loader materializes the point-in-time view itself,
+ * so no wrapping or decoration is required.
+ *
+ * @internal
+ */
+export async function loadContainerToSequenceNumber(
```
anthony-murphy, line 67: Deep Review: `loadContainerToSequenceNumber` is exported here, and the PR description calls it the new entry point — but it never ships. `packages/loader/container-loader/src/index.ts:45` exports only `loadContainerPaused` from `./loadPaused.js`; there is no export of `./loadContainerToSequenceNumber.js` or `ILoadContainerToSequenceNumberProps`, and `package.json` maps both `.` and `./internal` to `lib/index.js`, so consumers cannot import the new API.

Add both symbols to `src/index.ts` (as `@internal`, matching their tags) and regenerate the package API report/artifacts — or confirm it is intentionally unexported for this draft.

anthony-murphy:
`packages/loader/container-loader/src/pointInTimeServices.ts:428`

Deep Review: `PointInTimeDocumentServiceFactory implements IDocumentServiceFactory` but forwards only `createDocumentService`/`createContainer` — it does not expose `ILayerCompatDetails`. On the ODSP driver that property lives on the factory itself (`odspDocumentServiceFactoryCore.ts:267`: `public readonly ILayerCompatDetails?: unknown = odspDriverCompatDetailsForLoader`), and the loader reads compat off the factory (`container.ts:749-755` casts `documentServiceFactory as FluidObject<ILayerCompatDetails>` and calls `validateDriverCompatibility`). Because `loadContainerToSequenceNumber` swaps the real factory for this wrapper, the loader validates against `undefined`.

It is bypassed today (`minSupportedGeneration` 0 → logs `LayerIncompatibilityDetectedButBypassed`), but once the loader generation advances past the `LoaderDriver: 12` window, point-in-time loading will silently begin throwing a "layers are not compatible" `UsageError`. Existing wrappers forward it (`DocumentServiceFactoryProxy.ts:27-33`, `ReplayDocumentServiceFactory.ts:35-42`).

Forward the inner factory's value, e.g. `public get ILayerCompatDetails() { return (this.inner as FluidObject<ILayerCompatDetails>).ILayerCompatDetails; }` — matching the convention established in #24855.

anthony-murphy:
`packages/drivers/odsp-driver/src/odspDocumentServiceFactoryCore.ts:138`

Deep Review: `resolveFileVersion` rebuilds a URL string from only `driveId`, `itemId`, `fileVersion`, and optionally `dataStorePath`/`codeHint.containerPackageName`, then returns `new OdspDriverUrlResolver().resolve({ url })`. It neither passes `request.headers` nor preserves properties from the original `odspResolvedUrl`, so `isClpCompliantApp` (enterprise CLP compliance) and `fileMetadata` are dropped on the historical document service. `epochTracker.ts:385-412` reads `getOdspResolvedUrl(...).isClpCompliantApp` to add `X-CLP-Compliant-App` headers, so losing it can affect historical snapshot requests (see #10692).

Preserve the resolved-URL context deliberately — clone `odspResolvedUrl` and set `fileVersion` (carrying `isClpCompliantApp`, `shareLinkInfo`, etc.) rather than round-tripping through a freshly built URL string. Decide explicitly whether `fileMetadata.eTag` (live-item derived) should carry to the historical version URL at all.

packages/drivers/odsp-driver/src/odspDocumentServiceFactoryCore.ts
```diff
@@ -269,6 +279,57 @@ export class OdspDocumentServiceFactoryCore
...
+			throw new UsageError(
+				`No ODSP file version is available at or before sequence number ${targetSequenceNumber}.${oldestResolvedSequenceDetail}`,
+			);
+		}
+
+		const historicalResolvedUrl = await this.resolveFileVersion(
+			resolvedUrl,
+			baseResult.base.versionId,
+		);
+		const historicalDocumentService = await this.createDocumentService(
```
anthony-murphy: Deep Review: In `createPointInTimeDocumentService`, `historicalDocumentService` is created first and `liveDocumentService` second, then both are handed to `new OdspPointInTimeDocumentService(...)`. The two inner services are only disposed inside `OdspPointInTimeDocumentService.dispose()`. If `this.createDocumentService(resolvedUrl, ...)` for the live service rejects, that instance is never constructed, so the already-created `historicalDocumentService` — which owns caches, token fetchers, and epoch trackers — is never disposed. No try/catch guards the second creation.

Wrap the live-service creation in try/catch, disposing `historicalDocumentService` on failure before rethrowing:

```ts
try {
  const liveDocumentService = await this.createDocumentService(resolvedUrl, ...);
  return new OdspPointInTimeDocumentService(...);
} catch (error) {
  historicalDocumentService.dispose();
  throw error;
}
```

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentService.ts
```diff
@@ -0,0 +1,91 @@
...
+
+	// storageOnly makes the connection manager synthesize a read-only frozen delta stream (no live
+	// socket) and force the container read-only - see the class remarks.
+	public readonly policies: IDocumentServicePolicies = { storageOnly: true };
+
+	public dispose(): void {
+		this.liveDocumentService.off("metadataUpdate", this.metadataUpdateHandler);
+		this.historicalDocumentService.dispose();
+		this.liveDocumentService.dispose();
+	}
```
anthony-murphy, line 59: Deep Review: `IDocumentService.dispose` is `dispose(error?: any): void`, but this override takes no parameter and calls the inner services' `dispose()` with no argument. `container.ts:1127-1130` forwards the container error to the document service, and `odspDocumentService.ts:303-315` flushes ops only on normal dispose but removes epoch entries when `error` is defined — so dropping the signal changes ODSP disposal behavior (cache invalidation vs. flush). Accept `error` and forward it to both underlying services.

```suggestion
	public dispose(error?: any): void {
		this.liveDocumentService.off("metadataUpdate", this.metadataUpdateHandler);
		this.historicalDocumentService.dispose(error);
		this.liveDocumentService.dispose(error);
	}
```

anthony-murphy:
## Deep Review

*Reviewed commit `dd21d95` on 2026-07-20.*

**Readiness: 3/10 — GETTING STARTED**

Not ready for sign-off. The prior review's headline blocker is fixed — `loadContainerToSequenceNumber` now ships from the package barrel (`index.ts`). The approach remains sound: reusing the `storageOnly` frozen path plus `loadContainerPaused`, with a base snapshot chosen at-or-before the target via `findBaseForSeq`. What holds it back is that a new read-only loading paradigm ships with zero automated tests, on top of the correctness defects already flagged inline.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add automated tests for the new paradigm — the delta clamp (`Math.min(to, target+1)`), the `noBaseVersion` error path, input validation, the capability detector (`asPointInTimeCapableFactory`), and dispose/error-forwarding of both inner services. Implement `IPointInTimeCapableDocumentServiceFactory` on a mock/local factory to enable an e2e replay test without a live ODSP endpoint. If tests are intentionally deferred for this draft, state that and scope them as a merge-blocking follow-up.

### Context for Reviewers

- Re-introduces load-to-sequence-number that #20793 removed from core — but as a layer on top of the loader, the direction #20793 asked for. The old "materialized from a newer snapshot" trap is guarded by at-or-before base selection plus bounding delta fetch to `target+1`.
- The base-snapshot ≤ target invariant is enforced and unit-tested one level down at the version-manager layer (`odspVersionManager.spec.ts:63-99`), so the #20793 trap is closed for the returned read-only container.
- The ODSP-scoped factory + dedicated `loadContainerToSequenceNumber` entry point were examined against a generic target-sequence field threaded through core; the PR's scoped choice is the better-reasoned one — the generic contract would live in `@legacy @beta` driver-definitions and carry a large blast radius for an ODSP-only feature. Not a quality deduction.
- Correctness hinges on the connection manager honoring `storageOnly: true` to synthesize a frozen delta stream (`connectionManager.ts:518-523`); a regression there turns point-in-time load into a thrown error. Only `lindsnguyen` is on the review — suggested reviewers by area: **agarwal-navin** (compat), **kian-thompson** (CLP), **vladsud**/**alexvy86** (factory surface).

- **Needs human judgment** — factory-method + duck-typed capability detection vs. a typed optional capability or flat factory function. The runtime-guard approach is defensible; vladsud/alexvy86 have enforced factory-surface conventions here.
- **Needs human judgment** — whether `fileMetadata.eTag` (live-item derived) should carry to the historical version URL at all; kian-thompson is the best fit for the CLP/resolved-URL decision.
- **Needs human judgment** — whether standing up two full document services per point-in-time load is acceptable for this rarely-used read-only path, or worth collapsing to a single storage/version capability.
- **Needs human judgment** — whether the `no-internal-modules`-suppressed package-root export of `OdspPointInTimeDocumentServiceFactory` is an intentional supported host surface; confirm with the ODSP export-surface owner (vladsud).
- **Cannot be assessed by the pipeline** — that the `storageOnly` + `loadContainerPaused` path actually pauses at the target sequence number and surfaces gapped/truncated op streams as hard failures rather than hangs.

</details>

- `7729b4c` 2026-07-17 · **4/10** — export blocker + five inline defects flagged

</details>

packages/loader/container-loader/src/pointInTimeServices.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Adapts a point-in-time capable factory to a plain {@link IDocumentServiceFactory} that always
+ * materializes the configured target sequence number.
+ *
+ * @remarks
+ * The loader plumbs the caller's target through this adapter so the container's normal
+ * `createDocumentService(resolvedUrl)` call routes to `createPointInTimeDocumentService`. This keeps
+ * point-in-time loading a single-call experience for the caller (see
+ * {@link loadContainerToSequenceNumber}) - they never build or pass a decorated factory themselves.
+ */
+export class PointInTimeDocumentServiceFactory implements IDocumentServiceFactory {
```
jatgarg, line 60: I think this file could live in the odsp driver in the pointInTimeDriver. There is nothing related to loader here.

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentServiceFactory.ts
```diff
@@ -0,0 +1,173 @@
...
+		);
+		return new OdspPointInTimeDocumentService(
+			historicalResolvedUrl,
+			historicalDocumentService,
+			liveDocumentService,
+			targetSequenceNumber,
+		);
+	}
+
+	private async createVersionManager(
```
jatgarg, line 126: Can you please add some more comments in this file explaining a bit what these functions are doing

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentServiceFactory.ts
```diff
@@ -0,0 +1,173 @@
...
+			throw new UsageError(
+				`No ODSP file version is available at or before sequence number ${targetSequenceNumber}.${oldestResolvedSequenceDetail}`,
+			);
+		}
+
+		const historicalResolvedUrl = await this.resolveFileVersion(
+			resolvedUrl,
+			baseResult.base.versionId,
+		);
+		const historicalDocumentService = await this.createDocumentService(
```
jatgarg: Please add comment what is historicalDocumentService

packages/loader/container-loader/src/pointInTimeServices.ts
sonalideshpandemsft: I added this to container loader only for the prototype. Even if we move this to odsp driver, container-loader can’t import it because it’s driver-agnostic. I’ll likely need to create a separate package for this later. For now, I’m keeping it here for the prototype and move later

jatgarg (approved):

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentServiceFactory.ts
```diff
@@ -0,0 +1,185 @@
...
+		const recoverableDocumentService = await this.createDocumentService(
+			recoverableResolvedUrl,
+			logger,
+			clientIsSummarizer,
+		);
+		const liveDocumentService = await this.createDocumentService(
+			resolvedUrl,
+			logger,
+			clientIsSummarizer,
+		);
```
lindsnguyen, line 105: Based on what we learned about ODSP's restore and the epoch lineage stuff, I think we need to validate that the base version and the live doc's lineage is the same otherwise I think the ops replayed could be incorrect? (I'm not 100% sure how this part works) If you don't want to do it now, we can file a follow-up? but we should definitely think about how we want to handle it.

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentService.ts
```diff
@@ -0,0 +1,94 @@
...
+	public async connectToStorage(): Promise<IDocumentStorageService> {
+		return this.recoverableDocumentService.connectToStorage();
+	}
+
+	public async connectToDeltaStorage(): Promise<IDocumentDeltaStorageService> {
+		const liveDeltaStorage = await this.liveDocumentService.connectToDeltaStorage();
+		// The exclusive upper bound needed to include the target op itself.
+		const boundedTo = this.targetSequenceNumber + 1;
+		return {
+			fetchMessages: (from, to, abortSignal, cachedOnly, fetchReason) => {
```
lindsnguyen, line 70: I think we need to handle the specific scenario when not all the ops we want are correctly returned? it's especially important if we really only have 7 days of ops, copilot proposed something like this:

```
fetchMessages: (from, to, abortSignal, cachedOnly, fetchReason) => {
    const bounded = to === undefined ? boundedTo : Math.min(to, boundedTo);
    const inner = liveDeltaStorage.fetchMessages(from, bounded, abortSignal, cachedOnly, fetchReason);
    let maxSeq = from - 1;
    return {
        read: async () => {
            const result = await inner.read();
            if (!result.done && result.value.length > 0) {
                maxSeq = Math.max(maxSeq, result.value[result.value.length - 1].sequenceNumber);
            }
            if (result.done && maxSeq < bounded - 1) {
                throw new UsageError(
                    `Cannot materialize sequence number ${targetSequenceNumber}: delta storage returned ops through ${maxSeq}, ` +
                    `short of ${bounded - 1} (retention/trim or target beyond tip).`,
                );
            }
            return result;
        },
    };
}
```

I haven't tested it or anything but it might be a good start

sonalideshpandemsft: I'll follow up on this: https://dev.azure.com/fluidframework/internal/_workitems/edit/77927

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentServiceFactory.ts
sonalideshpandemsft: I'll follow up on this: https://dev.azure.com/fluidframework/internal/_workitems/edit/77927

lindsnguyen, line 105: Can be done as follow-up as well but rn since the recoverable and live document services are created sequentially, if the live  `createDocumentService` throws after the recoverable one succeeds, the recoverable service is never disposed. I think we need to wrap the second call in a try to dispose the first service if it fails.

lindsnguyen (approved):
one last comment, can be done in followup.

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentServiceFactory.ts
sonalideshpandemsft: https://dev.azure.com/fluidframework/internal/_workitems/edit/77964

PR #27704, opened 2026-07-16
feat: add imperative payload sharing for local Fluid handles

WillieHabi:
## Description

Fixes [AB#77539](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/77539)

This change adds an optional `ILocalFluidHandle.sharePayload()` API and implements it for blob handles.

This imperative API lets a consumer start the existing asynchronous payload upload and `BlobAttach` flow before serializing the handle into a DDS/marking the handle as attached to the Fluid object graph.

`attachGraph()` preserves the existing behavior by starting payload sharing when needed but does not start a second upload if `sharePayload()` was called first. Pending state continues to include only blobs whose handles have been attached to the Fluid object graph while payload sharing remains unfinished, calling `sharePayload()` alone does not make an unattached blob part of pending state.

The interface member is optional to preserve compatibility with older `ILocalFluidHandle` implementations.

### DIFF
--- a/packages/runtime/container-runtime/src/blobManager/blobManager.ts
+++ b/packages/runtime/container-runtime/src/blobManager/blobManager.ts
@@ -102,6 +102,7 @@ export class BlobHandle
 		public get: () => Promise<ArrayBufferLike>,
 		public readonly payloadPending: boolean,
 		private readonly sharePayloadCallback?: () => void,
+		private readonly attachGraphCallback?: () => void,
 	) {
 		super();
 		this._payloadState = payloadPending ? "pending" : "shared";
@@ -119,7 +120,7 @@ export class BlobHandle
 	};
 
 	public sharePayload(): void {
-		if (this.payloadSharingStarted) {
+		if (this.payloadState === "shared" || this.payloadSharingStarted) {
 			return;
 		}
 		this.payloadSharingStarted = true;
@@ -129,6 +130,7 @@ export class BlobHandle
 	public attachGraph(): void {
 		if (!this.attached) {
 			this.attached = true;
+			this.attachGraphCallback?.();
 			this.sharePayload();
 		}
 	}
@@ -262,11 +264,10 @@ export class BlobManager {
 	 */
 	private readonly localBlobCache: Map<string, LocalBlobRecord> = new Map();
 	/**
-	 * Blobs whose payload sharing has started but that have not finished blob-attaching are the set we need to
-	 * provide from getPendingState(). This stores their local IDs, and then we can look them up against the
-	 * localBlobCache.
+	 * Blobs with an attached handle that have not finished blob-attaching are the set we need to provide from
+	 * getPendingState(). This stores their local IDs, and then we can look them up against the localBlobCache.
 	 */
-	private readonly pendingBlobsWithSharingStarted: Set<string> = new Set();
+	private readonly pendingBlobsWithAttachedHandles: Set<string> = new Set();
 	/**
 	 * Local IDs for any pending blobs we loaded with and have not yet started the upload/attach flow for.
 	 */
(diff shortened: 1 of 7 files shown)


packages/runtime/container-runtime/src/blobManager/blobManager.ts
```diff
@@ -253,10 +262,11 @@ export class BlobManager {
...
 	private readonly localBlobCache: Map<string, LocalBlobRecord> = new Map();
 	/**
-	 * Blobs with an attached handle that have not finished blob-attaching are the set we need to provide from
-	 * getPendingState().  This stores their local IDs, and then we can look them up against the localBlobCache.
+	 * Blobs whose payload sharing has started but that have not finished blob-attaching are the set we need to
+	 * provide from getPendingState(). This stores their local IDs, and then we can look them up against the
+	 * localBlobCache.
 	 */
-	private readonly pendingBlobsWithAttachedHandles: Set<string> = new Set();
+	private readonly pendingBlobsWithSharingStarted: Set<string> = new Set();
```
ChumpChief: Hmm, this is probably a complexity to deal with - we don't really want to include blobs in the pending state if their handles aren't actually attached (because those blobs are instantly garbage if their handles aren't actually referenced, so there's no point in uploading them).  So pendingBlobsWithAttachedHandles is probably still the right semantic to track here.

You might consider splitting the callback in two - having a specific callback for when the handle is attached vs. shared.  Then the attached callback would need to only conditionally add the id to this collection if the blob was _still_ in pending state.

packages/runtime/container-runtime/src/blobManager/blobManager.ts
```diff
@@ -117,10 +118,18 @@ export class BlobHandle
 		this._events?.emit("payloadShareFailed", error);
 	};
 
+	public sharePayload(): void {
+		if (this.payloadSharingStarted) {
```
ChumpChief: Probably also conditionalize on whether the payload is already shared

```suggestion
		if (this.payloadState === "shared" || this.payloadSharingStarted) {
```

ChumpChief (commented):
I didn't think about how this gives new permutations on the ordering of attaching, sharing, etc.  Hopefully not too bad to deal with by being explicit about the actions of attaching the handle vs. sharing the payload?

packages/runtime/container-runtime/src/blobManager/blobManager.ts
WillieHabi: Makes sense, this actually makes the change simpler I think. restored the `pendingBlobsWithAttachedHandles`  semantics and split the callbacks so payload sharing and graph attachment are handled independently.

The attachment callback now adds the blob to pending-state tracking only when its local record still exists and has not already reached the 'attached' state (maybe need to double check this! this should avoid adding blobs who sharing has completed or failed before attatchment).

WillieHabi: It's kind of confusing, but my understanding now is "attached handle" means handle is stored in a DDS/referenced in document and "attached blob" means BlobAttach op has been processed from service. 

So we need to do something like

```typescript
if (localBlobRecord.state !== "attached") {
	this.pendingBlobsWithAttachedHandles.add(localId);
}
```

ChumpChief: That's right, and I think what you have here looks good.

In general we overuse the word "attach" throughout Fluid :-P. Unfortunately both of these usages (blobs, handles) were preexisting so wasn't worth trying to rename.

ChumpChief (approved):
Looks good to me, thank you!

Josmithr (approved):
Docs changes look good ✅

PR #27706, opened 2026-07-16
feat(tree): alpha findLast functions for arrayNode

brrichards:
## Description
Adds `findLast()` and `findLastIndex()` methods for ArrayNode. This is part of the effort to update arrayNode apis to contain apis similar to a normal ts array. Results reflect the state of the array when called. Items are read live, so if the callback edits the array remaining reads see the shifted array, and `findLast` returns the value captured at match time. 

Added alpha methods:

- `findLast(callbackFunction)` - returns the last item for which `callbackFunction` returns `true`, or `undefined` if there is no such item.
- `findLastIndex(callbackFunction)` - similar to `findLast()` instead returning the index and not the item. Returns -1 if no valid item.

## Reviewer Guidance


packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
```diff
@@ -482,6 +482,33 @@ export interface TreeArrayNodeAlpha<
...
+	findLast(
+		callbackFunction: (value: T, index: number, array: readonly T[]) => boolean,
+	): T | undefined;
+
+	/**
+	 * Returns the index of the last item in the array for which the given callbackFunction returns `true`,
+	 * searching from the last item to the first.
+	 * @remarks
+	 * Results reflect the state of the array when called.
+	 * If `callbackFunction` edits the array, behavior matches {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLastIndex | Array.prototype.findLastIndex}.
```
Josmithr: We should add a similar note to `findLast`.

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -253,6 +253,112 @@ describe("ArrayNode", () => {
...
+					const array = buildAlphaArray([1, 2, 3]);
+					assert.equal(
+						array.findLast((value) => value > 10),
+						undefined,
+					);
+				});
+
+				it("visits items from last to first, passing the index and array to the callbackFunction", () => {
+					const array = buildAlphaArray([5, 6, 7]);
+					const visited: [number, number][] = [];
```
Josmithr: Nit: rather than representing the value and visited order as a tuple, it would probably be a bit better to use an object with named properties. Helps better differentiate the two elements at-a-glance.

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -253,6 +253,112 @@ describe("ArrayNode", () => {
 				});
 			});
 
+			describe("findLast", () => {
```
Josmithr, line 256: A thought just occurred to me for these tests. While we aren't promising at an API level that our array nodes perfectly match the behavior of standard javascript arrays, we are generally trying to uphold this as an invariant. It might be nice to update our infra here to execute the same test scenarios against both our array nodes and TS arrays, and assert that they demonstrate the same behavior. We can then opt out of this validation on a case-by-case or method-by-method basis, and document why the deviation is expected/intended.

This also will give us a signal any time we update our TS or ecmascript versions if our behaviors have deviated from the broader ecosystem, at which point we can update our documentation to make the differentiation more evident to users.

Doesn't need to be in this PR, but is probably something for us to consider here.

Josmithr (commented):
Left a couple of quick suggestions. Otherwise looks good to me! Feel free to ping me when you need another round of review :)

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
brrichards: good call. updated!

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
brrichards: I agree this would be useful. I won't include it in this PR, but I can look into getting what you're talking about set up before the next round of alpha API's

Josmithr (approved):

PR #27707, opened 2026-07-16
feat(publish): Removed public publishing from pipeline

MarioJGMsoft:
## Description

Policy changes block registry.npmjs.org in various contexts. Now we have to update our publishing process based on these new policies. This change will unblock our build pipelines.

## Reviewer Guidance


tools/pipelines/templates/include-publish-npm-package.yml
```diff
@@ -75,7 +75,7 @@ stages:
 - stage: publish_npm_public
   dependsOn: build
   displayName: Publish to public npmjs.org
-  condition: and(succeeded(), or(eq(variables['release'], 'release'), eq(variables['release'], 'prerelease')))
+  condition: false # Disabled until we onboard to the new Public Publishing process. See #26466
```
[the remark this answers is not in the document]
MarioJGMsoft: I'm referring to an ADO item

dannimad (approved):

tools/pipelines/templates/build-npm-client-package.yml
```diff
@@ -115,6 +115,11 @@ parameters:
 - name: publish
   type: boolean
 
+# Indicates if publishing to the public npmjs.org registry is allowed.
+- name: allowPublishingToNpmjs
+  type: boolean
+  default: false
```
jason-ha, line 121: This should not have a default. All pipelines using this template should be making an explicit decision.
If it needs a default per staging with ff/internal pipelines, then comment the need to stage and why false is okay default. Then fix consumers and remove the default.

tools/pipelines/templates/include-publish-npm-package.yml
```diff
@@ -15,6 +15,10 @@ parameters:
   type: boolean
   default: false
 
+- name: allowPublishingToNpmjs
+  type: boolean
+  default: false
```
jason-ha, line 20: This really should not have a default of `false`. If it needs a default per staging with ff/internal pipelines, then default to true until that is updated. Then remove the default.
Since there are no updates setting this to true in the repo, all pipelines except client build can't publish to npmjs.org now. That is 13 pipelines.

tools/pipelines/templates/include-publish-npm-package.yml
```diff
@@ -75,7 +79,7 @@ stages:
 - stage: publish_npm_public
   dependsOn: build
   displayName: Publish to public npmjs.org
-  condition: and(succeeded(), or(eq(variables['release'], 'release'), eq(variables['release'], 'prerelease')))
+  condition: and(succeeded(), eq(${{ parameters.allowPublishingToNpmjs }}, true), or(eq(variables['release'], 'release'), eq(variables['release'], 'prerelease')))
```
jason-ha, line 82: no need to change:
It would be better to wholly exclude this stage than to do so condition.
Alternatively have a dummy stage using the same name to avoid complicated depends/conditions referencing this stage. The dummy can use a different displayName to make it clear that npmjs publish won't happen.

jason-ha, line 82: This is a boolean parameter. So, there is no need to use `eq(foo, true)` to deal with variable patterns.
Replace `eq(${{ parameters.allowPublishingToNpmjs }}, true)` with `${{ parameters.allowPublishingToNpmjs }}` (assuming not wholly excluding as noted in other comment).

jason-ha (commented):
Some things to follow up on

PR #27712, opened 2026-07-17
build: Add `docs` directory with lint/formatting tooling, prepopulated with a subset of content from our github wiki

Josmithr:
"Guidelines" content copied from github wiki to serve as initial validation of the new directory + tooling. More content will be ported in future PRs.


tools/pipelines/docs-lint.yml
```diff
@@ -0,0 +1,136 @@
...
+# Validates formatting (Prettier) and linting (markdownlint) for the `docs` directory, which is a
+# standalone pnpm workspace. Scoped to changes under `docs/` via path filters.
+
+trigger:
+  branches:
+    include:
+    - main
+    - next
+    - lts
+    - release/*
```
alexvy86: I think we only need `main` for this pipeline, no? Same in PR triggers below.

tools/pipelines/docs-lint.yml
```diff
@@ -0,0 +1,136 @@
...
+  paths:
+    include:
+    - docs
+    - common/build/build-common/prettier.config.cjs # The "docs" workspace depends on this file via file path, so we need to run the docs-lint pipeline if it changes.
```
alexvy86: Should also be in the `trigger` paths above. And both need this yml file itself too, as well as any templates it uses (e.g. `tools/pipelines/templates/include-use-node-version.yml`, `tools/pipelines/templates/include-install-pnpm.yml`, and anything they use recursively.

tools/pipelines/docs-lint.yml
```diff
@@ -0,0 +1,136 @@
...
+            workingDir: $(Pipeline.Workspace)/${{ variables.docsDirectory }}
+            customCommand: 'run lint'
+
+        - task: Bash@3
+          displayName: Prune pnpm store
+          inputs:
+            targetType: 'inline'
+            script: |
+              set -eu -o pipefail
+              pnpm store prune
```
alexvy86: Is this one really necessary here?

docs/pnpm-workspace.yaml
```diff
@@ -0,0 +1,21 @@
...
+minimumReleaseAge: 1440
+resolutionMode: highest
+blockExoticSubdeps: true
+trustPolicy: no-downgrade
+strictDepBuilds: true
+
+engineStrict: true
+frozenLockfile: true
+strictPeerDependencies: true
+updateNotifier: false
```
alexvy86, line 26: We probably want to align with the root `pnpm-workspace.yaml` for the supply chain security settings, e.g. `trustPolicy: off` because of the funky situation we have with needing to do things through different feeds in different places. In the root one they're also alphabetized, which I don't feel that strongly about.

tools/pipelines/docs-lint.yml
Josmithr: I'm not sure. Presumably Copilot copied this from one of our other pipelines. I can remove if we think it's unneeded.

tools/pipelines/docs-lint.yml
Josmithr: [Copilot]: Done — both the `trigger` and `pr` sections now target only `main` (removed `next`, `lts`, and `release/*`).

tools/pipelines/docs-lint.yml
Josmithr: [Copilot]: Done — both `trigger.paths` and `pr.paths` now include the same set: `docs`, the pipeline definition itself (`tools/pipelines/docs-lint.yml`), the two templates it uses (`include-use-node-version.yml` and `include-install-pnpm.yml` — confirmed neither references any further sub-templates), and `common/build/build-common/prettier.config.cjs`.

tools/pipelines/docs-lint.yml
Josmithr: [Copilot]: Removed the `Prune pnpm store` step. It was copied over from `repo-policy-check.yml` and is not needed for this lightweight lint job.

docs/pnpm-workspace.yaml
Josmithr: [Copilot]: Done — aligned the supply-chain settings with the root `pnpm-workspace.yaml`: set `trustPolicy: off` (with a comment noting the registry-mirror rationale), added `verifyDepsBeforeRun: false`, and alphabetized the settings to match the root ordering. Verified that `pnpm install --frozen-lockfile` still passes the supply-chain policies and `pnpm lint` is clean.

alexvy86 (approved):

PR #27714, opened 2026-07-17
feat(container-loader): add createEmptyRuntimeCodeLoader

anthony-murphy:
## Description

Adds a new `@alpha` `createEmptyRuntimeCodeLoader` free function to
`@fluidframework/container-loader`. It returns an `ICodeDetailsLoader` whose loaded
module instantiates a minimal, no-op container runtime.

This gives hosts the full set of `IContainer` capabilities (load, connect,
audience/quorum, read pending state) without needing to wire up a real container
runtime. The empty runtime:

- ignores all incoming ops and signals,
- never sends any ops or signals of its own,
- throws from `createSummary` (so attach/serialize are unsupported, since it has no
  content), and
- echoes back whatever pending local state was provided on the container context.

The composition follows the existing container-loader pattern: free function →
`ICodeDetailsLoader` → `IRuntimeFactory` → no-op `IRuntime`.

## Reviewer Guidance

This is a **draft**: additional `local-server-tests` coverage is in progress. Unit
tests in `container-loader` are included and passing.

### DIFF
--- a/.changeset/empty-runtime-code-loader.md
+++ b/.changeset/empty-runtime-code-loader.md
@@ -4,20 +4,23 @@
 ---
 Add support for containers backed by an empty, no-op runtime
 
-A new `@alpha` `createEmptyRuntimeCodeLoader` function is now exported from `@fluidframework/container-loader`.
-It returns an `ICodeDetailsLoader` whose loaded module produces a minimal, no-op container runtime.
+Two new `@alpha` functions are now exported from `@fluidframework/container-loader`: `createEmptyRuntimeCodeLoader`, which returns an `ICodeDetailsLoader` whose loaded module produces a minimal, no-op container runtime, and `createEmptyRuntimeFactory`, which returns just the underlying `IRuntimeFactory` for composing into an existing code loader.
 
 This is useful when you need the capabilities of an `IContainer` — such as loading, connecting to, or reading the pending state of an existing container — without wiring up a real container runtime.
 The empty runtime ignores all incoming ops and signals, and it never sends any ops or signals of its own.
 Because it has no content, it can only be used to load existing containers: creating a new (detached) container with it throws.
 
+Note that the "never sends ops or signals" guarantee comes from the runtime never invoking the `submit*` callbacks on its container context — the container still connects as a normal (potentially writable) client at the network layer.
+This complements the frozen container-loader utilities (such as `createFrozenDocumentServiceFactory`), which instead enforce read-only behavior at the network/driver layer.
+Callers who need network-layer read-only enforcement should pair the empty runtime with those utilities.
+
 ```typescript
 // ...
-const loader = new Loader({
+const container = await loadExistingContainer({
 	codeLoader: createEmptyRuntimeCodeLoader(),
 	documentServiceFactory,
 	urlResolver,
+	request: { url },
 });
-const container = await loader.resolve({ url });
 // ...
 ```
--- a/.changeset/empty-runtime-code-loader.md
+++ b/.changeset/empty-runtime-code-loader.md
@@ -7,9 +7,9 @@ Add support for containers backed by an empty, no-op runtime
 A new `@alpha` `createEmptyRuntimeCodeLoader` function is now exported from `@fluidframework/container-loader`.
 It returns an `ICodeDetailsLoader` whose loaded module produces a minimal, no-op container runtime.
 
-This is useful when you need the capabilities of an `IContainer` — such as loading, connecting to, or reading the pending state of a container — without wiring up a real container runtime.
+This is useful when you need the capabilities of an `IContainer` — such as loading, connecting to, or reading the pending state of an existing container — without wiring up a real container runtime.
 The empty runtime ignores all incoming ops and signals, and it never sends any ops or signals of its own.
-Because it has no content, it does not support summarization: attaching or serializing the container will throw.
+Because it has no content, it can only be used to load existing containers: creating a new (detached) container with it throws.
 
 ```typescript
 // ...
--- a/.changeset/empty-runtime-code-loader.md
+++ b/.changeset/empty-runtime-code-loader.md
@@ -1,0 +1,23 @@
+---
+"@fluidframework/container-loader": minor
+"__section": legacy
+---
+Add support for containers backed by an empty, no-op runtime
+
+A new `@alpha` `createEmptyRuntimeCodeLoader` function is now exported from `@fluidframework/container-loader`.
+It returns an `ICodeDetailsLoader` whose loaded module produces a minimal, no-op container runtime.
+
+This is useful when you need the capabilities of an `IContainer` — such as loading, connecting to, or reading the pending state of a container — without wiring up a real container runtime.
+The empty runtime ignores all incoming ops and signals, and it never sends any ops or signals of its own.
+Because it has no content, it does not support summarization: attaching or serializing the container will throw.
+
+```typescript
+// ...
+const loader = new Loader({
+	codeLoader: createEmptyRuntimeCodeLoader(),
+	documentServiceFactory,
+	urlResolver,
+});
+const container = await loader.resolve({ url });
+// ...
+```
--- a/packages/loader/container-loader/api-report/container-loader.legacy.alpha.api.md
+++ b/packages/loader/container-loader/api-report/container-loader.legacy.alpha.api.md
@@ -27,6 +27,9 @@ export interface ContainerAlpha extends IContainer {
 export function createDetachedContainer(createDetachedContainerProps: ICreateDetachedContainerProps): Promise<IContainer>;
 
 // @alpha @legacy
+export function createEmptyRuntimeCodeLoader(): ICodeDetailsLoader_2;
+
+// @alpha @legacy
 export function createFrozenDocumentServiceFactory(factory?: IDocumentServiceFactory | Promise<IDocumentServiceFactory>, readOnly?: boolean): IDocumentServiceFactory;
 
 // @beta @legacy (undocumented)
--- a/packages/loader/container-loader/api-report/container-loader.legacy.alpha.api.md
+++ b/packages/loader/container-loader/api-report/container-loader.legacy.alpha.api.md
@@ -30,6 +30,9 @@ export function createDetachedContainer(createDetachedContainerProps: ICreateDet
 export function createEmptyRuntimeCodeLoader(): ICodeDetailsLoader_2;
 
 // @alpha @legacy
+export function createEmptyRuntimeFactory(): IRuntimeFactory;
+
+// @alpha @legacy
 export function createFrozenDocumentServiceFactory(factory?: IDocumentServiceFactory | Promise<IDocumentServiceFactory>, readOnly?: boolean): IDocumentServiceFactory;
 
 // @beta @legacy (undocumented)
--- a/packages/loader/container-loader/src/emptyRuntime.ts
+++ b/packages/loader/container-loader/src/emptyRuntime.ts
@@ -28,12 +28,14 @@ import { loaderCoreCompatDetails } from "./loaderLayerCompatState.js";
  * A minimal, no-op {@link @fluidframework/container-definitions#IRuntime} implementation.
  *
  * @remarks
- * This runtime exists to give a host the full set of {@link @fluidframework/container-definitions#IContainer}
- * capabilities (connect, attach, storage, quorum, audience, etc.) without wiring up a real container runtime.
+ * This runtime exists to give a host the {@link @fluidframework/container-definitions#IContainer} capabilities
+ * (connect, storage, quorum, audience, etc.) of an already-existing container without wiring up a real container
+ * runtime.
  *
  * Nearly every member is a no-op, and it crucially never invokes any of the `submit*` callbacks on the
  * {@link @fluidframework/container-definitions#IContainerContext} it is given, so it can never send an op or a signal.
- * It has no content, so {@link EmptyRuntime.createSummary} throws; and {@link EmptyRuntime.getPendingLocalState}
+ * It has no content, so it can only load existing containers (its factory throws when asked to instantiate for a
+ * newly created container) and {@link EmptyRuntime.createSummary} throws; {@link EmptyRuntime.getPendingLocalState}
  * simply echoes back whatever pending state was provided on the context.
  */
 class EmptyRuntime implements IRuntime {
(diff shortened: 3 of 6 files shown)


packages/loader/container-loader/src/emptyRuntime.ts
```diff
@@ -0,0 +1,127 @@
...
+		return {};
+	}
+
+	public close(): void {}
+}
+
+/**
+ * An {@link @fluidframework/container-definitions#IRuntimeFactory} that only ever produces {@link EmptyRuntime}s.
+ */
+class EmptyRuntimeFactory implements IRuntimeFactory {
```
anthony-murphy, line 91: can we also expose a createEmptyRuntimeFactory

packages/loader/container-loader/src/emptyRuntime.ts
```diff
@@ -0,0 +1,136 @@
...
+	public get IRuntimeFactory(): IRuntimeFactory {
+		return this;
```
anthony-murphy, line 93: Deep Review: `EmptyRuntime` re-derives `{ ...loaderCoreCompatDetails, supportedFeatures: new Set<string>() }` inline, but the identical object is already exported `@internal` in this package as `loaderCompatDetailsForRuntime` (`loaderLayerCompatState.ts:39-45`) — the canonical value the Loader advertises across the Loader/Runtime boundary and feeds into `validateRuntimeCompatibility` (`loaderLayerCompatState.ts:100`, used at `container.ts:2468-2470`). Re-deriving it duplicates the definition; if the canonical constant ever gains a supported feature, this hand-rolled copy silently drifts and the empty runtime could pass/fail the compat gate inconsistently with a real runtime. Import and reuse `loaderCompatDetailsForRuntime` from `./loaderLayerCompatState.js` as the value of `ILayerCompatDetails`, or add a comment documenting the intentional divergence.

packages/loader/container-loader/src/emptyRuntime.ts
```diff
@@ -0,0 +1,136 @@
...
+		return this.pendingLocalState;
+	}
+
+	public async notifyOpReplay(_message: ISequencedDocumentMessage): Promise<void> {}
+
+	public async getEntryPoint(): Promise<FluidObject> {
+		return {};
+	}
+
+	public close(): void {}
```
anthony-murphy, line 85: Deep Review: The class-level `@remarks` uses `{@link EmptyRuntime.createSummary}` and `{@link EmptyRuntime.getPendingLocalState}`, but `EmptyRuntime` is module-private — only `createEmptyRuntimeCodeLoader` is re-exported (index.ts:194), and the api-report contains no `EmptyRuntime` class. A `{@link}` to a member of a non-exported symbol won't render as a navigable link and is a plausible `ae-unresolved-link`-style API-Extractor/tsdoc warning vector in CI. Reword to plain prose or drop the `{@link}` wrappers. If the two references span separate lines, apply the reword across both rather than committing the suggestion verbatim.

```suggestion
     * newly created container) and `createSummary` throws; `getPendingLocalState`
```

anthony-murphy:
`packages/test/local-server-tests/src/test/emptyRuntimeCodeLoader.spec.ts:378`

Deep Review: The op spy filters to attributable traffic — `counts.appOps += messages.filter((m) => m.type === MessageType.Operation).length` (line 373) with a comment that system ops the container layer may send on its own (join, noop, …) are excluded. The `submitSignal` branch (lines 378-382) increments `counts.signals` for *every* signal with no comparable filtering, then line 463 asserts `counts.signals === 0`. If the container/loader layer ever emits a system signal of its own, the assertion fails even though the empty runtime sent nothing. If any signals are attributable to the loader layer, filter them out (or add a comment documenting why no loader-originated signals are expected) to mirror the op-filtering rationale.

.changeset/empty-runtime-code-loader.md
```diff
@@ -0,0 +1,23 @@
+---
```
anthony-murphy, line 1: Deep Review: The frozen-container series enforces "load an existing container and send nothing" at the network layer (#27211 `WritableFrozenDeltaStream` + `ConnectionManager.sendMessages` short-circuit; #25653 `createFrozenDocumentServiceFactory`). This PR instead guarantees "no ops/signals" purely by the runtime never invoking `submit*`, while connecting as a normal (potentially writable) client. The changeset and PR description never mention that series, so a reviewer coming from that history won't know why network-layer enforcement is unnecessary here, nor whether callers needing read-only network enforcement should still pair this with `createFrozenDocumentServiceFactory`. Cross-reference the frozen-container series and state explicitly that "no ops" is guaranteed by the runtime never calling the context `submit*` callbacks (not by the delta stream / document service), so callers who also need network-layer read-only enforcement should still pair this with `createFrozenDocumentServiceFactory`.

packages/loader/container-loader/src/emptyRuntime.ts
anthony-murphy: Deep Review: Good question — whether to also expose a `createEmptyRuntimeFactory` is an API-surface decision that a human reviewer should weigh rather than a mechanical call. Balance a single loader helper against also exposing the factory in light of the prior guidance on this surface that a second export "should become an overload" with a more-specific name (#25513). Surfaced for reviewer sign-off in the review summary.

packages/loader/container-loader/api-report/container-loader.legacy.alpha.api.md
```diff
@@ -26,6 +26,12 @@ export interface ContainerAlpha extends IContainer {
 // @beta @legacy
 export function createDetachedContainer(createDetachedContainerProps: ICreateDetachedContainerProps): Promise<IContainer>;
 
+// @alpha @legacy
+export function createEmptyRuntimeCodeLoader(): ICodeDetailsLoader_2;
```
anthony-murphy, line 30: Deep Review: This new `@alpha` export ships as `createEmptyRuntimeCodeLoader(): ICodeDetailsLoader_2;` in the committed api-report — the `_2` suffix leaks into the public signature. `emptyRuntime.ts` imports `ICodeDetailsLoader` from `@fluidframework/container-definitions/internal`, while this package also exports its own deprecated same-named `ICodeDetailsLoader` (from `src/loader.ts`, re-exported via `index.ts`). API Extractor disambiguates the two reachable declarations by renaming one to `ICodeDetailsLoader_2`.

The returned type is functionally correct — it is the canonical container-definitions `ICodeDetailsLoader`; the suffix is purely an API-Extractor disambiguation artifact, so no callsite or consumer type-resolution is wrong. Confirm whether it's intentional: either align the export to the package-local `ICodeDetailsLoader`, or add a note that the collision is a transitional artifact that disappears when the deprecated interface is removed. Worth ChumpChief's eye given the prior api-report scaling/naming guidance (#25513).

packages/loader/container-loader/src/emptyRuntime.ts
```diff
@@ -0,0 +1,149 @@
...
+	return new EmptyRuntimeFactory();
+}
+
+/**
+ * Creates an {@link @fluidframework/container-definitions#ICodeDetailsLoader} whose loaded module produces an empty,
+ * no-op container runtime.
+ *
+ * @remarks
+ * Use this when you need the {@link @fluidframework/container-definitions#IContainer} capabilities of an
+ * already-existing container (for example to load, connect to, or read the pending state of a container) but do not
```
anthony-murphy, line 130: Deep Review: `getPendingLocalState(_props?: IGetPendingLocalStateProps): unknown { return this.pendingLocalState; }` ignores `_props` and returns the value captured at construction. `SerializedStateManager.getPendingLocalState()` passes `snapshotSequenceNumber` via `IGetPendingLocalStateProps`, which a real `ContainerRuntime` uses to derive fresh pending state keyed to the current snapshot baseline. If a host serializes via the empty runtime after the base snapshot advances, then restores with a *real* runtime, stale runtime pending state could be packaged against a newer snapshot. No test exercises this load→advance→re-serialize→restore-with-real-runtime path.

This is a genuine but extreme edge case at the boundary of the supported contract — the empty runtime holds no runtime-owned mutable state it *could* recompute from `snapshotSequenceNumber`. If mixing the empty runtime with a later real-runtime restore is unsupported, a one-line `@remarks`/guard note closes the concern; if supported, the props-ignoring echo needs revisiting.

anthony-murphy:
`packages/test/local-server-tests/src/test/emptyRuntimeCodeLoader.spec.ts:448`

Deep Review: `spyOnOutbound` builds a fresh `IDocumentServiceFactory` literal with only `createContainer`/`createDocumentService` instead of proxying `inner`. The sibling service and connection wrappers in the same function use `new Proxy(...)` with `Reflect.get` fall-through, so any property the inner factory advertises — notably `ILayerCompatDetails` via `IProvideLayerCompatDetails` — is silently dropped. The test passes today only because the local-server factory advertises no compat details; the wrapper would not catch a driver-compat regression and could bypass or spuriously fail compat validation if reused against a factory that does.

Wrap `inner` in a `Proxy` with `Reflect.get` fall-through (mirroring the sibling wrappers in the same function) instead of constructing a fresh object literal, so unforwarded properties pass through.

anthony-murphy:
## Deep Review

*Reviewed commit `8c7187d` on 2026-07-21.*

No sign-off yet, but the PR is close. This purely-additive `@alpha` change (`createEmptyRuntimeCodeLoader`, `createEmptyRuntimeFactory`) mounts an existing container on a minimal no-op `IRuntime` — no correctness, security, or API-breakage concerns surfaced. Two low-cost polish items remain: the entry-point docstring and PR description overpromise "read the pending state", and there is no disconnect→reconnect test on the writable network path where prior work concentrated reconnect bugs.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Reword the `createEmptyRuntimeCodeLoader` TSDoc and the PR description's "read pending state" phrasing to state that `getPendingLocalState` echoes the pending state supplied on the container context and does not incorporate ops received after load; cross-link `captureFullContainerState` (#27220)
- [ ] Add a local-server test that disconnects and reconnects the empty container mid-session, asserting it stays open and emits zero ops/signals across the reconnect
- [ ] Take the PR out of draft once the remaining local-server reconnect coverage lands

### Context for Reviewers

- Newest member of the established container-loader "no-runtime / frozen-container" alpha series (#25513, #25538, #25590, #25653, #25742, #27211, #27220) — the `create*` free-function → `ICodeDetailsLoader`/factory shape matches the convention. Read it as an additive increment, not a new abstraction.
- The "no ops/signals" guarantee is runtime-layer only: it comes from the runtime never invoking `submit*`, while the container still connects as a normal (potentially writable) client at the network layer. Callers needing network-layer read-only enforcement should pair this with `createFrozenDocumentServiceFactory` (#25653).
- The `getPendingLocalState` opaque echo is the intentional, lossless design — echoing `context.pendingLocalState` unchanged preserves stashed offline edits (an integration test resumes a real runtime from the empty runtime's output and recovers both attached and offline edits).
- Suggested reviewers: ChumpChief / dannimad for the `@legacy @alpha` api-report surface; dannimad / markfields for the frozen-series interaction.

- **Needs human judgment** — the two-export decision (`createEmptyRuntimeCodeLoader` + `createEmptyRuntimeFactory`) the author raised on [an inline thread](https://github.com/microsoft/FluidFramework/pull/27714#discussion_r3605788896); ChumpChief should sign off on adding two exports to this alpha api-report against the prior "should become an overload" guidance (#25513).
- **Needs human judgment** — the committed api-report ships `createEmptyRuntimeCodeLoader(): ICodeDetailsLoader_2;`; the `_2` suffix is an API-Extractor disambiguation artifact against the package-local deprecated `ICodeDetailsLoader`. Confirm this transitional artifact is acceptable ([posted thread](https://github.com/microsoft/FluidFramework/pull/27714#discussion_r3626019825)).
- **Needs human judgment** — an `IRuntime` adapter lives in `@fluidframework/container-loader` rather than `@fluidframework/container-runtime`. Both classes are module-private, so relocation stays cheap, but an owner (ChumpChief / markfields) should sign off before the alpha surface locks in.
- **Cannot be assessed by the pipeline** — the "never sends ops or signals over the wire" guarantee is exercised only against a local server; a reviewer with runtime/protocol expertise should confirm no loader-layer system signal path could invalidate the empty-runtime attribution in production topologies.

</details>

- `fe5220b` 2026-07-21 · **8/10** — prior inline threads and signal-filter concern cleared; three polish/edge-case items flagged inline
- `d3d2ecb` 2026-07-17 · **8/10** — draft with promised local-server-tests coverage outstanding; polish on four inline threads

</details>

anthony-murphy:
`packages/test/local-server-tests/src/test/emptyRuntimeCodeLoader.spec.ts:459`

Deep Review: The empty container "connects as a normal (potentially writable) client at the network layer" rather than using a frozen delta stream, and #27211 documents that this exact configuration is where reconnect edge cases live — `WritableFrozenDeltaStream` minted a per-instance `clientId` to avoid `pendingStateManager` `0x173` (`replayPendingStates called twice for same clientId!`) on reconnect with dirty pending ops, and outbound writes were short-circuited before the read-mode reconnect branch, which under `ReconnectMode.Never` closes the container. This local-server test only exercises a single connect (`waitForContainerConnection`) and never disconnects/reconnects the empty container; the round-trip test disconnects the real container, not the empty one.

Add a local-server test that disconnects and reconnects the empty container while the session is live, asserting it stays open and still emits zero ops/signals across the reconnect. This hardens the PR's headline "never sends ops or signals" guarantee across the transition where prior work found regressions, and fits the in-progress local-server coverage noted in the PR description.

anthony-murphy:
`packages/loader/container-loader/src/emptyRuntime.ts:184`

Deep Review: The `createEmptyRuntimeCodeLoader` TSDoc — and the PR description's "reading the pending state of an existing container" — advertise that a host can read a container's pending state through this runtime. But `EmptyRuntime.getPendingLocalState` only returns `this.pendingLocalState`, captured once at construction from `context.pendingLocalState`; `process` is a no-op, so ops arriving after load are never reflected (the unit test asserts the return is `strictEqual` to the exact input object). #27220 introduced `captureFullContainerState` as the actual "capture fresh pending state including ops without a runtime" primitive, so a host that reads this docstring first can be steered into a misuse on an `@alpha` surface meant for adoption.

Reword the TSDoc (and the PR description) to state that `getPendingLocalState` echoes the pending state supplied on the container context and does not incorporate ops received after load, and cross-link `captureFullContainerState` (#27220) for hosts needing freshly-captured full state.

PR #27716, opened 2026-07-17
feat(client-tree): [dark] transaction builds minimization

jason-ha:
Adds logic required to minimize builds during transaction post-processing via `minimize` but does not make changes.

build minimization:
1. Split builds where root sequence is partially used.
2. Trim builds that are fully detached.
3. Rebuild partially used builds to only include the used portion.

[AB#76928](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/76928)


packages/dds/tree/src/core/rebase/types.ts
```diff
@@ -126,8 +126,15 @@ export function taggedOptAtomId(
 	return taggedAtomId(id, revision);
 }
 
+export function offsetChangesetLocalId(
```
yann-achard-MS, line 129: You can now replace the existing one in src/feature-libraries/modular-schema/defaultRevisionReplacer.ts #Closed

packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
```diff
@@ -3,31 +3,189 @@
...
- * future change.
+ * Every node created during the change contributes a `build`. Once the change is squashed, a build is only meaningful
+ * for nodes that remain attached in the resulting document. This function inspects the change's
+ * {@link intoDelta | delta} to determine which built nodes end up attached ("transient" nodes are those that do not),
+ * then:
+ *
+ * - drops any build whose nodes are entirely unused, and splits any partially-used build so that only the runs of used
+ * nodes are retained,
+ * - trims transient content nested within a surviving node's build tree, and
+ * - prunes destroys for the removed builds, since destroying a node that was never built has no effect.
```
yann-achard-MS: There are no destroys, right?

jason-ha: Correct - stale comment (asserted on line 166)

packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
```diff
@@ -3,31 +3,189 @@
...
+/**
+ * A set of node IDs, keyed by revision then by the numeric portion (`localId`/`minor`) of the ID.
+ */
+type ChangeAtomIdSet = NestedSet<RevisionTag | undefined, ChangesetLocalId>;
+
+/**
+ * Indexes a delta's {@link DeltaRoot.global | global} detached-node changes by their node ID.
+ *
+ * @remarks
+ * `DeltaRoot.global` describes modifications to nodes that are built and/or removed by the change,
```
yann-achard-MS:
```suggestion
 * `DeltaRoot.global` describes modifications to nodes that are built or preexist the change as detached roots,
```

packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
```diff
@@ -3,31 +3,189 @@
...
+ * @remarks
+ * These are the "used" detached nodes: any build whose nodes are not in this set has no observable
+ * effect on the resulting document and can be dropped.
+ *
+ * The delta's marks describe changes to a field: an `attach` brings a detached node into the live
+ * tree, while `mark.fields` describes edits to the *pre-existing* content of the cell (which is being
+ * detached/removed when `mark.detach` is set). Nested content of freshly built/attached nodes is not
+ * carried on the attach mark itself; it is expressed via `DeltaRoot.global` keyed by the node's ID.
+ * Consequently this walk only descends into a node's nested content when that node itself remains in
+ * the live tree, so that content attached beneath a removed node is correctly treated as unused.
```
yann-achard-MS: This is hard to follow and probably not 100% correct. I think I would rather not have this comment block. It shouldn't be the source of truth since it will just become stale.

packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
```diff
@@ -3,31 +3,189 @@
...
+	// pre-rename ID). Iterate to a fixed point.
+	while (worklist.length > 0) {
+		const next = worklist.pop();
+		if (next === undefined) {
+			break;
+		}
+		const { revision: major, localId: minor } = next;
+		visitLiveFields(globalById.get(major)?.get(minor));
+		if (delta.rename !== undefined) {
+			for (const { oldId, newId, count } of delta.rename) {
```
yann-achard-MS: Perf: We probably want to put the `delta.rename` data in a `RangeMap` so we don't iterate the list of renames over and over.

packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
```diff
@@ -3,31 +3,189 @@
...
+				if (newId.major === major && minor >= newId.minor && minor < newId.minor + count) {
+					markLive({
```
yann-achard-MS: When there is a rename, it probably makes sense to remove `newId` from the `attached` set.

packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
```diff
@@ -3,31 +3,189 @@
...
+		for (const { id, fields } of delta.global) {
+			setInNestedMap(globalById, id.major, id.minor, fields);
+		}
+	}
+	return globalById;
+}
+
+/**
+ * Collects the set of detached node IDs whose content ends up attached within the live document tree
+ * once the given change is applied.
```
yann-achard-MS: It't not clear from this description whether the set will include nodes that start out as attached (which it will, based on the current implementation). Would be good to call that out. #Closed

packages/dds/tree/src/feature-libraries/modular-schema/minimizeBuilds.ts
```diff
@@ -0,0 +1,197 @@
...
+ * @param isLive - Predicate reporting whether the node with the given ID ends up attached in the resulting document.
+ * @returns The minimized builds.
+ */
+export function computeMinimizedBuilds(
+	buildsIn: ChangeAtomIdBTree<TreeChunk>,
+	globalById: ChangeAtomIdMap<DeltaFieldMap>,
+	isLive: (id: ChangeAtomId) => boolean,
+): ChangeAtomIdBTree<TreeChunk> {
+	const buildsOut = newChangeAtomIdBTree<TreeChunk>();
+	const droppedBuildIds: ChangeAtomIdSet = new Map();
```
yann-achard-MS: `droppedBuildIds` is not consumed anywhere at the moment. I'm guessing that will change, but just in case, putting a comment on it.

packages/dds/tree/src/feature-libraries/modular-schema/minimizeBuilds.ts
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * @returns The number of children removed from the node's fields.
+ *
+ * @remarks
+ * `deltaFields` is the {@link DeltaFieldMap} describing the modifications made to the built node (as
+ * produced in the change's delta `global` section, keyed by the node's build ID). It is walked in
+ * lockstep with `node`'s fields: a `detach` mark whose target cell is not live removes the corresponding
+ * (transient) child from the tree, and a `fields` mark descends into a surviving child to trim its own
+ * transient descendants. Content brought in by `attach`-only marks lives in a separate build and is not
+ * inlined here, so those marks consume no existing child.
```
yann-achard-MS: I think this block comment has negative value.

packages/dds/tree/src/feature-libraries/modular-schema/minimizeBuilds.ts
```diff
@@ -0,0 +1,197 @@
...
+				childIndex += 1;
+			} else if (mark.attach === undefined) {
+				// Unchanged run of existing children.
+				for (let offset = 0; offset < mark.count; offset += 1) {
+					newChildren.push(children[childIndex] ?? oob());
+					childIndex += 1;
+				}
+			}
+			// Otherwise the mark only attaches new content from a separate build,
+			// which is not inlined in this node, so there is nothing to copy over.
```
yann-achard-MS: Not quite right: the mark attaches content but that content need not be from a separate build. It could instead be from a preexisitng detached root or a node that gets detached as part of the change.

packages/dds/tree/src/feature-libraries/modular-schema/minimizeBuilds.ts
```diff
@@ -0,0 +1,197 @@
...
+function mapTreeFromNodeChunk(chunk: TreeChunk): ExclusiveMapTree {
+	const cursor = chunk.cursor();
+	cursor.firstNode();
+	return mapTreeFromCursor(cursor);
+}
+
+/**
+ * Trims transient content from a built node's in-memory tree, in place.
+ *
+ * @returns The number of children removed from the node's fields.
```
yann-achard-MS:
```suggestion
 * @returns The number of subtrees removed from the tree's fields.
```

packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
jason-ha: Look at https://github.com/microsoft/FluidFramework/blob/b40ce86d76560d7617f2ac5d08c9f044dfc1d4dd/packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts where buildRenameGraph is introduced. Let me know if you like that infrastructure better all-up. I didn't bother bringing that change in here since this doesn't yet build up what is attached as needed for edit minimize.

packages/dds/tree/src/feature-libraries/modular-schema/minimizeBuilds.ts
jason-ha: Won't be needed - I thought I cleaned all those up. Good eye.

packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
jason-ha: Do we need to? I'm not sure why the word Detached is in here at all. I'm going to rename to collectAttachedNodeIds and otherwise remove mention of detached.

packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
jason-ha: Is that because builds are always supposed to use their original id?

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -2368,4 +2378,138 @@ describe("transaction minimize post-processor", () => {
...
+				}
+			}
+		};
+		for (const chunk of minimized.builds?.values() ?? []) {
+			const cursor = chunk.cursor();
+			cursor.firstNode();
+			for (let index = 0; index < chunk.topLevelLength; index += 1) {
+				checkNode(mapTreeFromCursor(cursor));
+				cursor.nextNode();
+			}
```
yann-achard-MS, line 2423: There are a bunch of helpers in src/core/tree/cursor.ts (e.g., `forEachNode`) that might help reduce boilerplate here.

jason-ha:
No longer think that this half implementation is worthwhile.

PR #27721, opened 2026-07-20
fix(ci): update labeler config files to actions/labeler v6 schema

dannimad:
## Description

The `pr-labeler.yml` workflow uses `actions/labeler@v6` (pinned via SHA, ratchet comment `v6`), but `.github/labeler-areas.yml` and `.github/labeler-changesets.yml` were still written in the older `actions/labeler@v5` schema, where each label maps directly to a `changed-files` object. v6 requires each label to map to a **list** of match config objects instead.

This schema mismatch made every run of the "Pull Request Labeler" workflow fail with:
```
found unexpected type for label 'area: build' (should be array of config options)
```
As a result, no `area: *` labels, `changeset-present`, or `changeset-required` labels have been applied automatically since ~Dec 2, 2025, and the workflow was subsequently disabled (`disabled_manually`) after the repeated failures. I've verified this against several recent merged PRs that touched `.changeset/*.md` files but never received the `changeset-present` label.

This PR fixes both YAML files to the v6 array schema and re-enables the workflow.

- Before: `gh run list --workflow=pr-labeler.yml` shows 100% failures with the annotation above.
- After: this PR itself is the test, since it modifies `.github/**`, which triggers `pr-labeler.yml`. Check the Actions tab on this PR — the "Label areas" and "Label changeset-required" jobs should now succeed, and this PR should pick up the `area: build` label.

## Reviewer Guidance

- Please confirm the "Pull Request Labeler" workflow run on this PR succeeds and applies the expected labels before merging.

### DIFF
--- a/.github/labeler-areas.yml
+++ b/.github/labeler-areas.yml
@@ -4,7 +4,7 @@
 # see how this config file is used.
 
 "area: build":
-  changed-files:
+  - changed-files:
     - any-glob-to-any-file:
       - ".github/*"
       - ".github/workflows/**"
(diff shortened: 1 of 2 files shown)


dannimad:
Note on the currently-failing "Pull Request Labeler" check on **this** PR: `actions/labeler` fetches its configuration file via the GitHub API using `ref: github.context.sha`, which for `pull_request_target` events resolves to the **base branch (`main`)**, not this PR's branch (this repo's `pr-labeler.yml` has no checkout step, so there's no local file override). That means this check will keep failing with the old error until this PR is merged into `main` — it's testing the *old* broken config, not the fix.

I validated the fix directly instead: I extracted the exact `getLabelConfigMapFromObject` validation function from `actions/labeler`'s bundled source at the pinned commit (`634933edcd8ababfe52f92936142cc22ac488b1b`, i.e. the exact version this workflow uses) and ran it locally against both updated YAML files — both now parse successfully (24 labels in `labeler-areas.yml`, 1 in `labeler-changesets.yml`), where before they threw `found unexpected type for label '...' (should be array of config options)`.

Recommended verification after merge: confirm the next PR touching `.github/**` or `.changeset/**` picks up the expected `area: *` / `changeset-present` labels.

Abe27342 (approved):
I haven't validated all the claims in this PR but they look plausibly correct and the fix looks reasonable.

PR description mentions that labeler actions are pinned to certain SHAs / versions. I'm assuming that means we bumped it at some point and didn't realize these changes were necessary?

(If the claim isn't true that they are pinned and instead the problem is that we got auto-updated, we should fix that by pinning them: not doing so is a supply chain risk)

PR #27722, opened 2026-07-20
Revert feed changes and add utility script for internal developers

Abe27342:
## Description

Latest guidance is that external developers should still use registry.npmjs.org to install packages.

This reverts #27705 and provides a reasonably convenient way for Microsoft developers to bootstrap a valid environment.

This should also unblock CI from dependency updates, as the mirror we use in those builds shouldn't have the same 7 day latency.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -227,7 +227,14 @@ peerDependencyRules:
 publicHoistPattern:
   - '@arethetypeswrong/cli'
 
-registry: https://packagefeedproxy.microsoft.io/npm/
+# No `registry` is set here on purpose. pnpm resolves the built-in default (registry.npmjs.org),
+# which keeps external/OSS contributions working out of the box, and also keeps lockfile updates
+# pointed at the public registry.
+# NOTE: a `registry` value in pnpm-workspace.yaml takes precedence over a project-level `.npmrc`,
+# so setting it here (even to registry.npmjs.org) would silently defeat the per-developer override
+# below. Leave it unset instead.
+# Developers who must use an npmjs mirror or team-managed feed can make use of
+# `node ./scripts/set-dev-registry.cjs <registry-url>`.
 
 resolutionMode: highest
 
@@ -235,7 +242,8 @@ strictDepBuilds: true
 
 strictPeerDependencies: true
 
-# The registry.npmjs mirror does not preserve all the npm package metadata necessary for this pnpm feature to function
+# The artifact feeds used by CI and by internal developers (see the registry note above)
+# do not preserve all the npm package metadata necessary for this pnpm feature to function
 # correctly as of 7/16/2026. If this is turned back on...
 # See: https://github.com/orgs/pnpm/discussions/11084 for some discussion.
 # Enabling no-downgrade requires every transitive trust-policy violation to either be remediated
--- a/scripts/set-dev-registry.cjs
+++ b/scripts/set-dev-registry.cjs
@@ -41,13 +41,9 @@ const repoRoot = path.resolve(__dirname, "..");
 // Directory names that should never be traversed while discovering workspaces.
 const SKIP_DIR_NAMES = new Set(["node_modules", ".git"]);
 
-// Path fragments that identify workspaces which are test fixtures or managed separately with their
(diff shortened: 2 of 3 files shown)


pnpm-workspace.yaml
```diff
@@ -227,15 +227,23 @@ peerDependencyRules:
...
 
-registry: https://packagefeedproxy.microsoft.io/npm/
+# No `registry` is set here on purpose. pnpm resolves the built-in default (registry.npmjs.org),
+# which keeps external/OSS contributions working out of the box, and also keeps lockfile updates
+# pointed at the public registry.
+# NOTE: a `registry` value in pnpm-workspace.yaml takes precedence over a project-level `.npmrc`,
+# so setting it here (even to registry.npmjs.org) would silently defeat the per-developer override
+# below. Leave it unset instead.
+# Developers who must use an npmjs mirror or team-managed feed can make use of
+# `node ./scripts/set-dev-registry.cjs <registry-url>`.
```
CraigMacomber, line 237: This approach needs the `minimumReleaseAge: 1440` above to be changed to 0 for managed/internal users and, and also 7 days (with exceptions for our own packages set to 1440) for external users. This is impossible I think, so we might just want to set it to 0 for now as nearly all dependency impacting changes are from first party devs and they will be using a repo with strictly more delay than that CI requires?

scripts/set-dev-registry.cjs
```diff
@@ -0,0 +1,182 @@
...
+	if (entries.some((e) => e.isFile() && e.name === "pnpm-workspace.yaml")) {
+		found.push(dir);
+	}
+
+	for (const entry of entries) {
+		if (!entry.isDirectory() || SKIP_DIR_NAMES.has(entry.name)) {
+			continue;
+		}
+		const child = path.join(dir, entry.name);
+		if (EXCLUDED_FRAGMENTS.some((fragment) => child.includes(fragment))) {
```
CraigMacomber: Wouldn't it be simpler and more robust and faster to have a single list of SKIP_DIR_NAMES, and not use this EXCLUDED_FRAGMENTS approach? Basic string containment seems likely to have false matches. What if we had a release group for test-tooling, or something where "src" ended up in the middle or a word or something.

Is there a reason we have two different check mechanisms here?

scripts/set-dev-registry.cjs
```diff
@@ -0,0 +1,182 @@
...
+			const existing = fs.readFileSync(npmrcPath, "utf8");
+			if (!existing.startsWith(MARKER)) {
+				console.warn(`Skipping ${rel}/.npmrc (not managed by this script).`);
+				continue;
+			}
+			fs.rmSync(npmrcPath);
+			console.log(`Removed ${rel}/.npmrc`);
+			changed++;
+		} else {
+			const existing = fs.existsSync(npmrcPath)
```
CraigMacomber: This could be hoisted out of the if to deduplicate the existence check

scripts/set-dev-registry.cjs
Abe27342: note that the ignore list as implemented protects against the entire string "build-intrascture/src/test/data" (well, that or with backslashes), not any of the individual parts.

That said, we probably don't actually need to exclude the test workspaces in build-infrastructure, since if tests install dependencies in those on local machines we'd want similar registry config there as well... so I've removed the second mechanism and just dropped that bit completely.

pnpm-workspace.yaml
Abe27342: Hmm, good point that we probably need to update this setting.

I'd prefer tackling this in a follow-up since looks like all the pnpm-workspaces have it.

Abe27342: https://github.com/microsoft/FluidFramework/pull/27732 for that. Does the rationale there match your understanding? It's been a while since I played with `minimumReleaseAge`. If we're pretty confident this will be a problem, makes sense to check that in now. If we're less confident I can try and do some more local testing and decide if it will be a problem in practice

CraigMacomber (approved):

CraigMacomber (approved):
Would be good to note in the description that this should unbreak CI for dependency changes.

Issue #27180, opened 2026-04-27
Make minVersionForCollab required


scottn12:
## Broken API or pattern

Optional `minVersionForCollab` on the runtime entry points. Currently `LoadContainerRuntimeParams.minVersionForCollab` accepts `undefined` and the runtime falls back to `defaultMinVersionForCollab`. This should be made a required param.

Patterns to remove:

- `minVersionForCollab` being optional on `LoadContainerRuntimeParams`.
- `defaultMinVersionForCollab` when the customer does not set the value (will no longer be needed).

**Preparation References**

N/A since this is not a deprecated API

**Packages**

- `@fluidframework/container-runtime` - `LoadContainerRuntimeParams.minVersionForCollab` (and the default that resolves it to `defaultMinVersionForCollab`).
- `@fluidframework/runtime-utils` - remove `defaultMinVersionForCollab`.

**Expected Timeline**

Make required in 3.0.0

PR #27726, opened 2026-07-20
Add minDocumentRuntimeVersion compatibility aliases

WillieHabi:
## Description

Adds `minDocumentRuntimeVersion` as the preferred, additive name for configuring container-runtime cross-client compatibility while preserving compatibility with `minVersionForCollab`.

This focused replacement for #27656:

- Exports the public `MinDocumentRuntimeVersion` type alias.
- Adds the optional `LoadContainerRuntimeParams.minDocumentRuntimeVersion` beta property.
- Deprecates `LoadContainerRuntimeParams.minVersionForCollab` in 2.114.0, with removal planned for the 3.10.0 beta break under #27180.
- Rejects calls that specify both parameter names.
- Preserves existing persisted-schema and telemetry names.
- Ensures the Attributor runtime wrapper forwards the new parameter.
- Updates focused tests, API reports, documentation, and release notes.

`MinimumVersionForCollab` remains supported and is not deprecated in this PR. Tree, datastore, service-client, and test-utility APIs are intentionally outside its scope.

### DIFF
--- a/.changeset/bright-tigers-retire.md
+++ b/.changeset/bright-tigers-retire.md
@@ -1,0 +1,10 @@
+---
+"@fluidframework/container-runtime": minor
+"__section": deprecation
+---
+
+The minVersionForCollab compatibility alias is deprecated
+
+`LoadContainerRuntimeParams.minVersionForCollab` is deprecated in 2.114.0 in favor of `minDocumentRuntimeVersion` and is planned for removal in 3.10.0.
+
+See [#27180](https://github.com/microsoft/FluidFramework/issues/27180) for the cleanup plan.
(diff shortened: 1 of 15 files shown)


WillieHabi:
Closing based on the compatibility-layer review. The proposed minDocumentRuntimeVersion alias would imply a runtime-specific contract, while this value is intentionally propagated through the Runtime and DataStore layers and consumed by DDSes such as SharedTree for format and feature selection. We will retain minVersionForCollab / MinimumVersionForCollab for now and revisit naming only if the cross-layer contract is redesigned or split into layer-specific concepts. This does not change the independent 3.0 work in #27180 to make the existing parameter required.

PR #27730, opened 2026-07-20
test(client-tree): sentinel tests for yet to be supported scenarios

jason-ha:
Adds test cases that test limits of edits to establish boundary for minimize test cases should cover.

### DIFF
--- a/packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
+++ b/packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
@@ -554,22 +554,22 @@ const arrayScenarios = {
 	} as const,
 
 	/**
-	 * Starts from pre-existing content `["X", "Y"]` and inserts a transient "A☠️" that is removed before the
+	 * Starts from pre-existing content `["X🕰️", "Y🕰️"]` and inserts a transient "A☠️" that is removed before the
 	 * transaction closes, leaving the document unchanged.
 	 * @remarks
-	 * The pre-existing nodes "X" and "Y" are not created by this transaction, so they contribute no builds to its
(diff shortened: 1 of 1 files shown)


packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -1075,6 +1075,52 @@ describe("transaction minimize post-processor", () => {
...
+
+		// If any of these tests start to fail, the system has new capabilities
+		// and additional scenarios should be added to verify minimize handles
+		// them correctly.
+		describe("existing content re-inserted raises exception", () => {
+			it("nesting original box under new parent in array", () => {
+				assert.throws(() => {
+					const { view, stringifiedChange } = runScenario({
+						schema: BoxArray,
+						initialContent: [new Box({ value: "A☠️" })],
```
yann-achard-MS: Why the transient marker on A since it is not transient?
(Same below)

jason-ha: Well, in this case we'd never expect the content to show up in the change - it predates change.
Could use a third emoji and check that it doesn't show. Wasn't feeling the value, but can.?

yann-achard-MS: It just seems inconsistent with the other tests which have preexisting content that isn't labeled this way. I would be happy with any of the following options:
* Remove the marker
* Rename the marker to a more generic name
* Introduce a new marker and use it for all preexisting content

jason-ha: new marker employed

yann-achard-MS (approved):

PR #27731, opened 2026-07-21
docs: Port remaining github wiki documentation to `docs` directory

Josmithr:
Also enables a couple more linter rules to enforce common structure (single top-level heading as document title).

Contents from the wiki were copied as-is, save for the following changes:
- Links were updated to use relative file paths where appropriate
- Line formatting was updated to follow our guidelines
- All documents were given top-level headings
- Other misc. formatting/linting violations were fixed per the linter configurations
- An obsolete document on our breaking changes policy (pre-2.0) was removed, and redirect notice pages were removed

This PR also sets up LFS for binary asset storage, with associated documentation.


docs/content/Guidelines/Documentation-Guidelines/Where-To-Put-Documentation.md
```diff
@@ -17,7 +17,7 @@ They are not presented in any particular order.
 - [fluidframework.com](https://fluidframework.com/)
     - This is our customer-facing website.
       It is the public face of our product and is home to high-level overviews, samples, etc. to get users up-and-running on the framework.
-- [repo wiki](https://github.com/microsoft/FluidFramework/tree/main/docs) (here)
+- [repo wiki](../../Home.md) (here)
```
alexvy86:
```suggestion
- [repo itself](../../Home.md) (here)
```

alexvy86 (commented):
One small thing below. I'm also blindly trusting the new md files.

The bigger thing I'd bring up for discussion is the image assets. I think our repo used to use git-lfs way back (maybe for what today became the snapshot tests content submodule), but AFAIK it doesn't use it today. Should we re-evaluate that? It's not a huge number of images that we'd be committing here, but as the docs evolve the storage footprint could get bigger.

Josmithr:
> I'm also blindly trusting the new md files.

I added notes to the PR description about how the wiki contents were copied.

> The bigger thing I'd bring up for discussion is the image assets. I think our repo used to use git-lfs way back (maybe for what today became the snapshot tests content submodule), but AFAIK it doesn't use it today. Should we re-evaluate that? It's not a huge number of images that we'd be committing here, but as the docs evolve the storage footprint could get bigger.

Yeah, this thought occurred to me as well. Wasn't sure if we wanted to incur the extra complexity for these few images or not. I could go either way. We can always make that change in the future if we want.

alexvy86:
Just for reference, I chatted with Josh and I'm still a bit concerned about the binary files (images), and whether we should start using git-lfs for this repo again, which seems a question worth addressing if we plan to have docs here instead of the wiki.

alexvy86 (approved):

PR #27733, opened 2026-07-21
Improve website API search results

MarioJGMsoft:
## Description

Generated API symbols were difficult to find reliably because Pagefind indexed display titles such as `SchemaFactory Class` rather than canonical symbol names. Search result badges also inherited incorrect colors because their custom classes used Pagefind's reserved `pf-` prefix.

This change adds API symbol names as searchable Pagefind metadata and marks punctuation- and case-insensitive exact symbol matches in the results. It also updates Pagefind-specific CSS documentation and moves custom result badges outside Pagefind's reserved class namespace so version and exact-match labels respect light and dark themes.

Search coverage now includes generated API results and exact matches across classes, interfaces, type aliases, namespaces, and enums.

### Before

Website search before exact API matches and themed result badges

### After

Website search showing exact API matches with themed result badges

## Reviewer Guidance

The exact-match behavior intentionally strips punctuation before comparing the query with the canonical API symbol name, so queries such as `SchemaFactory/` still surface the exact match.

Fixes [AB#77406](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/77406)


website/src/css/custom.scss
```diff
@@ -16,6 +16,8 @@
 	--ifm-code-font-size: 95%;
 	--docusaurus-highlighted-code-line-bg: rgba(0, 0, 0, 0.1);
 	--docsearch-muted-color: #666c77;
+	--pf-exact-match-background: var(--ifm-color-primary);
```
WayneFerrao: These new custom properties use the `--pf-*` prefix, which is Pagefind's reserved theming namespace
(--pf-text etc. below). 
This PR's whole point is moving custom styling out of the reserved pf- namespace — the CSS vars reintroduce the same collision risk. I'd suggest --api-exact-match-background / --api-exact-match-text as alternatives

WayneFerrao (commented):
Great improvement!

website/src/css/custom.scss
MarioJGMsoft: Applied changes

WayneFerrao (approved):

WayneFerrao (approved):

PR #27734, opened 2026-07-21
Fix multiple insert assert

CraigMacomber:
## Description

Fix bug where `array.insertAtEnd(item, item);` would hit assert  `0xa2b` instead of a `UsageError`.

[AB#77712](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/77712)

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
+++ b/packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
@@ -13,8 +13,8 @@ import {
 import { asAlpha } from "../../../api.js";
 import {
 	SchemaFactory,
+	TreeArrayNode,
 	TreeViewConfiguration,
-	type TreeArrayNode,
 	type TreeArrayNodeAlpha,
 	type FixRecursiveArraySchema,
 	type InsertableTreeFieldFromImplicitField,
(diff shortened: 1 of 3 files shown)


jason-ha (approved):

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -94,6 +94,63 @@ describe("ArrayNode", () => {
...
+			const item = new Item({});
+			const array = init(ItemArray, [item]);
+			assert.throws(
+				() => array.insertAtEnd(item),
+				validateUsageError(
+					hydrated
+						? // The case of hydrating a node has extra context and stricter validation using that context.
+							// One sideeffect of that is we give nicer errors.
+							`A node with schema "ArrayNodeTest.Item" (name: "Item") was inserted into the tree more than once. This is not supported.`
+						: "A node may not be in more than one place in the tree",
```
Josmithr, line 136: Nit (and not necessarily for this PR): could we update the error message in the unhydrated case to offer more context to the user? The specificity gap between these 2 error messages is pretty wide.

Josmithr (approved):

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
CraigMacomber: That code doesn't have access to the context from which it gets the view schema, so I couldn't find a practical way to do that. Also layering wise we generally try and not have the unhydrated tree depend on the simple tree, and that tends to create dependency cycles.

PR #27737, opened 2026-07-22
Remove unreachable branch in getCreateDetailsForChild FromAttach case

agarwal-navin:
## Description

Removes a dead `if` block in `SummarizerNode.getCreateDetailsForChild` for the
`CreateSummarizerNodeSource.FromAttach` case. The assignment to
`childLastSummaryReferenceSequenceNumber` inside that block is unreachable.

`parentLastSummaryReferenceSequenceNumber` refers to the latest sequence number that was processed in the last successful summary. So, for `createParam.sequenceNumber <= parentLastSummaryReferenceSequenceNumber` to be true, the attach message would have to have been sequenced before the last summary and somehow not be part of that summary. This is not possible.

The reason this check exists is because of a legacy feature called "failure summaries" where if a data store failed to summarize, the summarize node would include its previous summary + pending ops into the current summary. In the next summary, when this data store was being summarized, the pending ops would be processed and the above condition could be true if there was an attach op for a DDS in that data store.

That feature was deleted years ago, so this check is no longer needed and only adds to confusion. Converted that check to as assert to validate the same. If this condition does happen somehow, we would see the assert.

### DIFF
--- a/packages/runtime/container-runtime/src/summary/summarizerNode/summarizerNode.ts
+++ b/packages/runtime/container-runtime/src/summary/summarizerNode/summarizerNode.ts
@@ -570,13 +570,6 @@ export class SummarizerNode implements IRootSummarizerNode {
 		const parentLastSummaryReferenceSequenceNumber = this._lastSummaryReferenceSequenceNumber;
 		switch (createParam.type) {
 			case CreateSummarizerNodeSource.FromAttach: {
-				if (
-					parentLastSummaryReferenceSequenceNumber !== undefined &&
-					createParam.sequenceNumber <= parentLastSummaryReferenceSequenceNumber
-				) {
-					// Prioritize latest summary if it was after this node was attached.
-					childLastSummaryReferenceSequenceNumber = parentLastSummaryReferenceSequenceNumber;
-				}
 				changeSequenceNumber = createParam.sequenceNumber;
 				break;
 			}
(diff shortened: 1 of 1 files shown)


packages/runtime/container-runtime/src/summary/summarizerNode/summarizerNode.ts
```diff
@@ -570,13 +570,15 @@ export class SummarizerNode implements IRootSummarizerNode {
...
-					// Prioritize latest summary if it was after this node was attached.
-					childLastSummaryReferenceSequenceNumber = parentLastSummaryReferenceSequenceNumber;
-				}
+				// `parentLastSummaryReferenceSequenceNumber` is the latest sequence number processed in the last
+				// successful summary. So, for `createParam.sequenceNumber <= parentLastSummaryReferenceSequenceNumber`
+				// to be true, the attach message would have to have been sequenced before the last summary and somehow
+				// not be part of that summary, which is not possible.
+				assert(
+					parentLastSummaryReferenceSequenceNumber === undefined ||
+						createParam.sequenceNumber > parentLastSummaryReferenceSequenceNumber,
```
markfields, line 579: What if they're equal? Seems like that should be ok and allowed by the assert (maybe the old check should have been a strict inequality)

agarwal-navin: It cannot be equal. Equal still means this op would have been in the summary.

markfields: Oh, yes of course, I was mixed up

markfields (approved):

jzaffiro (approved):

PR #27743, opened 2026-07-22
test(client-tree): test minimize content move scenarios

jason-ha:
Add additional cases that explicitly rearrange content (moves) that is detached at end of transaction.

### DIFF
--- a/packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
+++ b/packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
@@ -1031,10 +1031,10 @@ const objectScenarios = {
 	 *
 	 * 0. initial                                      -\> `Box: <empty>`
 	 * 1. insert nested Box with tags ["x☠️", "y☠️"]  -\> `Box: { nested: Box: { tags: ["x☠️", "y☠️"] } }`
-	 * 2. remove tag at 0                              -\> `Box: { nested: Box: { tags: ["y☠️"] } }`
-	 * 3. remove root box                              -\> `undefined`  |: `Box: { nested: Box: { tags: ["y☠️"] } }`
+	 * 2. remove tag at 0                              -\> `Box: { nested: Box: { tags: ["y☠️"] } }`         |: "x☠️"
+	 * 3. remove root box                              -\> `undefined`                                       |: "x☠️", `Box: { nested: Box: { tags: ["y☠️"] } }`
 	 *
-	 * Classification: x☠️ and y☠️ come in as new nested content and leave as nested under [detached] prior node (same parent)
+	 * Classification: x☠️ and y☠️ come in as new nested content and leave as detached root and nested under [detached] prior node (same parent), respectively
 	 */
 	nest_Box_with_tags_then_root_Box_removed: {
 		schema: OptionalBox,
--- a/packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
+++ b/packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
@@ -79,6 +79,7 @@ const RootStringArray = sf.array("RootArray", sf.string);
 class Box extends sf.objectRecursive("Box", {
 	value: sf.optional(sf.string),
 	nested: sf.optionalRecursive([() => Box]),
+	tags: sf.optional(sf.array("tags", sf.string)),
 }) {}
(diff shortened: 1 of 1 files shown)


packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -962,6 +964,118 @@ const objectScenarios = {
...
+	 * Steps:
+	 *
+	 * 0. initial           -\> `Box: { nested: Box: { tags: ["a🕰️"] } } }`
+	 * 1. insert tag "x☠️"  -\> `Box: { nested: Box: { tags: ["x☠️", "a🕰️"] } }`
+	 * 2. insert tag "y☠️"  -\> `Box: { nested: Box: { tags: ["y☠️", "x☠️", "a🕰️"] } }`
+	 * 3. remove root box   -\> `undefined`  |: `Box: { nested: Box: { tags: ["y☠️", "x☠️", "a🕰️"] } }`
+	 *
+	 * Classification: x☠️ and y☠️ come in as new roots and leave as nested under [detached] prior node
+	 */
+	nested_Box_tag_inserted_then_another_tag_inserted_causing_move_and_root_Box_removed: {
```
yann-achard-MS: The name for this scenario indicates that a move is occurring, but there is no movement in this scenario.
Are you perhaps operating under the assumption that inserting `y` moves `x` from index 0 to index 1? That does not count a moving `x` in our system. What happens is that a new cell is introduced for `y` before the existing cell that contains `x`. #Closed

packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts
```diff
@@ -1303,6 +1417,32 @@ describe("transaction minimize post-processor", () => {
...
+			assert.ok(view.root.nested);
+			assert.equal(view.root.nested.tags, undefined);
+		});
+
+		it("reflects an undefined root when a nested object with tags is added and then the root object is removed", () => {
+			const { view } = runScenario(objectScenarios.nest_Box_with_tags_then_root_Box_removed);
+			assert.equal(view.root, undefined);
+		});
+
+		it("reflects both tags of a newly inserted root object when a surviving tag is inserted before the existing one", () => {
```
yann-achard-MS:
```suggestion
		it("reflects the replacement tag of a newly inserted root object when the existing tag is removed and a new one is inserted", () => {
``` #Closed

File: packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts [hunk 3]
```diff
@@ -1303,6 +1417,32 @@ describe("transaction minimize post-processor", () => {
...
+			);
+			assert.equal(view.root, undefined);
+		});
+
+		it("reflects an empty nested tags field when tags are inserted, rearranged, and then removed", () => {
+			const { view } = runScenario(
+				objectScenarios.nested_Box_tags_inserted_then_tags_rearranged_and_removed,
+			);
+			assert.ok(view.root);
+			assert.ok(view.root.nested);
```
yann-achard-MS:
```suggestion
			assert.ok(view.root.nested);
			assert.equal(view.root?.nested?.value, "z❤️");
``` #Closed

File: packages/dds/tree/src/test/shared-tree/transactionMinimize.spec.ts [hunk 4]
```diff
@@ -1703,6 +1843,47 @@ describe("transaction minimize post-processor", () => {
...
+		});
+
+		it("carries no build when nested tags are inserted, rearranged, and then removed", () => {
+			const { view, stringifiedChange } = runScenario(
+				objectScenarios.nested_Box_tags_inserted_then_tags_rearranged_and_removed,
+			);
+			assert.doesNotMatch(stringifiedChange, transientMarkerRegex);
+			const change = getHeadChange(view);
+			// All inserted tags are removed, so no created node is present in the final document and no builds should remain.
+			assert.deepEqual(countBuilds(change), { builds: 0, tops: 0 });
```
yann-achard-MS:
```suggestion
			// All inserted tags are removed, so only the nested value is present in the final document and only that build should remain.
			assert.deepEqual(countBuilds(change), { builds: 1, tops: 1 });
``` #Closed

yann-achard-MS (approved):

PR #27747, opened 2026-07-23
feat(core-interfaces): add FluidReadonlyArray type

jenn-le:
## Description

Add `FluidReadonlyArray<T>` to `@fluidframework/core-interfaces` as a TypeScript-version-independent equivalent of the built-in `ReadonlyArray` type, following the same pattern as the existing `FluidReadonlyMap` and `FluidMap` types.

This includes stable methods through ES2023 (`at()`, `findLast()`, `findLastIndex()`) but deliberately excludes newer copy-on-write methods (`toReversed()`, `toSorted()`, `toSpliced()`, `with()`) that our implementations do not yet provide. This ensures Fluid Framework types remain safe to implement without being broken by TypeScript updates that add new members to `ReadonlyArray`.

The type is introduced at `@beta` and re-exported from `fluid-framework`. It will be used in a follow-up PR to replace the `extends ReadonlyArray<T>` in `ReadonlyArrayNode`/`TreeArrayNode`.

## Reviewer Guidance

[Reviewer guidance wiki](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#reviewer-guidance)

- The new `FluidReadonlyArray` interface is in `packages/common/core-interfaces/src/fluidArray.ts`
- Shared type test utilities were extracted to `src/test/types/typeTestUtils.ts` (also used by existing `fluidMapTypes.ts`)
- Type tests verify assignability relationships under the repo's ES2020 lib setting


packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
```diff
@@ -0,0 +1,84 @@
...
+import type { isAssignableTo, requireFalse, requireTrue } from "./typeTestUtils.js";
+
+// Under this repo's ES2020 lib, native ReadonlyArray lacks at()/findLast()/findLastIndex()
+// that FluidReadonlyArray requires, so it is NOT assignable to FluidReadonlyArray.
+// (Under ES2023+ libs, this would be true instead.)
+declare type _readonlyArray_to_fluidReadonlyArray = requireFalse<
+	isAssignableTo<readonly string[], FluidReadonlyArray<string>>
+>;
+
+// FluidReadonlyArray is NOT assignable to ReadonlyArray due to minor structural differences
```
CraigMacomber: This makes adopting this type a breaking chain which could be quite problematic for users to adopt.

I think we need to fix that.

packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
```diff
@@ -0,0 +1,84 @@
...
+// in Symbol.unscopables typing. This is acceptable — the important direction is that
```
CraigMacomber: Can you put the sentence breaks not in the middle of lines, it makes commenting on sentences hard.

packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
```diff
@@ -0,0 +1,84 @@
...
+// native arrays (which ARE ReadonlyArray) remain assignable to FluidReadonlyArray under ES2023+.
```
CraigMacomber: This is something we specifically do not support users doing, so having it work is probably more harmful then useful. Our type is sealed documenting that this is no allowed: we want this type to be one which only we are allowed to implement it so we can add things to it as needed, such as new API s that get added in 2-35 or whatever which would break code assigning 2023 arrays to our sealed type.

packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
```diff
@@ -0,0 +1,84 @@
...
+declare type _fluidReadonlyArray_to_readonlyArray = requireFalse<
+	isAssignableTo<FluidReadonlyArray<string>, readonly string[]>
+>;
+
+// Native mutable Array is NOT assignable to FluidReadonlyArray under ES2020
+// (missing at/findLast/findLastIndex).
+declare type _array_to_fluidReadonlyArray = requireFalse<
```
CraigMacomber: This, and the above non-assignability checks are not very useful. There are tons of reasons types might not be assignable ( if we implemented our array type as something totally wrong, like `number` those tests would all still pass, and give 0 useful information).

Additionally this test really tells us pretty much nothing, and also will break when we update to a newer es standard, which we plan to do soon. A test which doesn't validate anything we care about, and will have to be modified in the future has negative value.

packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
CraigMacomber: I think the main requirement for this change is that we make this true: as long as its not passing, we don't have a good way to validate that we got our interface correct, and it will break customers which uses our arrays as readonly arrays (which used to be supported) even with typescript configurations where  that worked before.

What we want to get out of this change is to keep such users working, but make the case where they change their tsconfig and start getting errors to report the errors in their code and not ours.

packages/common/core-interfaces/src/test/types/typeTestUtils.ts
```diff
@@ -0,0 +1,25 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Compile-time type utilities for manual type tests in this package.
+ * These mirror the pattern used in the auto-generated type test files.
```
CraigMacomber: Can't this package just import those the same way the generated type tests do?

packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
```diff
@@ -0,0 +1,84 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import type { FluidIterableIterator, FluidReadonlyArray } from "../../index.js";
```
CraigMacomber: Looks like these are the only test changes here: you should also update the tests in import-testing which are skipped due to the issue you are fixing.

dannimad (commented):

packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
jenn-le: It looks like these won't be fixed until TreeArrayNode is updated to use FluidReadonlyArray which would be a breaking change

.changeset/add-fluid-readonly-array.md
```diff
@@ -0,0 +1,9 @@
+---
+"@fluidframework/core-interfaces": minor
+"fluid-framework": minor
+"__section": Add FluidReadonlyArray type
```
CraigMacomber: I don't think that's a valid section

.changeset/add-fluid-readonly-array.md
```diff
@@ -0,0 +1,9 @@
...
+---
+
+Add `FluidReadonlyArray<T>` interface to `@fluidframework/core-interfaces` and re-export from `fluid-framework`.
+
+`FluidReadonlyArray<T>` provides a TypeScript-version-independent equivalent of the built-in `ReadonlyArray` type, following the same pattern as `FluidReadonlyMap` and `FluidMap`. The interface includes stable methods through ES2023 (`at()`, `findLast()`, `findLastIndex()`) but excludes newer copy-on-write methods (`toReversed()`, `toSorted()`, `toSpliced()`, `with()`) that Fluid Framework implementations don't yet support. This ensures these types remain safe to implement without TypeScript updates breaking them.
```
CraigMacomber: It's not just typescript version, but also typescript configuration that can impact this (it would be just typescript configuration if not for targets like ESNext which can change with version). I also wouldn't say its configuration independency since other parts of the config still impact this.

To be specific, I think we should say "independent of TypeScript [`lib`](https://www.typescriptlang.org/tsconfig/#lib)".

This more specific phrasing should probably be used in a few places (here, the doc comment etc)

packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
CraigMacomber: I guess those changes will be in a separate PR for 3.0 then?

.changeset/add-fluid-readonly-array.md
jenn-le: nice catch, fixed :)

packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
```diff
@@ -0,0 +1,66 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import type { requireAssignableTo } from "@fluidframework/build-tools";
+import type { FluidIterableIterator, FluidReadonlyArray } from "../../index.js";
+
+// FluidReadonlyArray is assignable to readonly T[].
+// This is the key invariant: users who treat our array nodes as readonly arrays must not break.
```
CraigMacomber: This isn't really true.

```suggestion
// Ensure FluidReadonlyArray is assignable to readonly T[] with our current TypeScript configuration.
// Technically customers should not rely on this as doing so makes their code likely
// to break if they change the `lib` in their TypeScript configuration.
// This is however a good way to ensure we are including everything we should,
// and that the transition from us declaring we implement `readonly T[]` to `FluidReadonlyArray` is unlikely to break code
// which matches our current `lib` selection.
```

packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
jenn-le: Yup, I'm also planning to promote the interfaces to public in a separate pr before 3.0. Let me know if you think that should just be part of 3.0 though.

packages/common/core-interfaces/src/test/types/fluidArrayTypes.ts
```diff
@@ -0,0 +1,66 @@
...
+declare type _fluidReadonlyArray_to_readonlyArray = requireAssignableTo<
+	FluidReadonlyArray<string>,
+	readonly string[]
+>;
+
+// FluidReadonlyArray iteration methods return FluidIterableIterator
```
CraigMacomber: I don't think this check actually works. If keys returned the TypeScript built in iterator, its decently likely that iterator type is assignable to ours, which would allow this to pass even if what it claims to test is false.

Realistically I think the way to check "FluidReadonlyArray iteration methods return FluidIterableIterator" is the API report, not a type test.

```suggestion
// FluidReadonlyArray iteration methods return types compatible with FluidIterableIterator.
// They should explicitly return `FluidIterableIterator` but that can't be tested here as it's a structural type.
// That can instead be validated in the api reports.
```

packages/common/core-interfaces/src/test/types/fluidMapTypes.ts
```diff
@@ -3,92 +3,88 @@
...
-	isAssignableTo<FluidMap<string, number>, FluidReadonlyMap<string, number>>
+declare type _fluidMap_to_fluidReadonlyMap = requireAssignableTo<
+	FluidMap<string, number>,
+	FluidReadonlyMap<string, number>
 >;
 
 // FluidReadonlyMap is assignable to ReadonlyMap (the extra Symbol.toStringTag is compatible).
-declare type _fluidReadonlyMap_to_readonlyMap = requireTrue<
-	isAssignableTo<FluidReadonlyMap<string, number>, ReadonlyMap<string, number>>
+declare type _fluidReadonlyMap_to_readonlyMap = requireAssignableTo<
```
CraigMacomber, line 26: This should get the same clarifications I suggested for the array version.

packages/common/core-interfaces/src/test/types/fluidMapTypes.ts
```diff
@@ -3,92 +3,88 @@
...
 // Native iterables are assignable to FluidIterable
-declare type _map_to_fluidIterable = requireTrue<
-	isAssignableTo<Map<string, number>, FluidIterable<[string, number]>>
->;
-declare type _set_to_fluidIterable = requireTrue<
-	isAssignableTo<Set<string>, FluidIterable<string>>
->;
-declare type _array_to_fluidIterable = requireTrue<
-	isAssignableTo<string[], FluidIterable<string>>
+declare type _map_to_fluidIterable = requireAssignableTo<
```
CraigMacomber, line 42: I don't think we should actually require either of these, or at the very least they should get the disclaimers like I suggested for array that this is a sanity check, not an invariant users can depend on. I think its perfectly valid for us to implement standard APIs which get added to the build in types before we update our typescript lib to include them (to help users with newer libs than use) which would cause these checks to fail.

packages/common/core-interfaces/src/test/types/fluidMapTypes.ts
```diff
@@ -3,92 +3,88 @@
...
-		ReturnType<Map<string, number>["entries"]>,
-		FluidIterableIterator<[string, number]>
-	>
+declare type _mapEntries_to_fluidIterableIterator = requireAssignableTo<
+	ReturnType<Map<string, number>["entries"]>,
+	FluidIterableIterator<[string, number]>
 >;
-declare type _setKeys_to_fluidIterableIterator = requireTrue<
-	isAssignableTo<ReturnType<Set<string>["keys"]>, FluidIterableIterator<string>>
+declare type _setKeys_to_fluidIterableIterator = requireAssignableTo<
```
CraigMacomber: If the built in iterator can be assigned to our iterator (which we check above), then these check seem worthless as it would pass even if our keys method returned the built in type.

.changeset/add-fluid-readonly-array.md
```diff
@@ -0,0 +1,11 @@
+---
+"@fluidframework/core-interfaces": minor
+"fluid-framework": minor
+"__section": feature
+---
+
```
Josmithr:
```suggestion
```

packages/common/core-interfaces/src/fluidArray.ts
```diff
@@ -0,0 +1,353 @@
...
+	 */
+	[Symbol.iterator](): FluidIterableIterator<T>;
+
+	/**
+	 * Returns a string representation of an array.
+	 */
+	toString(): string;
+
+	/**
+	 * Returns a string representation of an array. The elements are converted to string using their toLocaleString methods.
```
Josmithr: Nit
```suggestion
	 * @remarks
	 * Returns a string representation of an array. The elements are converted to string using their `toLocaleString` methods.
```

packages/common/core-interfaces/src/fluidArray.ts
```diff
@@ -0,0 +1,353 @@
...
+	 */
+	toLocaleString(): string;
+
+	/**
+	 * Combines two or more arrays.
+	 * This method returns a new array without modifying any existing arrays.
```
Josmithr, line 51: Nit
```suggestion
	 * @remarks
	 * This method returns a new array without modifying any existing arrays.
```

packages/common/core-interfaces/src/fluidArray.ts
```diff
@@ -0,0 +1,353 @@
...
+	 * @param items - Additional arrays and/or items to add to the end of the array.
+	 */
+	concat(...items: ConcatArray<T>[]): T[];
+
+	/**
+	 * Combines two or more arrays.
+	 * This method returns a new array without modifying any existing arrays.
```
Josmithr, line 59: Nit
```suggestion
	 * @remarks
	 * This method returns a new array without modifying any existing arrays.
```

packages/common/core-interfaces/src/fluidArray.ts
```diff
@@ -0,0 +1,353 @@
...
+	): U[];
+
+	/**
+	 * Returns a new array with all sub-array elements concatenated into it recursively up to the specified depth.
+	 * @param depth - The maximum recursion depth. Defaults to 1.
+	 */
+	flat<A, D extends number = 1>(this: A, depth?: D): FlatArray<A, D>[];
+
+	/**
+	 * {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/unscopables | Symbol.unscopables}
```
Josmithr: Nit
```suggestion
	 * See {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/unscopables | Symbol.unscopables}
```

packages/common/core-interfaces/src/fluidArray.ts
```diff
@@ -0,0 +1,353 @@
...
+	 */
+	filter(
+		predicate: (value: T, index: number, array: readonly T[]) => unknown,
+		// Typing inherited from ReadonlyArray.
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		thisArg?: any,
+	): T[];
+
+	/**
+	 * Calls the specified callback function for all the elements in an array. The return value of the callback function is the accumulated result, and is provided as an argument in the next call to the callback function.
```
Josmithr: This is a bit of pedantry on my part, especially since I'm assuming these docs were copied from TypeScript's. But I don't think this pair of sentences captures the intention of "reduce" very well. The second sentence feels like an afterthought / clarification - something that would go in a `@remarks` block. But really, it's the crux of what "reduce" does (and what differentiates it from `map`. Personally, I think the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) for this are better.

Paraphrased:
> Executes the provided callback function on each element of the array, in order, passing in the return value from the calculation on the preceding element. The final result of running the reducer across all elements of the array is a single value.

Josmithr: (Same note for the other reducer methods)

Josmithr (commented):
Left a few docs suggestions 🤓

Josmithr (approved):

CraigMacomber (approved):

PR #27748, opened 2026-07-23
Validate base version lineage before point-in-time replay

sonalideshpandemsft:
The ODSP point-in-time document service materializes a document at a target sequence number by replaying the live op stream on top of a recoverable base version. That replay is only valid if the chosen base is actually on the live document's lineage and the bridging ops are still retained.

** Lineage (epoch) validation when picking a base**

`OdspVersionManager.findBaseForSeq` checks the chosen base shares the live document's ODSP epoch (`x-fluid-epoch`) before returning it, so, a cross-lineage base can never be replayed.

- Added `getLiveDocumentEpoch` / `getRecoverableVersionEpoch` to the file version fetcher, reading `x-fluid-epoch` from the live and version-scoped snapshot endpoints (`blobs=0`).
- These reads deliberately use the raw `fetchArray` helper instead of `epochTracker.fetch`: the shared `EpochTracker` pins to the first epoch it sees and throws on divergence, so it could never yield two epochs to compare.
- On mismatch, throws the driver's non-retryable `fileOverwrittenInStorage` error (matching what `EpochTracker` raises), so the loader sees one consistent errorType. 

The service factory also builds a single shared `EpochTracker` and routes all reads (version history, base snapshot, and live ops) through it, giving an extra guard that every read is against the same file lineage.

Adds unit tests for the fetcher and version manager, and updates `DEV.md`.


packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -123,12 +148,61 @@ export class OdspVersionManager implements IOdspVersionManager {
...
 		}
 		return { kind: "noBaseVersion", oldestResolvedSeq };
 	}
 
+	private async validateLineageEpoch(base: ResolvedVersion): Promise<void> {
+		const [liveEpoch, baseEpoch] = await Promise.all([
+			this.fetcher.getLiveDocumentEpoch(),
+			this.fetcher.getRecoverableVersionEpoch(base.versionId),
+		]);
+		// Log the observed epochs so real traffic can confirm whether the version-scoped read actually
```
lindsnguyen: rather than rely on logs, could we not just manually test the API in the loop app with dev tools console, and do a restore (using the versions in the loop app, not pages app) to change the lineage and see if they api you're using is the old lineage or not?

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentService.ts
lindsnguyen, line 40: should we add unit tests for this service? I know youre looking into the local server tests but i think there's enough logic going on here that unit tests of the different edge cases are good too? Unless you're thinking it can be done in a followup PR? Not sure if there's a code coverage requirement for these changes.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -123,12 +148,61 @@ export class OdspVersionManager implements IOdspVersionManager {
...
+			baseVersionId: base.versionId,
+			baseEpoch,
+			liveEpoch,
+			epochsMatch:
+				baseEpoch !== undefined && liveEpoch !== undefined && baseEpoch === liveEpoch,
+		});
+		// Fail closed when either epoch is unknown: without both we cannot prove the base shares the
+		// live document's lineage, and replaying across lineages silently corrupts the result.
+		if (liveEpoch === undefined || baseEpoch === undefined) {
+			throw new UsageError(
```
lindsnguyen: in the scenario that either epoch is undefined, I'm not sure we should be throwing a `UsageError`? I would expect usage errors to mean that a user could do something different and fix the issue (I could be assuming the wrong meaning based on the error's name though so feel free to drop), but if we're hitting this then maybe somehow the api call's return is wrong, unrelated to what the user was asking for. Wondering also if this is something we'd also want to retry getting in case the header isn't returning properly for some transient reason? maybe not something we need here or in the near future but it would be good to note?

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
sonalideshpandemsft: I'll remove this log. I have tested restoring a previous version in loop app and the epoch id gets updated. This was left since I have been testing the changes in the real svc e2e pipeline

packages/drivers/odsp-driver/src/odspVersionManager/odspFileVersionFetcher.ts
```diff
@@ -134,5 +136,40 @@ export function createOdspFileVersionFetcher(
...
+			const response = await fetchArray(url, { method, headers });
+			return response.headers.get("x-fluid-epoch") ?? undefined;
+		});
+
+	const getLiveDocumentEpoch = async (): Promise<string | undefined> =>
+		// The (unversioned) live snapshot endpoint is a current-file read, so its epoch is the live
+		// document's epoch. `blobs=0` keeps the response to the tree metadata.
+		readEpoch(`${itemRoot}/opStream/snapshots/trees/latest?blobs=0`, "LiveEpoch");
+
+	const getRecoverableVersionEpoch = async (versionId: string): Promise<string | undefined> =>
```
anthony-murphy, line 203: we should cache this result, as a numbered versions epoch should never change.

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -75,6 +80,16 @@ export interface IOdspFileVersionFetcher {
 	 * wrong value.
 	 */
 	resolveSequenceNumber(versionId: string): Promise<number>;
+	/**
+	 * Read the live document's current ODSP epoch (`x-fluid-epoch`), or `undefined`. Epoch identifies
+	 * the file's binary lineage and changes on a version restore or download-then-reupload; compared
+	 * with {@link getRecoverableVersionEpoch} to confirm a base is on the live document's lineage.
+	 */
+	getLiveDocumentEpoch(): Promise<string | undefined>;
```
anthony-murphy: nit. i feel like this interface should live in odspFileVersionFetcher, as keeping it here leads to a circular dependency which can cause problem. should be a small followup

anthony-murphy: baiscally odspFileVersionFetcher shouldn't reference this file. looks like atleast one other type should move too

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentService.ts
```diff
@@ -66,15 +70,56 @@ export class OdspPointInTimeDocumentService
...
+						// Only enforce on a complete, non-cached, non-aborted pass: a cachedOnly pass
+						// legitimately ends short (it only drains local cache), and an abort is not a
+						// retention failure.
+						if (
+							result.done &&
+							opsNeeded &&
+							cachedOnly !== true &&
+							abortSignal?.aborted !== true &&
+							(firstSeq !== from || maxSeq < bounded - 1)
+						) {
```
anthony-murphy: isn't it possible the ops have yet to flush to storage and only exist via the deltastream/websocket? this seem like it could be a common case where the service update the document and the user is there ready to load the version right after the change.

sonalideshpandemsft: I think it's possible and the point in time serivce is storage-only so it does not read in-flight ops off the websocket. instead, this is already handled by [getSingleOpBatch  (with  ParallelRequests )](https://github.com/microsoft/FluidFramework/blob/97e7938697d4266aa20c7a2c25f2817596ef6ddb/packages/loader/driver-utils/src/parallelRequests.ts#L418) -- when delta storage returns fewer ops than expected, it's not treated as an error rather the driver's op fetch layer keeps polling storage for missing ops. If they flush within ~30sec,  the load succeds otherwise it fails with `cannotCatchup` error

packages/drivers/odsp-driver/src/odspVersionManager/odspVersionManager.ts
```diff
@@ -96,12 +84,14 @@ export interface IOdspVersionManager {
 export class OdspVersionManager implements IOdspVersionManager {
 	private versionsCache: Promise<OdspFileVersionRef[]> | undefined;
 	private readonly seqByVersion = new Map<string, Promise<number>>();
+	private readonly epochByVersion = new Map<string, Promise<string | undefined>>();
```
lindsnguyen, line 87: There's a PromiseCache in core-utils internal you can use. I didn't realize it originally, but planning to switch to that in my followup pr, I can probably make this change too if you don't wanna do it in this diff.

Planning to go back to that followup once I post my other PR

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentService.ts
```diff
@@ -66,15 +76,76 @@ export class OdspPointInTimeDocumentService
...
+									{ driverVersion },
+								);
+							}
+							// Top not reached: the low end is intact but the stream stopped before the
+							// target's op (`bounded - 1`). This is transient - those ops have been sequenced
+							// on the live document but are not yet flushed to (or visible from) delta storage,
+							// the common "load a version right after the change" race - so a retry can succeed
+							// once they land. A target legitimately beyond the live tip lands here too and will
+							// keep retrying until the loader's own bounds give up.
+							if (maxSeq < bounded - 1) {
```
lindsnguyen: could be wrong, this was a suggestion by AI but it makes sense to me: inside the call to `fetchMessages` (`odspDeltaStorageService.ts:161`), the `validateMessages` call (defined in `driverUtils.ts:125`) does:
```
// If not strict, then return the first consecutive sub-block. If strict or start
// seq number is not what we expected, then return no ops.
if (strict || from !== start) {
    messages.length = 0;
```
this empties the entire batch whenever the returned ops don't start exactly at `from`. Because of that (plus the poll-then-throw a layer down), I think both of the if branches here are unreachable in prod:
•  `firstSeq > from`  (retention → non-retryable): `validateMessages` empties any chunk that doesn't start at `from`, so the only non-empty chunk `read()` can only ever observe starts at `from`. This means `firstSeq` is always  === `from` , never  > `from`. The one input that would trip this (a non-empty batch starting above `from`) is exactly what the layer below deletes.
•  `maxSeq < bounded - 1`  (not-flushed → retryable): since we always pass a bounded `to`,  `getSingleOpBatch`(`parallelRequests.ts:418`) never returns a clean `done` on an incomplete range: on any shortfall it polls ~30s then throws ( :468 ), so  `read()` rejects instead of ever reaching `done` with  maxSeq < bounded - 1 .

This means the only `done` we'd get is when it's a complete one where both conditions are false.
I looked at the code personally and I'm pretty sure this is the case but please do look and verify? I think based on your response to Tony above though, the retrying already happens so maybe we just remove these branches?

lindsnguyen (approved):
couple comments but looks good otherwise.

packages/drivers/odsp-driver/src/pointInTimeDriver/odspPointInTimeDocumentService.ts
sonalideshpandemsft: checked this logic, and the cannotCatchUp case can’t actually happen because delta storage already promises that a completed stream contains the entire requested range of ops

lindsnguyen (approved):

PR #27754, opened 2026-07-24
build: bump client patch version to 2.112.2

dannimad:
## Description

Bumps the client release group version on `release/client/2.112` from `2.112.1` to `2.112.2`.


jason-ha (approved):
Changes look good for when the time comes.
Note: we won't need to follow-up with prior version bump since there are no differences in packages from 2.112.0 to 2.112.1.

shlevari (approved):

PR #27755, opened 2026-07-24
feat(publish): Removed public publishing from pipeline (#27707) [2.112]

MarioJGMsoft:
## Description

Cherry-picks [PR #27707](https://github.com/microsoft/FluidFramework/pull/27707) (`feat(publish): Removed public publishing from pipeline`) onto the `release/client/2.112` release branch.

Policy changes block registry.npmjs.org in various contexts, so the publishing process needed to be updated to unblock build pipelines. This backports that change to the 2.112 release.

Cherry-picked from upstream commit 92c8d24ed1c2933a18e5a8386293bea95fa19947.

## Note

Only the client pipeline can publish to registry.npmjs.org, and only when the corresponding option is set. No other pipeline run from this branch can publish to registry.npmjs.org. That is acceptable because this branch is intended for client releases only.

### DIFF
--- a/tools/pipelines/build-client.yml
+++ b/tools/pipelines/build-client.yml
@@ -35,6 +35,10 @@ parameters:
     - default
     - skip
     - force
+- name: allowPublishingToNpmjs
+  displayName: Allow publishing to npmjs.org
+  type: boolean
+  default: false
 - name: interdependencyRange
   displayName: Range to use for interdependencies (only affects releases) (default = ~)
   type: string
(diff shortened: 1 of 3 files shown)


jason-ha (approved):
Update the PR note to call out that only client pipeline can publish to npmjs.org if the option is set. No other pipeline run from branch can publish to npmjs.org and that is okay because this branch is intended for client releases only.

shlevari (approved):

PR #27756, opened 2026-07-24
feat(tree): emit telemetry when healing unresolvable identifiers on decode

CraigMacomber:
## Description

When `SharedTreeOptionsBeta.healUnresolvableIdentifiersOnDecode` is enabled, SharedTree now records a `HealUnresolvableIdentifierOnDecode` telemetry event (at `LogLevel.info`) each time a non-final identifier is healed while loading a summary. This lets applications relying on the healing workaround observe which documents actually required healing.

This is a follow-up to the identifier-healing work: the heal and throw paths only exist because of a prior bug where non-finalized identifiers could be written into summaries. Instrumenting the recovery path makes those occurrences visible in the wild. The error/throw path is intentionally left un-instrumented — the thrown exception already surfaces via the application's own error telemetry and must not be silently swallowed.

Changes:
- Add an optional `logger` to the internal `IdentifierHealingConfig`; `forceDecodeEncodedIdWithoutSession` sends the event directly via `logger?.send(..., LogLevel.essential)`.
- Thread the shared object's logger into the healing config in `SharedTreeKernel` and `SharedTreeCore`.
- Document the telemetry event and level on `healUnresolvableIdentifiersOnDecode`.
- Add unit tests and a changeset.

Because it changes observable behavior for consumers who opted into the healing feature (they now get telemetry), this ships with a changeset.

## Reviewer Guidance

- The logger lives on `IdentifierHealingConfig` (recovery path only) rather than being threaded independently; the throw path is deliberately not instrumented. Confirm that split is acceptable.

### DIFF
--- a/packages/dds/tree/src/test/util/compressedIds.spec.ts
+++ b/packages/dds/tree/src/test/util/compressedIds.spec.ts
@@ -267,7 +267,7 @@ describe("compressedIds", () => {
 				);
 			});
 
-			it("does not log when the id is resolvable", () => {
+			it("does not log when the ID is resolvable", () => {
 				const compressedId = testIdCompressor.generateCompressedId();
 				const opSpaceId = testIdCompressor.normalizeToOpSpace(compressedId);
 				const logger = new MockLogger();
--- a/packages/dds/tree/src/test/util/compressedIds.spec.ts
+++ b/packages/dds/tree/src/test/util/compressedIds.spec.ts
@@ -7,6 +7,7 @@ import { strict as assert } from "node:assert";
 
 import type { OpSpaceCompressedId, SessionId } from "@fluidframework/id-compressor";
 import { createIdCompressor, createSessionId } from "@fluidframework/id-compressor/internal";
+import { MockLogger } from "@fluidframework/telemetry-utils/internal";
 import { validateAssertionError } from "@fluidframework/test-runtime-utils/internal";
 
 import {
@@ -251,6 +252,31 @@ describe("compressedIds", () => {
 			// The id is final, so the result is a session-space id (numeric), not a v5 UUID string.
 			assert.equal(result, compressedId);
 			assert.equal(typeof result, "number");
+		});
+
+		describe("telemetry", () => {
+			it("records a recovery event on the heal path via the healing config logger", () => {
+				const { opSpaceId } = makeUnresolvableOpSpaceId();
+				const logger = new MockLogger();
+				forceDecodeEncodedIdWithoutSession(opSpaceId, testIdCompressor, {
+					sharedObjectId,
+					logger,
+				});
+				assert(
+					logger.events.some((e) => e.eventName === "HealUnresolvableIdentifierOnDecode"),
+				);
+			});
+
+			it("does not log when the id is resolvable", () => {
+				const compressedId = testIdCompressor.generateCompressedId();
+				const opSpaceId = testIdCompressor.normalizeToOpSpace(compressedId);
+				const logger = new MockLogger();
+				forceDecodeEncodedIdWithoutSession(opSpaceId, testIdCompressor, {
+					sharedObjectId,
+					logger,
+				});
+				assert.equal(logger.events.length, 0);
+			});
 		});
 	});
 
--- a/packages/dds/tree/src/util/compressedIds.ts
+++ b/packages/dds/tree/src/util/compressedIds.ts
@@ -212,7 +212,11 @@ export function forceDecodeEncodedIdWithoutSession(
 				category: "generic",
 				eventName: "HealUnresolvableIdentifierOnDecode",
 			},
-			LogLevel.info,
+			// This telemetry should be very low-volume (and possibly never happen at all),
+			// as should only occur when loading documents that were corrupted by a now fixed bug and have not been re-summarized since healing was added.
+			// It is useful for monitoring and diagnosing for this to be reported in the same cases in which the error case would be reported,
+			// which is the essential level.
+			LogLevel.essential,
 		);
 		return healed;
 	}
--- a/packages/dds/tree/src/util/compressedIds.ts
+++ b/packages/dds/tree/src/util/compressedIds.ts
@@ -191,9 +191,9 @@ export interface IdentifierHealingConfig {
  * not instrumented here: the thrown exception is expected to surface via the application's own error
  * telemetry, and should never be silently swallowed.
  *
- * @param id - The op-space compressed id to decode.
- * @param idCompressor - The ID compressor used to normalize the id.
- * @param healing - Heal-on-decode configuration. Presence enables healing of non-final ids.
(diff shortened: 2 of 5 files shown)


packages/dds/tree/src/util/compressedIds.ts
```diff
@@ -189,7 +206,15 @@ export function forceDecodeEncodedIdWithoutSession(
 	}
 	// `id` is a non-final op-space compressed id.
 	if (healing !== undefined) {
-		return uuidV5(`${healing.sharedObjectId}|${id}`, healingNamespace);
+		const healed = uuidV5(`${healing.sharedObjectId}|${id}`, healingNamespace);
+		healing.logger?.send(
```
CraigMacomber, line 210: I haven't made a change which added telemetry before, so please review the details here carefully.

I'm not confident any of these values are what we want, or if there is additional process that is necessary to follow.

dannimad (commented):

packages/dds/tree/src/util/compressedIds.ts
Josmithr: @markfields maybe you can advise here

packages/dds/tree/src/util/compressedIds.ts
```diff
@@ -177,6 +184,16 @@ export interface IdentifierHealingConfig {
...
+ *
+ * The recovery (heal) path only occurs because of a prior bug where non-finalized identifiers were
+ * written into summaries. When {@link IdentifierHealingConfig.logger} is supplied, a telemetry event
+ * is recorded on that path so heals can be observed in the wild. The error/throw path is intentionally
+ * not instrumented here: the thrown exception is expected to surface via the application's own error
+ * telemetry, and should never be silently swallowed.
+ *
+ * @param id - The op-space compressed id to decode.
+ * @param idCompressor - The ID compressor used to normalize the id.
+ * @param healing - Heal-on-decode configuration. Presence enables healing of non-final ids.
```
Josmithr: Nit
```suggestion
 * @param id - The op-space compressed ID to decode.
 * @param idCompressor - The ID compressor used to normalize the ID.
 * @param healing - Heal-on-decode configuration. Presence enables healing of non-final IDs.
```

packages/dds/tree/src/test/util/compressedIds.spec.ts
```diff
@@ -252,6 +253,31 @@ describe("compressedIds", () => {
...
+				forceDecodeEncodedIdWithoutSession(opSpaceId, testIdCompressor, {
+					sharedObjectId,
+					logger,
+				});
+				assert(
+					logger.events.some((e) => e.eventName === "HealUnresolvableIdentifierOnDecode"),
+				);
+			});
+
+			it("does not log when the id is resolvable", () => {
```
Josmithr: Nit
```suggestion
			it("does not log when the ID is resolvable", () => {
```

packages/dds/tree/src/util/compressedIds.ts
```diff
@@ -189,7 +206,15 @@ export function forceDecodeEncodedIdWithoutSession(
...
 	// `id` is a non-final op-space compressed id.
 	if (healing !== undefined) {
-		return uuidV5(`${healing.sharedObjectId}|${id}`, healingNamespace);
+		const healed = uuidV5(`${healing.sharedObjectId}|${id}`, healingNamespace);
+		healing.logger?.send(
+			{
+				category: "generic",
+				eventName: "HealUnresolvableIdentifierOnDecode",
+			},
+			LogLevel.info,
```
markfields: I do think this should be essential.  Rationale -the other side of the `if (healing !== undefined)` check throws an error which we will for sure log at essential level (probably with multiple events as it propagates).

So we should match both branches to be able to definitively monitor the fix.  I'm not worried about the overall volume since there's natural attrition due to not using the document or subsequent summaries persisting the fixed ID.

packages/dds/tree/src/util/compressedIds.ts
```diff
@@ -189,7 +206,15 @@ export function forceDecodeEncodedIdWithoutSession(
 	}
 	// `id` is a non-final op-space compressed id.
 	if (healing !== undefined) {
-		return uuidV5(`${healing.sharedObjectId}|${id}`, healingNamespace);
+		const healed = uuidV5(`${healing.sharedObjectId}|${id}`, healingNamespace);
+		healing.logger?.send(
+			{
+				category: "generic",
```
markfields, line 212: If you switch the type of `logger` to be `ITelemetryLoggerExt` and use `sendEvent` or `sendTelemetryEvent` or whatever then you won't need this category ("generic" is default).  That's the interface we typically use for intrumenting.  (The base logger is what is exposed across seams for passing logs around)

packages/dds/tree/src/test/util/compressedIds.spec.ts
```diff
@@ -252,6 +253,31 @@ describe("compressedIds", () => {
...
+				forceDecodeEncodedIdWithoutSession(opSpaceId, testIdCompressor, {
+					sharedObjectId,
+					logger,
+				});
+				assert(
+					logger.events.some((e) => e.eventName === "HealUnresolvableIdentifierOnDecode"),
+				);
+			});
+
+			it("does not log when the ID is resolvable", () => {
```
markfields, line 270: Want to add a test case for not logging when it throws (not resolvable and not healed)

dannimad:

packages/dds/tree/src/test/util/compressedIds.spec.ts
CraigMacomber: That case is only possible when forceDecodeEncodedIdWithoutSession is passed no IdentifierHealingConfig, and since the logger is part of IdentifierHealingConfig, that would involve testing that an API which is not passed the longer does not log anything.

I don't think thats helpful for two reasons:
1. If would actually be ok if we logged telemetry in the error case.
2. Its impossible that it would log anything as it has no access to the logger.

We could add a higher level integration case for that case, but that would amount to testing that loading a shared tree with an invalid summary throws a data corruption error (which would be logged), and I hope we have testing for that somewhere else. We do have testing in this file that an unrecovered ID does throw.

packages/dds/tree/src/util/compressedIds.ts
CraigMacomber: Done.

packages/dds/tree/src/util/compressedIds.ts
CraigMacomber: There are reasons not to do this, mostly related to potential future refactors and shared-tree internals which can accept loggers directly from users which might not implement ITelemetryLoggerExt.
Specifically if this code becomes reachable from IndependentViews which don't have a main fluid logger, we could run into issues.

markfields (approved):

Josmithr (approved):

PR #27759, opened 2026-07-24
docs(tree): RFC for shared branch config

yann-achard-MS:
## Description

This PR serves as a request for comments on the potential vision for how to model shared branch capabilities.
It may never be checked in and that's fine.

## Breaking Changes

None

## Reviewer Guidance

I am first and foremost interested in opinions on whether the implied system seems desirable and approachable from an application author's point of view. I also welcome opinions on the proposed factoring of the system and how it is explained. Such discussions may lead to questions about or new proposals for how this system may be implemented wherever that may have an impact on the functionalities or performance characteristics described here.

I did not bother to factor out from `SharedBranchConfig` those settings that may be applicable to local branches. I am not seeking comments on such a factoring.

### DIFF
--- a/packages/dds/tree/src/shared-tree-core/sharedBranchConfig.ts
+++ b/packages/dds/tree/src/shared-tree-core/sharedBranchConfig.ts
@@ -1,0 +1,196 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Configuration for a branch.
+ * This has important performance and functional implications.
+ *
+ * See {@link SAMPLE_BRANCH_CONFIGS} for some example configurations you may wish to use.
+ */
+export interface SharedBranchConfig {
+	/**
+	 * Controls when to retain the commit history that relates the tip of the parent branch to the tip of this child branch.
+	 * Retaining this history allows the child branch to be rebased onto or merged into the parent branch.
+	 * This history grows as new commits are added to either the parent branch or child branch.
+	 * This growth negatively impacts the following areas:
+	 * - The main memory on running clients that have checked out the child branch
+	 * - The size of summaries (in memory and over the wire) and the time required to generate, upload, download, and process them.
+	 * - The size of the document at rest (in storage and on the wire).
+	 *
+	 * @remarks This configuration does not affect the retention of commits necessary for the purpose of merging concurrent edits made to the child branch.
+	 */
+	readonly retainRelationToParent: HistoryRetentionConfig;
+
+	/**
+	 * The maximum number of divergent commits on the parent branch before the child branch is automatically rebased to the tip of the parent branch.
+	 * Note that once the rebase occurs, the number of divergent commits on the parent branch comes back down to zero.
+	 *
+	 * Setting this field to `Number.POSITIVE_INFINITY` disables automatic rebasing entirely.
+	 *
+	 * Setting this field to `0` will cause the child branch to be automatically rebased onto the parent branch after every new commit on the parent branch.
+	 *
+	 * Considering the following example graph:
+	 * ```text
+	 * A──B──C──D──E     <- The tip of the parent branch points to commit `E`
+	 *       └──X──Y──Z  <- The tip of the child branch points to commit `Z`
+	 * ```
+	 * Setting this field to `2` will cause the child branch to be automatically rebased onto the parent branch after the next commit is added to the parent branch.
+	 *
+	 * @remarks Note that auto-rebasing will only occur if the history linking the tip of both branches happens to be retained at that point in time.
+	 * Most notably, auto-rebasing can never occur if {@link HistoryRetentionConfig.untilDivergentCommitCountOnParentBranchExceeds} is configured with a smaller value than the one in this field.
+	 * See {@link HistoryRetentionConfig} for more information on how to configure history retention.
+	 */
+	readonly autoRebaseWhenDivergentCommitCountOnParentBranchExceeds: number;
+
+	/**
+	 * The maximum number of commits between the tip of the main branch and the tip of this branch before the tip state of this branch is included in summaries.
+	 *
+	 * Considering the following example graph:
+	 * ```text
+	 * A──B──C──D──E           <- The tip of the main branch points to commit `E`
+	 *       └──R──S──T        <- The tip of a child branch points to commit `T`
+	 *             └──X──Y──Z  <- The tip of a grandchild branch points to commit `Z`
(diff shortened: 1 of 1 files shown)


packages/dds/tree/src/shared-tree-core/sharedBranchConfig.ts
```diff
@@ -0,0 +1,196 @@
...
+	 *
+	 * Considering the following example graph:
+	 * ```text
+	 * A──B──C──D──E     <- The tip of the parent branch points to commit `E`
+	 *       └──X──Y──Z  <- The tip of the child branch points to commit `Z`
+	 * ```
+	 * This field would have to be set to at least `3` in order to retain the history linking both branches, since there are three divergent commits on the child branch (`X`, `Y`, and `Z`).
+	 * Setting it to `2` would cause the whole history to become GC-able as soon as `Z` had been added to the child branch.
+	 */
+	readonly untilDivergentCommitCountOnChildBranchExceeds: number;
```
noencke, line 123: This is clearly documented; I think I understand what these numbers do after just one read. However, the documentation isn't _prescriptive_ and the API isn't _suggestive_. I don't know what values to choose for a given scenario.

Suppose I have a branch that I expect to get "longish". Let's say it can get up to 500 divergent commits - or maybe it can get arbitrarily long, but 500 commits is where I go "okay that's pretty extreme, that will never happen in practice". Do I set the max child divergent commits to 500? Or should I set it to 250 - something closer to the P90 or something? Or should I be thinking in terms of performance rather than usage expectation - my _service_ gets slow after 350 commits, so I should only go up to 350 commits, regardless of how I expect my users/app to behave?

Then, suppose a user reaches the limit. Should I garbage their branch and throw an error? Should I "freeze" their branch - keep it alive, but block any more edits with a user-facing message? Is that even possible robustly with concurrent edits? Should I warn them when they get to 90% of the limit?

Now, I know the instinct here is to say - "that's all fine. The app can do whichever of those things it decides. It depends on the app and the service and what user experience they want. This API just needs to be flexible enough to allow them any policy." But it's still sort of unclear what to do, I think. Something is strange about the state transition from "branch is under the limit" to "branch exceeds the limit". How is the user supposed to respond to that transition?

I'm still working on modeling this in my head.

noencke: Maybe this number is usually supposed to be set to infinity and we rely on apps to dispose branches - and setting this number to an actual integer is like an "emergency stop" rather than a tuned policy choice. If that's the case, I think it makes sense as-is...

packages/dds/tree/src/shared-tree-core/sharedBranchConfig.ts
```diff
@@ -0,0 +1,196 @@
...
+	 * Retaining this history allows the child branch to be rebased onto or merged into the parent branch.
+	 * This history grows as new commits are added to either the parent branch or child branch.
+	 * This growth negatively impacts the following areas:
+	 * - The main memory on running clients that have checked out the child branch
+	 * - The size of summaries (in memory and over the wire) and the time required to generate, upload, download, and process them.
+	 * - The size of the document at rest (in storage and on the wire).
+	 *
+	 * @remarks This configuration does not affect the retention of commits necessary for the purpose of merging concurrent edits made to the child branch.
+	 */
+	readonly retainRelationToParent: HistoryRetentionConfig;
```
noencke, line 24: Have you thought about whether we might want to retain shared branch history for the sake of undo/history, regardless of whether or not we also care about merging/rebasing with the parent branch? The current dials are all about keeping the history around for the sake of merge/rebase with the parent. But, might I just want a long shared branch where I don't necessarily intend to merge it ever (or maybe I do in some cases, but not others, who knows) but I do want to be able to undo at least some amount backwards. In that case, I think we'd use these same dials (well, specifically the "child" divergent commit count dial) to model that. If that's true, then that should probably make its way into the docs in the final version as a second effect of turning that dial.

noencke: Or perhaps it's a separate dial if we want to be able to undo _past_ (before) the fork point.

packages/dds/tree/src/shared-tree-core/sharedBranchConfig.ts
yann-achard-MS: While reading this comment, I initially thought you may not have had the chance to read the docs on the sample configs yet, but I think I see how even these don't address what you're getting at.

Part of the problem is that we haven't made performance investigations to be able to say how different scenarios perform under different configurations. Linking to that would answer a lot of questions, then the app author can work out what perf they should expect for a given percentile. It would be even cooler if we could make a test harness so that an app author can get perf numbers for their own app.

> Something is strange about the state transition from "branch is under the limit" to "branch exceeds the limit". How is the user supposed to respond to that transition?

Just in case that's not clear, it's OK for a branch to go beyond the numbers set in the config. The consequences are not that dire: auto-rebasing (if it was enabled at all) will stop happening, and you won't be able to manually rebase or merge the branch anymore (the merge/rebase APIs would need be changed to tryMerge/tryRebase, or maybe we would just add isRebeasableOnto/isMergableInto predicates and make merge/rebase throw when the history is missing). 

Back to the question... it's up to the app to decides whether the end user should even be made aware of such a limit. With a branch that is never rebased but is meant to be mergeable upon user request, the app could either warn the user about the limit, propose that the branch be rebased there and then, or say nothing to the user and use content-based merge if and when the user does request the merge.

I think I agree with you about the strangeness but only insofar as configurations that lead to a switch are themselves strange because it means the purpose of the branch is not clear at the outset. If the app author is not confident about the purpose of the shared branches created through a given code path, then it's the app author's responsibility to choose a configuration that will lead to the least regret, potentially by picking something conservative.

This all leads me to contemplate whether to allow the configuration to be amended after the fact and what that would mean (e.g., increasing the limits on history retention on an existing branch only makes sense before that branch punches through the limit). If the configuration is changeable on the fly, then I suspect we could get away with a dramatically simpler configuration.

I will keep thinking about this.

packages/dds/tree/src/shared-tree-core/sharedBranchConfig.ts
yann-achard-MS: Ah, good point. We currently retain historical data for undo purposes if you get a revertible, so local branches wouldn't be affected by this unless we change how that works. For shared/persisted branches though, I agree you'd have to start by asking for the history to be preserved. It likely you'd want to make that independent of the fork point though (in the sense that you might want to retain the ability to undo things that were done on the parent branch).

PR #27766, opened 2026-07-27
fix(ci): stop fetching assets from a public CDN at runtime

MarioJGMsoft:
## Description

Newly enforced policies no longer permit outbound calls to `cdn.jsdelivr.net`. Three places in the repo fetch assets from that CDN at runtime; the browser-based ones originate inside the Puppeteer/Chrome process, which makes them hard to attribute from a violation report alone. This PR makes all three load locally instead.

**1. presence-tracker emoji picker (the actual source of the reports).**
`examples/apps/presence-tracker/src/view.ts` constructs `new Picker()` from `emoji-picker-element`, whose default data source is `https://cdn.jsdelivr.net/npm/emoji-picker-element-data@^1/en/emojibase/data.json` (`emoji-picker-element/database.js`). `tests/presenceTracker.test.ts` launches Chrome with `headless: "shell"`, which matches the reported process path `~/.cache/puppeteer/chrome-headless-shell/linux-131.0.6778.87/...` — the Chrome version `puppeteer@23.10.3` installs via the root `test:jest` script. The `/proc/self/exe` entries are that browser's renderer/zygote children. The fix adds `emoji-picker-element-data` as a dependency, emits it as a webpack `asset/resource`, and passes the resulting local URL via `dataSource`.

**2. Build-perf dashboard.** `build-tools/.../templates/dashboard.ejs` loaded `chart.js` and `chartjs-adapter-date-fns` from jsdelivr via `<script src>`. They are now added as dependencies of `build-cli` and inlined into the generated HTML by `htmlGenerator.ts`. Note this template was *not* the source of the reports — nothing in `build-performance-observability.yml` opens the generated HTML in a browser — but it is a real CDN reference in shipped output.

**3. smde example.** SimpleMDE's spell checker lazily downloads a dictionary from a CDN; `smdeView.tsx` now passes `spellChecker: false`.

## Reviewer Guidance

-   `src/emojiDataUrl.d.ts` is an ambient declaration for the webpack-aliased module. TypeScript can't resolve the alias, and importing the JSON directly produces a ~430 KB literal type, hence the `import-x/no-unresolved` suppression.
-   Locally verified: `tsc`, `webpack` (emits the data JSON as a hashed asset), `biome check`, `eslint`, `flub check policy`, and the 11 `htmlGenerator` tests. The Puppeteer run itself could not be exercised locally (Chrome download blocked) — the `JestTest` job in the client build is the decisive validation.

Fixes [AB#73159](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/73159)


.changeset/clean-build-dashboards.md
```diff
@@ -0,0 +1,8 @@
+---
+"@fluid-tools/build-cli": patch
```
CraigMacomber: This changeset directory is for the client release group. `@fluid-tools/build-cli` is not part of the client release group, so this does not belong here.

CraigMacomber: Generally the change type should always be minor for our tooling to work correctly, not patch. Setting patch here will cause this to not get included in the release notes when we do a minor release (I'm unclear if thats working at intended or a bug)

MarioJGMsoft: I realized that the changeset might be unnecessary for this PR. Opted for removing it completely from the PR.

CraigMacomber (approved):

PR #27767, opened 2026-07-27
Add end-to-end test for a data store realized between summarize and refresh

agarwal-navin:
## Description

Adds an end-to-end test to `summarizeWithOutOfOrderDataStoreRealization.spec.ts` that reproduces a summarizer behavior where a data store is realized on the summarizer in the window *between* a summary being generated / uploaded and that summary being acked (refreshed).

The scenario is modeled on a real application:

- A "table" data object (the default data store) eagerly loads any "cell" data store whose handle is added to it by another client, via an op handler installed in `hasInitialized` and kept alive on the summarizer through `mixinSummaryHandler`.
- A new cell is attached so that its attach op reaches the summarizer *after* the summary's reference sequence number is fixed but *before* the ack. The summarizer therefore realizes the newly attached cell — creating summarizer nodes for its DDSes — before the summary is refreshed.

With the current summarizer-node behavior, the cell's DDS then tracks a reference sequence number for a summary in which its content was not directly written, and the *next* summary emits an unresolvable summary handle, so the upload fails (in production this surfaces as a 404 `fluidElementNotFound`).

This is intentionally **not** being fixed: when it happens the summarizer restarts and the issue self-heals on the fresh summarizer. The test asserts the current (rejecting) behavior with `assert.rejects`, documents this in a comment, and uses `itExpects` to allow the expected `Summarize_cancel` / `SummarizeFailed` error telemetry.

Test-only change (no product code touched), so no changeset is included.

## Reviewer Guidance


packages/test/test-end-to-end-tests/src/test/summarization/summarizeWithOutOfOrderDataStoreRealization.spec.ts
```diff
@@ -289,3 +296,207 @@ describeCompat(
...
+					// Restore the original upload before the final summary, even if the summary above throws.
+					summarizerRuntime.storage.uploadSummaryWithContext = originalUpload;
+				}
+
+				// Ideally this summary should NOT reject. But with the current summarizer node behavior it
+				// does: the cell's DDS emits an unresolvable summary handle and the upload fails. We have
+				// decided not to fix this - when it happens the summarizer restarts and the issue fixes
+				// itself on the fresh summarizer - so this test asserts the current (rejecting) behavior.
+				await provider.ensureSynchronized();
+				await assert.rejects(async () => summarizeNow(summarizer));
```
markfields: Would be good to assert on the specific error that we're expecting

agarwal-navin: Done

markfields (approved):

PR #27770, opened 2026-07-27
Misc Formatted Text work

CraigMacomber:
## Description

Misc work before we want to take text to alpha.

1. Disable expensive asserts outside of text tests.
2. Use javascript private properties where APIs could collide with user subclass properties.
3. Note risk of name coillision for properties of erased types
4. add reformat API
5. Clarify formatRange API with docs and tests.
6. Fill in some lacking docs

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/text/textDomain.ts
+++ b/packages/dds/tree/src/text/textDomain.ts
@@ -34,6 +34,20 @@ import type {
 // eslint-disable-next-line @typescript-eslint/no-unused-vars, unused-imports/no-unused-imports, import-x/no-duplicates
 import type { NodeKind, TreeNodeSchema } from "../simple-tree/index.js";
 
+/**
+ * Enables expensive debug assertions for formatted text.
+ */
+export let enableTextExpensiveDebugAsserts = false;
+
+/**
+ * Enables or disables expensive debug assertions for formatted text.
+ * @remarks
+ * These should remain disabled except when testing implementation details of this code.
+ */
+export function setEnableExpensiveDebugAsserts(value: boolean): void {
+	enableTextExpensiveDebugAsserts = value;
+}
+
 const sf = new SchemaFactoryAlpha("com.fluidframework.text");
 
 class TextNode
@@ -122,7 +136,8 @@ class TextNode
 		const result = this.content.charactersCopy();
 		debugAssert(
 			() =>
-				compareArrays(result, this.charactersCopy_reference()) ||
+				(enableTextExpensiveDebugAsserts &&
+					compareArrays(result, this.#charactersCopy_reference())) ||
 				"invalid charactersCopy optimizations",
 		);
 		return result;
@@ -131,7 +146,9 @@ class TextNode
 	public fullString(): string {
 		const result = this.content.fullString();
 		debugAssert(
-			() => result === this.fullString_reference() || "invalid fullString optimizations",
+			() =>
+				(enableTextExpensiveDebugAsserts && result === this.#fullString_reference()) ||
+				"invalid fullString optimizations",
 		);
 		return result;
 	}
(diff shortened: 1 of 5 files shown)


packages/dds/tree/src/text/textDomain.ts
```diff
@@ -122,7 +136,8 @@ class TextNode
 		const result = this.content.charactersCopy();
 		debugAssert(
 			() =>
-				compareArrays(result, this.charactersCopy_reference()) ||
+				(enableTextExpensiveDebugAsserts &&
```
brrichards: since `enableTextExpensiveDebugAsserts` defaults to false would this check always fail by default with the && gate?

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -548,6 +601,16 @@ export namespace FormattedTextAsTree {
 	 *
 	 * @see {@link FormattedTextAsTree.Statics.fromString} for construction.
 	 * @see {@link FormattedTextAsTree.createSchema} for creating schemas whose nodes implement this.
+	 *
+	 * Rather than referring to this type, use the the instance type of the subclass of the schema from {@link FormattedTextAsTree.createSchema}.
```
brrichards:
```suggestion
	 * Rather than referring to this type, use the instance type of the subclass of the schema from {@link FormattedTextAsTree.createSchema}.
```

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -548,6 +601,16 @@ export namespace FormattedTextAsTree {
...
+	 * When a generic interface is required, use {@link FormattedTextAsTree.FormattedTextMembers}.
+	 * The generic type parameterization of this interface is is not stable, and simply an implementation detail of {@link FormattedTextAsTree.FormattedTextMembers}.
```
brrichards:
```suggestion
	 * The generic type parameterization of this interface is not stable, and is simply an implementation detail of {@link FormattedTextAsTree.FormattedTextMembers}.
```

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -558,13 +621,17 @@ export namespace FormattedTextAsTree {
 		 * This is not persisted in the tree, and observation of it is not tracked by the tree observation tracking.
 		 * @privateRemarks
 		 * Opt this into observation tracking.
+		 * TODO: This being an mutable instance property is not required, and might not be the best API design choice.
```
brrichards:
```suggestion
		 * TODO: This being a mutable instance property is not required, and might not be the best API design choice.
```

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -190,21 +211,79 @@ export namespace FormattedTextAsTree {
...
+						);
+					}
+					for (const [key, value] of fieldFormats) {
+						const field = atomFormatSchema.fields.get(key);
+						if (field === undefined) {
+							throw new UsageError(`Unknown format key: ${key}`);
+						}
+
+						// Ensures that if the input is a node, it is cloned before being inserted into the tree.
+						// Note that since this used field schema, `undefined` can pass through this if allowed by the schema.
```
brrichards:
```suggestion
						// Note that since this uses field schema, `undefined` can pass through this if allowed by the schema.
```

packages/dds/tree/src/text/textDomain.ts
CraigMacomber: Yes, this is wrong.

Fortunately some tests also caught it, but not dedicated tests for this. I have better encapsulated this logic, and added dedicated testing for it. Should be better now.

brrichards (approved):

PR #27772, opened 2026-07-28
Emit manifest metadata on all build pipelines

Abe27342:
## Description

Moves the manifest generation introduced in #27752 to the build-npm-package template rather than only include it in the client build.

This will allow standing up separate publishing pipelines for all of our other build pipelines.

Test links (only visible internally) for a reasonable representative subset of these pipelines:

https://dev.azure.com/fluidframework/internal/_build/results?buildId=412689&view=results
https://dev.azure.com/fluidframework/internal/_build/results?buildId=412693&view=results
https://dev.azure.com/fluidframework/internal/_build/results?buildId=412691&view=results


MarioJGMsoft (approved):
The move looks good to me :)

PR #27774, opened 2026-07-28
treeshape equals short circuit

daesunp:
## Description

Add a reference-identity short circuit for `TreeShape.equals` so the common case of comparing a shape against itself skips the full structural walk

### DIFF
--- a/packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
+++ b/packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
@@ -197,6 +197,11 @@ export class TreeShape {
 			return true;
 		}
 
+		// A given type usually only has one shape, so check it next to speed up the false case.
+		if (this.type !== other.type) {
+			return false;
+		}
+
 		// TODO: either dedupe instances and/or store a collision resistant hash for fast compare.
 
 		if (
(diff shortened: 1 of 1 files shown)


packages/dds/tree/src/feature-libraries/chunked-forest/uniformChunk.ts
```diff
@@ -191,6 +191,12 @@ export class TreeShape {
 	}
 
 	public equals(other: TreeShape): boolean {
+		// Shapes are commonly shared (e.g. all chunks from one chunker reference the same instance), so
+		// short-circuit on reference identity to avoid the full structural walk in the common case.
+		if (this === other) {
+			return true;
+		}
```
CraigMacomber, line 198: This looks good, but the false case is still going to be slow in a lot of cases, as its going to recursively walk the whole tree in compareArrays (unless the array lengths differ). I feel like we might as well fix that while we are here since its related.

Pulling the `this.type` check before the array check would make this O(1) in most cases (as in practice the shapes are either the same object, or have different types in almost all cases currently)
```suggestion
		}
		// A given type usually only has one shape, so check it next to speed up the false case.
		if (this.type !== other.type) {
			return false;
		}
```

Of course you also want to remove that check from the end if adding it here.

daesunp: Updated with suggestion

CraigMacomber (approved):

PR #27781, opened 2026-07-28
Cleanup FormattedTextAsTree API surface a bit

CraigMacomber:
## Description

Cleanup FormattedTextAsTreeDefault API surface a bit, preparing for promotion to alpha.

This mostly just combines FormattedTextMembers and Members, and adds the missing optional format parameter to Statics.fromString. Additionally includes some better API tagging, comments minor tweaks.

## Reviewer Guidance


packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -651,7 +658,12 @@ export namespace FormattedTextAsTree {
 		 */
 		insertWithFormattingAt(
```
Josmithr, line 661: While we're simplifying, is there a reason this shouldn't be called just "insertAt" (presumably this conflicts with ArrayNode)? Or "insertCharactersAt"?

Josmithr (approved):

packages/dds/tree/src/text/textDomainFormatted.ts
CraigMacomber: We extend TextAsTree.Members which has insertAt already, which used the defaultFormatting (as APIs without explicitly formatted are documented to do).

We could override that with one that takes optional formatting, but if we did that, it would take an optional formatting parameter for all the text, not formatting per character like this does, so this would still be a different API which we need.

Josmithr: Okay. Just wondering if we can simplify the name a bit.

Josmithr (approved):

PR #27785, opened 2026-07-29
improvement(client-tree): pipe ChangeFamily context to transaction post-processors

jason-ha:
Transaction post-processors previously received only the squashed change to
process. Some post-processing (e.g. minimization) needs access to
change-family–specific helpers such as the `ModularChangeFamily`, which were
not reachable from the processor callback.

This change threads a new `TChangeProcessingContext` type parameter through the
change-processing pipeline so a `ChangeFamily` can supply a context to its
related processors:

- `ChangeFamily` gains a `buildProcessor` method and a `ProcessChangeFn<TChange,
  TChangeProcessingContext>` helper type. When `TChangeProcessingContext` is
  `never` (default), the processor keeps its original `(change) => change` shape;
  otherwise, it receives `(change, context)`.
- `ChangeProcessor`, `SquashingTransactionOptions`, `SquashingTransactionStack`,
  `SharedTreeBranch`, `EditManager`, and the transaction plumbing are
  parameterized on `TChangeProcessingContext` and pass the context through.
- `SharedTreeChangeFamily` implements `buildProcessor` and exposes a
  context with `ModularChangeFamily` now available to post-processors.
- `DefaultChangeFamily`, `ModularChangeFamily`, and `mitigatedChangeFamily` are
  updated to the new signatures.
- `EditManager`'s `TChangeFamily` parameter that allowed concrete `ChangeFamily` to be read from public `changeFamily` has been removed as no callers required the concrete type. `TChangeProcessingContext` with a default of `never` appears in its place.

The context is now piped end-to-end so transaction post-processors (such as the
minimizer) can access the change family that produced the change.

### DIFF
--- a/packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
+++ b/packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
@@ -20,7 +20,7 @@ import type { ModularChangeset } from "./modularChangeTypes.js";
  * future change.
  *
  * @param change - The change to minimize.
- * @param fieldKinds - The field kinds to delegate to when computing the change's delta.
+ * @param changeFamily - The change family used to compute the change's delta and identify built nodes.
  */
 export function minimizeModularChangeset(
 	change: ModularChangeset,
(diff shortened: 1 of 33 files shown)


packages/dds/tree/src/feature-libraries/modular-schema/minimizeModularChange.ts
```diff
@@ -26,7 +24,7 @@ import type { ModularChangeset } from "./modularChangeTypes.js";
  */
```
yann-achard-MS, line 24: Update doc comment

yann-achard-MS (approved):

yann-achard-MS (approved):

PR #27787, opened 2026-07-29
feat(tree): getOrInsert and getOrInsertComputed mapNode alpha API

brrichards:
## Description
Adds `getOrInsert` and `getOrInsertedComputed` alpha APIs for `mapNode`

## Reviewer Guidance


packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -261,6 +277,18 @@ abstract class CustomMapNodeBase<const T extends ImplicitAllowedTypes> extends T
 		this.editor(key).set(mapTree, field.length === 0);
 		return this;
 	}
+	public getOrInsert(
+		key: string,
+		defaultValue: InsertableTreeNodeFromImplicitAllowedTypes<T>,
+	): TreeNodeFromImplicitAllowedTypes<T> {
+		if (!this.has(key)) {
+			if (defaultValue === undefined) {
```
yann-achard-MS: We should catch this `defaultValue === undefined` case at compile-time instead of checking it at runtime.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -180,6 +180,22 @@ export interface TreeMapNodeAlpha<T extends ImplicitAllowedTypes = ImplicitAllow
...
+	 * Returns the value at `key`, first inserting `defaultValue` if this map has no entry for `key`.
+	 *
+	 * @param key - The key of the element to return or insert at.
+	 * @param defaultValue - The value to insert if `key` has no entry. May not be `undefined`.
+	 * @returns The existing value at `key`, or the newly inserted `defaultValue`.
+	 *
+	 * @remarks
+	 * The merge semantics of this operation are loosely specified:
+	 * the check for insertion may happen either at edit authoring time only,
+	 * or at both edit authoring time and sequencing time, yielding different outcomes.
```
yann-achard-MS: After chatting more about this at standup, I realized that we'll never implement the "at sequencing time" version of this API. This means we can fully specifiy the exact semantics right now. Here's what we want to say about them:

The check for the presence of an existing entry with the given `key` is performed at the time the edit is authored. This has implications for the merge semantics of this operation:
* If no entry is present for `key` at authoring time, thus leading to an insert, while a peer concurrently inserts a value for the same `key`, then the two inserts will race and the one that is sequenced second will overwrite the value inserted by the one that is sequenced first.
* If an entry is present for `key` at authoring time, thus leading to no insert, while a peer concurrently deletes the entry for `key`, then the map will end up with no entry for `key` no matter how the two edits are sequenced.
* If an entry is present for `key` at authoring time, thus leading to no insert, while a peer concurrently inserts a new entry for `key`, then the map will end up with the entry set by the peer no matter how the two edits are sequenced.

These last two points mean that, upon sequencing of an edit made with this API, there is no guarantee that the entry for the given key will be the current one (if any) or the fallback one. If such a guarantee is important, then 
consider using constraints to ensure your edit only applies when appropriate.

@remarks
This API is not equivalent to the following alternative:
```
map.set(key, map.get(key) ?? fallbackValue);
```
They differ in the following ways:
* This API treats entries containing `null` values as populated. By contrast, the above alternative's usage of `??` means that a `null` entry is treated as equivalent a missing entry.
* This API only inserts/sets the entry for the given `key` when there is no current one. By contrast, the above alternative always performs an insert. This will throw an error for non-leaf types like objects, maps, array, and records in cases where the entry is already populated because we do not support inserting a node that was already inserted. This will always succeed for leaf types like number, string, bool, null, and handle since a new node will be created based on the value. Note that even in the cases where it does not throw, the `set` operation has the potential to overwrite entries that are concurrently inserted even when the map had an entry for `key`, which is not the case with this API.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -180,6 +180,22 @@ export interface TreeMapNodeAlpha<T extends ImplicitAllowedTypes = ImplicitAllow
...
+	 */
+	getOrInsert(
+		key: string,
+		defaultValue: InsertableTreeNodeFromImplicitAllowedTypes<T>,
```
yann-achard-MS: We should rename `defaultValue` to something else (`fallbackValue`?) since we don't think this API should be used for writing default values (see comment above).

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -180,6 +180,22 @@ export interface TreeMapNodeAlpha<T extends ImplicitAllowedTypes = ImplicitAllow
...
+	 * @param key - The key of the element to return or insert at.
+	 * @param defaultValue - The value to insert if `key` has no entry. May not be `undefined`.
+	 * @returns The existing value at `key`, or the newly inserted `defaultValue`.
+	 *
+	 * @remarks
+	 * The merge semantics of this operation are loosely specified:
+	 * the check for insertion may happen either at edit authoring time only,
+	 * or at both edit authoring time and sequencing time, yielding different outcomes.
+	 */
+	getOrInsert(
```
yann-achard-MS, line 242: Separately from merge semantics, we also want to caution people against writing default values to their document if the application logic could instead have logic that automatically returns the default value when reading (which is much less costly). The more legitimate use of this API is when the value to be optionally inserted can vary with the key.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -261,6 +277,18 @@ abstract class CustomMapNodeBase<const T extends ImplicitAllowedTypes> extends T
 		this.editor(key).set(mapTree, field.length === 0);
 		return this;
 	}
+	public getOrInsert(
+		key: string,
+		defaultValue: InsertableTreeNodeFromImplicitAllowedTypes<T>,
+	): TreeNodeFromImplicitAllowedTypes<T> {
+		if (!this.has(key)) {
```
yann-achard-MS: We want to minimize the number of accesses to the map, so I think that means we want to call `get` instead of `has` at the top here, and avoid calling `get` at the bottom.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
brrichards: Thanks for the thorough documentation of the merge semantics! updated the jsdoc to reflect this.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
brrichards: updated to reflect this along with the name change from "defaultValue"

.changeset/neat-rings-ask.md
```diff
@@ -3,16 +3,18 @@
...
 
-The merge semantics of `clear` are loosely specified: either of the following may occur:
+The merge semantics of `clear` are loosely specified. Either of the following may occur:
 
 - `clear` may remove all elements that were in the map when the edit was authored, even if some of those elements have since been moved elsewhere in the tree (in which case they are removed from their new location).
 - `clear` may remove all elements that are in the map when the edit is sequenced, even if some of those elements were not yet in the map when the edit was authored.
 
-This method is available on `TreeMapNodeAlpha`, which can be obtained from an existing `TreeMapNode` via `asAlpha`, or by declaring the schema with `SchemaFactoryAlpha`'s `mapAlpha`.
+`getOrInsert` returns the value at a key, first inserting a fallback value if the map has no entry for that key. The check for an existing entry is performed when the edit is authored, which affects merge outcomes: a concurrent insert by a peer can overwrite a value inserted this way, and when no insert occurs, a concurrent delete or insert by a peer determines the final state of the entry.
```
yann-achard-MS: Instead of attempting to give a condensed version of the merge semantics, which is hard to do without making mistakes and is hard to consume, I think we can rely on the users to read the doc comments on the method. So, I would just remove any mention of the merge semantics here.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -327,7 +327,7 @@ abstract class CustomMapNodeBase<const T extends ImplicitAllowedTypes> extends T
 			return existing;
 		}
 		this.set(key, fallbackValue);
-		return this.get(key);
+		return fallbackValue as TreeNodeFromImplicitAllowedTypes<T>;
```
yann-achard-MS: I'm concerned the cast may hide some cases where we can't actually rely on the `fallbackValue` being the same thing as what `this.get(key)` might return. @CraigMacomber - is there a safe way to avoid having to call `this.get(key)` here?

brrichards: It wasn't letting the return happen without the cast unfortunately so curious about this as well

CraigMacomber: This code is incorrect, and the typing is right about your value being wrong.

fallbackValue is an insertable value, which might not be a node, but might implicitly get constructed into one.

You have 4 choices for how to handle this:
1. force the insertable value into a node before inserting (Like TreeBeta.create can do), then insert that and return it
2. let `set` do the implicit construction, then look it up with `get`
3. Require input to be a node/tree value and not an insertable value
4. Remove the return value

2 is the simplest, so I'd recommend it, but 1 is also reasonable.

CraigMacomber: https://fluidframework.com/docs/api/fluid-framework/treenode-class has some details about implicit node construction.

yann-achard-MS: Thanks Craig.
@brrichards - Please have a look at option 1, but if it seems too cumbersome then I'd be ok with option 2.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -180,6 +180,63 @@ export interface TreeMapNodeAlpha<T extends ImplicitAllowedTypes = ImplicitAllow
 	 * even if some of those elements were not yet in the map when the edit was authored.
 	 */
 	clear(): void;
+	/**
+	 * Returns the value at `key`, first inserting `fallbackValue` if this map has no entry for `key`.
+	 *
+	 * @param key - The key of the element to return or insert at.
+	 * @param fallbackValue - The value to insert if `key` has no entry.
+	 * @returns The existing value at `key`, or the newly inserted `fallbackValue`.
```
Josmithr: Nit: it might be a bit better to reword this as something like:
> The value at `key`, which may be the provided `fallbackValue` if the map had no previous entry for `key`.

The current wording doesn't clearly indicate under which circumstances the user gets one or the other condition.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -180,6 +180,63 @@ export interface TreeMapNodeAlpha<T extends ImplicitAllowedTypes = ImplicitAllow
...
+	 *
+	 * @remarks
+	 * The check for the presence of an existing entry with the given `key` is performed at the time the edit is authored.
+	 * This has implications for the merge semantics of this operation:
+	 *
+	 * - If no entry is present for `key` at authoring time, thus leading to an insert,
```
Josmithr: Opinionated nit: this might flow a bit nicer with parens instead of commas for the aside about the insert case
```suggestion
	 * - If no entry is present for `key` at authoring time (thus leading to an insert)
```

(Same for the other examples below)

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -180,6 +180,63 @@ export interface TreeMapNodeAlpha<T extends ImplicitAllowedTypes = ImplicitAllow
...
+	 * while a peer concurrently deletes the entry for `key`,
+	 * then the map will end up with no entry for `key` no matter how the two edits are sequenced.
+	 *
+	 * - If an entry is present for `key` at authoring time, thus leading to no insert,
+	 * while a peer concurrently inserts a new entry for `key`,
+	 * then the map will end up with the entry set by the peer no matter how the two edits are sequenced.
+	 *
+	 * These last two points mean that, upon sequencing of an edit made with this API,
+	 * there is no guarantee that the entry for the given key will be the current one (if any) or the fallback one.
+	 * If such a guarantee is important, then consider using constraints to ensure the edit only applies when appropriate.
```
Josmithr: Is there an API we can link to for constraints here?

Josmithr: @yann-achard-MS

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -180,6 +180,63 @@ export interface TreeMapNodeAlpha<T extends ImplicitAllowedTypes = ImplicitAllow
...
+	 *
+	 * This API is not equivalent to the following alternative:
```
Josmithr: Nit
```suggestion
	 * This API is **not** equivalent to the following alternative:
```

Josmithr (commented):
Left a couple of docs suggestions. But overall they look great! Thanks for being so thorough!

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
yann-achard-MS: `TreeContextAlpha.runTransaction` takes a `RunTransactionParamsAlpha` which as a `preconditions` field. Not sure which one of those we should link. Maybe `RunTransactionParamsAlpha.preconditions`?

dannimad:

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
brrichards: updated with option 1 following sort of what set does, but if this isnt ideal can change to option 2.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
brrichards: changed the wording

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
brrichards: Thanks for providing the link. Added!

yann-achard-MS (approved):

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -180,6 +184,90 @@ export interface TreeMapNodeAlpha<T extends ImplicitAllowedTypes = ImplicitAllow
...
+	 * `callback` is only invoked when the map has no entry for `key`.
+	 * Prefer this API over {@link TreeMapNodeAlpha.getOrInsert} when producing the fallback value is expensive.
+	 *
+	 * The check for the presence of an existing entry with the given `key` is performed at the time the edit is authored.
+	 * See {@link TreeMapNodeAlpha.getOrInsert} for the implications this has on the merge semantics of this operation.
+	 *
+	 * If `callback` throws, no edit is made and the error is propagated to the caller.
+	 * If `callback` sets an entry for `key` in this map, that entry is overwritten with the value `callback` returned.
+	 */
+	getOrInsertComputed(
```
Josmithr, line 267: Be sure to update the changeset and PR title/description to cover this addition as well

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
brrichards: option 2 it is!

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
brrichards: thank you! updated

.changeset/witty-otters-insert.md
```diff
@@ -0,0 +1,32 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": tree
+---
+Add getOrInsert and getOrInsertComputed methods to TreeMapNodeAlpha
+
+[`TreeMapNodeAlpha`](https://fluidframework.com/docs/api/fluid-framework/treemapnodealpha-interface) now has `getOrInsert` and `getOrInsertComputed` methods, further aligning it with JavaScript's built-in Map API. Both return the value at a key, first inserting a value if the map has no entry for that key: `getOrInsert` takes the fallback value directly, while `getOrInsertComputed` takes a callback which is only invoked (with the key) when an insert is needed, which is preferable when producing the fallback value is expensive.
```
Josmithr: Nit: our Markdown best-practices docs generally recommend line-breaking on sentence boundaries to make future diffs easier to parse. Probably not too important for this changeset, but just a general suggestion.
```suggestion
[`TreeMapNodeAlpha`](https://fluidframework.com/docs/api/fluid-framework/treemapnodealpha-interface) now has `getOrInsert` and `getOrInsertComputed` methods, further aligning it with JavaScript's built-in Map API.
Both return the value at a key, first inserting a value if the map has no entry for that key: `getOrInsert` takes the fallback value directly, while `getOrInsertComputed` takes a callback which is only invoked (with the key) when an insert is needed, which is preferable when producing the fallback value is expensive.
```

Josmithr (approved):
Docs and API changes look good. I'll let Craig and/or Yann approve the code changes though, since they've been more involved on that side of things.

yann-achard-MS (approved):

PR #27790, opened 2026-07-29
refactor(tree): Compose IdDecodingContext in FieldBatchDecodingContext

CraigMacomber:
## Description

Refactors `FieldBatchDecodingContext` to **compose** an `IdDecodingContext` (as a public `idDecodingContext` field) instead of **subclassing** it.

Previously `FieldBatchDecodingContext extends IdDecodingContext`, so the decode context was passed directly wherever an `IdDecodingContext` was expected. That coupling only mattered at a single boundary (the field-batch codec's `decode`), while the generic `DecoderContext` already composed an `idDecodingContext` separately. Subclassing also meant that reusing an existing `IdDecodingContext` to build a `FieldBatchDecodingContext` would have required copy-constructing the base context.

With composition:

- `FieldBatchDecodingContext` holds `idDecodingContext` and `incrementalDecoder`; it is no longer an `IdDecodingContext`.
- The codec decode boundary passes `context.idDecodingContext` to `decode(...)`.
- `IdDecodingContext.hasOriginatorSessionId` is now `public` (consulted by `withIncrementalDecoder`, which previously relied on protected subclass access).
- A handful of tests that abused the subclassing to use a `FieldBatchDecodingContext` where an `IdDecodingContext` was wanted now construct an `IdDecodingContext` directly.

This is a pure internal refactor with no runtime, wire-format, or public-API change. It unblocks upcoming adoption of `IdDecodingContext` in the change codecs, where an existing `IdDecodingContext` can be reused to build a `FieldBatchDecodingContext` without copying.

### DIFF
--- a/packages/dds/tree/src/test/feature-libraries/chunked-forest/codec/checkEncode.ts
+++ b/packages/dds/tree/src/test/feature-libraries/chunked-forest/codec/checkEncode.ts
@@ -95,7 +95,7 @@ function testDecode(
 	expectedTree: JsonableTree[][],
 	identifierFilter: CounterFilter<string>,
 	version: FieldBatchFormatVersion,
-	idCompressor?: IIdCompressor,
+	idCompressor: IIdCompressor = testIdCompressor,
 	incrementalDecoder?: IncrementalDecoder,
 ): EncodedFieldBatchV1OrV2 {
 	const chunk = updateShapesAndIdentifiersEncoding(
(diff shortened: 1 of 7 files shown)


packages/dds/tree/src/test/feature-libraries/chunked-forest/codec/checkEncode.ts
```diff
@@ -111,11 +110,11 @@ function testDecode(
 	const result = decode(
 		chunk,
 		idCompressor === undefined
-			? FieldBatchDecodingContext.forOp({
+			? new IdDecodingContext({
```
Josmithr: Nit: this ternary could be moved into the constructor for the `idCompressor` property only.

Josmithr: Same below.

packages/dds/tree/src/util/compressedIds.ts
```diff
@@ -261,9 +261,13 @@ export interface IdDecoderOptionsWithOriginator {
  */
 export class IdDecodingContext {
 	/**
-	 * Used internally to prevent the use of this decoder in incremental chunks if it has a session id (which would be wrong in those chunks).
+	 * Whether this context resolves identifiers using an originator session id.
```
Josmithr: Nit
```suggestion
	 * Whether this context resolves identifiers using an originator session ID.
```

packages/dds/tree/src/test/feature-libraries/chunked-forest/codec/checkEncode.ts
CraigMacomber: Moved to a default parameter as thats even cleaner

Josmithr (approved):

PR #27791, opened 2026-07-29
tree: Move some ModularChangeFamily method implementations into static functions

alex-pardes:
## Description

Refactored the implementation of `ModularChangeFamily.invert` into a static function, along with some other modular changeset utilities.


packages/dds/tree/src/test/feature-libraries/modular-schema/modularChangeFamily.spec.ts
```diff
@@ -1607,14 +1606,14 @@ function normalizeChangeset(change: ModularChangeset): ModularChangeset {
...
 			// TODO: This relies on field kinds calling prune child on all changes,
 			// while pruning is supposed to be an optimization which could be skipped.
-			normalizedFieldChanges.set(
-				field,
-				changeHandler.rebaser.prune(fieldChange.change, normalizeNodeChanges),
-			);
+			normalizedFieldChanges.set(field, {
+				...fieldChange,
+				change: brand(changeHandler.rebaser.prune(fieldChange.change, normalizeNodeChanges)),
+			});
```
jason-ha, line 1616: What is going on here?

jason-ha (approved):
Assuming most is a refactor with really just relocated logic, then I have just one question.

packages/dds/tree/src/test/feature-libraries/modular-schema/modularChangeFamily.spec.ts
yann-achard-MS: The normalization use to strip the wrapper (I'm assuming by mistake). I'm curious why this compiled before.

yann-achard-MS (approved):

packages/dds/tree/src/test/feature-libraries/modular-schema/modularChangeFamily.spec.ts
alex-pardes: I'm also not sure why this compiled before, but I've looked into this kind of thing years ago, and typically the root cause is that Typescript's type system is just weird. I know Craig has a list of scenarios where the type checker is more permissive than you would expect. I think he's concluded that for most of these quirks, there's no convenient way to avoid these issues.

PR #27792, opened 2026-07-30
chore: enable pnpm dedupePeers to shrink lockfile

alexvy86:
## Description

Sets `dedupePeers: true` in `pnpm-workspace.yaml` and regenerates the lockfile.

`dedupePeers` (added in pnpm v10.33.0) changes the format of peer-dependency identifiers in the lockfile from fully-nested paths to flat `name@version` keys. Example:

- Before: `eslint-config-prettier 10.1.8(eslint@9.39.1(jiti@2.6.1)(supports-color@10.2.2))`
- After:  `eslint-config-prettier 10.1.8(eslint@9.39.1)`

The nested peer information still lives on each peer's own lockfile entry; it's just no longer re-encoded into every consumer's identifier. Same resolved graph, smaller lockfile.

### Impact on this repo

- `pnpm-lock.yaml`: **40,068 -> 39,881 lines** (-187), **1.69 MB -> 1.62 MB** (-72 KB, ~4.3%).
- No `package.json` changes, no dependency version changes, no changes to what actually lands in `node_modules`.

## Reviewer Guidance

Main things to sanity-check:
- The lockfile diff should be almost entirely peer-suffix flattening on existing entries - no unexpected version bumps or removed packages.
- Anyone with a stale local `pnpm-lock.yaml` on an in-flight branch will need to rebase (or accept a lockfile-config-mismatch on `--frozen-lockfile`) once this merges. Worth calling out in the merge announcement.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2,6 +2,7 @@ lockfileVersion: '9.0'
 
 settings:
   autoInstallPeers: true
+  dedupePeers: true
   excludeLinksFromLockfile: false
 
 catalogs:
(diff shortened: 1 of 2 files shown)


jason-ha (approved):
I like it.
I think lockfile will be easier to review with this.

PR #27795, opened 2026-07-30
feat(publish): record publish ownership in publish-metadata.json

MarioJGMsoft:
## Description

We are moving npm publishing out of the build pipelines and into a dedicated publish pipeline, and that needs to roll out gradually rather than as a single flag day. The publish pipeline already consumes the `publish-metadata.json` emitted alongside the packed tarballs, but nothing in that file says whether the build pipeline is still publishing the packages itself, so both could decide they own a release.

This adds a `publishAsPartOfBuild` boolean to the schemaVersion 1 metadata:

- `true` - the build pipeline publishes through its own publish stages. Current behavior, and the default.
- `false` - publishing is delegated to the dedicated publish pipeline.

It is a new parameter on the shared `include-generate-publish-metadata.yml` template, so every pipeline that emits publish metadata gets the field. The default of `true` means no pipeline changes behavior. `build-client.yml` plumbs it through `build-npm-client-package.yml` and sets it explicitly, so ownership can be handed over one release group at a time by flipping a single line. A matching consumer change is needed on the publish pipeline side to read the field and stand down when it is `true`.

The field is additive and the schema version is unchanged, so publishing behavior is unchanged here.

## Reviewer Guidance

Pipeline YAML only, so there is no unit test coverage to point at. Two things worth a look:

- **Fail fast instead of defaulting.** If the value is empty or malformed, the step errors out rather than picking a default. The reason is that this field only means anything if it agrees with how the pipeline's publish stages are actually configured, and a silent default can disagree with that in either direction. Defaulting to `true` on a pipeline that has handed publishing off means the metadata claims the build already published, the publish pipeline stands down, and the release is silently never published. Defaulting to `false` on a pipeline that still publishes inline means both sides think they own it and the packages get published twice. There is no safe guess, so failing the build is the only option that cannot produce a wrong release. Worth noting this makes it inconsistent with the adjacent `isLatest` block, which does coerce an empty value to `false`.
- **Polarity.** `publishAsPartOfBuild` is phrased from the build pipeline's perspective so the default matches today's behavior and migration means turning it off. The inverse framing would have made the safe default `false`, which seemed worse.

Verified by running the generated inline bash against real `jq`: `true`/`True`/`false`/`False` produce a genuine JSON boolean, and empty/unset/garbage fail the step.


Abe27342 (approved):
this looks fine as-is, but can you make a follow-up that drops the default on `publishAsPartOfBuild` and makes everyone explicitly pass true? It will make auditing "what's left" easier.

PR #27796, opened 2026-07-30
feat(tree): getConstraintStatus

yann-achard-MS:
## Description

Adds a `getConstraintStatus` free function that reports on the status of a SharedTreeChange.
This is a prerequisite for reporting this information to the application when a change is sequenced.

## Breaking Changes

None

### DIFF
--- a/packages/dds/tree/src/shared-tree/sharedTreeChangeFamily.ts
+++ b/packages/dds/tree/src/shared-tree/sharedTreeChangeFamily.ts
@@ -358,15 +358,17 @@ export enum ConstraintStatus {
 	Satisfied,
 	/**
 	 * At least one implicit constraint has been violated.
+	 * @remarks
 	 * Implicit constraints are those that are automatically enforced by SharedTree on all changes.
 	 *
 	 * Such a violation typically arises in one of two scenarios:
-	 * 1. A schema change conflicts with a concurrent data or schema that was sequenced before it.
+	 * 1. A schema change conflicts with a concurrent data or schema change that was sequenced before it.
 	 * 2. A data change conflicts with a concurrent schema change that was sequenced before it.
 	 */
 	ImplicitViolation,
 	/**
 	 * At least one explicit constraint has been violated.
+	 * @remarks
 	 * Explicit constraints are those that are explicitly added
 	 * through {@link RunTransactionParamsAlpha.preconditions | preconditions}
 	 * or {@link TransactionCallbackStatusAlpha.preconditionsOnRevert | preconditionsOnRevert}.
--- a/packages/dds/tree/src/shared-tree/sharedTreeChangeFamily.ts
+++ b/packages/dds/tree/src/shared-tree/sharedTreeChangeFamily.ts
@@ -347,3 +347,57 @@ export function updateRefreshers(
 		}
 	});
 }
+
+/**
+ * The status of a {@link SharedTreeChange} with respect to its constraints.
+ */
+export enum ConstraintStatus {
+	/**
+	 * All constraints are satisfied.
+	 */
+	Satisfied,
+	/**
+	 * At least one implicit constraint has been violated.
+	 * Implicit constraints are those that are automatically enforced by SharedTree on all changes.
+	 *
+	 * Such a violation typically arises in one of two scenarios:
+	 * 1. A schema change conflicts with a concurrent data or schema that was sequenced before it.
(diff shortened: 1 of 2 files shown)


packages/dds/tree/src/shared-tree/sharedTreeChangeFamily.ts
```diff
@@ -361,3 +347,57 @@ export function updateRefreshers(
...
+ * The status of a {@link SharedTreeChange} with respect to its constraints.
+ */
+export enum ConstraintStatus {
+	/**
+	 * All constraints are satisfied.
+	 */
+	Satisfied,
+	/**
+	 * At least one implicit constraint has been violated.
+	 * Implicit constraints are those that are automatically enforced by SharedTree on all changes.
```
Josmithr, line 362: Nit
```suggestion
	 * @remarks
	 * Implicit constraints are those that are automatically enforced by SharedTree on all changes.
```

packages/dds/tree/src/shared-tree/sharedTreeChangeFamily.ts
```diff
@@ -361,3 +347,57 @@ export function updateRefreshers(
...
+	 *
+	 * Such a violation typically arises in one of two scenarios:
+	 * 1. A schema change conflicts with a concurrent data or schema that was sequenced before it.
+	 * 2. A data change conflicts with a concurrent schema change that was sequenced before it.
+	 */
+	ImplicitViolation,
+	/**
+	 * At least one explicit constraint has been violated.
+	 * Explicit constraints are those that are explicitly added
```
Josmithr, line 372: Nit
```suggestion
	 * @remarks
	 * Explicit constraints are those that are explicitly added
```

packages/dds/tree/src/test/shared-tree/sharedTreeChangeFamily.spec.ts
```diff
@@ -95,6 +99,9 @@ const stDataChange1: SharedTreeChange = {
 const stDataChange2: SharedTreeChange = {
 	changes: [{ type: "data", innerChange: dataChange2 }],
 };
+const stConstraint: SharedTreeChange = {
+	changes: [{ type: "data", innerChange: constraint }],
+};
```
Josmithr, line 104: Nit: our guidelines call to avoid abbreviations like "st". This is an existing pattern here, so not worth blocking on. But we should generally try to avoid.

packages/dds/tree/src/shared-tree/sharedTreeChangeFamily.ts
```diff
@@ -361,3 +347,57 @@ export function updateRefreshers(
...
+	/**
+	 * All constraints are satisfied.
+	 */
+	Satisfied,
+	/**
+	 * At least one implicit constraint has been violated.
+	 * Implicit constraints are those that are automatically enforced by SharedTree on all changes.
+	 *
+	 * Such a violation typically arises in one of two scenarios:
+	 * 1. A schema change conflicts with a concurrent data or schema that was sequenced before it.
```
Josmithr: Nit
```suggestion
	 * 1. A schema change conflicts with a concurrent data or schema change that was sequenced before it.
```
?

packages/dds/tree/src/shared-tree/sharedTreeChangeFamily.ts
```diff
@@ -361,3 +347,57 @@ export function updateRefreshers(
...
+	 * 2. A data change conflicts with a concurrent schema change that was sequenced before it.
```
Josmithr: For my understanding, how conservative is this? Do we unilaterally consider concurrent schema/data changes to be conflicting even if the data change is compatible with both versions of the schema?

yann-achard-MS: > Do we unilaterally consider concurrent schema/data changes to be conflicting even if the data change is compatible with both versions of the schema?

Yes. This is indeed very conservative. Some day we might invest in making that less conservative.

Josmithr (approved):

packages/dds/tree/src/test/shared-tree/sharedTreeChangeFamily.spec.ts
yann-achard-MS: Good point. Will make a separate PR.

PR #27800, opened 2026-07-30
feat(tree): better formatted text insertion APIs

CraigMacomber:
## Description

Remove `defaultFormat` property from FormattedTextAsTree.Members.
Extend `insertAt` so it can insert either strings or typed text atoms with optional formatting.

## Breaking Changes

APIs are internal only (so no changeset).

Callers using `FormattedTextAsTree.Members.defaultFormat` must pass formatting directly to `insertAt` or `reformat`. References to `FormattedTextAtoms` must use `TextAtomSchemas`.

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/text/textDomainFormatted.ts
+++ b/packages/dds/tree/src/text/textDomainFormatted.ts
@@ -665,7 +665,7 @@ export namespace FormattedTextAsTree {
 		>[],
 	> extends TextAsTree.Members {
 		/**
-		 * {@link FormattedTextAsTree.Members.insertAt} with optional formatting to apply to all additional characters,
+		 * {@link TextAsTree.Members.insertAt} with optional formatting to apply to all additional characters,
 		 * and allowing an array of atoms instead of a string.
 		 * @param format - Optional formatting to apply to all additional characters. If not specified, the default formatting (from {@link FormattedTextAsTree.createSchema}) will be used.
 		 * @remarks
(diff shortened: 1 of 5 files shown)


packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -619,14 +665,20 @@ export namespace FormattedTextAsTree {
 		>[],
 	> extends TextAsTree.Members {
 		/**
-		 * Format to use by default for text inserted with non-formatted APIs.
+		 * {@link FormattedTextAsTree.Members.insertAt} with optional formatting to apply to all additional characters,
+		 * and allowing an array of atoms instead of a string.
+		 * @param format - Optional formatting to apply to all additional characters. If not specified, the default formatting (from {@link FormattedTextAsTree.createSchema}) will be used.
```
brrichards, line 670: nit: it would be nice to have a test for this path with the update to the current tests. Maybe with a schema whose configured default isn't all false, otherwise the assertion passes against any form

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -619,14 +665,20 @@ export namespace FormattedTextAsTree {
 		>[],
 	> extends TextAsTree.Members {
 		/**
-		 * Format to use by default for text inserted with non-formatted APIs.
+		 * {@link FormattedTextAsTree.Members.insertAt} with optional formatting to apply to all additional characters,
```
brrichards:
```suggestion
		 * {@link TextAsTree.Members.insertAt} with optional formatting to apply to all additional characters,
```
Should this be `{@link TextAsTree.Members.insertAt}`? As written it links to the member it's documenting.

packages/dds/tree/src/text/textDomainFormatted.ts
CraigMacomber: I have now added a "default formatting"

brrichards (approved):

PR #23796, opened 2025-02-05
refactor(odsp-driver): Remove node-fetch usage

WayneFerrao:
[AB#29142](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/29142)

## Description

This PR removes the node-fetch dependency in favour of the native fetch function.

PR #27802, opened 2026-07-30
docs: Update Client Requirements template for 3.0 release

### LINKED ISSUE #27458: TypeScript minimum version will be 6.0 with build target es2022 in v3.0
## Broken API or pattern

Fluid Framework client SDK will drop support for TypeScript below 6.0 and lib/target below `es2022`.

**Preparation References**

Use TypeScript v6.0+ or compatible compiler.
Set TypeScript configuration `"lib"` and `"target"` properties with `"ES2022"` or later.

**Packages**

All client packages

**Expected Timeline**

v3.0

Josmithr:
Updates the Client Requirements template with updates for 3.0. This content is embedded in the majority of our library package READMEs.

[AB#78964](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/78964)


jason-ha (approved):
I wonder if we should call out Jest+require-ESM issue in here.
NodeJS supports require-ESM fully as of 22.12.
Jest want's NodeJS 24.9 or later.

tools/markdown-magic/src/templates/Client-Requirements-Template.md
```diff
@@ -32,11 +33,6 @@ Node10 resolution is not supported as it does not support Fluid Framework's API
...
-    ES Modules are the preferred way to consume our client packages (including in NodeJs) and consuming our client packages from ES Modules is fully supported.
+    ES Modules are the required way to consume our client packages (including in NodeJs).
 -   CommonJs:
-    Consuming our client packages as CommonJs is supported only in NodeJS and only for the cases listed below.
-    This is done to accommodate some workflows without good ES Module support.
-    If you have a workflow you would like included in this list, file an issue.
-    Once this list of workflows motivating CommonJS support is empty, we may drop support for CommonJS one year after notice of the change is posted here.
-
-    -   Testing with Jest (which lacks [stable ESM support](https://jestjs.io/docs/ecmascript-modules) due to [unstable APIs in NodeJs](https://github.com/nodejs/node/issues/37648))
+    CommonJs is no longer officially supported as of version 3.0.
```
CraigMacomber, line 41: We might want to explicitly support require(ESM) (see https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require ) and ensure our ESM modules are compatible with that (in our case that mainly means avoiding top level await, which we currently do not use).

```suggestion
    We no longer publish CommonJs modules as of 3.0, however we do support loading our ES Modules from CommonJs using [require(ESM)](https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require) which is supported in NodeJS `>=22.12.0`.
```

tools/markdown-magic/src/templates/Client-Requirements-Template.md
```diff
@@ -14,10 +14,11 @@ When making such a request please include if the configuration already works (an
```
CraigMacomber, line 16: The above statement about NodeJS version implies we currently support Node 22 and node 24 (since node 22 is less than a year from EOL).

That's fine I guess, but we might want to adjust the phrasing to just say we support Node22 and Node24 as long as they are not EOL, or something like that.

CraigMacomber, line 16: Is the:

>- Running Fluid in a Node.js environment with the `--no-experimental-fetch` flag is not supported.

line above still relevant? Perhaps we can remove it?

tools/markdown-magic/src/templates/Client-Requirements-Template.md
CraigMacomber: Note that the change you made is stricter, which means we could do what you said for now, and then if customers need require(ESM) we could add support for it (as I suggested).

My suggestion would ban top level await in our code, where yours would not.

Thus maybe we want to do with your phrasing for now as that reserves our ability to do top level await, which is useful (like async loading wasm modules)

CraigMacomber: Either way, eventually we want to allow us to use top level await, so if we support require(ESM) in 3.0, we likely want to drop it in 4.0.

Josmithr: > Note that the change you made is stricter, which means we could do what you said for now, and then if customers need require(ESM) we could add support for it (as I suggested).

This was my thinking. I generally erred on the side of under-promising with the expectation that we may end up wanting or needing to actually deliver more. 

I'd be fine applying your suggested changes if folks would prefer, but my inclination would be to err on the side of being stricter and coming back to this as needed.

tools/markdown-magic/src/templates/Client-Requirements-Template.md
Josmithr: I'm not sure. @jason-ha any idea?

Josmithr: Updated.

tools/markdown-magic/src/templates/Client-Requirements-Template.md
```diff
@@ -8,16 +8,19 @@ When making such a request please include if the configuration already works (an
 
 # Supported Runtimes
 
--   NodeJs ^22.22.2 except that we will drop support for it [when NodeJs 22 loses its upstream support on 2027-04-30](https://github.com/nodejs/release#release-schedule), and will support a newer LTS version of NodeJS at least 1 year before 22 is end-of-life.
+-   NodeJS versions 22 and 24 for as long as they are receiving [upstream support](https://nodejs.org/en/about/previous-releases).
```
Josmithr: @jason-ha @CraigMacomber should we also be adding support for 26? Perhaps at the same time that we drop support for 22?

tools/markdown-magic/src/templates/Client-Requirements-Template.md
jason-ha: Where is copilot integration such that we can ask it right here what the history of that option is in nodejs?
From FF side, this was added Feb 2025 via #23796 per @WayneFerrao.
And per copilot the option is relevant to v22 but no longer in v24+. So, it should remain for now.

tools/markdown-magic/src/templates/Client-Requirements-Template.md
jason-ha: I think we should support all current stable versions. But until we've done at least one test I don't think we can commit to that. Checking with 26 and adding support note should be added to the backlog.

tools/markdown-magic/src/templates/Client-Requirements-Template.md
```diff
@@ -8,9 +8,11 @@ When making such a request please include if the configuration already works (an
...
+    -   Support for version 22 will be dropped [when it loses upstream support on 2027-04-30](https://github.com/nodejs/release#release-schedule).
     -   Running Fluid in a Node.js environment with the `--no-experimental-fetch` flag is not supported.
--   Modern browsers supporting the es2022 standard library: in response to asks we can add explicit support for using babel to polyfill to target specific standards or runtimes (meaning we can avoid/remove use of things that don't polyfill robustly, but otherwise target modern standards).
+-   Modern browsers supporting the es2022 standard library.
+    -   In response to requests, we can add explicit support for using babel to polyfill to target specific standards or runtimes (meaning we can avoid/remove use of things that don't polyfill robustly, but otherwise target modern standards).
```
jason-ha: I'm having trouble following this statement. Who is "we"? I think this should be addressing consumers - if this means they need to make a request, then be clear about how that request is to be made.

tools/markdown-magic/src/templates/Client-Requirements-Template.md
jason-ha: We can say not officially supported and then go on to note that conditionally (experimentally?) require-ESM may be used to use FF from CJS. We'll want to add a note about using tsconfig options required. I think those might be:
```json
  "compilerOptions": {
    "module": "Node20", // or NodeNext
    "customConditions": ["module-sync"]
  }
```
but I need to recheck. I will reply when I confirm.

tools/markdown-magic/src/templates/Client-Requirements-Template.md
jason-ha: We should add Node20 to the list.
I'd like to drop Node16, but bohemia still has that set broadly.
I haven't read all of the differences, but I know that Node20 is needed to allow ESM use by CJS. That isn't enough motivation to drop 16. I plan to use Node20 in FF when doing some CJS using ESM packages testing.

tools/markdown-magic/src/templates/Client-Requirements-Template.md
Josmithr: I'm just going to remove this statement altogether. Since we're no longer prioritizing 3P customers, we can avoid encouraging requests. If something is critical for a 3P customer, they will likely file an issue regardless.

tools/markdown-magic/src/templates/Client-Requirements-Template.md
```diff
@@ -1,23 +1,26 @@
-These are the platform requirements for the current version of Fluid Framework Client Packages.
-These requirements err on the side of being too strict since within a major version they can be relaxed over time, but not made stricter.
-For Long Term Support (LTS) versions this can require supporting these platforms for several years.
+The following are the platform requirements for using Fluid Framework client libraries.
+These requirements err on the side of being conservative, policies may be relaxed within a major version series, but not made stricter.
```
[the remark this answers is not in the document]
Josmithr:
```suggestion
These requirements err on the side of being conservative.
Policies may be relaxed within a major version series, but not made stricter.
```

jason-ha (approved):

tools/markdown-magic/src/templates/Client-Requirements-Template.md
jason-ha: While a bit dangerous, it appears that only `"module": "node20"` (or later) is required for TypeScript to be happy and import ESM for CJS. I didn't have to use `"customConditions": ["module-sync"]`.

jason-ha (commented):
We might bring some qualifications back, but we can do so as needed. Clean and clear is a good place to be.

jason-ha (approved):

PR #27804, opened 2026-07-31
build(client): rename api tasks

jason-ha:
Use new "build:entrypoints" naming.

Existing "api" and "api-extractor" tasks are preserved for other workspaces.

In 3.0 client workspace will drop CommonJS support but keep some scripts in place to make revival easy should the drop not stick.

Co-authored-by: Copilot 223556219+Copilot@users.noreply.github.com
Copilot-Session: ba45ddac-4ae6-4b63-82e6-196c38d7d485

[AB#79048](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/79048)


WillieHabi:
Ran the PR through copilot and found that since Tree does not define `build:entrypoints:esm`  or  `build:entrypoints:cjs`  (because its entrypoint files are produced directly by `build:esnext ` and  `tsc`) then running `npm --filter @fluidframework/tree build:entrypoints` generates the root Node10 redirects without producing `lib/entrypoints/*`  or  `dist/entrypoints/*`  files.

fluidBuild.config.cjs
```diff
@@ -120,27 +139,54 @@ module.exports = {
...
+		"build:entrypoints": {
+			dependsOn: [
+				"build:entrypoints:esm",
+				"build:entrypoints:cjs",
+				"build:entrypoints:node10",
+			],
+			script: false,
+		},
+		"build:entrypoints:esm": ["build:esnext"],
+		"build:entrypoints:cjs": ["tsc"],
```
WillieHabi, line 159: Tree doesn't have "build:entrypoints:esm" or "build:entrypoints:cjs" so Tree might need specific `build:entrypoints` override

jason-ha: It shouldn't need special casing. Since not all packages have entrypoint generation we have the pattern of consumers depending on build:entrypoints:esm and build:esnext or build:entrypoints:cjs and tsc.
Prior to this it would have been api-extractor:esnext and build:esnext or api-extractor:comonjs and tsc. Those original combos are still in place.

jason-ha: (I agree it is funny. `build:entrypoints` is a portion of work and not an endcap completion of being able to use the entrypoints generated.)

WillieHabi (approved):

PR #27807, opened 2026-07-31
Type erase the details of StringTextAtom to prevent implicit construction with invalid strings

CraigMacomber:
## Description

Type erase the details of StringTextAtom to prevent implicit construction with invalid strings.
This also erases its schema from the API, making schema evolution practical and property key renames non-breaking.

## Breaking Changes

These are all internal, so no changeset, and should not break any apps:

1. Users of StringTextAtomNode's constructor have to move to the static factories.
2. Users of instanceof can use Tree.is instead.
3. Users modifiying the content of a StringTextAtom must repalce the whole node now.

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/test/text/textDomainFormatted.spec.ts
+++ b/packages/dds/tree/src/test/text/textDomainFormatted.spec.ts
@@ -170,6 +170,8 @@ describe("textDomainFormatted", () => {
 				atoms.map((atom) => atom.content),
 				["a", "😀", "b"],
 			);
+			const empty = FormattedTextAsTree.StringTextAtom.fromString("");
+			assert.deepEqual(empty, []);
 		});
 	});
 
--- a/packages/dds/tree/src/test/text/textDomainFormatted.spec.ts
+++ b/packages/dds/tree/src/test/text/textDomainFormatted.spec.ts
@@ -20,6 +20,7 @@ import { setEnableExpensiveDebugAsserts } from "../../text/textDomain.js";
 import type { TextAsTree } from "../../text/textDomain.js";
 import {
 	FormattedTextAsTree,
+	StringTextAtomNode,
 	// eslint-disable-next-line import-x/no-internal-modules -- Importing code being tested
 } from "../../text/textDomainFormatted.js";
 import { describeHydration, hydrateNode } from "../simple-tree/index.js";
@@ -146,6 +147,32 @@ describe("textDomainFormatted", () => {
 		);
 	});
 
+	describe("StringTextAtom", () => {
+		it("fromCharacter creates one atom", () => {
+			assert.equal(FormattedTextAsTree.StringTextAtom.fromCharacter("a").content, "a");
+			assert.equal(FormattedTextAsTree.StringTextAtom.fromCharacter("😀").content, "😀");
+		});
+
+		it("fromCharacter rejects values which are not one character", () => {
+			assert.throws(
+				() => FormattedTextAsTree.StringTextAtom.fromCharacter(""),
+				validateUsageError(/exactly one Unicode character/),
+			);
+			assert.throws(
+				() => FormattedTextAsTree.StringTextAtom.fromCharacter("ab"),
+				validateUsageError(/exactly one Unicode character/),
+			);
+		});
+
+		it("fromString creates one atom per character", () => {
+			const atoms = FormattedTextAsTree.StringTextAtom.fromString("a😀b");
+			assert.deepEqual(
+				atoms.map((atom) => atom.content),
+				["a", "😀", "b"],
+			);
+		});
+	});
+
 	describe("formatRange", () => {
 		it("basic use", () => {
 			const text = FormattedTextAsTreeDefault.Tree.fromString("hello");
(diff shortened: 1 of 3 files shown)


packages/dds/tree/src/test/text/textDomainFormatted.spec.ts
```diff
@@ -146,6 +147,32 @@ describe("textDomainFormatted", () => {
...
+				() => FormattedTextAsTree.StringTextAtom.fromCharacter(""),
+				validateUsageError(/exactly one Unicode character/),
+			);
+			assert.throws(
+				() => FormattedTextAsTree.StringTextAtom.fromCharacter("ab"),
+				validateUsageError(/exactly one Unicode character/),
+			);
+		});
+
+		it("fromString creates one atom per character", () => {
```
Josmithr, line 167: Nit: probably worth a test for empty string input

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -68,6 +75,44 @@ const sfStatic = new SchemaFactoryAlpha("com.fluidframework.text.formatted");
...
+ * @privateRemarks
+ * This is outside the namespace so it can be exported for testing, but not package exported.
+ */
+export class StringTextAtomNode
+	extends sfStatic.object("StringTextAtom", {
+		/**
+		 * The underlying text content of this atom.
+		 * @remarks
+		 * This is typically a single Unicode code point, and thus may contain multiple UTF-16 surrogate pair code units.
+		 * Using longer strings is still valid. For example, so users might store whole grapheme clusters here, or even longer sections of text.
```
Josmithr: Nit
```suggestion
		 * Longer strings are still valid. For example, users might store whole grapheme clusters here, or even longer sections of text.
```

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -68,6 +75,44 @@ const sfStatic = new SchemaFactoryAlpha("com.fluidframework.text.formatted");
...
+		 * Anything combined into a single atom will be treated atomically, and can not be partially selected or formatted.
+		 * Using larger atoms and splitting them as needed is NOT a recommended approach, since this will result in poor merge behavior for concurrent edits.
+		 * Instead atoms should always be the smallest unit of text which will be independently selected, moved or formatted.
```
Josmithr: Nit
```suggestion
		 * Instead, atoms should always be the smallest unit of text which will be independently selected, moved or formatted.
```

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -583,16 +625,30 @@ export namespace FormattedTextAsTree {
...
 	 */
 	export interface FormattedAtom<TFormat, TText> {
+		/**
+		 * Content which is formatted.
+		 */
 		readonly content: TText;
+		/**
+		 * Formatting which is applied to the content.
+		 * @remarks
+		 * Can be reassigned of deeply mutated to edit the formatting of the content.
```
Josmithr:
```suggestion
		 * Can be reassigned or deeply mutated to edit the formatting of the content.
```

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -583,16 +625,30 @@ export namespace FormattedTextAsTree {
...
+		 */
 		format: TFormat;
 	}
 
 	/**
 	 * Portion of a string.
+	 * @remarks
+	 * Additional kinds text atoms (also known as embedded objects) which can occur inside a string can implement this.
```
Josmithr:
```suggestion
	 * Additional kinds of text atoms (also known as embedded objects) which can occur inside a string can implement this.
```

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -603,26 +659,46 @@ export namespace FormattedTextAsTree {
 	}
 
 	/**
-	 * Unit in the string representing a single character.
+	 * Static factory functions for {@link FormattedTextAsTree.(StringTextAtom:variable)}.
+	 * @privateRemarks
+	 * We type erase StringTextAtom and only provide these static factories for construction
```
Josmithr: Nit
```suggestion
	 * We type-erase `StringTextAtom` and only provide these static factories for construction
```

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -603,26 +659,46 @@ export namespace FormattedTextAsTree {
...
+	 * to reduce the chance of someone accidentally creating a text atom for a string other than a single unicode code point.
+	 * Other strings should work, but our intention is to provide no type safe API which can produce them, so an application can take their lack of existence as an invariant if they want.
```
Josmithr: Nit
```suggestion
	 * Other strings should work, but our intention is to provide no type-safe API which can produce them, so an application can take their lack of existence as an invariant if they want.
```

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -603,26 +659,46 @@ export namespace FormattedTextAsTree {
...
+	 * It is still however possible to produce them, like export/import round trips with editing in the middle of the process, of collaboration with an equivalent schema which doesn't enforce this invariant.
```
Josmithr:
```suggestion
	 * It is still however possible to produce them, like export/import round trips with editing in the middle of the process, or collaboration with an equivalent schema which doesn't enforce this invariant.
```

packages/dds/tree/src/test/text/textDomainFormatted.spec.ts
CraigMacomber: Added.

Josmithr (approved):

PR #27809, opened 2026-07-31
Include final retain in tree deltas

CraigMacomber:
## Description

Previously our array node deltas were not required to include the final retain. This is now alwayse included.

Addationally handling of the extra new line required to make quill a happe is now more robust, fixing a buig where remote copllaborators should end up with extra new lines.

## Reviewer Guidance

The review process is outlined on [this wiki page](https://github.com/microsoft/FluidFramework/wiki/PR-Guidelines#guidelines).

PR #27810, opened 2026-07-31
Better docs for tree delta types

CraigMacomber:
## Description

While authoring https://github.com/microsoft/FluidFramework/pull/27809 I had to make some assumptions about deltas, which were undocumented.

This fills in the missing documentation.

## Reviewer Guidance

This documentes my AI assisted guesses about the behaviour: this needs review from an expert who knows if these comments are correct.

### DIFF
--- a/packages/dds/tree/src/core/tree/delta.ts
+++ b/packages/dds/tree/src/core/tree/delta.ts
@@ -122,7 +122,7 @@ export interface Root<TTrees = ProtoNodes> {
 export type ProtoNodes = TreeChunk;
 
 /**
- * Represents a change to a contiguous range of a field,
+ * A change to a contiguous range of a field,
  * including nested changes described by {@link Mark.fields}.
  *
  * @remarks
--- a/packages/dds/tree/src/core/tree/delta.ts
+++ b/packages/dds/tree/src/core/tree/delta.ts
@@ -67,7 +67,7 @@ import type { TreeChunk } from "./chunk.js";
  */
 
 /**
- * Represents the change made to a document.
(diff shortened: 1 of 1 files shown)


packages/dds/tree/src/core/tree/delta.ts
```diff
@@ -122,20 +122,48 @@ export interface Root<TTrees = ProtoNodes> {
...
+ * - Attach:`attach` only: inserts `count` new nodes.
+ * `fields` must be undefined.
+ * - Replace: `attach` and `detach`: replaces `count` existing nodes with `count` attached nodes.
+ * If count is 1, `fields` may describe nested changes to the detached node.
  */
 export interface Mark {
 	/**
-	 * The number of nodes affected.
-	 * When `isAttachMark(mark)` is true, this is the number of new nodes being attached.
-	 * When `isAttachMark(mark)` is false, this the number of existing nodes affected.
```
CraigMacomber, line 131: These functions do not exist, so I rewrote the docs to not depend on them.

packages/dds/tree/src/core/tree/delta.ts
```diff
@@ -67,7 +67,7 @@ import type { TreeChunk } from "./chunk.js";
  */
 
 /**
- * Represents the change made to a document.
```
CraigMacomber, line 70: I generally dislike the pattern of putting "represents" in every single type definition's docs. All of them are inside the representation so that's implied in my opinion, and we can be more concise and to the point without confusion by omitting that.

yann-achard-MS (approved):

PR #27812, opened 2026-07-31
fix (FF AI Codespaces): Remove claude and add `-i /yolo auto`

markfields:
## Description

Changes: 
- The `dev` alias in the FF AI Codespace used `agency claude`, so switching that to `agency copilot`.
- Added `-i "/yolo auto"` to `dev` and `oce` aliases to get auto mode (LLM-monitored auto-approval)
- Removed `claude` and `dev-copilot` aliases.


alexvy86 (approved):
Seems fine to me. If someone wants to use Claude, they can still set it up manually.

PR #27814, opened 2026-07-31
feat(tree): settled event for local commits

yann-achard-MS:
## Description

`LocalChangeMetadata` now exposes an `events: Listenable<LocalCommitEvents>` property that fires a `"settled"` event.
The `"settled"` event provides details about the outcome of applying this commit.
This can be used by an application to determine whether any constraints associated with the commits were violated.

## Breaking Changes

None


packages/dds/tree/src/core/rebase/types.ts
```diff
@@ -311,6 +334,47 @@ export interface LabelTree {
...
+	 * This occurs when an implicit constraint has been violated.
+	 * Implicit constraints are those that are automatically enforced by SharedTree on all changes.
+	 *
+	 * Such a violation typically arises in one of two scenarios:
+	 * 1. A schema change conflicts with a concurrent data or schema change that was sequenced before it.
+	 * 2. A data change conflicts with a concurrent schema change that was sequenced before it.
+	 */
+	FullyDropped,
+	/**
+	 * Only the creation of new content was applied.
```
yann-achard-MS, line 402: Yep, this is a bit surprising. There are good reasons why it has to be this way, but it doesn't feel like a good idea to try to explain them to the application author.

dannimad:
Wair for https://github.com/microsoft/FluidFramework/pull/27824 to merge

packages/dds/tree/src/core/rebase/types.ts
```diff
@@ -271,6 +272,28 @@ export interface LocalChangeMetadata extends CommitMetadata {
...
+	/**
+	 * Fired once a commit has been ordered by the sequencing service.
+	 * @param outcome - information about what changes from the commit were applied or not
+	 *
+	 * @remarks
+	 * Once a commit is sequenced, the following guarantees hold:
+	 * 1. The changes carried by the commit have been persisted and other peers are able to see them.
+	 * 2. There can be no more concurrent changes sequenced before this commit, which means this commit has reached its settled form.
+	 */
+	settled(outcome: CommitOutcome): void;
```
Josmithr, line 340: What are our use-cases for this feature? Is the only goal right now to communicate whether or not associated transaction constraints were violated? Or that some/all of the changes were dropped? If so, should we consider narrowing the semantic scope here to that, rather than this more general API?

packages/dds/tree/src/core/rebase/types.ts
```diff
@@ -271,6 +272,28 @@ export interface LocalChangeMetadata extends CommitMetadata {
...
+	 * Events related to a local change that has been applied.
+	 */
+	readonly events: Listenable<LocalCommitEvents>;
+}
+
+/**
+ * Events related to a local commit that has been applied.
+ * @sealed @alpha
+ */
+export interface LocalCommitEvents {
```
Josmithr, line 286: Since we only expect this event to fire (exactly?) once per commit, would it make sense to model this as a promise rather than an event?

packages/dds/tree/src/core/rebase/types.ts
Josmithr: Might be good to add a `@privateRemarks` block that explains why here, that way devs on our team looking at this in the future have that context.

packages/dds/tree/src/core/rebase/types.ts
yann-achard-MS: The use-case is: the app wants to be able to tell when a given change has been persisted. In the presence of constraint, that's not very useful unless you also have a way of telling what exactly has been persisted (which is dependent on whether the constraints were violated and how).

Can you give me an example of what narrowing the semantic scope might look like? I don't think I really understand what that means here. Are you just wanting to change naming and docs or also wanting to make functional changes?

packages/dds/tree/src/core/rebase/types.ts
yann-achard-MS: Good idea, will do.

packages/dds/tree/src/core/rebase/types.ts
yann-achard-MS: That seems like a good question for group input. Will bring it up at standup.

packages/dds/tree/src/core/rebase/types.ts
Josmithr: How we might narrow it depended on the use-cases we're trying to enable. I was under the impression that we basically just wanted an API that answered the question "was my transaction dropped, and if so: why?". If our use-cases are wider than that, then narrowing probably doesn't make sense. It just wasn't immediately clear to me what problems we are aiming to solve with this API.

Adding examples to these docs (via `@example` blocks) and/or examples in the changeset would be helpful, I think.

yann-achard-MS: How about adding the following to the current @remarks:
```
This event can be used by applications to inform the end user that their changes have been saved (`CommitOutcome.FullyApplied`) or rejected (`CommitOutcome.FullyDropped` and `CommitOutcome.NewContentOnly`).
It can also be used to queue up a new attempt at making the rejected changes. Note however that new edits must be made outside of the event callback.
```
I can also add something similar in the changeset.

packages/dds/tree/src/core/rebase/types.ts
yann-achard-MS: Update: settled on the side of modeling this as an event.

packages/dds/tree/src/core/rebase/types.ts
Josmithr: I still think examples would be useful. E.g. having an example that demonstrates how a user might leverage this to detect a rejected transaction and attempt to re-apply its changes (especially since the changes can't be applied _within_ the associated callback). I think other examples would be useful as well.

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -845,13 +852,19 @@ export type Listeners<T extends object> = Listeners_2<T>;
 
 // @alpha @sealed
 export interface LocalChangeMetadata extends CommitMetadata {
+    readonly events: Listenable_2<LocalCommitEvents>;
```
Josmithr, line 855: Nit: I'm surprised this got emitted as `Listenable_2` given that you directly import `Listenable` in the module that contains `LocalChangeMetadata`. This isn't a big deal, but you might try importing from `core-interfaces/internal` in `types.ts` and see if that cleans this up.

yann-achard-MS: Importing it from `core-interfaces/internal` does not make a difference.

Josmithr: Okay. Thanks for trying :)

packages/dds/tree/src/core/rebase/types.ts
```diff
@@ -271,6 +272,72 @@ export interface LocalChangeMetadata extends CommitMetadata {
...
+	 * @param outcome - information about what changes from the commit were applied or not
+	 *
+	 * @remarks
+	 * Once a commit is sequenced, the following guarantees hold:
+	 * 1. The changes carried by the commit have been persisted and other peers are able to see them.
+	 * 2. There can be no more concurrent changes sequenced before this commit, which means this commit has reached its settled form.
+	 *
+	 * This event can be used by applications to inform the end user that their changes have been saved (`CommitOutcome.FullyApplied`) or rejected (`CommitOutcome.FullyDropped` and `CommitOutcome.NewContentOnly`).
+	 * It can also be used to queue up a new attempt at making the rejected changes. Note however that new edits must be made outside of the event callback.
+	 * @example
```
Josmithr: Nit: adding a preceding intro/overview description for the example would be good

packages/dds/tree/src/core/rebase/types.ts
```diff
@@ -271,6 +272,72 @@ export interface LocalChangeMetadata extends CommitMetadata {
...
+	 * 	if (metadata.isLocal) {
+	 * 		metadata.events.on("settled", (outcome) => {
+	 * 			if (outcome === CommitOutcome.FullyApplied) {
+	 * 				alert("Clear operation succeeded.");
+	 * 			} else {
+	 * 				const shouldTryAgain = confirm(
+	 * 					"The contents have changed. Do you still want to clear everything?",
+	 * 				);
+	 * 				if (shouldTryAgain) {
+	 * 					// Try again asynchronously
```
Josmithr: Nit: probably worth expanding on this to note that it is invalid to do this synchronously in the event handler.

packages/dds/tree/src/core/rebase/types.ts
Josmithr: I see that you added the example for retry on rejected - are there other examples that we should add?

Bringing the discussion back to my original question about narrowing the scope of the API: if on-rejected behavior is the only known use-case we have right now, I'm wondering if it would make more sense to just surface a "rejected" or "transactionRejected" event (at least for now) instead of "settled". If we have known use-cases for responding to non-rejected settled statuses, then this API as is probably makes sense. But if not, we could narrow the API to start and expand later as needed. If we _do_ have known use-cases for the non-rejected case, then I think we should add example docs for that as well.

yann-achard-MS: We do have another use case besides the rejected one: notifying the user that the change has succeeded and is persisted. This is called out both in the docs and in the example already.

packages/dds/tree/src/core/rebase/types.ts
yann-achard-MS: The paragraph above ("This event can be used by applications...") was meant to fulfilling that role. Is there something missing to make that clearer? Add "See example below." to the end of that paragraph?

packages/dds/tree/src/core/rebase/types.ts
Josmithr: I guess the example shows logging in response to success; I'm looking for a real application behavior that would be predicated on success. I guess maybe some UI could show a spinner or something to indicate that content is in a transient state until the dust settles?

packages/dds/tree/src/core/rebase/types.ts
Josmithr: Ah, I see. In the resulting API documentation, the order that different tags' content is rendered is not necessarily the same as how it is ordered in the comment, so that relationship may not be clear in the final state. It would probably be sufficient to just add a brief "title" to the example by adding text on the same line as the tag.

Josmithr (approved):

packages/dds/tree/src/core/rebase/types.ts
yann-achard-MS: `alert()` is a UI popup, not logging. I agree a spinner is a more realistic UI choice, but I think that will unnecessarily complicate the example.

packages/dds/tree/src/core/rebase/types.ts
yann-achard-MS: Done

Josmithr (approved):

PR #27815, opened 2026-08-03
fix(client): update TypeScript 6 compatibility

jason-ha:
In TypeScript 6, `Uint8Array` is generic over `ArrayBufferLike` which currently may be either `ArrayBuffer` or `SharedArrayBuffer`. And `ArrayBuffer` is refined to disallow using `Uint8Array` as it implicitly.

Supporting changes:
- Remove unused `@types/base64-js`
- Add helpers for careful, optimal extraction of `ArrayBuffer` or `ArrayBufferLike` from `ArrayBufferLike` or `Uint8Array`.
  - `Uint8ArrayToArrayBuffer` is split to
    1. `AnyUint8ArrayToArrayBuffer` and explicitly produces an `ArrayBuffer` that some cases expect (they don't handle `SharedArrayBuffer`)
    2. `Uint8ArrayToArrayBufferLike` that preserves the underlying buffer type (identical execution as before)

If `ArrayBufferLike` was expected and `UInt8Array` was provided, now `Uint8ArrayToArrayBufferLike` wraps the use and handles any offsets that the view may have had. (TS 5.4 was unsound for `ArrayBufferLike` allowing `Uint8Array` to satisfy that constraint. In common use of wrapping `ArrayBuferLike` with `Uint8Array` there would be no issue as `Uint8Array` constructor accepts either as single arguments. Other uses like given to `DataView` would compile fine but throw.)

If `ArrayBuffer` was expected and `ArrayBufferLike` was provided, now `ArrayBufferLikeToArrayBuffer` wraps the use and will convert from `SharedArrayBuffer` if that was the case.

In the future:
- some helpers could be removed if `ArrayBuffer` expectations were related to `ArrayBufferLike` or `Uint8Array`.
- under actual TypeScript 6 use:
  - further optimization of `AnyUint8ArrayToArrayBuffer` is possible without type check error.
  - `Uint8ArrayToArrayBufferLike` and `IsoBuffer` can be made generic.
- Blob APIs should be reviewed for ability to support `SharedArrayBuffer` and use more views (`Uint8Array`) that allow segmented references without cloning data.

[AB#46433](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/46433)

Copilot-Session: 62172399-447d-4b34-a88c-1fbb53b94268


examples/apps/blobs/src/container/blobCollection/blobCollection.ts
```diff
@@ -45,7 +45,7 @@ class BlobCollection implements IBlobCollection {
 				const newBlob: IBlobRecord = {
 					id: key,
 					// Blobs in Fluid are retrieved as ArrayBuffers, this translates it back to a Blob
-					blob: new Blob([arrayBuffer]),
+					blob: new Blob([Uint8ArrayToArrayBuffer(new Uint8Array(arrayBuffer))]),
```
ChumpChief, line 48: Is this conversion required in actuality, or just to make TS happy?  Maybe just update the `IFluidHandle<ArrayBufferLike>` cast on line 41 rather than a runtime change if the real object was already correct in practice.

packages/common/client-utils/src/bufferShared.ts
```diff
@@ -7,11 +7,18 @@
  * Converts a Uint8Array array to an ArrayBuffer.
  * @param array - Array to convert to ArrayBuffer.
  *
+ * @remarks The returned buffer may be the input array's backing buffer or a copy of its bytes.
+ * Callers should not mutate the returned buffer unless they own the input array and its backing
+ * storage, and should not rely on mutations being reflected in the input array.
```
Josmithr, line 12: This is gross. Thanks for documenting.

packages/dds/shared-object-base/src/test/attachingBindingAndConnecting.spec.ts
```diff
@@ -78,11 +79,36 @@ function createTestSharedObject(overrides: OverridableType): {
...
+			return overrides.summarizeCore.call(
+				this,
+				serializer,
+				telemetryContext,
+				incrementalSummaryContext,
+				fullTree,
+			);
+		}
+		protected async loadCore(services: IChannelStorageService): Promise<void> {
+			await overrides.loadCore?.call(this, services);
```
Josmithr, line 98: This is the only updated method here that doesn't assert the presence of its dispatch target. Intentional?

packages/dds/tree/src/test/shared-tree/schematizeTree.spec.ts
```diff
@@ -183,10 +183,10 @@ describe("schematizeTree", () => {
 			fork(): ITreeCheckout {
 				throw new Error("Function not implemented.");
 			},
-			isBranch(): boolean {
+			isBranch() {
 				return true;
 			},
-			hasRootSchema(): boolean {
+			hasRootSchema() {
```
Josmithr, line 189: Why remove the explicit return types?

examples/apps/blobs/src/container/blobCollection/blobCollection.ts
CraigMacomber, line 48: This looks like an obfuscated way to cast from ArrayBufferLike to ArrayBuffer.

I don't know why this code was introduced, but it seems like its one:
1. A hack to workaround a newly introduced typing issue which should probably be fixed rather than hidden like this.
2. Unnecessary and should be removed
3. Something else?

experimental/PropertyDDS/packages/property-dds/src/propertyTree.ts
```diff
@@ -119,8 +123,8 @@ export interface ISharedPropertyTreeEncDec {
 		decode: (IPropertyTreeMessage) => IPropertyTreeMessage;
 	};
 	summaryEncoder: {
-		encode: (ISnapshotSummary) => IsoBuffer;
-		decode: (IsoBuffer) => ISnapshotSummary;
+		encode: (ISnapshotSummary) => Uint8Array;
```
CraigMacomber, line 126: This seems like a functional design change as there are rather different types. A `Uint8Array` is a wrapper around an ArrayBuffer, with a rather different API.

IsoBuffer on the other hand was a platform specific type for ArrayBuffer in web and node's buffer locally IIRC.

I'd understand if we wanted to move from IsoBuffer to ArrayBuffer to get off of special cases for NodeJs, but that seems like if done, something that should be a dedicated change to do only that, and unrelated to TypeScript 6.

The choice to move to Uint8Array instead of a buffer type here seems like a subjective API design choice, not a TypeScript version related thing.

I'm not sure what exactly changed in TypeScript 6, but perhaps we could try and remove/replace ArrayBufferLike and IsoBuffer with the browser standard types and do so without making unrelated changes, like moving to typed arrays where we used to use buffers? I'm guessing (but not confident) such a cleanup would avoid the break caused by TS 6 entirely.

packages/common/client-utils/src/bufferShared.ts
```diff
@@ -7,11 +7,18 @@
...
-	if (array.byteOffset === 0 && array.byteLength === array.buffer.byteLength) {
-		return array.buffer;
+	if (array.buffer instanceof ArrayBuffer) {
+		if (array.byteOffset === 0 && array.byteLength === array.buffer.byteLength) {
+			return array.buffer;
+		}
+		return array.buffer.slice(array.byteOffset, array.byteOffset + array.byteLength);
 	}
-	return array.buffer.slice(array.byteOffset, array.byteOffset + array.byteLength);
+	return new Uint8Array(array).buffer;
```
CraigMacomber, line 23: If this instanceof check fails, how do we know its safe to ignore offsets? Shouldn't we worry about https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/byteOffset (and length) regardless of the buffer's concreate type?


Also, isn't `new Uint8Array(array).buffer` the same as just `array.buffer`?

Having two codepaths here with different behavior is seems error prone.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -234,9 +234,9 @@ abstract class CustomMapNodeBase<const T extends ImplicitAllowedTypes> extends T
 	public has(key: string): boolean {
 		return this.innerNode.tryGetField(brand(key)) !== undefined;
 	}
-	public keys(): IterableIterator<string> {
+	public *keys(): IterableIterator<string> {
 		const node = this.innerNode;
-		return node.keys();
+		yield* node.keys();
```
ChumpChief, line 239: Is this preferable?  I'm not familiar with the pros/cons.

packages/dds/tree/src/test/shared-tree/schematizeTree.spec.ts
jason-ha: Because the actual return type should be `this is Foo` and that is not supported in 5.4.5. 6 will support it. But leaving it implicit works for both.

packages/drivers/odsp-driver/src/compactSnapshotParser.ts
```diff
@@ -63,7 +64,10 @@ function readBlobSection(node: NodeTypes): {
 		) {
 			// "id": <node name>
 			// "data": <blob>
-			blobContents.set(blob.getString(1), blob.getBlob(3).arrayBuffer);
+			blobContents.set(
+				blob.getString(1),
+				Uint8ArrayToArrayBuffer(new Uint8Array(blob.getBlob(3).arrayBuffer)),
```
ChumpChief, line 69: I'm not familiar with the BlobDeepCopy/BlobShallowCopy but it seems like this approach defeats whatever memory strategy BlobShallowCopy was designed to provide?  No idea if that represents real value but might be worth considering a type-only fix here and logging a TODO if we're not sure.

packages/dds/shared-object-base/src/test/attachingBindingAndConnecting.spec.ts
jason-ha: This one doesn't check because it doesn't have to return anything and use of `?` is sufficient.
I think my mind was mottled when I looked at these. Only the first needs the override. For the rest they should also be able to us `?`. I will double check with TS 6 but I expect the ones below will look like this.

packages/common/client-utils/src/bufferShared.ts
jason-ha: If not `ArrayBuffer`, it could be `ShareArrayBuffer` (or some other future type). `Uint8Array` _will_ make a copy and give us confirmed `ArrayBuffer`.

packages/drivers/file-driver/src/fileDocumentStorageService.ts
```diff
@@ -120,7 +120,7 @@ export class FluidFetchReader
 			const fileName = `${this.path}/${this.versionName}/${sha}`;
 			if (fs.existsSync(fileName)) {
 				const data = fs.readFileSync(fileName);
-				return data;
+				return Uint8ArrayToArrayBuffer(data);
```
ChumpChief, line 123: I imagine this sort of change is fine (I think expectation in this particular case for example is just equivalent to `data.buffer`), but maybe consider whether you want to defer behavioral changes in favor of making this a types-only PR.

packages/runtime/runtime-utils/src/summaryUtils.ts
```diff
@@ -372,21 +372,20 @@ export function convertSnapshotTreeToSummaryTree(
...
 		let decoded: string | undefined;
-		if (snapshot.blobsContents !== undefined) {
-			// TODO Why are we non null asserting here?
-			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-			const content: ArrayBufferLike = snapshot.blobsContents[id]!;
+		if (snapshot.blobsContents === undefined) {
+			// 0.44 back-compat We still put contents in same blob for back-compat so need to add blob
+			// only for blobPath -> blobId mapping and not for blobId -> blob value contents.
+			const encodedBlob = snapshot.blobs[id];
+			if (encodedBlob !== undefined) {
```
ChumpChief, line 379: Is silent continue a good option here?  Currently this will throw with a friendly error if encodedBlob is undefined, maybe the fail-fast is valuable?

ChumpChief (commented):
Just a rough review - main feedback is that it's hard to know the implications of some of the functional changes (either in functional behavior or possible memory implications), so if the primary motivator is about TS compatibility maybe consider whether you want to separate behavioral changes.

Also, we currently declare support down to TS 5.4 (https://github.com/microsoft/FluidFramework/blob/main/ClientRequirements.md#supported-tools).  When I last looked at this, some of the typing around buffers changed in TS 5.7, so probably just be careful that your solution maintains our support guarantees.  It also might be time to revisit what versions of TS we support.

packages/dds/tree/src/test/shared-tree/schematizeTree.spec.ts
jason-ha: Well, it must have been a build fluke when I tried to use removal to work for both 5.4.5 and 6.0.3. Upon rebuild this doesn't work. So I'll remove it and we'll fix up as part of actual migration.

packages/runtime/runtime-utils/src/summaryUtils.ts
jason-ha: If we want that then that would be fine but should be its own change. This is really just a refactor with no logic change. `content` was already being checked for `!== undefined` despite the `!`.

Once shifting other things with ArrayBuffer, there doesn't appear to be any need to change this code for TS 6.

jason-ha:
@ChumpChief , @CraigMacomber , @Josmithr , I've split out the non-ArrayBuffer changes we want to keep to #27884. Once that merges we'll use this to deal with the ArrayBuffer requirements.

Preview: I rolled all of the changes back and more targeted changes with more helpers.

packages/common/client-utils/src/bufferShared.ts
jason-ha: The view handling was incorrect. Fixed now and tests added.

examples/apps/blobs/src/container/blobCollection/blobCollection.ts
jason-ha: New explicit helper added to handle this: `ArrayBufferLikeToArrayBuffer`

experimental/PropertyDDS/packages/property-dds/src/propertyTree.ts
jason-ha: This was not the best and has been rolled back. The changes are reduced to narrower set.
When we actually have TS6 we can (probably should) make `IsoBuffer` generic over `ArrayBufferLike` as `Uint8Array` is. That means no change here.

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
jason-ha: This lets the generation happen iteratively instead of collecting all of the keys and then iterating over that. This brings keys() inline with entries(). Handled in separate PR.

packages/drivers/odsp-driver/src/compactSnapshotParser.ts
jason-ha: There is a helper now that should be a no-op when the instance type is correct.
I'd like to just use `ArrayBufferLike` in the map, but `ISnapshot.blobContents: Map<string, ArrayBuffer>`. So that change is a breaking change. Probably we're always providing `ArrayBuffer` here (but I couldn't prove that against all paths). If we're good, then helper is quick check. Otherwise, it will clone to get to the correct type.

packages/drivers/file-driver/src/fileDocumentStorageService.ts
jason-ha: .buffer would be unsafe if the data were a limit view.
Passing `Uint8Array` as `ArrayBufferLike` is a problem in *some* cases but TS 5.4 is unsound and allows it.
If it just is viewed by `Uint8Array` (probably the most common), then things would be okay.
So, either we update places to allow views and not require whole buffers or we check and pay some penalty for correctness that TS6 wants.

packages/common/client-utils/package.json
```diff
@@ -264,7 +263,11 @@
 		"releasePkgVersion": "2.115.0"
 	},
 	"typeValidation": {
-		"broken": {},
+		"broken": {
```
CraigMacomber, line 266: `Uint8ArrayToArrayBuffer` was marked as `@internal`, and we claim to only type test `"entrypoint": "public"` so why is it showing up as a broken API?

packages/common/client-utils/src/bufferShared.ts
```diff
@@ -4,14 +4,73 @@
  */
 
 /**
- * Converts a Uint8Array array to an ArrayBuffer.
- * @param array - Array to convert to ArrayBuffer.
+ * Distills an ArrayBuffer from an ArrayBufferLike or creates one from SharedArrayBuffer if needed.
+ *
+ * @param buffer - ArrayBuffer or any ArrayBufferLike to convert to ArrayBuffer.
```
Josmithr: Nit (we should use `inlineCode` syntax to differentiate code references from English):
```suggestion
 * Distills an `ArrayBuffer` from an `ArrayBufferLike` or creates one from `SharedArrayBuffer` if needed.
 *
 * @param buffer - `ArrayBuffer` or any `ArrayBufferLike` to convert to `ArrayBuffer`.
```

packages/common/client-utils/src/bufferShared.ts
```diff
@@ -4,14 +4,73 @@
...
+	if (buffer instanceof ArrayBuffer) {
+		return buffer;
+	}
+	// eslint-disable-next-line unicorn/prefer-spread -- spread is not the same as slice for Uint8Array
+	return new Uint8Array(buffer).slice().buffer;
+}
+
+/**
+ * Distills a Uint8Array array to an ArrayBuffer | SharedArrayBuffer copying
+ * data as needed per the array's byteOffset and byteLength.
```
Josmithr: Nit
```suggestion
 * Distills a `Uint8Array` array to an `ArrayBuffer | SharedArrayBuffer` copying
 * data as needed per the array's `byteOffset` and `byteLength`.
```

packages/common/client-utils/src/bufferShared.ts
```diff
@@ -4,14 +4,73 @@
...
+ *
+ * @param array - Array to extract buffer from.
+ *
+ * @remarks The returned buffer may be the input array's backing buffer or a copy of its bytes.
+ * Callers should not mutate the returned buffer unless they own the input array and its backing
+ * storage, and should not rely on mutations being reflected in the input array.
+ *
+ * @privateRemarks
+ * Upon update to TypeScript 5.9 or higher, this should become a generic function
+ * over ArrayBufferLike that parameterizes Uint8Array.
```
Josmithr: Nit
```suggestion
 * over `ArrayBufferLike` that parameterizes `Uint8Array`.
```

packages/common/client-utils/src/bufferShared.ts
```diff
@@ -4,14 +4,73 @@
...
 		return array.buffer;
 	}
 	return array.buffer.slice(array.byteOffset, array.byteOffset + array.byteLength);
 }
+
+/**
+ * Distills a Uint8Array array to an ArrayBuffer copying data as needed per the array's
+ * byteOffset and byteLength or if underlying buffer is not ArrayBuffer.
+ *
+ * @param array - Array to convert to ArrayBuffer.
```
Josmithr: Nit
```suggestion
 * Distills a `Uint8Array` array to an `ArrayBuffer` copying data as needed per the array's
 * `byteOffset` and `byteLength` or if underlying buffer is not `ArrayBuffer`.
 *
 * @param array - Array to convert to `ArrayBuffer`.
```

packages/common/client-utils/src/bufferShared.ts
```diff
@@ -4,14 +4,73 @@
...
+ *
+ * @remarks The returned buffer may be the input array's backing buffer or a copy of its bytes.
+ * Callers should not mutate the returned buffer unless they own the input array and its backing
+ * storage, and should not rely on mutations being reflected in the input array.
+ *
+ * Prefer to use Uint8ArrayToArrayBufferLike if callsite doesn't require
+ * an explicit ArrayBuffer (can also handle SharedArrayBuffer).
```
Josmithr: Nit
```suggestion
 * Prefer to use {@link Uint8ArrayToArrayBufferLike} if the call site doesn't require
 * an explicit `ArrayBuffer` (can also handle `SharedArrayBuffer`).
```

packages/common/client-utils/src/bufferShared.ts
```diff
@@ -4,14 +4,73 @@
...
+		array.byteOffset === 0 &&
+		array.byteLength === array.buffer.byteLength &&
+		array.buffer instanceof ArrayBuffer
+	) {
+		return array.buffer;
+	}
+	// In TypeScript 5.9 or higher, ArrayBufferLikeToArrayBuffer can be removed.
+	// slice is used to ensure underlying buffer is ArrayBuffer.
+	// eslint-disable-next-line unicorn/prefer-spread -- spread is not the same as slice for Uint8Array
+	return ArrayBufferLikeToArrayBuffer(Uint8ArrayToArrayBufferLike(array.slice()));
```
ChumpChief: Maybe preference here, but I think it reads easier without the wrapping (mostly no-op) functions?


```suggestion
	// slice copies the buffer to ensure it's an ArrayBuffer, not a SharedArrayBuffer.
	// eslint-disable-next-line unicorn/prefer-spread -- spread is not the same as slice for Uint8Array
	return array.slice().buffer;
```

packages/common/client-utils/src/bufferShared.ts
```diff
@@ -4,14 +4,73 @@
...
+ * storage, and should not rely on mutations being reflected in the input array.
+ *
+ * @internal
+ */
+export function ArrayBufferLikeToArrayBuffer(buffer: ArrayBufferLike): ArrayBuffer {
+	if (buffer instanceof ArrayBuffer) {
+		return buffer;
+	}
+	// eslint-disable-next-line unicorn/prefer-spread -- spread is not the same as slice for Uint8Array
+	return new Uint8Array(buffer).slice().buffer;
```
ChumpChief: Does this need the Uint8Array construction or can it just slice the buffer?  Maybe also a small comment?


```suggestion
	// slice copies the buffer (a SharedArrayBuffer) to produce an ArrayBuffer.
	// eslint-disable-next-line unicorn/prefer-spread -- spread is not the same as slice for Uint8Array
	return new buffer.slice();
```

examples/apps/blobs/src/container/blobCollection/blobCollection.ts
```diff
@@ -45,7 +45,7 @@ class BlobCollection implements IBlobCollection {
 				const newBlob: IBlobRecord = {
 					id: key,
 					// Blobs in Fluid are retrieved as ArrayBuffers, this translates it back to a Blob
-					blob: new Blob([arrayBuffer]),
+					blob: new Blob([ArrayBufferLikeToArrayBuffer(arrayBuffer)]),
```
ChumpChief: This is in an example - is it representative of guidance we'd promote to customers (i.e. use our new helper function `ArrayBufferLikeToArrayBuffer`)?  I see it's `@internal` so if not maybe consider instead using whatever recommendable pattern here (perhaps just calling the buffer methods directly here)

packages/drivers/odsp-driver/src/createFile/createNewUtils.ts
```diff
@@ -87,7 +91,7 @@ function convertCreateNewSummaryTreeToTreeAndBlobsCore(
 				const contentBuffer =
 					typeof summaryObject.content === "string"
 						? stringToBuffer(summaryObject.content, "utf8")
-						: summaryObject.content;
+						: AnyUint8ArrayToArrayBuffer(summaryObject.content);
```
ChumpChief, line 94: I think use of `AnyUint8ArrayToArrayBuffer` implies that it's important to have an ArrayBuffer here and not a SharedArrayBuffer, why is that?

packages/loader/container-loader/src/utils.ts
```diff
@@ -172,7 +172,7 @@ function convertSummaryToISnapshot(
 				blobContents.set(
 					blobId,
 					summaryObject.content instanceof Uint8Array
-						? Uint8ArrayToArrayBuffer(summaryObject.content)
+						? AnyUint8ArrayToArrayBuffer(summaryObject.content)
```
ChumpChief, line 175: Similar question here re: why disallow SharedArrayBuffer

packages/common/client-utils/package.json
jason-ha: This is an internal package `@fluid-internal/client-utils` and we type check its entire surface (intentionally). It would be better if the typeValidation said "./" or "" or whatever the alternative is for the default entrypoint.
We aren't check `@public` we are checking `import * from "@fluid-internal/client-utils"`.

packages/loader/driver-utils/src/test/summaryCompresssionTester.spec.ts
```diff
@@ -137,7 +137,7 @@ class InternalTestStorage implements IDocumentStorageService {
 	}
 	async readBlob(id: string): Promise<ArrayBufferLike> {
 		return id === misotestid
-			? new TextEncoder().encode(abcContent)
+			? Uint8ArrayToArrayBufferLike(new TextEncoder().encode(abcContent))
```
ChumpChief: Suggest just accessing the buffer since it's clear here that we have a new Uint8Array

```suggestion
			? new TextEncoder().encode(abcContent).buffer
```

packages/loader/driver-utils/src/test/summaryCompresssionTester.spec.ts
```diff
@@ -581,9 +581,8 @@ async function testEncDecBinaryLoop(
 	}
 }
 
-function compareTwoBlobs(blob1: ArrayBufferLike, blob2: ArrayBufferLike): boolean {
+function compareTwoBlobs(blob1: ArrayBufferLike, blob2View: Uint8Array): boolean {
```
ChumpChief, line 582: Similar here, maybe just pass `originBlob.buffer` at the call site to keep the two sides of a `compare*` function parallel.

packages/common/client-utils/src/bufferShared.ts
ChumpChief: Doc nit: maybe clarify "as needed".  I think it's "if needed to ensure the resulting buffer exactly represents the Uint8Array's view into the data"?

ChumpChief (commented):
I think seems mostly fine, I can see there's a lot of legacy stuff here that really just wants another look so I don't want to block this PR on too much of that.  My main thoughts are:

1. Callsites that demand an ArrayBuffer (not an ArrayBufferLike) probably deserve inline comment explaining why the tighter requirement
2. Maybe prefer to just call `.buffer` in more places where it's locally apparent that the buffer is correct.
3. Some of the fixes add extra "wrapping and unwrapping" with array views and buffer access.  Probably just make sure these are as minimal as possible

I also think it's totally valid if you want to add some type casting in here (with relevant comment) to preserve current behavior if we have any doubts about the functional/performance implications of a given change (e.g. whether adding a buffer copy is acceptable).  I think the "real fix" for a lot of these places is to push `ArrayBufferLike` out of our API surfaces, so I think it'd be fair to similarly defer real typeguarding to go along with that effort.

packages/loader/driver-utils/src/test/summaryCompresssionTester.spec.ts
jason-ha: There is no guarantee that the encoder didn't allocate a working buffer that is larger than the minimum required.
In general is seems that more code should use views and not buffers. Only then convert to a whole buffer in the places that need it. But that goes beyond here.
IF the encoder always gives us a whole buffer then the penalty is just a little property check.

packages/loader/container-loader/src/utils.ts
jason-ha: I originally was more aggressive at trying to relax the typing but would hit places that didn't seem safe to alter - API exposure. So now we are conservative. If all paths are really respecting the rules, then there shouldn't be a penalty.

packages/drivers/odsp-driver/src/createFile/createNewUtils.ts
jason-ha: `Blob`s require `ArrayBuffer`. Here it is a matter of what gets surfaced at API. These changes are conservatively not changing APIs and being safe.

packages/loader/driver-utils/src/test/summaryCompresssionTester.spec.ts
jason-ha: That isn't safe. If you want symmetry, then the view wrapping can be moved out. Didn't seem worth it to me.

packages/loader/driver-utils/src/test/summaryCompresssionTester.spec.ts
ChumpChief: AFAICT the TextEncoder standard specifies it matches the length:
1. ["...result of creating a Uint8Array object given output and this’s relevant realm..."](https://encoding.spec.whatwg.org/#dom-textencoder-encode)
2. ["...result of creating a Uint8Array object from bytes in realm..."](https://encoding.spec.whatwg.org/#create-a-uint8array-object)
3. ["...result of creating an ArrayBuffer from bytes in realm..."](https://webidl.spec.whatwg.org/#arraybufferview-create)
4. ["...Let jsArrayBuffer be ? AllocateArrayBuffer(realm.[[Intrinsics]].[[%ArrayBuffer%]], bytes’s length)..."](https://webidl.spec.whatwg.org/#arraybuffer-create)

Agree on preferring not to work with buffers in so many places.  This feedback is optional either way.

packages/common/client-utils/src/bufferShared.ts
jason-ha: If you can find documentation that says slice() always copies then we could use that. If array was already ArrayBuffer backed, then slice could return a new Uint8Array with shallow clone.
This changes a bit with TS6. Here is a revised version of the function that more or less does so with an assert:
```typescript
export function AnyUint8ArrayToArrayBuffer(array: Uint8Array): ArrayBuffer {
	// Shortcut to provider underlying buffer as-is when possible.
	if (isArrayBufferBackedUint8Array(array)) {
		// Use Uint8ArrayToArrayBufferLike to deal with a partial view.
		return Uint8ArrayToArrayBufferLike(array);
	}

	// slice is used to ensure underlying buffer is ArrayBuffer.
	// It will also account for a partial view.
	// eslint-disable-next-line unicorn/prefer-spread -- spread is not the same as slice for Uint8Array
	const arrayBufferBacked = array.slice();
	assert(
		arrayBufferBacked.byteOffset === 0 && arrayBufferBacked.byteLength === array.length,
		"slice should return a view with byteOffset 0 and same length",
	);
	return arrayBufferBacked.buffer;
}
```
Instead of an assert perhaps we could have unit test to guard. That would mean that if it fails, then there is risk of code already shipped failing in tricky ways.

ChumpChief: MDN says it here "returns a copy of a portion of a typed array into a new typed array object": https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice

The ECMAScript standard is harder to follow but it does indicate that it does a byte-by-byte copy ([23.2.3.27 step 8.f.ix](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.slice)) as opposed to pointing to the same buffer.  Supporting this being a new backing buffer is step 7 of the slice algorithm which indicates creation of a new typed array with countBytes (which will do an allocation per [23.2.5.1 step 10](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-typedarray))

packages/loader/driver-utils/src/test/summaryCompresssionTester.spec.ts
jason-ha: I found that we have plenty of _test_ code that does TextEncoder...buffer.
When I found this instance of failure, I looked at how it was succeeding. Turns out it was (is) never used so I originally just added .buffer and commented that it was never used. I lost the comment when I redid things and had the safer helper.
Now skirting this altogether with asserting that path isn't taken and have removed the TextEncoder use.

packages/common/client-utils/src/bufferShared.ts
jason-ha: Documented and now using the simpler path with just `.buffer`.

ChumpChief (approved):

CraigMacomber (approved):

PR #27817, opened 2026-08-03
build(client): fix depend on everything for build:entrypoints:node10

jason-ha:
Without any dependency config, `build:entrypoints:node10` task depended on all tasks before (default of `after: ["^*"]`).


MarioJGMsoft (approved):
lgtm

PR #27820, opened 2026-08-03
build(client-PropertyDDS): Node16 and module support

jason-ha:
Before this change property-properties and property-common only supported CommonJS. property-dds ESM would load CommonJS and ESM versions of property-changeset.

ts2esm was used to update sources within property-properties and property-common. Those were then further revised to handle destructuring that was not possible using "import". (The prior ESM output was mostly just another version of CommonJS where there were some dynamic "import"s that probably didn't work.)

For tests, converted in body `require` calls to static imports making ESM tests viable.

property-properties CommonJS type interface src/index.d.ts has been rewritten for ESM as src/api.ts.

## Reviewer Guidance

Reviewing each commit one at a time may be easier than all at once. Note that some clumsiness of property-properties src/index.d.ts requiring CommonJS as primary in first commit ([Node16 and module support](https://github.com/microsoft/FluidFramework/commit/ed59823de395cf4dc914edfc7b0b6a4975682646)) was resolved only in fifth commit ([refactor CJS api.ts to ESM](https://github.com/microsoft/FluidFramework/commit/d5270c6ad0cd410bf2ea54a7c441b251f15b588f)).


ChumpChief (approved):
I see there's a test failure, but otherwise seems fine I think

jason-ha:
> I see there's a test failure, but otherwise seems fine I think

Yeah - build config issue (that is only likely to happen on CI) should be fixed now

ChumpChief (approved):

PR #27821, opened 2026-08-03
fix(server): validate Historian document ownership

zhangxin511:
## Description
[MSRC] [130146], fix cross tenant access

## Breaking Changes

N/A

## Reviewer Guidance

- **fix(server): isolate document static cache by tenant**
- **fix(server): add Historian document ownership boundary**
- **fix(historian): protect summary reads and deletes**
- **fix(historian): protect non-initial summary writes**
- **test(historian): complete ownership regression matrix**


zhangxin511:
Addressed the review findings in `6feb200cfb` and `6c088c98c4`.

- **H1 — fixed:** `initial=true` no longer unconditionally bypasses validation. Initial summary upload now performs a fresh Alfred document lookup before any tenant lookup, service construction, cache access, or storage access. Bootstrap is allowed only when Alfred explicitly reports absence (`null` or a direct 404). Any existing document, including one scheduled for deletion, is rejected as an initial replay. Dependency failures and malformed successful responses fail closed.
- **H2 — evaluated; authorization caching was not added:** the authorization decision depends on mutable lifecycle state, especially `scheduledDeletionTime`. The existing static/ephemeral cache does not contain that authoritative state, so using it for authorization could allow summary access after soft deletion. Protected summary GET/POST/DELETE therefore retain the fresh Alfred read. Any performance optimization should happen in the authoritative provider/Alfred path rather than caching an authorization decision that can become stale.
- **M1 — fixed:** `ignoreEphemeralFlag` is now preserved on the ownership-validated path. It does not bypass tenant/document identity, scheduled-deletion, or replay checks; it only disables ephemeral expiration/cache behavior and constructs the validated document as durable. Coverage verifies the ignored-expiry behavior, no ephemeral cache write, and continued scheduled-deletion rejection.

The Historian package build succeeds and the full suite passes with **90 tests**.

server/historian/packages/historian-base/src/routes/summaries.ts
```diff
@@ -125,25 +164,53 @@ export function create(
...
-			postEphemeralContainerChecker,
-			query,
-		});
+		let service: RestGitService;
+		if (initial === true) {
+			await utils.validateInitialSummaryUpload({
+				tenantId,
+				authorization,
+				documentManager,
+			});
```
Toritos01, line 172: Question for my understanding: I see that we are trying to replace the createGitService with createProtectedSummaryService in these changes. What is the reason for not using createProtectedSummaryService in the initial summary upload case?

Is the functionality of creating the protected service replaced by this validateInitialSummaryUpload call?

server/historian/packages/historian-base/src/routes/summaries.ts
```diff
@@ -152,17 +219,13 @@ export function create(
...
-			tenantService,
-			storageNameRetriever,
-			documentManager,
-			cache,
-			allowDisabledTenant: true,
-			ephemeralDocumentTTLSec,
-		});
+			"delete",
+			"notApplicable",
+			true,
```
Toritos01, line 227: Another question to make sure my understanding is correct:
So the purpose is to replace the generalized "createGitService" with a more restricted client that can only do actions that are necessary by the API that is using that gitService? For example, this case where deleteSummary is given access to "delete" permission.

Toritos01 (approved):
Changes look good, just added two comments for some questions I had :)

yunho-microsoft (approved):

server/historian/packages/historian-base/src/routes/summaries.ts
```diff
@@ -96,25 +102,55 @@ export function create(
...
-		sha: string,
-		useCache: boolean,
+		operation: utils.SummaryOperation,
+		routeType: utils.SummaryRouteType,
+		allowDisabledTenant = false,
 		query?: Query,
-	): Promise<IWholeFlatSummary> {
-		const service = await utils.createGitService({
+	): Promise<RestGitService> {
+		await utils.validateSummaryDocument({
```
zhajie, line 113: IMO it's better to check if it's ephemeral somewhere and exclude ephemeral for this case, as ephemeral has more traffic but ephemeral is not on azure storage.

server/historian/packages/historian-base/src/routes/summaries.ts
zhangxin511: `validateInitialSummaryUpload` replaces only the normal ownership validation during bootstrap, not service creation. A legitimate initial upload occurs before Alfred persists the document record, so `validateSummaryDocument` would reject it as missing. The bootstrap check allows only exact absence, rejects any existing record as replay, and propagates unexpected Alfred failures. The route then constructs the normal service through `createGitService`.

server/historian/packages/historian-base/src/routes/summaries.ts
zhangxin511: `createProtectedSummaryService` does not return a restricted Git client. It performs the fresh validation and then constructs the same `RestGitService` through the existing routing path. `"delete"` is validation/telemetry context; it does not grant a capability. The `true` argument is `allowDisabledTenant`, which permits cleanup for a disabled tenant. Deletion still occurs only through the explicit `service.deleteSummary(...)` call.

server/historian/packages/historian-base/src/routes/summaries.ts
zhangxin511: Good point—the reported Azure-storage disclosure does not apply identically to ephemeral containers because their Redis repositories are tenant-scoped, and the additional traffic is a valid concern. I would avoid exempting them in this security fix, though, because the fresh lookup also enforces current lifecycle state. Before that lookup, Historian only has cached/static ephemeral classification, which does not include mutable state such as `scheduledDeletionTime`; skipping validation could leave a soft-deleted ephemeral summary accessible until Redis TTL expiry and make authorization semantics depend on the storage backend. I suggest keeping the uniform gate here and handling performance separately by optimizing the authoritative Alfred/provider lookup.

PR #27827, opened 2026-08-03
Improve comparison base resolution for review skills

CraigMacomber:
## Description

The existing review skill was often counterproductive for me as it was very specific about what to review against, often worse than what the model would come up with on its own. These changes correct its assumptions, making the skill much more generally useful, even with less powerful models, explicitly directing it to handle removes with different names (like my upstream), compare to the merge point with the target branch not its head, handle local changes (untracked, modifications and staged changes) etc.

The new approach checks for an existing PRs as a heuristic to find the correct upstream branch and remote, then looks for an upstream for the https://github.com/microsoft/FluidFramework repo, regardless of its name.

It also warns if the merge base and target branch are too out of sync by commit count, not just diff size (which was slow and confusing when I hit it in my fork due to using the wrong remote).

I also refactored our two existing skills which hard coded origin/main to both use this extracted common logic for improved consistency and maintainability.

I used this skill to review itself, with a couple models and at a couple points in the process (like when it was just local changes on main, when it was staged on a branch, and after the PR was created. Even Haiku 4.5 could follow the steps and produce a review using this skill (its review quality wasn't great, but it found the correct diff to review).

## Reviewer Guidance


.claude/skills/comparison-base/SKILL.md
```diff
@@ -0,0 +1,124 @@
...
+
+Otherwise, query the GitHub commit-to-PR API (`repos/{owner}/{repo}/commits/$REVIEW_OID/pulls`) for each GitHub repository named by configured remotes, including `microsoft/FluidFramework`. Treat HTTP 404 or 422 as no commit candidates and continue; locally-ahead or unpublished commits may not exist in the queried repository. If `$REVIEW_REPOSITORY` and `$REVIEW_HEAD_BRANCH` are known, also query open PRs by that full head repository owner plus branch so locally-ahead commits can resolve their published PR. Read `state`, `base.repo.clone_url`, `base.ref`, `head.repo.full_name`, `head.ref`, `head.sha`, and `html_url` from the REST response. Only accept open PRs whose head repository and branch match the known review identity and whose head SHA matches `$REVIEW_OID` or is its ancestor.
+
+Do not rely on bare `gh pr view` or an unqualified branch name. These are ambiguous in fork checkouts. If multiple open PRs match, ask the user which PR to use.
```
tylerbutler: I know `gh` won't work in all cases, but I think these `gh` commands will reveal the right info to the agent in many cases and may be easier for it to follow:

`gh pr view --json url,state,baseRefName,headRefName,headRefOid,headRepository`
`gh repo view --json nameWithOwner,defaultBranchRef`

tylerbutler (commented):
It seems like the bundle size test has to do something similar where it figures out the base commit - but I guess it's easier in CI where the remotes are fixed.

CraigMacomber:
> It seems like the bundle size test has to do something similar where it figures out the base commit - but I guess it's easier in CI where the remotes are fixed.

It's not just the fixed remotes, but in our PR CI you know that there is always a PR with a clear merge target. This skill can work without a PR, and even without a branch (review local changes). It only uses PRs as an optional heuristic mostly so that if your branch is merging into something other than main it can still give a good review.

.claude/skills/comparison-base/SKILL.md
CraigMacomber: I have incorporated this approach as an option (with some tweaks to make it robust with no default repository), and also adjusted the process so it stop when it find a good branch to reduce extra tool calls.

zhangxin511 (approved):

PR #22391, opened 2024-09-04
Revert test timeout back to 2 seconds for local server

tyler-cai-microsoft:
Fixes [AB#8942](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/8942)

Reverts local server e2e test timeout back to 2 seconds.

Depends on this PR: https://github.com/microsoft/FluidFramework/pull/22390

PR #21091, opened 2024-05-15
Use public client auth for odsp e2e flows

Abe27342:
## Description

Currently, we use the ropc flow with a confidential client to authenticate in our e2e/stress tests against odsp. This is less appropriate than using a public client flow, since our application registration really only needs to have delegated permissions.
This PR adjusts things to use a public flow--I've updated the relevant app registrations to allow both forms of auth already.

Using a public flow also means our infrastructure setup here aligns exactly with [this relatively recent MSFT guidance](https://learn.microsoft.com/en-us/entra/identity-platform/test-automate-integration-testing?tabs=dotnet).

PR #7549, opened 2021-09-22
Split e2e different version testing in separate jobs.

curtisman:
- Get quicker result for non-compat version
- Make it easy to tell which which "version" pass or fail
- Finish faster with parallel jobs.

Also change the mocha-common config to reuse the same config parsing code to get the test variant and report prefix.
This also allow switches to be used to select driver instead of environment variables.

PR #27828, opened 2026-08-03
test(odsp): add end-to-end coverage for point-in-time container load

sonalideshpandemsft:
Adds real-service end-to-end coverage for loading an ODSP container at a target sequence number
(`loadContainerToSequenceNumber` + `OdspPointInTimeDocumentServiceFactory`). 

New specs under `test-end-to-end-tests/src/test/pointInTime/`:

- **`loadSuccess`** — loads at the earliest recoverable version, determinism of loading the same target twice, that the result is a disconnected read-only view pinned at the target, and replay across a deep history.
- **`loadToSequenceNumber`** — targets on a version boundary, targets mid-stream (requiring op replay from the base), and two distinct targets each reflecting their own state.
- **`epochMismatch`** — restoring a previous or middle version bumps the file's epoch, so a load whose base predates the restore fails rather than replaying ops across a lineage boundary.
- **`loadFailure`** — restore-induced epoch mismatch, a target beyond retained ops, aborting an in-flight load mid-replay, and malformed targets rejected up front with a `UsageError`.
- **`odspVersionApi`** — the ODSP version REST surface these tests depend on: snapping a version via metadata and restoring a previous version.

Supporting changes:

- `pointInTimeTestUtils.ts` / `odspVersionTestApi.ts` — shared helpers for driving version history (list, snap, restore) and building a point-in-time load.
- `OdspTestDriver` gains `createPointInTimeDocumentServiceFactory` (NoCompat only — it imports the factory directly rather than through the versioned driver api) and `getStorageTokenForResource`, needed for the raw ODSP REST calls the version setup makes.
- `odsp-driver` now exports `getApiRoot` and `getHeadersWithAuth` so the test API can build authenticated version-history requests.

[Pipeline run](https://dev.azure.com/fluidframework/internal/_build/results?buildId=413697&view=logs&j=70bcf72b-4313-5650-47ec-21056eb37a56&t=a18b750b-42b2-5f72-ae5c-19a5e9502048)


packages/test/test-end-to-end-tests/src/test/pointInTime/loadFailure.spec.ts
```diff
@@ -0,0 +1,378 @@
...
+				);
+			};
+
+			await assert.rejects(
+				loadTo(-1),
+				(error: Error) => /non-negative integer/i.test(error.message),
+				"a negative target should be rejected up front",
+			);
+			await assert.rejects(
+				loadTo(1.5),
```
anthony-murphy, line 372: Deep Review: The malformed-target assertions here (`loadTo(-1)`, `loadTo(1.5)` rejecting with `UsageError`) test pure loader input-validation — `loadContainerToSequenceNumber` validates `Number.isInteger(...)` and `< 0` before any driver/network work. But they live inside `setupPointInTimeSuite`, whose `before` hook skips every non-odsp driver (`pointInTimeTestUtils.ts` — `if (provider.driver.type !== "odsp") { this.skip(); }`), so this public-contract check only runs on credentialed ODSP legs. The suite's own doc-comment calls it "a cheap, deterministic guard ... before any network work," yet it never runs on the default non-ODSP legs.

Add a lightweight always-run (no-credentials, no-network) test asserting `loadContainerToSequenceNumber` rejects negative / non-integer / `NaN` targets with a `UsageError`, outside the odsp-gated suite, so the validation is covered everywhere.

packages/test/test-end-to-end-tests/src/test/pointInTime/pointInTimeTestUtils.ts
```diff
@@ -0,0 +1,337 @@
...
+	const documentId = createDocumentId();
+	const container = await createAttachedPointInTimeContainer(
+		provider,
+		runtimeFactory,
+		tracker,
+		documentId,
+	);
+	const dataObject = (await container.getEntryPoint()) as IPointInTimeTestObject;
+	const versionApi = createOdspVersionTestApiProps(provider, container);
+	const summarizer = options.withSummarizer
```
anthony-murphy: Deep Review: `createPointInTimeTestContext` creates a summarizer when `options.withSummarizer` is set, stores it on the context, and uses it, but exposes no teardown. The only registered teardown is `afterEach(() => tracker.reset())`, and `reset()` only closes containers in the tracker's `containers` map — the summarizer's container (destructured away when the `IContainer` from `createSummarizerFromFactory` is discarded) is never registered. Every `withSummarizer: true` test therefore leaves a live summarizer client connected to real ODSP.

Close the summarizer (and its discarded container) in the fixture teardown, matching the manual `summarizer.close()` lifecycle in `summarizeWithSearch.spec.ts`.

anthony-murphy:
`packages/drivers/odsp-driver/src/index.ts:29`

Deep Review: This adds `getApiRoot` to the `odspUrlHelper.js` export line and a new `export { getHeadersWithAuth }` — both `@internal`, exposed solely so `odspVersionTestApi.ts` can build raw version-history REST requests. This re-expands an export surface that owners deliberately trimmed: #20488 "Stopped exporting the `@internal` items that didn't need to be exported," and #5968 pushed back on new `index.ts` exports ("Who (outside of driver package) needs this?"). It doesn't break api-report (`@internal` exports are untracked there), so this is a contract/ownership matter, not a red build.

Add a short comment on both exports noting they exist solely for point-in-time test infrastructure, so a future trimming pass doesn't silently remove them and break the test build — and get sign-off from an export-surface owner (vladsud / ChumpChief). Inlining the trivial `getHeadersWithAuth` is a lower-preference alternative, but it would duplicate the driver's `assert(!!authHeader, ...)` auth-header invariant in test code.

packages/test/test-drivers/src/odspTestDriver.ts
```diff
@@ -451,4 +473,18 @@ export class OdspTestDriver implements ITestDriver {
...
+
+	/**
+	 * Fetches a storage-scoped access token for the given ODSP resource.
+	 *
+	 * @remarks
+	 * Exposed for test infrastructure that needs to make raw ODSP REST calls outside the driver
+	 * (e.g. point-in-time version setup: listing, restoring, and snapping file versions). The
+	 * returned value is the raw access token, not an `Authorization` header value.
+	 */
+	public async getStorageTokenForResource(
```
shlevari, line 485: Is there anything from a security standpoint we need to be doing with this token passing it across the boundary like this? I think it should be fine, but just want to call it out. Like it needs to be clear to callers they can't log it, etc

packages/test/test-end-to-end-tests/src/test/pointInTime/epochMismatch.spec.ts
```diff
@@ -0,0 +1,260 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/*
+ * Point-in-time failure scenarios (real service).
```
shlevari: nit: unclear what is meant by (real service) in this comment

packages/test/test-end-to-end-tests/src/test/pointInTime/epochMismatch.spec.ts
```diff
@@ -0,0 +1,260 @@
...
+				const ctx = await createPointInTimeTestContext(suite, apis, { withSummarizer: false });
+				const { container, versionApi, documentId, incrementAndSync, snapVersion } = ctx;
+
+				// Arrange two snapped versions so there is an older, non-tip version to restore to. The
+				// target seq is captured before the restore so the load is bound to the pre-restore lineage.
+				await incrementAndSync(2);
+				const targetSequenceNumber = container.deltaManager.lastSequenceNumber;
+				await snapVersion("snap-a");
+				await incrementAndSync(2);
+				await snapVersion("snap-b");
```
shlevari, line 70: is it guaranteed that once snapVersion is awaited, it will be returned in listFileVersions? I thought this wasn't a guarantee from ODSP and want to make sure we don't create a flaky test. I definitely could be wrong though

packages/test/test-drivers/src/odspTestDriver.ts
sonalideshpandemsft: I wonder if it's worth moving the odsp apis under odsp-driver. that way, the `getStorageTokenForResource` method is not exported

packages/test/test-end-to-end-tests/src/test/pointInTime/epochMismatch.spec.ts
sonalideshpandemsft: snapVersion calls the odsp api to trigger version by updating the metadata `description`. This immediately snaps a new odsp version and this is the same api being used by page history service. I can add an assert to check if a new version is snapped

anthony-murphy:
## Deep Review

*Reviewed commit `2ce6a94` on 2026-08-04.*

No correctness, security, or API-break defects. This is a well-structured, additive, ODSP-only NoCompat real-service E2E suite for point-in-time container load. Both substantive inline items from the prior review are resolved in this commit: the `withSummarizer` teardown leak is closed via `fixture.onCleanup`, and the input-validation coverage gap is closed by the new always-run `loadContainerToSequenceNumber.spec.ts`. What remains is surface-widening and pipeline-visibility polish that needs owner sign-off or an author pointer — not code rework.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add a test-infrastructure-only comment on the two new `@internal` exports (`index.ts`, `getUrlAndHeadersWithAuth.ts`) so a future trimming pass doesn't silently remove them, or move the ODSP test apis under `odsp-driver` so neither symbol needs exporting
- [ ] Add a JSDoc note on `getStorageTokenForResource` that the return is a raw access token (not an `Authorization` header) that callers must not log
- [ ] Link the recurring pipeline that runs `test:realsvc:odsp:pointInTime` so the ~1500 lines of new coverage aren't at risk of silent bit-rot

### Context for Reviewers

- `createPointInTimeDocumentServiceFactory` deliberately bypasses the versioned `this.api` test-driver indirection and imports `OdspPointInTimeDocumentServiceFactory` directly from `@fluidframework/odsp-driver/internal` — NoCompat-only by design, since the factory is a brand-new capability with no prior versions to cross-test. Confirm every point-in-time spec runs through the NoCompat provider so the direct-import factory is never exercised under cross-version compat.
- The two new `index.ts` exports (`getApiRoot`, `getHeadersWithAuth`) re-expand an `@internal` surface that #20488 and #5781 deliberately trimmed and #5968 previously pushed back on. They land on the `/internal` entrypoint (not the public surface) and reuse existing production helpers. Owners to tag: vladsud / ChumpChief.
- `OdspTestDriver` token/factory additions build on the FIC-based test-tenant token work in #26714 (Abe27342). The point-in-time script (`--driver=odsp --timeout=60s`, no `--odspEndpointName`) and per-case 120s timeouts follow the slow-suite isolation pattern from #7549 / #22391; #21105 fixed e2e tests that ran against ODSP PROD when the endpoint was unset — confirm the intended endpoint when wiring into CI.
- The suite's real-service fidelity hinges on a subtlety: a driveItem version snapshot captures only *persisted* Fluid state, so `snapVersion` forces a summary before PATCHing item `description`, then id-diffs `/versions` before/after to prove a genuinely new version appeared rather than the seq-0 creation snapshot.

- **Needs human judgment** — export-surface sign-off: vladsud or ChumpChief should confirm the two new `@internal` re-exports (`getApiRoot`, `getHeadersWithAuth`) are acceptable for test-infra use, given #20488 / #5781 deliberately trimmed such exports. This is an ownership/taste call the author can't settle alone.
- **Needs human judgment** — an ODSP auth owner (per the FIC rework #26714) should confirm exposing `getStorageTokenForResource` publicly on the shared `OdspTestDriver` is acceptable, or agree on scoping the raw-token path to the point-in-time helper module. A reviewer already flagged the security-awareness angle inline.
- **Needs human judgment** — confirm the NoCompat guarantee: every point-in-time spec must be wired through the NoCompat provider so the direct-`/internal`-import factory is never run under cross-version compat.
- **Cannot be assessed by the pipeline** — the suite's correctness depends on live ODSP `/versions` newest-first ordering, epoch-bump-on-restore semantics, and op-replay determinism. Validate via the linked Azure pipeline run.

</details>

- `d1600c1` 2026-08-03 · **8/10** — No correctness, security, or API-break defects; three additive polish items flagged inline.

</details>

packages/test/test-drivers/src/odspTestDriver.ts
anthony-murphy: Deep Review: Building on the security-awareness point — this line promotes a private token accessor to public API. `getStorageTokenForResource` (`odspTestDriver.ts:485-489`) is a pure visibility promotion: its whole body is `return this.getStorageToken(options);`, and `getStorageToken` (:455) plus every other token accessor on the class are `private`. The promotion exists so out-of-driver test infra (`odspVersionTestApi.ts:151-180`) can make raw ODSP REST calls. `OdspTestDriver` has historically been kept narrow — auth reworks #26714, #21091, #6569 never exposed raw token accessors.

Mitigations: the package is `@fluid-private/test-drivers`, the method requires explicit resource-scoped `OdspResourceTokenFetchOptions` (not a parameterless token dump), and the same callback already flows into `OdspDocumentServiceFactory`. So this is an internal-surface ownership question, not a user-facing break.

Two concrete asks: (1) keep a JSDoc note that the return is a raw access token (not an `Authorization` header) that callers must not log — addressing shlevari's point directly; (2) have an ODSP auth owner familiar with the FIC rework (#26714) confirm the raw-token REST path, and weigh sonalideshpandemsft's proposal to move the ODSP apis under `odsp-driver` so `getStorageTokenForResource` need not be exported at all — that also resolves the `getApiRoot` / `getHeadersWithAuth` export-widening in one move.

anthony-murphy (approved):

PR #27829, opened 2026-08-03
feat(tree): add isStagedUpgradeEnabled query

jenn-le:
## Description

Adds `isStagedUpgradeEnabled` to `TreeViewAlpha` — a query API that checks whether a given staged schema upgrade has already been applied to a document's stored schema. Returns a three-value `StagedUpgradeStatus` (`"disabled"` | `"partial"` | `"enabled"`).

This is useful for rollout safety checks (e.g. conditionally including an upgrade token after a feature flag rollback) and conditionally enabling UI features based on the document's schema state.

**Key changes:**
- `TreeViewAlpha.isStagedUpgradeEnabled(upgrade)` — returns `StagedUpgradeStatus` from cached upgrade data
- `StagedUpgradeStatus` type — `"disabled" | "partial" | "enabled"` (exported from `@alpha`)
- Upgrade location tracking integrated into the existing `getDiscrepanciesInAllowedContent` schema walk via `UpgradeLocationCollector` interface (single-pass, no duplicate traversal)
- `checkSchemaCompatibility` now returns `enabledUpgrades` alongside its existing compatibility status
- `SchematizingSimpleTreeView` caches enabled upgrades computed during compatibility checks
- `ITreeViewConfigurationAlpha` tagged `@input`
- Website docs updated with `isStagedUpgradeEnabled` usage and `ApiLink` components

## Reviewer Guidance

- Upgrade location collection happens as a side effect during the existing discrepancy generator walk (`discrepancies.ts`) — the `UpgradeLocationCollector` callbacks fire for each staged type/optional field encountered.
- When `canView` is false, the schema walk breaks early so `enabledUpgrades` may be incomplete (documented on the API).
- `computeUpgradeStatuses` is a private helper in `schemaCompatibilityTester.ts` that converts raw location counts into the three-value status.


.changeset/enable-staged-schema-upgrades.md
```diff
@@ -0,0 +1,161 @@
+---
```
Josmithr: This looks like an out-of-date changeset? Seems to be related to your previous PR, rather than this one. Or am I missing something?

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
Josmithr: Should we instead spread these other properties in? That way we don't accidentally forget to include some new optional property added to the tree view configuration types?

Josmithr, line 456: Nit: docs? 😁

packages/dds/tree/src/simple-tree/api/configuration.ts
Josmithr, line 164: This type is exclusively used as input to the tree library, right? If so, we should tag it as `@input`

packages/dds/tree/src/simple-tree/api/configuration.ts
```diff
@@ -167,13 +168,14 @@ export interface ITreeViewConfigurationAlpha<
 	 * Policy for generating stored schema from the view schema.
```
Josmithr: This seems a bit too general. This property is specific to staged upgrades.

packages/dds/tree/src/simple-tree/api/configuration.ts
```diff
@@ -167,13 +168,14 @@ export interface ITreeViewConfigurationAlpha<
...
 	 *
 	 * @remarks
+	 * This policy is fixed at view construction time and cannot be changed afterwards.
+	 *
+	 * If provided, this policy is used directly for compatibility checks and for
```
Josmithr: Nit: can we link to the mentioned APIs here?

packages/dds/tree/src/simple-tree/api/configuration.ts
```diff
@@ -167,13 +168,14 @@ export interface ITreeViewConfigurationAlpha<
...
+	 * `initialize` / `upgradeSchema` schema generation.
+	 *
 	 * If omitted or `undefined`, defaults to {@link StagedSchemaUpgradePolicyFactory.restrictive}
```
Josmithr: Nit: You can use the `@defaultValue` tag for this.
```suggestion
	 * @defaultValue {@link StagedSchemaUpgradePolicyFactory.restrictive}
```

packages/dds/tree/src/simple-tree/api/stagedUpgradeQuery.ts
```diff
@@ -0,0 +1,207 @@
...
+	isMapNodeSchema,
+	isObjectNodeSchema,
+	isRecordNodeSchema,
+} from "../node-kinds/index.js";
+import type { TreeSchema } from "../treeSchema.js";
+
+import { tryStoredSchemaAsArray } from "./customTree.js";
+
+/**
+ * Returns true iff at least one location guarded by `upgrade` in `viewSchema` is already enabled
```
Josmithr: Nit: let's avoid the "iff" abbreviation.
```suggestion
 * Returns true if and only if at least one location guarded by `upgrade` in `viewSchema` is already enabled
```

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -644,6 +622,15 @@ export interface TreeViewAlpha<
 	 */
 	initialize(content: InsertableField<TSchema>): void;
 
+	/**
+	 * Returns true iff at least one location guarded by the provided staged `upgrade` token
+	 * is already enabled in the document's stored schema.
```
Josmithr: Nit: I would recommend breaking this up into 2 separate statements: the summary element should be expressed semantically, and then a `@returns` block can be expressed in terms of the value being returned. E.g.
> Checks if the provided staged `upgrade` token is enabled in the document's stored schema.
and
> @returns `true` if and only if...

(also we should avoid "iff")

.changeset/staged-upgrade-query-api.md
```diff
@@ -0,0 +1,28 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+
```
Josmithr:
```suggestion
```

.changeset/staged-upgrade-query-api.md
```diff
@@ -0,0 +1,28 @@
...
+TreeViewAlpha can now query whether a staged schema upgrade has been applied
+
+A new `isStagedUpgradeEnabled` method on `TreeViewAlpha` checks whether a given `SchemaUpgrade` token has already been applied to a document's stored schema.
```
Josmithr: Nit: link to `TreeViewAlpha` API docs?

.changeset/staged-upgrade-query-api.md
```diff
@@ -0,0 +1,28 @@
...
+
+Additionally, `initialize` and `upgradeSchema` now automatically preserve already-enabled upgrades.
```
Josmithr: Nit: API docs links?

packages/dds/tree/docs/main/runtime-schema-upgrade.md
```diff
@@ -206,3 +205,91 @@ Clients should not need their local feature flag to be enabled in order to read
 
 The `upgrades` property names are not persisted.
 Compatibility depends on the schema and the `SchemaUpgrade` values, not on the application-specific flag names used at call sites.
+
+## Query If an Upgrade is Enabled
+
+The core question is, given a `SchemaUpgrade`, is it enabled in this document's stored schema?
+A `SchemaUpgrade` is considered enabled when, for at least one location where it is used in view schema, the result of the upgrade exists in the corresponding stored schema location.
```
Josmithr: Can this ever be partial? If an upgrade has been applied anywhere, doesn't that mean it's been applied _everywhere_?

Josmithr: Given some of the below details, it seems like the answer to this is no. I'm a bit surprised by this. What does it mean for a document schema to be partially upgraded?

packages/dds/tree/docs/main/runtime-schema-upgrade.md
```diff
@@ -206,3 +205,91 @@ Clients should not need their local feature flag to be enabled in order to read
...
+
+- `true`: at least one staged member guarded by that specific `SchemaUpgrade` is present in stored schema.
+	- the decision to define true as being at least one location rather than all of them is due to the usage of this api for guarding against downgrades during rollout, if any staged member has already been upgraded, it must be continued to be upgraded, if the user does not wish for this to happen, they should use a different `SchemaUpgrade` for any members that should not be upgraded as a result
+- `false`: otherwise.
+
+Important constraints:
+
+- This query requires both view schema and stored schema context.
+- It cannot be answered from stored schema alone because guarded locations are discovered from staged annotations in view schema.
+- Other upgrade tokens are irrelevant to the result for the queried token.
```
Josmithr: What does this mean? It isn't immediately clear to me.

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
```diff
@@ -97,6 +98,11 @@ export class SchematizingSimpleTreeView<
 	 * Undefined iff uninitialized or disposed.
 	 */
 	private currentCompatibility: SchemaCompatibilityStatus | undefined;
+	/**
+	 * Cached set of enabled schema upgrades, computed alongside compatibility.
+	 * Undefined iff uninitialized or disposed.
```
Josmithr:
```suggestion
	 * @remarks `undefined` only when uninitialized or disposed.
```

website/docs/data-structures/tree/schema-evolution/feature-flag-schema-upgrades.mdx
```diff
@@ -162,16 +162,14 @@ const config = new TreeViewConfiguration({ schema: Board });
 
 ## Understanding Irreversibility
 
-:::warning
-Once a document's stored schema has been upgraded, it cannot be downgraded. Do not disable a feature flag that has already been enabled in production — there is currently no API to determine which documents have been upgraded, so you cannot safely roll back without risking incompatibility for clients that encounter already-upgraded documents. An API for querying per-document upgrade status is in progress.
-:::
+Once a document's stored schema has been upgraded, it cannot be downgraded. However, rolling back a feature flag is safe — this is handled automatically. When a document has already been upgraded, `initialize` and `upgradeSchema` automatically detect the enabled upgrades in stored schema and continue including them, even if the feature flag is off. You do not need to manually check or pass upgrade tokens for already-upgraded documents. With the feature flag rolled back, new documents and any documents that haven't been upgraded yet will not have the upgrade enabled.
```
Josmithr: Nit: our Markdown guidelines generally recommend line-breaking along sentence boundaries. Makes future diffs a bit easier to reason about.

Also, can we link to the API docs for `initialize` and `upgradeSchema`?

Josmithr: We have an `ApiLink` helper component you can leverage for this (instead of hard-coded URLs). See `shortLinks.tsx`.

packages/dds/tree/docs/main/runtime-schema-upgrade.md
jenn-le: It can if an upgrade was added, enabled, and then also added to another location later. This practice is discouraged in our docs but is technically possible. I added a test for it called "counts an upgrade as enabled if it's enabled in at least one location"

jenn-le: Also, it doesn't stay in this state for very long because we're using this api to automatically include already enabled upgrades so any new locations an upgrade gets added to will be almost immediately enabled.

Josmithr: Is there any possibility of documents being left in a partially upgraded state such that future steps in the migration could leave them unreadable?

jenn-le: Nope, being partially upgraded just means that the type can be written in one location but not another but can be read from all locations

.changeset/staged-upgrade-query-api.md
```diff
@@ -0,0 +1,27 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+TreeViewAlpha can now query whether a staged schema upgrade has been applied
+
+A new [`isStagedUpgradeEnabled`](https://fluidframework.com/docs/api/tree/treeviewalpha-interface#isstagedupgradeenabled-methodsignature) method on [`TreeViewAlpha`](https://fluidframework.com/docs/api/tree/treeviewalpha-interface) checks whether a given `SchemaUpgrade` token has already been applied to a document's stored schema.
+
+Additionally, [`initialize`](https://fluidframework.com/docs/api/tree/treeviewalpha-interface#initialize-methodsignature) and [`upgradeSchema`](https://fluidframework.com/docs/api/tree/treeviewalpha-interface#upgradeschema-methodsignature) now automatically preserve already-enabled upgrades.
```
CraigMacomber: I'm not sure this is the behavior that we want.

I imagine this makes "canUpgrade" true in more cases, and makes it a lot easier to end up in cases where your schema don't follow a clear linear path which we can easily reason about and test.

CraigMacomber: Its great "automatically preserve already-enabled upgrades" becomes practical to implement, and it might be a good default, but making it default requires some real careful consideration. The more conservative approach, which never writes anything but the restrictive version, has basically no risk of causing documents to become unusable if you use schema snapshot compatibility testing, and avoids making any further changes to documents if they have unexpected (possibly future, possible messed up) schema, even if valid. Those limitations and thus simplicity and predictability have some benefits.

Maybe adding a build in restrictive except for already enabled option, without changing the default, might be a safer option.

Also, now that I think about it, isn't it still impossible to implement "automatically preserve already-enabled upgrades" as a policy, since doing so requires access to the view, and forces the functions to be in-pure which violates the API requirements and can break caching and cause potential stale results?

.changeset/staged-upgrade-query-api.md
```diff
@@ -0,0 +1,27 @@
...
+The effective upgrade policy is the union of the configured `stagedUpgradePolicy` and any upgrades already present in the stored schema, preventing accidental regression of previously applied upgrades.
+
+```typescript
+const view = tree.viewWith(
+	new TreeViewConfigurationAlpha({
```
CraigMacomber, line 12: Generally I'd consider this an anti-pattern, as it makes schema snapshot compatibility testing impossible: there is no TreeViewConfiguration which you can use in your test.

I think the main use-case for this should be things like deciding which content type creation buttons to put in the tool bar so you only enable creating types allowed in the document.

.changeset/staged-upgrade-query-api.md
```diff
@@ -0,0 +1,27 @@
...
+		schema: mySchema,
+		stagedUpgradePolicy: featureFlag.isEnabled
+			? StagedSchemaUpgradePolicy.enabledStagedUpgrades(myUpgrade)
+			: StagedSchemaUpgradePolicy.restrictive,
+	}),
+);
+
+// Check if the upgrade was already applied to this document
+if (view.isStagedUpgradeEnabled(myUpgrade)) {
```
CraigMacomber, line 21: I think it would be best to have this feature in its own PR without changing the default behavior of can upgrade.

Its not risky to introduce an API like this to query the state, and its useful even without runtime upgrades.

Consider code populating a toolbar with kinds of content to insert somewhere: it can use this API to gate including of the new content type. If you do it that way, new features are more fully available sooner (when some add using the document upgrades it, not just when your version of the app is upgraded), and the dynamic upgrade trick is enough to test them.

It might be a good idea to make the change to the defaults you mentioned above, but I think that really is its own thing which we should review and consider independently.

.changeset/staged-upgrade-query-api.md
CraigMacomber: With the change to the default behavior, this pattern might be ok, except schema compatibility testing: I'm not sure how to handle that. I think we should avoid presenting patterns we don't have a good design for testing in our changesets when there are other options. Given that in this example we are opening a preexisting tree which a future client might have upgraded, I don't think we need to complicate this example with complex patterns like this.

packages/dds/tree/docs/main/runtime-schema-upgrade.md
```diff
@@ -206,3 +205,95 @@ Clients should not need their local feature flag to be enabled in order to read
...
+Recommended API shape:
+
+```typescript
+// On TreeViewAlpha:
+isStagedUpgradeEnabled(upgrade: SchemaUpgrade): boolean
+```
+
+Boolean meaning:
+
+- `true`: at least one staged member guarded by that specific `SchemaUpgrade` is present in stored schema.
```
CraigMacomber: I don't like the convention of having non-human facing documentation where it being for AI is unclear from the file name, path and diff and thus the intended audience is unclear in PRs. How am I supposed to review this? Check that its accurately redundant with the APi docs? Happen to memorize that the ~3rd line in this file is a note saying this is for AI?

If this disagrees with he API docs, which is supposed to change?

I'd rather have a single source of truth for documentation of a parameter in a given API, which should the API doc comments.

What does it mean to have a "recommended API shape" for something that is implemented (that will be the case if this merges). Wouldn't this document be out of date if it merges, so I should object to it?

packages/dds/tree/docs/main/runtime-schema-upgrade.md
```diff
@@ -206,3 +205,95 @@ Clients should not need their local feature flag to be enabled in order to read
...
+Additionally, `initialize` and `upgradeSchema` automatically include all already-enabled upgrades in the effective upgrade policy.
+This ensures that schema operations never regress upgrades that have already been applied to the stored schema — even if the caller's `stagedUpgradePolicy` does not explicitly include them.
+The effective policy is the union of the configured `stagedUpgradePolicy` and the cached set of enabled upgrades.
+
+Conceptually, the algorithm is:
+
+1. Find all staged members in view schema guarded by the queried token.
+2. If none exist, return `false`.
+3. Check whether guarded members are present in stored schema at their corresponding locations.
+4. Return `true` if at least one guarded member is present.
```
CraigMacomber: I don't think that's the right policy.

I think there is a huge difference between one location in a document where a schema is being staged happening to have been upgraded, and it being fully upgraded. Returning true when doing a upgrade to enable it would not be a no-op is concerning.

I think it might make more sense to return a enum (or string union) with three values (disabled, partially enabled, and enabled, or something like that)

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
```diff
@@ -94,9 +95,14 @@ export class SchematizingSimpleTreeView<
 	private flexTreeContext: Context | undefined;
 
 	/**
-	 * Undefined iff uninitialized or disposed.
+	 * Undefined only when uninitialized or disposed.
```
CraigMacomber: Technically, this is less specific than "if and only if" as it does not actually specify if the "uninitialized or disposed" stage will always have this as undefined, or simply allows it to be undefined.

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
```diff
@@ -280,6 +284,33 @@ export class SchematizingSimpleTreeView<
 		this.runSchemaEdit(() => this.checkout.updateSchema(newSchema));
 	}
 
+	public isStagedUpgradeEnabled(upgrade: SchemaUpgrade): boolean {
+		this.ensureUndisposed();
+		if (this.currentEnabledUpgrades === undefined) {
+			// Force computation if not yet computed (shouldn't normally happen after init).
```
CraigMacomber: this sounds like its not considering the case where someone does a schema change (for example a remote client, or even the local client). Not sure if you have a cache invalidation bug, or this comment is just choosing not to consider such cases (considering them nor normal).

I'd prefer documentation to not focus on what is normal, but what is allowed/valid as sating what normally happens is an easy way to mislead people into being able to assume its the only supported case.

Generally edge cases matter just as much for correctness, so pointing out whats normal isn't very helpful.

The exception to that is for performance optimizations, where noting the common case can make a lot of sense (But I think common case is better terminology than normal as it better implies that uncommon cases are valid and supported where one might think abnormal cases can assert/error)

packages/dds/tree/docs/main/runtime-schema-upgrade.md
jenn-le: I'll just get rid of it. I was thinking it could be useful but it's usefulness would prob lie more in something like a progress doc e.g. this is what is implemented and this is what is planned.

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
```diff
@@ -280,6 +284,33 @@ export class SchematizingSimpleTreeView<
...
+	 * and any upgrades already enabled in the current stored schema.
+	 * This ensures that initialize/upgradeSchema never regresses already-enabled upgrades.
+	 */
+	private get effectiveUpgradePolicy(): StagedSchemaUpgradePolicy {
+		const base = this.stagedUpgradePolicy;
+		const enabled = this.currentEnabledUpgrades;
+		if (enabled === undefined || enabled.size === 0) {
+			return base;
+		}
+		return {
```
CraigMacomber: If I'm remembering the implementation of the stored schema conversion correctly, this is likely to have one of two issues (or manage to have both I suppose):

1. Allocating a new object for this policy every time will cause cache misses.
2. This policy might report different things in the future if the set is modified. I'm guessing currentEnabledUpgrades is replaced if a remote client does an upgrade for example, but if not, that's a problem.

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
```diff
@@ -280,6 +284,33 @@ export class SchematizingSimpleTreeView<
...
+		}
+		return this.currentEnabledUpgrades?.has(upgrade) ?? false;
+	}
+
+	/**
+	 * Returns an effective upgrade policy that includes both the configured `stagedUpgradePolicy`
+	 * and any upgrades already enabled in the current stored schema.
+	 * This ensures that initialize/upgradeSchema never regresses already-enabled upgrades.
+	 */
+	private get effectiveUpgradePolicy(): StagedSchemaUpgradePolicy {
```
CraigMacomber: This approach seems rather dependent on the fact that a given upgrade is either enabled of disabled, and doesn't handle one which is partially enabled. That seems like a problem.

Its perfectly possible to use one staged upgrade in two places in your schema, then for the next version of the app to bake in one of the upgrades and not the other. I think this will cause the older version of the app to upgrade the schema to one newer than the future version would normally do, causing a second upgrade implicitly that could break support for even older clients unexpecidly and in a way missed by our snapshot testing.

packages/dds/tree/src/simple-tree/api/stagedUpgradeQuery.ts
```diff
@@ -0,0 +1,245 @@
...
+import {
+	isArrayNodeSchema,
+	isMapNodeSchema,
+	isObjectNodeSchema,
+	isRecordNodeSchema,
+} from "../node-kinds/index.js";
+import type { TreeSchema } from "../treeSchema.js";
+
+import { tryStoredSchemaAsArray } from "./customTree.js";
+
```
CraigMacomber: Isn't most of this file just reimplementing schema traversal? We already have utilities for walking schema?

Also I think these might not handle recursive types correctly like our preexisting utilities do, since i don't see them taking in visited sets to detect that.

.changeset/staged-upgrade-query-api.md
jenn-le: I think staged upgrades as a concept, even their enablement, is a view schema concept so access to the view schema is always required when we're dealing with them. Without access to it, we wouldn't know what parts of the stored schema are staged in the view, unless we're persisting more info to the stored schema which I chose to avoid.

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
jenn-le: This was handled wrong, it should have the same lifetime as currentCompatibility and so in this case, should call failDispose

packages/dds/tree/src/simple-tree/api/stagedUpgradeQuery.ts
jenn-le: changed to use the same walk as get discrepancies and added a test for recursive types

.changeset/staged-upgrade-query-api.md
```diff
@@ -0,0 +1,24 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+TreeViewAlpha can now query whether a staged schema upgrade has been applied
+
+A new [`isStagedUpgradeEnabled`](https://fluidframework.com/docs/api/tree/treeviewalpha-interface#isstagedupgradeenabled-methodsignature) method on [`TreeViewAlpha`](https://fluidframework.com/docs/api/tree/treeviewalpha-interface) checks whether a given `SchemaUpgrade` token has already been applied to a document's stored schema.
+
```
Josmithr, line 11: Nit: adding a brief note about how/why a user may want to leverage this new capability would be useful. We give an example below, but giving it a bit of an English lead-in would be good.

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -489,7 +489,7 @@ export namespace FluidSerializableAsTree {
     export type Data = JsonCompatible<IFluidHandle>;
     const // @system
     _APIExtractorWorkaroundObjectBase: TreeNodeSchemaClass<"com.fluidframework.serializable.object", NodeKind.Record, TreeRecordNodeUnsafe<readonly [() => typeof FluidSerializableObject, () => typeof Array, LeafSchema<"string", string>, LeafSchema<"number", number>, LeafSchema<"boolean", boolean>, LeafSchema<"null", null>, LeafSchema<"handle", IFluidHandle<unknown>>]> & WithType<"com.fluidframework.serializable.object", NodeKind.Record, unknown>, {
-    readonly [x: string]: string | number | IFluidHandle<unknown> | System_Unsafe.InsertableTypedNodeUnsafe<LeafSchema<"boolean", boolean>, LeafSchema<"boolean", boolean>> | FluidSerializableObject | Array | null;
+    readonly [x: string]: string | number | IFluidHandle<unknown> | FluidSerializableObject | Array | System_Unsafe.InsertableTypedNodeUnsafe<LeafSchema<"boolean", boolean>, LeafSchema<"boolean", boolean>> | null;
```
Josmithr: You'll need to regenerate the reports via a clean build

packages/dds/tree/src/simple-tree/api/discrepancies.ts
```diff
@@ -48,6 +49,25 @@ import type { TreeSchema } from "../treeSchema.js";
 
 import { tryStoredSchemaAsArray } from "./customTree.js";
 
+/**
+ * Collects upgrade location data during a schema walk.
+ * Called for each staged type or staged optional field encountered.
```
Josmithr: Nit
```suggestion
 * @remarks Called for each staged type or staged optional field encountered.
```

packages/framework/fluid-framework/api-report/fluid-framework.alpha.api.md
```diff
@@ -680,7 +680,7 @@ export namespace FluidSerializableAsTree {
     export type Data = JsonCompatible<IFluidHandle>;
     const // @system
     _APIExtractorWorkaroundObjectBase: TreeNodeSchemaClass<"com.fluidframework.serializable.object", NodeKind.Record, TreeRecordNodeUnsafe<readonly [() => typeof FluidSerializableObject, () => typeof Array, LeafSchema<"string", string>, LeafSchema<"number", number>, LeafSchema<"boolean", boolean>, LeafSchema<"null", null>, LeafSchema<"handle", IFluidHandle<unknown>>]> & WithType<"com.fluidframework.serializable.object", NodeKind.Record, unknown>, {
-    readonly [x: string]: string | number | IFluidHandle<unknown> | System_Unsafe.InsertableTypedNodeUnsafe<LeafSchema<"boolean", boolean>, LeafSchema<"boolean", boolean>> | FluidSerializableObject | Array | null;
+    readonly [x: string]: string | number | IFluidHandle<unknown> | FluidSerializableObject | Array | System_Unsafe.InsertableTypedNodeUnsafe<LeafSchema<"boolean", boolean>, LeafSchema<"boolean", boolean>> | null;
```
Josmithr: Looks like these reports need to be regenerated too

packages/dds/tree/src/simple-tree/api/schemaCompatibilityTester.ts
```diff
@@ -24,23 +39,53 @@ import type { SchemaCompatibilityStatus } from "./tree.js";
  * @param viewSchema - Schema for the view
  * @param stored - The stored schema to check compatibility against
  * @param stagedSchemaUpgrades - Staged schema upgrades enabled for this view, or explicit stored-schema generation options
+ *
+ * @remarks
+ * When `canView` is false, `enabledUpgrades` may be incomplete because the schema walk is
+ * interrupted at the first discrepancy rather than exhausting all locations.
```
Josmithr, line 130: Is this going to be a problem? Would it make sense to parameterize this API such that we ensure a full walk when `enabledUpgrades` is needed, and omit it entirely otherwise (or something)?

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -644,6 +645,26 @@ export interface TreeViewAlpha<
...
 
+	/**
+	 * Checks whether a staged schema upgrade has been applied to the document's stored schema.
+	 *
+	 * @param upgrade - The upgrade token to check.
+	 *
+	 * @returns The enablement status of the upgrade:
+	 * - `"disabled"` — no locations guarded by the upgrade are enabled.
+	 * - `"partial"` — at least one location is enabled but not all of them.
+	 * - `"enabled"` — all locations guarded by the upgrade are enabled.
```
Josmithr: Nit: these details are already captured by the `StagedUpgradeStatus` type's docs right? Should we just omit them here?

packages/dds/tree/src/test/simple-tree/api/schemaCompatibilityTester.spec.ts
```diff
@@ -558,9 +576,212 @@ describe("checkSchemaCompatibility", () => {
...
+						enabledUpgrades: new Map(),
+					},
 				);
 			});
 		});
 	});
 });
+
+describe("checkSchemaCompatibility enabledUpgrades", () => {
+	const sf = new SchemaFactoryAlpha("findEnabledUpgradesTest");
```
Josmithr: Nit: our guidelines call to generally avoid abbreviations. Could we call this "schemaFactory" (or something) instead?

packages/dds/tree/src/test/simple-tree/api/schemaCompatibilityTester.spec.ts
```diff
@@ -558,9 +576,212 @@ describe("checkSchemaCompatibility", () => {
...
+		assert(stored.tryUpdateRootFieldSchema(toUpgradeSchema(baseSchema).rootFieldSchema));
+		assert(stored.tryUpdateTreeSchema(schemaStatics.number));
+
+		const config = new TreeViewConfigurationAlpha({ schema: baseSchema });
+		const { enabledUpgrades } = checkSchemaCompatibility(config, stored);
+		assert.equal(enabledUpgrades.size, 0);
+	});
+
+	it("detects enabled staged allowed type upgrade", () => {
+		const stagedStr = sf.staged(sf.string);
```
Josmithr: Nit: our guidelines call to generally avoid abbreviations. Could we call this "stagedString" instead?

Josmithr: (Same feedback for other occurences)

Josmithr (approved):
I left a question and a couple of nitpick suggestions. Otherwise, this looks good to me.

packages/dds/tree/src/simple-tree/api/schemaCompatibilityTester.ts
jenn-le: I think I prefer that we pick one policy or the other. This seems like a level of configuration that I don't think we should add until someone explicitly asks for it. I went with this one because I think it's reasonable for us to say that if the schemas are incompatible then this information is incomplete and if you want to use this information to see if the incompatibility is due to upgrade enablement, this incompleteness is actually better since it would include the first enabled upgrade that caused a problem. Also then we don't have to complete the full schema walk. I think the argument for completing the walk to get the rest of this information is also reasonable, we don't do the walk often enough that it would be a perf bottleneck.

Josmithr (approved):

PR #21837, opened 2024-07-10
Offline: Compare incoming Batch ID with pending local Batch ID to detect container forking

markfields:
## Description

Fixes [AB#8285](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/8285)

#21767 added Batch ID to batch metadata, but didn't do anything with it besides preserve it across reconnect/rehydrate.  This PR looks for Batch ID on incoming batches and compares the ID with the next Batch ID in the PendingStateManager.  Two interesting cases:

* For a `local` message, if the batchIds don't match then something has gone terribly wrong (similar to comparing op contents or `batchStartCsn`. We just log here for now.
* For a non-`local` message, if the batchIds _do_ match, then the container has forked and this container's pending local content has already been sequenced (at least partially - we will revisit this in the future).  We throw a DataProcessingError here now to protect from double-applying those changes.

## Reviewer Guidance

TODO:

- [x] punted: Unit Test updates for PSM (I saw the new code exercised stepping through e2e tests). Will make a note in the Test Improvements feature on the backlog
- [x] Fix e2e test with timing issue

This is nearly the final piece of this prototype PR: https://github.com/microsoft/FluidFramework/pull/21266

PR #13273, opened 2022-12-06
Some refactoring related to the runtime's incoming op processing pipeline

andre4i:
Part of ADO:2089.

This is mainly extracting the op pre-processing logic out of the container runtime and adding some more test coverage to the chunking logic.

This is part 2 of 3 from https://github.com/microsoft/FluidFramework/pull/13131

PR #27834, opened 2026-08-04
feat(container-runtime): add internal version mark resolver

lindsnguyen:
## Description

Adds a runtime-owned **version mark resolver** (all `@internal`) that turns a version mark's pending
`batchId` into a durable global `sequenceNumber`, so a host can view or restore a document as of a
specific edit — including **across sessions** and after the client that made the edit is gone. This is
the Fluid-side half of "version history for long-running / agentic (LWS) page edits": mark **storage
and the stored locator shape stay app-side**; Fluid owns only resolution.

`ContainerRuntime` exposes an `@internal` `versionMarkResolver` (`IVersionMarkResolver`). A host:

- captures a mark from `getCurrentPendingBatchId()` / `getCurrentSequenceNumber()`,
- live-promotes pending marks via `onBatchSequenced(...)` (fires for any connected client), and
- resolves a stored mark via `resolve(batchId, referenceSequenceNumber)` -> `ResolveResult`
  (`resolved` / `pending` / `unresolvable`), feeding a `resolved` sequence number to the loader's
  point-in-time load (`loadContainerToSequenceNumber`).

`resolve()` is live-then-history: an ephemeral in-session `batchId -> sequenceNumber` map (bounded by
evicting below the MSN, mirroring `DuplicateBatchDetector`), then an out-of-session scan of historical
ops via a driver-agnostic reader injected through `IContainerContext.fetchOps` (the read counterpart of
`submitFn`). On a scan miss, `classifyMiss` distinguishes `pending` (not sequenced yet) from
`unresolvable` (ops trimmed) using a read-derived availability signal.

No public API surface is added — everything is `@internal`, so there is no api-report change and no
changeset. Design rationale and the consumer/graduation surface are documented in
`packages/runtime/container-runtime/src/versionMarks/DEV.md`.

## Reviewer Guidance

- **Release tags:** everything is `@internal` (only `IVersionMarkResolver` + `ResolveResult` are exported
  from the package index; `fetchOps` is an `@internal` member on the `@legacy @beta` `IContainerContext`).
  This is intended for a first-party dogfood consumer, so no api-report change / changeset — please confirm
  that framing.
- **`classifyMiss` (pending vs unresolvable):** the availability signal is a documented **interim**
  heuristic derived from how the driver responds to a trimmed range; a dedicated driver
  op-availability/retention API is the intended follow-up. See DEV.md.
- **Design context** (mark-agnostic loader, the `fetchOps` seam, MSN-bounded cache, the app-owned
  locator shape) lives in `versionMarks/DEV.md`.
- **Validation:** 31 version-marks unit tests plus the inbound-batch mapping tests; full
  `container-runtime` mocha suite passes.


packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -3288,6 +3311,20 @@ export class ContainerRuntime
 				}
 			}
 
+			// Record the version-mark point for a completed batch (its last op's sequence number),
+			// carrying the batch id across a piecemeal batch's messages.
+			const versionMarkUpdate = inboundVersionMarkUpdate(
```
anthony-murphy: Deep Review: The version-mark update fires before pending-state / fork validation. In `processInboundMessageOrBatch`, `inboundVersionMarkUpdate(inboundResult, this.versionMarkInboundBatchId)` then `this.versionMarkResolverInternal.processInboundBatch(...)` (3316-3324) mutate the map **and** synchronously invoke every `onBatchSequenced` listener (versionMarkResolver.ts:205-207). Only afterward does `this.pendingStateManager.processInboundMessages(inboundResult, local)` run (3334) — the point that throws `DataProcessingError` for a remote batch matching a pending local batchId (fork detection, #21837) and for local pending-content mismatches (pendingStateManager.ts:649-657, 723-771).

Because the broadcast fires first, an `onBatchSequenced` listener can observe and promote a batch validation is about to reject. The notification is synchronous and irreversible to an external app store.

Move the `processInboundBatch` update + `onBatchSequenced` notification to after `processInboundMessages` succeeds, reusing the same `inboundResult`, so a fork/pending-content mismatch that throws cannot leave an app having already promoted the mark. Add a test asserting no notification fires when validation throws.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,224 @@
...
+			let inTargetBatch = false;
+			let firstScannedSequenceNumber: number | undefined;
+			while (true) {
+				const result = await stream.read();
+				if (result.done) {
+					break;
+				}
+				for (const op of result.value) {
+					firstScannedSequenceNumber ??= op.sequenceNumber;
+					const metadata = asBatchMetadata(op.metadata);
```
anthony-murphy: Deep Review: The cross-session history scan silently misses chunked batches. `resolveFromHistory` derives each op's batch identity inline over raw fetched ops — `const opBatchId = metadata?.batchId ?? (typeof op.clientId === "string" ? generateBatchId(op.clientId, op.clientSequenceNumber) : undefined)` (127-134). These ops come straight from `IHistoricalOpReader.fetchMessages` and are **not** run through `RemoteMessageProcessor` / `OpSplitter` unpacking.

Chunking deliberately removes `batchId` from the final chunk's wire metadata and restores it only after `OpSplitter.processChunk` reassembly (opSplitter.ts:174-189, 231-242), which runs only on the live inbound path. So a resubmitted chunked batch is promoted correctly live, but a later cross-session `resolve(batchId, referenceSequenceNumber)` never sees that `batchId` and returns `pending`/`unresolvable` — the exact departed-client scenario this feature exists to serve. Grouped and compressed batches preserve identity in wire metadata (opGroupingManager.ts:144-173, opCompressor.ts:52-58), so chunked is the surviving correctness hole.

Route the history scan through the shared batch-unpack pipeline so it observes the same restored `batchId` the live inbound path sees, or reconstruct chunk identity here. Add a scan-path test that exercises a chunked batch end-to-end (fetch → resolve) — no such test exists today, so the gap is currently unguarded.

packages/common/container-definitions/src/runtime.ts
```diff
@@ -399,6 +400,19 @@ export interface IContainerContext {
 		referenceSequenceNumber?: number,
 	) => number;
 	readonly submitSignalFn: (contents: unknown, targetClientId?: string) => void;
+	/**
+	 * Reads a range of sequenced ops from delta storage, the read counterpart of `submitFn`. The container
+	 * owns delta storage and injects this so the runtime can pull historical ops (e.g. to resolve a batch
+	 * identity to a sequence number). `abortSignal` cancels an in-flight fetch when the runtime stops reading
+	 * early. Optional: absent means the runtime has no historical-op read.
+	 *
+	 * @internal
```
sonalideshpandemsft: you cannot mark individual methods as internal. IContainerContext is marked legacy beta. you might have to create a new internal interface IContainerContextInternal extending IContainerContext and add fetchOps to it

packages/common/container-definitions/src/runtime.ts
```diff
@@ -399,6 +400,19 @@ export interface IContainerContext {
...
+	 */
+	readonly fetchOps?: (
```
sonalideshpandemsft: can we avoid adding fetchOps to IContainerContext?

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,238 @@
...
+		const existingSequenceNumber = this.sequenceNumberByBatchId.get(batchId);
+		if (existingSequenceNumber === sequenceNumber) {
+			return;
+		}
+
+		this.sequenceNumberByBatchId.set(batchId, sequenceNumber);
+		this.evictBelowMinimumSequenceNumber();
+		for (const listener of this.batchListeners) {
+			listener(batchId, sequenceNumber);
+		}
```
anthony-murphy, line 311: Deep Review: `onBatchSequenced` listeners run un-isolated on the inbound-op hot path — a throwing subscriber crashes the container.

`processInboundBatch` does `this.sequenceNumberByBatchId.set(...)`, `this.evictBelowMinimumSequenceNumber()`, then `for (const listener of this.batchListeners) { listener(batchId, sequenceNumber); }` with no try/catch. It runs synchronously on the core inbound path — `containerRuntime.ts` calls `versionMarkResolverInternal.processInboundBatch(...)` from `processInboundMessageOrBatch`, now after `pendingStateManager.processInboundMessages`. `onBatchSequenced` is an app-facing subscription on `IVersionMarkResolver`. A throwing listener propagates out of `processInboundBatch` → `process()`, aborting inbound op processing and taking down the container; an early throw also silently skips later listeners even though the map was already mutated. The PR's own comment notes the notification promotes a mark in an external store — an irreversible side effect.

Wrap each listener invocation in try/catch and route failures to a logger (e.g. `this.mc.logger`), mirroring how the runtime isolates handler faults elsewhere, so one misbehaving subscriber can't abort op processing or starve later subscribers. Add a test asserting a throwing listener neither aborts inbound processing nor blocks other listeners.

Distinct from the now-resolved [prior thread](https://github.com/microsoft/FluidFramework/pull/27834#discussion_r3708759991) on update-fires-before-validation ordering (the update was moved after validation) — this is a residual isolation concern in the same region.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,238 @@
...
+					// piecemeal batch's messages.
+					const update = inboundVersionMarkUpdate(inboundResult, carriedBatchId);
+					carriedBatchId = update.carriedBatchId;
+					if (update.sequenced?.batchId === batchId) {
+						return { status: "resolved", sequenceNumber: update.sequenced.sequenceNumber };
+					}
+				}
+			}
+			// Not found: distinguish "not sequenced yet" from "trimmed" via the tip and where the scan landed.
+			return this.classifyMiss(from, firstScannedSequenceNumber);
```
anthony-murphy, line 238: Deep Review: History scan asserts `0x9d5` and rejects `resolve()` when the reference point falls inside a prior multi-op batch.

`resolveFromHistory()` starts a fresh `RemoteMessageProcessor` scan at `from = referenceSequenceNumber + 1` and feeds every fetched op into the unpacker. `referenceSequenceNumber` is an arbitrary app-captured coordinate with no guarantee of being a batch boundary. If the scan begins on the final op of a multi-op batch that started before the reference point, `RemoteMessageProcessor.getResultBasedOnBatchMetadata()` sees `metadata.batch === false` while `batchInProgress` is false and asserts `0x9d5` (Unexpected batch end marker), rejecting `resolve()` instead of scanning forward. `resolve()` is contracted to return `resolved`/`pending`/`unresolvable`, not throw. Existing tests cover full multi-op batches and batches starting at the first scanned op, but not a scan that begins inside an earlier multi-op batch.

Make historical scanning tolerant of a leading partial batch — skip leading `batch !== true` fragments until the next batch boundary before using the processor, or fetch from a known earlier batch boundary rather than `referenceSequenceNumber + 1`. Add a scan-path test where the scan anchor lands mid-batch.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1879,6 +1891,52 @@ export class ContainerRuntime
...
 			},
 		);
+		// fetchOps is an internal-only capability the loader injects (IContainerContextInternal), not part
+		// of the public IContainerContext contract, so narrow to read it.
+		const fetchOps = (context as IContainerContextInternal).fetchOps;
+		this.versionMarkResolverInternal = new VersionMarkResolver({
+			getCurrentSequenceNumber: () => this.deltaManager.lastSequenceNumber,
+			getCurrentMinimumSequenceNumber: () => this.deltaManager.minimumSequenceNumber,
+			getCurrentPendingBatchId: () => this.pendingStateManager.getMostRecentPendingBatchId(),
+			// Wire the container-provided op reader (if any) so resolution can read historical ops.
```
anthony-murphy, line 1908: Deep Review: Mark capture misses same-turn edits still buffered in the Outbox, persisting a wrong "already resolved" coordinate.

The capture hook wires `getCurrentPendingBatchId` to `this.pendingStateManager.getMostRecentPendingBatchId()`, which only peeks `pendingMessages` for the most recently flushed batch. Local submits go to `this.outbox.submit(...)` then `scheduleFlush()`; in `FlushMode.TurnBased` the flush is deferred with `Promise.resolve().then(() => this.flush())`. A host capturing a mark immediately after an edit in the same JS turn sees `undefined` for the pending batchId even though the edit is pending. DEV.md tells apps to store a resolved locator when `getCurrentPendingBatchId()` is absent, so the miss becomes a durable wrong coordinate — a silent data-integrity bug on the PR's stated primary capture flow.

Ensure capture observes unflushed local work — flush before reading the pending batchId when safe, or extend the capture hook to include the Outbox's current batch identity rather than only `PendingStateManager`. Add a same-turn capture test under `FlushMode.TurnBased`.

sonalideshpandemsft: I think a single capture API would make this contract more reliable. If the desired edit is still in the outbox,  getCurrentPendingBatchId()  may currently return an older pending batch or  undefined.

Could the runtime expose an async api such as:
`captureVersionMark(): Promise<VersionMarkLocator>`

It could wait for the current outbox batch to be flushed into PendingStateManager , then return that specific batch’s  batchId and referenceSequenceNumber together

anthony-murphy:
## Deep Review

*Reviewed commit `d2b53ac` on 2026-08-05.*

Not ready for sign-off. Two contained correctness bugs remain, both flagged inline: the history scan still crashes (`resolve()` rejects) when the anchor lands inside a clipped *chunked* batch, and pending marks are permanently orphaned across reconnect when `batchIdTrackingEnabled` is false. The three bugs from the prior review are addressed — `onBatchSequenced` dispatch is now wrapped in try/catch, the same-turn capture miss is closed by the new async `captureVersionMark()` flush-then-read seam, and the non-chunk mid-batch marker is now skipped. CI build is still failing.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Fix the failing CI build so the bundle-size comparison runs and CI is green before human review
- [ ] Confirm in the PR description's Validation section that the new non-chunk mid-batch skip, listener isolation, and async-capture paths are test-covered

### Context for Reviewers

- Architecture converged: three independent blind design passes landed on this PR's shape — the app owns mark storage and the stored locator, the runtime owns only resolution, `batchId` is the identity handle, and the three-state `ResolveResult` (`resolved`/`pending`/`unresolvable`) is the durable contract.
- `fetchOps` is added on `IContainerContextInternal` (not the public/legacy `IContainerContext`) and read at construction without retaining `context`, following #16369's "don't retain IContainerContext". The "should we add `fetchOps` at all" question ([thread](https://github.com/microsoft/FluidFramework/pull/27834#discussion_r3713730368)) is a design call for the area owner.
- The `isTracking` per-inbound-batch gate mirrors #22497's "don't pay for a disabled feature"; the scan's copy-before-process and fresh `RemoteMessageProcessor` follow #13273.

- **Needs human judgment** — confirm the all-`@internal` framing (no api-report change / no changeset) is acceptable for a first-party dogfood consumer, including `fetchOps` reached via `IContainerContextInternal`. Open reviewer question: [can we avoid adding fetchOps to IContainerContext?](https://github.com/microsoft/FluidFramework/pull/27834#discussion_r3713730368) (with a related [note on not marking individual methods internal](https://github.com/microsoft/FluidFramework/pull/27834#discussion_r3713648291)).
- **Needs human judgment** — the always-on `batchId → sequenceNumber` map plus per-inbound-batch gating reverses #22497's "don't track unless enabled"; confirm the perf tradeoff was measured, not assumed, and whether gating on an active version-mark consumer is acceptable.
- **Needs human judgment** — whether the loader-injected `fetchOps` needs a #22877 supported-features/compatibility entry, or is an intentional documented best-effort exemption.
- **Needs human judgment** — explicit sign-off on shipping the driver-dependent `classifyMiss` `pending`/`unresolvable` signal as a documented interim heuristic, with a dedicated driver op-availability/retention API as the follow-up.
- **Cannot be assessed by the pipeline** — the per-inbound-batch hot-path allocation/eviction cost for containers that never use version marks is unmeasured; runtime/perf profiling on the inbound path is needed to confirm it is negligible.

</details>

- `b14776c` 2026-08-04 · **4/10** — two blockers cleared; three new contained bugs flagged
- `e157473` 2026-08-03 · **4/10** — chunked-batch scan miss and update-before-validation ordering flagged

</details>

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: Resolved. The new async `captureVersionMark()` flush-then-read seam does what was proposed here — it flushes the current Outbox batch into `PendingStateManager` before reading the pending `batchId`, so a same-turn capture no longer sees `undefined` or a stale batch. Thanks sonalideshpandemsft — the single async capture API you suggested is the shape that landed.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
anthony-murphy: Deep Review: The non-chunk case is handled now — `if (!inBatch && asBatchMetadata(op.metadata)?.batch === false) continue;` (versionMarkResolver.ts:206-215) skips the stray `batch === false` end marker that produced the `0x9d5` assert. The chunked variant of the same defect survives, though.

The scan still routes every op through a fresh `RemoteMessageProcessor` (empty `OpSplitter`) via `unpack?.(op)`, and that guard doesn't skip chunk continuations. When `referenceSequenceNumber + 1` lands on chunk 2+ of an unrelated in-flight chunked batch, `OpSplitter.processChunk`/`addChunk` fails its strict ordering check (`chunkId !== map.length + 1`) and throws `DataCorruptionError("Chunk Id mismatch")` (opSplitter.ts:90-100). Nothing wraps `unpack?.(op)` — the outer `try`'s only `finally` action is `abortController.abort()` — so the rejection bubbles out of `resolve()`, which is contracted to return `resolved`/`pending`/`unresolvable`, not throw.

Make the historical unpack tolerant of a clipped leading chunk stream — skip chunk continuations until a fresh complete chunk sequence starts (analogous to the `batch === false` skip), or wrap `unpack?.(op)` to convert the chunk-mismatch error into a scan-continue. Add a scan-path test whose anchor lands mid-chunked-batch.

anthony-murphy:
`packages/runtime/container-runtime/src/containerRuntime.ts:5306`

Deep Review: Pending version marks are permanently orphaned across reconnect when `batchIdTrackingEnabled` is false.

`reSubmitBatch` builds `resubmitInfo = { batchId: this.batchIdTrackingEnabled ? batchId : undefined, staged }` (containerRuntime.ts:5306), and `batchIdTrackingEnabled` is `!disableBatchIdTracking && this._flushMode === FlushMode.TurnBased && this.groupedBatchingEnabled` (containerRuntime.ts:2029-2032). With grouped batching disabled, `FlushMode.Immediate`, or the `DisableBatchIdTracking` kill-switch, resubmit passes `batchId: undefined` and `Outbox.flush` mints a brand-new `batchId` for the resubmitted batch.

A mark captured via `captureVersionMark()` before the reconnect holds the *original* pending `batchId` (`getMostRecentPendingBatchId()`), which never appears on any sequenced op — the live map never keys it and the history scan reconstructs the *new* identity — so `resolve()` returns `pending`/`unresolvable` forever, with no error surfaced. The kill-switch replay test at `containerRuntime.spec.ts:481-550` already shows the replay path stamps no `metadata.batchId`.

Either preserve `batchId` across resubmit unconditionally (decouple it from `batchIdTrackingEnabled`), or have `captureVersionMark()`/`resolve()` explicitly refuse or signal when tracking is disabled so the feature never hands out a mark it cannot honor. Document the coupling and add a test that captures a mark, forces a reconnect with tracking disabled, and asserts the resulting resolution behavior.

packages/common/container-definitions/src/runtime.ts
```diff
@@ -482,6 +483,26 @@ export interface IContainerContext {
...
+ * @internal
+ */
+export interface IContainerContextInternal extends IContainerContext {
+	/**
+	 * Reads a range of sequenced ops from delta storage, the read counterpart of `submitFn`. The container
+	 * owns delta storage and injects this so the runtime can pull historical ops (e.g. to resolve a batch
+	 * identity to a sequence number). `abortSignal` cancels an in-flight fetch when the runtime stops reading
+	 * early. Optional: absent means the runtime has no historical-op read.
+	 */
+	readonly fetchOps?: (
```
shlevari, line 499: Would be good to have a comment saying whether from and to are inclusive or exclusive. I feel like my default assumption is [) but that should be clearly defined

packages/loader/container-loader/src/container.ts
```diff
@@ -2443,6 +2443,17 @@ export class Container
 					this.submitBatch(batch, referenceSequenceNumber),
 				submitSignalFn: (content, targetClientId) =>
 					this.submitSignal(content, targetClientId),
+				// Read counterpart of submitFn: the container owns delta storage, so it injects a reader for
+				// the runtime to pull historical ops. Connects per call so it always uses the current service
+				// (no stale handle); cost is one connect per out-of-session mark resolution. Future: cache the
+				// handle (invalidated on reconnect/epoch-change/dispose) if a caller batch-resolves many marks.
+				fetchOps: async (from, to, abortSignal) => {
```
shlevari, line 2450: nit. Can we break this into a helper function like the rest of the items in the constructor? Helps with readability.

packages/loader/container-loader/src/containerContext.ts
```diff
@@ -72,6 +73,9 @@ export interface IContainerContextConfig
 	readonly taggedLogger: TelemetryLoggerExt;
 	// This "overrides" IContainerContext.snapshotWithContents to be required but allow `undefined`.
 	readonly snapshotWithContents: IContainerContext["snapshotWithContents"] | undefined;
+	// fetchOps is an internal-only capability (IContainerContextInternal), not part of the public
+	// IContainerContext contract. Optional: hosts may not provide op reading.
```
shlevari, line 77: the comment says optional but the interface is non optional. Is this on purpose?

sonalideshpandemsft: The config key is required, but  IContainerContextInternal["fetchOps"] includes undefined because the source property is optional

packages/runtime/container-runtime/src/test/versionMarks/versionMarkResolver.spec.ts
```diff
@@ -0,0 +1,737 @@
...
+				[makeOp({ sequenceNumber: 12, clientId: "clientA", clientSequenceNumber: 7 })],
+			]);
+			const resolver = makeResolver({ reader });
+			assert.deepEqual(await resolver.resolve(batchId, 0), {
+				kind: "resolved",
+				sequenceNumber: 12,
+			});
+		});
+
+		it("resolves a resubmitted single-op batch by its explicit batchId metadata", async () => {
```
markfields, line 270: Could we add an integration test for the complete stash → rehydrate → resubmit → resolve path? These unit tests verify that explicit original `batchId` metadata resolves correctly, but not that pending-state rehydration preserves and stamps that identity end to end. The test should capture/store a pending mark, stash the originating batch, rehydrate and resubmit it from a new client, then verify that the original mark resolves to the resubmitted batch's final sequence number.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,329 @@
...
+ * work, so the mark already points at a durable sequence number. The app packs its own stored record from
+ * this — the runtime does not define the stored locator shape.
+ *
+ * @internal
+ */
+export type VersionMarkCapture =
+	| {
+			readonly kind: "pending";
+			readonly batchId: string;
+			readonly referenceSequenceNumber: number;
```
markfields, line 55: Could we rename this to `sequenceNumberLowerBound` (and likewise rename the `resolve` parameter)? `referenceSequenceNumber` carries the usual op refSeq semantic elsewhere in the runtime, whereas this value is specifically an exclusive search anchor: historical resolution begins at `value + 1`. A more descriptive name would help prevent callers from assigning the normal refSeq meaning to it.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,329 @@
...
+
+	/** Whether inbound batches need tracking (see {@link tracking}). Read by the runtime to gate the hot path. */
+	public get isTracking(): boolean {
+		return this.tracking;
+	}
+
+	public async captureVersionMark(): Promise<VersionMarkCapture> {
+		// Seal the current batch so a just-submitted local edit is flushed into the pending state with a
+		// stable batchId (which is only assigned at flush) before we read it.
+		this.hooks.flushPendingBatch();
```
markfields, line 146: I was nervous about the flush side effect, but I do think it's okay. But it raises some interesting corner cases to explicitly handle, and/or cover in tests+docs.

- Ideally, preflight checks for reentrant/inbound processing and `orderSequentially` so capture returns a failure result without flushing or closing the container.
  - NOTE: This is an API change, it adds a new "kind" to VersionMarkCapture
- For Staging Mode, test commit and discard: commit should preserve and resolve its ID; after discard, the captured batch will never sequence, so the resulting mark behavior should be documented.

Suggested test coverage to add:
- **Real batch-cut test:** Submit an op through `ContainerRuntime`, call capture before the TurnBased flush, and verify capture flushes it into PSM and returns that exact batch’s ID—not a mocked ID.
- **Unsafe-context tests:** Call capture during inbound processing and inside `orderSequentially`; verify an explicit failure result, no flush, and no container closure.
- **Staging test:** Capture staged edits and verify nothing is sent immediately; commit preserves the captured ID through resubmit and resolution. Also define/test the result after discard.
- **One offline integration test:** Capture while disconnected, stash/rehydrate, resubmit from the new client, and resolve using the original ID. This can satisfy the separate rehydration comment too.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,329 @@
...
+	private tracking = false;
+
+	public constructor(private readonly hooks: VersionMarkResolverRuntimeHooks) {}
+
+	/** Whether inbound batches need tracking (see {@link tracking}). Read by the runtime to gate the hot path. */
+	public get isTracking(): boolean {
+		return this.tracking;
+	}
+
+	public async captureVersionMark(): Promise<VersionMarkCapture> {
```
markfields, line 143: Doesn't need to be (and thus shouldn't be) async

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,329 @@
...
+	 * Captures a version mark at the current point. Seals the current outbound batch first (so a just-made
+	 * local edit has a stable `batchId`, which is only assigned when a batch is flushed), then returns the
+	 * mark data atomically: a `pending` capture (`batchId` + `referenceSequenceNumber`) when there is an
+	 * unacked local batch, or a `resolved` capture (`sequenceNumber`) when there is no in-flight local work.
+	 *
+	 * @remarks Async because a correct capture must wait for the flush before the `batchId` exists — a
+	 * synchronous read would return an older batch or `undefined`. Sealing the batch is a side effect
+	 * (it submits the current batch), so capture at savepoint boundaries, not per keystroke.
+	 */
+	captureVersionMark(): Promise<VersionMarkCapture>;
```
markfields, line 75: Could we rename this to `sealAndCaptureVersionMark`? That makes the flush side effect less surprising

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,329 @@
...
+	onBatchSequenced(listener: (batchId: string, sequenceNumber: number) => void): () => void;
+}
+
+/**
+ * @internal
+ */
+export interface VersionMarkResolverRuntimeHooks {
+	readonly getCurrentSequenceNumber: () => number;
+	readonly getCurrentMinimumSequenceNumber: () => number;
+	readonly getCurrentPendingBatchId: () => string | undefined;
```
markfields, line 94: Consider merging this into `flushPendingBatch` so that returns the string, rather than this code having to call them in sequence

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,329 @@
...
+		} finally {
+			// Cancel any in-flight fetch once we stop reading (found or exhausted).
+			abortController.abort();
+		}
+	}
+
+	/**
+	 * Classifies a history-scan miss as `pending` or `unresolvable`. See {@link resolveFromHistory}.
+	 */
+	private classifyMiss(
```
markfields, line 248: This code feels over-fit to driver behaviors.  Is it that important to distinguish 'pending' from 'unresolvable'?  Relative to how confident we are that this code gets it right - I believe it is making some educated guesses in some cases?

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,329 @@
...
+		}
+		if (firstScannedSequenceNumber > from) {
+			// A trim gap at the anchor: the mark's ops (just after the reference point) are gone.
+			return { kind: "unresolvable" };
+		}
+		// Ops are present from the reference point and the batch is not among them: not yet sequenced.
+		return { kind: "pending" };
+	}
+
+	public onBatchSequenced(
```
markfields, line 277: Future: before promoting this API to beta, we should try extending the existing `batchEnd` event to expose the effective stable batch ID and reuse it here, rather than finalizing a parallel batch-sequenced notification API.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,329 @@
...
+		}
+		return this.resolveFromHistory(reader, batchId, referenceSequenceNumber);
+	}
+
+	/**
+	 * Scans sequenced ops from `referenceSequenceNumber` forward for `batchId`, returning its last op's
+	 * sequence number when found and aborting the fetch once found or exhausted. Each op is routed through
+	 * the same unpack pipeline the live inbound path uses (chunk reassembly, ungroup, decompress) and
+	 * {@link inboundVersionMarkUpdate} derives its batch identity, so chunked batches resolve here too. The
+	 * scan anchor is not guaranteed to be a batch boundary, so a leading batch clipped by the anchor is
```
markfields, line 182: This is true for chunk streams, not ordinary multi-op batches: `InboundBatchAggregator` keeps normal batches atomic, but DeltaManager advances `lastSequenceNumber` for each intermediate chunk before reconstruction completes. The current orphan batch-end guard does not handle a scan starting mid-chunk stream; please either anchor on the last fully reconstructed runtime batch or make the history scan tolerate missing leading chunks, and add coverage for that case.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	/**
+	 * Bounds the map to the collaboration window by dropping entries below the MSN (mirrors
+	 * `DuplicateBatchDetector`). Safe because below the MSN every client has processed the batch, so its
+	 * `onBatchSequenced` already fired; a dropped entry falls back to the history scan. Entries insert in
+	 * sequence order, so iteration stops at the first retained one.
+	 */
+	private evictBelowMinimumSequenceNumber(): void {
+		const minimumSequenceNumber = this.hooks.getCurrentMinimumSequenceNumber();
```
markfields, line 321: Using MSN doesn't feel intuitive to me, but I think I may be missing something about how this will be used and where the markers will come from.  Can you briefly explain?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1541,11 +1547,17 @@ export class ContainerRuntime
 	private readonly blobManager: BlobManager;
 	private readonly pendingStateManager: PendingStateManager;
 	private readonly duplicateBatchDetector: DuplicateBatchDetector | undefined;
+	/**
+	 * Host-facing version mark resolver.
+	 */
+	public readonly versionMarkResolver: IVersionMarkResolver;
```
markfields, line 1553: Could this be a getter that returns `versionMarkResolverInternal` as `IVersionMarkResolver`? That avoids storing and assigning the same object twice while preserving the narrower host-facing type.

packages/common/container-definitions/src/runtime.ts
```diff
@@ -482,6 +483,26 @@ export interface IContainerContext {
 	readonly snapshotWithContents?: ISnapshot;
 }
 
+/**
+ * Internal extension of {@link IContainerContext} for capabilities the loader injects into the runtime
+ * that are not part of the public container/runtime contract.
+ *
+ * @internal
+ */
+export interface IContainerContextInternal extends IContainerContext {
```
markfields, line 492: We should run this by Navin to get his input on best practice for cross-layer features like this.  Maybe there is something explicit we should/could do for layer-compat rather that would be beneficial.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1879,6 +1891,59 @@ export class ContainerRuntime
...
+				namespace: "VersionMarkResolver",
+			}),
+			// Seal the current outbound batch so a just-submitted edit gets a stable batchId in the pending
+			// state before captureVersionMark reads it (a batchId is only assigned when a batch is flushed).
+			flushPendingBatch: () => this.flush(),
+			// Wire the container-provided op reader (if any) so resolution can read historical ops.
+			getHistoricalOpReader: fetchOps ? () => ({ fetchMessages: fetchOps }) : undefined,
+			// Unpack scanned historical ops through the same pipeline the live inbound path uses, so a
+			// chunked batch's batchId (only restored after reassembly) is observed by the history scan too.
+			createHistoricalOpUnpacker: fetchOps
```
markfields, line 1912: Future: consider preserving the effective `batchId` in the final chunk's wire metadata. That would let historical resolution identify batches directly from metadata instead of constructing a second stateful un-virtualization pipeline, and would avoid the clipped-leading-chunk problem.

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -0,0 +1,125 @@
...
+
+The reader is wired through the container→runtime seam, mirroring `submitFn` (the write direction): the loader injects an optional `fetchOps(from, to?, abortSignal?) => Promise<IStream<ISequencedDocumentMessage[]>>` via **`IContainerContextInternal`** — an `@internal` interface extending `IContainerContext`, so `fetchOps` stays off the public/legacy `IContainerContext` contract. The `Container` implements it via its own delta storage (`service.connectToDeltaStorage().fetchMessages`, `container.ts`); and `ContainerRuntime` narrows the context to `IContainerContextInternal` to read it and passes it into the resolver as both `getHistoricalOpReader` and (wrapped in a `RemoteMessageProcessor`) `createHistoricalOpUnpacker`. It stays generic — `fetchOps` is backed by the driver-agnostic `IDocumentDeltaStorageService`, so the runtime never imports a driver. When a host provides no `fetchOps`, the hooks are absent and unknown batchIds return `pending`. The scan threads an `AbortSignal` and aborts the in-flight fetch once the batch is found (or the range is exhausted).
+
+### `pending` vs `unresolvable` on a miss (read-derived availability)
```
markfields, line 70: Could we define "trimmed" here? It means older sequenced ops are no longer retained or returned by the service's delta storage, not eviction from the resolver cache or advancement past MSN.

markfields (approved):
Lots to discuss in parallel to merging this, but overall the direction is good.

PR #27843, opened 2026-08-04
feat(tree): Promote `FormattedTextAsTree` to alpha

Josmithr:
See the changeset for more details.

[AB#78598](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/78598)


packages/dds/tree/src/entrypoints/alpha.ts
```diff
@@ -330,7 +332,6 @@ export {
 	replaceConciseTreeHandles, 
 	replaceHandles, 
 	replaceVerboseTreeHandles, 
-	trackDirtyNodes, 
-	defineTreeDataStore
```
Josmithr: TODO: this seems unrelated. Investigate.

Josmithr: This file was stale on main. Fix here: https://github.com/microsoft/FluidFramework/pull/27844

.changeset/bold-letters-dance.md
```diff
@@ -0,0 +1,53 @@
+---
+"@fluidframework/tree": minor
+"__section": tree
+---
+Promote FormattedTextAsTree APIs to alpha
```
Josmithr: **Question for reviewers**

"FormattedTextAsTree" is quite verbose. The plain-text counterpart ("TextAsTree") is similarly verbose. This might be a good time to chat about renaming (I can do this in a separate PR first to avoid user-facing churn for the formatted text API).

Potential alternatives off the top of my head:
- "FormattedText" (and "PlainText"?)
    - This is perhaps too simplified?
- "FormattedTextSchema" (and "PlainTextSchema")
    - An existing beta API with comparable scope is called "TableSchema". I don't have a particularly strong preference for this naming, but it would at least cause our names to consistent.

I'm open to alternatives :)

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -115,14 +115,17 @@ export class StringTextAtomNode
 
 /**
  * A collection of text related types, schema and utilities for working with text beyond the basic {@link SchemaStatics.string}.
+ *
  * @remarks
  * This is generic over formatting an embedded object/atom types.
- * See {@link FormattedTextAsTreeDefault} for a default parameterization.
+ *
  * @privateRemarks
+ * See {@link FormattedTextAsTreeDefault} for an example parameterization.
```
Josmithr: Note for reviewers: This links to an internal API, so this needed to be moved into a private remarks block. Also updated comment, since we don't plan on stabilizing that API - it exists as an example.

.changeset/bold-letters-dance.md
CraigMacomber: This pattern aligns with JsonAsTree, which I think makes sense, and was done because naming a bunch of stuff just JSON was way to general/ambiguous.

I prefer to put the relevant thing first for autocomplete/discoverability/sorting reasons (hence asTree being a suffix)

I think the same argument applies to Text. TextAsTree seems reasonable as there is lot of text stuff, but that is special as it puts it in a tree.

For FormattedTextAsTree, I'd be tempted to just go with `FormattedText` or `TextFormatted`.

`TextFormatted` and `TextPlain` (and `TextFormattedDefault`) might be decent names.

Its worth noting however that FormattedTextAsTree uses some things from TextAsTree, so adding "plain" to it might be a little bit odd (not too bad) unless we refactor it a bit.

For Tables we went with TableSchema. Unlike strings/test and JSON, tables are less of a built in TypeScript thing (though are an HTML thing...) so simply `Table` might have been ok.

As these namespaces contain interfaces which nodes implement (like FormattedTextAsTree.members), naming it "Schema" seems maybe not ideal.

Anyway, I agree we should probably do a rename, and I think all the various Ideas and motives I have are noted above.

I'd be fine with PlainText and FormattedText, or TextPlain and TextFormatted. TextSchema and FormattedTextSchema could also work, but I prefer the other options.

.changeset/bold-letters-dance.md
```diff
@@ -0,0 +1,53 @@
+---
+"@fluidframework/tree": minor
```
CraigMacomber, line 2: you forgot fluid-framework, which is the package we actually want people to use this from.

Josmithr: Yup. Good catch.

Josmithr:
```suggestion
"@fluidframework/tree": minor
"fluid-framework": minor
```

CraigMacomber (approved):

.changeset/bold-letters-dance.md
```diff
@@ -0,0 +1,54 @@
...
+Promote FormattedTextAsTree APIs to alpha
+
+The [`FormattedTextAsTree`](https://fluidframework.com/docs/api/fluid-framework/formattedtextastree-namespace) namespace is now available from the `@fluidframework/tree/alpha` entrypoint.
+It provides a generic, collaborative rich-text domain built on SharedTree, parameterized by the formatting you want to associate with each unit of text and by any extra "atom" (embedded object) types you want to allow alongside plain characters.
+
+Use `FormattedTextAsTree.createSchema` to generate a text schema for your chosen formatting, then treat the resulting node like a formatted string.
+
+```typescript
+import { SchemaFactory, SchemaFactoryBeta } from "@fluidframework/tree";
+import { FormattedTextAsTree } from "@fluidframework/tree/alpha";
```
Josmithr:
```suggestion
import { SchemaFactory} from "fluid-framework";
import { SchemaFactoryBeta } from "fluid-framework/beta";
import { FormattedTextAsTree } from "fluid-framework/alpha";
```

CraigMacomber (approved):

.changeset/bold-letters-dance.md
Josmithr: I think I have a preference for "PlainText" and "FormattedText" over "TextPlain" and "TextFormatted", only because they flow like you would expect the English descriptions to. And intellisense/autocomplete should handle either just fine.

@brrichards @daesunp what do you guys think?

Josmithr: > For Tables we went with TableSchema. Unlike strings/test and JSON, tables are less of a built in TypeScript thing (though are an HTML thing...) so simply `Table` might have been ok.

Yeah, I think I agree. And if we're not including the "Schema" postfix in the other domains like this, I think I might pursue that rename (in a staged fashion).

Josmithr: I went ahead and created a PR that applies my preferred renames: https://github.com/microsoft/FluidFramework/pull/27853

Still open to alternatives if folks feel strongly.

.changeset/bold-letters-dance.md
```diff
@@ -0,0 +1,55 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+Promote FormattedText APIs to alpha
+
+The [`FormattedText`](https://fluidframework.com/docs/api/fluid-framework/formattedtext-namespace) namespace is now available from the `@fluidframework/tree/alpha` entrypoint.
```
Josmithr:
```suggestion
The [`FormattedText`](https://fluidframework.com/docs/api/fluid-framework/formattedtext-namespace) namespace is now available from the `fluid-framework/alpha` entrypoint.
```

packages/dds/tree/src/text/textDomainFormatted.ts
```diff
@@ -139,9 +142,10 @@ export namespace FormattedText {
 	 * @param defaultFormatInsertable - The formatting applied to text inserted via non-formatted APIs
 	 * (for example {@link FormattedText.Members.insertAt} and {@link FormattedText.Statics.fromString} when no explicit format is provided).
 	 * @returns The schema for the formatted text node, whose nodes implement {@link FormattedText.Members} and whose statics implement {@link FormattedText.Statics}.
-	 * @remarks
-	 * See {@link FormattedTextDefault} for a default parameterization of this factory.
+	 *
 	 * @privateRemarks
+	 * See {@link FormattedTextDefault} for a default parameterization of this factory.
```
Josmithr:
```suggestion
	 * See {@link FormattedTextDefault} for an example parameterization of this factory.
```

CraigMacomber (approved):

PR #27844, opened 2026-08-04
build(tree): Update out-of-date auto-generated alpha entrypoint file

Josmithr:
Contents were out of date on main. Regenerated via `npm run generate:entrypoint-sources`.


Josmithr:
@jason-ha it looks like we don't currently validate that these ts entrypoints remain up-to-date. Seems like we should be running this set as a part of the build in this package.

CraigMacomber (approved):

PR #27849, opened 2026-08-04
build(client): remove Node10 module resolution compatibility

### LINKED ISSUE #27457: Removal of Node10 resolutions in v3.0
## Broken API or pattern

Compile settings with Node10 resolution requirements (not Node16, NodeNext, or bunder) will no longer be supported.

**Preparation References**

Update compiler settings to use `"module": "Node16" | "NodeNext"`. When using a bundler, TypeScript compilerOptions with `"module": "esnext"` and `"moduleResolution": "bundler"` remain supported.

**Packages**

All client packages

**Expected Timeline**

v3.0

WillieHabi:
## Description

Removes TypeScript Node10 (`moduleResolution: "node"` / `"node10"`) compatibility from Fluid Framework Client packages for the 3.0 release.

Client packages no longer generate or publish root declaration shims for legacy resolution. Package validation now requires modern Node16, NodeNext, or Bundler resolution while preserving current ESM and CommonJS support.

Note: This also adds explicit NodeNext cross-package compilation coverage and keeps Tree's aggregate entrypoint task producing its modern ESM and CommonJS entrypoints after the Node10 task is removed.

Closes #27457. 

## Breaking Changes

Applications using TypeScript's Node10 resolver must update to one of these supported configurations:

- `"module": "Node16"` with `"moduleResolution": "Node16"`
- `"module": "NodeNext"` with `"moduleResolution": "NodeNext"`
- `"module": "ESNext"` with `"moduleResolution": "Bundler"`

Existing package entrypoints such as `/alpha`, `/beta`, `/legacy`, and `/internal` remain available under supported module resolution modes.

See [Breaking vs. Non-breaking Changes](https://github.com/microsoft/FluidFramework/wiki/Breaking-vs-Non-breaking-Changes).

## Reviewer Guidance

This PR is staged for Client 3.0 and should remain a draft until the major-break window opens.


examples/utils/import-testing/src/test/crossPackage/nodeNextImporter.ts
```diff
@@ -0,0 +1,29 @@
...
+/*
+Compile-time test for cross-package schema consumption where schemaDefinitions.d.ts
+was generated under NodeNext module resolution.
+
+Each tree export tier has its own JS entrypoint, so TypeScript correctly
+resolves import paths in .d.ts files. These imports should compile without error.
+*/
+import { TreeViewConfiguration } from "@fluidframework/tree";
+
+/* eslint-disable import-x/no-internal-modules */
```
jason-ha: It is pretty obvious. However, these should be commented to avoid the linter warnings.

packages/dds/claims/.gitignore
```diff
@@ -1,6 +1,3 @@
 # Compiled TypeScript and CSS
 dist
 lib
-
-# specialized Node10 support
-!internal.d.ts
```
jason-ha, line 6: no corresponding file to be deleted?

packages/dds/tree/package.json
```diff
@@ -221,6 +220,13 @@
 	},
 	"fluidBuild": {
 		"tasks": {
+			"build:entrypoints": {
+				"dependsOn": [
+					"build:esnext",
+					"tsc"
+				],
+				"script": false
+			},
```
jason-ha: This doesn't add up. Why is it needed when not here before? `script false` seems wrong since it is a defined script entry.
Perhaps we can just remove `build:entrypoints` from tree?
If we were to have rules like this it would probably be in global config - script false on `build:entrypoints*` task would probably give you what you are looking for. (But shouldn't be needed at this time.)

.gitignore
```diff
@@ -16,12 +16,6 @@ nyc
 .vale/config/vocabularies/*
 !.vale/config/vocabularies/fluid/
 
-# Generated Node10 module resolution compatibility files
-alpha.d.ts
-beta.d.ts
-internal.d.ts
-legacy.d.ts
```
jason-ha, line 23: What is the plan with `clean` scripts?
They all still reference these files.
Note that I am not suggesting that the clean scripts be updated in this PR. But I do want to know the plan.

packages/dds/claims/.gitignore
WillieHabi: The file was deleted in this change #27569 the unignore was left behind.

.gitignore
WillieHabi: The idea here was to keep the clean scripts thorugh 3.0 "transition" so `pnpm clean` removes existing stale previously generated  .d.ts files. Think they are harmless to keep for now (?)

Maybe followup removal of clean scripts after 3.0 ships and they are no longer generated.

examples/utils/import-testing/src/test/crossPackage/nodeNextImporter.ts
WillieHabi: added comment to suppression

packages/dds/tree/package.json
WillieHabi: Right yes, the override here was trying to preserve the `build:entrypoints` wrapper but realizing there's actually no need...better to just to remove build:entrypoints altogether from tree.

WillieHabi: updated

WillieHabi:
Superseded by central Client 3.0 staging PR #27970.

PR #27852, opened 2026-08-04
test(tree): add tree change event ordering tests

jenn-le:
## Description

Adds tests that assert the ordering guarantees of SharedTree's change events (`nodeChanged`, `treeChanged`, `rootChanged`) and their interaction with `withBufferedTreeEvents`. These tests document and lock in the following behaviors:

- `nodeChanged` fires before `treeChanged` on the same node
- Events propagate bottom-up through the tree hierarchy
- `rootChanged` fires after `nodeChanged`/`treeChanged` (via `afterBatch`)
- `withBufferedTreeEvents` inverts `rootChanged` relative to buffered node events
- Node events respect ordering within buffered flushes

Also updates a TODO comment in `treeChangeEvents.ts` to document the confirmed bottom-up ordering behavior.

## Reviewer Guidance

[Reviewer Guidance Wiki](https://dev.azure.com/fluidframework/internal/_wiki/wikis/FF%20Internal%20Wiki/1217/Reviewer-guidance)

The test file is self-contained. The only production code change is replacing a TODO with documentation (line 69 of `treeChangeEvents.ts`).

### DIFF
--- a/packages/dds/tree/src/test/simple-tree/api/treeChangeEventOrdering.spec.ts
+++ b/packages/dds/tree/src/test/simple-tree/api/treeChangeEventOrdering.spec.ts
@@ -186,15 +186,13 @@ describe("Tree change event ordering", () => {
 			subscribeToNodeEvents(root.child, log, "child");
 			subscribeToNodeEvents(root, log, "parent");
 
-			// Edit parent FIRST, then child — events follow edit order, not tree depth
 			withBufferedTreeEvents(() => {
 				root.ownProp = "world";
 				root.child.value = 42;
 			});
 
 			// Events fire in the order nodes were first edited (parent before child),
-			// NOT bottom-up by tree depth. The "bottom-up" behavior observed in other tests
-			// is a consequence of editing the child first.
+            // but nodeChanged still comes before treeChanged
 			assert.deepEqual(log, [
 				"parent:nodeChanged",
 				"parent:treeChanged",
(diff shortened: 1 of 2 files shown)


Josmithr:
@jenn-le Looks like this PR has some extraneous changes FYI

packages/dds/tree/src/test/simple-tree/api/treeChangeEventOrdering.spec.ts
```diff
@@ -0,0 +1,477 @@
...
+
+			const log: string[] = [];
+
+			subscribeToNodeEvents(root.child, log, "child");
+			subscribeToNodeEvents(root, log, "parent");
+
+			// Both parent and child change in the same buffered batch
+			withBufferedTreeEvents(() => {
+				root.child.value = 42;
+				root.ownProp = "world";
```
daesunp, line 156: I tried running this test with the edit order switched, and the test failed, so we may have to find a different way to test this (which I don't know either). I chatted with Craig about this a while ago, so I think he may have some good ideas on how to test this

daesunp: Actually I made a mistake, it still orders nodeChanged before treeChanged, so this test is good :)

jenn-le: I added another test and clarified some docs showing that the event order is based on edit order but nodeChanged comes before treeChanged for each node

daesunp (approved):

daesunp (approved):

daesunp (approved):

PR #27853, opened 2026-08-04
refactor(tree): Rename internal and alpha text APIs

Josmithr:
Applies the following API renames:
- `TextAsTree` (alpha) -> `PlainText`
- `FormattedTextAsTree` (internal) -> `FormattedText`
- `FormattedTextAsTreeDefault` (internal) -> `FormattedTextDefault`

This PR also removes the invalid re-exports of `FormattedTextAsTree` in the `quill-react` package. Note that while these re-exports appeared in our API reports, they *did not* appear in any of the generated entrypoints, so they could not actually be used by consumers.


CraigMacomber (approved):

packages/framework/quill-react/api-report/quill-react.public.api.md
```diff
@@ -4,8 +4,6 @@
 
 ```ts
 
-export { FormattedTextAsTree }
```
CraigMacomber, line 7: How did we allow an internal API to be reexported as public? Seems like an API extractor config bug which we should fix? Or maybe a namespace handling bug?

Did this show up in our generated entry point types and code as well, or is it just in the API report?

CraigMacomber: Resolving the cause of this shouldn't block the PR, I just want to ensure we follow-up on this.

For this PR though, it would be good to clarify if this is a real app facing public export which could have been used and worked, which we are removing, or something broken that would have never worked at runtime.

Josmithr: > How did we allow an internal API to be reexported as public? Seems like an API extractor config bug which we should fix? Or maybe a namespace handling bug?

It is a config bug. Any time we re-export contents from another package, we need to be sure to configure the model/report configurations to bundle those packages in the analysis. Otherwise, api-extractor treats external packages as foreign and doesn't attempt to reason about their release tags (which may have different semantics), implicitly treating everything foreign as _public_.

I _thought_ I had already updated our base configs to do this automatically for packages in our various namespaces, but either that work was incomplete or was reverted. I'll investigate.

> Did this show up in our generated entry point types and code as well, or is it just in the API report?

I'm investigating. I suspect the roll-ups are fine, since that logic doesn't use api-extractor. But I will confirm.

Josmithr: Confirmed that rollups were unaffected. This is just an issue with how the reports are configured, which I'm still investigating.

Josmithr: Okay, the reports issue is now understood. It looks like the base api-extractor "report" config that ensures cross-package re-exports are handled correctly is no longer used by most packages when generating their reports. I'm not sure when this happened, but it explains what's happening here. We'll need to fix this.

Abe27342 (approved):
api files looks consistent with description and reasonable renames. Didn't review the implementation files.

CraigMacomber (approved):

PR #27855, opened 2026-08-05
fix: order ESLint after generated test outputs

CraigMacomber:
## Description

Adds explicit Fluid build dependencies for ESLint tasks that resolve generated test-only package exports.

Without these edges, a clean parallel build can run ESLint before a producer's test declarations exist. This causes nondeterministic `import-x/no-unresolved` errors and follow-on unsafe-type errors. The affected relationships are:

- Sequence to Merge Tree test exports
- Test DDS Utils to Stochastic Test Utils test exports
- Tree to ID Compressor test utilities
- Presence Runtime to its self-imported test utilities
- Local Server Stress Tests to its DDS test exports

Each failure was reproduced by cleaning the producer output and running the consumer's ESLint task through Fluid build. After adding the dependencies, Fluid build schedules the required test compilation before ESLint and all targeted checks pass.

## Reviewer Guidance

Please verify that the explicit task dependencies are preferable to broadening the repository-wide ESLint dependency graph.


packages/test/local-server-stress-tests/package.json
```diff
@@ -124,6 +127,20 @@
...
+				"@fluidframework/sequence#build:test:esm",
+				"@fluidframework/map#build:test:esm",
+				"@fluidframework/matrix#build:test:esm",
+				"@fluidframework/tree#build:test:esm",
+				"@fluidframework/task-manager#build:test:esm",
+				"@fluidframework/legacy-dds#build:test:esm",
+				"@fluidframework/ordered-collection#build:test:esm",
+				"@fluidframework/counter#build:test:esm",
+				"@fluidframework/register-collection#build:test:esm",
+				"@fluid-internal/claims#build:test:esm"
```
Josmithr: This seems brittle to me. Should we instead just depend on *all* dependencies' `build:test:esm` tasks for this package, since it has so many test dependencies?

CraigMacomber: I have cleaned up the duplication in what I think is the best way.

Josmithr: The explicit list of dependencies still seems brittle to me for this package; it seems likely that additional test build dependencies could be added that would easily break this. In my opinion, it would be safer to just eat the perf cost in this package and wait for all test builds to complete before linting.

Josmithr (approved):
Left some more thoughts in the existing comment thread, but I'm not considering them blocking.

packages/test/local-server-stress-tests/package.json
CraigMacomber: Ah, you mean `^build:test:esm`. I read that wrong and was thinking you meant `build:test:esm` (or its dependencies, which we don't have a clean way to express other than that)

CraigMacomber: I'll leave it for now: this package is already pretty near the end of the build dependency wise, and that would significantly reduce what can build in parallel with it (force it after the tests for a lot of random stuff). Might be worth doing if we have issues in the future, but I'd have to measure build parallelization impact.

PR #27856, opened 2026-08-05
build(client): bump build-tools to 0.66.0

MarioJGMsoft:
## Description

Bumps the build-tools catalog entries in pnpm-workspace.yaml from ^0.65.0 to ^0.66.0 and updates the lockfile accordingly.

Ran `pnpm dedupe` that also cleans up the root lockfile of the repo.

### Note on the `fflate` pin in `pnpm-lock.yaml`

`@fluid-tools/build-cli` 0.66.0 raised its `fflate` dependency from `^0.8.2` to `^0.8.3`, so this bump introduces `fflate@0.8.3` into the tree. On its own that is harmless — attw stayed on its own `0.8.2`. But a `pnpm dedupe` collapses the two onto `0.8.3`, because attw's `^0.8.2` range also accepts it, and that breaks the `are-the-types-wrong` CI check for every package:

    Error: Cannot read properties of undefined (reading 'filename')

**Root cause (attw's bug, exposed by fflate).** `@arethetypeswrong/core` decompresses the packed tarball with fflate's streaming API and keeps only the last emitted chunk instead of accumulating:

```js
let unzipped;
new Gunzip((chunk) => (unzipped = chunk)).push(tarball, /* final */ true);
const data = untar(unzipped);
const prefix = data[0].filename; // ← throws
```

This has already been solved in newer versions of the `attw` package. Will open a follow up task to bump that version but will leave the pinned version for now.
## Reviewer Guidance

There were some dependencies removed from the lockfile, my guess is that these were related to the bundle-size tooling, but I'd like confirmation from reviewers if this is expected or not.

Fixes: [AB#77295](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/77295)


steffenloesch (approved):

examples/utils/import-testing/src/test/build.spec.ts
```diff
@@ -13,7 +13,17 @@ import { promisify } from "node:util";
...
+/**
+ * Minimal shape of the parts of package.json this test reads.
+ *
+ * @remarks
+ * Declared locally rather than imported from `@fluidframework/build-tools` because that package is
+ * ESM-only and this test is also compiled as CommonJS, where its type surface is unavailable.
+ */
+interface PackageJsonDevDependencies {
+	readonly devDependencies: Readonly<Record<string, string>>;
+}
```
jason-ha, line 32: Can this be restored adding `customConditions: ['module-sync']` to tsconfig?
Otherwise, there should be a TODO to restore the original import.

jason-ha: It isn't when this stops building CommonJS, but close enough.
(When we do the CJS removal for libraries, we'll still build CJS examples. But those will be able to import ESM. If module-sync isn't enough then we need new version of TypeScript, which is also coming.)

jason-ha (approved):

examples/utils/import-testing/src/test/build.spec.ts
MarioJGMsoft: It can't be restored because the new build-tools version doesn't declare a package that can be used in CJS. 
I created AB#80348 to track the cleanup.

jason-ha: right - you also need to say Node20 or NodeNext

pnpm-lock.yaml
```diff
@@ -40069,6 +39686,19 @@ snapshots:
...
+      cwd: 0.10.0
+      find-process: 1.4.7(supports-color@10.2.2)
+      prompts: 2.4.2
+      spawnd: 10.1.4
+      tree-kill: 1.2.2
+      wait-on: 8.0.1(supports-color@10.2.2)
+    transitivePeerDependencies:
+      - debug
+      - supports-color
+
```
jason-ha, line 39429: These new jest deps come from build-tools bump? Please double check

MarioJGMsoft: Redid it with a clean `pnpm i` and got the same lockfile, so it's not merge churn. I think it's from `bundle-size-tools` being dropped in 0.66.0 — that removes `danger` → `debug@4.4.3`, which is an optional peer of `jest-dev-server`/`wait-on`, so those entries get re-keyed without the `(debug@4.4.3)` suffix.

jason-ha: I think I looked the wrong direction when looking to see if we had the version but with different qualifiers.
I think build-tools may be too big if it can impact jest deps. Oh well.

BTW did you run `pnpm dedupe`?

jason-ha (approved):

pnpm-lock.yaml
MarioJGMsoft: I hadn't run `pnpm dedupe`, but that also brought up something interesting, which is that current main has some changes when you run`pnpm dedupe` 😮. This PR should take care of that because I ran it in `fix: ran pnpm dedupe`. https://github.com/microsoft/FluidFramework/pull/27856/commits/919a63c8836e029852b5f3ed7fe7c31194c35a95

jason-ha (approved):

jason-ha (commented):
Ideally there would be comments in the repo to note this issue. But at the very least note the issue in the description. File attw issue if there isn't one.

jason-ha (approved):

jason-ha (approved):

PR #27857, opened 2026-08-05
feat: stop implementing TypeScript built-in container types

### LINKED ISSUE #27459: Stop implementing TypeScript built in types like ReadonlyArray and ReadonlyMap in v3.0
## Broken API or pattern

Currently we explicitly declare that some of our types implement built in ReadonlyArray, ReadonlyMap (and some related types like their iterators).
This results in our code failing to type check for users using typescript configured such that these types require members we do not implement.
We will instead migrate our implementations to refer to our own equivalent types which we can ensure they implement: this will not change what behavior is actually implemented in any way.
Since these types are structural and not nominal, our implementations (Like ArrayNode, MapNode, and SharedMap) will still be assignable to the built in types in cases where libCheck did not report any errors.
Only cases where libCheck produced compile errors should be impacted, and these will either, depending on usage, be fixed with no changed to the user code OR require them to use the types from Fluid Framework.

**Preparation References**

Migration guidance:
No runtime impact. If type errors are hit, adjust usage to use the new types as needed to fix them.

**Packages**

Mainly tree and shared map.

**Expected Timeline**

v3.0

jenn-le:

## Description

Resolves https://github.com/microsoft/FluidFramework/issues/27459.

Fluid's public map and array APIs currently declare conformance to TypeScript built-in container and iterator types. Changes to a consumer's configured TypeScript library can therefore add requirements that Fluid's implementations do not satisfy.

This change migrates the affected Tree, SharedMap, Presence, and container-loader APIs to Fluid-owned `FluidReadonlyArray`, `FluidReadonlyMap`, `FluidMap`, and `FluidIterableIterator` contracts. It also removes the now-redundant `IDirectoryBeta` interface. Runtime behavior is unchanged.

BubbleBench's shared model contracts are migrated to Fluid-owned iterable types and a minimal read-only collection shape. This keeps both Tree arrays and native arrays compatible and verifies how applications can model collection APIs without depending on TypeScript's evolving built-in container declarations.

## Breaking Changes

This changes public TypeScript declarations and is intended for the Fluid Framework 3.0 release. See [Breaking-vs-Non-breaking-Changes](https://github.com/microsoft/FluidFramework/wiki/Breaking-vs-Non-breaking-Changes) for details.

Most existing assignments remain structurally compatible. Code that calls methods available only on newer built-in `Array` or `Map` types should first copy the Fluid container into a native container, for example:

```typescript
const reversed = [...arrayNode].toReversed();
```

Shared application interfaces that accept both Fluid containers and native containers should use `FluidIterable<T>` or declare only the collection members they require instead of requiring a built-in container type. Consumers using `IDirectoryBeta` should migrate to `IDirectory`, which now provides the Fluid-owned map contract.

## Reviewer Guidance

Please focus on completeness of the public container/iterator migration and the generated API report changes. This public breaking change also requires API Council approval.

All five BubbleBench packages compile with the updated shared collection contracts.


jenn-le:
Replaced by #27908, which uses the upstream `test/client/3.0/stop-built-in-type-implementations` branch.

PR #27858, opened 2026-08-05
Fix(test): fix for idCompressor flaky summarization test

brrichards:
## Description

Fixes the flaky test `r11s-frs_0 | IdCompressor Summaries Non-Compat | [r11s-frs] Includes ack'd ids in summary`. This seemed to be a test bug and not a production bug. 

These are summarizer tests, but the suite predates [WritingTestsThatTakeSummaries.md](https://github.com/microsoft/FluidFramework/blob/main/packages/test/test-end-to-end-tests/WritingTestsThatTakeSummaries.md#the-rules-that-keep-these-tests-deterministic) that has useful info for flaky tests . This PR updates the suite to follow the doc.

The fix for the reported flake is `syncSummarizer: true`. The doc describes this exact failure mode:

- Without it, `ensureSynchronized` does not wait for the summarizer, so a subsequent `summarizeNow` may run before the summarizer has processed your latest ops.

The remaining rules the suite didn't meet are fixed for compliance while touching it

The full list of rules and if they were updated in this PR: 
1.  Use syncSummarizer: true - updated
2. Disable automatic summaries on every interactive container - updated
3. Call ensureSynchronized() before every summary - met
4. Use summaryVersion to chain loads - updated
5. Close one summarizer before starting another - met

This bug can be replicated locally by artifically delaying the summarizers inbound op delivery time. Which is exactly what [Rule #1](https://github.com/microsoft/FluidFramework/blob/main/packages/test/test-end-to-end-tests/WritingTestsThatTakeSummaries.md#1-use-syncsummarizer-true) solves. 

## Reviewer Guidance


packages/test/test-end-to-end-tests/src/test/idCompressor.spec.ts
```diff
@@ -805,9 +806,22 @@ describeCompat("IdCompressor Summaries", "NoCompat", (getTestObjectProvider, com
 		dds: { SharedDirectory },
 	} = compatAPIs;
 	const disableConfig: ITestContainerConfig = {
-		runtimeOptions: { enableRuntimeIdCompressor: undefined },
+		runtimeOptions: {
+			enableRuntimeIdCompressor: undefined,
+			summaryOptions: { summaryConfigOverrides: { state: "disabled" } },
```
taylorsw04, line 811: What is this line added for? I get the enabling/disabling (what was there previously), but what's this for?

taylorsw04 (approved):

packages/test/test-end-to-end-tests/src/test/idCompressor.spec.ts
brrichards: This line is from [Rule #2](https://github.com/microsoft/FluidFramework/blob/main/packages/test/test-end-to-end-tests/WritingTestsThatTakeSummaries.md#2-disable-automatic-summaries-on-every-interactive-container) and is for disabling auto summaries of containers so that summaries only come from `summarizeNow`. This one wasn't causing issues. Maybe there was one flaky fail from this scenario, but for the most part this line is just to update the tests to what the .md said about summarization test. The summarizer configs below this dont have this line so summarizeNow still works as intended

PR #27864, opened 2026-08-05
Reduce redundant logger plumbing in tree

CraigMacomber:
## Description

Leverage the logger in the breaker in more places in tree, reducing getting loggers in multiple ways and potential issues of using the wrong one.

## Reviewer Guidance


packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -547,9 +545,7 @@ export class TreeCheckout implements ITreeCheckout {
 		private readonly revisionTagCodec: RevisionTagCodec,
 		private readonly idCompressor: IIdCompressor,
 		private readonly _removedRoots: DetachedFieldIndex = makeDetachedFieldIndex("repair"),
-		/** Optional logger for telemetry. */
-		private readonly logger?: TelemetryLoggerExt,
-		public readonly breaker: Breakable = new Breakable("TreeCheckout", logger),
+		public readonly breaker: Breakable = new Breakable("TreeCheckout"),
```
Josmithr, line 548: Nit: might still be worth exposing a private getter property for `logger` that pulls it off of the `breaker`. Then we have a place to document why we're using the breaker's logger rather than taking our own (which would likely be non-obvious to someone reading this code for the first time).

Josmithr: Might also be worth adding a note to `breaker` that its logger is used for other telemetry too.

packages/dds/tree/src/shared-tree-core/sharedTreeCore.ts
Josmithr, line 151: Nit: probably worth a quick note that `breaker`'s logger is also used for other telemetry logging.

Josmithr (commented):
I'm not sure how I feel about the pattern of re-using the `breaker`'s logger for other telemetry logging. I could imagine a world in which the breaker wrapped its input logger with some breaker-specific context. If we did that, we would end up affecting unrelated logs.

packages/dds/tree/src/shared-tree/treeCheckout.ts
CraigMacomber: Done.

Josmithr (approved):

PR #27869, opened 2026-08-06
fix(codespaces): resolve Copilot CLI for start

steffenloesch:
## Description

> [!IMPORTANT]
> This PR is blocked on publication of `@fluid-tools/build-cli@0.67.0`.
> Do not merge it until that package is available from npm and the manual Codespace validation below passes.

The stable AI Codespace currently installs the latest published `flub`, whose `flub ai` command cannot resolve the Copilot CLI under pnpm's scoped package layout. This causes `start` to fail before opening its interactive session:

`Copilot CLI not found at .../node_modules/@github/index.js`

Screenshot of the error from a new Codespace:
Copilot CLI resolution error

The Copilot CLI resolution fix will be released separately in build-tools 0.67.0. Once that release is available, this PR:

- Installs `@fluid-tools/build-cli@>=0.67.0` in the stable AI Codespace so the launcher fix is guaranteed while later releases remain eligible.
- Invokes the `flub` executable installed by each Codespace profile directly, preventing a repository-local older package from shadowing it after `pnpm install`.
- Bumps both AI Codespace prebuild versions so stable and Insiders prebuilds include the corrected launcher.

This PR intentionally does not install Copilot CLI separately or depend on Agency's internal Copilot cache layout.

### Blocking dependency

- Build-cli fix PR: _to be added_
- Build-tools 0.67 release branch: _to be added_
- Release tag: `build-tools_v0.67.0` _not yet available_
- npm package: `@fluid-tools/build-cli@0.67.0` _not yet available_

### Validation after publication

Create fresh stable and Insiders AI Codespaces and verify:

```bash
flub --version
type -a flub
start
pnpm install
flub --version
start
```

Both `start` runs must authenticate, create an interactive session, and display the workflow-selection prompt. The selected `flub` must remain the profile-installed version after `pnpm install`.

## Reviewer Guidance

The intended final architecture is that build-cli owns its SDK-compatible Copilot CLI dependency. Codespaces install a released build-cli version and do not perform their own Copilot CLI discovery.


scripts/codespace-setup/install-build-cli.sh
```diff
@@ -2,12 +2,12 @@
...
 # Prerequisites: node + corepack must already be set up (on-create.sh), and
-# PNPM_HOME must be on PATH (set in .devcontainer/Dockerfile).
+# PNPM_HOME/bin must be on PATH (set in .devcontainer/Dockerfile).
 
 set -euo pipefail
 
-echo "Installing @fluid-tools/build-cli globally..."
-pnpm add -g @fluid-tools/build-cli
+echo "Installing @fluid-tools/build-cli and its Copilot CLI runtime globally..."
+pnpm add -g @fluid-tools/build-cli '@github/copilot@^1.0.43'
```
markfields: Ok so we're pinning to copilot v1.x?  I guess that's fine

markfields (commented):
The comment about the redundant block was a conclusion Copilot came to after I asked a few "why" quetions about the code.  Please consider carefully before incorporating the feedback.  Otherwise, looks good!

scripts/codespace-setup/agent-aliases.sh
```diff
@@ -56,7 +56,22 @@ flub-ai() {
...
+				copilot_cli_path="$candidate"
+				break
+			fi
+		done
+	fi
+	if [[ -n "$copilot_cli_path" ]]; then
+		COPILOT_CLI_PATH="$copilot_cli_path" pnpm exec flub ai --launchFile "$launch_file" "$@"
+	else
+		pnpm exec flub ai --launchFile "$launch_file" "$@"
+	fi
```
markfields: This block is redundant. `CopilotClient` prioritizes the explicit `cliPath` passed by `runAiSession` over `COPILOT_CLI_PATH`, and the Copilot loader does not read this variable, so these branches always produce the same behavior. Please remove this path-discovery logic and keep the original `pnpm exec flub ai ...` call.

steffenloesch: I've started to work on this, because `start` didn't work on a new stable Codespace. I've added a screenshot to the description.

There a different settings between the stable codespace and the experiemental one. Here's the Copilot response:

> This block is required for the stable AI Codespace, which installs published  flub  0.65 and therefore does not pass an explicit  cliPath . The SDK reads  COPILOT_CLI_PATH  before falling back to its broken pnpm package discovery. I verified published  flub  fails without the variable and starts successfully with it. Insiders uses the locally built  flub , where the explicit  cliPath  handles resolution.

However, I'm also not a big fan of the path magic here, and installing the Copilot CLI separately. The CLI should come with agency, and things work with the `copilot` command. Will look into this some more to better understand the setup, and see if there's a better fix.

scripts/codespace-setup/install-build-cli.sh
steffenloesch: It's a fair concern. I think the CLI version should match agency's.

scripts/codespace-setup/agent-aliases.sh
steffenloesch: I don't see a good agency route:

> I verified that Agency manages its own Copilot CLI and currently selects 1.0.78. However, it exposes neither a stable executable path nor a machine-readable resolver command. Its  ~/.copilot-cli/<version>  layout and resolver logs are internal, and  agency copilot  cannot serve directly as the SDK subprocess because Agency injects  --session-id , which conflicts with headless mode. For now, the separately installed 1.x CLI is the safer SDK-owned dependency. Reusing Agency’s copy would require a supported resolver API from Agency.

steffenloesch: I think if we release a new build-tools version, we can get things working in a cleaner way. Will transform this PR to illustrate how, but it'll need to be held until a new build-tools version is out.

steffenloesch:
PR #27869 is now intentionally held as the final Codespace consumer change and is blocked on publication of @fluid-tools/build-cli@0.67.0. The branch is being rewritten to remove the temporary separate Copilot installation and path-discovery logic; runtime validation will resume after the build-tools release is available.

steffenloesch:
The branch has been rewritten to the final held architecture: the PR now contains only Codespace setup changes, requires @fluid-tools/build-cli >=0.67.0, invokes the profile-installed flub directly, and removes the temporary separate Copilot CLI dependency and path-discovery logic. It remains draft and blocked until build-tools 0.67.0 is published and fresh stable/Insiders Codespaces pass the documented validation.

scripts/codespace-setup/agent-aliases.sh
steffenloesch: Made the changes. Will now work on the prerequisite `build-tools` change.

steffenloesch:
The blocking build-cli fix is now available for review in [PR #27886](https://github.com/microsoft/FluidFramework/pull/27886). PR #27869 remains draft until that fix is merged, included in build-tools 0.67.0, published, and validated in fresh Codespaces.

steffenloesch:
This PR should not get completed before #27886 is completed, and the next 0.67.0 version of `build-cli` is published with it.

tylerbutler:
Unless something changed, it should be possible to use the built-in version of build-tools so you don't have to publish first. Regrardless of whether you do it this time, it would be useful to think through the "speed of light" for deploying changes in this area.

steffenloesch:
> Unless something changed, it should be possible to use the built-in version of build-tools so you don't have to publish first. Regrardless of whether you do it this time, it would be useful to think through the "speed of light" for deploying changes in this area.

The stable codespace installs from npm, see https://github.com/microsoft/FluidFramework/blob/main/.devcontainer/ai-agent/devcontainer.json#L45, whereas the insiders one installs the built-in one: https://github.com/microsoft/FluidFramework/blob/main/.devcontainer/ai-agent-insiders/devcontainer.json#L45

PR #25425, opened 2025-09-09
build(client): Include re-exports in API reports

Josmithr:
Updates our API-Extractor configurations for report generation to include other local packages' exports in the generation. This makes reviewing the cross-package implications of API changes easier to review.

PR #27873, opened 2026-08-06
build(client): Update API-Extractor config setup to ensure correct report generation and rename configs more intuitively

Josmithr:
Most of the repo's packages were not correctly configured with regards to their report generation. Most of the configs used for report generation were also used for model generation, but report generation has some different needs that were captured by `api-extractor-report-base.esm.json`, which was not being extended in these cases.

This PR does 3 things to resolve the issues here:
1. For the packages that used a single config for model and report generation, it splits those into 2 separate configs (and moves them into the `api-extractor` directory in each package.
2. Updates "report" configs to correctly extend our base report config.
    - The impact of this change can be seen in the handful of updated API reports. Cross-package re-exports from within the client release group now appear in full detail in reports, and are correctly filtered by release tag to only appear in the _correct_ reports (matching the policies leveraged by our own internal tooling used for rollup generation).
3. Updates config names to ensure "model" and "report" are included in the names to make the set of configs easier to differentiate.

This PR also does some other misc. cleanup (e.g. removing configs that were unused after the above changes), and expands repo documentation about API-Extractor and how it's used.


docs/content/Contributing/Maintaining-API-Support-Levels.md
```diff
@@ -27,7 +27,7 @@ Packages using standard build tooling require two manual configurations:
      	},
     ```
 
-2. For packages with "./legacy" APIs, "api-extractor/api-extractor.current.json"'s "mainEntryPointFilePath" value should be set to the least stable .d.ts types path in use. E.g. this preference order: `"<projectFolder>/lib/alpha.d.ts"`, `"<projectFolder>/lib/beta.d.ts"`, and finally `"<projectFolder>/lib/public.d.ts"`. Eventually, this configuration could be automated via policy checker. (Packages without "./legacy" will reference `"<projectFolder>/lib/index.d.ts"` in "api-extractor.json".)
+2. For packages with "./legacy" APIs, "api-extractor/api-extractor-model.current.json"'s "mainEntryPointFilePath" value should be set to the least stable .d.ts types path in use. E.g. this preference order: `"<projectFolder>/lib/alpha.d.ts"`, `"<projectFolder>/lib/beta.d.ts"`, and finally `"<projectFolder>/lib/public.d.ts"`. Eventually, this configuration could be automated via policy checker. Packages without "./legacy" use "api-extractor-model.json" for model generation.
```
jason-ha: For `Packages without "./legacy" use "api-extractor-model.json" for model generation.`, what is "api-extractor-model.json"? Is that the local package's path/name? Should it be "api-extractor/api-extractor-model.json" for consistent location?

examples/utils/example-utils/api-extractor/api-extractor-report.json
```diff
@@ -0,0 +1,8 @@
+{
+	"$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+	"extends": "<projectFolder>/../../../common/build/build-common/api-extractor-report.esm.current.json",
+	"mainEntryPointFilePath": "<projectFolder>/lib/index.d.ts",
+	"apiReport": {
+		"reportVariants": ["public", "beta", "alpha"]
+	}
```
jason-ha: These could be simpler if they used "api-extractor-report-base.esm.json" instead of ".current" version (which is meant to be paired with packages that also to legacy things).
(Or could add an explicit "-report.esm.no-legacy" config if directly to "-report-base.esm" feels wrong to better match docs in common/build/build-common/README.md. Jason's preference.)

examples/utils/example-utils/package.json
```diff
@@ -29,14 +29,18 @@
 	"types": "lib/index.d.ts",
 	"scripts": {
 		"build": "fluid-build . --task build",
+		"build:api-reports": "npm run build:api-reports:current",
+		"build:api-reports:current": "api-extractor run --local --config api-extractor/api-extractor-report.json",
```
jason-ha: There should just be "build:api-reports" (preferred simplicity)
or it should be
>		"build:api-reports": "concurrently \"npm:build:api-reports:*\"",
In either case fluidBuild.config.cjs should be updated.

packages/framework/client-logger/fluid-telemetry/api-extractor/api-extractor-report.json
```diff
@@ -0,0 +1,30 @@
...
+				"logLevel": "none"
+			}
+		},
+		"extractorMessageReporting": {
+			// Require documentation on all package-exported API items.
+			"ae-undocumented": {
+				"logLevel": "error",
+				"addToApiReportFile": false
+			}
+		}
```
jason-ha, line 24: api-extractor really needs multi extend support

packages/framework/presence-definitions/api-report/presence-definitions.alpha.api.md
```diff
@@ -118,9 +118,9 @@ export namespace InternalPresenceTypes {
 // @alpha @system
 export namespace InternalPresenceUtilityTypes {
     // @system
-    export type IfNotificationParametersSignature<Event, IfParametersValid, Else> = Event extends (...args: infer P) => void ? InternalCoreInterfacesUtilityTypes.IfSameType<P, JsonSerializable<P>, IfParametersValid, Else> : Else;
+    export type IfNotificationParametersSignature<Event, IfParametersValid, Else> = Event extends (...args: infer P) => void ? InternalUtilityTypes.IfSameType<P, JsonSerializable<P>, IfParametersValid, Else> : Else;
```
jason-ha, line 121: This is not correct.
The namespace exported from core-interfaces is `InternalCoreInterfacesUtilityTypes`. (Was correct before.)

jason-ha (commented):
I guess there was never a regression. It is just that #25425 didn't account for the `api-extractor-base.esm.no-legacy.json` for combined model+report.
I think there are probably no uses left for that config and it should be removed.

docs/content/Contributing/Maintaining-API-Support-Levels.md
Josmithr: Yeah, this was intended to be the package-local config for model generation. Looks like the more qualified path made its way to one mention here but not the other.
```suggestion
2. For packages with "./legacy" APIs, "api-extractor/api-extractor-model.current.json"'s "mainEntryPointFilePath" value should be set to the least stable .d.ts types path in use. E.g. this preference order: `"<projectFolder>/lib/alpha.d.ts"`, `"<projectFolder>/lib/beta.d.ts"`, and finally `"<projectFolder>/lib/public.d.ts"`. Eventually, this configuration could be automated via policy checker. Packages without "./legacy" use a "api-extractor/api-extractor-model.json" configuration for model generation.
```

packages/framework/presence-definitions/api-report/presence-definitions.alpha.api.md
Josmithr: Will investigate.

packages/framework/client-logger/fluid-telemetry/api-extractor/api-extractor-report.json
Josmithr: Hrm. I'm also not sure I buy the documentation here. I'll see if we can remove this workaround.

packages/framework/presence-definitions/api-report/presence-definitions.alpha.api.md
Josmithr: This appears to be an artifact of `core-interfaces` being included in `bundledPackages` - API-Extractor is deduplicating the symbols, and our aliasing trick is being effectively discarded. This strictly affects the report though, and has no impact on our rollups (obviously, since those don't use api-extractor), nor the generated docs model (which intentionally _does not_ leverage the bundling functionality). So this should be safe, if a bit confusing.

packages/framework/client-logger/fluid-telemetry/api-extractor/api-extractor-report.json
Josmithr: Yeah, I was able to clear all of this out.

packages/framework/presence-definitions/api-report/presence-definitions.alpha.api.md
jason-ha: Seems bug worthy - anywhere in repo we can note the funny things with see with api-extractor?

examples/utils/migration-tools/api-extractor/api-extractor-report.json
```diff
@@ -0,0 +1,5 @@
+{
+	"$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+	"extends": "<projectFolder>/../../../common/build/build-common/api-extractor-report-base.esm.json",
+	"mainEntryPointFilePath": "<projectFolder>/lib/alpha.d.ts"
```
jason-ha: Why do we need alpha.d.ts instead of index.d.ts when there is no legacy?

fluidBuild.config.cjs
```diff
@@ -206,12 +206,11 @@ module.exports = {
...
 		},
-		// build:api-reports may be handled in one step with build:docs when a
-		// package only uses api-extractor supported exports, which is a single
-		// export/entrypoint. For packages with /legacy exports, we need to
-		// generate reports from legacy entrypoint as well as the "current" one.
+		// Packages with a single API report may handle build:api-reports directly.
+		// Packages with /legacy exports generate reports from legacy and "current" entrypoints using child tasks.
 		// The "current" entrypoint should be the broadest of "public.d.ts",
 		// "beta.d.ts", and "alpha.d.ts".
+		"build:api-reports": ["build:entrypoints:esm", "api-extractor:esnext", "build:esnext"],
```
jason-ha: Companion question to the one about some packages using alpha.d.ts/public.d.ts instead of index.d.ts when there is no legacy?
Without legacy and report tasks using index.d.ts, the deps should be reducible to just "build:esnext".

examples/utils/example-utils/package.json
jason-ha: It looks like we ended up with the bad combination of deps and change here. There are two options:
1. simple script set (Jason's preferred simplicity):
```text
package.json[scripts]:
		"build:api-reports:current": "api-extractor run --local --config api-extractor/api-extractor-report.json",

fluidBuild.config.json:
		"build:api-reports": ["build:esnext"],
  ...
		"ci:build:api-reports": ["build:esnext"],
```
2. multiple scripts (the "current" name is a little funky)
```text
package.json[scripts]:
		"build:api-reports": "concurrently \"npm:build:api-reports:*\"",
		"build:api-reports:current": "api-extractor run --local --config api-extractor/api-extractor-report.json",

fluidBuild.config.json
<no new additions>
```

jason-ha (commented):
There some merge conflicts - why do we let others do work while we're working?

examples/utils/migration-tools/api-extractor/api-extractor-report.json
Josmithr: Fair point. No reason other than Copilot making assumptions and me not thinking about it 😉 Will update.

packages/framework/presence-definitions/api-report/presence-definitions.alpha.api.md
Josmithr: I can file a bug, though I doubt there will be any action on it. Still, it's something we can reference in our own documentation, so probably worth doing. I've added this to my TODO list.

examples/utils/migration-tools/api-extractor/api-extractor-report.json
Josmithr: Updated.

fluidBuild.config.cjs
Josmithr: Updated.

packages/framework/presence-definitions/api-report/presence-definitions.alpha.api.md
Josmithr: Rushstack issue: https://github.com/microsoft/rushstack/issues/5920

I've also added some docs to our repo about API-Extractor, including notes about this issue.

docs/content/Development-Environment/API-Extractor.md
```diff
@@ -0,0 +1,63 @@
...
+
+API reports and API models intentionally use separate configurations.
+Reports are generated for individual release levels and bundle local dependencies so cross-package re-exports can be reviewed.
+Models retain the complete symbol graph and stricter link validation needed to generate the API reference documentation.
+
+## Working with API-Extractor
+
+The package-level `build:api-reports` task regenerates committed `*.api.md` review files after `build:esnext` produces declarations.
+CI runs the corresponding `ci:build:api-reports` task and fails when committed reports are out of date.
+Packages with "current" and "legacy" surfaces run child tasks for each generated entrypoint.
```
jason-ha: I know line 26 qualifies things, but it feels misleading since so many packages of interest have legacy. Maybe remove the after `build:esnext` content - then it should all be accurate.

docs/content/Development-Environment/API-Extractor.md
```diff
@@ -0,0 +1,63 @@
...
+
+Fluid Framework uses [API-Extractor](https://api-extractor.com/) to validate package API surfaces, generate API reports for code review, and produce the API models consumed by our API documentation tooling.
+
+## Configuration
+
+Most packages keep their API-Extractor configuration in an `api-extractor` directory and extend shared configuration from [`common/build/build-common`](../../../common/build/build-common/).
+The shared configuration is layered so common validation rules and package-specific output concerns stay separate:
+
+- [`api-extractor-base.json`](../../../common/build/build-common/api-extractor-base.json) defines validation, message reporting, output locations, and other defaults, but leaves all outputs disabled.
+- [`api-extractor-base.esm.json`](../../../common/build/build-common/api-extractor-base.esm.json) selects `lib/index.d.ts` as the standard ESM entrypoint and enables TSDoc metadata.
```
jason-ha: That is true but it also enables both report and model at once which is a no no. Seems like that file should be removed.

docs/content/Development-Environment/API-Extractor.md
```diff
@@ -0,0 +1,63 @@
...
+- [`api-extractor-report-base.esm.json`](../../../common/build/build-common/api-extractor-report-base.esm.json) configures API reports and bundles Fluid package dependencies so re-exported APIs appear in review files.
+- Package-level model configurations enable API model generation for the complete API surface used by the documentation toolchain.
+
+Packages without a `./legacy` export generally use a `api-extractor-report.json` config file for report generation, and `api-extractor-model.json` config file for model generation.
+Packages _with_ a `./legacy` export use separate `api-extractor-report.current.json` and `api-extractor-report.legacy.json` report configurations because the two entrypoints expose different API surfaces.
+
+API reports and API models intentionally use separate configurations.
+Reports are generated for individual release levels and bundle local dependencies so cross-package re-exports can be reviewed.
+Models retain the complete symbol graph and stricter link validation needed to generate the API reference documentation.
```
jason-ha: Not sure what of this belongs here versus in common/build/build-common/README.md. Feels like more should be in the README.

examples/utils/migration-tools/package.json
```diff
@@ -37,8 +37,10 @@
 	},
 	"scripts": {
 		"build": "fluid-build . --task build",
+		"build:api-reports": "concurrently \"npm:build:api-reports:*\"",
+		"build:api-reports:current": "api-extractor run --local --config api-extractor/api-extractor-report.json",
```
jason-ha: example/utils/example-utils/package.json was fixed but this non-preferred script set remains in many places.
ref: https://github.com/microsoft/FluidFramework/pull/27873#discussion_r3739493460

examples/utils/migration-tools/api-extractor/api-extractor-report.json
jason-ha: I know why now.
We want to have those there because the generated guys are the truth about what we export.
Using index.d.ts allows fluidBuild.config.json:
```
		"build:api-reports": ["build:esnext"],
  ...
		"ci:build:api-reports": ["build:esnext"],
```
instead of also depending on the entrypoint generation.
If our "roll-up" differs, then we don't have an accurate listing of APIs. So, it is definitely safer with the before pattern. Probably they agree.
The better option is to have an additional not `--local` report run using index.d.ts to confirm there aren't any discrepancies.

docs/content/Development-Environment/API-Extractor.md
Josmithr: Ah, good call.

docs/content/Development-Environment/API-Extractor.md
Josmithr: Fair. I can move, and maybe just leave a link from `docs` for discoverability.

examples/utils/migration-tools/api-extractor/api-extractor-report.json
Josmithr: I think I'll just restore the earlier behavior here for now. I'd like to avoid adding too many newly required API-Extractor runs.

Longer term, I think we may want to create a fluid-build task built around API-Extractor's programmatic API so we can do more of the stuff we want all at once, and hopefully re-use computation more effectively.

examples/utils/migration-tools/package.json
Josmithr: I had Copilot do another cleanup pass.

examples/utils/example-utils/package.json
Josmithr: Copilot: Resolved using the preferred simple script set. Single-report packages now invoke API Extractor directly through `build:api-reports`; `fluidBuild.config.cjs` supplies the declaration and generated-entrypoint prerequisites for both local and CI report tasks. Packages with actual current and legacy surfaces retain the wildcard wrapper and child tasks. The full client `build:api-reports` graph passed (230 tasks across 169 packages).

docs/content/Development-Environment/API-Extractor.md
Josmithr: Copilot: Resolved. The wording that tied report generation specifically to running after `build:esnext` has been removed. The page now states that the required declaration files and generated release-level entrypoints must exist before report generation, and separately notes that packages with current and legacy surfaces use child tasks for each entrypoint.

docs/content/Development-Environment/API-Extractor.md
Josmithr: Copilot: Resolved. The detailed configuration hierarchy, package setup patterns, report/model distinction, task structure, and known limitation now live in `common/build/build-common/README.md`. The docs page is a concise discoverability and workflow overview that links to that README as the authoritative technical reference.

examples/utils/migration-tools/package.json
Josmithr: Copilot: Resolved repository-wide. All single-report packages now use a direct `build:api-reports` script; only packages with genuine current and legacy report surfaces retain the wildcard wrapper and child tasks. A repository scan found 0 forwarding wrappers, 44 direct single-report packages, and 47 multi-report packages. `migration-tools` now uses the direct form, and the full client report build passed.

tools/benchmark/api-extractor.json
```diff
@@ -1,7 +1,24 @@
 {
 	"$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
-	"extends": "../../common/build/build-common/api-extractor-base.cjs.no-legacy.json",
+	"extends": "../../common/build/build-common/api-extractor-base.json",
```
Josmithr, line 3: Note for reviewers: this was the only remaining reference to `api-extractor-base.cjs.no-legacy.json`. I removed that config and inlined its details here.

packages/drivers/driver-base/package.json
jason-ha: This is confusing. It lists outFileLegacyBeta but there is no such entrypoint.
It seems there are a bunch of these; so, nothing to do now.

packages/dds/cell/api-extractor/api-extractor-report.json
```diff
@@ -1,4 +1,5 @@
 {
 	"$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
-	"extends": "<projectFolder>/../../../common/build/build-common/api-extractor-report-base.esm.json"
+	"extends": "<projectFolder>/../../../common/build/build-common/api-extractor-report-base.esm.json",
+	"mainEntryPointFilePath": "<projectFolder>/lib/public.d.ts"
```
jason-ha, line 4: no need to change: it seems odd that we generate .beta and .alpha reports when we use public.d.ts and clearly there is nothing below.

jason-ha (approved):
Needs merge fix but otherwise looks good

packages/dds/cell/api-extractor/api-extractor-report.json
Josmithr: I have a follow-up branch that adds some tooling to help prevent unnecessary report generation based on which entrypoints exist in the package.json. I'll put that up shortly after this PR lands.

alexvy86 (approved):
Stamping for devtools. My understanding of the space is limited, but trusting others' in-depth review.

Abe27342 (approved):
didn't review in detail but changes to id-compressor are in line with other packages and API reports look good.

jzaffiro (approved):
Approving for docs and merge-tree

jason-ha (approved):

PR #27874, opened 2026-08-06
ci(lts): emit publish-metadata.json and delegate client publishing

MarioJGMsoft:
## Description

Onboards the LTS `Build - client packages` build to the same publishing process `main` already uses, where the build produces artifacts and a separate pipeline publishes them.

On `main`, the client build is artifact-only for publishing: it builds, packs, and writes `pack/publish-metadata.json`, and the publish pipeline consumes that artifact and owns all publish, promotion, and tag decisions. The `lts` branch had not been onboarded, so it still published in-build and emitted no metadata.

## Changes

* **New** `templates/include-generate-publish-metadata.yml`, which writes `pack/publish-metadata.json` next to the packed tarballs. Byte-identical to the copy on `main`.
* `templates/build-npm-package.yml` gains a `publishAsPartOfBuild` parameter, generates the metadata right after `npm pack`, and gates its in-build publish stage on `publish && publishAsPartOfBuild`.
* `build-client.yml` sets `publishAsPartOfBuild: false`. The other 14 consumers set `true`, keeping their current behavior.

## Behavioral parity

The publish pipeline recomputes eligibility from the candidate's branch and the `publishOverride` in the metadata, applying the same rule `include-vars.yml` applies here: only `main`, `release/*`, and `test/*` publish by default, and `force` still overrides. Manifest upload stays off for LTS.

| Case | Result |
| --- | --- |
| `lts`, override `default` | no publication |
| `lts`, override `force` | internal feeds, no manifest |
| `release/*`, `test/*` | unchanged |

LTS candidates may now reach the internal-dev feed, which the in-build LTS publish template had no stage for. Per review this is acceptable.

## Reviewer guidance

`publishAsPartOfBuild` deliberately has **no default**, so every pipeline must declare which side owns its publishing and a search for `publishAsPartOfBuild: true` is an accurate list of what has not been migrated yet.

Fixes [AB#79900](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/79900)

### DIFF
--- a/tools/pipelines/templates/include-generate-publish-metadata.yml
+++ b/tools/pipelines/templates/include-generate-publish-metadata.yml
@@ -53,14 +53,13 @@ steps:
           ;;
       esac
 
-      case "${IS_LATEST:-}" in
-        ""|'$(SetVersion.isLatest)'|false|False) isLatest=false ;;
-        true|True) isLatest=true ;;
-        *)
-          echo "##vso[task.logissue type=error]Invalid isLatest value '$IS_LATEST'"
-          exit 1
-          ;;
-      esac
+      # SetVersion.isLatest is currently only emitted for latest release builds, so an
+      # unresolved macro means false. Remove this normalization after
+      # https://github.com/microsoft/FluidFramework/pull/27840 is available in build-tools.
+      isLatest=false
+      if [ "${IS_LATEST:-}" = "true" ]; then
+        isLatest=true
+      fi
 
       # Unlike the other fields there is no safe fallback here: guessing wrong either
       # double-publishes or drops the release entirely, so require an explicit boolean.
--- a/tools/pipelines/templates/include-generate-publish-metadata.yml
+++ b/tools/pipelines/templates/include-generate-publish-metadata.yml
@@ -1,0 +1,94 @@
+# Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+# Licensed under the MIT License.
+
+parameters:
+- name: publishOverride
+  type: string
+
+# Indicates who owns the publish step for this build:
+#   true  - publishing happens as part of this build pipeline's own publish stages.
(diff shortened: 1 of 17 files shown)


tools/pipelines/templates/include-generate-publish-metadata.yml
```diff
@@ -0,0 +1,94 @@
...
+      case "$PUBLISH_OVERRIDE" in
+        default|force|skip) ;;
+        *)
+          echo "##vso[task.logissue type=error]Invalid publishOverride value '$PUBLISH_OVERRIDE'"
+          exit 1
+          ;;
+      esac
+
+      case "${IS_LATEST:-}" in
+        ""|'$(SetVersion.isLatest)'|false|False) isLatest=false ;;
```
Abe27342: this is the same bug I fixed in https://github.com/microsoft/FluidFramework/pull/27841

I guess in practice it shouldn't matter because the bug means "isLatest" always reports false, which should be correct for LTS. but probably better to keep the branches aligned :)

markfields (approved):

tools/pipelines/templates/include-generate-publish-metadata.yml
MarioJGMsoft: Applied change, but I do have a follow up question, would it be possible to bump the LTS branch to the newest version of Build Tools? Is that okay?

Abe27342: In theory, yeah. Might be a can of worms to sort out if you try. Up to you (looks like you took my fix which should not require it; dunno if you want latest build tools for other reasons)

alexvy86 (approved):
Mostly stamping for cr-infra. I looked through the files, overall idea makes sense, but didn't think too hard about potential issues nor tried to nitpick logic. Since this is mostly emulating main, I trust it's at least good enough :)

PR #24422, opened 2025-04-22
container-runtime: Add API for default configuration

scottn12:
## Description

This PR adds an API that allows customers to set leverage the default configuration logic added in https://github.com/microsoft/FluidFramework/pull/24379. The API is an **optional** property added to the `LoadContainerRuntimeParams` interface, which is used when instantiating a container runtime.

This PR also renames the API from `compatibilityVersion` -> `minVersionForCollab` (it was changed from `compatibilityMode` -> `compatibilityVersion` in https://github.com/microsoft/FluidFramework/pull/24407).

### Considerations

The API was added to `LoadContainerRuntimeParams` for two reasons:
1. Allow us to give clear guidance to customers. For example, "Pass in `minVersionForCollab` X to ensure collaboration works with clients running runtime version X. 
2. It provides a simple path to requiring customers to pass in a `minVersionForCollab` in the future (by making the param required).
 
One alternative is to expose a helper function that will generate the `runtimeOptions` object that is used when instantiating a container runtime. However, this may complicate the guidance given to customers and does not provide a clear path to making this a required step in the future.

## Example

A customer may use the API like so:
```ts
const containerRuntime = await ContainerRuntime.loadRuntime({
    context,
    registryEntries,
    existing,
    provideEntryPoint,
    runtimeOptions: {
	enableGroupedBatching: false,
    },
    minVersionForCollab: "2.20.0", 
});
```
There are two noteworthy observations:
1. `minVersionForCollab` is set to `"2.20.0"`. This will ensure the default configurations for `IContainerRuntimeOptionsInternal` are set such that we will ensure that clients running version 2.20.0 or later of the runtime can collaborate.
5. We also manually set `enableGroupedBatching` in the `runtimeOptions` param. Batching is normally enabled by default when `minVersionForCollab` is set to `"2.0.0"` or later, but any configurations manually set in `runtimeOptions` will override the default configurations.

## Next Steps
- Add "fail fast" mechanism that prevents explicity setting `runtimeOptions` in such a way that contradicts `minVersionForCollab`. For example, setting `enableGroupedBatching` to be `true` and `minVersionForCollab` to `"1.4.0"`. This should not be allowed since clients running v1.4.0 of the runtime will not be able to collaborate if batching is enabled. See [AB#37810](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/37810)

## Misc
[...]

PR #27880, opened 2026-08-06
feat: support single-request create with detached blobs

anthony-murphy:

## Description

Adds an opt-in BlobManager summary format that allows a detached container containing blobs to be created in a single request.

### Runtime-owned attach flow

With `inlineDetachedBlobsAsSummaryBlobs: true`, BlobManager intercepts `uploadBlob` while the runtime is detached, keeps the bytes in runtime-owned state, and does not call detached blob storage. Detached blob storage therefore reports size zero, so the existing loader no-blob path passes the runtime attachment summary unchanged to `createContainer`. The feature-disabled path is unchanged.

```mermaid
flowchart TD
    A[uploadBlob while detached] --> B{Summary-blob option enabled?}
    B -- No --> C[Existing detached blob storage path]
    B -- Yes --> D[Keep bytes in BlobManager]
    D --> E[Detached storage remains empty]
    E --> F[Existing loader no-blob attach path]
    F --> G[createContainer with runtime summary]
```

### Summary format

All blobs created while detached share one loading-group tree. Each BlobManager `localId` is used directly as its stable summary entry name:

```text
.blobs/
├── .redirectTable                    SummaryType.Blob containing []
└── .detached/                         groupId: fluid-internal:detached-blobs
    ├── <localId-1>                    SummaryType.Blob containing base64 text
    ├── <localId-2>                    SummaryType.Blob containing base64 text
    └── ...
```

BlobManager always writes `.redirectTable` when `.detached` is present. It contains `[]` for a summary-only BlobManager and contains the ordinary legacy redirects for mixed state. This preserves the loader's existing `.blobs` invariant without teaching it about `.detached`.

The complete tree is returned on snapshot load, including every `localId -> service blob ID` entry; loading-group filtering applies only to blob contents. BlobManager therefore discovers all IDs without loading payloads and reads an individual payload later through `readBlob(serviceBlobId)`.

No manifest or detached-storage ID is needed. BlobManager alone interprets the entry names and base64 payloads; other layers see ordinary UTF-8 summary blobs in an ordinary child tree.

BlobManager has no separate persistent contents map. `IBlobManagerLoadInfo` carries the `.detached` subtree as one `ISnapshotTreeWithBlobContents`: `blobs` is the canonical `localId -> snapshot blob ID` mapping, and optional `blobsContents` is populated only during detached rehydrate. BlobManager derives `detachedBlobSummaryIds` from `blobs` and seeds `localBlobCache` from `blobsContents`; attached loads retain the mapping without local bytes and read and decode storage on demand.

BlobManager merges the `.detached` mapping into its runtime `redirectTable`, so every local blob ID resolves through the same mechanism once a remote ID is known. `detachedBlobSummaryIds` only classifies which redirect targets contain base64 summary content rather than attachment bytes. Serialized `.redirectTable` content still omits those entries because `.detached` is the authoritative persisted mapping and summary blob IDs may change when a full-tree summary writes new content.

The creating runtime retains its detached contents in `localBlobCache` after attach because it has no base-summary context that safely anchors handles. Reloaded clients use handles to:

```text
/.blobs/.detached/<localId>
```

The creator continues emitting content from its local cache, while reloaded clients reuse those handles in normal summaries. Full-tree summaries cannot use prior-summary handles, so BlobManager materializes the required contents for that summary and emits fresh base64 text.

### Compatibility and loader boundary

The format is persisted in container schema as `inlineDetachedBlobsAsSummaryBlobs` and requires `explicitSchemaControl` with `minVersionForCollab: "2.115.0"`. There is no runtime-loader compatibility handshake.

The loader has no `.detached` naming, local-ID, base64, or attach-flow logic. Its existing live pending-state `.blobs` special case is unchanged: it serializes `.redirectTable` but not attachment payloads or child-tree payloads. Consequently, a frozen load can reconstruct the container, but—as with existing attachment blobs—cannot dereference a payload that was not captured.

`captureFullContainerState` has the targeted loader change: it captures structural blobs in child trees under `.blobs` without treating direct legacy attachment entries as structural content or mistaking the self-contained `.blobs` group for a datastore loading group.

`captureReferencedAttachmentBlobs` continues to capture direct attachment entries separately. When loader-managed attachments are removed, the entire `.blobs` special case can be deleted and the ordinary recursive walker will cover this format.

### Blob lifecycle and GC

Detached-created payloads remain summary blobs for their full lifetime; a service summary-blob ID is never reinterpreted as an attachment ID. BlobManager GC identity remains `/_blobs/<localId>`, and sweep removes the local ID and summary entry from later summaries.

A rehydrated detached container may contain detached summary blobs together with legacy detached-storage blobs. BlobManager patches only the external detached-storage entries during attach and leaves summary blobs intact.

### Testing

Focused local-server scenarios cover single-request creation, binary and multiple blobs in the shared group, detached rehydrate, the legacy fallback, and `captureFullContainerState`. Portable E2E coverage exercises serialized detached creation plus later normal and full-tree summaries.

Local-server stress now uploads blobs while detached with this option enabled. The service-load pairwise matrix covers `[true, undefined]`, enables explicit schema control for the new format, and permits non-ODSP detached-blob coverage when summary-blob storage is selected.

## Reviewer Guidance

Suggested review order:

1. `blobManager.ts` for detached upload interception, lazy content ownership, mixed-state patching, GC, and full-tree materialization.
2. `blobManagerSnapSum.ts` for the shared summary format, attached-load ID reconstruction, and handles.
3. `documentSchema.ts` and `containerCompatibility.ts` for runtime format negotiation.
4. The focused `captureFullContainerState` loader diff.
5. The focused local-server and E2E lifecycle tests.


anthony-murphy:

- 🤖 **Mechanical Review** — launching the ROV…
- 🌑 **Historical / Contextual Review** — into the midnight zone…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:
## Deep Review

*Reviewed commit `3c4f5a5` on 2026-08-19.*

Not ready for sign-off. This is a well-designed, opt-in, schema-gated feature, but one Tier 2 compatibility gap blocks it: the schema gate does not force `minVersionForCollab: 2.115.0` when a caller omits it, so the new `.detached` write format can be persisted while the document still advertises 2.0 compatibility — letting a pre-2.115 runtime load a format it cannot resolve. Details are in the inline thread on `containerCompatibility.ts`. Everything else is polish, held until this is resolved.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- Follows the document-schema rollout model (#20174): this is a genuine *write-format* flag, so enabling it must kick incompatible clients out. Check every runtime/test factory that enables the option sets both `explicitSchemaControl` **and** an appropriate `minVersionForCollab` — forcing `explicitSchemaControl` alone is not the same as forcing the version.
- `.blobs` child-tree capture re-reads detached payloads on full-container-state capture — the same download-all profile #16031 deliberately narrowed away, now scoped to genuine pending state rather than all attachments. Confirm only structural child-tree blobs under `.blobs` are captured and direct attachment entries stay excluded.
- Detached-created payloads stay summary blobs for their entire lifetime (base64 +33% on every summary, full-tree re-materialization) rather than converging to `SummaryType.Attachment`. Two independent blind proposals preferred convergence; this is a deliberate schema-versioned tradeoff that needs owner confirmation as terminal vs. interim.
- BlobManager summary internals stay behind the BlobManager boundary (the `.detached` schema is hidden from the loader, continuing the #21729 refactor and answering vladsud's #6762 layering question), and flat GC node paths (`/_blobs/<localId>`) are preserved despite the nested summary tree.
- Suggested reviewers: **ChumpChief** (blob schema-flag semantics), **wes-carlson** (`.blobs` capture/throttling), **vladsud** (schema gate/layering), **markfields** (flat-GC-path), **jason-ha** (telemetry).

- **Needs owner sign-off** — the permanent summary-blob lifetime for detached-created payloads (no convergence to `SummaryType.Attachment`). ChumpChief/vladsud should confirm it is the intended terminal format or an acknowledged interim step.
- **Needs human judgment** — confirm `validateConfigMapOverrides` genuinely early-returns for `2.0.0-defaults` so the 2.115 minimum is unenforced when `minVersionForCollab` is omitted. The pipeline could not read that function's body (not in the diff); the diff and the external reviewer thread strongly suggest a caller-dependent gate, but a domain owner should verify the exact runtime behavior.
- **Cannot be assessed by the pipeline** — atomic single-request create, base64 request-size limits, and lazy loading-group fetch on capture are only exercised on local-server and compat providers. Real-service (esp. ODSP) request-size and throttling behavior needs runtime validation.
- **Risk-signal** — the acknowledged unawaited-blob-upload replay race (#23701) is now reachable in the detached path. Watch for stress flakes on this branch before attributing them to a DDS regression.

</details>

packages/runtime/container-runtime/src/containerCompatibility.ts
```diff
@@ -209,6 +212,10 @@ const runtimeOptionsAffectingDocSchemaConfigValidationMap: ConfigValidationMap<R
 			[undefined, "1.0.0"],
 			[true, "2.40.0"],
 		]),
+		inlineDetachedBlobsAsSummaryBlobs: configValueToMinVersionForCollab([
+			[undefined, "1.0.0"],
+			[true, "2.115.0"],
+		]),
```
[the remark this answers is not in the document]
anthony-murphy: Deep Review: Confirming this gap — the diff corroborates that the gate is caller-dependent rather than schema-enforced. The service-load pairwise matrix enables `inlineDetachedBlobsAsSummaryBlobs: [true, undefined]` (optionsMatrix.ts) setting only `explicitSchemaControl` and never `minVersionForCollab`, while the local-server stress factory does pair it with `2.115.0` (stressDataObject.ts:1243) — the exact split you flagged.

Beyond passing `2.115.0` from the service-load factory, add an explicit guard so enabling this option forces the document's effective `minVersionForCollab` to at least `2.115.0` (or hard-fails when a lower version is advertised), independent of whether the caller passes it. Otherwise any consumer omitting `minVersionForCollab` can persist the `.detached` format under a 2.0-compatible document, and a pre-2.115 runtime that lacks the `.detached`/base64/`detachedBlobSummaryIds` logic will load it and be silently unable to dereference the handles. Worth a test asserting a pre-2.115 runtime is rejected when opening such a container.

packages/runtime/container-runtime/src/blobManager/blobManager.ts
```diff
@@ -792,7 +852,67 @@ export class BlobManager {
...
+	}
+
+	/**
+	 * Materializes payloads that would otherwise be summarized by handle.
+	 * Used only when producing a full-tree summary.
+	 */
+	private async loadFullTreeContents(): Promise<Map<string, ArrayBufferLike>> {
+		const fullTreeContents = new Map(this.getDetachedBlobSummaryContentsFromLocalCache());
+		await Promise.all(
+			[...this.detachedBlobSummaryIds].map(async (localId) => {
```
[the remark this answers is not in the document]
anthony-murphy: Addressed with a convergence model in dec7626da2. When an attached container loads `.blobs/.detached`, the snapshot blob IDs are now normalized into ordinary local-ID-to-service-blob-ID redirects. Subsequent incremental and full-tree summaries emit `SummaryType.Attachment` entries plus the redirect mapping, so full-tree summarization performs no `readBlob` calls. Detached/rehydrating containers still retain local bytes and emit `SummaryType.Blob` for the initial single-request create. Added a 65-blob test that fails on any storage read and verifies attachment convergence.

anthony-murphy:

- 🤖 **Mechanical Review** — launching the ROV…
- 🌊 **Solution / Problem Space Review** — drifting with the current…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:
## Deep Review

*Reviewed commit `dec7626` on 2026-08-20.*

Not ready for sign-off. The prior blocking gap is fixed — enabling `inlineDetachedBlobsAsSummaryBlobs` now throws a `UsageError` unless `minVersionForCollab >= 2.115.0`, and the read-storm concern is resolved via the attached-client convergence model. But a new, equally-blocking correctness bug surfaced this run: the frozen-state capture path recurses into `.blobs/.detached` and round-trips raw binary summary blobs through UTF-8, silently corrupting non-UTF-8 bytes on the feature's headline path. Details are in the inline thread on `captureReferencedContents.ts`.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Confirm whether the shipped `Uint8Array([0x80, 0xff, 0x00, 0x7f])` byte-equality test actually exercises the `.detached` recursion + `readReferencedSnapshotBlobs` UTF-8 path (or add a test that does), and document whether the unconditional base64 switch on `IPendingDetachedContainerState` is an intended `@internal` same-version-only serialized-state format bump

### Context for Reviewers

- The prior blocking compatibility gate is now enforced in code: `validateRuntimeOptions` throws a `UsageError` when the option is enabled with the default `minVersionForCollab`, and the service-load factory now passes `2.115.0`. The earlier caller-dependent-gate concern is resolved.
- The blob-read-storm concern is resolved via a convergence model: when an attached container loads `.blobs/.detached`, snapshot blob IDs are normalized into ordinary local-ID→service-blob-ID redirects and re-summarized as `SummaryType.Attachment`, so full-tree summaries perform no storage reads (65-blob test asserts zero reads).
- Detached-created payloads stay summary blobs for their entire lifetime (base64 +33% per summary, full-tree re-materialization) rather than converging to `SummaryType.Attachment`. This is a deliberate schema-versioned tradeoff — owner confirmation needed on terminal vs. interim.
- Follows the document-schema rollout model (#20174): this is a genuine *write-format* flag with a hard `minVersionForCollab >= 2.115.0` gate, deliberately choosing a persisted-format + hard version gate over the collaborate-anyway handshake used for `createBlobPayloadPending` (#24320).
- The `inlineDetachedBlobsAsSummaryBlobs: undefined` option-assertion literals edited here have no compile check and reportedly aren't run in CI (jason-ha, #24422) — verify them manually.
- Stress coverage now enters the detached-blob path (the prior local-server skip guard is removed); the acknowledged unawaited-blob-upload replay race (#23701) is now reachable here — inspect the detached-blob operation distribution before attributing stress flakes to a DDS regression.
- Suggested reviewers: **ChumpChief** (BlobManager mechanics, the S1 fix + convergence classification), **vladsud** (schema/version gate), **jason-ha** (opt-in default + un-CI'd option assertions), **dannimad** (BlobManager construction), **wes-carlson** (`.blobs` capture).

- **Needs human judgment** — Confirm the permanent summary-blob lifetime for detached-created payloads (the creator re-emits inline bytes on every summary; the `addHandle` reuse branch is unreachable) is the intended terminal format or an acknowledged interim step. ChumpChief/vladsud should confirm, or the PR description's handle-reuse claim should be corrected.
- **Needs human judgment** — Whether the unconditional base64 switch on `IPendingDetachedContainerState` is an accepted `@internal` serialized-state format bump. The forward direction (old loader reads new base64) is ungated and undocumented; the author himself set the sticky-serialized-state rule (#20198).
- **Cannot be assessed by the pipeline** — Real-service (esp. ODSP) request-size and throttling behavior for the atomic single-request create with large inline payloads. Only exercised on local-server/compat providers.
- **Risk-signal** — The acknowledged unawaited-blob-upload replay race (#23701) is now reachable on the detached stress path. Watch for stress flakes on this branch before attributing them to a DDS regression.

</details>

- `3c4f5a5` 2026-08-19 · **5/10** — `minVersionForCollab` gate unenforced when the caller omits it (Tier 2)

</details>

packages/loader/container-loader/src/captureReferencedContents.ts
```diff
@@ -194,6 +195,9 @@ function collectReferencedBlobIds(
 			if (tableBlobId !== undefined) {
```
anthony-murphy, line 195: Deep Review: This path silently corrupts binary `.detached` blobs on the feature's headline frozen-capture flow.

This PR adds recursion so `collectReferencedBlobIds` now walks child trees under the root `.blobs` tree: `for (const childTree of Object.values(subTree.trees)) { collectReferencedBlobIds(childTree, false, ids); }`. That pulls in the `.detached/<localId>` blob ids. `readReferencedSnapshotBlobs` then serializes every collected blob with `bufferToString(data, "utf8")`, and the frozen-load side decodes with `stringToBuffer(serializedContent, "utf8")` (serializedStateManager.ts:317-318).

The `.detached` entries hold arbitrary binary `Uint8Array` payloads — written raw at `blobManagerSnapSum.ts:189` via `addBlob(localId, new Uint8Array(detachedBlobSummaryContent))` — so the UTF-8 round-trip is lossy for non-UTF-8 bytes (`0x80`, `0xff` → U+FFFD → `EF BF BD`). These blobs are *not* rescued by the binary-safe base64 attachment path: `captureReferencedAttachmentBlobs` reads only `.redirectTable`, which for a summary-only detached container is empty (`if (localIdToStorageId.size === 0) { return {}; }`).

Fix: don't route `.blobs` child-tree payloads through the UTF-8 `snapshotBlobs` path in `collectReferencedBlobIds`. Either skip recursing into `.detached` (its bytes are meant to be captured base64 by `captureReferencedAttachmentBlobs`), or base64-encode the collected child-tree blobs onto the `IBase64BlobContents` field with the `snapshotBlobsEncoding` marker and decode them with the base64 decoder on frozen load — matching the serialized-state path this PR already fixes for the same binary-safety reason.

The PR's own test uploads `Uint8Array([0x80, 0xff, 0x00, 0x7f])` and asserts byte-equality after `captureFullContainerState` + frozen load, which cannot hold under a UTF-8 round-trip. Confirm that test actually exercises the summary-only detached-container case, and add one that does if it doesn't.

anthony-murphy:

- ✨ **Solution / Problem Space Review** — tracking bioluminescence…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:

- 🛰️ **Mechanical Review** — calibrating the sonar…
- ⬇️ **Historical / Contextual Review** — dropping through the layers…
- 🌊 **Solution / Problem Space Review** — drifting with the current…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

packages/runtime/container-runtime/src/blobManager/blobManager.ts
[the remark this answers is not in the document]
anthony-murphy: Scope update: the attachment-convergence implementation was removed because it depended on unrelated loader binary serialization changes. Full-tree summaries now materialize the base64-backed blobs with at most 32 concurrent reads; focused coverage verifies the bound. Implemented in `dc3c48bf3d`.

packages/runtime/container-runtime/src/containerCompatibility.ts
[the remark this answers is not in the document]
anthony-murphy: Tracked the default-sentinel validation gap in https://github.com/microsoft/FluidFramework/issues/28028. This PR now leaves `validateConfigMapOverrides` unchanged, retains map-driven validation for explicit versions, and configures service-load coverage with `2.115.0`. The omitted/default behavior remains tracked separately.

packages/loader/container-loader/src/captureReferencedContents.ts
anthony-murphy: Tracked the generic UTF-8 snapshot-serialization defect in https://github.com/microsoft/FluidFramework/issues/28030. This PR removes the raw-binary loader changes and restores runtime-local base64 encoding, so its `.blobs/.detached` structural blobs are UTF-8-safe through frozen capture. The generic loader fix remains out of scope.

anthony-murphy:

- 🪣 **Mechanical Review** — trimming ballast…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:

- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:

- ⏭️ Mechanical Review
- 🗺️ **Solution / Problem Space Review** — venturing into uncharted water…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:
## Deep Review

*Reviewed commit `bf032ba` on 2026-08-20.*

The two prior blocking concerns — UTF-8 corruption in frozen capture and the min-version gate — are resolved on this commit, and no correctness or design gaps remain. Only Tier 3 polish is left, flagged inline. The PR's CI build is currently failing (likely the un-tagged asserts hitting the policy gate), which also blocks the bundle-size comparison.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Get the CI build green — the bundle-size comparison is blocked on the failing build; running `flub generate assertTags` (see the assert-tagging inline thread) is the likely fix.

### Context for Reviewers

- Schema-gated write-format rollout matches the established lineage (#20174 / #24320): `inlineDetachedBlobsAsSummaryBlobs` is wired through document schema and validated `true` only for `minVersionForCollab: "2.115.0"` — a require-to-read gate for a format older clients can't read.
- BlobManager remains the owner of the summary format (#6762): ContainerRuntime threads a targeted `blobManagerSummaryOverride` and is never taught about local IDs, base64 payloads, or redirect-table interpretation.
- containerRuntime.ts is the highest-churn changed file (36 events over 6 months, 11 related PRs) — tag runtime/schema owners (vladsud, jason-ha) for the summarize-path change.
- Detached-created payloads stay summary blobs for their full lifetime (base64 +33% per summary, full-tree re-materialization) rather than converging to `SummaryType.Attachment` — a deliberate schema-versioned tradeoff; owner confirmation needed on terminal vs. interim.
- The `inlineDetachedBlobsAsSummaryBlobs: undefined` option-assertion literals have no compile check and reportedly aren't run in CI (jason-ha, #24422) — verify them manually.
- Stress coverage now enters the detached-blob path (the prior local-server skip guard is removed); the acknowledged unawaited-upload replay race (#23701) is now reachable — inspect the detached-blob operation distribution before attributing stress flakes to a DDS regression.
- Suggested reviewers: **vladsud** (schema controller/rollout, BlobManager boundary), **jason-ha** (`true | undefined` telemetry convention, un-CI'd option assertions), **ChumpChief** (BlobManager mechanics), **wes-carlson** (`.blobs` capture), **Abe27342** (unawaited-upload stress race).

- **Needs human judgment** — Confirm the permanent summary-blob lifetime for detached-created payloads (the creator re-emits inline bytes on every summary; the `addHandle` reuse branch is reportedly unreachable) is the intended terminal format or an acknowledged interim step. ChumpChief/vladsud should confirm, or the PR description's handle-reuse claim should be corrected.
- **Needs human judgment** — Whether the unconditional base64 switch on `IPendingDetachedContainerState` is an accepted `@internal` serialized-state format bump; the forward direction (old loader reads new base64) is ungated and undocumented, and the author himself set the sticky-serialized-state rule (#20198).
- **Needs human judgment** — Whether the default-sentinel min-version validation gap (#28028) and the generic UTF-8 snapshot-serialization defect (#28030), both tracked out of scope, are acceptable to defer, given this PR keeps `validateConfigMapOverrides` unchanged and relies on runtime-local base64 for UTF-8 safety.
- **Cannot be assessed by the pipeline** — Real-service (esp. ODSP) request-size and throttling behavior for the atomic single-request create with large inline payloads; only exercised on local-server/compat providers. Full-tree summaries re-read/re-encode all detached payloads (bounded at 32 concurrent reads).
- **Cannot be assessed by the pipeline** — The `inlineDetachedBlobsAsSummaryBlobs: undefined` option-assertion literals reportedly aren't run in CI (jason-ha, #24422); verify them manually.

</details>

- `dec7626` 2026-08-20 · **5/10** — frozen-capture UTF-8 corruption flagged; min-version gate fixed
- `3c4f5a5` 2026-08-19 · **5/10** — minVersionForCollab gate unenforced when the caller omits it

</details>

packages/loader/container-loader/src/captureReferencedContents.ts
anthony-murphy: Deep Review: Confirmed resolved on `bf032ba`. Removing the raw-binary loader changes and restoring runtime-local base64 encoding makes the `.blobs/.detached` structural blobs UTF-8-safe through frozen capture, so the byte-corruption path this thread flagged no longer applies. The generic loader-level UTF-8 snapshot-serialization defect is appropriately tracked out of scope in #28030.

packages/runtime/container-runtime/src/blobManager/blobManager.ts
```diff
@@ -331,13 +341,48 @@ export class BlobManager {
...
 		});
 
 		this.redirectTable = toRedirectTable(blobManagerLoadInfo, this.mc.logger);
+		const detachedBlobSummary = blobManagerLoadInfo.detachedBlobSummary;
+		const detachedBlobSummaryEntries = Object.entries(detachedBlobSummary?.blobs ?? {});
+		this.detachedBlobSummaryIds = new Set(
+			detachedBlobSummaryEntries.map(([localId]) => localId),
+		);
+		for (const localId of this.detachedBlobSummaryIds.keys()) {
+			assert(
```
anthony-murphy, line 358: Deep Review: Six new asserts carry plain-string messages with no hex short code — e.g. `blobManager.ts:358` `assert(!this.redirectTable.has(localId), "Detached summary IDs cannot also be attachment redirects")`, `blobManager.ts:915` `assert(storageId !== undefined, "Detached blob summary ID must have a storage ID")`, and `blobManagerSnapSum.ts:93` `assert(Object.keys(detachedBlobSummaryTree.trees).length === 0, "Detached blob summary cannot contain child trees")`.

Adjacent asserts in these same files carry hex codes (`0xc7e`, `0x11f`, `0x384`, `0x5bd`, `0xc83`), and the prior refactor of this file (#21729) preserved hex-tagged asserts. Untagged asserts fail the repo's assert-tagging policy gate and produce no stable telemetry code for triaging the new detached-blob invariants — plausibly the cause of the currently-failing CI build.

Run `flub generate assertTags` so every new assert receives a hex code before merge.

packages/runtime/container-runtime/src/test/documentSchema.spec.ts
```diff
@@ -57,6 +57,7 @@ describe("Runtime", () => {
 		opGroupingEnabled: false,
```
anthony-murphy, line 57: Deep Review: The PR adds `inlineDetachedBlobsAsSummaryBlobs` to `IDocumentSchemaFeatures` (documentSchema.ts:152), `documentSchemaSupportedConfigs` (documentSchema.ts:304), and `summarizeDocumentSchema` (documentSchema.ts:681), but the only documentSchema.spec.ts change is the default fixture value `inlineDetachedBlobsAsSummaryBlobs: undefined`. No case asserts that `true` is serialized, validated, and round-tripped by `DocumentsSchemaController`, unlike the existing `createBlobPayloadPending`/`disallowedVersions` coverage.

Adjacent coverage (min-version allow/reject at containerRuntime.spec.ts:4711-4740, local-server serialize/rehydrate, E2E persisted-format round-trip) lowers regression risk but does not fill the controller-level unit gap; #20678 fixed a document-schema regression that non-targeted tests missed.

Add a focused `DocumentsSchemaController` test that enables `inlineDetachedBlobsAsSummaryBlobs: true` and asserts the summarized/current schema includes the property, mirroring the existing schema-property cases.

packages/test/local-server-stress-tests/src/test/localServerStress.spec.ts
```diff
@@ -40,8 +40,8 @@ describe("Local Server Stress", () => {
 		},
 		// Minimization is slow with many seeds; use only to minimize specific failing seeds.
 		skipMinimization: true,
-		// Pre-existing DDS bugs: seeds 0 and 54 (ConsensusOrderedCollection), seed 92 (Matrix).
-		skip: [0, 54, 92],
+		// Pre-existing ConsensusOrderedCollection bug under the current operation distribution.
```
anthony-murphy: Deep Review: Removing the `local server doesn't support detached blobs` guard in baseModel.ts (now uploading blobs while detached) shifts the operation distribution, and the skip list is rewritten from `skip: [0, 54, 92]` ("Pre-existing DDS bugs: seeds 0 and 54, seed 92 (Matrix)") to `skip: [0, 180]` ("Pre-existing ConsensusOrderedCollection bug under the current operation distribution"). The Matrix bug (old seed 92) and a ConsensusOrderedCollection seed (54) lose their comment attribution.

Keep the skip-list comment's bug attribution complete — confirm the pre-existing Matrix (old seed 92) and ConsensusOrderedCollection (seed 54) bugs are still tracked via issue links, rather than merely no longer hit under the reshuffled distribution.

packages/runtime/container-runtime/src/test/containerRuntime.spec.ts
```diff
@@ -2013,6 +2013,7 @@ describe("Runtime", () => {
 				enableGroupedBatching: true, // Redundant, but makes the JSON.stringify yield the same result as the logs
 				explicitSchemaControl: false,
 				createBlobPayloadPending: undefined,
+				inlineDetachedBlobsAsSummaryBlobs: undefined,
```
anthony-murphy: Deep Review: The PR adds `inlineDetachedBlobsAsSummaryBlobs: undefined` to test objects typed `satisfies ContainerRuntimeOptionsInternal` (here at containerRuntime.spec.ts:2016, and at AzureClient.spec.ts:420 and :439). The `true | undefined` typing (documentSchema.ts:152) exists specifically so options need not be explicitly assigned `undefined` — per jason-ha's guidance in #24320: "we can list options as `true | undefined` and then we don't have to pollute telemetry. All of the following line additions could be removed."

Remove the redundant `inlineDetachedBlobsAsSummaryBlobs: undefined` entries; optionally clean up the sibling `createBlobPayloadPending: undefined` at the same time.

packages/runtime/container-runtime/src/blobManager/blobManager.ts
anthony-murphy: this is not done manually. it is automated before release. assert tags should be strings at checkin

anthony-murphy:
## Deep Review

*Reviewed commit `e7d4c5d` on 2026-08-21.*

Not full sign-off, but close. The two prior blocking concerns — UTF-8 corruption in frozen capture and the unenforced min-version gate — are resolved, and no correctness or design gaps remain. Only Tier 3 polish and hardening is left: two hardening items are flagged inline (a defensive assert/negative test on the loader `.blobs` recursion, and telemetry/ceiling on inlined detached blobs) alongside three open test/cleanup threads.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Confirm the cause of any remaining CI build failure so the bundle-size comparison completes — the assert-tag theory is ruled out (string-message asserts are the correct checked-in state; `flub` assigns hex codes automatically before release).

### Context for Reviewers

- The `.blobs` child-tree recursion refines the #16031 special case (still skips direct root-level attachment payloads) without reopening the offline-load throttling regression #15831 — the hardening residual is flagged inline.
- Schema-gated write-format rollout matches the #20174 / #24320 lineage: `inlineDetachedBlobsAsSummaryBlobs` is validated `true` only for `minVersionForCollab: "2.115.0"` under explicit schema control — a require-to-read gate for a format older clients can't read.
- BlobManager remains the owner of the summary format (#6762): ContainerRuntime threads a targeted override and is never taught local IDs, base64 payloads, or redirect-table interpretation.
- The no-driver-change design was independently validated over three blind proposals, all of which instead added a driver capability plus a loader attach-state branch the PR avoids.
- Detached-created payloads stay base64 summary blobs for their full lifetime (~+33% per summary, full-tree re-materialization) rather than converging to `SummaryType.Attachment` — a deliberate schema-versioned tradeoff; owner confirmation needed on terminal vs. interim.
- Stress coverage now enters the detached-blob path (the local-server skip guard is removed); the acknowledged unawaited-upload replay race (#23701) is now reachable — inspect the detached-blob operation distribution before attributing stress flakes to a DDS regression.
- The `inlineDetachedBlobsAsSummaryBlobs: undefined` option-assertion literals reportedly aren't run in CI (jason-ha, #24422) — verify them manually.
- Suggested reviewers: **wes-carlson** (`.blobs` capture, #16031), **ChumpChief** (BlobManager mechanics), **vladsud** (schema/compat, BlobManager boundary #6762), **jason-ha** (`true | undefined` telemetry convention), **scottn12** (documentSchema), **Abe27342** (stress replay race).

- **Needs human judgment** — Confirm the permanent summary-blob lifetime for detached-created payloads (base64 +33%, full-tree re-materialization) is the intended terminal format rather than an interim step, or correct the PR description's handle-reuse claim. ChumpChief/vladsud should weigh in.
- **Needs human judgment** — Whether the default-sentinel min-version validation gap (#28028) and the generic UTF-8 snapshot-serialization defect (#28030), both tracked out of scope, are acceptable to defer, given the PR leaves `validateConfigMapOverrides` unchanged and relies on runtime-local base64 for UTF-8 safety.
- **Needs human judgment** — Whether extending `loadV1`/`summarizeV1` in place (vs. a V2 boundary) is the right long-term call for an evolving `.detached` format.
- **Cannot be assessed by the pipeline** — Real-service (esp. ODSP) request-size and throttling behavior for the atomic single-request create with large inline payloads; only exercised on local-server/compat providers.
- **Cannot be assessed by the pipeline** — The `inlineDetachedBlobsAsSummaryBlobs: undefined` option-assertion literals reportedly aren't run in CI (jason-ha, #24422); verify them manually.

</details>

- `bf032ba` 2026-08-20 · **8/10** — prior blocking concerns (UTF-8 corruption, min-version gate) resolved; Tier 3 polish only
- `dec7626` 2026-08-20 · **5/10** — frozen-capture UTF-8 corruption flagged; min-version gate fixed
- `3c4f5a5` 2026-08-19 · **5/10** — minVersionForCollab gate unenforced when the caller omits it

</details>

anthony-murphy:
Deep Review: Understood — agreed that string-message asserts are the correct checked-in state and `flub` assigns hex short codes automatically before release. Withdrawing this as the CI-failure cause; no action needed on this thread.

packages/loader/container-loader/src/captureReferencedContents.ts
```diff
@@ -194,6 +195,9 @@ function collectReferencedBlobIds(
...
 				ids.add(tableBlobId);
 			}
+			for (const childTree of Object.values(subTree.trees)) {
+				collectReferencedBlobIds(childTree, false, ids);
```
anthony-murphy, line 199: Deep Review: `collectReferencedBlobIds` now recurses into every child tree of the root `.blobs` subtree (`for (const childTree of Object.values(subTree.trees)) { collectReferencedBlobIds(childTree, false, ids); }`). Once `isRoot === false`, the general path collects all `tree.blobs` values unconditionally, and `readReferencedSnapshotBlobs` reads every collected id.

PR #16031 introduced the opposite restriction: over-collection (#15831) caused containers with offline load enabled to download all attachment blobs in the snapshot, resulting in stress-test throttling and an attach-time assert. There is no current defect — root-level attachment entries are still skipped, `blobManagerSnapSum.ts:93-96` asserts the `.detached` tree cannot nest further child trees, and the writer never places attachment-classified payloads in a `.blobs` child tree.

The residual risk is that the loader-side recursion is coupled to the writer's format by prose only: there is no assert at the consumption site, and the spec test (`captureReferencedContents.spec.ts:93-114`) exercises only one small structural child blob — not a negative case proving a nested attachment payload is still skipped.

Add an assert/guard at the recursion site that child trees under root `.blobs` contain only structural summary blobs (or explicitly skip attachment-classified entries), plus a spec case proving a direct attachment payload nested beside a `.detached` tree is still skipped and a large attachment set is not bulk-read. Non-blocking to the current format.

anthony-murphy:
`packages/runtime/container-runtime/src/blobManager/blobManager.ts:540`

Deep Review: `uploadBlob` short-circuits to inline every detached blob into `localBlobCache` with an identity redirect and no size/count check. Full-tree summaries re-materialize missing detached payloads in batches of `maxConcurrentDetachedBlobReads` = 32, and the creator retains all bytes in memory for the container lifetime.

Both cross-family design evaluators independently ranked this the top operational gap — inlining degrades badly for large blob payloads, and the opt-in schema gate limits blast radius but gives no runtime visibility into inlined bytes/count. The related unbounded-concurrency concern is already resolved via the 32-read cap ([that thread](https://github.com/microsoft/FluidFramework/pull/27880#discussion_r3817565135)); the telemetry/ceiling gap remains.

Add telemetry on total inlined bytes / blob count at the `uploadBlob` interception, and consider a documented size/volume ceiling above which the inline path declines. Keep the core mechanism.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1236,6 +1240,7 @@ export class ContainerRuntime
 				idCompressorMode,
```
vladsud, line 1240: This forwards the host option into desired document schema even for already-attached existing documents. Those documents cannot use the detached-create optimization, but can acquire a sticky schema feature that excludes older clients. Request the writer feature only while detached, while continuing to recognize an already-persisted property.

packages/runtime/container-runtime/src/blobManager/blobManager.ts
```diff
@@ -854,6 +982,7 @@ export class BlobManager {
 			}
 			maybeUnusedStorageIds.add(storageId);
 			this.redirectTable.delete(localId);
```
vladsud, line 984: This removes the summary classification and redirect but leaves the creator's `localBlobCache` entry alive. The payload remains retained and `hasBlob(localId)` remains true after sweep. Remove the cache entry when deleting a summary-backed blob and cover creator and freshly loaded attached clients.

packages/runtime/container-runtime/src/blobManager/blobManager.ts
```diff
@@ -384,7 +429,11 @@ export class BlobManager {
 			return undefined;
```
vladsud, line 429: This adds a permanent third `undefined` case to `lookupTemporaryBlobStorageId`: summary-backed blobs never acquire an attachment ID. Update this TSDoc and the runtime interface contract so consumers do not wait for a temporary URL that will never become available.

vladsud (commented):
Fresh attached pending state is not self-contained: the live pending-state capture retains `.blobs/.redirectTable` but does not fetch payloads under `.blobs/.detached`. Online reload can read them lazily from service, but offline/cache-independent rehydrate can reconstruct handles whose payloads are unavailable. Capture and serialize these structural child blobs losslessly, or explicitly reject that offline workflow.

packages/runtime/container-runtime/src/containerCompatibility.ts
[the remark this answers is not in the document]
vladsud: The required floor is higher than 2.115.0: neither `client_v2.115.0` nor `client_v2.116.0` contains the schema key or `.detached` reader. After rebasing, the first compatible release line is 3.0.0. Please update the mapping and compatibility tests accordingly rather than only enforcing 2.115.0.

PR #27884, opened 2026-08-07
fix(client): update TypeScript 6 compatibility

jason-ha:
Update iterator interfaces, test mocks, extraneous casts, and undefined test variables to match the newer standard library and compiler checks.

ArrayBuffer incompatibility and use of built-in types to be addressed separately via #27815 and #27857.

### DIFF
--- a/packages/dds/shared-object-base/src/test/attachingBindingAndConnecting.spec.ts
+++ b/packages/dds/shared-object-base/src/test/attachingBindingAndConnecting.spec.ts
@@ -62,6 +62,7 @@ type OverridableType = Overridable<{
 		serializer: IFluidSerializer,
 		telemetryContext?: ITelemetryContext | undefined,
 		incrementalSummaryContext?: IExperimentalIncrementalSummaryContext | undefined,
+		fullTree?: boolean,
 	) => ISummaryTreeWithStats;
 	loadCore: (this: SharedObject, services: IChannelStorageService) => Promise<void>;
 	processMessagesCore: (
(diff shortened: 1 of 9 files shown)


packages/dds/shared-object-base/src/test/attachingBindingAndConnecting.spec.ts
```diff
@@ -78,11 +79,36 @@ function createTestSharedObject(overrides: OverridableType): {
...
-		protected processMessagesCore = overrides?.processMessagesCore?.bind(this);
-		protected onDisconnect = overrides?.onDisconnect?.bind(this);
-		protected applyStashedOp = overrides?.applyStashedOp?.bind(this);
+		protected summarizeCore(
+			serializer: IFluidSerializer,
+			telemetryContext?: ITelemetryContext,
+			incrementalSummaryContext?: IExperimentalIncrementalSummaryContext,
+			fullTree?: boolean,
+		): ISummaryTreeWithStats {
+			assert(overrides.summarizeCore !== undefined, "summarizeCore not set");
```
Josmithr, line 88: In the previous PR, you indicated in [this thread](https://github.com/microsoft/FluidFramework/pull/27815#discussion_r3706572126) that these overrides could be simplified.

jason-ha: Upon further consideration, most of the changes are the best.
Before any change there would be an exception trying to call something that was not defined. To preserve that we need a failing assertion if the call happens.
I fixed the loadCore case that was silent and now will raise assert if not set and still called.

Josmithr (approved):

PR #14150, opened 2023-02-13
Move the reference sequence number to the runtime

andre4i:
## Description

Related to ADO:3397

There is a known issue regarding reference sequence number mismatches. It is possible that the inbound queue is scheduled to start processing ops before the flush (which creates a batch) is scheduled. This may create the situation in which an op is created after the inboud queue processes, and the op will have a different reference sequence number than the rest of the batch.

Currently, the loader does not care about this, as it is not receiving the reference sequence number from the runtime at all. This will still be the case with old loaders.

This change does the following:

- moves the reference sequence number stamping at the batch level in the runtime
- it guarantees that all batches will have the same reference sequence numbers by flushing the partial batches whenever we detect a sequence number mismatch (unless the feature is disabled by feature gate). The change is correct from the perspective of batching, as if the deltamanager started processing ops, it is clear that the current batch should be over, as the current JS turn is over as well.
- this change is required for both compression/chunking correctness and for the [experimental flushmode async](https://github.com/microsoft/FluidFramework/pull/14106), which may create more situations like described above.

PR #12938, opened 2022-11-15
Some refactoring related to the runtime's outgoing op pipeline

andre4i:
Related to ADO:2465

- the container runtime defers all outgoing message logic to the outbox
- batch managers don't do compression anymore
- compression (and other batch processors) are part of the outbox
- some test coverage

This is in preparation for chunking compressed but large batches, which would be another `processor` which would chunk the first op, send the chunks immediately as individual 'batches', then send the last chunk + empty ops as part of a single batch.

PR #12592, opened 2022-10-19
Simplify batching logic in PendingStateManager

kian-thompson:
# [AB#1833](https://dev.azure.com/fluidframework/internal/_workitems/edit/1833)

## Description

Batch tracking inside the `PendingStateManager` will now be done through the batch metadata on ops to indicate starts and ends of batches.

Before, we were using `flush` calls to indicate a batch had ended because the `batch: false` metadata would not be updated in messages in the `PendingStateManager`. Due to changes in the batching workflow in `ContainerRuntime`, the `batch: false` metadata is available.

## Does this introduce a breaking change?

`IPendingFlush` has been deprecated. Use batch metadata on `IPendingMessage` instead to indicate the end of a batch.

PR #27885, opened 2026-08-07
feat(container-runtime): add internal version mark resolver

sonalideshpandemsft:
This PR applies feedback from previous PRs 

Key changes include:

• Preserving reconnect-stable batch IDs when capturing resubmitted multi-op batches.
• Renaming  referenceSequenceNumber  to  sequenceNumberLowerBound  to clarify its role in historical resolution.
• Replacing direct construction of  OdspPointInTimeDocumentServiceFactory  with  getOdspPointInTimeDocumentServiceFactory  and an interface.
• Expanding unit coverage and design documentation across the runtime, loader, and ODSP driver.

These changes keep mark persistence and locator shape application-owned while Fluid provides the runtime resolver and read-only loading of a document at an exact sequence number.


anthony-murphy:

- 🔦 **Mechanical Review** — testing the lights…
- ✨ **Solution / Problem Space Review** — tracking bioluminescence…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1879,6 +1891,59 @@ export class ContainerRuntime
...
+				namespace: "VersionMarkResolver",
+			}),
+			// Seal the current outbound batch so a just-submitted edit gets a stable batchId in the pending
+			// state before captureVersionMark reads it (a batchId is only assigned when a batch is flushed).
+			flushPendingBatch: () => this.flush(),
+			// Wire the container-provided op reader (if any) so resolution can read historical ops.
+			getHistoricalOpReader: fetchOps ? () => ({ fetchMessages: fetchOps }) : undefined,
+			// Unpack scanned historical ops through the same pipeline the live inbound path uses, so a
+			// chunked batch's batchId (only restored after reassembly) is observed by the history scan too.
+			createHistoricalOpUnpacker: fetchOps
```
shlevari, line 1914: Is there ever a situation where fetchOps is undefined?

shlevari (approved):

anthony-murphy:
## Deep Review

*Reviewed commit `04bbfea` on 2026-08-07.*

Not ready for sign-off. The core architecture is strong — a runtime-owned, `@internal` version-mark resolver that converts app-stored pending `batchId` locators into durable sequence numbers, with mark storage kept app-owned. Two contained correctness bugs, both conceded crash paths, are flagged inline; each has a localized guard fix.

### Path to Ready

- [ ] Resolve inline threads

### Context for Reviewers

- Version-mark tracking gates per-inbound-batch work behind `isTracking`, mirroring the unused-feature gating `DuplicateBatchDetector` established in #22497; markfields (the requested reviewer) owns both subsystems.
- `Container.fetchOps` calls `connectToDeltaStorage()` on every invocation so it uses the current service rather than a stale handle, matching the connection-scoped-state care in `submitBatch`'s history (#14150).
- Flush-before-capture ordering is load-bearing: the resolver flushes before reading the current sequence number and pending batch id so a stale/undefined batch id is never paired with a mismatched reference sequence number. Scrutinize any change that splits capture into separate getters or moves the sequence-number read before the flush.
- Cross-session resolution deliberately uses an out-of-session historical delta-storage scan (`fetchOps` + `IHistoricalOpReader` through the live unpack pipeline) rather than persisting the tracker into the summary — a design divergence by intent that avoids summary bloat and honors the app-owned-storage constraint.

- **Needs human judgment** — markfields sign-off on the `@internal` incubation strategy and the version-mark tracking/gating design, which mirrors the #22497 `DuplicateBatchDetector` pattern he owns.
- **Needs human judgment** — whether the interim ODSP-driver-dependent empty-read → `unresolvable` classification is acceptable as a documented limitation, or whether the terminal `unresolvable` verdict should wait for the planned contractual op-availability/retention API. This is a risk-tolerance call, not a mechanical fix.
- **Cannot be assessed by the pipeline** — cross-fork / driver determinism of the Delta Storage read semantics underpinning the historical scan; the pipeline cannot verify runtime storage-consistency behavior across drivers.

</details>

packages/runtime/container-runtime/src/containerRuntime.ts
anthony-murphy: Deep Review: Good question. The absent-reader case is a designed fallback — when no `IHistoricalOpReader` is available, `fetchOps` is `undefined` and resolution returns `pending` rather than a terminal verdict, as documented in `DEV.md`. A maintainer should confirm this fallback satisfies the concern, but it is not an unhandled path.

packages/runtime/container-runtime/src/pendingStateManager.ts
```diff
@@ -366,6 +366,24 @@ export class PendingStateManager implements IDisposable {
...
+		const lastPendingMessage = this.pendingMessages.peekBack();
+		if (lastPendingMessage === undefined) {
+			return undefined;
+		}
+		const batchStart = this.pendingMessages.get(
+			this.pendingMessages.length - lastPendingMessage.batchInfo.length,
+		);
+		assert(batchStart !== undefined, "pending batch start message is missing");
+		return getEffectiveBatchId(batchStart);
+	}
```
anthony-murphy, line 385: Deep Review: `getMostRecentPendingBatchId()` can assert/crash on back-compat stashed pending ops.

The batch-start index is computed as `this.pendingMessages.length - lastPendingMessage.batchInfo.length` (lines 380-382). Back-compat stashed ops predating the `batchInfo` format are patched with `length: -1` (`message.batchInfo = { clientId: uuid(), batchStartCsn: -1, length: -1, staged: false }`, lines 1044-1047) and pushed into `pendingMessages` during `applyStashedOp` (lines 577-578, 591-592). If such a message is the `peekBack()` when capture runs, the index becomes `length - (-1) = length + 1`, out of bounds → `get()` returns `undefined` → `assert(batchStart !== undefined, "pending batch start message is missing")` (line 383) fires, throwing out of `captureVersionMark`.

The `@remarks` claim it "Ignores `initialMessages`" (lines 372-373), but that guarantee no longer holds once stashed ops are patched into `pendingMessages`. `applyStashedOpsAt` runs during runtime load before the runtime is returned (`containerRuntime.ts:1310-1314`), and existing tests (`pendingStateManager.spec.ts:986-990`) cover only the pre-apply state, not the post-apply legacy `length: -1` state.

Recommendation: guard `batchInfo.length <= 0` before computing the index — walk backward past legacy `length === -1` entries and return `undefined` when no current-session flushed batch remains, treating legacy/unknown-length pending batches as non-capturable. Add a test covering the post-apply legacy stashed-op state.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -0,0 +1,329 @@
...
+					firstScannedSequenceNumber ??= op.sequenceNumber;
+					if (!inBatch && asBatchMetadata(op.metadata)?.batch === false) {
+						// The anchor (`referenceSequenceNumber + 1`) may land inside a batch whose start
+						// precedes the window; its orphan end marker would trip the processor's 0x9d5 assert.
+						// Drop the clipped tail. The target batch is fully in-window, so this can't skip it.
+						continue;
+					}
+					// undefined = a system op, or an incomplete chunk awaiting later fragments.
+					const inboundResult = unpack?.(op);
+					if (inboundResult === undefined) {
```
anthony-murphy, line 228: Deep Review: `resolveFromHistory` can throw `"Chunk Id mismatch"` when the scan anchor lands mid-chunked-op.

The scan starts at `referenceSequenceNumber + 1` (line 192) and only pre-filters orphan `batch:false` markers before calling `unpack?.(op)` (lines 206-216). The container constructs a fresh `RemoteMessageProcessor` with a fresh `OpSplitter` for each scan (`containerRuntime.ts:1912-1924`). `OpSplitter.addChunk` throws `DataCorruptionError("Chunk Id mismatch")` when the incoming chunk id is not the next expected chunk (`opSplitter.ts:90-99`). The scan anchor has no guarantee of landing on a chunk boundary, so it can land after chunk 1 of an unrelated in-flight chunked op, and `unpack` throws before the scan reaches the target batch — aborting resolution.

The inline comment at line 214 addresses only "an incomplete chunk awaiting later fragments," not a leading/clipped chunk with `chunkId !== 1`. PR tests cover complete chunk reassembly (`versionMarkResolver.spec.ts:380-424`) and clipped non-chunk batches (`versionMarkResolver.spec.ts:427-510`), but not a scan beginning mid-chunk.

Recommendation: before `unpack(op)`, detect a chunked op when the scan has no chunk state and skip the remainder of that clipped chunk sequence, or reset the historical unpacker so mid-stream chunks neither throw nor abort resolving later batches. Add a test for a scan anchor that starts mid-chunk.

anthony-murphy:
## Deep Review

*Reviewed commit `4755ce2` on 2026-08-09.*

Not ready for sign-off. The core architecture is strong — a runtime-owned, `@internal` version-mark resolver that converts app-stored pending `batchId` locators into durable sequence numbers while keeping mark storage app-owned. One contained correctness bug remains: `getMostRecentPendingBatchId` asserts/crashes on back-compat stashed ops carrying the `batchInfo.length === -1` sentinel (flagged inline). The earlier mid-chunk history-scan concern is now a self-documented known limitation on the non-graduated `@legacy @alpha` surface, not a blocker.

### Path to Ready

- [ ] Resolve inline threads
- [ ] File a tracking issue for the mid-chunk history-scan limitation and confirm the `@internal`/`@legacy @alpha` non-graduation gating stays until it is closed

### Context for Reviewers

- Version-mark tracking gates per-inbound-batch work behind `isTracking`, mirroring the unused-feature gating `DuplicateBatchDetector` established in #22497; markfields (the requested reviewer) owns both subsystems.
- The `getMostRecentPendingBatchId` rewrite fixes a real bug, not a refactor: the old `getEffectiveBatchId(peekBack())` read a batch's last message and returned the wrong id whenever a batch carried an explicit stamped `batchId` after resubmit/reconnect. The change navigates to the batch-start message.
- The `referenceSequenceNumber` → `sequenceNumberLowerBound` rename disambiguates from the protocol property; `VersionMarkCapture` is a transient in-memory result, not a persisted locator, so no migration is required.
- Cross-session resolution deliberately uses an out-of-session historical delta-storage scan (`fetchOps` + `IHistoricalOpReader`) rather than persisting the tracker into the summary, to avoid summary bloat and honor the app-owned-storage constraint; `Container.fetchOps` calls `connectToDeltaStorage()` on every invocation to avoid a stale handle.
- Root cause and fix direction were independently triple-confirmed: three solution-blind proposers reproduced the exact `batchId`-on-start-message defect and prescribed reading the batch-start message plus the lower-bound rename.

- **Needs human judgment** — markfields sign-off on the `@internal` incubation strategy and the version-mark tracking/gating design, which mirrors the #22497 `DuplicateBatchDetector` pattern he owns. Consider tagging kian-thompson (authored the `batchInfo`/`length: -1` sentinel, #12592) or andre4i/vladsud (batch/chunk runtime semantics, #12938/#14150) for the `getMostRecentPendingBatchId` change.
- **Needs human judgment** — whether the interim mid-chunk history-scan limitation and any ODSP-driver-dependent empty-read classification are acceptable as documented known limitations on the non-graduated alpha surface, or whether terminal resolution semantics should wait for a planned contractual op-availability/retention API. This is a risk-tolerance call, not a mechanical fix.
- **Cannot be assessed by the pipeline** — cross-fork / driver determinism of the Delta Storage read semantics underpinning the historical scan; the pipeline cannot verify runtime storage-consistency behavior across drivers.

</details>

- `04bbfea` 2026-08-07 · **5/10** — two conceded crash paths flagged inline

</details>

packages/runtime/container-runtime/src/test/versionMarks/versionMarkResolver.spec.ts
```diff
@@ -545,15 +545,15 @@ describe("VersionMarkResolver", () => {
...
+			// tip (5) == sequenceNumberLowerBound, so `from` (6) is beyond the tip: the batch cannot have landed.
 			const reader = makeReader([]);
 			const resolver = makeResolver({ reader, currentSequenceNumber: 5 });
 			assert.deepEqual(await resolver.resolve(generateBatchId("missing", 9), 5), {
 				kind: "pending",
 			});
 		});
 
-		it("reads from referenceSequenceNumber + 1 (the exclusive lower-bound anchor)", async () => {
+		it("reads from sequenceNumberLowerBound + 1 (the exclusive lower-bound anchor)", async () => {
```
markfields, line 556: All else being equal, I would move the "+1" out of here.  i.e. let the "lower bound" be the first possible value of the sequenceNumber, thus setting it to `refSeq+1`.  In other words, based on the name it feels more natural to be an _inclusive_ lower bound not _exclusive_.

It's ok to leave it as-is for time's sake.

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -242,37 +247,78 @@ This makes cleanup proportional to the number of entries actually evicted (amort
...
+11. **Sequenced-before-persisted restore:** Resolve a mark from the live map as soon as its batch sequences, then immediately start a point-in-time load before ODSP has flushed that op to durable delta storage. Define whether the host waits, explicitly requests an op flush, or retries later, and verify a resolved locator never implies that the target is already materializable by a storage-only loader.
+
+### Resolver correctness and robustness gaps
+
+1. **Chunk-stream scan anchors:** The history scan can begin after one or more chunks of a message have already been trimmed from the requested range. `RemoteMessageProcessor` cannot currently reconstruct that incomplete leading chunk stream. Either capture a lower bound that is known to follow the last fully reconstructed runtime batch, or teach historical resolution to discard an incomplete leading chunk stream before processing the next complete batch. Cover anchors on every intermediate chunk, including a target batch immediately after the clipped stream.
```
markfields, line 279: Let's open a DF-blocking bug for this. This is a core functional gap: if `sequenceNumberLowerBound + 1` lands after the first chunk of an existing chunk stream, the fresh `OpSplitter` receives chunk 2+ with no prior state and throws `DataCorruptionError("Chunk Id mismatch")`. `resolve()` then rejects instead of resolving the target batch or returning a defined `pending`/`unresolvable` result. The bug should cover skipping the incomplete leading chunk stream and add tests for every intermediate-chunk anchor, including a target batch immediately afterward.

markfields (approved):

packages/runtime/container-runtime/src/test/versionMarks/versionMarkResolver.spec.ts
sonalideshpandemsft: https://dev.azure.com/fluidframework/internal/_workitems/edit/80024

packages/runtime/container-runtime/src/versionMarks/DEV.md
sonalideshpandemsft: https://dev.azure.com/fluidframework/internal/_workitems/edit/80025

PR #27886, opened 2026-08-07
fix(build-cli): resolve the SDK-provided Copilot CLI

steffenloesch:
## Description

`flub ai` cannot start its Copilot SDK session when dependencies are installed with pnpm. The SDK's default bundled-CLI resolver constructs an invalid path for pnpm's scoped package layout:

`.../node_modules/@github/index.js`

This change resolves `@github/copilot/npm-loader.js` relative to the installed `@github/copilot-sdk` module and passes that entry point explicitly to `CopilotClient`. This keeps CLI-version ownership with the SDK and avoids adding a separate direct Copilot dependency.

A behavioral test launches the resolved entry point with `--version` and verifies that it is a runnable GitHub Copilot CLI.

This fix unblocks the held Codespace integration in [PR #27869](https://github.com/microsoft/FluidFramework/pull/27869) and must be included in build-tools 0.67.0.

### Why not use Agency's Copilot CLI?

Agency manages its own Copilot CLI version, but it does not expose a supported machine-readable command for obtaining the resolved executable path. Its versioned `~/.copilot-cli/<version>` cache is an internal implementation detail without a stable symlink or public resolver contract.

Using `agency copilot` itself as the SDK subprocess is also incompatible. Agency injects `--session-id`, while `CopilotClient` starts the CLI with `--headless`; Copilot CLI rejects that combination because SDK headless mode manages sessions through JSON-RPC.

Resolving the SDK-owned dependency therefore provides the stable compatibility boundary: the SDK selects its compatible Copilot CLI version, and build-cli launches that executable directly.

### Manual Validation

```
cd /workspaces/FluidFramework/build-tools
pnpm install --frozen-lockfile
pnpm build:fast

node packages/build-cli/bin/dev ai
```

## Reviewer Guidance

The important dependency-resolution detail is that the resolver is rooted at the SDK module using `createRequire`. It therefore selects the Copilot CLI installed for that SDK even when pnpm isolates scoped dependencies.

### DIFF
--- a/build-tools/packages/build-cli/src/library/ai/copilotSession.ts
+++ b/build-tools/packages/build-cli/src/library/ai/copilotSession.ts
@@ -150,7 +150,8 @@ export async function runAiSession(
 }
 
 export function resolveCopilotCliPath(): string {
-	// The SDK's default resolver constructs an invalid path for pnpm's scoped package layout.
+	// Resolve from the SDK's dependency context so pnpm selects its nested Copilot CLI.
+	// import.meta.resolve's parentURL overload is experimental, so use createRequire here.
 	const requireFromSdk = createRequire(import.meta.resolve("@github/copilot-sdk"));
 	return requireFromSdk.resolve("@github/copilot/npm-loader.js");
 }
--- a/build-tools/packages/build-cli/src/library/ai/copilotSession.ts
+++ b/build-tools/packages/build-cli/src/library/ai/copilotSession.ts
@@ -3,6 +3,7 @@
  * Licensed under the MIT License.
  */
 
+import { createRequire } from "node:module";
 import { approveAll, CopilotClient, defineTool } from "@github/copilot-sdk";
 
 /**
@@ -86,6 +87,7 @@ export async function runAiSession(
 
 	const client = new CopilotClient({
 		...(githubToken !== undefined ? { githubToken } : {}),
+		cliPath: resolveCopilotCliPath(),
 		// Suppress Node's "ExperimentalWarning: SQLite is an experimental feature"
 		// that the bundled Copilot CLI subprocess emits on stderr.
 		env: {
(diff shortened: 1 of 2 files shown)


build-tools/packages/build-cli/src/library/ai/copilotSession.ts
```diff
@@ -3,6 +3,7 @@
  * Licensed under the MIT License.
  */
 
+import { createRequire } from "node:module";
```
tylerbutler, line 6: Are you sure require is needed here? Can't you do an ESM import?

tylerbutler:
It's worth noting that this only affects the copilot session that is used byt the `flub ai` command. The standard coding CLI sessions are still going through agency which is important for MCP config etc.

build-tools/packages/build-cli/src/library/ai/copilotSession.ts
steffenloesch: I'm not super familiar with the resolution details. Copilot thinks we need `createRequire` because we need to set the parent url, and doing that with import is experimental in our current node version:

> We need the resolved filesystem path rather than importing the loader, and resolution must be rooted at  @github/copilot-sdk  so pnpm selects that SDK’s nested  @github/copilot  dependency. The single-argument  import.meta.resolve()  would resolve from build-cli instead. Its  parentURL  overload could provide the required context, but that overload remains experimental and requires  --experimental-import-meta-resolve , including in current Node versions.  createRequire()  gives us the equivalent contextual resolution using a stable Node API.

tylerbutler: Hmmmm, OK. Worth a comment so someone does not try to "fix" it. :)

steffenloesch: Added a comment explaining both the pnpm dependency context and why `createRequire` is intentional.

tylerbutler (approved):

PR #27889, opened 2026-08-07
Clarify tree schema type branding

CraigMacomber:
## Description

Clarifications around `typeNameSymbol` brand. Turns out we can't just remove it, so this clarifies why.

A future change (for 3.0) will make breaking API changes to further clean this up (stop non-type exporting the symbol, rename it, remove the deprecations)

## Reviewer Guidance

### DIFF
--- a/packages/dds/tree/src/simple-tree/core/treeNode.ts
+++ b/packages/dds/tree/src/simple-tree/core/treeNode.ts
@@ -9,8 +9,8 @@ import { UsageError } from "@fluidframework/telemetry-utils/internal";
 import { markEager } from "./flexList.js";
 import { tryGetTreeNodeSchema } from "./treeNodeKernel.js";
 import { NodeKind, type TreeNodeSchemaClass } from "./treeNodeSchema.js";
-// eslint-disable-next-line import-x/no-deprecated
-import { type WithType, typeNameSymbol, type typeSchemaSymbol } from "./withType.js";
+// eslint-disable-next-line import-x/no-deprecated -- Required to implement the deprecated compatibility member.
+import type { WithType, typeNameSymbol, typeSchemaSymbol } from "./withType.js";
 
 /**
  * A non-{@link NodeKind.Leaf|leaf} SharedTree node. Includes objects, arrays, and maps.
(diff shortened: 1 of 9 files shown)


packages/dds/tree/src/simple-tree/core/treeNode.ts
```diff
@@ -9,8 +9,8 @@ import { UsageError } from "@fluidframework/telemetry-utils/internal";
 import { markEager } from "./flexList.js";
 import { tryGetTreeNodeSchema } from "./treeNodeKernel.js";
 import { NodeKind, type TreeNodeSchemaClass } from "./treeNodeSchema.js";
-// eslint-disable-next-line import-x/no-deprecated
-import { type WithType, typeNameSymbol, type typeSchemaSymbol } from "./withType.js";
+// eslint-disable-next-line import-x/no-deprecated -- Required to implement the deprecated compatibility member.
```
Josmithr: Nit: probably worth calling out exactly what import this is referring too, just in case other of these imports become deprecated in the future. Unlikely to be a problem, but probably a good general practice to follow IMO (at least when the comment appears over an entire import block, rather than an individual imported member).

I'm assuming our formatter would complain if you tried to line-wrap and put this on the individual item?

packages/dds/tree/src/simple-tree/core/withType.ts
```diff
@@ -96,6 +94,11 @@ export const contentSchemaSymbol: unique symbol = Symbol("SharedTree Schema");
  * 	}
  * }
  * ```
+ * @privateRemarks
+ * The properties of this are implemented as getters rather than instance properties as that is the easiest way to make them non-enumerable,
+ * and also saves memory and time as they only have to be setup on the prototype.
+ * We want these to be non-enumerable as they should not be shown to users as they are internal typing brands.
```
Josmithr: Nit
```suggestion
 * We want these to be non-enumerable, and they should not be shown to users as they are internal typing brands.
```

Josmithr (approved):

packages/dds/tree/src/simple-tree/core/withType.ts
CraigMacomber: I think is is correct as is. The reason we want them to be non enumerable is to surface them to users less (for example when users enumerate properties using one of the APIs that includes enumerable symbol properties). I have rewritten this to be more clear.

packages/dds/tree/src/simple-tree/core/treeNode.ts
CraigMacomber: It actually is allowed. I have updated them to use this pattern as well as have a more specific message.

Josmithr (approved):

PR #27892, opened 2026-08-07
fix(ci): disable compatibility installs on LTS

markfields:
## Description

Keep LTS CI and PR builds network-isolated while preserving the existing non-compatibility tests.

The LTS test package declares version `1.4.0`, but the repository root/release-group version is `0.14.0`. CI generates build versions such as `0.14.0-<build>`, so the test scripts' hard-coded `--baseVersion=1.4.0` caused the non-compatibility path to dynamically install the published `1.4.0` packages.

This change:

- Removes the hard-coded `--baseVersion=1.4.0` from the local and Tinylicious real-service test scripts, allowing the harness to use the current CI build version by default.
- Sets `fluid__test__compatVersion` to `["0"]` for regular and coverage test invocations. `0` is the existing non-compatibility configuration; this excludes N-1, N-2, and LTS compatibility configurations while leaving current-version tests enabled.

The compatibility harness remains unchanged for nonzero compatibility versions; those paths still install and test historical versions when explicitly selected outside this LTS CI configuration.

## Reviewer Guidance

This PR targets `lts`. That branch predates the parallel `include-test-task.yml` topology, so its shared build template contains only the regular and coverage invocations updated here.


packages/test/test-end-to-end-tests/package.json
```diff
@@ -28,7 +28,7 @@
     "test:realsvc": "npm run test:realsvc:local && npm run test:realsvc:tinylicious",
     "test:realsvc:frs": "npm run test:realsvc:run -- --driver=r11s --r11sEndpointName=frs --timeout=20s",
     "test:realsvc:frs:report": "cross-env FLUID_TEST_REPORT=1 npm run test:realsvc:frs --",
-    "test:realsvc:local": "npm run test:realsvc:run -- --driver=local --baseVersion=1.4.0",
```
markfields, line 31: This is needed because the version used in the pipeline is 0.14.0-xxxxxx, and we just want to test against the already-installed bits, so if we specify this then it goes and downloads it.

alexvy86: 0.14.0-xxxxx is the version of what, that gets installed? Is that a typo for 1.4.0-xxxxxx? Otherwise I'm confused :)

markfields: Haha I was too.  It's not a typo.  From Copilot:

The pipeline treats the repository as one release group. include-set-package-version.yml runs:

Then update-package-version.sh runs flub bump ... for the release group. That makes the CI workspace use the root/release-group build version:

So:

1.4.0 is the declared/published LTS package version.
0.14.0 is the private root/release-group version used to generate CI build versions.
0.14.0-416026 is not a prerelease of 1.4.0; it is a build version derived from the root’s 0.14.0 baseline.
That also explains why the historical --baseVersion=1.4.0 mattered: it explicitly anchored compatibility tests to the published LTS package version, rather than the synthetic CI workspace version. Removing it changes that behavior to test against the generated CI build version.

tools/pipelines/templates/build-npm-package.yml
```diff
@@ -342,6 +342,10 @@ extends:
           - ${{ if ne(variables['testCoverage'], true) }}:
             - task: Npm@1
               displayName: npm run ${{ parameters.taskTest }}
+              env:
+                # Avoid dynamic npm installs of old package versions during network-isolated CI/PR runs.
+                # This keeps current-version tests running while skipping cross-version compat tests.
```
alexvy86, line 347: I think a bit more context about why it's ok to deliberately forego compat testing in the LTS branch would be useful. Worst case scenario is that we have to put out a patch for LTS and inadvertently break its collab with an older version, but it's unlikely we'll have to do it, and it's unlikely that the kind of patch we would have to put out would compromise collab?

markfields: I don't think we care about collab with older versions. Those older versions aren't supported in any way. Right?

alexvy86 (approved):
Chatted with Mark offline, seems fine to merge this and we'll discuss more in Teams if necessary.

package.json
```diff
@@ -53,8 +53,8 @@
     "test:mocha": "pnpm run -r --no-sort --stream --no-bail test:mocha --color",
     "test:mocha:bail": "pnpm run -r --no-sort --stream test:mocha",
     "test:mocha:report": "pnpm run -r --no-sort --stream --no-bail test:mocha:multireport -- --timeout 4s",
-    "test:realsvc": "pnpm run -r --stream --no-bail --no-sort test:realsvc",
-    "test:realsvc:report": "pnpm run -r --stream --no-bail --no-sort test:realsvc:report",
+    "test:realsvc": "pnpm run -r --workspace-concurrency=1 --stream --no-bail --no-sort test:realsvc",
```
markfields, line 56: Tinylicious was timing out because the e2e tests were running at the same time as the stress tests.

Before this change, local e2e tests ran in parallel with t9s stress, but since we're not running compat anymore they finished before the stress tests did and the e2e tests against t9s were getting starved.

tools/pipelines/templates/build-npm-package.yml
markfields: I added to this comment explaining this.  I'll also get feedback from the team if this is ok, and if it's a problem we have an alternative PR here that we can pursue:  #27931

steffenloesch (approved):

PR #27893, opened 2026-08-07
fix(ci): keep pnpm installs on authenticated registries

markfields:
## Description

Prevent the network-isolated build-tools and client pipelines from contacting `registry.npmjs.org`.

The shared pnpm setup continues to create and authenticate a temporary user-level `.npmrc` for the configured Azure Artifacts registry. It now:

- passes that userconfig explicitly while installing pnpm;
- fails fast unless both npm and pnpm resolve the expected registry;
- publishes the existing CI mirror workarounds (`trustPolicy=off` and `minimumReleaseAge=0`) as job variables; and
- sets `trustLockfile=true` for subsequent CI installs so pnpm 11 does not query registry metadata to reapply those policies to every dependency already represented by a committed lockfile.

`trustLockfile` does not affect lockfile creation or updates, frozen-lockfile consistency, structural validation, or downloaded tarball integrity. The setting is emitted only by the ADO template, so local dependency-update workflows retain their existing behavior.

Shared build, lint, and test scripts now run through `Bash@3` instead of `Npm@1`. `Npm@1` replaces the user-level npm configuration, which allowed nested registry operations such as compatibility-test `pnpm view` calls to fall back to npmjs. Bash preserves the authenticated job-level userconfig. The npmrc setup documentation now calls out this behavior for scripts that directly or indirectly perform registry operations.

## Reviewer Guidance

Please focus on:

- registry verification in `include-install-pnpm.yml`;
- the CI-only `trustLockfile` tradeoff described above; and
- the intentional use of Bash for build and test scripts that may perform nested registry operations.

### Validation

Build Tools and Client Packages pipelines all show CFS Client COMPLIANT ✅

Prior validation done with forced cold pnpm cache, will be done again after review signoff, before merging:

- Internal cold-cache Build - build-tools run 415527: cache miss and CFSClean compliant.
- Internal cold-cache Build - client packages run 415518: cache miss and CFSClean compliant. The unrelated RealsvcTinyliciousTest job failed; Build and Coverage tests succeeded.


tools/pipelines/templates/include-install-pnpm.yml
```diff
@@ -120,19 +120,36 @@ steps:
...
   displayName: Install pnpm
+  env:
+    NPM_CONFIG_USERCONFIG: ${{ parameters.userNpmrcPath }}
+    NPM_REGISTRY: ${{ parameters.primaryRegistry }}
   inputs:
     targetType: 'inline'
     workingDirectory: '$(Pipeline.Workspace)/${{ parameters.buildDirectory }}'
     script: |
       set -eu -o pipefail
+      expected_registry="${NPM_REGISTRY%/}/"
```
markfields, line 131: I wonder if NPM_REGISTRY var can be inlined here I think it's the only usage

markfields: Copilot says keep it:

> Inlining the template parameter would put an ADO runtime value directly into generated shell source, creating quoting or injection risk. Environment-variable passing treats it strictly as data.

examples/apps/claims-example/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@fluid-example/claims-example",
-	"version": "2.115.0",
+	"version": "2.116.0",
```
markfields, line 3: (Cherry-picked from #27916)

Abe27342: fyi - merged main into your branch to remove this from the diff :P

tools/pipelines/templates/include-install-pnpm.yml
Abe27342: copilot is correct :)

Abe27342 (approved):
approving to unblock--seems good. I have some potential suggestions I'll follow up with outside of PR discussion

PR #27896, opened 2026-08-09
improvement(client-container-runtime): cleanup from internal ContainerRuntime

jason-ha:
`ContainerRuntime` is still exported but is now only `@internal` instead of `@legacy` and it may implement `IFluidRootParentContextPrivate`.


jason-ha:
It is too clumsy to deal with the additional api-extractor export requirements while ContainerRuntime is exported. (The original comments were decent.)

PR #27902, opened 2026-08-10
Add ServiceClient session and tinylicious support, as well as example utils, and use for inventory-app

CraigMacomber:
## Description

Adds the foundations for examples to use the `ServiceClient` API without the legacy loader host:

- Adds an alpha session-storage-backed local service.
- Adds an alpha Tinylicious service client, including forwarded HTTPS endpoint support.
- Adds loader-free shared webpack and example startup/rendering helpers.
- Migrates the inventory app to the shared helpers with ephemeral, session, and Tinylicious modes.
- Adds focused tests, generated API reports, and changesets for the new driver APIs.

These additions are non-breaking and retain the browser `process` polyfill required by existing dependencies.

## Reviewer Guidance


packages/drivers/tinylicious-driver/src/insecureTinyliciousUrlResolver.ts
```diff
@@ -31,8 +33,12 @@ export const defaultTinyliciousEndpoint = "http://localhost";
  */
 export class InsecureTinyliciousUrlResolver implements IUrlResolver {
 	private readonly tinyliciousEndpoint: string;
-	public constructor(port = defaultTinyliciousPort, endpoint = defaultTinyliciousEndpoint) {
-		this.tinyliciousEndpoint = `${endpoint}:${port}`;
+	public constructor(port?: number, endpoint?: string) {
+		const endpointUrl = new URL(endpoint ?? defaultTinyliciousUrl);
+		if (port !== undefined) {
```
CraigMacomber: There was what I consider a bug in this code before, where if you provided no port, but the URL included one, it would put the default port on the end, which gives a malformed URL.

This new approach makes more sense in my opinion, where we allow URLs which contain ports, but can optionally override those.

Tests are included to ensure this works as expected.

packages/common/driver-definitions/src/serviceClient.ts
```diff
@@ -185,9 +185,8 @@ export interface ServiceOptions {
 	 * Choosing an older version may limit the features and write formats the application can use to
 	 * those supported by that version.
 	 *
-	 * A service may provide a default when this option is omitted.
 	 */
-	readonly oldestSupportedClient?: OldestSupportedServiceClientVersion;
+	readonly oldestSupportedClient: OldestSupportedServiceClientVersion;
 }
```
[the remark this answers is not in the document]
CraigMacomber: This fixes a regression, and is an alpha API so its fine. ALso only services added in this require it: the preexisting one did not.

examples/data-objects/inventory-app/src/inventoryList.ts
```diff
@@ -3,14 +3,14 @@
...
-// eslint-disable-next-line import-x/no-internal-modules
-import { treeDataObjectInternal } from "@fluidframework/react/internal";
+import { defineTreeDataStore } from "fluid-framework/alpha";
 
 import { Inventory, treeConfiguration } from "./schema.js";
 
-export const InventoryListFactory = treeDataObjectInternal(
-	treeConfiguration,
-	() =>
+export const inventoryDataStoreKind = defineTreeDataStore({
```
Josmithr: Not for this PR, but I continue to find the "Kind" terminology confusing. I wonder if we should migrate to something like "Definition" instead.

PACKAGES.md
```diff
@@ -171,25 +171,25 @@ The dependencies between layers are enforced by the layer-check command._
...
+| - [@fluidframework/routerlicious-driver](/packages/drivers/routerlicious-driver)</br>- [@fluidframework/routerlicious-urlresolver](/packages/drivers/routerlicious-urlResolver)</br>&nbsp;</br>&nbsp;</br>&nbsp;</br>&nbsp;</br>&nbsp;</br>&nbsp; | - [Core-Interfaces](#Core-Interfaces)</br>- [Driver-Definitions](#Driver-Definitions)</br>- [Core-Utils](#Core-Utils)</br>- [Client-Utils](#Client-Utils)</br>- [Telemetry-Utils](#Telemetry-Utils)</br>- [Driver-Utils](#Driver-Utils)</br>- [Driver](#Driver)</br>- [Server-Shared-Utils](#Server-Shared-Utils) |
 
-### Test-Utils
+### ServiceClients
```
Josmithr, line 176: Is this re-ordering of sections expected?

examples/utils/example-utils/src/exampleApp.ts
```diff
@@ -0,0 +1,152 @@
...
+ * compatibility required by the local-driver services.
+ * Reads the `fluidClient` URL query parameter.
+ * Accepts `ephemeral`, `session`, or `tinylicious`:
+ * missing and unknown values default to the session service when session storage is available,
+ * or the ephemeral service otherwise.
+ *
+ * This is intended to be invoked once on startup:
+ * it may start a local service as a side-effect.
+ *
+ * When used in testing, and cleanup is required, use the ephemeral service,
```
Josmithr: Nit
```suggestion
 * When used in testing and cleanup is required, use the ephemeral service,
```

examples/utils/example-utils/src/exampleApp.ts
```diff
@@ -0,0 +1,152 @@
...
+ * This helper is designed to work with
+ * {@link @fluid-example/webpack-fluid-loader#exampleAppConfig}, which provides the browser
+ * compatibility required by the local-driver services.
+ * Reads the `fluidClient` URL query parameter.
+ * Accepts `ephemeral`, `session`, or `tinylicious`:
+ * missing and unknown values default to the session service when session storage is available,
+ * or the ephemeral service otherwise.
+ *
+ * This is intended to be invoked once on startup:
+ * it may start a local service as a side-effect.
```
Josmithr: Nit
```suggestion
 * This is intended to be invoked once on startup.
 * It may start a local service as a side-effect.
```

examples/utils/example-utils/src/exampleApp.ts
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * Configures a service client based on the `fluidClient` URL query parameter.
+ *
+ * @remarks
+ * This helper is designed to work with
+ * {@link @fluid-example/webpack-fluid-loader#exampleAppConfig}, which provides the browser
+ * compatibility required by the local-driver services.
+ * Reads the `fluidClient` URL query parameter.
+ * Accepts `ephemeral`, `session`, or `tinylicious`:
+ * missing and unknown values default to the session service when session storage is available,
```
Josmithr: Nit
```suggestion
 * Accepts `ephemeral`, `session`, or `tinylicious`.
 * Missing and unknown values default to the session service when session storage is available,
```

examples/utils/example-utils/src/exampleApp.ts
```diff
@@ -0,0 +1,152 @@
...
+ * @typeParam T - The type of the root data store.
+ * @param client - The service client used to load or create the container.
+ * @param rootStore - The kind of data store to use as the container root.
+ * @returns The loaded or newly attached container.
+ * @remarks
+ * This function assumes its the only thing using the `location.hash`: if your application uses it for anything else,
+ * it should replace use of this function with an appropriate alternative scheme.
+ * @internal
+ */
+export async function loadExampleContainer<T>(
```
Josmithr: Nit: I think elsewhere in examples we have tried to be explicit about differentiating "load" from "create". In cases where both are handled dynamically, I think we've generally tried to use naming like "createOrLoad...".

examples/utils/webpack-fluid-loader/src/baseConfig.ts
```diff
@@ -0,0 +1,138 @@
...
+ */
+
+import path from "path";
+
+import HtmlWebpackPlugin from "html-webpack-plugin";
+import type { Configuration as WebpackConfiguration } from "webpack";
+import type { Configuration as DevServerConfiguration } from "webpack-dev-server";
+
+/**
+ * Axios publishes its browser CommonJS bundle with a reference to `axios.cjs.map`, but does not
```
Josmithr: Nit: This documents why this is needed, but not what it is / how it should be consumed. I'd recommend adding that and moving these notes into a `@remarks` block.

examples/utils/webpack-fluid-loader/src/baseConfig.ts
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Creates a loader-agnostic webpack configuration suitable for an example application.
+ * @param baseDir - The application directory containing its source and output folders.
+ * @param env - Environment options for the webpack build.
+ * @param options - Options for generated assets and other base configuration behavior.
+ * @returns A webpack configuration for the example application.
+ * @internal
+ */
+export function baseExampleConfig(
```
Josmithr: Nit: our guidelines call to name functions using verb phrases. These should probably be "createBaseDevServerConfig" and "createBaseExampleConfig"

packages/drivers/local-driver/src/ephemeralService.ts
```diff
@@ -102,24 +130,91 @@ export function getDefaultEphemeralService(): EphemeralService {
...
+	deleteDocument(id: string): Promise<void>;
+
+	/**
+	 * Deletes all documents stored by this service.
+	 *
+	 * @remarks
+	 * Deletion is only allowed when this service has no open containers because resetting the local
+	 * server invalidates all of its active connections.
+	 */
+	deleteAllDocuments(): Promise<void>;
```
Josmithr, line 196: Nit: should we add `@throws` documentation to these?

packages/drivers/local-driver/src/ephemeralService.ts
```diff
@@ -102,24 +130,91 @@ export function getDefaultEphemeralService(): EphemeralService {
...
+
+	/**
+	 * Drives all containers connected to this service toward convergence, processing pending operations and
+	 * waiting for all dirty containers to save.
+	 *
+	 * @param timeoutMilliseconds - The maximum time to wait for containers to quiesce, in milliseconds. Defaults to 30_000.
```
Josmithr, line 202: What is expected to happen on timeout? Throw an error? No-op? We should document this explicitly.

packages/drivers/local-driver/src/ephemeralService.ts
```diff
@@ -159,106 +254,97 @@ export interface EphemeralServiceOptions extends ServiceOptions {
...
 {
 	// A single server is shared by all containers connected to this service so they can communicate with each other.
-	private readonly server: ILocalDeltaConnectionServer =
-		LocalDeltaConnectionServer.create(
-			// new LocalSessionStorageDbFactory(),
-		);
-	private readonly documentServiceFactory = new LocalDocumentServiceFactory(this.server);
+	private server: ILocalDeltaConnectionServer;
+	private documentServiceFactory: LocalDocumentServiceFactory;
+	private readonly databaseFactory: ITestDbFactory;
```
Josmithr, line 325: Docs?

packages/drivers/local-driver/src/ephemeralService.ts
```diff
@@ -279,6 +365,64 @@ class EphemeralServiceImplementation
...
+	private ensureAvailable(): void {
+		if (this.closed) {
+			throw new UsageError("Local service is closed");
+		}
+		if (this.maintenanceInProgress) {
+			throw new UsageError("Local service document maintenance is already in progress");
+		}
+	}
+
+	private async deleteDocuments(id?: string): Promise<void> {
```
Josmithr, line 397: Why is "documents" plural here? Is this just referring to the fact that a single "document" has data spread across multiple DB collections?

packages/drivers/tinylicious-driver/src/insecureTinyliciousUrlResolver.ts
```diff
@@ -31,8 +31,12 @@ export const defaultTinyliciousEndpoint = "http://localhost";
  */
 export class InsecureTinyliciousUrlResolver implements IUrlResolver {
 	private readonly tinyliciousEndpoint: string;
-	public constructor(port = defaultTinyliciousPort, endpoint = defaultTinyliciousEndpoint) {
-		this.tinyliciousEndpoint = `${endpoint}:${port}`;
+	public constructor(port?: number, endpoint?: string) {
```
Josmithr: Nit: param docs?

packages/drivers/tinylicious-driver/src/tinyliciousService.ts
```diff
@@ -0,0 +1,201 @@
...
+import { InsecureTinyliciousTokenProvider } from "./insecureTinyliciousTokenProvider.js";
+import {
+	createInsecureTinyliciousTestUrlResolver,
+	createTinyliciousCreateNewRequest,
+	InsecureTinyliciousUrlResolver,
+} from "./insecureTinyliciousUrlResolver.js";
+
+/**
+ * Options for configuring a {@link createTinyliciousServiceClient}.
+ * @alpha
```
Josmithr, line 45: `@input`?

packages/drivers/tinylicious-driver/src/tinyliciousService.ts
```diff
@@ -0,0 +1,201 @@
...
+		const aliasResult = await dataStore.trySetAlias(rootDataStoreId);
+		assert(aliasResult === "Success", "Should be able to set alias on new data store");
+	}
+	return runtime;
+};
+
+function makeContainerLoaderOptions(options: TinyliciousServiceOptions): {
+	urlResolver: IUrlResolver;
+	documentServiceFactory: RouterliciousDocumentServiceFactory;
+	clientDetailsOverride: { capabilities: { interactive: boolean } };
```
Josmithr: Nit: presumably there is a real type somewhere and we're just capturing a subset of it here? Would it be better to `Pick` it or define a new type or something? Two-tiered inline type with no docs seems worth some refactoring IMO

tools/markdown-magic/src/transforms/exampleAppReadmeHeader.ts
```diff
@@ -19,12 +19,14 @@ export const exampleAppReadmeHeaderTransform: Transform = transform(
 		...headingLevelSchema,
 		gettingStarted: { type: "boolean", default: true },
 		usesTinylicious: { type: "boolean", default: true },
+		serviceClient: { type: "boolean", default: false },
```
Josmithr, line 22: Would it be better to model these as orthogonal options? It looks like we treat them as such. Having a single optional "serviceMode" (or something) enum might be better.

This could be done in a follow-up to avoid creating lots of README churn.

PACKAGES.md
CraigMacomber: The reorder was an expected generated side effect.

This PR moves @fluidframework/tinylicious-driver from Routerlicious-Driver to ServiceClients in layerInfo.json:130-154. Because @fluid-private/test-drivers belongs to Test-Utils and depends on tinylicious-driver (package.json:48-65), this created:

Test-Utils → ServiceClients

The PACKAGES.md generator topologically orders dependencies before dependents (layerGraph.ts:517-553). Consequently, ServiceClients moved before Test-Utils in PACKAGES.md:170-188.

examples/data-objects/inventory-app/src/inventoryList.ts
CraigMacomber: Used `InventoryDataStore` as discussed.

packages/drivers/local-driver/src/ephemeralService.ts
CraigMacomber: added docs

tools/markdown-magic/src/transforms/exampleAppReadmeHeader.ts
CraigMacomber: Added a TODO over in tools/markdown-magic/src/transforms/exampleGettingStarted.ts

packages/drivers/tinylicious-driver/src/tinyliciousService.ts
CraigMacomber: Replaced with Pick.

examples/data-objects/inventory-app/src/inventoryList.ts
```diff
@@ -7,7 +7,7 @@ import { defineTreeDataStore } from "fluid-framework/alpha";
 
 import { Inventory, treeConfiguration } from "./schema.js";
 
-export const inventoryDataStoreKind = defineTreeDataStore({
+export const InventoryDataStore = defineTreeDataStore({
```
Josmithr, line 10: Nit: docs?

Josmithr (approved):

Josmithr (approved):

PR #27908, opened 2026-08-10
feat: stop implementing TypeScript built-in container types

### LINKED ISSUE #27459: Stop implementing TypeScript built in types like ReadonlyArray and ReadonlyMap in v3.0

jenn-le:
## Description

Resolves https://github.com/microsoft/FluidFramework/issues/27459.

Fluid's public map and array APIs currently declare conformance to TypeScript built-in container and iterator types. Changes to a consumer's configured TypeScript library can therefore add requirements that Fluid's implementations do not satisfy.

This change migrates the affected Tree, SharedMap, Presence, and container-loader APIs to Fluid-owned `FluidReadonlyArray`, `FluidReadonlyMap`, `FluidMap`, and `FluidIterableIterator` contracts. It also removes the now-redundant `IDirectoryBeta` interface. Runtime behavior is unchanged.

## Breaking Changes

This changes public TypeScript declarations and is intended for the Fluid Framework 3.0 release. See [Breaking-vs-Non-breaking-Changes](https://github.com/microsoft/FluidFramework/wiki/Breaking-vs-Non-breaking-Changes) for details.

Most existing assignments remain structurally compatible. Code that calls methods available only on newer built-in `Array` or `Map` types should first copy the Fluid container into a native container, for example:

```typescript
const reversed = [...arrayNode].toReversed();
```

Consumers using `IDirectoryBeta` should migrate to `IDirectory`, which now provides the Fluid-owned map contract.

## Reviewer Guidance

Please focus on completeness of the public container/iterator migration and the generated API report changes. This public breaking change also requires API Council approval.


.changeset/replace-builtin-map-array-types.md
```diff
@@ -0,0 +1,47 @@
...
+
+- `ReadonlyArrayNode` now extends `FluidReadonlyArray<T>` instead of `ReadonlyArray<T>`
+- `TreeMapNode` now extends `FluidReadonlyMap` instead of `ReadonlyMap`
+- `TreeArrayNode` and `TreeMapNode` iterator methods now return `FluidIterableIterator` instead of `IterableIterator`
+- `TreeRecordNode[Symbol.iterator]` and `TreeRecordNodeUnsafe[Symbol.iterator]` now return `FluidIterableIterator`
+- `IDirectory` and `ISharedMap` now extend `FluidMap` instead of `Map`
+- `IDirectoryBeta` has been removed. Use `IDirectory`, which now provides the Fluid-owned map contract
+- `StateMap.keys()` now returns `FluidIterableIterator` instead of `IterableIterator`
+- `PendingLocalStateStore` iteration methods now return `FluidIterableIterator` instead of built-in iterator types
+- BubbleBench's shared model contracts now use Fluid-owned iteration types and a minimal read-only collection shape so both Tree arrays and native arrays remain compatible
```
Josmithr, line 25: We should split this into separate changesets by affected package(s) IMO. That way the resulting changelogs are scoped appropriately to the affected packages.

It _might_ make sense to also do a separate one for `fluid-framework` that aggregates all of the DDS changes in one, rather than including it in all of those separate changesets.

examples/benchmarks/bubblebench/common/src/types.ts
```diff
@@ -3,6 +3,8 @@
  * Licensed under the MIT License.
  */
 
+import type { FluidIterable } from "@fluidframework/core-interfaces";
```
Josmithr, line 6: Nit: could we make the bubblebench changes in a separate PR? These presumably don't need to wait for the 3.0 breaking change window.

packages/dds/map/src/interfaces.ts
```diff
@@ -71,6 +74,18 @@ export interface IDirectory
 	 */
 	set<T = unknown>(key: string, value: T): this;
 
+	/**
+	 * Removes all entries from the directory.
+	 */
+	clear(): void;
+
+	/**
+	 * Removes the specified element from this IDirectory by its key.
```
Josmithr, line 83: Nit
```suggestion
	 * Removes the specified element from this directory by its key.
```

packages/dds/tree/src/test/simple-tree/api/typesUnsafe.spec.ts
```diff
@@ -20,7 +25,12 @@ import type {
 
 type MapInlined = System_Unsafe.ReadonlyMapInlined<string, typeof numberSchema>;
 
-type _check = requireTrue<areSafelyAssignable<MapInlined, ReadonlyMap<string, number>>>;
+type _check = requireAssignableTo<MapInlined, FluidReadonlyMap<string, number>>;
+
+type _recordNodeUnsafeIterator = requireAssignableTo<
```
Josmithr, line 30: Nit: probably worth a quick comment

Josmithr (commented):
Left a few minor suggestions. Overall looks good

examples/benchmarks/bubblebench/common/src/types.ts
jenn-le: Opened here: https://github.com/microsoft/FluidFramework/pull/27928

.changeset/replace-builtin-map-array-types.md
jenn-le: Split the combined entry into package-scoped changesets for Tree, Map, Presence, and container-loader, plus a separate fluid-framework aggregator changeset.

packages/dds/tree/src/test/simple-tree/api/typesUnsafe.spec.ts
jenn-le: Added a concise comment explaining that the assertion verifies unsafe record iteration uses the Fluid-owned iterator type.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
jason-ha:
```suggestion
	implements Pick<FluidReadonlyArray<T>, (typeof arrayPrototypeKeys)[number]>
```

jason-ha:
```suggestion
	every<S extends T>(
		predicate: (value: T, index: number, array: readonly T[]) => value is S,
		thisArg?: any,
	): this is FluidReadonlyArray<S>;
```

jason-ha (commented):
merge needs fixe - taking combination via VS Code merge editor should be correct.
Additionally, to stay more independent (compile okay against TS6) `NodeWithArrayFeatures` should be slightly altered as noted below.

packages/dds/tree/src/simple-tree/node-kinds/array/arrayNode.ts
jenn-le: Updated the type-guard overload to narrow to `FluidReadonlyArray<S>`.

jenn-le: Updated `NodeWithArrayFeatures` to implement the selected members from `FluidReadonlyArray<T>`.

.changeset/use-fluid-pending-state-iterator-types.md
```diff
@@ -0,0 +1,13 @@
+---
+"@fluidframework/container-loader": minor
+"__section": breaking
+---
+Pending local state uses Fluid-owned iterator types
+
+`PendingLocalStateStore` iteration methods now return `FluidIterableIterator` instead of TypeScript's built-in iterator types.
```
Josmithr: Nit: API docs links for the APIs referenced in the changesets would be useful.

jenn-le: Added links to the API documentation for both `PendingLocalStateStore` and `FluidIterableIterator` in the changeset.

.changeset/use-fluid-presence-iterator-types.md
```diff
@@ -0,0 +1,14 @@
+---
+"@fluid-internal/presence-definitions": minor
+"@fluidframework/presence": minor
+"__section": breaking
+---
+Presence maps use Fluid-owned iterator types
+
+`StateMap.keys()` now returns `FluidIterableIterator` instead of TypeScript's built-in `IterableIterator`.
```
Josmithr: Nit: API docs links?

.changeset/use-fluid-shared-map-types.md
```diff
@@ -0,0 +1,13 @@
+---
+"@fluidframework/map": minor
+"__section": breaking
+---
+Shared maps use Fluid-owned map types
+
+`IDirectory` and `ISharedMap` now extend the Fluid-owned `FluidMap` interface instead of TypeScript's built-in `Map` interface.
```
Josmithr: Nit: API docs links?

Josmithr (approved):

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -326,9 +346,11 @@ abstract class CustomMapNodeBase<const T extends ImplicitAllowedTypes> extends T
 	public has(key: string): boolean {
 		return this.innerNode.tryGetField(brand(key)) !== undefined;
 	}
-	public keys(): FluidIterableIterator<string> {
+	public *keys(): FluidIterableIterator<string> {
 		const node = this.innerNode;
-		return node.keys();
+		for (const key of node.keys()) {
+			yield key;
+		}
```
jason-ha: Was this more efficient before? What is the difference?
(I don't know what `yield* node.keys()` actually did.)

.changeset/use-fluid-shared-map-types.md
```diff
@@ -4,10 +4,22 @@
...
+[`IDirectoryBeta`](https://fluidframework.com/docs/api/map/idirectorybeta-interface) has been removed because `IDirectory` now provides the Fluid-owned map contract.
 
 #### Migration
 
 Consumers using `IDirectoryBeta` should use `IDirectory` instead.
 Methods available only on newer built-in map types are not available on shared maps.
+
+```typescript
+// Before
+// import type { IDirectoryBeta as Directory } from "@fluidframework/map/legacy";
```
jason-ha: uncomment?

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
jenn-le: I don't think there is any difference, this is an unnecessary change. I changed it back.

.changeset/use-fluid-shared-map-types.md
jason-ha: Resolved with no comment or change? Forgot to push?

jenn-le: Yup, forgot to push :)

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -342,9 +346,8 @@ abstract class CustomMapNodeBase<const T extends ImplicitAllowedTypes> extends T
 	public has(key: string): boolean {
 		return this.innerNode.tryGetField(brand(key)) !== undefined;
 	}
-	public *keys(): IterableIterator<string> {
-		const node = this.innerNode;
-		yield* node.keys();
+	public keys(): FluidIterableIterator<string> {
+		return this.innerNode.keys();
```
jason-ha: This feels like a change - what is the difference?

packages/loader/container-loader/src/pendingLocalStateStore.ts
```diff
@@ -151,20 +152,23 @@ export class PendingLocalStateStore<TKey> {
 				}
 				return { done, value: [value[0], JSON.stringify(value[1])] };
 			},
+			[Symbol.iterator](): FluidIterableIterator<[TKey, string]> {
+				return this;
+			},
 		};
```
jason-ha, line 158: What was going on here? I guess it was change that we don't need. Perhaps it was simplification that can follow.?

jason-ha (approved):

packages/loader/container-loader/src/pendingLocalStateStore.ts
jenn-le: This is required for it to conform to the FluidIterableIterator interface, I changed it to this to keep it more in line with how the implementation was previously, returning an iterator explicitly rather than yielding.

PR #13329, opened 2022-12-09
When enabled, chunk compressed batches larger than 1MB

andre4i:
Part of ADO:2089.

This adds the functionality of accepting batches which exceed the 1MB limit after compression, making it possible to submit payloads of any size. It is based on some of the previous chunking work which has been decommissioned a while back as it became incompatible with the default flush mode.

What currently happens is that if a batch is larger than the minimum compression threshold and gets compressed, yet its post-compression size is still larger than the maximum payload limit (1MB), the container will close with an error, as the payload is going to be rejected by the socket anyway. 

This change leverages the form of the compressed batch (which has one large message at the beginning and empty messages in the tail in order to reserve sequence numbers) by chunking the first op in small pieces which can be sent as separate ops (technically small one op batches) and then sending a batch with the last chunk as the first op along with the empty messages.

The feature is disabled by default, and it requires both compression to be enabled (via `compressionOptions` in `IContainerRuntimeOptions`) along with chunking (via specifying a value for `chunkSizeInBytes` in `IContainerRuntimeOptions`).

**The feature is experimental and changes the document schema**. Sending chunked messages has been disabled for a while, but the runtime is still able to process them. The chunked messages added with this change will require a runtime which is able to process them. Therefore, it is recommended that the feature is enabled only after the version of this change reaches full saturation, to ensure that all clients are able to interpret and process the new schema. The config enabling the feature (`chunkSizeInBytes` in `IContainerRuntimeOptions`) is relative only to producing the ops and not consuming the ops. The consumer is able to interpret these ops regardless of configuration.

This is part 3 of 3 from https://github.com/microsoft/FluidFramework/pull/13131

PR #27919, opened 2026-08-10
fix(container-runtime): skip clipped leading chunk streams when resol…

lindsnguyen:
## Description

Fixes [AB#80025](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/80025) — a Data-Refresh-blocking bug in the version-mark (savepoint) resolver.

`VersionMarkResolver.resolveFromHistory` scans forward from `sequenceNumberLowerBound + 1`
using a fresh historical-op unpacker (which holds a fresh `OpSplitter`). When the scan anchor
lands *mid-chunk-stream* — i.e. a chunked op whose leading chunk(s) sequenced at/before the
lower bound, outside the scan window — the fresh `OpSplitter` receives chunk 2+ first and throws
`DataCorruptionError("Chunk Id mismatch")`. That throw was uncaught, so `resolve()` **rejected**
instead of returning a resolved/pending/unresolvable result, breaking mark resolution for any
history whose window opened inside a chunk stream.

The fix detects and skips such a **clipped leading chunk stream** (per `clientId`) before feeding
ops to the unpacker, until the stream completes. This is provably safe: a version mark's target
batch is *pending* at capture, so all its ops — including chunks — sequence strictly after the
lower bound; the target's own chunk 1 is therefore always in-window and never clipped. A clipped
stream (chunk 1 ≤ lower bound) is by definition older/other traffic, never the target, so skipping
it can never skip the target or change the resolved sequence number.

The chunk-detection knowledge is factored into `shouldSkipClippedChunk` (in `inboundBatch.ts`,
alongside `inboundVersionMarkUpdate`), which mirrors the live unpacker's invariant: only runtime
(`type === "op"`) ops are deserialized, and a copy is deserialized rather than the reader-owned op.
A new exported `tryGetChunkedOp` helper (internal `opLifecycle` barrel only — not public API)
inspects an op's chunk descriptor.

**Repro / verification:** covered by new unit tests plus an end-to-end test driving the real
`RemoteMessageProcessor`/`OpSplitter` pipeline, which asserts a clipped leading chunk stream is
skipped (rather than throwing "Chunk Id mismatch") and the following target batch still resolves.
Full version-marks suite passes.

## Reviewer Guidance

- The safety argument for skipping (target is pending at capture ⇒ never clipped) is the crux —
  see `shouldSkipClippedChunk` and the guard in `resolveFromHistory`.
- `shouldSkipClippedChunk` deliberately re-uses the live unpacker's "runtime-ops-only, deserialize
  a copy" invariant (`containerRuntime.ts` `createHistoricalOpUnpacker`); confirming that mirror is
  faithful is worth a look.


packages/runtime/container-runtime/src/test/versionMarks/versionMarkResolver.spec.ts
```diff
@@ -155,6 +191,34 @@ function makeRecordingUnpackerFactory(
 	};
 }
 
+/**
+ * Builds the REAL inbound unpack pipeline, matching the container's `createHistoricalOpUnpacker`: a
+ * fresh `RemoteMessageProcessor` over a real `OpSplitter`/`OpDecompressor`/`OpGroupingManager`. Unlike
+ * the stand-in unpackers, this actually reassembles chunks — so a clipped leading chunk stream that
+ * reaches it throws `DataCorruptionError("Chunk Id mismatch")`, exercising the guard end-to-end.
+ */
+function makeRealUnpackerFactory(): () => (
```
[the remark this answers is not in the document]
anthony-murphy: Deep Review: Confirming and extending this. `makeRealUnpackerFactory` (spec line 200) is the only helper that builds the real `RemoteMessageProcessor`/`OpSplitter` reassembly path that actually throws `DataCorruptionError("Chunk Id mismatch")` — the exact regression AB#80025 exists to prevent — yet a search across the file returns only its definition; it is never called. Every test in the new `resolveFromHistory - mid-chunk-stream scan anchor` block instead uses `makeRecordingUnpackerFactory`, whose stand-in unpacker treats a chunk op as an ordinary single-message result and never reassembles chunks or throws. Those tests assert the skip *decision* (`fed` / `!fed.includes(...)`) but never drive an op through the throwing pipeline, so they cannot confirm the real pipeline no longer throws. This contradicts the PR description's claim of "an end-to-end test driving the real `RemoteMessageProcessor`/`OpSplitter` pipeline."

Preferred fix (a): wire `makeRealUnpackerFactory()` into a regression test in this block that feeds a clipped leading chunk stream through the real pipeline and asserts the following target resolves *without* throwing — ideally also asserting it *does* throw when the guard is bypassed, proving the harness reproduces AB#80025. This delivers the verification the PR promises and clears the likely `noUnusedLocals`/`no-unused-vars` build failure (CI is currently red). Fallback (b): delete the helper plus its five now-unused imports (`RemoteMessageProcessor`, `OpSplitter`, `OpDecompressor`, `OpGroupingManager`, `ensureContentsDeserialized`) and correct the description's end-to-end claim. Option (a) is strongly preferred.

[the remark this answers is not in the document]
lindsnguyen: updated

anthony-murphy:
`packages/runtime/container-runtime/src/versionMarks/inboundBatch.ts:449`

Deep Review: `shouldSkipClippedChunk` re-encodes the live-path gate — `op.type !== MessageType.Operation || typeof op.clientId !== "string"` plus `const messageCopy = { ...op }; ensureContentsDeserialized(messageCopy)` — that `createHistoricalOpUnpacker` (`containerRuntime.ts:1932-1944`) also encodes, and it maintains its own per-client `ChunkStreamAlignment` map. The PR's own Risks section concedes both sites must move together if the live unpacker's gating changes.

This is correct and the safest option as shipped: `OpSplitter`/`RemoteMessageProcessor` are untouched, there's zero new live-path risk, chunk wire-format recognition is not duplicated (`tryGetChunkedOp` is reused from the `opLifecycle` barrel), and the alignment map holds only leading-clip state. The residual is a documented two-place invariant, now guarded by the end-to-end regression test.

To retire the drift hazard, either route the runtime-op/clientId gate + deserialize-a-copy rule through a single shared/exported predicate so `createHistoricalOpUnpacker` and `shouldSkipClippedChunk` cannot silently drift, or file a tracked follow-up to migrate the skip into an opt-in `OpSplitter` mode and delete the parallel `ChunkStreamAlignment` map. Filing the follow-up (referenced from the code/DEV.md) is enough to clear this. The pipeline-vs-resolver placement itself is a design tradeoff for an area owner (anthony-murphy / markfields).

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -276,10 +277,9 @@ The existing real-service ODSP suites under `packages/test/test-end-to-end-tests
...
 5. **Conflicting batch identity:** `processInboundBatch` suppresses an identical repeated `(batchId, sequenceNumber)` pair but currently overwrites the map if the same `batchId` appears at a different sequence number. The protocol says that remapping is impossible, and map insertion order is used by MSN eviction. Assert or fail explicitly on a conflicting remap and add a regression test so corruption cannot invalidate the eviction ordering assumption.
 6. **Historical stream contract:** Add direct coverage for duplicated, decreasing, gapped, and malformed batch/chunk sequences, plus a target found after a trim gap. Define which violations are rejected by the driver, which are rejected by the resolver, and which conservatively return `pending`; never infer `unresolvable` from a stream that violated the requested range contract.
```
anthony-murphy: Deep Review: The "Resolver correctness and robustness gaps" list renumbers to 1, 2, 3 after deleting the now-fixed chunk-stream item, but the two trailing items stay labeled `5.` and `6.`, so the rendered list skips item 4. Renumber them to 4 and 5 for a contiguous list.

```suggestion
4. **Conflicting batch identity:** `processInboundBatch` suppresses an identical repeated `(batchId, sequenceNumber)` pair but currently overwrites the map if the same `batchId` appears at a different sequence number. The protocol says that remapping is impossible, and map insertion order is used by MSN eviction. Assert or fail explicitly on a conflicting remap and add a regression test so corruption cannot invalidate the eviction ordering assumption.
5. **Historical stream contract:** Add direct coverage for duplicated, decreasing, gapped, and malformed batch/chunk sequences, plus a target found after a trim gap. Define which violations are rejected by the driver, which are rejected by the resolver, and which conservatively return `pending`; never infer `unresolvable` from a stream that violated the requested range contract.
```

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -308,6 +366,12 @@ export class VersionMarkResolver implements IVersionMarkResolver {
 		if (existingSequenceNumber === sequenceNumber) {
 			return;
 		}
+		// The protocol guarantees a `batchId` maps to one sequence number; a conflicting remap would also
```
anthony-murphy, line 325: Deep Review: This PR adds `assert(existingSequenceNumber === undefined, 0xd1d /* version mark batchId remapped to a different sequence number */)` in `processInboundBatch`, after the early return when `existingSequenceNumber === sequenceNumber`. That assert hard-faults the container on the non-explicit-duplicate path the inbound detector deliberately tolerates.

The live inbound path in `containerRuntime.ts` tolerates the same effective `batchId` at two different sequence numbers for NON-explicit duplicates: `const shouldThrowOnDuplicate = batchIdExplicit || otherBatchIdExplicit;` then `if (shouldThrowOnDuplicate) { throw error; }`. A non-explicit duplicate (server-outage scenario) does not throw — it falls through to `versionMarkResolverInternal.processInboundBatch(...)` with a NEW, different sequence number for the same effective `batchId = generateBatchId(clientId, batchStartCsn)`. So `existingSequenceNumber` is defined and differs, and `0xd1d` faults the container. Before this PR the same line was a silent `Map.set` overwrite (non-fatal).

This is exercised by an existing test that requires no-throw for "a duplicate batch with no explicit batchId on either side (service-outage scenario)" (`containerRuntime.spec.ts:2848-2890`). The inline protocol-single-mapping justification here is contradicted by `DuplicateBatchDetector`'s non-throwing path (`duplicateBatchDetector.ts:120-132`). It only triggers when version-mark tracking is enabled — i.e. the Data-Refresh apps this PR targets.

The assert direction was reviewer-requested (#27885 DEV.md), so keep it — but scope it to match the inbound throw-policy. Either (a) gate the assert on the same `batchIdExplicit || otherBatchIdExplicit` condition the inbound `DuplicateBatchDetector` uses, so the resolver only faults when the duplicate path already decided it is fatal; or (b) preserve the first-seen mapping and return early on a remap collision (a server-induced duplicate resolves to the same batch content). Add a regression test covering the non-explicit-duplicate + version-mark-tracking path, extending the existing non-throw contract at `containerRuntime.spec.ts:2848-2890`.

packages/runtime/container-runtime/src/versionMarks/inboundBatch.ts
```diff
@@ -68,3 +75,47 @@ export function inboundVersionMarkUpdate(
...
+/**
+ * Decides whether `op` belongs to a clipped *leading* chunk stream a history scan must skip. A scan
+ * anchored mid-stream can hand a fresh OpSplitter chunk 2+ with no chunk-1 state, throwing
+ * `DataCorruptionError("Chunk Id mismatch")`. The first chunk seen per client decides: chunkId 1 is a
+ * clean start (`aligned`, feed it); higher means the anchor clipped this stream (`skipping`, drop until
+ * its final chunk, then realign). Non-runtime, non-chunk, and clientless ops never skip. Mutates `alignment`.
+ *
+ * @internal
+ */
+export function shouldSkipClippedChunk(
```
sonalideshpandemsft: Could  shouldSkipClippedChunk  return the deserialized op alongside its skip decision? It currently obtains a deserialized copy through  tryGetDeserializedRuntimeOpCopy , but discards it when returning the boolean, causing the unpacker to copy and deserialize the original op again. Returning  { skip, deserializedOp }  would allow the scan and unpacker to share that work.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
lindsnguyen: updated to return early instead of assert

packages/runtime/container-runtime/src/versionMarks/inboundBatch.ts
lindsnguyen: this makes sense, it does deserialize twice. But since this only runs during `resolveFromHistory` which doesn't get called often (`fetchMessages` dominates, not the extra `JSON.parse`), I think leaving it as a clean boolean return is cleaner rn. If it gets called more often we can change it then? I can file a followup (backlog though, I don't think it's needed currently)

lindsnguyen: https://dev.azure.com/fluidframework/internal/_workitems/edit/80352

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -143,24 +143,26 @@ The runtime does not store or mutate app marks. The listener only lets the app r
 3. Set `from = sequenceNumberLowerBound + 1`, create a fresh unpacker, and create an `AbortController`.
 4. Request `fetchMessages(from, undefined, abortSignal)`.
 5. Read every stream chunk until the target is found or the stream returns `done`.
-6. Return the matched completed batch's last sequence number, or classify the miss.
+6. Return the matched completed batch's last sequence number (with the count of leading ops skipped), or classify the miss.
 7. Abort the controller in `finally`, both on success and on exhaustion/error, so the underlying fetch can stop any remaining work.
+8. `resolve()` emits one `Resolve` telemetry event (`outcome`, `path`, `clippedLeadingOpsSkipped`, `durationMs`) before returning — on every path, including a thrown scan (`outcome: "error"`, emitted via `finally`). See [Telemetry](#telemetry).
```
markfields: I'm not sure we need to coin a new term "clipped leading op" here.  It's a pretty scoped concept and easy to understand, whereas "clipped" feels like a new concept that needs a whole doc explaining what they are and all the nuance.

Can we just say "opsSkipped"?  And remove the word "clipped" elsewhere as well

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -143,24 +143,26 @@ The runtime does not store or mutate app marks. The listener only lets the app r
...
 
 For each raw op, the scan records the first returned sequence number before filtering because that value is also the trim-availability signal. It then:
 
-1. Drops a `batch: false` marker when no batch is currently in progress. The scan anchor may land on the clipped tail of an ordinary batch whose start was before `from`; feeding that orphan end marker to `RemoteMessageProcessor` would violate its batch-state invariant.
-2. Passes the op to the unpacker. `undefined` means a filtered system op or an incomplete chunk waiting for more fragments.
-3. Mirrors the unpacker's batch-in-progress state from `batchStartingMessage` and final `nextBatchMessage` results.
-4. Calls `inboundVersionMarkUpdate`, carrying the batch id across piecemeal results exactly as the live path does.
-5. Returns immediately when the completed batch id matches the requested id.
+1. Drops a `batch: false` marker when no batch is currently in progress (counting it into `clippedLeadingOpsSkipped`). The scan anchor may land on the clipped tail of an ordinary batch whose start was before `from`; feeding that orphan end marker to `RemoteMessageProcessor` would violate its batch-state invariant.
```
markfields: I don't think an ordinary batch can be clipped here: `InboundBatchAggregator` waits for the complete batch, and `DeltaQueue` drains it synchronously, so a runtime-observed `lastSequenceNumber` should not land mid-batch. Please remove this guard and its synthetic coverage, and keep `clippedLeadingOpsSkipped` specific to clipped chunk streams.

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -143,24 +143,26 @@ The runtime does not store or mutate app marks. The listener only lets the app r
...
...unk-reassembly state must not be shared across scans). If the scan starts mid-stream, that splitter would receive chunk 2+ with no chunk-1 state and throw `DataCorruptionError("Chunk Id mismatch")`, rejecting `resolve()` instead of returning a defined result. The scan therefore **detects and skips a clipped leading chunk stream** before feeding the op to the unpacker (via `tryGetChunkedOp`, using deserialized contents): the first chunk seen for a `clientId` with `chunkId > 1` marks that client `skipping`; its chunks are dropped until the final chunk (`chunkId >= totalChunks`) passes, after which the client is `aligned` and its subsequent in-window streams are fed normally. Alignment is tracked **per `clientId`** because chunk streams from different clients can interleave. This is safe for the same reason as the orphan batch-end guard: the target batch was pending at capture, so its own chunk 1 is after `from`, and skipping a clipped *leading* stream can never skip the target. See the `resolveFromHistory - mid-chunk-stream scan anchor` tests, which cover every intermediate-chunk anchor, an in-window stream that must not be skipped, interleaved per-client streams, and a defined miss.
```
markfields: I think this logic belongs in `OpSplitter` rather than a resolver-side pre-filter that duplicates OpSplitter logic (tracking in-progress batches, deserialization, etc). We can add a flag to gate the assertion so the regular live op processing path still does that, but it can also handle partial chunked batches.

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -143,24 +143,26 @@ The runtime does not store or mutate app marks. The listener only lets the app r
...
```
markfields, line 161: A test case to add (I think) - a complete chunked batch interleaved with a partial chunked batch.  Maybe the complete one should be the target batch.

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -184,6 +186,36 @@ This is an **interim, read-derived** signal: it infers availability from how the
...
+| --- | --- | --- | --- |
+| `Resolve` **(implemented)** | end of `resolve()` (via `finally`, so a thrown scan is still reported) | `outcome` (`resolved`\|`pending`\|`unresolvable`\|`error`), `path` (`session`\|`history`\|`noReader`), `clippedLeadingOpsSkipped` (number), `durationMs`, `sequenceNumber` (when resolved) | success rate; how often history is needed; latency; the **near-miss** signal; `unresolvable` = data-loss KPI; `error` = the scan threw (e.g. an uncovered chunk-mismatch or driver failure) |
+| `Capture` *(planned — AB#80270)* | `sealAndCaptureVersionMark()` | `kind` (`pending`\|`resolved`) | capture volume; pending ratio |
+
+`clippedLeadingOpsSkipped` counts leading orphan-batch-end and clipped-chunk ops dropped by the scan-anchor guards. The static invariant (the target batch is pending at capture, so it sequences at/after `from` and is never clipped) means a *correct* skip never drops the target. The one thing the invariant cannot self-verify in the field is a future violation of it; the **near-miss filter `clippedLeadingOpsSkipped > 0 && outcome !== "resolved"`** surfaces exactly that (kept informational because `pending` can be legitimate).
```
markfields: Not sure how useful this is to track and log. How do you imagine we'll interpret this?  What's really the difference between a skipped chunk and any other non-target op?

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -252,7 +284,8 @@ Per-inbound-batch work (deriving the batch identity and populating the map/notif
 ## Current test map
```
markfields, line 279: (in a separate PR) Do we want to try out per-spec.ts md files, like versionMarkResolver.spec.md, with the labeled catechism-style mirror of the tests? Not a leading question - what do you think?

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -276,12 +309,10 @@ The existing real-service ODSP suites under `packages/test/test-end-to-end-tests
...
-6. **Historical stream contract:** Add direct coverage for duplicated, decreasing, gapped, and malformed batch/chunk sequences, plus a target found after a trim gap. Define which violations are rejected by the driver, which are rejected by the resolver, and which conservatively return `pending`; never infer `unresolvable` from a stream that violated the requested range contract.
+1. **Live/history resolution races:** A batch can be recorded by `processInboundBatch` while a history scan for the same `batchId` is in progress. Recheck the live map before returning a miss so `resolve()` does not return stale `pending` or `unresolvable` after the batch has resolved in-session. Also decide whether concurrent calls for the same locator should share one scan rather than issuing duplicate delta-storage reads.
```
markfields, line 308: Nit: If you number these all "1." it'll make the diff cleaner (and markdown renderers compute the actual number typically).  Or just do a bulleted list

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -171,45 +175,85 @@ export class VersionMarkResolver implements IVersionMarkResolver {
...
-	 * See DEV.md.
+	 * Also reports `clippedLeadingOpsSkipped` (leading orphan-batch / clipped-chunk ops dropped by the
+	 * anchor guards) for the resolve telemetry. See DEV.md.
 	 */
 	private async resolveFromHistory(
 		reader: IHistoricalOpReader,
 		batchId: string,
 		sequenceNumberLowerBound: number,
-	): Promise<ResolveResult> {
+	): Promise<{ result: ResolveResult; clippedLeadingOpsSkipped: number }> {
```
markfields: (if we don't remove it) put clippedops number on ResolveResult rather than a nested result type?

markfields (commented):
I'm suggesting we try to reduce duplicate logic and update the main op processing path to be able to handle partial chunked batches,

Really, it would be amazing to be able to ignore virtualization completely.  This would require putting the batch metadata on the final chunk.  We should discuss if this is feasible, that would be ideal.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
lindsnguyen: planning to remove

packages/runtime/container-runtime/src/versionMarks/DEV.md
lindsnguyen: Decide to remove it because you're right, it doesn't actually provide anything addtional by storing this info

anthony-murphy:
## Deep Review

*Reviewed commit `d92984c` on 2026-08-14.*

Not ready for sign-off, but close. Area owner markfields' 2026-08-13 review directed the two changes earlier reviews flagged as risky — moving the tolerance into `OpSplitter` (`allowInitialPartialChunkStream`) and removing the resolver's ordinary-batch orphan-end guard — and this commit implements that owner-directed redesign, which demotes the headline correctness concerns from blockers to held polish. What remains is contained: a disputed `0xd1d` fault on the non-explicit-duplicate path, zero regression coverage for the removed guard, a stale PR description, and a red broken-links CI check.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Fix the red **broken-links** CI check (issue comment [#5299090857](https://github.com/microsoft/FluidFramework/pull/27919#issuecomment-5299090857))
- [ ] Update the PR description and Reviewer Guidance to match the shipped design: `allowInitialPartialChunkStream` on `OpSplitter` and `tryGetDeserializedRuntimeOpCopy` in `opSerialization.ts`. Remove the references to `shouldSkipClippedChunk`, `tryGetChunkedOp`, `inboundBatch.ts`, and the removed `resolveFromHistory` guard — those describe an abandoned intermediate design that is not in the diff. Use the Suggested description below.

### Context for Reviewers

- First relaxation of the always-on "Chunk Id mismatch" receiver guard that has been ungated since #13329. `allowInitialPartialChunkStream` defaults to `false`, is enabled only for the historical scan unpacker, and the live inbound `OpSplitter` still throws on first mismatch — verify the flag can never leak onto the live path and that discarding a clipped stream can't mask corruption on the target batch.
- New `tryGetDeserializedRuntimeOpCopy` (`opLifecycle/opSerialization.ts`) preserves the runtime-op-only / shallow-copy-then-deserialize invariant from #13273/#22750 — keep it as the single mirror of the live-path invariant.
- `opSplitter.ts`/`opSerialization.ts`/duplicate-batch have an established reviewer cohort (andre4i, kian-thompson, markfields, vladsud); `versionMarkResolver.ts`/`.spec.ts`/`DEV.md` are greenfield single-author code now under active markfields review. markfields and anthony-murphy are the natural sign-off owners.

- **Needs owner judgment (markfields)** — whether the `0xd1d` assert is correct on the version-mark path or must be scoped to match the inbound `batchIdExplicit || otherBatchIdExplicit` throw-policy. markfields holds "remapping is impossible"; the prior review demonstrated a non-explicit-duplicate mechanism reaching it. Owner is the authority on `DuplicateBatchDetector` (#22310).
- **Needs owner judgment (markfields)** — sign-off that the removed ordinary-batch guard is genuinely dead code given `InboundBatchAggregator`/`DeltaQueue` atomicity on the replayed (not just live) op stream.
- **Cannot be assessed by the pipeline** — runtime behaviour of the real `RemoteMessageProcessor`/`OpSplitter` reassembly against production chunk streams, and that the capture-time "target pending ⇒ never clipped" invariant holds for every anchor case in the field.

</details>

### Suggested description

```markdown
## Description

Fixes [AB#80025](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/80025) — a Data-Refresh-blocking bug in the version-mark (savepoint) resolver.

`VersionMarkResolver.resolveFromHistory` scans forward from `sequenceNumberLowerBound + 1` using a
fresh historical-op unpacker (holding a fresh `OpSplitter`). When the scan anchor lands *mid-chunk-stream*
— a chunked op whose leading chunk(s) sequenced at/before the lower bound, outside the scan window — the
fresh `OpSplitter` receives chunk 2+ first and throws `DataCorruptionError("Chunk Id mismatch")`. That
uncaught throw made `resolve()` reject instead of returning resolved/pending/unresolvable.

The fix adds an opt-in `OpSplitterOptions.allowInitialPartialChunkStream` flag on `OpSplitter`. It
defaults to `false` (live inbound path stays strict); it is enabled only for the historical scan unpacker
created by `createHistoricalOpUnpacker`. When set, `OpSplitter` drains a leading partial chunk stream
per client until it completes, then re-enforces strict chunk-ID checking. This is provably safe: a version
mark's target batch is *pending* at capture, so all its ops — including chunk 1 — sequence strictly after
the lower bound and are never clipped; any clipped stream (chunk 1 <= lower bound) is older/other traffic.

Supporting changes:
- New `tryGetDeserializedRuntimeOpCopy` helper in `opLifecycle/opSerialization.ts` (exported from the
  internal `opLifecycle` barrel only) centralizes the "runtime ops only, deserialize a copy" invariant so
  `createHistoricalOpUnpacker` and its tests share one definition.
- Per area-owner review, the resolver's previous ordinary-batch orphan-end-marker guard was removed:
  `InboundBatchAggregator` aggregates ordinary batches atomically and `DeltaQueue` drains them
  synchronously, so `lastSequenceNumber` cannot land mid-ordinary-batch; only chunk streams can be clipped.
- `processInboundBatch` now fails explicitly (`0xd1d`) on a conflicting `batchId -> sequenceNumber` remap
  rather than silently overwriting the map that MSN eviction ordering depends on.

## Reviewer Guidance

- Crux of safety: target batch is pending at capture => its chunk 1 is always in-window => skipping a
  clipped leading stream can never skip the target. See `allowInitialPartialChunkStream` in
  `opSplitter.ts` and its wiring in `createHistoricalOpUnpacker` (`containerRuntime.ts`).
- Confirm the flag can never leak onto the live inbound `OpSplitter`.
- Confirm the `0xd1d` remap assert cannot fire on the tolerated non-explicit-duplicate (service-outage)
  path where version-mark tracking is enabled.

## Verification

New unit tests for `tryGetDeserializedRuntimeOpCopy` and `OpSplitter` skip/re-sync behaviour, plus an
end-to-end resolver test (`makeRealUnpackerFactory`) driving the real `RemoteMessageProcessor`/`OpSplitter`
pipeline. Full version-marks suite passes.
```

- `e5db88f` 2026-08-11 · **5/10** — new `0xd1d` assert flagged as a Tier 2 crash regression
- `8182059` 2026-08-11 · **8/10** — prior CI/test-wiring blocker resolved; no correctness blockers found
- `62f06a9` 2026-08-10 · **5/10** — unwired real-pipeline test left the anti-throw fix unverified

</details>

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
anthony-murphy: Deep Review: Thanks — but on the current commit `d92984c` the `0xd1d` assert is still present in `processInboundBatch`. The early return only covers the `existingSequenceNumber === sequenceNumber` case; the assert still fires when the same effective `batchId` remaps to a *different* sequence number. markfields' "remapping is impossible" premise resolves the design question, but it doesn't engage the non-explicit-duplicate (service-outage) mechanism: `DuplicateBatchDetector` deliberately tolerates the same effective `batchId` at a different sequence number (`shouldThrowOnDuplicate = batchIdExplicit || otherBatchIdExplicit`), and the no-throw test at `containerRuntime.spec.ts:2848-2890` requires exactly that — but only when version-mark tracking is enabled (the Data-Refresh apps this PR targets).

Either land the return-early change you described (it isn't in the current diff), or gate the assert on the same `batchIdExplicit || otherBatchIdExplicit` condition the inbound path uses, and add a regression test extending the non-throw contract at `containerRuntime.spec.ts:2848-2890`. If the invariant truly holds on the version-mark path, keep the assert but document why it is unreachable in the comment.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -208,35 +239,18 @@ export class VersionMarkResolver implements IVersionMarkResolver {
 			const stream = await reader.fetchMessages(from, undefined, abortController.signal);
```
anthony-murphy, line 239: Deep Review: The ordinary-batch orphan-end guard (`inBatch` tracking plus the `!inBatch && asBatchMetadata(op.metadata)?.batch === false` skip) is removed here per markfields' direction, and its two regression tests (`versionMarkResolver.spec.ts:499-627`) are deleted — so the clipped ordinary-batch-tail path now has zero coverage. `RemoteMessageProcessor.getResultBasedOnBatchMetadata()` still asserts `0x9d5` unconditionally when a `batch:false` marker arrives with no batch in progress, and the new `OpSplitter` `allowInitialPartialChunkStream` tolerance covers chunk streams only — not ungrouped multi-op batch tails.

Accept the owner-directed removal, but capture the invariant so it can't silently regress: record markfields' rationale in code/DEV.md (`InboundBatchAggregator` aggregates atomically + `DeltaQueue` drains synchronously => `lastSequenceNumber` never lands mid-ordinary-batch), and confirm it holds for the stored/replayed stream the resolver scans via `fetchMessages`, not just the live inbound queue. Consider retaining one lightweight assertion or test that fails loudly if a clipped ordinary-batch tail ever reaches the unpacker, so the deleted coverage can't mask a future replay/capture-path change.

packages/runtime/container-runtime/src/test/versionMarks/versionMarkResolver.spec.ts
```diff
@@ -425,88 +477,74 @@ describe("VersionMarkResolver", () => {
...
-			// to find the target.
+	describe("resolveFromHistory - partial initial chunk stream", () => {
+		it("resolves a complete chunked target interleaved with another client's partial stream", async () => {
+			const targetContents = JSON.stringify({
+				type: ContainerMessageType.FluidDataStoreOp,
+				contents: {},
+			});
+			const splitAt = Math.ceil(targetContents.length / 2);
 			const reader = makeReader([
 				[
```
anthony-murphy, line 488: Deep Review: `makeRealUnpackerFactory` is meant to mirror the container's `createHistoricalOpUnpacker`, but it re-derives the runtime-op invariant inline (`op.type !== MessageType.Operation || typeof op.clientId !== "string"` => `{ ...op }` => `ensureContentsDeserialized`) instead of calling the shared `tryGetDeserializedRuntimeOpCopy(op)` that production was refactored to use. The helper exists precisely so callers don't re-derive this invariant ("Shared so callers don't re-derive the invariant"); a future change to the helper would silently desync this "faithful mirror" — the very test that validates the clipped-stream skip behaviour. Have `makeRealUnpackerFactory` call `tryGetDeserializedRuntimeOpCopy(op)` so the mirror stays faithful to `createHistoricalOpUnpacker` by construction.

packages/runtime/container-runtime/src/versionMarks/DEV.md
lindsnguyen: yeah i think this makes sense, got copilot to file with some context: https://dev.azure.com/fluidframework/internal/_workitems/edit/80578

markfields (approved):
I have some comments but they're having trouble posting, none were blocking, I'll follow up after you merge

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1965,13 +1967,11 @@ export class ContainerRuntime
 							),
 						);
 						return (op) => {
-							// Mirror the live path: only modern runtime-envelope ops (type Operation, with a
-							// client id) carry batches; skip system/server ops.
-							if (op.type !== MessageType.Operation || typeof op.clientId !== "string") {
+							// Only runtime ops carry batches; deserialize a copy so the reader-owned op stays untouched.
+							const messageCopy = tryGetDeserializedRuntimeOpCopy(op);
```
markfields, line 1980: Could we use this helper at the live `process` boundary too, so modern live and historical messages share the same copy/deserialization path? We should avoid a second shallow copy, keep the legacy unpacked-message handling separate, and preserve the live assertion for malformed modern ops without a client ID.

packages/runtime/container-runtime/src/opLifecycle/opSplitter.ts
```diff
@@ -214,6 +244,33 @@ export class OpSplitter {
...
+				if (
+					nextInitialChunkId !== undefined &&
+					chunkedContent.chunkId !== nextInitialChunkId
+				) {
+					this.throwChunkIdMismatch(message, chunkedContent, nextInitialChunkId);
+				}
+
+				if (
+					nextInitialChunkId === undefined &&
+					(chunkedContent.chunkId <= 1 || chunkedContent.chunkId > chunkedContent.totalChunks)
```
markfields, line 259: Could we make these states explicit instead of using `1` as an aligned sentinel? I suggest `chunkMap` for an active complete stream, `nextInitialChunkIdByClient` only for an actively discarded partial stream, and an `alignedClientIds` set so later partial starts still throw; also validate `chunkId <= totalChunks` directly rather than routing that invalid case through `addChunk`.

packages/runtime/container-runtime/src/opLifecycle/opSplitter.ts
```diff
@@ -45,6 +59,8 @@ function isChunkedContents(contents: unknown): contents is IChunkedContents {
 export class OpSplitter {
 	// Local copy of incomplete received chunks.
 	private readonly chunkMap: Map<string, string[]>;
+	// No entry = unseen; 1 = aligned; >1 = the next initial partial chunk to discard.
+	private readonly nextInitialChunkIdByClient = new Map<string, number>();
```
markfields, line 63: Test case to consider: a malformed tolerated initial stream (a gap, duplicate, or restart) should throw

markfields (approved):

packages/runtime/container-runtime/src/containerRuntime.ts
lindsnguyen: will do in followup

packages/runtime/container-runtime/src/opLifecycle/opSplitter.ts
lindsnguyen: will do in followup

markfields (approved):

PR #27920, opened 2026-08-11
fix(tree): Better asserts, and fix latent optimized forest array corruption bug

CraigMacomber:
## Description

Improves validation and diagnostics for the optimized SharedTree forest. Internal attach, detach, create, and destroy operations are now validated before mutating forest state, preventing a failed operation from partially modifying the forest or consuming detached content. Additional assertions fail closer to the source of inconsistent operations, making failures such as later out-of-bounds chunk indexes easier to triage.

This PR also corrects chunk indexing when a shared multi-node `SequenceChunk` is normalized for editing. The previous implementation could clone the selected child into the wrong slot and replace a sibling. This issue is currently unreachable through `ForestTypeOptimized`: its current chunk policy does not create sequence chunks, decoded fields are deaggregated, and delta builds split sequence content into individual node cursors. The fix and targeted tests preserve correctness if sequence chunks are used by this forest in the future or through internal/custom chunk configurations.

Other changes ensure replaced chunk references are released correctly (improving our ability to edit in place) and add focused coverage for chunk lookup, copy-on-write normalization, operation atomicity, reference ownership, and field splitting.

## Reviewer Guidance

The publicly reachable behavior changes are the visitor operation validation and improved assertions in `chunkedForest.ts`. The `SequenceChunk` indexing correction is defensive because optimized forests do not currently use sequence chunks.


packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -236,49 +375,24 @@ export class ChunkedForest implements IEditableForest {
...
-					// TODO: this could fail for really long chunks being split (due to argument count limits).
-					// Current implementations of chunks shouldn't ever be that long, but it could be an issue if they get bigger.
-					chunks.splice(indexOfChunk, 1, ...newChunks);
-					found.referenceRemoved();
 
-					found = newChunks[indexWithinChunk] ?? oob();
-				}
-				assert(found instanceof BasicChunk, 0x536 /* chunk should have been normalized */);
-				if (found.isShared()) {
-					this.mutableChunk = chunks[indexOfChunk] = found.clone();
```
CraigMacomber, line 277: This line is the main bug. If we split a chunk and somehow got back a shared BasicChunk, this index is wrong. It should be `indexOfChunk + indexWithinChunk` instead.

I found expressing the fix that way to be really confusing, so I refactored this code so it has a lot less different indexes in scope, less mutable variables, and overall clearer more documented flow and more testable free functions.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -144,10 +262,20 @@ export class ChunkedForest implements IEditableForest {
 			},
 			destroy(detachedField: FieldKey, count: number): void {
 				this.forest.#events.emit("beforeChange");
+				const field = this.forest.roots.fields.get(detachedField);
+				assert(field !== undefined, "Destroyed field must exist");
```
brrichards, line 266: I notice this is the one new assert without a trigger test. would be nice just to have them all covered

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -123,11 +228,24 @@ export class ChunkedForest implements IEditableForest {
 			this.roots = this.roots.clone();
 		}
 
+		// The `CombinableVisitor` which can apply deltas to this forest.
+		// Annoyingly declaring the type here or in a satisfies clause causes this code not to build,
+		// so the type is inferred.
+		//
+		// This visitor works by shallow copying (where needed) the forest as its walked down
+		// to ensure the location being traversed (and thus possibly edited) is an exclusively owned BasicChunk.
+		// Future optimizations may improve thus, but currently that is a requirement for the editing code to work.
```
brrichards:
```suggestion
		// Future optimizations may improve this, but currently that is a requirement for the editing code to work.
```

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -123,11 +228,24 @@ export class ChunkedForest implements IEditableForest {
 			this.roots = this.roots.clone();
 		}
 
+		// The `CombinableVisitor` which can apply deltas to this forest.
+		// Annoyingly declaring the type here or in a satisfies clause causes this code not to build,
+		// so the type is inferred.
+		//
+		// This visitor works by shallow copying (where needed) the forest as its walked down
```
brrichards:
```suggestion
		// This visitor works by shallow copying (where needed) the forest as it's walked down
```

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -140,6 +151,151 @@ describe("ChunkedForest", () => {
...
+
+			const result = ensureExclusiveBasicChunk(chunks, 0, makeCompressor());
+
+			assert.notEqual(result, basic);
+			assert.equal(chunks[0], result);
+			assert.equal(result.isShared(), false);
+			assert.equal(basic.isShared(), false);
+		});
+
+		it("normalizes a UniformChunk and selects the requested node", () => {
```
brrichards, line 271: nit: It looks like every test that calls to `ensureExclusiveBasicChunk` are using a `chunks` array that is a single chunk so `indexOfChunk = 0` always. Might be nice to include a test that exercises the offset portion of `finalINdexOfChunk = indexOfChunk + indexWithinChunk` since this is the area the bug being fixed was in.

brrichards (approved):
Couple small nits and potential coverage suggestions, but overall LGTM

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
```diff
@@ -272,5 +464,158 @@ describe("ChunkedForest", () => {
 			}
 			assert.deepEqual(values, [0, 1, 99, 2, 3, 4]);
 		});
+
+		it("rejects an attach whose destination exceeds the field length without consuming its source", () => {
```
CraigMacomber: This is actually something we don't require: recovery from failed application of malformed deltas is not something we support, and should probably mark the forest as broken. I should cleanup this test.

packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
CraigMacomber: Added

Josmithr (commented):
Can we / should we update the object forest implementation with the same validation enhancements?

packages/dds/tree/src/test/feature-libraries/chunked-forest/chunkedForest.spec.ts
CraigMacomber: Added

CraigMacomber (commented):
> Can we / should we update the object forest implementation with the same validation enhancements? 

Done.

.changeset/fix-chunked-forest-editing.md
```diff
@@ -0,0 +1,11 @@
...
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": fix
+---
+Improve validation in SharedTree forests
+
+Our [ForestType](https://fluidframework.com/docs/api/tree/foresttype-interface) implementations now have better and more consistent validation of changes being applied.
+This should only impact cases which have hit bugs, resulting in edits which are not valid for the tree they are being applied to.
+Now the asserts should be more specific, helping to triage such issues.
+This also reduces he risk of of document corruption by catching invalid data earlier and more consistently.
```
brrichards:
```suggestion
This also reduces the risk of document corruption by catching invalid data earlier and more consistently.
```

packages/dds/tree/src/test/forestTestSuite.ts
```diff
@@ -403,6 +406,159 @@ export function testForest(config: ForestTestConfiguration): void {
 		assert.equal(detachedCursor.getFieldLength(), 0);
 	});
 
+	/**
+	 * This suite checks that forests assert on invalid deltas.
+	 *
+	 * Generally asserts like this are not part of the contract we would expect form an interface implementation,
```
brrichards:
```suggestion
	 * Generally asserts like this are not part of the contract we would expect from an interface implementation,
```

brrichards (approved):

packages/dds/tree/src/test/forestTestSuite.ts
CraigMacomber: Your suggestion converted indent to spaces, so making a fixed version of it:
```suggestion
	 * Generally asserts like this are not part of the contract we would expect from an interface implementation,
```

brrichards (approved):

Josmithr (approved):
Approving docs changes. I'll defer to @brrichards to sign off on the code/test changes.

Update: which I see he has already done :)

PR #27926, opened 2026-08-11
fix: preserve BubbleBench container compatibility

jenn-le:
## Description

BubbleBench needs to support both Fluid Tree arrays and native arrays. This PR introduces a minimal shared collection contract so the benchmark remains compatible with both container implementations without requiring the full built-in array interface.

Because the affected BubbleBench packages are private, the standalone changeset records the fix without adding package changelog entries, following the repository changeset guidance.

Validation performed:

- `git diff --check upstream/main...HEAD`
- `build:compile` for all five BubbleBench packages: `common`, `baseline`, `ot`, `experimental-tree`, and `shared-tree`
- Jest suites run sequentially for `ot`, `experimental-tree`, and `shared-tree` (all passed)

## Reviewer Guidance

The diff is limited to the shared BubbleBench collection type, its three Tree/OT adapters, and a standalone no-package changeset.


jenn-le:
Superseded by a main-targeting PR because GitHub does not allow changing the base of a pull request already registered in a native stack.

jenn-le:
Superseded by #27928, which targets main independently.

PR #27933, opened 2026-08-11
fix(tinylicious): support waitContainerToCatchUp

alexvy86:
## Description

Tinylicious did not fully support `waitContainerToCatchUp` for read-mode containers loading from the beginning of an unsummarized document.

Two distinct server behaviors contributed to the problem:

- Tinylicious's in-memory database treated a delta range lower bound of `0` as absent, so `{ $gt: 0 }` returned no operations. This query bug is specific to Tinylicious; Routerlicious's Mongo-backed storage handles the range correctly.
- The local orderer did not report a checkpoint sequence number when clients connected. A read-mode client could fetch available operations but could not prove how far it needed to catch up, so `waitContainerToCatchUp` returned `false` as permitted by its contract.

This change preserves zero-valued range bounds, tracks the latest sequence number in the local orderer, and includes it in connected messages. It adds focused database and local-server tests plus an end-to-end regression test.

### Why the Nexus change does not alter deployed Routerlicious behavior

`IConnected` already defines the optional `checkpointSequenceNumber` field, and clients already understand it. This PR does not introduce a new wire-protocol field; it makes the local server populate an existing capability.

Nexus is shared transport plumbing and receives its orderer manager through dependency injection. Tinylicious supplies `LocalOrdererManager`, which now implements the optional `getCheckpointSequenceNumber` capability. Deployed Routerlicious supplies `OrdererManager`, which does not implement it. Nexus uses optional chaining, so those Routerlicious deployments continue omitting the field and retain their existing behavior.

The end-to-end test remains skipped until a Tinylicious version containing these changes is published and consumed by the end-to-end test package. When run manually unskipped, it fails against published Tinylicious 7.0.1 with `false !== true` and passes against the locally built server.

## Reviewer Guidance

Please focus on the checkpoint sequence-number propagation through the local orderer and Nexus connection path. The intended rollout is to merge and publish Tinylicious and its updated server dependencies first, then update the end-to-end dependency and unskip the regression test.


server/routerlicious/packages/memory-orderer/src/localOrderer.ts
```diff
@@ -179,7 +186,33 @@ export class LocalOrderer implements IOrderer {
 	) {
 		this.existing = details.existing;
 		this.dbObject = this.getDeliState();
-		this.socketPublisher = new LocalSocketPublisher(this.pubSub);
+		const deliState: IDeliState =
+			typeof this.dbObject.deli === "string"
+				? JSON.parse(this.dbObject.deli)
+				: this.dbObject.deli;
+		this.latestSequenceNumber = deliState.sequenceNumber;
```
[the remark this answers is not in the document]
alexvy86: Fixed in latest

jzaffiro (approved):
docs look good!

server/routerlicious/.changeset/tidy-pandas-obey.md
```diff
@@ -0,0 +1,22 @@
+---
+"@fluidframework/server-lambdas": patch
```
CraigMacomber: IIRC, our tooling tends to screw things up if you don't leave these at the default "minor" that out tooling always uses. I think it breaks the release notes generation.

server/routerlicious/.changeset/tidy-pandas-obey.md
```diff
@@ -0,0 +1,22 @@
...
+"@fluidframework/server-local-server": patch
+"@fluidframework/server-memory-orderer": patch
+"@fluidframework/server-services-core": patch
+"tinylicious": patch
```
CraigMacomber: Title below only describes a change in tinylicious, so its a bit odd for so many packages to be listed here.

Maybe make the title more general, or list less impacted packages?

CraigMacomber: Maybe something like "Fix checkpoint catchup bugs" then list the various user facing impacts in the description?

server/routerlicious/packages/local-server/src/localOrdererManager.ts
```diff
@@ -76,6 +76,22 @@ export class LocalOrdererManager implements IOrdererManager {
...
 
+	/**
+	 * Gets the latest sequence number known by the document's local orderer. Nexus includes this
+	 * value in the connected message so read-mode clients can reliably wait until they are caught up.
+	 *
+	 * @param tenantId - The tenant that owns the document.
+	 * @param documentId - The document whose latest sequence number should be returned.
+	 * @returns The latest sequence number known to the local orderer.
+	 */
+	public getCheckpointSequenceNumber?: (
```
CraigMacomber: Is there a reason this is a writable optional property (so users of it can set it to undefined or another function for example) rather than a method?

From the doc comment, I don't see a reason we would want this to be optional (its always defined), nor writable, nor an instance property.

Defining a single method on the prototype using the method syntax is more efficient than allocating a copy of the function for every instance, and would better match the style of the methods before and after.

If there is a reason its not taking that pattern, it should probably document that.

server/routerlicious/packages/memory-orderer/src/localOrderer.ts
```diff
@@ -198,6 +233,13 @@ export class LocalOrderer implements IOrderer {
 		return orderer;
 	}
 
+	/**
+	 * Gets the latest sequence number broadcast by this orderer.
+	 *
+	 * @returns The sequence number a connecting client must process through to be caught up.
+	 */
+	public getCheckpointSequenceNumber?: () => number | undefined = () => this.latestSequenceNumber;
```
CraigMacomber: Again, why is this an optional mutable instance property?

If we expect users of this object to modify/replace/clear this it should document why/when that's expected. If not, it should not use a type which specifically allows that.

CraigMacomber: Note this approach also makes it enumerable. Instead of an enumerable optional mutable instance property, I think you want the opposite in all those regards, an read only required inherited non-enumerable property (aka a method)

server/routerlicious/packages/services-core/src/orderer.ts
```diff
@@ -81,5 +81,16 @@ export interface IOrderer {
...
+
+	/**
+	 * Gets the latest sequence number known to the orderer so a connecting client can determine
+	 * whether it has processed all operations known to the service.
+	 *
+	 * @param tenantId - The tenant that owns the document.
+	 * @param documentId - The document whose latest sequence number should be returned.
+	 * @returns The latest known sequence number, or `undefined` when the orderer cannot provide one.
+	 */
+	getCheckpointSequenceNumber?(tenantId: string, documentId: string): Promise<number | undefined>;
```
CraigMacomber: We keep having mutable optional properties with this name, that make no reference to eachother. Perhaps the others are implementing this interface, win which case duplicating docs on them is counterproductive in my opinion, as you should let them be implicitly inherited, or explicitly inherited, or at least linked if you also need to add extra details.

Interestingly this one is an optional method, not a property. Seems like an odd mismatch between it and the implementations.

This should document the semantics of when its undefined (does that mean there are no checkpoints, to that this feature is unsupported? As this is internal, is there a reason we have to allow this to not always exist?

server/routerlicious/.changeset/tidy-pandas-obey.md
alexvy86: Updated the Tinylicious bump to minor.

server/routerlicious/.changeset/tidy-pandas-obey.md
alexvy86: Updated the title to “Fix checkpoint catch-up bugs.” I also kept the changeset focused on Tinylicious; the other packages are internal implementation details and do not need changeset entries to be published.

server/routerlicious/packages/local-server/src/localOrdererManager.ts
alexvy86: Converted this to a prototype method. The optional marker remains only for type compatibility with the previously exported class: making the method required fails the generated compatibility tests against 7.0.0. Every instance from this version defines the method, and the TSDoc now states that explicitly.

server/routerlicious/packages/memory-orderer/src/localOrderer.ts
alexvy86: Converted this to a prototype method as well, so it is inherited and non-enumerable rather than an allocated mutable instance property. It remains optional in the exported class type solely to preserve compatibility with 7.0.0; every current instance defines it, as documented.

server/routerlicious/packages/services-core/src/orderer.ts
alexvy86: Aligned both implementations with the interface method shape and linked the local manager documentation back to this contract. The shared docs now specify that absence means the backend does not support reporting a checkpoint; when implemented, the method returns a definite number. Optionality is needed both for existing exported class compatibility and for deployed orderer managers that do not support this capability.

server/routerlicious/packages/local-server/src/localOrdererManager.ts
```diff
@@ -76,6 +76,24 @@ export class LocalOrdererManager implements IOrdererManager {
 		return orderer;
 	}
 
+	/**
+	 * Implements
+	 * {@link @fluidframework/server-services-core#IOrdererManager.getCheckpointSequenceNumber}.
+	 *
+	 * @remarks This method is optional in the class type to preserve compatibility with previous
```
ChumpChief: The class is internal, do we need this compatibility or can we jump to required?

server/routerlicious/packages/memory-orderer/src/localOrderer.ts
```diff
@@ -198,6 +233,18 @@ export class LocalOrderer implements IOrderer {
 		return orderer;
 	}
 
+	/**
+	 * Gets the latest sequence number broadcast by this orderer.
+	 *
+	 * @remarks This method is optional in the class type to preserve compatibility with previous
```
ChumpChief: Same here, internal class?

server/routerlicious/packages/tinylicious/src/services/inMemorycollection.ts
```diff
@@ -144,15 +144,23 @@ export class Collection<T> implements ICollection<T> {
...
-				if (query[key].$gt > 0) {
+			// A range bound of 0 is meaningful and must be distinguished from an absent bound. For
+			// example the routerlicious driver asks for the ops of a document from the very beginning
+			// with `{ $gt: 0 }`; treating that as "not a range query" would degrade it to an equality
+			// match and return no ops at all, leaving clients unable to ever catch up.
+			const lowerBound: unknown = query[key].$gt;
+			const upperBound: unknown = query[key].$lt;
+			const hasLowerBound = typeof lowerBound === "number";
+			const hasUpperBound = typeof upperBound === "number";
+			if (hasLowerBound || hasUpperBound) {
```
ChumpChief: This nesting feels like it can be simplified, maybe also avoid the double-filter case by combining into one filter pass?

server/routerlicious/packages/local-server/src/localOrdererManager.ts
```diff
@@ -76,6 +76,24 @@ export class LocalOrdererManager implements IOrdererManager {
...
+	 * versions of the exported class. Every instance created by this version defines it.
+	 */
+	public async getCheckpointSequenceNumber?(
+		tenantId: string,
+		documentId: string,
+	): Promise<number> {
+		const orderer = (await this.getOrderer(tenantId, documentId)) as LocalOrderer;
+		if (orderer.getCheckpointSequenceNumber === undefined) {
```
ChumpChief: Feels like since this class is in charge of producing the orderers, it should be able to know whether they will have this functionality or not.

server/routerlicious/packages/memory-orderer/src/localOrderer.ts
```diff
@@ -179,7 +191,30 @@ export class LocalOrderer implements IOrderer {
...
+					"sequenceNumber" in message &&
+					typeof message.sequenceNumber === "number"
+				) {
+					this.latestSequenceNumber = Math.max(
+						this.latestSequenceNumber,
+						message.sequenceNumber,
+					);
+				}
+			}
+		});
```
ChumpChief: Nit: Maybe consider using `messages.findLast()` if it's sorted, or `messages.reduce()` if not.

server/routerlicious/packages/local-server/src/localOrdererManager.ts
alexvy86: Agreed. The manager now stores `Promise<LocalOrderer>` rather than the broader interface type, so it calls the required method directly without a cast or runtime presence check.

server/routerlicious/packages/memory-orderer/src/localOrderer.ts
alexvy86: I kept the loop intentionally: these batches are not guaranteed to be sorted, and `findLast()` is unavailable with the repo's ES2020 library target. A `reduce()` would not simplify the scan and would still need the same malformed-entry handling.

server/routerlicious/packages/memory-orderer/src/localOrderer.ts
alexvy86: Agreed; this is required now as well, with the intentional `@internal` forward-compatibility break recorded in the generated type-test metadata.

server/routerlicious/packages/tinylicious/src/services/inMemorycollection.ts
alexvy86: Simplified this to a single `filter` call with one predicate that applies both optional bounds.

server/routerlicious/packages/local-server/src/localOrdererManager.ts
alexvy86: Agreed. I made this method required. Since the exported `@internal` class is still covered by our generated forward-compatibility tests, I also recorded the intentional break in `typeValidation.broken`.

ChumpChief (approved):

server/routerlicious/packages/memory-orderer/src/localOrderer.ts
```diff
@@ -198,6 +233,15 @@ export class LocalOrderer implements IOrderer {
 		return orderer;
 	}
 
+	/**
+	 * Gets the latest sequence number broadcast by this orderer.
+	 *
+	 * @returns The sequence number a connecting client must process through to be caught up.
+	 */
+	public getCheckpointSequenceNumber(): number {
+		return this.latestSequenceNumber;
```
znewton: Latest SN is not equal to checkpoint SN typically. Is this because LocalOrderer does not actually perform checkpointing?

server/routerlicious/packages/routerlicious-base/src/nexus/ordererManager.ts
```diff
@@ -129,6 +129,19 @@ export class OrdererManager implements IOrdererManager {
 		return kafkaOrderer;
 	}
 
+	/**
+	 * The orderer manager in Nexus routes messages to Kafka or a local or proxy orderer, but does
+	 * not observe the resulting sequenced operations. An in-process local orderer can track the
+	 * sequence numbers it broadcasts; this manager cannot reliably obtain an equivalent current
```
znewton, line 135: Checkpoints in R11s are stored in a DB, so they _should_ be able to reliably be found. We can't reliably get _latest_ (like you used for local), but latest checkpoint should be findable.

znewton: I'd be happy with this being a no-op (current impl) behind a Feature Flag in the Nexus config, but I want to make sure we're being accurate here with feature capabilities. It seems to me that the same bug in T9s is present in R11s and this is not addressing it, so we should at least _try_ to address it if we can.

alexvy86: Good point. Nexus cannot cheaply determine the exact current head, but R11s should be able to read the latest persisted Deli checkpoint from its checkpoint storage. That lagging value is still consistent with the `checkpointSequenceNumber` wire contract and would improve read-client catch-up behavior.

For this PR, we want to keep Nexus runtime behavior unchanged: returning `undefined` causes the field to remain omitted just as it is today, while this PR fixes Tinylicious. We'll explore wiring the R11s checkpoint into Nexus—likely initially behind a feature flag, with coverage for local/global checkpoint selection and read connections—in a separate PR. I'll also make sure that follow-up treats this as an R11s improvement rather than claiming Nexus is incapable of reporting any useful checkpoint.

server/routerlicious/packages/memory-orderer/src/localOrderer.ts
alexvy86: Correct: `LocalOrderer` does perform Deli checkpointing, but this value intentionally is not the persisted checkpoint. Because the local orderer directly observes the sequenced operations it broadcasts, it can provide the latest sequence number it has observed as a fresher (and stronger) catch-up fence. The existing `checkpointSequenceNumber` wire field permits such a value even though its name suggests a persisted checkpoint; its contract is the service's last known sequence number and allows it to lag the actual head.

server/routerlicious/packages/routerlicious-base/src/nexus/ordererManager.ts
alexvy86: Also of note, I think R11s suffers from one of the two issues we fixed for tinylicious here, i.e. not providing a `checkpointSequenceNumber` in the connect message to clients. But if I understood things right, the driver then submits a request for all ops, which R11s handles correctly and tinylicious didn't. So clients hitting R11s still behave correctly, while clients hitting t9s never got the latest ops.

znewton (approved):

PR #27938, opened 2026-08-12
fix(tree): preserve stored keys and staged optionals in schema compatibility snapshots

daesunp:
## Description

Updates schema compatibility snapshots to use persisted stored keys instead of property keys and preserve staged-optional fields, preventing property renames from appearing as schema changes.

Existing V1 snapshots remain supported, while `VersionDispatchingCodecBuilder` selects the appropriate format and enables compatible snapshots to be normalized to v2.


.changeset/tall-clocks-scream.md
```diff
@@ -0,0 +1,9 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+
```
Josmithr:
```suggestion
```

.changeset/tall-clocks-scream.md
```diff
@@ -0,0 +1,9 @@
...
+Schema compatibility snapshots ignore property keys
```
Josmithr: I would probably rephrase this as 
```suggestion
Schema compatibility snapshots use stored keys instead of ephemeral property keys for objects
```

.changeset/tall-clocks-scream.md
```diff
@@ -0,0 +1,9 @@
...
+
+The schema compatibility snapshot format now keys object fields by their persisted stored keys, so renaming a developer-facing property no longer appears to change schema compatibility. Existing version 1 snapshots remain supported, and the new normalize mode can rewrite the latest compatible snapshot into the current format to minimize future diffs.
```
Josmithr, line 8: We should add guidance for developers who want to migrate from v1 to v2.

packages/dds/tree/src/simple-tree/api/snapshotCompatibilityChecker.ts
```diff
@@ -802,6 +821,19 @@ export class SnapshotCompatibilityChecker {
 	}
 }
 
+function getSnapshotFormatVersion(snapshot: JsonCompatibleReadOnly): number {
```
Josmithr, line 834: Nit: docs?

packages/dds/tree/src/simple-tree/api/simpleSchemaCodec.ts
```diff
@@ -104,6 +129,43 @@ export function decodeSchemaCompatibilitySnapshot(
 	};
 }
 
+function upgradeFormat(
```
Josmithr, line 141: Nit: docs?

Josmithr: (Same suggestion for the other new functions here)

packages/dds/tree/src/test/simple-tree/api/snapshotCompatibilityChecker.spec.ts
```diff
@@ -41,6 +45,56 @@ const nodeFileSystem = {
 	...path,
 };
 
+function isJsonObject(
```
Josmithr: Nit: do we have an existing helper somewhere for this? If not, should we define this next to `JsonCompatibleReadonlyObject` so others can use it as needed in the future?

packages/dds/tree/src/simple-tree/api/snapshotCompatibilityChecker.ts
```diff
@@ -524,14 +528,17 @@ export function snapshotSchemaCompatibility(
...
-			`Invalid mode: ${JSON.stringify(mode)}. Must be either "assert" or "update".`,
+			`Invalid mode: ${JSON.stringify(mode)}. Must be "assert", "update", or "normalize".`,
 		);
 	}
 
 	const currentEncodedForSnapshotting = exportCompatibilitySchemaSnapshot(currentViewSchema);
 	const snapshots = checker.readAllSchemaSnapshots(versionComparer);
+	const rawSnapshots = transformMapValues(snapshots, (_snapshot, version) =>
+		checker.readSchemaSnapshotRaw(version),
+	);
```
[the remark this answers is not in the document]
daesunp: replaced eager rereading/parsing of snapshots with cache.

.changeset/tall-clocks-scream.md
daesunp: Updated with suggestion

.changeset/tall-clocks-scream.md
daesunp: Updated with suggestion

.changeset/tall-clocks-scream.md
daesunp: Updated with instructions for migration

packages/dds/tree/src/simple-tree/api/snapshotCompatibilityChecker.ts
daesunp: Added docs

packages/dds/tree/src/simple-tree/api/simpleSchemaCodec.ts
daesunp: Added docs

packages/dds/tree/src/test/simple-tree/api/snapshotCompatibilityChecker.spec.ts
daesunp: Removed and used existing `isJsonObject` utility

CraigMacomber:
PR titles should generally say what the PR does. "schema compat property keys" does not do this. I suggest a name like "Fix tree schema compatibility snapshots to always use stored keys"

.changeset/tall-clocks-scream.md
```diff
@@ -0,0 +1,10 @@
+---
+"@fluidframework/tree": minor
+"fluid-framework": minor
+"__section": tree
+---
+Schema compatibility snapshots use stored keys instead of ephemeral property keys for objects
```
CraigMacomber, line 6:
```suggestion
New schema compatibility snapshots will use stored keys instead of property keys
```

packages/dds/tree/src/simple-tree/api/schemaFromSimple.ts
```diff
@@ -120,10 +120,12 @@ function generateNode(
...
-			const fields: Record<string, FieldSchema> = {};
-			for (const [key, field] of schema.fields) {
-				fields[key] = generateFieldSchema(field, context, field.storedKey);
-			}
+			const fields: Record<string, FieldSchema> = Object.fromEntries(
+				Array.from(schema.fields, ([key, field]) => [
+					key,
+					generateFieldSchema(field, context, field.storedKey),
+				]),
+			);
```
CraigMacomber, line 128: This change seems like a refactor, unrelated to the new format.

Is this intended to have any impact? If not, what's the motivation for this change.

packages/dds/tree/src/simple-tree/api/simpleSchemaCodec.ts
```diff
@@ -85,15 +86,45 @@ export function decodeSchemaCompatibilitySnapshot(
 	validator?: FormatValidator,
 ): SimpleTreeSchema {
 	const effectiveValidator = validator ?? FormatValidatorNoOp;
-	const compiledValidator = extractJsonValidator(effectiveValidator).compile(
-		Format.SimpleTreeSchemaFormat,
+	const jsonValidator = extractJsonValidator(effectiveValidator);
+	const formatVersion = getFormatVersion(encodedSchema);
+	if (formatVersion === Format.SimpleSchemaFormatVersion.v2) {
```
CraigMacomber, line 91: Rather then implementing your own custom logic for detecting versions and selecting the right decoder, please use VersionDispatchingCodecBuilder.

It does a little more than we need, since we only ever need to encode v2, and don't need min version for collab logic, but we can simply hard some values for the min version for colab, and it should do what we need, including nice error messages for version from the future and such.

packages/dds/tree/src/simple-tree/api/simpleSchemaCodec.ts
```diff
@@ -187,30 +258,18 @@ function encodeSimpleAllowedTypes(
...
-
 	return {
 		kind: schema.kind,
-		fields: encodedFields,
+		fields: Object.fromEntries(
+			Array.from(schema.fields.values(), (fieldSchema) => [
+				fieldSchema.storedKey,
+				encodeField(fieldSchema),
+			]),
+		),
```
CraigMacomber, line 268: Why are you converting this from the old style? This seems like a much larger change than is needed, which makes the actual change much harder to see. Couldn't this have just changed `encodedFields[fieldKey] = encodeObjectField(fieldSchema);` to `encodedFields[fieldSchema.storedKey] = encodeField(fieldSchema);`

Probably would be nicer also include a comment noting that this intentionally only preserves the stored key, which is just one of the ways this encoding is lossy.

packages/dds/tree/src/simple-tree/api/simpleSchemaCodec.ts
```diff
@@ -85,15 +86,45 @@ export function decodeSchemaCompatibilitySnapshot(
...
-	if (!compiledValidator.check(encodedSchema)) {
-		throw new UsageError(
-			"The provided simple schema is not valid according to the schema format.",
-		);
+}
+
+/**
+ * Gets the format version from an encoded schema, if present.
+ */
+function getFormatVersion(encodedSchema: JsonCompatibleReadOnly): number | undefined {
```
CraigMacomber, line 111: Didn't you implement this twice, see getSnapshotFormatVersion. Both should be redundant with the logic in VersionDispatchingCodecBuilder I think.

packages/dds/tree/src/test/simple-tree/api/snapshotCompatibilityChecker.spec.ts
```diff
@@ -41,6 +43,50 @@ const nodeFileSystem = {
 	...path,
 };
 
+function convertObjectSnapshotToV1(
```
CraigMacomber, line 46: I don't think we should need this.

I think it would make more sense more directly test support for loading v1 snapshots by simply hard coding some minimal ones in unit tests to validate it.

packages/dds/tree/src/test/simple-tree/api/snapshotCompatibilityChecker.spec.ts
```diff
@@ -295,6 +341,203 @@ Snapshots exist for versions: [
...
+			class Renamed extends factory.object("User", {
+				displayName: factory.required(factory.string, { key: "name" }),
+			}) {}
+
+			class ChangedStoredKey extends factory.object("User", {
+				displayName: factory.string,
+			}) {}
+
+			class PrototypeStoredKey extends factory.object("PrototypeStoredKey", {
+				value: factory.required(factory.string, { key: "__proto__" }),
```
CraigMacomber, line 360: Ah, I'm starting to see why maybe you changed how we construct our objects!

That really seems like a separate issue from this PR which should get its own separate tests and likely separate PR. Adding a test schema in packages/dds/tree/src/test/testTrees.ts with stored key "__proto__" (likely after I land https://github.com/microsoft/FluidFramework/pull/28200 ) would a cleaner way to do that I think.

Also if you are testing this special magic string, a comment on why this is being testing in this specific code/test would  be a good idea.

.changeset/tall-clocks-scream.md
daesunp: Updated changeset

packages/dds/tree/src/simple-tree/api/schemaFromSimple.ts
daesunp: restored after `__proto__` related changes are being factored out into a separate follow up PR

packages/dds/tree/src/simple-tree/api/simpleSchemaCodec.ts
daesunp: Updated to use `VersionDispatchingCodecBuilder`

packages/dds/tree/src/simple-tree/api/simpleSchemaCodec.ts
daesunp: restored after `__proto__ `related changes are being factored out into a separate follow up PR

packages/dds/tree/src/simple-tree/api/simpleSchemaCodec.ts
daesunp: Updated to use `VersionDispatchingCodecBuilder`

packages/dds/tree/src/test/simple-tree/api/snapshotCompatibilityChecker.spec.ts
daesunp: removed after `__proto__` related changes are being factored out into a separate follow up PR

packages/dds/tree/src/test/simple-tree/api/snapshotCompatibilityChecker.spec.ts
daesunp: removed `__proto__` specific tests and implementation changes from this PR

PR #27941, opened 2026-08-12
Preserve SharedTree no-change constraints during refresher updates

Abe27342:
## Description

SharedTree recomputes refresher data when pending edits are resubmitted after reconnecting. `updateRefreshers` reconstructed the modular changeset without preserving its no-change constraints, which allowed a constrained transaction to be applied after a concurrent sequenced edit.

This change preserves `noChangeConstraint` and `noChangeConstraintOnRevert` while updating refreshers. It adds a focused unit test for the reconstructed changeset and a local-server end-to-end test covering the disconnected edit and resubmit flow.

## Reviewer Guidance

Please verify that preserving the global no-change constraint fields is the appropriate behavior whenever refresher data is recomputed.

### DIFF
--- a/packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
+++ b/packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
@@ -1673,34 +1673,9 @@ export function updateRefreshers(
 		}
 	}
 
-	const {
-		fieldChanges,
-		nodeChanges,
-		nodeToParent,
-		nodeAliases,
-		crossFieldKeys,
-		maxId,
-		revisions,
-		constraintViolationCount,
-		noChangeConstraint,
-		noChangeConstraintOnRevert,
-		builds,
-		destroys,
-	} = change;
-
 	return makeModularChangeset({
-		fieldChanges,
-		nodeChanges,
-		nodeToParent,
-		nodeAliases,
-		crossFieldKeys,
-		maxId: maxId as number,
-		revisions,
-		constraintViolationCount,
-		noChangeConstraint,
-		noChangeConstraintOnRevert,
-		builds,
-		destroys,
+		...change,
+		maxId: change.maxId,
 		refreshers,
 	});
 }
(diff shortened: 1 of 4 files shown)


packages/dds/tree/src/feature-libraries/modular-schema/modularChangeFamily.ts
yann-achard-MS: We probably should transition to this:
```suggestion
	return makeModularChangeset({
		...change,
		maxId: change.maxId,
		refreshers,
	});
```
You'll have to update `makeModularChangeset` so that `maxId` in `props` is still a required key, but one that can be undefined:
`maxId: number | undefined;`

yann-achard-MS: Note: looks like the diff for the suggested change is bad, so applying the suggestion may not work out.

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeUtils.ts
```diff
@@ -53,7 +53,7 @@ export function makeModularChangeset(props?: {
 	nodeToParent?: ChangeAtomIdBTree<FieldId>;
 	nodeAliases?: ChangeAtomIdBTree<NodeId>;
 	crossFieldKeys?: CrossFieldKeyTable;
-	maxId: number;
+	maxId: number | undefined;
```
Josmithr: Nit: I realize none of the param properties here are documented, but can we document what this is and what undefined means for it?

Josmithr:
Should this have a changeset?

Abe27342:
> Should this have a changeset?

Depends how noisy we want to be. It fixes a bug that should only be possible to hit if you're consuming alpha APIs. No expected change on the consuming side, either. I thought about it but think "no" is reasonable

yann-achard-MS (approved):

packages/dds/tree/src/feature-libraries/modular-schema/modularChangeUtils.ts
Abe27342: I documented the top-level function instead and tweaked parameter typing to make things more clear (semantics of all of these fields should inherit semantics of the same fields in `ModularChangeset`, which is now more obviously the case)

Josmithr: Thanks!

PR #27943, opened 2026-08-12
build(eslint-config-fluid): consolidate custom rules into config package

alexvy86:
## Description

Moves the custom Fluid ESLint rules into `@fluidframework/eslint-config-fluid`, removing the separate `@fluid-internal/eslint-plugin-fluid` package and release pipeline. Existing `@fluid-internal/fluid/*` rule names are preserved, so consumers of the shared config do not need to migrate rule overrides or suppression comments.

The config package is updated to 14.1.0 and now owns the rule dependencies, documentation, and ESLint 8/9 test coverage. Repository package metadata and pipeline-trigger guidance no longer reference the removed plugin.

Implements [AB#79903](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/79903).

## Reviewer Guidance

The npm deprecation of `@fluid-internal/eslint-plugin-fluid` should happen after `@fluidframework/eslint-config-fluid` 14.1.0 is published.


common/build/eslint-config-fluid/src/test/fluid-rules.test.mts
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { strict as assert } from "node:assert";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+import { recommended } from "../../flat.mjs";
+import { createESLintForConfig } from "./eslintConfigHelper.mjs";
```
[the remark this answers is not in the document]
alexvy86: This is intentional. The package consistently uses `.mjs` specifiers for `.mts` sources (including the existing `no-only-tests.test.mts`), and the integration test passes under the package's Mocha setup.

common/build/eslint-config-fluid/src/rules/test/eslintConfigHelper.cjs
```diff
@@ -14,7 +14,7 @@ const path = require("node:path");
 const eslintPackage = process.env.ESLINT_PACKAGE || "eslint";
```
Josmithr, line 14: Could we move the tests for rules under `rules/test` instead of `test/rules`? I personally prefer tests being as close to the source code they validate as possible.

Josmithr (approved):

alexvy86:
While smoke-testing the packed `@fluidframework/eslint-config-fluid@14.1.0` package against the client release group, I found two React-rule compatibility issues that are unrelated to this package-consolidation change:

- `@eslint-react/eslint-plugin` 5.9 removes `@eslint-react/jsx-key-before-spread`; the replacement `@eslint-react/jsx-no-key-after-spread` is already enabled by its `recommended-typescript` preset, so the explicit old-rule overrides in `inventory-app` and `text-editor` will need to be removed.
- The 5.9 preset also enables `@eslint-react/rules-of-hooks`, duplicating the existing `react-hooks/rules-of-hooks` enforcement. The `@eslint-react` copy should be disabled so the repo continues using the canonical `eslint-plugin-react-hooks` rule and its existing suppressions.

These failures come from the pre-existing 14.0 React-plugin upgrade and only surface when the client moves off config 13.x. They are not caused by consolidating `eslint-plugin-fluid` into this package, so the fixes were intentionally left out of this PR and should be handled as part of the client release group's 14.1 integration.

Josmithr (approved):

common/build/eslint-config-fluid/src/rules/test/eslintConfigHelper.cjs
alexvy86: Done. I moved the rule unit tests and fixtures from `src/test/rules` to `src/rules/test`, updated the test scripts and helper paths, and kept the exported-config integration test under `src/test`. The ESLint 8 and 9 suites both pass with the new layout.

Abe27342 (approved):

PR #27946, opened 2026-08-12
feat(tree): Throw descriptive TypeError for unsupported Array/Map prototype methods

jenn-le:
## Description

When calling standard `Array.prototype` or `Map.prototype` methods that aren't supported on `ArrayNode` or `MapNode` (e.g., `copyWithin`, `fill`, `reverse`, `sort`), users previously got a generic "is not a function" error. This is especially confusing for LLM-generated JavaScript code that may attempt to use these standard methods.

The proxy `get` traps now detect access to unsupported prototype methods and return a function that throws a `TypeError` with a descriptive message suggesting the correct ArrayNode/MapNode API to use instead. For example:

```
TypeError: ArrayNode does not support 'fill'. Use the ArrayNode API (e.g., insertAt, removeAt, moveToIndex, splice).
```

For MapNode, the same pattern is applied as a safety net for any future `Map.prototype` additions that diverge from the MapNode API.

## Reviewer Guidance

The detection is generic — it checks against the actual `Array.prototype`/`Map.prototype`, so future additions to these prototypes are automatically covered. Methods that ArrayNode already implements (like `push`, `pop`, `splice`) are on the dispatch target and are unaffected.


packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -1423,69 +1423,74 @@ describe("ArrayNode", () => {
...
+	describeHydration("unsupported Array.prototype methods", (init) => {
+		it("throws descriptive TypeError for copyWithin", () => {
+			const array = init(PojoEmulationNumberArray, [1, 2, 3]);
+			const copyWithin = (array as unknown as Record<string, unknown>).copyWithin as (
+				...args: unknown[]
+			) => unknown;
+			assert.equal(typeof copyWithin, "function");
+			assert.throws(
+				() => copyWithin.call(array, 0, 1),
+				(error: Error) => error instanceof TypeError && error.message.includes("copyWithin"),
```
Josmithr: Nit: this message check seems a bit weak to me. Might be worth validating the entire message, or at least the first part of it. We can always update the tests if/when we update the error messages.

Josmithr: (Same feedback for the other tests)

packages/dds/tree/src/test/simple-tree/mapNode.spec.ts
```diff
@@ -347,3 +281,48 @@ describeHydration(
...
+			assert.equal(typeof method, "function");
+			assert.throws(
+				() => method.call(node),
+				(error: Error) =>
+					error instanceof TypeError &&
+					error.message.includes(fakeMethodName) &&
+					error.message.includes("MapNode"),
+			);
+		} finally {
+			// eslint-disable-next-line @typescript-eslint/no-dynamic-delete
```
Josmithr: Nit: can we add a quick explanation to the disable comment?

Josmithr (approved):

packages/dds/tree/src/simple-tree/node-kinds/map/mapNode.ts
```diff
@@ -189,6 +189,27 @@ const handler: ProxyHandler<TreeMapNodeAlpha<any>> = {
...
+		}
+		// If the property is a function on Map.prototype but not on the target,
+		// return a function that throws a descriptive TypeError when called.
+		if (
+			typeof key === "string" &&
+			!(key in target) &&
+			key in Map.prototype &&
+			typeof (Map.prototype as unknown as Record<string, unknown>)[key] === "function"
+		) {
+			return (...args: unknown[]) => {
```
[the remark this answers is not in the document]
jenn-le: Good catch -- removed the unused `...args` parameter. The array version already didn't have it.

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
jenn-le: Good point. Updated all tests to validate the exact error message string instead of just checking for substring includes.

packages/dds/tree/src/test/simple-tree/mapNode.spec.ts
jenn-le: Good call. Added explanations to the eslint-disable comments in both test files -- they are cleaning up test-only methods added to the prototype.

packages/dds/tree/src/test/simple-tree/mapNode.spec.ts
```diff
@@ -345,5 +321,6 @@ describeHydration(
 				assert.deepEqual([...fromRecord.data], data);
 			});
 		});
+
```
Josmithr: Nit
```suggestion
```

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -1423,71 +1423,67 @@ describe("ArrayNode", () => {
...
-				validateUsageError(/non-negative/),
-			);
-		});
+		} finally {
+			// eslint-disable-next-line @typescript-eslint/no-dynamic-delete -- Cleaning up test-only method added to Array.prototype
+			delete (Array.prototype as unknown as Record<string, unknown>)[methodName];
+		}
+	});
+
+	it("does not interfere with existing ArrayNode methods", () => {
```
Josmithr: I'm not sure this test offers real value - we already have lots of tests that use the existing methods.

packages/dds/tree/src/test/simple-tree/node-kinds/arrayNode.spec.ts
```diff
@@ -1423,71 +1423,67 @@ describe("ArrayNode", () => {
...
-		it("tracks across a move within the array", () => {
-			const array = init(CustomizableNumberArray, [10, 20, 30, 40, 50]);
-			const anchor = createArrayInsertionAnchor(array, 2); // gap between 20 and 30
-			array.moveToIndex(0, 4); // move 50 to the front -> [50, 10, 20, 30, 40]
-			// The moved node is inserted before the anchor (shifting it right by one); its removal is
-			// after the anchor and so does not move it.
-			assert.equal(anchor.index, 3); // still in the gap between 20 and 30
-			anchor.dispose();
-		});
+	it("throws descriptive TypeError for an unsupported Array.prototype method", () => {
```
Josmithr, line 1493: Looks like we lost the map version of this test in the most recent changes?

Josmithr (approved):

Josmithr (approved):

PR #27947, opened 2026-08-12
fix(merge-tree): visit the starting reference in walkReferences

ChumpChief:
## Description

`LocalReferenceCollection.walkReferences` never visited the reference passed as `start`.

References in a collection are stored per-offset in three buckets (`before`/`at`/`after`). When `start` is supplied, the walk is supposed to discard the buckets preceding the one containing `start`, then resume mid-bucket at `start` itself. Instead, the guard compared `offsetPositions.first`/`.last` — `ListNode`s of the *outer* bucket list, typed `ListNode<DoublyLinkedList<LocalReference> | undefined>` — against `startNode`, a `ListNode<LocalReference>` from an *inner* bucket. Those are never the same object, so the loop drained `offsetPositions` entirely and the walk began at the next offset.

The effect: a walk with an explicit `start` skipped `start` and every other reference at its offset. It also made the `startList === pos ? startNode : undefined` argument passed to `walkList` unreachable — that line is the clearest evidence this was a typo rather than intent, since it only makes sense if the loop stops at the bucket holding `start`.

The fix compares bucket to bucket (`offsetPositions.first?.data !== startList`), so the loop stops at the right bucket and `walkList` resumes from `startNode`.

### Why this went unnoticed

- **It type-checks.** Because `before`/`at`/`after` are optional, both operands are unions containing `undefined`, which gives them enough overlap to slip past TypeScript's `TS2367` no-overlap check. Without the `undefined` on both sides the same comparison is a compile error.
- **No caller exercises it.** Every in-repo caller (`revertibles.ts`, `mergeTree.ts`) passes `undefined` for `start`, and the method is `@internal`, so the parameter has been effectively dead since it was written.

The typo has been present since `walkReferences` was introduced in #11899 (Sept 2022); the lines have only seen formatting churn since.

### Testing

Adds `packages/dds/merge-tree/src/test/localReference.spec.ts` with 13 tests covering:

- Forward and backward walks with and without `start`, mid-bucket resume, `start` at the first and last offsets, and early termination via the visitor.
- Offsets holding **multiple buckets** — a `before` bucket coexisting with `at` at offset 0, and an `after` bucket coexisting with `at` at the last offset, populated through `addBeforeTombstones`/`addAfterTombstones`. These exercise the case the fix is really about: the discard loop must skip sibling buckets at the same offset before stopping at the one holding `start`. Walks starting from each of the three buckets are checked in both directions.

Ten of the thirteen tests fail against the unfixed comparison, including all four multi-bucket `start` cases. The full merge-tree suite (1568 tests) passes.

## Reviewer Guidance

### DIFF
--- a/.changeset/great-walls-walk.md
+++ b/.changeset/great-walls-walk.md
@@ -1,0 +1,14 @@
+---
+"@fluidframework/merge-tree": patch
+"__section": fix
+---
+Fix `LocalReferenceCollection.walkReferences` skipping the starting reference
+
+`LocalReferenceCollection.walkReferences` compared the wrong objects when locating the
+before/at/after bucket containing the supplied `start` reference. It compared list nodes of the
+outer bucket list against the list node of `start` itself, which never match, so the walk discarded
+every bucket at `start`'s offset and began at the next offset instead. As a result, a walk with an
+explicit `start` never visited `start` or any other reference at its offset.
+
+The walk now stops at the bucket containing `start` and resumes from `start` within that bucket, so
+`start` is visited first in both the forward and backward directions.
--- a/.changeset/great-walls-walk.md
+++ b/.changeset/great-walls-walk.md
@@ -2,7 +2,7 @@
 "@fluidframework/merge-tree": patch
 "__section": fix
 ---
-Fix `LocalReferenceCollection.walkReferences` skipping the starting reference
+Fix LocalReferenceCollection.walkReferences skipping the starting reference
 
 `LocalReferenceCollection.walkReferences` compared the wrong objects when locating the
 before/at/after bucket containing the supplied `start` reference. It compared list nodes of the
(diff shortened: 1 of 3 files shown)


.changeset/great-walls-walk.md
```diff
@@ -0,0 +1,14 @@
+---
+"@fluidframework/merge-tree": patch
+"__section": fix
+---
+Fix `LocalReferenceCollection.walkReferences` skipping the starting reference
```
Josmithr: Our changeset guidelines call to avoid using `inlineCode` syntax in changeset titles.
```suggestion
Fix LocalReferenceCollection.walkReferences skipping the starting reference
```

ChumpChief: Fixed in latest

shlevari (approved):

anthony-murphy (approved):

Josmithr (approved):

PR #27948, opened 2026-08-12
feat(container-runtime): Default document schema to declare createBlobPayloadPending support at 2.40.0+

dannimad:
## Description

When `minVersionForCollab` is set to `"2.40.0"` or later, the document schema now declares support for the
`createBlobPayloadPending` capability by default, instead of requiring callers to opt in explicitly.

Importantly, this only affects what gets negotiated into the document schema (which all collaborating clients
must understand) — it does **not** change any observable behavior. `BlobManager`'s pending-payload upload
behavior remains off unless a caller explicitly sets `createBlobPayloadPending: true` in their
`ContainerRuntimeOptions`. This split is deliberate: a document schema declaration is effectively permanent
once other clients observe it in a document, whereas client-side behavior can be safely rolled forward or back
via a package update. Pre-enabling the schema declaration now means documents will already be schema-compatible
by the time we're ready to turn on the corresponding client behavior by default in a future release, at which
point that change becomes a simple code change rather than a slow, fleet-wide document schema migration.

## Reviewer Guidance

- Please double check the reasoning for decoupling the schema default from the client-side behavior default —
  documented in `CrossClientCompatibilityDevGuide.md` as a pattern to reuse for future schema-affecting options.
- `containerRuntime.ts` tracks whether `createBlobPayloadPending` was explicitly requested (mirroring the
  existing `enableRuntimeIdCompressor` explicit-vs-defaulted handling) and only that explicit value gates
  `BlobManager`'s actual behavior.

### DIFF
--- a/packages/runtime/container-runtime/src/containerCompatibility.ts
+++ b/packages/runtime/container-runtime/src/containerCompatibility.ts
@@ -157,10 +157,15 @@ const runtimeOptionsAffectingDocSchemaConfigMap: ConfigMap<RuntimeOptionsAffecti
 			// "3.0.0": { enableGCSweep: true },
 		},
 		createBlobPayloadPending: {
-			// This feature is new and disabled by default. In the future we will enable it by default, but we have not
-			// closed on the version where that will happen yet.  Probably a .10 release since blob functionality is not
-			// exposed on the `@public` API surface.
+			// NOTE: This default only controls what gets declared/negotiated in the document schema (i.e. whether
+			// documents record that clients here understand the pending-blob-payload format). It does NOT turn on
+			// the actual pending-payload behavior in BlobManager - that remains opt-in (see the "explicit vs
+			// defaulted" handling in containerRuntime.ts) until we have higher confidence in the client-side
+			// behavior. Pre-enabling the schema declaration now means documents will already be schema-compatible
+			// by the time we're ready to flip the behavior default too, which then becomes a simple code change
+			// rather than requiring a slow, fleet-wide document schema migration.
 			"1.0.0": undefined,
+			"2.40.0": true,
 		},
 	};
 
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -481,8 +481,18 @@ export interface ContainerRuntimeOptions {
 	readonly explicitSchemaControl: boolean;
 
 	/**
-	 * Create blob handles with pending payloads when calling createBlob (default is `undefined` (disabled)).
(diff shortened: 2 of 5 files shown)


packages/runtime/container-runtime/src/containerCompatibility.ts
```diff
@@ -157,10 +157,15 @@ const runtimeOptionsAffectingDocSchemaConfigMap: ConfigMap<RuntimeOptionsAffecti
...
-			// exposed on the `@public` API surface.
+			// NOTE: This default only controls what gets declared/negotiated in the document schema (i.e. whether
+			// documents record that clients here understand the pending-blob-payload format). It does NOT turn on
+			// the actual pending-payload behavior in BlobManager - that remains opt-in (see the "explicit vs
+			// defaulted" handling in containerRuntime.ts) until we have higher confidence in the client-side
+			// behavior. Pre-enabling the schema declaration now means documents will already be schema-compatible
+			// by the time we're ready to flip the behavior default too, which then becomes a simple code change
+			// rather than requiring a slow, fleet-wide document schema migration.
 			"1.0.0": undefined,
+			"2.40.0": true,
```
ChumpChief, line 168: I think this should just be 3.0.0 and drop the "explicitly requested" aspect.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1656,6 +1679,10 @@ export class ContainerRuntime
 		featureGatesForTelemetry: Record<string, boolean | number | undefined>,
 		provideEntryPoint: (containerRuntime: IContainerRuntime) => Promise<FluidObject>,
 		public readonly minVersionForCollab: MinimumVersionForCollab,
+		// Whether the caller explicitly requested `createBlobPayloadPending` behavior (as opposed to it merely
+		// being the defaulted value for the document schema declaration). See the comment where this is computed
+		// in `loadRuntime2` for why this is tracked separately from `runtimeOptions.createBlobPayloadPending`.
+		private readonly createBlobPayloadPendingExplicitlyRequested: boolean,
```
ChumpChief, line 1685: This feature should remain driven purely on the document schema.

ChumpChief (commented):
We shouldn't introduce a matrix of schema+explicit request.  Just enabling at 3.0.0 should be sufficient.

packages/runtime/container-runtime/src/containerRuntime.ts
dannimad: @anthony-murphy maybe I misunderstood you. What did you mean by "schema change" vs "behavior change". Looking at this change more closely, it would just make the feature enablement/disablement more ambiguous, but that was the way I mapped that schema/behavior distinction.

PR #27949, opened 2026-08-12
chore: remove Vale changeset linting

Abe27342:
## Description

Removes the `@vvago/vale` development dependency and the repository integration built around it. This includes the local changeset-check script, Vale configuration and vocabulary, pnpm build allowance and lockfile entries, and the GitHub workflow that ran `errata-ai/vale-action` against changesets.

After this change, changeset Markdown is no longer linted by Vale locally or in pull requests.


pnpm-lock.yaml
```diff
@@ -26912,15 +26861,6 @@ packages:
     deprecated: Rimraf versions prior to v4 are no longer supported
     hasBin: true
 
-  rimraf@2.7.1:
```
CraigMacomber, line 26915: Wow, vale pulled in two additional old rimraf versions? That's a lot of deps gone :)

CraigMacomber (approved):
That's 20 dependencies gone, not including the ones downloaded at post install time.

If having vale had no cost, I'd say keep it, but it has a lot of false positives the way we use it, a lot of deps, and come rather concerning post install steps. I think it makes sense to remove.

alexvy86 (approved):

PR #27950, opened 2026-08-12
fix(tree): improve schema errors

jenn-le:
## Description

Improves SharedTree schema diagnostics while protecting potentially sensitive document data.

Schema validation errors now:

- Report a stable mismatch category in production.
- Attach diagnostic values as structured telemetry properties.
- Tag paths and field names as `UserData`.
- Tag schema and node identifiers as `CodeArtifact`.
- Include detailed, actionable messages only in debug builds, where they are reconstructed lazily from the structured properties and stripped from production bundles.

When `TreeView.compatibility.canView` is false, accessing `TreeView.root` also reports the first stored/view schema mismatch and recommends initialization, schema upgrade, a compatible view schema, or explicit migration as appropriate.

Tests cover telemetry categorization, production and debug messages, nested validation paths, and schema compatibility states.

## Reviewer Guidance

The validation changes are primarily in `schemaChecker.ts`. In particular, review the classification of each diagnostic property as `UserData`, `CodeArtifact`, or safe untagged data.

### DIFF
--- a/packages/dds/tree/src/feature-libraries/schemaChecker.ts
+++ b/packages/dds/tree/src/feature-libraries/schemaChecker.ts
@@ -84,9 +84,10 @@ function formatSchemaValidationError(
 	error: SchemaValidationError,
 	context: SchemaValidationErrorContext | undefined,
 ): string {
-	const nodeDesc = context?.nodeType === undefined ? "unknown" : `"${context.nodeType}"`;
+	const nodeDesc =
+		context?.nodeType === undefined ? "unknown" : JSON.stringify(context.nodeType);
 	const fieldKindDesc =
-		context?.fieldKind === undefined ? "unknown" : `"${context.fieldKind}"`;
+		context?.fieldKind === undefined ? "unknown" : JSON.stringify(context.fieldKind);
 	const childCountDesc = context?.childCount ?? "unknown";
 	switch (error) {
 		case SchemaValidationError.Field_KindNotInSchemaPolicy: {
@@ -119,7 +120,7 @@ function formatSchemaValidationError(
 					? "unknown"
 					: `[${[...context.allowedTypes]
 							.sort()
-							.map((type) => `"${type}"`)
+							.map((type) => JSON.stringify(type))
 							.join(", ")}]`;
 			return (
 				`Tree does not conform to schema. ` +
@@ -146,7 +147,7 @@ function formatSchemaValidationError(
 					? "unknown"
 					: `[${[...context.unexpectedFields]
 							.sort()
-							.map((field) => `"${field}"`)
+							.map((field) => JSON.stringify(field))
 							.join(", ")}]`;
 			return (
 				`Tree does not conform to schema. ` +
@@ -174,7 +175,7 @@ function formatSchemaValidationError(
 					? undefined
 					: `[${[...context.unexpectedFields]
 							.sort()
-							.map((field) => `"${field}"`)
+							.map((field) => JSON.stringify(field))
 							.join(", ")}]`;
 			const fieldsPart =
 				fieldsDesc === undefined
--- a/packages/dds/tree/src/feature-libraries/schemaChecker.ts
+++ b/packages/dds/tree/src/feature-libraries/schemaChecker.ts
@@ -34,163 +34,34 @@ export enum SchemaValidationError {
 }
 
 /**
- * Additional context about a schema validation error.
- * @remarks
- * Not all fields are populated for every error type — only those relevant to the specific error.
+ * Information about a schema validation error.
  */
-export interface SchemaValidationErrorContext {
-	/**
-	 * The type identifier of the node that failed validation.
-	 */
-	readonly nodeType?: string;
-	/**
-	 * The set of type identifiers allowed in the field, if applicable.
-	 */
-	readonly allowedTypes?: ReadonlySet<string>;
-	/**
-	 * Field keys on the node that were not recognized by its schema.
-	 */
-	readonly unexpectedFields?: ReadonlySet<string>;
-	/**
-	 * The field kind identifier, if the error applies to a field.
-	 */
-	readonly fieldKind?: string;
-	/**
-	 * The number of children found in the field, if its multiplicity is invalid.
-	 */
-	readonly childCount?: number;
-	/**
-	 * The value schema required by a leaf node.
-	 */
-	readonly expectedValueSchema?: ValueSchema;
-	/**
-	 * A description of the runtime type of an invalid node value.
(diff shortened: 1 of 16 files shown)


packages/dds/tree/src/feature-libraries/schemaChecker.ts
```diff
@@ -67,34 +84,98 @@ function formatSchemaValidationError(
 	error: SchemaValidationError,
 	context: SchemaValidationErrorContext | undefined,
 ): string {
+	const nodeDesc = context?.nodeType === undefined ? "unknown" : `"${context.nodeType}"`;
+	const fieldKindDesc =
+		context?.fieldKind === undefined ? "unknown" : `"${context.fieldKind}"`;
+	const childCountDesc = context?.childCount ?? "unknown";
```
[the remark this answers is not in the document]
jenn-le: Accepted. Updated node type, field kind, allowed type, and unexpected field formatting to use JSON.stringify, with a regression test covering quotes and newlines.

packages/dds/tree/src/feature-libraries/schemaChecker.ts
```diff
@@ -67,34 +84,98 @@ function formatSchemaValidationError(
...
+			return (
+				`Tree does not conform to schema. ` +
+				`A field of kind ${fieldKindDesc} allows at most one child, but found ${childCountDesc}.`
+			);
+		}
+		case SchemaValidationError.Field_ChildInForbiddenField: {
+			return (
+				`Tree does not conform to schema. ` +
+				`A forbidden field of kind ${fieldKindDesc} must be empty, but found ${childCountDesc} children.`
+			);
```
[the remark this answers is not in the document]
jenn-le: Accepted. The diagnostic now uses "child" when the count is 1 and "children" otherwise, with the focused test updated to cover the singular case.

packages/dds/tree/src/feature-libraries/schemaChecker.ts
```diff
@@ -67,34 +84,100 @@ function formatSchemaValidationError(
 	error: SchemaValidationError,
 	context: SchemaValidationErrorContext | undefined,
 ): string {
+	const nodeDesc =
```
CraigMacomber: Use whole words, not `Desc`

packages/dds/tree/src/feature-libraries/schemaChecker.ts
```diff
@@ -50,6 +51,22 @@ export interface SchemaValidationErrorContext {
 	 * Field keys on the node that were not recognized by its schema.
 	 */
 	readonly unexpectedFields?: ReadonlySet<string>;
+	/**
+	 * The field kind identifier, if the error applies to a field.
+	 */
+	readonly fieldKind?: string;
```
CraigMacomber: I know this design of having SchemaValidationError and SchemaValidationErrorContext as separate things is a preexisting design, but its also a bad design as it results in the places that produce the errors and those that format them having to agree on what data to provide, and having no typing help with this.

If we want high quality errors, I think it would make way more sense to, at the location the error is discovered, generate the description of the error there when we have the full context of exactly what the error is.

We pack that description into some object, like this context, but it includes the actual error description string, maybe some extra fields for a code friendly description if you want (include SchemaValidationError in this), and a path to that node (which starts empty, and gets appended to as the validation logic returns for each layer).

That fixes the type safety annoyance, keeps the logic for describing the details of the error more localized without having to pass all this data in-between, and allows you to include the actual path to the node with the error in the message.

CraigMacomber: More concretely, if you inline the logic for making each description into the relevant locations in which call onError, its the same ammoutn of code, but much stronger typed and doesn't require this pile of optional fields inbetween.

If multiple places need to format the same error or similar errors, you can make a function for formatting that specific error, but no need to bundle all of them together like this.

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
```diff
@@ -78,6 +79,28 @@ import type { TreeCheckout } from "./treeCheckout.js";
...
+	viewSchema: TreeSchema,
+	storedSchema: TreeCheckout["storedSchema"],
+): void {
+	if (compatibility.canView) {
+		return;
+	}
+
+	const discrepancy = getSchemaCompatibilityError(viewSchema, storedSchema);
+	const details =
+		discrepancy === undefined ? "" : ` The first schema mismatch is: ${discrepancy}.`;
```
agarwal-navin, line 93: Will this only report the first schema mismatch? What if there are multiple? Should we consider logging all (or limit it to 10 or something)

packages/dds/tree/src/feature-libraries/schemaChecker.ts
```diff
@@ -33,83 +47,139 @@ export enum SchemaValidationError {
...
  */
-export function throwOutOfSchema(
-	maybeError: SchemaValidationError,
-	context?: SchemaValidationErrorContext,
-): never {
-	throw new UsageError(formatSchemaValidationError(maybeError, context));
+export function throwOutOfSchema(details: SchemaValidationErrorDetails): never {
+	const message = `Tree does not conform to schema. ${getSchemaValidationErrorMessage(details.error)}`;
+	throw new UsageError(
+		appendDebugMessage(message, () => getSchemaValidationDebugMessage(details)),
```
agarwal-navin, line 64: A couple of questions:
1. Why use `appendDebugMessage`? Basically, what's the rationale behind the decision to log these for non-production builds only?
2. Can we log the fields / types as telemetry properties (instead of part of the message) such that they can be tagged? That way they can still be part of prod.

agarwal-navin (approved):

packages/dds/tree/src/shared-tree/schematizingTreeView.ts
jenn-le: Yes, only the first mismatch. We can change the policy after some more discussion. I'm not sure what the best policy is but I think this is good enough for now.

packages/dds/tree/src/feature-libraries/schemaChecker.ts
jenn-le: 1. The rationale is to reduce production bundle size by including more details in only debug builds.
2. Created a work item tracking making decisions around what exactly we want to include as telemetry properties vs part of the message.

jenn-le:
Merging this as is so we have this for the 3.0 release, I have an item tracking additional clean up and cementing the decision about what information we include where.

jzaffiro (approved):

PR #27951, opened 2026-08-13
refactor(tree): Reserve positional argument in `treeChanged` event listener signature

Josmithr:
`TreeAlpha`'s `treeChanged` events already propogate delta events through this surface at runtime, so this is strictly formalizing an expectation that must already be satisfied to in order to safely use these events.


.changeset/reserve-tree-changed-argument.md
```diff
@@ -0,0 +1,19 @@
...
+"fluid-framework": minor
+"__section": tree
+---
+The treeChanged event reserves its listener argument for event data
+
+[`TreeChangeEvents.treeChanged`](https://fluidframework.com/docs/api/tree/treechangeevents-interface#treechanged-method) now declares its first listener argument as optional `unknown`.
+The event's runtime behavior has not changed, but the declaration reserves that position for event data that experimental or future APIs may provide.
+
+Most listeners require no changes.
+Listeners that declare their own optional first parameter should remove it or use a wrapper so they do not interpret event data as application data.
```
CraigMacomber, line 12: Might be worth adjusting the phrasing in this changeset to better communicate that this is not changing runtime behavior, and the runtime behavior which such code would mishandle is already happening.

Josmithr:
```suggestion
Listeners that declare their own optional first parameter should remove it or use a wrapper so they do not interpret event data as application data.
Such listeners were already unlikely to function correctly - these changes just make the contract for use more explicit.
```

CraigMacomber (approved):

noencke (approved):

PR #27955, opened 2026-08-13
If editing a forest fails, mark it as broken

CraigMacomber:
## Description

If editing a forest fails, mark it as broken.
This helps ensure that failed edits can't result in inconsistent forest states which get persisted spreading the corruption.

This is done at two layers for improved robustness: the forest will make itself broken if its visitor crashes doing the updates.
Addationally I have added higher level logic which will mark a forest as broken if we fail to load it, or fail to apply a change (from op or local) to it.

This should have no impact in cases where no errors which should be fatal occur.

This could prevent corruption from making it into an op or summary if we have a bug (in addition to the one causing the corruption) which tries to continue using the forest after an edit failed to apply, as it would make the later attempt to read the invalid forest throw indicating that it's broken by a previous fatal error.

## Reviewer Guidance


packages/dds/tree/src/feature-libraries/chunked-forest/chunkedForest.ts
```diff
@@ -422,9 +436,12 @@ export class ChunkedForest implements IEditableForest {
 		for (const getVisitor of this.deltaVisitors) {
 			announcedVisitors.push(getVisitor());
 		}
-		const combinedVisitor = combineVisitors([forestVisitor, ...announcedVisitors]);
-		this.activeVisitor = combinedVisitor;
-		return combinedVisitor;
+		const visitor = combineVisitors([
+			makeBreakingVisitor(forestVisitor, this.breaker),
+			...announcedVisitors,
+		]);
```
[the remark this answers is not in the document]
CraigMacomber: I have added additional use of breaker at higher levels which should handle such cases

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -497,6 +496,21 @@ export class TreeCheckout implements ITreeCheckout {
...
+	 * The breaker for this checkout.
+	 * @remarks
+	 * As the forest is part of the mutable state of the checkout,
+	 * we want to be broken if it is broken, so we use the forest's breaker as our own.
+	 *
+	 * This does mean that if something goes wrong in the checkout, we break the forest,
+	 * which is not strictly necessary, but may be desirable as if something is wrong with the checkout,
+	 * while the forest might be internally consistent, it likely doesn't contain the correct content
+	 * (since the checkout is keeping it up to date properly anymore).
+	 */
```
Josmithr, line 507: Nit: these docs are great - would the other `breaker` getters benefit from a similar level of detail?

Josmithr (approved):

packages/dds/tree/src/shared-tree/treeCheckout.ts
CraigMacomber: I think the others are less nuanced as there is a clearer scope.

Josmithr (approved):

PR #27861, opened 2026-08-05
feat(eslint-config-fluid): typescript 6 support

jason-ha:
Update typescript to v6 and eslint packages to versions supporting v6.

And run `pnpm dedupe`.

Additionally, workaround npmjs quarantine for eslint-plugin-fluid and provide some guidance on updating dependencies in alignment to repository policies.

[AB#79313](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/79313)

PR #27956, opened 2026-08-13
build(eslint-config-fluid): remove eslint-plugin-fluid feed ingestion workaround

MarioJGMsoft:
## Description

`eslint-config-fluid` declares `@fluid-internal/eslint-plugin-fluid@^0.5.0`, but pnpm `overrides` added in #27861 silently downgraded installs back to `0.4.1` while `0.5.0` was still waiting to be ingested into the ADO `publicPackages` upstream feed. That block was documented as only being needed until 2026-08-10, which has now passed, so it and its two companion `@typescript-eslint` transitive pins are removed.

The lockfile was regenerated with the `install-no-frozen` script and now resolves `0.5.0` directly, matching the declared specifier.

The `@rushstack/eslint-plugin` and `eslint-plugin-tsdoc` overrides are intentionally left in place — those cover third-party packages lagging on TypeScript 6, not feed ingestion.

## Reviewer Guidance

-   The lockfile also picks up `@typescript-eslint/types@8.65.0` (was `8.58.2`) under `project-service` — incidental churn from re-resolving with `resolutionMode: highest`, not related to the plugin bump.
-   Worth confirming the eslint-config-fluid pipeline installs cleanly against the feed, since that is the scenario the workaround existed to paper over.

Fixes: [AB#79930](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/79930)

### DIFF
--- a/common/build/eslint-config-fluid/pnpm-lock.yaml
+++ b/common/build/eslint-config-fluid/pnpm-lock.yaml
@@ -7,9 +7,6 @@ settings:
 overrides:
   '@rushstack/eslint-plugin>@typescript-eslint/utils@<8.58.0': ~8.58.0
   eslint-plugin-tsdoc>@typescript-eslint/utils@<8.58.0: ~8.58.0
-  '@fluid-internal/eslint-plugin-fluid@^0.5.0': ^0.4.1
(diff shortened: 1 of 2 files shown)


jason-ha (approved):
❤️

PR #27957, opened 2026-08-13
chore: remove unused transitive dependency subtrees

CraigMacomber:
## Description

Some development tools pull in dependency subtrees for optional functionality that Fluid never uses. Add targeted pnpm overrides to omit those subtrees, reducing install surface, native build scripts, platform-specific packages, and lockfile churn.

| Removal | Workspaces | Net reduction |
|---|---|---:|
| `jssm-viz-cli>sharp` | build-tools | 26 |
| `sass>@parcel/watcher` | client | 14 |
| `pm2>pm2-sysmonit` | client and Routerlicious | 8 |
| `tar-fs>bare-fs` and `tar-fs>bare-path` | client and common-utils | 18 |
| **Total** | | **66** |

Sharp requires `npm:empty-npm-package@1.0.0` because `jssm-viz-cli` imports it eagerly, although Fluid only generates SVG diagrams. The other dependencies support unused optional features and can be dropped with `"-"`.

Validation included frozen-lockfile installs for every affected workspace, repository quick checks, production webpack builds using Sass, PM2 start/stop, Node-mode tar packing, and SVG diagram generation without Sharp.

## Reviewer Guidance

The override comments document the omitted functionality, measured dependency reduction, and relevant native build behavior.

I have no idea if this reduction is worth the complexity: I simply wondered how much impact an agent could find doing this. In my opinion this is enough that it might be worth considering keeping, but likely not.

There is also risk that some of this breaks dev workflows not tested by PR CI.


alexvy86 (approved):
If nothing else, seems like a good experiment to me. But I suspect it'll stick and I like it.

PR #27964, opened 2026-08-13
test: Update example packages to use playwright instead of jest + puppeteer

Josmithr:
Migrate the repository's example packages and related DevTools end-to-end tests from Jest + Puppeteer to [Playwright](https://playwright.dev/). This removes the Jest-Puppeteer configuration and type workaround package, adds shared Playwright configuration, and updates build and CI wiring to run and publish Playwright test results.

The `client-utils` browser hash tests bundle and execute the production browser entry point so regressions in the browser implementation cannot be masked by copied test logic.

To facilitate the new Playwright configurations, Tinylicious now supports selecting its listening port via command line argument.

[AB#78597](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/78597)


build-tools/packages/build-cli/src/library/repoPolicyCheck/npmPackages.ts
```diff
@@ -1372,7 +1372,12 @@ export const handlers: Handler[] = [
 			}
 			const testScript = scripts.test;
 
-			const splitTestScriptNames = ["test:mocha", "test:jest", "test:realsvc"];
+			const splitTestScriptNames = [
+				"test:mocha",
+				"test:jest",
+				"test:realsvc",
+				"test:playwright",
```
Josmithr:
```suggestion
				"test:playwright",
				"test:realsvc",
```

examples/playwright.config.base.ts
```diff
@@ -0,0 +1,22 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { devices, type PlaywrightTestConfig } from "@playwright/test";
+
+export const baseConfig: PlaywrightTestConfig = {
+	testDir: "tests",
```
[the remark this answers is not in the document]
Josmithr: That is incorrect.

alexvy86:
I found two issues in the `client-utils` portion of this migration:

1. `packages/common/client-utils/src/test/playwright/gitHash.test.ts`: the replacement no longer exercises the production browser hashing implementation. The previous Jest test used `rewire` to invoke `digestBuffer` and `encodeDigest` from `hashFileBrowser.js`; the new test imports the Node entry point and independently reimplements those browser operations in `browserHash`/`encodeDigest`. A regression in the production browser implementation can therefore pass while the copied implementation remains correct. Could we run the actual browser entry point in the page, for example through a small browser bundle that imports `indexBrowser.js`?

2. `packages/common/client-utils/package.json`: `fluidBuild.tasks` now defines `build:test:mocha:cjs`, `build:test:mocha:esm`, and `build:test:types` twice. The values currently match, so behavior is unchanged, but JSON parsing silently keeps only the later copies and future edits to the earlier definitions would be ignored. Please keep one definition for each key and add only `build:test:playwright` in the new block.

alexvy86:
One behavior change seems worth calling out in the PR description: the old Jest-Puppeteer configs used `usedPortAction: "error"`, whereas the Playwright configs use `reuseExistingServer: !process.env.CI`. CI still fails on a port conflict, but local test runs will now reuse a process already listening at the assigned URL instead of rejecting it. That is probably a useful local-development improvement, but it could also connect a test to a stale or unrelated server, so documenting the intentional tradeoff would help reviewers and future maintainers.

alexvy86:
I did a joint pass over the whole PR with an agent (didn't look at every single file myself) and I think the change is sound. I can approve once the merge conflict is addressed and the 2 points in https://github.com/microsoft/FluidFramework/pull/27964#issuecomment-5330615422 have been considered.

Josmithr:
> One behavior change seems worth calling out in the PR description: the old Jest-Puppeteer configs used `usedPortAction: "error"`, whereas the Playwright configs use `reuseExistingServer: !process.env.CI`. CI still fails on a port conflict, but local test runs will now reuse a process already listening at the assigned URL instead of rejecting it. That is probably a useful local-development improvement, but it could also connect a test to a stale or unrelated server, so documenting the intentional tradeoff would help reviewers and future maintainers.

I went ahead and reverted this behavioral change. We can always reconsider this later, but that minimizes the scope of change here.

Josmithr:
@copilot resolve the merge conflicts in this pull request

pnpm-lock.yaml
```diff
@@ -20057,6 +19211,11 @@ packages:
     resolution: {integrity: sha512-+1VkjdD0QBLPodGrJUeqarH8VAIvQODIbwh9XpP5Syisf7YoQgsJKPNFoqqLQlu+VQ/tVSshMR6loPMn8U+dPg==}
     engines: {node: '>=14'}
 
+  '@playwright/test@1.62.1':
```
alexvy86, line 19214: If we're already updating the package.json version and it resolved to 1.62.1, might as well declare that one as minimum in package.json files?

build-tools/packages/build-cli/src/library/repoPolicyCheck/npmPackages.ts
```diff
@@ -1372,7 +1372,12 @@ export const handlers: Handler[] = [
 			}
 			const testScript = scripts.test;
 
-			const splitTestScriptNames = ["test:mocha", "test:jest", "test:realsvc"];
+			const splitTestScriptNames = [
```
alexvy86, line 1375: Update the error message on line 1395 to reflect the new `test:playwright` for completeness.

server/routerlicious/.changeset/calm-ports-listen.md
alexvy86, line 1: The changes to tinylicious in the PR can't be having an effect on the tests in the client release group since we don't use the in-repo version, right? So are they necessary at all? If they are, let's split them into their own PR, but it seems to me they're just a convenience improvement with no immediate use case.

Josmithr: That's a good point. I didn't immediately question if this was actually needed for the playwright infra, assuming it was. But it can't be... I'll investigate. Either way, this seems like a nice change to get in (I've wanted this capability for other reasons in the past anyways), but agreed that it should be a separate PR.

Josmithr: Confirmed unneeded. I wonder if the agent working on this thought they needed it initially, then pivoted its approach? I'll revert the changes here and put up a separate PR.

Josmithr: Changes reverted here. New PR: https://github.com/microsoft/FluidFramework/pull/28042

alexvy86 (approved):
🚀 . We should try to follow up soon with removing jest from the (non-example) packages that still use it, I think Abram had brought that up as one of his requests, that we don't end up in a limbo where we still have jest and playwright.

PR #27968, opened 2026-08-14
Expose point-in-time APIs through new legacy beta entrypoint

sonalideshpandemsft:
Moves the point-in-time loading and capture APIs from the legacy alpha to custom beta entrypoints:

- `@fluidframework/container-loader/legacy/beta/pointInTime`
- `@fluidframework/odsp-driver/legacy/beta/pointInTime`

This provides explicit, stable import paths for the point-in-time APIs and prevents them from being exported through the packages general-purpose entrypoints. Test utilities have been updated to use the new paths. 

The entrypoints follow the same consumption rules as beta APIs, but the APIs they expose are expected to evolve and may break without a general beta API break announcement


packages/drivers/odsp-driver/api-report/odsp-driver.legacy.alpha.api.md
```diff
@@ -65,7 +65,7 @@ export function getHashedDocumentId(driveId: string, itemId: string): Promise<st
 // @beta @legacy
 export function getLocatorFromOdspUrl(url: URL, requireFluidSignature?: boolean): OdspFluidDataStoreLocator | undefined;
 
-// @alpha @legacy
+// @beta @legacy
```
markfields: I wouldn't expect this to be in .alpha.api.md file, can you regenerate the api report somehow?

packages/drivers/odsp-driver/api-report/odsp-driver.legacy.beta.api.md
```diff
@@ -65,6 +65,9 @@ export function getHashedDocumentId(driveId: string, itemId: string): Promise<st
 // @beta @legacy
 export function getLocatorFromOdspUrl(url: URL, requireFluidSignature?: boolean): OdspFluidDataStoreLocator | undefined;
 
+// @beta @legacy
```
markfields: I'm not sure if these should be in the api report, I guess this detail didn't come up in the meeting this morning.

packages/drivers/odsp-driver/src/pointInTime.ts
```diff
@@ -0,0 +1,11 @@
+/*!
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/* eslint-disable import-x/no-internal-modules */
```
markfields, line 6: Consider adding justification in the comment?

packages/drivers/odsp-driver/package.json
```diff
@@ -43,6 +43,10 @@
 				"default": "./dist/index.js"
 			}
 		},
+		"./legacy/beta/pointInTime": {
```
markfields, line 46: What will a sample import look like in Bohemia?  `from '@fluidframework/odsp-driver/legacy/beta/pointInTime';`, right?

I was debating with myself whether to keep "beta" in there.  I guess it is good because it indicates it's a beta API in the general sense.  The only risk is that someone familiar with our current support structure for beta (including deprecation announcements) copy/pastes it and then is shocked when it disappears.

Although... in Bohemia itself, the integration pipeline will catch a rogue usage when we go to remove it.  So maybe it's ok.

An alternative would be to promote a new term here (by convention, not requirement) like `./dev/pointInTime` or `./temp/pointInTime`.  I think I like `dev`.

packages/drivers/odsp-driver/api-report/odsp-driver.legacy.beta.api.md
sonalideshpandemsft: It was part of beta api reports since I missed removing them from index.ts

packages/drivers/odsp-driver/api-report/odsp-driver.legacy.alpha.api.md
sonalideshpandemsft: alpha reports will include beta apis

packages/drivers/odsp-driver/package.json
sonalideshpandemsft: Based on the conversation offline -- keeping it `legacy/beta/<feature-name>` as it avoids confusion when these imports are integrated in bohemia

Josmithr:
What is the motivation for having separate entrypoints for these APIs? I.e., why are our standard tagging conventions not sufficient? We should avoid custom entrypoints whenever possible, because it adds non-trivial complexity to package configurations / builds.

packages/drivers/odsp-driver/api-report/odsp-driver.legacy.alpha.api.md
```diff
@@ -65,9 +65,6 @@ export function getHashedDocumentId(driveId: string, itemId: string): Promise<st
 // @beta @legacy
 export function getLocatorFromOdspUrl(url: URL, requireFluidSignature?: boolean): OdspFluidDataStoreLocator | undefined;
 
-// @alpha @legacy
```
Josmithr, line 68: Note my general comment [here](https://github.com/microsoft/FluidFramework/pull/27968#issuecomment-5296978312) first. We should avoid custom entrypoints whenever possible.

But if we _really_ do need a custom entrypoint, we need to make sure we are generating API reports for it. This PR removes report coverage of APIs we plan to continue to support, which I don't think is okay.

Josmithr: We also need to make sure we are running our API-Extractor _lint_ checks to prevent accidentally leaking release level incompatibilities across packages.

sonalideshpandemsft: Bohemia does not allow consuming alpha  APIs. Therefore, these APIs are being promoted to beta  and exposed through a dedicated entrypoint. Eventually, they will move under `legacy/beta`, and the temporary entrypoint will be removed. This keeps any breaking changes scoped to the feature that consumes these APIs

Josmithr: Why can't we just promote them to legacy/beta then? We can make any breaking changes needed now because the APIs are currently alpha. And future breaking changes can be made in accordance with our beta breaking policy (minor version increments of 10). The reason 1P partners doesn't want to use alpha APIs is because they want stability promises. This special entrypoint seems like we're just working around the specific alpha rule without actually addressing the underlying concern.

packages/drivers/odsp-driver/package.json
```diff
@@ -43,6 +43,10 @@
...
+			"import": "./lib/pointInTime.js",
+			"require": "./dist/pointInTime.js"
+		},
```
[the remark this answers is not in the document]
jason-ha: types is required for other exports because of a flawed resolution in build-tools. Here we don't need those tools to process these exports and simplicity is preferred.

packages/drivers/odsp-driver/package.json
```diff
@@ -86,8 +92,9 @@
 		"ci:build:api-reports": "concurrently \"npm:ci:build:api-reports:*\"",
 		"ci:build:api-reports:current": "api-extractor run --config api-extractor/api-extractor.current.json",
 		"ci:build:api-reports:legacy": "api-extractor run --config api-extractor/api-extractor.legacy.json",
+		"ci:build:api-reports:pointInTime": "api-extractor run --config api-extractor/api-extractor.pointInTime.json",
 		"ci:build:docs": "api-extractor run",
-		"clean": "rimraf --glob dist lib {alpha,beta,internal,legacy}.d.ts legacyAlpha.d.ts \"**/*.tsbuildinfo\" \"**/*.build.log\" _api-extractor-temp nyc",
+		"clean": "rimraf --glob dist lib {alpha,beta,internal,legacy}.d.ts legacy/alpha.d.ts legacyAlpha.d.ts \"**/*.tsbuildinfo\" \"**/*.build.log\" _api-extractor-temp nyc",
```
jason-ha, line 97: ?

packages/loader/container-loader/package.json
```diff
@@ -93,6 +93,10 @@
 				"default": "./dist/index.js"
 			}
 		},
+		"./legacy/beta/pointInTime": {
+			"import": "./lib/pointInTime.js",
+			"require": "./dist/pointInTime.js"
+		},
```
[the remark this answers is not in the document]
jason-ha: types is required for other exports because of a flawed resolution in build-tools. Here we don't need those tools to process these exports and simplicity is preferred.

packages/drivers/odsp-driver/api-report/odsp-driver.legacy.alpha.api.md
sonalideshpandemsft: The intent is these APIs are integrated and test in 1P partner codebases while they are still evolving. Breaks will be introduced in beta breaking releases but they won't need to be announced prior to the release

packages/drivers/odsp-driver/api-report/odsp-driver.pointInTime.beta.api.md
```diff
@@ -0,0 +1,17 @@
+## Beta API Report File for "@fluidframework/odsp-driver"
```
markfields, line 1: Future improvement - Might be good to put some words in here differentiating this from regular beta exports (besides just the filename)

packages/drivers/odsp-driver/package.json
markfields, line 97: Looks extraneous / unrelated?

packages/drivers/odsp-driver/package.json
```diff
@@ -61,6 +65,7 @@
 		"build:api-reports": "concurrently \"npm:build:api-reports:*\"",
 		"build:api-reports:current": "api-extractor run --local --config api-extractor/api-extractor.current.json",
 		"build:api-reports:legacy": "api-extractor run --local --config api-extractor/api-extractor.legacy.json",
+		"build:api-reports:pointInTime": "api-extractor run --local --config api-extractor/api-extractor.pointInTime.json",
```
markfields, line 68: Were these added automatically?  Feels like a maintenence burden we'd want to automate if not

packages/loader/container-loader/src/loadContainerToSequenceNumber.ts
```diff
@@ -27,7 +27,7 @@ import {
  * point-in-time capability the loader detects. For ODSP, pass an
  * `IPointInTimeDocumentServiceFactory` created by `getOdspPointInTimeDocumentServiceFactory`.
  *
- * @legacy @alpha
+ * @legacy @beta
```
markfields, line 30: Suggestion: Start a convention where we mention the custom export and explain the relaxed requirements around documenting breaks.  Something like

`@legacy @beta - Exported only under the legacy/beta/pointInTime namespace. No need to document breaking changes until promoted to general legacy/beta`

markfields (approved):

packages/drivers/odsp-driver/api-report/odsp-driver.legacy.alpha.api.md
Josmithr: So, the primary intended difference here is that we don't want to announce the breakages beforehand? Otherwise we're following the existing beta/legacy guidelines? If so, this seems like a lot of infrastructural churn / maintenance burden to support that. Seems like just making the APIs normal legacy/beta would be much easier, IMO

markfields: This idea was hatched this morning in a chat with @anthony-murphy and @jason-ha.

It's just as Sonali said - We hold ourselves to everything about the legacy-beta bar except the breaking change documentation guidelines (announcing breaks ahead of time).  We still must stage breaks gracefully, with one good option being just add `myUnderDevelopmentApi2` with the new shape and then remove the old one in the beta-breaking release.

We are adding a new thing because (A) alpha is not usable in practice in most scenarios and (B) beta is semi-permanent and harder to quickly remove/iterate.  This "beta but under a temporary export" solves those both.

Josmithr: But this setup doesn't really make things easier to iterate/remove. If something is tagged as `@beta` it means we only break it in minor versions that are increments of 10. Adding a new entrypoint doesn't change that. It sounds like the only difference is that we don't want to have to announce the changes ahead of time? I REALLY don't think that's a good reason to add all of this infra complexity.

sonalideshpandemsft:
closing this PR in favor: https://github.com/microsoft/FluidFramework/pull/27976

packages/drivers/odsp-driver/api-report/odsp-driver.legacy.alpha.api.md
lindsnguyen: We're making this change and not directly using `@beta` not because we don't want to annouce but because we're trying to solve for the fact these APIs has a high chance they might be changed. By having this temp entrypoint we were hoping to keep iterating without commiting them to the general `legacy/beta` surface, so nobody's taking a permanent beta dependency on something that's still moving. It still gives bohemia beta-grade dependency rules, just scoped to this feature. Then we graduate them into real `legacy/beta` once the shapes settle.
 
We were hoping to defer the commitment, but due to timing, we're just going to go ahead and promote it to beta/legacy and if there ends up requiring breaking changes we'll just stage those breaks accordingly.

PR #27969, opened 2026-08-14
test(external-data): Separate external data unit and UX test infrastructure

Josmithr:
Fixes the package's single app smoke test. Also updates the package's testing infra in the following ways, to clean things up:
- Splits unit tests from app UX tests
- Migrates unit tests from jest+puppeteer to mocha
- Renames test directories to match our expected conventions ("test" not "tests")


examples/external-data/tests/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+	"compilerOptions": {
+		"rootDir": "./",
+		"outDir": "../lib/test",
+		"types": ["mocha", "node"],
+	},
+	"include": [
+		"./customerService.test.ts",
+		"./externalDataService.test.ts",
+		"./taskData.test.ts",
+		"./utilities.ts",
```
Josmithr: Rather than enumerating specific files here and in the jest configuration, let's move the unit tests and app tests into separate directories so we can leverage standard glob patterns. My recommendation would be to move the unit tests under `src/test` and keeping the app tests here under `/tests`.

examples/external-data/.mocharc.cjs
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) Microsoft Corporation and contributors. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+"use strict";
+
+const getFluidTestMochaConfig = require("@fluid-internal/mocha-test-setup/mocharc-common");
+
+const config = getFluidTestMochaConfig(__dirname);
+config.timeout = 10_000;
```
Josmithr: Is this timeout override needed? If so, please document why.

examples/external-data/tests/tsconfig.json
Josmithr: Actually, let's also rename the "tests" directory to "test" while we're making changes here. That is the more standard naming for this repo.

examples/external-data/src/test/customerService.test.ts
```diff
@@ -268,23 +271,15 @@ describe("mock-customer-service", () => {
...
 			await closeServer(localService);
 		}
 	});
 
-	// Skipping to close off the broadcast-signal loop. Tested manually and it works well.
-	// Unclear why localServiceApp is failing to post right for this test.
-	it("events-listener: Complete data flow for session-end event", async () => {
+	// This test intentionally waits twice for webhook delivery and non-delivery.
+	it("events-listener: Complete data flow for session-end event", async function () {
+		this.timeout(5000);
```
Josmithr:
```suggestion
	it("events-listener: Complete data flow for session-end event", async function () {
		// This test intentionally waits twice for webhook delivery and non-delivery.
		this.timeout(5000);
```

examples/external-data/test/index.tsx
```diff
@@ -33,7 +33,7 @@ export async function createContainerAndRenderInElement(
 
 		id = await createResponse.attach();
 		model.baseDocument.addTaskList({
-			externalTaskListId: "task-list-test",
+			externalTaskListId: "task-list-1",
```
Josmithr, line 36: Note for reviewers: this was causing the app smoke test to fail - this was not the ID expected elsewhere in the tests.

alexvy86 (approved):
Looks good. Called out offline that maybe we just want to get rid of this example though.

Issue #27851, opened 2026-08-04
Deprecate minVersionForCollab and MinimumVersionForCollab in favor of oldestSupportedClient


WillieHabi:
## Deprecated API

The pre-rename names for the cross-client compatibility parameter:

- `minVersionForCollab` on `LoadContainerRuntimeParams` (`@fluidframework/container-runtime`) — `@legacy @beta`
- `minVersionForCollab` on `BaseContainerRuntimeFactoryProps` (`@fluidframework/aqueduct`) — `@legacy @beta`
- `minVersionForCollaboration` on `createTreeContainerRuntimeFactory` (`@fluidframework/fluid-static`) — `@legacy @beta`
- The `MinimumVersionForCollab` type (`@fluidframework/runtime-definitions`) — `@public`

**Context**

These are replaced by `oldestSupportedClient` with `OldestSupportedClientVersion` for runtime APIs in https://github.com/microsoft/FluidFramework/pull/27806. The deprecated public and beta names remain accepted so those renames are additive. The alpha driver APIs are replaced directly by `OldestSupportedServiceClientVersion` and `ServiceOptions.oldestSupportedClient`; indexed office-bohemia code does not use the removed alpha names.

The name is layer-neutral. The value is threaded through Runtime and DataStore layers and consumed by DDSes such as SharedTree for write-format and feature selection.

**Approach**

Consumers move to `oldestSupportedClient`, typed `OldestSupportedClientVersion` for runtime APIs or `OldestSupportedServiceClientVersion` for the minor-only service-client API.

**Dependencies**

https://github.com/microsoft/FluidFramework/pull/27806 introduces the replacement names, deprecates the public and beta old names, removes the superseded alpha driver names, migrates in-repo type references, updates service-client signatures, and adds the preferred `createTreeContainerRuntimeFactory` overload.

This is distinct from https://github.com/microsoft/FluidFramework/issues/27180, which makes the parameter required but does not remove either name.

https://github.com/microsoft/FluidFramework/pull/27896 changes which internal parent-context interface `ContainerRuntime` implements. Phase 2 should account for that final interface shape.

**Compatibility Concerns**

The deprecated public and beta names need to coexist while Runtime → DataStore → DDS and service-client internals migrate across package and API layers. Partners consume the beta properties directly, so their removal needs the standard beta-break lead time.

[...]

PR #27971, opened 2026-08-14
Require explicit oldestSupportedClient at public runtime boundaries

### LINKED ISSUE #27180: Make minVersionForCollab required

WillieHabi:
## Description

This PR requires applications to make an explicit cross-client compatibility choice when constructing a container runtime through the customer-facing Container Runtime and Aqueduct APIs.

### API behavior

- **Beta Container Runtime and Aqueduct APIs:** Callers must provide exactly one of `oldestSupportedClient` (preferred) or `minVersionForCollab` (deprecated). Supplying neither or both is rejected by TypeScript and at runtime.
- **Alpha Container Runtime API:** `loadContainerRuntimeAlpha` accepts only `oldestSupportedClient`. Missing values and the deprecated property are rejected at runtime for JavaScript and type-erased callers.
- **Internal static loaders:** `ContainerRuntime.loadRuntime` and `loadRuntime2` retain their historical fallback for direct internal test and replay callers.

### Scope

- The accepted compatibility floor does not change in this PR.
- The historical `2.0.0-defaults` sentinel remains distinct from explicit `2.0.0` because it preserves the runtime configuration used before compatibility versions were explicitly selected.
- Persisted document-schema fields and telemetry property names are unchanged.

Caller preparation landed separately in [PR #28024](https://github.com/microsoft/FluidFramework/pull/28024) and [PR #28025](https://github.com/microsoft/FluidFramework/pull/28025).

Fixes #27180. Tracks [AB#42061](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/42061).

## Breaking Changes

Client 3.0 callers must pass `oldestSupportedClient` when using these customer-facing APIs:

| API | Client 3.0 requirement |
| --- | --- |
| `loadContainerRuntime` | Pass `oldestSupportedClient` |
| `BaseContainerRuntimeFactory` | Pass `oldestSupportedClient` to the constructor |
| `ContainerRuntimeFactoryWithDefaultDataStore` | Pass `oldestSupportedClient` to the constructor |

Existing callers may temporarily use the deprecated `minVersionForCollab` overloads while migrating. Those overloads are scheduled for removal in the Client 3.10 beta-break window under [issue #27851](https://github.com/microsoft/FluidFramework/issues/27851).

```typescript
const runtime = await loadContainerRuntime({
	// Other runtime parameters...
	oldestSupportedClient: "2.0.0",
});
```

See [Breaking vs. Non-Breaking Changes](https://github.com/microsoft/FluidFramework/blob/main/docs/content/Contributing/Breaking-vs-Non-Breaking-Changes.md).

## Validation

- Container Runtime and Aqueduct source and test compilation passed for ESM and CommonJS.
- Container Runtime: 1,071 tests passed.
- Aqueduct: 8 tests passed.
- Package checks, ESLint, API reports, API docs, generated type tests, and CI readiness passed.

## Reviewer Guidance

The review process is outlined in [the pull request guidelines](https://github.com/microsoft/FluidFramework/blob/main/docs/content/Contributing/PR-Guidelines.md#guidelines).

- **Review surface:** 20 files.
- **Primary focus:** the customer-facing exact-one boundary, the canonical-only alpha boundary, and preservation of the internal static-loader fallback.
- **Type and runtime coverage:** invalid executable calls use `@ts-expect-error`, so the same expressions prove both TypeScript rejection and runtime rejection.
- **Migration API:** temporary deprecated parameter types keep guidance visible without marking canonical overloads as deprecated.
- **Follow-up:** [PR #27972](https://github.com/microsoft/FluidFramework/pull/27972) separately raises the deployed-client floor to 2.0.0 while preserving the historical `2.0.0-defaults` sentinel for internal and test behavior.
- **Required approvals:** the beta and legacy-alpha API changes require separate `fluid-cr-api` approval.
- **Integration:** the office-bohemia integration pipeline must pass before merge.


examples/version-migration/separate-container/src/modelVersion1/containerCode.ts
```diff
@@ -76,6 +76,12 @@ export class InventoryListContainerRuntimeFactory implements IRuntimeFactory {
 		compositeEntryPoint.addEntryPointPiece(modelEntryPointPiece);
 		const migratorEntryPointPiece = makeMigratorEntryPointPiece(exportDataCallback);
 		compositeEntryPoint.addEntryPointPiece(migratorEntryPointPiece);
-		return loadCompositeRuntime(context, existing, compositeEntryPoint, this.runtimeOptions);
+		return loadCompositeRuntime(
+			context,
+			existing,
+			compositeEntryPoint,
+			"2.0.0",
```
jason-ha: Wondering if there should be a common examples setting.
Will be annoying to visit a bunch of randoms in the future and wonder whether they need 2.0.0 for some reason and will be invalid.
I think our examples can be more recent and can use 3.0.0 unless they have reason otherwise. The guidance after all is to use most recent that is commonly out. These aren't out (deployed) so 3.0.0 should be good.

packages/framework/aqueduct/src/test/aqueduct.spec.ts
```diff
@@ -3,5 +3,132 @@
...
+		// @ts-expect-error -- exactly one compatibility version property is required.
+		acceptVersionProperties(commonProps);
+		acceptVersionProperties({
+			...commonProps,
+			oldestSupportedClient: "2.0.0",
+			// @ts-expect-error -- both compatibility version properties cannot be supplied.
+			minVersionForCollab: "2.0.0",
+		});
+	};
+	verifyVersionPropertiesAtCompileTime();
```
jason-ha: No need to change, but I think it would have been fine to have these in the runtime tests.
The runtime tests now are casting which makes it hard to see that there is also type check protections.

packages/framework/attributor/src/mixinAttributor.ts
```diff
@@ -83,12 +80,8 @@ export const mixinAttributor = (
 			}
 
 			const runtime = await Base.loadRuntime({
-				context,
+				...params,
```
jason-ha: This is different behavior and will allow all params properties to come through. Was there intentional filtering before? Is this change safe?

packages/framework/data-object-base/src/runtimeFactory.ts
```diff
@@ -65,6 +65,7 @@ export class RuntimeFactory extends RuntimeFactoryHelper {
 			registryEntries: this.registry,
 			existing,
 			provideEntryPoint: this.provideEntryPoint,
+			oldestSupportedClient: "2.0.0",
```
jason-ha: This does not look great. How can we do better? If 2.0.0 can be chosen, why not something else?

packages/runtime/container-runtime/src/test/batching.spec.ts
```diff
@@ -84,6 +84,7 @@ describe("Runtime batching", () => {
 			context: getMockContext(mockDeltaManager) as IContainerContext,
 			registry: new FluidDataStoreRegistry([]),
 			existing: false,
+			oldestSupportedClient: "2.0.0-defaults",
```
jason-ha: What is the plan? This will not be allowed. So why do you have to do it now?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -766,7 +765,7 @@ export interface UnknownIncomingTypedMessage extends TypedMessage {
 type UnsequencedSignalEnvelope = Omit<ISignalEnvelope, "clientBroadcastSignalSequenceNumber">;
 
 /**
- * This object holds the parameters necessary for the {@link loadContainerRuntime} function.
+ * This object holds the parameters necessary for the `loadContainerRuntime` function.
```
jason-ha: `loadContainerRuntime` does not exist any more?
Perhaps you just need to qualify the reference.?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -805,51 +804,105 @@ export interface LoadContainerRuntimeParams {
...
 	 * @example
 	 * oldestSupportedClient: "2.0.0"
 	 *
 	 * @privateRemarks
 	 * Used to determine the default configuration for {@link IContainerRuntimeOptionsInternal} that affect the document schema.
-	 * For example, let's say that feature `foo` was added in 2.0 which introduces a new op type. Additionally, option `bar`
-	 * was added to `IContainerRuntimeOptionsInternal` in 2.0 to enable/disable `foo` since clients prior to 2.0 would not
-	 * understand the new op type. If a customer were to set oldestSupportedClient to 2.0.0, then `bar` would be set to
-	 * enable `foo` by default. If a customer were to set oldestSupportedClient to 1.0.0, then `bar` would be set to
-	 * disable `foo` by default.
```
jason-ha, line 826: Was this removed per 2.0 and 1.0 use?
Comment still seems relevant.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -805,51 +804,105 @@ export interface LoadContainerRuntimeParams {
...
+ * @legacy @beta
+ */
+export function loadContainerRuntime(
+	params: Omit<LoadContainerRuntimeParams, "oldestSupportedClient" | "minVersionForCollab"> & {
+		readonly oldestSupportedClient?: never;
+		/**
+		 * @deprecated 2.116.0. To be removed in 3.10.0. Pass `oldestSupportedClient` instead.
+		 * See {@link https://github.com/microsoft/FluidFramework/issues/27851} for context.
+		 */
+		readonly minVersionForCollab: OldestSupportedClientVersion;
```
jason-ha: I don't know that those comments show up anywhere for anyone.
I think you are trying to avoid publishing `DeprecatedLoadContainerRuntimeParams` but I think that is the only way the comments will be seen. I could be wrong.
I don't think it is so horrible to expose `DeprecatedLoadContainerRuntimeParams` temporarily.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -805,51 +804,105 @@ export interface LoadContainerRuntimeParams {
...
  * @legacy @beta
  */
-export async function loadContainerRuntime(
+export function loadContainerRuntime(
 	params: LoadContainerRuntimeParams,
+): Promise<IContainerRuntime & IRuntime>;
+
+/**
+ * Overload accepting the deprecated `minVersionForCollab` property.
+ * @legacy @beta
```
jason-ha: Calling this version is deprecated, right? So, mark as such.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -858,7 +911,7 @@ export async function loadContainerRuntime(
 }
 
 /**
- * Alpha variant of {@link loadContainerRuntime} that returns the runtime in an
+ * Alpha variant of `loadContainerRuntime` that returns the runtime in an
```
jason-ha: Restore using reference qualifier.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -870,7 +923,29 @@ export async function loadContainerRuntime(
...
+		readonly oldestSupportedClient?: never;
+		/**
+		 * @deprecated 2.116.0. To be removed in 3.10.0. Pass `oldestSupportedClient` instead.
+		 * See {@link https://github.com/microsoft/FluidFramework/issues/27851} for context.
+		 */
+		readonly minVersionForCollab: OldestSupportedClientVersion;
+	},
+): Promise<{
+	runtime: IContainerRuntime & IRuntime;
+}>;
```
jason-ha: Is this in use in bohemia or across layer boundaries? It is alpha, so we should be able to just go with the non-deprecated version.

packages/runtime/datastore-definitions/src/dataStoreRuntime.ts
```diff
@@ -228,8 +228,8 @@ export interface IFluidDataStoreRuntimeInternalConfig {
 	 * choosing an older version may limit available features and write formats.
 	 * This property is consumed by `SharedObjectFactory` (which are implementations of
 	 * {@link @fluidframework/datastore-definitions#IChannelFactory}).
-	 * See {@link @fluidframework/container-runtime#LoadContainerRuntimeParams.oldestSupportedClient}
-	 * for more details.
+	 * See {@link @fluidframework/container-runtime#LoadContainerRuntimeParams} for more details
+	 * about `oldestSupportedClient`.
```
jason-ha: Was this broken but not called out before? Property remains intact from what I see.

packages/runtime/runtime-utils/src/compatibilityBase.ts
```diff
@@ -10,29 +10,6 @@ import { compare, gt, gte, lte, valid, parse } from "semver-ts";
...
- *
- * Further to distinguish unspecified `minVersionForCollab` from a specified
- * version and allow `enableExplicitSchemaControl` to default to `true` for
- * any 2.0.0+ version, we will use a special value of `2.0.0-defaults`, which
- * is semantically less than 2.0.0.
- *
- * @internal
- */
-export const defaultMinVersionForCollab =
-	"2.0.0-defaults" as const satisfies OldestSupportedClientVersion;
```
jason-ha, line 34: I see expression of this value around. So, we haven't really gotten rid of its use. Perhaps it was a good exercise to see where it was needed and if any cases could move on. But for the places that keep this it seems better to have a central constant. A rename would be in order but that can happen later.

packages/runtime/runtime-utils/src/compatibilityBase.ts
```diff
@@ -320,12 +294,6 @@ export function validateConfigMapOverrides<T extends Record<string, unknown>>(
 	overrides: Partial<T>,
 	validationMap: ConfigValidationMap<T>,
 ): void {
-	if (minVersionForCollab === defaultMinVersionForCollab) {
-		// If the minVersionForCollab is set to the default value, then we will not validate the runtime options
-		// This is to avoid disruption to users who have not yet set the minVersionForCollab value explicitly.
-		// TODO: This also skips validation for users which explicitly request defaultMinVersionForCollab which seems like a bug.
-		return;
-	}
```
jason-ha, line 328: Not sure this is the right time. "2.0.0-default" is still out there.

packages/test/snapshots/src/validateSnapshots.ts
jason-ha, line 1: Eh? Swept up collateral?

jason-ha (commented):
This is very large and there are some questions. It might be the wrong time to require `oldestClientSupported`. It doesn't appear like we couldn't change the requirement in a later beta breaking change.
Perhaps we only want to take some of these changes immediately. Then do internal cleanup. And finally make the break later.

packages/framework/aqueduct/src/container-runtime-factories/baseContainerRuntimeFactory.ts
```diff
@@ -73,22 +73,73 @@ export interface BaseContainerRuntimeFactoryProps {
...
+ * See {@link https://github.com/microsoft/FluidFramework/issues/27851} for context.
+ * @input
+ * @legacy
+ * @beta
+ */
+export type DeprecatedBaseContainerRuntimeFactoryProps = Omit<
+	BaseContainerRuntimeFactoryProps,
+	"oldestSupportedClient" | "minVersionForCollab"
+> & {
+	readonly oldestSupportedClient?: undefined;
```
Josmithr, line 102: Nit: docs? We can use `{@inheritDoc BaseContainerRuntimeFactoryProps .oldestSupportedClient}` to avoid duplication.

packages/framework/aqueduct/src/container-runtime-factories/containerRuntimeFactoryWithDefaultDataStore.ts
```diff
@@ -42,6 +46,36 @@ export interface ContainerRuntimeFactoryWithDefaultDataStoreProps
...
+ * See {@link https://github.com/microsoft/FluidFramework/issues/27851} for context.
+ * @input
+ * @legacy
+ * @beta
+ */
+export type DeprecatedContainerRuntimeFactoryWithDefaultDataStoreProps = Omit<
+	ContainerRuntimeFactoryWithDefaultDataStoreProps,
+	"oldestSupportedClient" | "minVersionForCollab"
+> & {
+	readonly oldestSupportedClient?: undefined;
```
Josmithr, line 63: Nit: docs?

CrossClientCompatibilityDevGuide.md
```diff
@@ -44,7 +44,7 @@ However, there are many other types of changes that could impact cross-client co
...
 
-`oldestSupportedClient` defines the oldest Fluid client version that must be able to access documents written by the runtime. Customers are encouraged to set `oldestSupportedClient` to the highest version their users are saturated on. If the customer does not set `oldestSupportedClient`, a default value is assigned. `oldestSupportedClient` controls the "default configurations" and "unsafe configuration prevention" mechanisms explained below.
+`oldestSupportedClient` defines the oldest Fluid client version that must be able to access documents written by the runtime. Callers must specify exactly one of `oldestSupportedClient` or the deprecated `minVersionForCollab`; omitting both is an error. Customers are encouraged to set `oldestSupportedClient` to the highest version their users are saturated on. `oldestSupportedClient` controls the "default configurations" and "unsafe configuration prevention" mechanisms explained below.
```
Josmithr: Nit: while we're here, would be a good time to update the line formatting to follow our [Markdown best practices](https://github.com/microsoft/FluidFramework/blob/main/docs/content/Guidelines/Documentation-Guidelines/Markdown-Best-Practices.md#line-breaks-along-sentence-boundaries) (line break on sentence boundaries):
```suggestion
`oldestSupportedClient` defines the oldest Fluid client version that must be able to access documents written by the runtime.
Callers must specify exactly one of `oldestSupportedClient` or the deprecated `minVersionForCollab`; omitting both is an error.
Customers are encouraged to set `oldestSupportedClient` to the highest version their users are saturated on.
`oldestSupportedClient` controls the "default configurations" and "unsafe configuration prevention" mechanisms explained below.
```

CrossClientCompatibilityDevGuide.md
```diff
@@ -67,7 +67,7 @@ const runtimeOptionsAffectingDocSchemaConfigMap: ConfigMap<RuntimeOptionsAffecti
 };
 ```
 
-> **Note on `"2.0.0-defaults"`:** This is a special version string (considered less than `"2.0.0"` by `semver`) used as the default when a customer does not explicitly set `oldestSupportedClient`. It exists to distinguish the unspecified case from an explicit `"2.0.0"` setting. Some options (e.g., `explicitSchemaControl`) use a threshold of `"2.0.0"` rather than `"2.0.0-defaults"`, meaning they only activate when the customer _explicitly_ sets `oldestSupportedClient` to `"2.0.0"` or higher. See `defaultMinVersionForCollab` in [compatibilityBase.ts](./packages/runtime/runtime-utils/src/compatibilityBase.ts) for more information.
+> **Note on `"2.0.0-defaults"`:** This special historical version string (considered less than `"2.0.0"` by `semver`) preserves the runtime option defaults that applied before compatibility versions were explicitly configured. It may be selected explicitly by internal and test callers that need those semantics. Some options (e.g., `explicitSchemaControl`) use a threshold of `"2.0.0"` rather than `"2.0.0-defaults"`, so they activate only when `oldestSupportedClient` is `"2.0.0"` or higher.
```
Josmithr: Nit: while we're here, would be a good time to update the line formatting to follow our [Markdown best practices](https://github.com/microsoft/FluidFramework/blob/main/docs/content/Guidelines/Documentation-Guidelines/Markdown-Best-Practices.md#line-breaks-along-sentence-boundaries) (line break on sentence boundaries):
```suggestion
> **Note on `"2.0.0-defaults"`:** This special historical version string (considered less than `"2.0.0"` by `semver`) preserves the runtime option defaults that applied before compatibility versions were explicitly configured.
> It may be selected explicitly by internal and test callers that need those semantics.
> Some options (e.g., `explicitSchemaControl`) use a threshold of `"2.0.0"` rather than `"2.0.0-defaults"`, so they activate only when `oldestSupportedClient` is `"2.0.0"` or higher.
```

packages/framework/aqueduct/src/container-runtime-factories/containerRuntimeFactoryWithDefaultDataStore.ts
```diff
@@ -55,7 +88,21 @@ export class ContainerRuntimeFactoryWithDefaultDataStore extends BaseContainerRu
...
+	public constructor(props: DeprecatedContainerRuntimeFactoryWithDefaultDataStoreProps);
+	/**
+	 * Creates a factory when the compatibility property is selected dynamically.
+	 */
+	public constructor(
+		props:
+			| ContainerRuntimeFactoryWithDefaultDataStoreProps
+			| DeprecatedContainerRuntimeFactoryWithDefaultDataStoreProps,
+	);
+	public constructor(props: ContainerRuntimeFactoryWithDefaultDataStorePropsInternal) {
```
jason-ha: `ContainerRuntimeFactoryWithDefaultDataStorePropsInternal` is `ContainerRuntimeFactoryWithDefaultDataStoreProps | DeprecatedContainerRuntimeFactoryWithDefaultDataStoreProps`.
Why do we declare this two ways? I'm guessing the `.. | ..` override is needed for some reason -- if not delete. There doesn't seem to be enough use of union type to justify it existing.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -825,41 +826,169 @@ export interface LoadContainerRuntimeParams {
...
+	 */
+	containerRuntimeCtor?: typeof ContainerRuntime;
+	/**
+	 * {@link @fluidframework/container-runtime#LoadContainerRuntimeParams.runtimeOptions}, except
+	 * with additional internal-only options.
+	 */
+	runtimeOptions?: IContainerRuntimeOptionsInternal;
+};
+
+function getExplicitOldestSupportedClient(
```
jason-ha: Duplicated in container-runtime-factory. Deduplicate to common location?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -825,41 +826,169 @@ export interface LoadContainerRuntimeParams {
...
+	params: Readonly<{
+		oldestSupportedClient?: OldestSupportedClientVersion;
+		minVersionForCollab?: OldestSupportedClientVersion;
+	}>,
+): OldestSupportedClientVersion {
+	if (params.minVersionForCollab !== undefined) {
+		throw new UsageError(
+			"minVersionForCollab is not supported by loadContainerRuntimeAlpha. Use oldestSupportedClient instead.",
+		);
+	}
```
jason-ha, line 948: This seems a little funny. It could be informative, I guess. Needs comments.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -1039,7 +1151,8 @@ export class ContainerRuntime
 		// with FF runtime 2.10.0 or later.
 		if (!isValidMinVersionForCollab(minVersionForCollab)) {
 			throw new UsageError(
-				`Invalid minVersionForCollab: ${minVersionForCollab}. It must be an existing FF version (i.e. 2.22.1).`,
+				`Invalid compatibility version: ${minVersionForCollab}. ` +
+					"`oldestSupportedClient` (or deprecated `minVersionForCollab`) must be an existing FF version (i.e. 2.22.1).",
```
jason-ha: maybe just list oldestSupportedClient

packages/runtime/container-runtime/src/test/loadContainerRuntime.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+		const acceptAlpha = (_params: Parameters<typeof loadContainerRuntimeAlpha>[0]): void => {};
+		const callDeprecatedOverload = async (): ReturnType<typeof loadContainerRuntime> =>
+			loadContainerRuntime({
+				...commonParams,
+				minVersionForCollab: "2.0.0",
+			});
+		const callEitherOverload = async (
+			params: LoadContainerRuntimeParams | DeprecatedLoadContainerRuntimeParams,
+		): ReturnType<typeof loadContainerRuntime> => loadContainerRuntime(params);
+
```
jason-ha: What are these tests for? They seem to be testing TypeScript type checking - mostly detached from implementation.

packages/runtime/container-runtime/src/test/loadContainerRuntime.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+			/Specify exactly one of oldestSupportedClient or minVersionForCollab/,
+		);
+	});
+
+	it("forwards the deprecated compatibility parameter", async () => {
+		const context = {
+			taggedLogger: { send: () => {} },
+		} as unknown as IContainerContext;
+
+		await assert.rejects(
```
jason-ha: doesn't look like forwarding - looks like rejection

packages/framework/aqueduct/src/container-runtime-factories/containerRuntimeFactoryWithDefaultDataStore.ts
```diff
@@ -76,8 +123,11 @@ export class ContainerRuntimeFactoryWithDefaultDataStore extends BaseContainerRu
 			return undefined; // continue search
 		};
 
+		const oldestSupportedClient = getExplicitOldestSupportedClient(props);
 		super({
 			...props,
+			oldestSupportedClient,
+			minVersionForCollab: undefined,
```
jason-ha: Why pass `undefined`? Can't just leave it out?

packages/framework/aqueduct/src/test/aqueduct.spec.ts
```diff
@@ -3,5 +3,139 @@
...
+
+describe("BaseContainerRuntimeFactory compatibility parameter", () => {
+	it("keeps the canonical construction properties extendable", () => {
+		interface ExtendedProps extends BaseContainerRuntimeFactoryProps {
+			readonly customProperty: boolean;
+		}
+		const props: ExtendedProps = {
+			...commonProps,
+			oldestSupportedClient: "2.0.0",
+			minVersionForCollab: undefined,
```
jason-ha: In common cases `undefined` won't be passed, right? Instead those will be just left out. Can we follow that pattern?

packages/framework/aqueduct/src/test/aqueduct.spec.ts
```diff
@@ -3,5 +3,139 @@
...
+		assert.doesNotThrow(() => new BaseContainerRuntimeFactory(props));
+	});
+
+	it("accepts a dynamically selected compatibility property", () => {
+		const createFactory = (
+			props: BaseContainerRuntimeFactoryProps | DeprecatedBaseContainerRuntimeFactoryProps,
+		): BaseContainerRuntimeFactory => new BaseContainerRuntimeFactory(props);
+
+		assert.equal(typeof createFactory, "function");
+	});
```
jason-ha: no actual call - looks like testing TypeScript

packages/framework/aqueduct/src/container-runtime-factories/containerRuntimeFactoryWithDefaultDataStore.ts
WillieHabi: The explicit union overload is needed so a caller holding  CanonicalProps | DeprecatedProps  can pass that variable directly.  

Deleted *PropsInternal alias and just use the explicit union.

packages/runtime/container-runtime/src/containerRuntime.ts
WillieHabi: getExplicitOldestSupportedClient now lives in Container Runtime’s internal entrypoint, and both Container Runtime and Aqueduct use that single implementation

packages/runtime/container-runtime/src/containerRuntime.ts
WillieHabi: Added comments.. this separate check is intentional. beta temporarily accepts the deprecated property, while alpha accepts only  oldestSupportedClient.

packages/runtime/container-runtime/src/containerRuntime.ts
WillieHabi: Updated

packages/runtime/container-runtime/src/test/loadContainerRuntime.spec.ts
WillieHabi: moved the overload, exact-one, explicit-undefined  alpha, and union-callability checks into src/test/types/loadContainerRuntime.ts . 

Mocha suite now contains only runtime behavior checks.

packages/runtime/container-runtime/src/test/loadContainerRuntime.spec.ts
WillieHabi: yeah name is misleading, renamed it to “normalizes the deprecated compatibility parameter before validation” and documented the mechanism.

packages/framework/aqueduct/src/test/aqueduct.spec.ts
WillieHabi: Yes, removed the hollow typeof assertion and moved the union-callability contract into src/test/types/containerRuntimeFactories.ts.

packages/framework/aqueduct/src/test/aqueduct.spec.ts
WillieHabi: Yes makes sense. Now follow the common pattern and omit the unused property. Explicit undefined coverage still exists, but it moved to the compile-only type tests.

packages/framework/aqueduct/src/container-runtime-factories/containerRuntimeFactoryWithDefaultDataStore.ts
WillieHabi: Updated, code now destructures the key out before forwarding instead of retaining it with an undefined value.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -978,10 +990,14 @@ export async function loadContainerRuntime(
 	params: LoadContainerRuntimeParams | DeprecatedLoadContainerRuntimeParams,
 ): Promise<IContainerRuntime & IRuntime> {
 	const oldestSupportedClient = getExplicitOldestSupportedClient(params);
+	const {
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars -- Omit the deprecated property after normalizing it.
+		minVersionForCollab: _deprecatedMinVersionForCollab,
```
jason-ha: Why does this need destructured at all?

packages/framework/aqueduct/src/test/types/containerRuntimeFactories.ts
```diff
@@ -0,0 +1,79 @@
...
+const defaultFactory = undefined as unknown as IFluidDataStoreFactory;
+
+const canonicalBaseProps: BaseContainerRuntimeFactoryProps = {
+	...commonProps,
+	oldestSupportedClient: "2.0.0",
+};
+const canonicalBasePropsWithExplicitUndefined: BaseContainerRuntimeFactoryProps = {
+	...canonicalBaseProps,
+	minVersionForCollab: undefined,
+};
```
jason-ha: This is not the preferred way to construct this. Omitting the property is more representative of actual use.
The use of `minVersionForCollab?: undefined` and `oldestSupportedClient?: undefined` is unfortunate. I haven't complained about them because I assumed it simplified overloading.
I would be better to just not have those of to have `foo?: never`.

packages/runtime/container-runtime/src/test/loadContainerRuntime.spec.ts
jason-ha: I've not getting the value of the type checks. But won't block on 'em.

packages/framework/aqueduct/src/test/aqueduct.spec.ts
jason-ha: but what/where is the value?

WillieHabi: Of the compile time tests? I think the value is protecting the consumer-facing TypeScript contract introduced by this PR.

packages/framework/aqueduct/src/test/types/containerRuntimeFactories.ts
WillieHabi: removed the explicit undefined from both

WillieHabi: and yeah the `?: undefined`  fields help the union overload reject an object containing both defined compatibility values.

WillieHabi: Using undefined also matches the runtime check, which treats an explicitly undefined property the same as an omitted property. 

Copilot review explained how `?: never` would distinguish those cases for consumers using exactOptionalPropertyTypes.

packages/runtime/container-runtime/src/containerRuntime.ts
WillieHabi: Yeah, looking at it now this is unnecessary. this came trying to to omit the old property rather than pass it as undefined, so destructuring seemed reasonable. removed now

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -873,6 +1012,7 @@ export async function loadContainerRuntime(
 export async function loadContainerRuntimeAlpha(params: LoadContainerRuntimeParams): Promise<{
 	runtime: IContainerRuntime & IRuntime;
 }> {
+	getAlphaOldestSupportedClient(params);
```
jason-ha: Why?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -825,41 +826,179 @@ export interface LoadContainerRuntimeParams {
...
+ * @legacy
+ * @beta
+ */
+export function loadContainerRuntime(
+	params: LoadContainerRuntimeParams | DeprecatedLoadContainerRuntimeParams,
+): Promise<IContainerRuntime & IRuntime>;
+export async function loadContainerRuntime(
+	params: LoadContainerRuntimeParams | DeprecatedLoadContainerRuntimeParams,
 ): Promise<IContainerRuntime & IRuntime> {
+	getExplicitOldestSupportedClient(params);
```
jason-ha: Why?

packages/runtime/container-runtime/src/test/loadContainerRuntime.spec.ts
```diff
@@ -22,25 +25,76 @@ const commonParams = {
...
+	afterEach(() => {
+		sandbox.restore();
+	});
+
+	const loadContainerRuntimeWithSelectedParams = (
+		params: LoadContainerRuntimeParams | DeprecatedLoadContainerRuntimeParams,
+	): Promise<unknown> => loadContainerRuntime(params);
+
+	it("forwards the canonical compatibility parameter to the internal loader", async () => {
+		const loadRuntime = sandbox.stub(ContainerRuntime, "loadRuntime").resolves(runtime);
```
jason-ha: This appears to now be testing itself

jason-ha (commented):
Prior to last commit I would have approved with: I still question the type tests being detached from runtime.
I don't think last test changes are better - or at least some appear worse.

packages/runtime/container-runtime/src/test/loadContainerRuntime.spec.ts
WillieHabi: removed

packages/runtime/container-runtime/src/containerRuntime.ts
WillieHabi: I renamed this to  validateAlphaOldestSupportedClient , changed it to return  void, and added a callsite comment

packages/runtime/container-runtime/src/containerRuntime.ts
WillieHabi: Added a comment

packages/runtime/container-runtime/src/containerRuntime.ts
jason-ha: There was a fix for `loadContainerRuntimeAlpha`, but `loadContainerRuntime` get... remains.

packages/runtime/container-runtime/src/test/loadContainerRuntime.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+			taggedLogger: { send: () => {} },
+		} as unknown as IContainerContext;
+
+		// Reaching version validation proves the public exact-one guard accepted the deprecated
+		// property. The invalid version avoids requiring a complete container context in this test.
+		await assert.rejects(
+			loadContainerRuntime({
+				...commonParams,
+				context,
+				minVersionForCollab: "1.2.3.4" as OldestSupportedClientVersion,
```
jason-ha: prefer `// @ts-expect-error` for consistency

packages/framework/aqueduct/src/test/aqueduct.spec.ts
```diff
@@ -3,5 +3,126 @@
...
+	new ContainerRuntimeFactoryWithDefaultDataStore(props);
+
+describe("BaseContainerRuntimeFactory compatibility parameter", () => {
+	it("keeps the canonical construction properties extendable", () => {
+		interface ExtendedProps extends BaseContainerRuntimeFactoryProps {
+			readonly customProperty: boolean;
+		}
+		const props: ExtendedProps = {
+			...commonProps,
+			oldestSupportedClient: "2.0.0",
```
jason-ha: Lots of hard to maintain "2.0.0" use.
`defaultTestOldestSupportedClient`  or similar?

packages/runtime/container-runtime/src/test/loadContainerRuntime.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+			/Specify exactly one of oldestSupportedClient or minVersionForCollab/,
+		);
+	});
+
+	it("rejects both compatibility parameters at runtime", async () => {
+		await assert.rejects(
+			// @ts-expect-error Exactly one compatibility property may be supplied.
+			loadContainerRuntime({
+				...commonParams,
+				oldestSupportedClient: "2.0.0",
```
jason-ha: more "2.0.0"

packages/runtime/container-runtime/src/containerRuntime.ts
jason-ha, line 1139: Seems like it would be better to replace this with `validateExplicitOldestSupportedClient` renamed back to `getExplicitOldestSupportedClient`. Am I missing a case when we'd still need `defaultMinVersionForCollab`? Doesn't the title of this imply no more default?
I'd love to see `defaultMinVersionForCollab` be removed -- was asking about the plan for that in the prior PR.
There isn't need to validate in flat function.

Also, the lines 1132-1134 comments suggest that `minVersionForCollab` from here is passed to other layers (from here) under that name, but I don't see that. `DocumentSchemaController` (this layer) does use that name because it is persisted and won't be changing. Otherwise, later `updatedMinVersionForCollab` is computed from it and that is passed as an unnamed parameter to `containerRuntimeCtor`.

jason-ha: (That grabbed a lot of lines - I was selecting 1124-1138 when I commented.)

WillieHabi: > Am I missing a case when we'd still need defaultMinVersionForCollab

You’re not missing a customer-facing case (implicit default is already gone from  loadContainerRuntime, alpha, and the Aqueduct factories), the remaining fallback applies only to direct internal static-loader callers. But I audited those callers with copilot and found 76 direct object-literal calls that currently omit both properties, (almost all in Container Runtime tests, plus replay-tool, mixinAttributor also currently calls the base loader without forwarding either property). Moving getExplicitOldestSupportedClient into loadRuntime2 is the cleaner end state, but it requires migrating those callers first,so  it isn’t a local replacement.

The other distinction I found when originally making this PR is that removing the implicit fallback is feasible after that migration, **but** removing 2.0.0-defaults is not. It remains a semantic historical value -- for example, it selects  explicitSchemaControl: false, while explicit  2.0.0  selects true. I remember removal caused issues with the replay tool tests due to this behavior. We could rename it separately, but we cannot collapse it into  2.0.0  without changing behavior.

I’d prefer to handle the internal caller migration as a focused follow-up so this PR remains scoped to the customer-facing break. I’ve updated the PR title accordingly and also rewrote the nearby comment to describe the actual propagation path.

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -825,41 +826,181 @@ export interface LoadContainerRuntimeParams {
...
-	 * Specifying both `oldestSupportedClient` and `minVersionForCollab` is an error.
+	 * This property must not be provided. Use `oldestSupportedClient` instead.
 	 *
-	 * @deprecated 2.116.0. To be removed in 3.10.0. Use
-	 * {@link LoadContainerRuntimeParams.oldestSupportedClient} instead.
+	 * @deprecated 2.116.0. To be removed in 3.10.0. Use `oldestSupportedClient` instead.
 	 * See {@link https://github.com/microsoft/FluidFramework/issues/27851} for context.
 	 */
-	minVersionForCollab?: OldestSupportedClientVersion;
+	minVersionForCollab?: undefined;
```
jason-ha, line 837: I was suspicious of the dual properties in both param types.
I confirmed that you can remove this and retain expected errors in test cases. But you would have to keep something on the deprecated (preferably `oldestSupportedClient?: never` but `undefined` is fine.)
With that update the errors are better. Setting both yields:
```text
No overload matches this call.
  Overload 1 of 3, '(params: LoadContainerRuntimeParams): Promise<IContainerRuntime & IRuntime>', gave the following error.
    Object literal may only specify known properties, and 'minVersionForCollab' does not exist in type 'LoadContainerRuntimeParams'.
  Overload 2 of 3, '(params: DeprecatedLoadContainerRuntimeParams): Promise<IContainerRuntime & IRuntime>', gave the following error.
    Type 'string' is not assignable to type 'never'.
  Overload 3 of 3, '(params: LoadContainerRuntimeParams | DeprecatedLoadContainerRuntimeParams): Promise<IContainerRuntime & IRuntime>', gave the following error.
    Object literal may only specify known properties, and 'minVersionForCollab' does not exist in type 'LoadContainerRuntimeParams'.
containerRuntime.ts(846, 11): The expected type comes from property 'oldestSupportedClient' which is declared here on type 'DeprecatedLoadContainerRuntimeParams'
```
Not required to be cleaned up in this PR.

.changeset/little-beers-guess.md
```diff
@@ -0,0 +1,34 @@
...
+[`BaseContainerRuntimeFactory`](https://fluidframework.com/docs/api/aqueduct/basecontainerruntimefactory-class)
+and
+[`ContainerRuntimeFactoryWithDefaultDataStore`](https://fluidframework.com/docs/api/aqueduct/containerruntimefactorywithdefaultdatastore-class)
+no longer choose a compatibility version when one is omitted. New code must specify
+[`oldestSupportedClient`](https://fluidframework.com/docs/api/container-runtime/loadcontainerruntimeparams-interface#oldestsupportedclient-propertysignature).
+Typed callers that temporarily retain the deprecated `minVersionForCollab` property must use
+[`DeprecatedLoadContainerRuntimeParams`](https://fluidframework.com/docs/api/container-runtime/deprecatedloadcontainerruntimeparams-typealias),
+[`DeprecatedBaseContainerRuntimeFactoryProps`](https://fluidframework.com/docs/api/aqueduct/deprecatedbasecontainerruntimefactoryprops-typealias),
+or
+[`DeprecatedContainerRuntimeFactoryWithDefaultDataStoreProps`](https://fluidframework.com/docs/api/aqueduct/deprecatedcontainerruntimefactorywithdefaultdatastoreprops-typealias),
```
jason-ha, line 24: The impacted non-deprecated types should be mentioned here. If the type (x3) was mentioned in package.json because things were broken, then call it out.
(Recommend removing the unneeded `minVersionForCollab?: undefined` from those types.)
Can both be done in follow-up.

packages/framework/aqueduct/src/container-runtime-factories/baseContainerRuntimeFactory.ts
```diff
@@ -73,24 +73,42 @@ export interface BaseContainerRuntimeFactoryProps {
...
-	 * Specifying both `oldestSupportedClient` and `minVersionForCollab` is an error.
+	 * This property must not be provided. Use `oldestSupportedClient` instead.
 	 *
-	 * @deprecated 2.116.0. To be removed in 3.10.0. Use
-	 * {@link BaseContainerRuntimeFactoryProps.oldestSupportedClient} instead.
+	 * @deprecated 2.116.0. To be removed in 3.10.0. Use `oldestSupportedClient` instead.
 	 * See {@link https://github.com/microsoft/FluidFramework/issues/27851} for context.
 	 */
-	minVersionForCollab?: OldestSupportedClientVersion | undefined;
+	minVersionForCollab?: undefined;
```
jason-ha, line 84: It doesn't look like this change can be made without first updating bohemia.
https://office.visualstudio.com/OC/_search?action=contents&text=BaseContainerRuntimeFactoryProps&type=code&lp=buildrelease-Team&filters=ProjectFilters%7BOC%7DRepositoryFilters%7BLoopSDK*office-bohemia%7D&pageSize=25&result=DefaultCollection/OC/office-bohemia/GBmaster//packages/octopus-document-model/src/fluid/fluid-container-wrapper.ts
Do you have a successful Pages integration run with these changes?

jason-ha (commented):
Suspect breaks Pages integration...
  otherwise could approve.

packages/framework/aqueduct/src/container-runtime-factories/baseContainerRuntimeFactory.ts
WillieHabi: I refreshed the prep branch I used for the initial FF -> Pages run and opened a PR to close gaps in test compatilituy floors. I reran Pages against the refreshed branch with your exact FF build. All typechecks now pass, including the MFS failure from your run: https://office.visualstudio.com/OC/_build/results?buildId=54274093.

.changeset/little-beers-guess.md
WillieHabi: updated the changeset to explicitly name LoadContainerRuntimeParams, BaseContainerRuntimeFactoryProps, and  ContainerRuntimeFactoryWithDefaultDataStoreProps as the existing inputs whose compatibility changes

packages/runtime/container-runtime/src/containerRuntime.ts
WillieHabi: Ok makes sense. One tradeoff I found is that a prebuilt or spread object containing both real properties can structurally match the canonical type once minVersionForCollab?: undefined is removed. The runtime exact-one check still rejects it, so nothing fails silently, but yes this can cleaned up in follow up and I’ll capture that case in the follow-up tests. Thanks

PR #27977, opened 2026-08-15
feat(tree): optional commit validation

yann-achard-MS:
## Description

Adds the following new options to `SharedTreeOptions` (alpha):
- 1validateCommitsOnFirstSubmission1: (default: `false`) When `true`, validates that commits being submitted for the first time can be applied without errors to a view.
- 1validateRebasedCommitsBeforeResubmission`: (default: `false`) When `true`, validates that the commits being resubmitted can be applied without errors to a view.

In the event that a commit cannot be applied, SharedTree will throw an error and will enter a "broken" state, preventing the offending commit (and any further commits) from being submitted.
This can be enabled (at the cost of performance) to improve safety against document corruption in the event of a bug in the SharedTree code:
when the additional validation is enabled, a client will error instead of potentially corrupting the document.

## Breaking Changes

None

## Reviewer Guidance

It's likely that we'll want to enable one or both flags by default in the future. This should be done after we've had the chance to use the new flags, evaluate/improve their perf impact, and consider different validation strategies.

### DIFF
--- a/packages/dds/tree/src/shared-tree-core/branchCommitEnricher.ts
+++ b/packages/dds/tree/src/shared-tree-core/branchCommitEnricher.ts
@@ -28,12 +28,18 @@ export class BranchCommitEnricher<TChange> {
 	/**
 	 * Process the given commits for later {@link BranchCommitEnricher.retrieveChange | retrieval}.
 	 * @param commits - The commits to prepare.
+	 * @param forceValidation - Whether to force the application of the enriched commits to a side checkout.
+	 * This is useful to reduce the likelihood of an invalid commit being generated
+	 * (either due to the given `commits` being invalid or because of an error in the enrichment logic).
 	 */
-	public prepareChanges(commits: readonly GraphCommit<TChange>[]): void {
+	public prepareChanges(
+		commits: readonly GraphCommit<TChange>[],
+		forceValidation: boolean,
+	): void {
 		if (hasSome(commits)) {
 			const startingState = commits[0].parent;
 			assert(startingState !== undefined, 0xcc1 /* New commits must have a parent. */);
-			const enrichedCommits = this.enricher.enrich(startingState, commits);
+			const enrichedCommits = this.enricher.enrich(startingState, commits, forceValidation);
 			for (const [index, commit] of commits.entries()) {
 				const enrichedCommit = enrichedCommits[index];
 				assert(enrichedCommit !== undefined, 0xcc2 /* Missing enriched commit. */);
--- a/packages/dds/tree/src/shared-tree/sharedTreeChangeEnricher.ts
+++ b/packages/dds/tree/src/shared-tree/sharedTreeChangeEnricher.ts
@@ -110,7 +110,10 @@ export class SharedTreeChangeEnricher {
 		this.changeQueue.push(typeof change === "function" ? change : () => change);
 	}
 
-	private purgeChangeQueue(): void {
+	/**
+	 * Applies all queued changes.
+	 */
+	public purgeChangeQueue(): void {
 		if (this.changeQueue.length === 0) {
 			return;
 		}
--- a/packages/dds/tree/src/shared-tree/treeCheckout.ts
+++ b/packages/dds/tree/src/shared-tree/treeCheckout.ts
@@ -1766,11 +1766,7 @@ export class TreeCheckout implements ITreeCheckout {
 			// This is useful for catching bugs in the enrichment logic and in the rebasing logic that may have been involved in producing the changes.
 			// Note that we cannot guarantee that the enriched change will be applied successfully even in the absence of concurrent changes,
 			// because the checkout state we are applying this enriched change to was derived from the state of the checkout, which may be different from that of a peer.
-			try {
-				enricher.purgeChangeQueue();
-			} catch (error) {
-				throw new Error(`Error during enrichment validation: ${error}`);
-			}
+			enricher.purgeChangeQueue();
 		}
 		enricher[disposeSymbol]();
 		return enrichedChanges;
--- a/packages/dds/tree/src/shared-tree/treeCheckout.ts
+++ b/packages/dds/tree/src/shared-tree/treeCheckout.ts
@@ -1720,6 +1720,7 @@ export class TreeCheckout implements ITreeCheckout {
 	public enrich(
 		context: GraphCommit<SharedTreeChange>,
 		changes: readonly TaggedChange<SharedTreeChange>[],
+		forceValidation: boolean,
 	): SharedTreeChange[] {
 		if (!hasSome(changes)) {
 			return [];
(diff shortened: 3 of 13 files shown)


packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1750,13 +1751,29 @@ export class TreeCheckout implements ITreeCheckout {
...
+		}
+		if (forceValidation) {
+			// Ensure that all changes can be applied successfully.
+			// This is useful for catching bugs in the enrichment logic and and in the rebasing logic that may have been involved in producing the changes.
+			// Note that we cannot guarantee that the enriched change will be applied successfully even in the absence of concurrent changes,
+			// because the checkout state we are applying this enriched change to was derived from the state of the checkout, which may be different from that of a peer.
+			try {
+				enricher.purgeChangeQueue();
+			} catch (error) {
+				throw new Error(`Error during enrichment validation: ${error}`);
```
yann-achard-MS: @CraigMacomber - Is there a better pattern to use here? See comment in the test. #Resolved

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
```diff
@@ -2130,6 +2166,46 @@ describe("sharedTreeView", () => {
...
+
+			fork.root.insertAtEnd({ id: "A" });
+
+			// This hides the commit that inserted "A".
+			fork.checkout.mainBranch.setHead(view1.checkout.mainBranch.getHead());
+
+			fork.root.removeAt(0);
+
+			// The merge will fail because only the commit that removes "A" will be merged.
+			assert.throws(() => view1.merge(fork));
```
yann-achard-MS, line 2204: @CraigMacomber - I tried to assert that the error mached `/Error during enrichment validation/` but that didn't work out. #Resolved

packages/dds/tree/src/shared-tree/treeCheckout.ts
CraigMacomber: Making use of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause might make sense for a case like this. We make some minimal use of it currently, but will be able to use it more generally after the es2022 bump in 3.0.

If this error is supposed to imply a bug, catching it like this might be harmful. If the underlying error is an assert for example, this will replace it with an error who's message is not known to be telemetry safe and can't be logged. Additionally you would lose some of the stack information. I'm not sure about the best option here, but rethrowing the same error, but adding scope information to it, and maybe flagging that information as telemetry safe might be useful, but if we get a call stack going through this code, that's actually redundant and maybe just not catching it at all would be better.

I generally try to avoid catch blocks, and almost always rethrow the same exception in a catch. If not, stashing the previous on in the cause is generally my preferred approach (thats what the Breakable code does)

packages/dds/tree/src/test/shared-tree/treeCheckout.spec.ts
CraigMacomber: Not sure how you tried to assert that, but it should be possible. Maybe look at how validateUsageError is used and implemented for an example?

packages/dds/tree/src/shared-tree/treeCheckout.ts
yann-achard-MS: I removed the try-catch block. The callstack should tell us what we need to know. Also, the same kind of error can arise without this forced validation being turned on and there's no reason to treat those differently.

packages/dds/tree/src/shared-tree-core/branchCommitEnricher.ts
```diff
@@ -28,12 +28,18 @@ export class BranchCommitEnricher<TChange> {
 	/**
 	 * Process the given commits for later {@link BranchCommitEnricher.retrieveChange | retrieval}.
 	 * @param commits - The commits to prepare.
+	 * @param forceValidation - Whether to force the application of the enriched commits to a side checkout.
+	 * This is useful to reduce the likelihood of an invalid commit being generated
```
CraigMacomber, line 32: This makes it sound like it reduces the likelihood of invalid commits being generated, which I think is inaccurate (or at least misleading)

Would it be more accurate to say:

```suggestion
	 * This is useful to crash on some cases of invalid commits instead of submitting them,
	 * keeping corruption and data loss from spreading from this client into the persisted document.
```

CraigMacomber: Looks like several other places have very similar documentation. I think it might be best to simply update the docs on the actual package API surface (the config property), including private remarks as appropriate, and link that from all these places.

Maybe something like: `Attempt to {link to validation flag|validate} commits before sending them to prefer crashing locally over corrupting persisted data.`

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1750,13 +1751,25 @@ export class TreeCheckout implements ITreeCheckout {
...
 		for (const change of changes) {
-			enriched.push(enricher.enrich(change.change));
-			enricher.enqueueChange(change);
+			const enrichedChange = enricher.enrich(change.change);
+			enrichedChanges.push(enrichedChange);
+			// We could either apply the original or the enriched change: either should allow subsequent changes to be enriched correctly.
+			// We choose to apply the enriched change in order to more closely replicate what a peer will experience (assuming no concurrent changes).
+			// This makes it more likely that we'll detect a bug in the enrichment logic.
+			// Note that enqueueing the change does not guarantee that it will be applied unless `forceValidation` is enabled.
+			enricher.enqueueChange(tagChange(enrichedChange, change.revision));
```
CraigMacomber, line 1762: This is a change in behavior, even when not opting in right? Meaning we might apply some changes (and those will use the enriched version), when force validation is off, just not all of them?

packages/dds/tree/src/shared-tree/sharedTreeChangeEnricher.ts
```diff
@@ -110,7 +110,10 @@ export class SharedTreeChangeEnricher {
 		this.changeQueue.push(typeof change === "function" ? change : () => change);
 	}
 
-	private purgeChangeQueue(): void {
+	/**
+	 * Applies all queued changes.
```
CraigMacomber, line 114: This could use more contexts. The class it's on has no doc comment, so its not easy to discover what this is applying the changes to, or what the implications of that are.

Maybe add a doc comment to SharedTreeChangeEnricher, and a remarks block here indicating if this results in modifying views, flushing pending work, validation etc.

CraigMacomber: Looks like this relies on cloning forests in the non-owned case. I'm not sure of how often the non owned case happens, but thats O(size of forest) for the reference forest implementation which is still the default.
Maybe we want to recommend using the optimized forest if opting into validation if that results in a lot of cloning?

One might generally expect an app wanting to maximize debugging info to explicitly use the reference or debug expensive forests, so this might be a good detail to include.

CraigMacomber (approved):
User facing docs look good, and behavior of code when not opting in seems probably safe, but I left a question and some suggestions for improving our internal facing docs.

I have not reviewed the tests in too much detail and am not really confident about the implications of opting in as the docs make it a bit hard to understand and I don't have a great preexisting understanding of this part of the code.

packages/dds/tree/src/shared-tree/treeCheckout.ts
yann-achard-MS: You are correct, I should have called that out in the PR description. The validation flag impacts whether changes are applied, while this code change impacts which version of the changes are applied when applied. Before this change, we would apply the original (non-enriched) version, while we now applied the enriched version.

PR #27979, opened 2026-08-17
build(client): bump build-tools to 0.67.0

steffenloesch:
## Description

Bumps the root `buildTools` catalog from 0.66.0 to the newly released 0.67.0 packages and regenerates the lockfile.

Build-tools 0.67.0 is published on npm, tagged `latest`, and backed by `build-tools_v0.67.0`.

## Reviewer Guidance

The lockfile diff contains only build-tools 0.66.0 → 0.67.0 references and the four corresponding tarball integrity hashes. No `pnpm dedupe` was run.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,17 +8,17 @@ settings:
 catalogs:
   buildTools:
     '@fluid-tools/build-cli':
-      specifier: ^0.66.0
-      version: 0.66.0
+      specifier: ^0.67.0
+      version: 0.67.0
     '@fluid-tools/build-infrastructure':
-      specifier: ^0.66.0
-      version: 0.66.0
+      specifier: ^0.67.0
+      version: 0.67.0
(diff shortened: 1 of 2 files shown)


jason-ha:
What is this comment about in the PR description:
> Unlike the prior 0.66.0 catalog bump, current main uses @arethetypeswrong/core 0.18.5 with fflate 0.8.3, so the earlier ATTW decompression failure is not present.
Feels like it might be noise.

steffenloesch:
> What is this comment about in the PR description:
> 
> > Unlike the prior 0.66.0 catalog bump, current main uses @arethetypeswrong/core 0.18.5 with fflate 0.8.3, so the earlier ATTW decompression failure is not present.
> > Feels like it might be noise.

That's fair. It's not too relevant and I've removed it. It was referring to the build break in the previous build-tools bump PR #27856, and pointed out that this issue is not present in this PR.

jason-ha (approved):

PR #27982, opened 2026-08-17
[3.0 Breaking Change] feat(core-interfaces): ITelemetryBaseLogger.send will require LogLevel

### LINKED ISSUE #27595: `ITelemetryBaseLogger.send` will require `LogLevel` in v3.0
## Deprecated API

Calling `ITelemetryBaseLogger.send` without defined `logLevel` parameter is deprecated.
`ITelemetryLoggerExt.send`, which is inherited from `ITelemetryBaseLogger` is similarly impacted.

**Context**

Explicit specification of a `LogLevel` ensures consideration of logging intent. With consistent specification of this property advance logging features such as filtering/sampling are possible.

**Approach**

Always pass `LogLevel` when calling `ITelemetryBaseLogger.send`.

**Dependencies**

none

**Compatibility Concerns**

Implementations of `ITelemetryBaseLogger` are not impacted. While the parameter will become required when `send` without `logLevel` is removed, it is best to leave implementations allowing it to be optional if any layers may be using older definitions of `ITelemetryBaseLogger` and not passing a defined `LogLevel`.

**Phases**

`ITelemetryBaseLogger.send` without defined `logLevel` is deprecated in v2.111.

**Expected Timeline**

`ITelemetryBaseLogger.send` without defined `logLevel` is expected to be removed in v3.0.

MarioJGMsoft:
## Description

Makes `logLevel` required on `ITelemetryBaseLogger.send`, so logging intent is explicit at every call site — which is what makes consistent filtering and sampling possible. This is the Client 3.0 removal of the deprecated "call `send` without a `logLevel`" behavior.

**The requirement is caller-facing and compile-time only.** No dispatch behavior changes: every logger in the repo still forwards the `logLevel` it was handed, so an older caller that invokes `send(event)` with one argument produces exactly the same result before and after this PR. What changes is that new code can no longer omit the level by accident.

**Implementations must keep tolerating a missing `logLevel`.** Fluid runs against a mix of package versions, so code compiled before this change still calls `send(event)` with one argument, and will for as long as those versions are supported. 

Call sites across the repo pass an explicit level, Fluid's own logger implementations are updated to the new signature, and API reports are regenerated.

Closes: #27595
Fixes: [AB#77258](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/77258)

## Breaking Changes

`ITelemetryBaseLogger.send` and the inherited `ITelemetryLoggerExt.send` now require a `LogLevel`.

**Callers** must pass a level. Use `LogLevel.essential` to preserve previous behavior:

```typescript
// Before
logger.send({ category: "generic", eventName: "ExampleEvent" });
// After
logger.send({ category: "generic", eventName: "ExampleEvent" }, LogLevel.essential);
```

See [Breaking-vs-Non-breaking-Changes](https://github.com/microsoft/FluidFramework/blob/main/docs/content/Contributing/Breaking-vs-Non-Breaking-Changes.md) for details.

## Reviewer Guidance

The review process is outlined in [the pull request guidelines](https://github.com/microsoft/FluidFramework/blob/main/docs/content/Contributing/PR-Guidelines.md#guidelines).

- `EventAndErrorTrackingLogger`, `createSampledLogger`, and the test loggers) to declare it required. That is safe here because each one only forwards the value it receives and never branches on it, but it does mean the repo does not model the guidance it gives. Worth deciding whether that is acceptable or whether these should match the documented shape.
- The `ChildLogger` back-compat tests keep exercising a one-argument `send` via `@ts-expect-error`. That preserves the runtime coverage for older callers, but it also means those call sites will go quiet if the underlying behavior ever regresses to something the compiler accepts. Open to a better way to express "this is what an older layer does".


Abe27342 (approved):
approving for tree and merge-tree. didn't look at unrelated files in detail

.changeset/shy-words-brush.md
```diff
@@ -0,0 +1,43 @@
...
+"@fluidframework/core-interfaces": minor
+"@fluidframework/telemetry-utils": minor
+"@fluidframework/azure-client": minor
+"fluid-framework": minor
+"__section": breaking
+"__highlight": true
+---
+Require a log level for every telemetry event
+
+The `logLevel` parameter of `ITelemetryBaseLogger.send` and `ITelemetryLoggerExt.send` is now required.
```
Josmithr: Nit: let's link to the API docs for these.

Josmithr: Also probably link to `LogLevel` below.

packages/framework/client-logger/app-insights-logger/src/fluidAppInsightsLogger.ts
```diff
@@ -180,7 +181,7 @@ class FluidAppInsightsLogger implements ITelemetryBaseLogger {
 	 * This method also uses the provided {@link FluidAppInsightsLoggerConfig} to
 	 * determine whether an event should be sent or not.
 	 */
-	public send(event: ITelemetryBaseEvent): void {
+	public send(event: ITelemetryBaseEvent, _logLevel: LogLevel): void {
```
Josmithr: Should this be updated to do something with the provided log level? If not, we should probably at least add documentation to this method noting that the parameter is ignored.

MarioJGMsoft: Applied changes and added comment

Josmithr (approved):
API and docs changes look good. Didn't review beyond those in detail.

.changeset/shy-words-brush.md
MarioJGMsoft: Linked the requested links to the changeset

jason-ha (dismissed):
Split this PR into two pieces.
Internal fixes to pass logLevel all the time should be in the first.
The breaking require PR should be tiny.
When you get to implementation, be very careful to not require logLevel as there is a long layer compat trail here and the implementations should continue handling `undefined` for some time. Worth commenting. (Ideally there'd be two interfaces - one for callers and one for implementors. Mocks are mostly okay to expect defined value -- so long as they don't work with old packages.)

packages/common/core-interfaces/src/logger.ts
```diff
@@ -100,15 +100,22 @@ export type LogLevel = (typeof LogLevel)[keyof typeof LogLevel];
 /**
  * Interface to output telemetry events.
  * Implemented by hosting app / loader
+ *
+ * @remarks
+ * Logger implementations that may be reached by callers compiled against older Fluid packages must
+ * continue to declare their `send` implementation with an optional `logLevel` parameter and treat
+ * an omitted level as {@link LogLevelConst.essential} before filtering or forwarding. This protects
+ * mixed-version deployments where older callers can still invoke `send(event)` with one argument.
+ *
```
jason-ha, line 109: I think this should also be marked as `@input` as mostly for customers they provide one.

packages/service-clients/end-to-end-tests/azure-client/src/test/AzureClientFactory.ts
```diff
@@ -178,9 +182,15 @@ export function createAzureClientLegacy(
...
 	};
+	const baseLogger = getLogger();
 	return new AzureClientLegacy({
 		connection: connectionProps,
-		logger: getLogger(),
+		logger:
+			baseLogger === undefined
+				? undefined
+				: {
+						send: (event: ITelemetryBaseEvent) => baseLogger.send(event, LogLevel.essential),
```
jason-ha, line 192: proxy minLogLevel?

packages/common/core-interfaces/src/logger.ts
MarioJGMsoft: So then we're fine with the api showing that it's `@public @input`?

packages/service-clients/end-to-end-tests/azure-client/src/test/AzureClientFactory.ts
MarioJGMsoft: Because this is the AzureClientLegacy, it uses the old ITelemetryBaseLogger which doesn't expect minLogLevel. So I think it's okay to leave it as is.

packages/common/core-interfaces/src/logger.ts
jason-ha: Matches my understanding of https://github.com/microsoft/FluidFramework/blob/main/docs/content/Guidelines/Documentation-Guidelines/Documenting-TypeScript/TSDoc-Guidelines.md#input

packages/service-clients/end-to-end-tests/azure-client/src/test/AzureClientFactory.ts
jason-ha: More that doesn't expect - it is not even optional

jason-ha (approved):

PR #27986, opened 2026-08-17
Set  createBlobPayloadPending minVersionForCollab>=3.0.0.

dannimad:
## Description

When `minVersionForCollab` is set to `"3.0.0"` or later, the document schema now declares support for the
`createBlobPayloadPending` capability by default, instead of requiring callers to opt in explicitly.

Introduced a separate, unexported `validateConfigMapKeyVersion` helper that validates configMap keys without
the upper-bound check. The `minVersionForCollab` value used to look up a config is still validated normally
(via `validateMinimumVersionForCollab`) and can never itself exceed the current package version, so a configMap
entry for a future version will simply never be selected until the package version catches up to it.

Added regression tests confirming:
- a configMap may contain an entry beyond the current package version without throwing, and that entry is
  never selected until reachable
- passing a future version as the `minVersionForCollab` argument itself still throws — i.e. the real
  upper-bound check on caller-supplied values was not bypassed
- `validateMinimumVersionForCollab` continues to reject out-of-range versions

## Reviewer Guidance

- Please double check the reasoning for decoupling the schema default from the client-side behavior default —
  documented in `CrossClientCompatibilityDevGuide.md` as a pattern to reuse for future schema-affecting options.
- Please double check the `validateConfigMapKeyVersion` split in `compatibilityBase.ts`: it intentionally skips
  the "not greater than current package version" check that `validateMinimumVersionForCollab` enforces, since
  configMap keys are static declarations of future defaults rather than live compatibility claims made by a
  caller.


packages/runtime/runtime-utils/src/compatibilityBase.ts
```diff
@@ -165,11 +165,13 @@ export function getConfigForMinVersionForCollabIterable<T>(
...
 ): T {
-	// Validate and strongly type the versions from the configMap.
+	// Validate and strongly type the versions from the configMap. configMap keys may describe defaults for
+	// versions later than the current package version (e.g. to prepare ahead of a future major release), so we
+	// don't enforce the upper-bound check here.
 	const versions: [OldestSupportedClientVersion, T][] = Array.from(
 		entries,
 		([version, value]) => {
-			validateMinimumVersionForCollab(version);
+			validateConfigMapKeyVersion(version);
```
WillieHabi, line 175: Seems like without validateMinimumVersionForCollab validating minVersionForCollab, we could have an out-of-range lookup could now select a future entry. 

Think we can call validateMinimumVersionForCollab(minVersionForCollab) at the start of the function(?)

WillieHabi:
> containerRuntime.ts tracks whether createBlobPayloadPending was explicitly requested (mirroring the
existing enableRuntimeIdCompressor explicit-vs-defaulted handling) and only that explicit value gates
BlobManager's actual behavior.

I see this in the PR description, is this explicit-request tracking in this PR? 
Old comment. Not included in the PR

packages/runtime/container-runtime/src/containerCompatibility.ts
```diff
@@ -157,10 +157,10 @@ const runtimeOptionsAffectingDocSchemaConfigMap: ConfigMap<RuntimeOptionsAffecti
 			// "3.0.0": { enableGCSweep: true },
 		},
 		createBlobPayloadPending: {
-			// This feature is new and disabled by default. In the future we will enable it by default, but we have not
-			// closed on the version where that will happen yet.  Probably a .10 release since blob functionality is not
-			// exposed on the `@public` API surface.
+			// This feature is disabled by default for clients older than 3.0.0, and enabled by default starting with
+			// minVersionForCollab>=3.0.0.
 			"1.0.0": undefined,
+			"3.0.0": true,
```
WillieHabi, line 163: nit: maybe add a changeset? or are we planning to add one in another PR

jason-ha, line 163: I believe the proposal is to set this true for 2.40.0.
@ChumpChief ?

PR #27988, opened 2026-08-17
feat(container-runtime): support reusable pending local state on containers without changes

Abe27342:

## Description

Allow pending local state without pending operations or attachment blobs to initialize multiple containers safely. Such captures omit session-scoped runtime state, including the ongoing ID compressor session, and instead reconstruct conceptual state from the snapshot and sequenced operations.

Add the `@legacy @beta` `isPendingLocalStateReusable` API so callers can conservatively detect reusable captures. Update pending-state documentation, generated API reports, tests, and release notes guidance accordingly.

This supports communicating an exact container state across processes or machines instead of loading the latest service state.

## Reviewer Guidance

Please review the conservative reusable-state predicate and the decision to omit all pending runtime state for quiescent captures.

### DIFF
--- a/packages/common/container-definitions/src/loader.ts
+++ b/packages/common/container-definitions/src/loader.ts
@@ -470,8 +470,14 @@ export interface IContainer extends IEventProvider<IContainerEvents> {
 	 * The container's {@link IContainer.closed} property must be false and the container must not
 	 * be {@link IContainer.disposed}; otherwise the implementation throws `UsageError`.
 	 *
-	 * WARNING: misuse of this API can result in duplicate op submission and potential
-	 * document corruption. To prevent container forking, callers must:
+	 * WARNING: Use `isPendingLocalStateReusable` from `@fluidframework/container-runtime/legacy` to
+	 * determine whether the returned state can safely be used to load multiple containers.
+	 * Reusable pending state is generally limited to containers that have not made edits. This can
+	 * be useful for communicating across processes or machines some exact container state to load
+	 *
+	 * Unless `isPendingLocalStateReusable` returns `true`, misuse of this API can result in
+	 * duplicate op submission and potential document corruption. To prevent container forking,
+	 * callers must:
 	 *
 	 * - Regenerate the serialization on every reconnect and replace any previously stored copy.
 	 * - Wait until the original container has been closed before rehydrating from its serialized state.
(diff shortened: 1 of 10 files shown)


packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -5657,14 +5657,24 @@ export class ContainerRuntime
...
 			(event) => {
+				const pendingAttachmentBlobs = this.blobManager.getPendingBlobs();
+				if (!this.hasPendingMessages() && pendingAttachmentBlobs === undefined) {
+					// Sequenced saved ops can reconstruct the runtime state without carrying
+					// session-scoped state such as the IdCompressor's ongoing session.
+					event.end({
+						attachmentBlobsSize: 0,
+						pendingOpsSize: 0,
+					});
+					return undefined;
```
Abe27342, line 5668: note: after some discussion with Mark, we think the GC timestamp may want to be preserved.

(and captureFullContainerState gets that wrong today if so, as it drops all this pending runtime state)

unsure on blobs.
id-compressor is undesirable (main actual issue with reusability when there are no other messages).

Not clear if the container-runtime check here is right (hasPendingMessages) or if pending state manager needs to be consulted directly.

packages/common/container-definitions/src/loader.ts
```diff
@@ -584,8 +590,12 @@ export interface ILoader extends Partial<IProvideLoader> {
 	 * When `pendingLocalState` is provided, the resolved container is rehydrated from a blob
 	 * previously produced by {@link IContainer.getPendingLocalState}.
 	 *
-	 * WARNING: misuse of `pendingLocalState` can result in duplicate op submission and potential
-	 * document corruption. To prevent container forking, callers must:
+	 * Use `isPendingLocalStateReusable` from `@fluidframework/container-runtime/legacy` to
```
Abe27342, line 593: Maybe we want a new API rather than allowing this one to be usable in this case. `serializeSavedContainerState` or something like that?

is it safe to call even if there are local changes?

(would route through same code paths as current getPendingLocalState on IContainer but drop pending runtime state)

Abe27342: main scenario this is plausibly useful over captureFullContainerState (besides that API having some bugs today) is if someone actually did legitimately make edits (which were now committed) to a container they have open and they want to pick up where they left off in multiple parallelizable follow-ups

Abe27342:
closing in favor of other changes - captureFullContainerState serves needs we have in this realm now.

PR #27995, opened 2026-08-17
fix(tinylicious): bound blob upload concurrency

alexvy86:
## Description

Recurring Tinylicious test stalls showed many requests completing together after multi-second quiet periods. This initially looked like an event-loop, garbage-collection, or summary-conversion pause, so the server was instrumented locally for timer delay, GC, CPU, sockets, and the lifecycle of outbound HTTP requests.

The reproductions instead showed a responsive, often mostly idle event loop while thousands of internal blob requests waited seconds for response headers. Tinylicious runs Alfred and Historian in one process but retains their localhost HTTP boundary, and summary upload uses an unbounded `Promise.all` over sibling entries. Each unique blob therefore creates a separate Alfred-to-Historian HTTP request. Larger workloads degraded with total upload fanout: roughly 4,000 uploads created about 4,000 sockets and made the slowest document creation request take 9-17 seconds, while roughly 8,000 uploads increased that maximum latency to 31 seconds.

This change bounds that fanout at the Tinylicious Git manager boundary, before Historian creates the HTTP request. A process-wide FIFO limit admits at most 50 concurrent blob uploads across documents. The limit is Tinylicious-specific so shared and distributed Routerlicious deployments are unaffected, and capacity is released correctly when an upload rejects.

Controlled summary workloads showed the following before/after results:

Each workload concurrently created four documents whose initial summaries contained the indicated number of unique blobs. "Maximum document creation latency" is the end-to-end HTTP response time of the slowest of those four document creation requests.

| Workload | Maximum document creation latency, before | Maximum document creation latency, after | Peak sockets, before | Peak sockets, after |
| --- | ---: | ---: | ---: | ---: |
| 4 documents x 1,000 blobs each | 16.8 s | 6.1 s | ~4,012 | 107 |
| 4 documents x 2,000 blobs each | 31.4 s | 9.0 s | ~8,974 | 107 |

A full local Tinylicious e2e run completed with 4,030 passing tests and no failures.

## Reviewer Guidance

Tests cover the process-wide concurrency bound across independent manager instances and rejection recovery. Fifty was selected because limits of 100 and 200 reintroduced response backlog without a reliable throughput improvement.


server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
```diff
@@ -20,6 +21,60 @@ import type {
...
+			return await operation();
+		} finally {
+			const next = this.waiters.shift();
+			if (next === undefined) {
+				this.activeCount--;
+			} else {
+				next();
+			}
+		}
+	}
```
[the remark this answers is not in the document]
alexvy86: `activeCount` represents occupied or synchronously reserved permits, not only operations whose continuations have resumed. When `finally` shifts a waiter, intentionally leaving `activeCount` unchanged transfers the completing operation’s permit to that waiter before `next()` resolves it. Therefore, even if several resolved waiters have not resumed yet, each remains represented in `activeCount`; a later completion that finds no waiter decrements only its own permit. Decrementing and immediately incrementing around `next()` would leave the same accounting state.

I also exercised the proposed interleaving directly with a limit of 2: both active operations completed in the same synchronous turn, additional runs were submitted before any limiter/waiter continuation resumed, and the observed maximum remained 2 (`activeCount` returned to 0 with no waiters). The existing package test similarly observes exactly 50 across 300 uploads and independent manager instances.

ChumpChief (approved):
Exciting!

server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
```diff
@@ -20,6 +21,60 @@ import type {
...
+/**
+ * Runs asynchronous operations with a fixed concurrency limit.
+ */
+class PromiseLimiter {
+	private activeCount = 0;
+	private readonly waiters: (() => void)[] = [];
+
+	public constructor(private readonly limit: number) {}
+
+	public async run<T>(operation: () => Promise<T>): Promise<T> {
```
markfields: Unit test would be great. Even in a follow-up PR to get the mitigation in ASAP. 

Also it's possible this utility exists already elsewhere in the codebase, I think I've seen it before.

markfields (approved):
Ship it!

server/routerlicious/.changeset/tinylicious-blob-upload-limit.md
```diff
@@ -0,0 +1,8 @@
+---
+"tinylicious": minor
+"__section": fix
+---
+
+Bound Tinylicious blob upload concurrency
```
CraigMacomber, line 6: This `.changeset` directory is for client: as this is not a client change it does not belong here. If you put it here, it will be listed as a change in the 3.0 release of client.

server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
```diff
@@ -20,6 +21,60 @@ import type {
...
+ * sockets at once, delaying all requests handled by the process.
+ *
+ * This limit is shared across Tinylicious tenants so concurrent documents use one request budget.
+ */
+const maxConcurrentBlobUploads = 50;
+
+/**
+ * Runs asynchronous operations with a fixed concurrency limit.
+ */
+class PromiseLimiter {
```
CraigMacomber: this quite a common thing to do. I know our flub does it in flub-exec, pnpm has an implementation and I suspect so does fluid build. I wouldn't be surprised if we already have an implementation of this logic in our dependency tree somewhere we can just use

CraigMacomber: flub exec used a third party library for this IIRC: not sure what the others do.

server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
alexvy86: Addressed in 9fd67a803d. The PR includes unit coverage in `tenantManager.spec.ts` for the process-wide concurrency cap across manager instances and for queued work continuing after an upload rejects. I also replaced the custom limiter with the server release group’s existing third-party approach, `async.queue`.

server/routerlicious/.changeset/tinylicious-blob-upload-limit.md
alexvy86: Fixed in 9fd67a803d: moved the changeset to `server/routerlicious/.changeset/` so it participates in the server release group rather than the client release.

server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
alexvy86: Addressed in 9fd67a803d. Tinylicious now depends directly on `async` and uses a process-wide `async.queue` with concurrency 50, matching the approach already used by other server packages instead of maintaining a custom limiter.

server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
```diff
@@ -18,8 +19,37 @@ import type {
...
+ * sockets at once, delaying all requests handled by the process.
+ *
+ * This limit is shared across Tinylicious tenants so concurrent documents use one request budget.
+ */
+const maxConcurrentBlobUploads = 50;
+
+const blobUploadQueue = queue(
+	async (upload: () => Promise<ICreateBlobResponse>) => upload(),
+	maxConcurrentBlobUploads,
+);
```
markfields: Nit: I just spent 5 minutes trying to parse this syntax, learning about queue and AsyncWorker in TypeScript playground.  Here on the other side, I might suggest making this a little bit more self-documenting and approachable, for instance:

```suggestion
const blobUploadQueue = queue<() => Promise<ICreateBlobResponse>>(
	async upload => upload(), // We'll push "upload" callbacks onto the queue, which will simply invoke them
	maxConcurrentBlobUploads,
);
```

markfields (approved):

server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
alexvy86: Didn't do exactly this but something to the same effect.

server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
```diff
@@ -29,39 +31,49 @@ import { default as Axios } from "axios";
...
+				queueLength,
+				queueLengthThreshold: threshold,
+				maxConcurrentBlobUploads,
+			});
 		}
 	}
 }
 
-const blobUploadLimiter = new PromiseLimiter(maxConcurrentBlobUploads);
+blobUploadQueue.saturated(recordBlobUploadQueueLength);
```
ChumpChief: I'm not familiar with the `async` library, but doesn't this mean recordBlobUploadQueueLength will always be called just once when blobUploadQueue.length is 50 (meaning it won't see when it hits one of the thresholds)?  I see that it also gets called in createBlob and looks effective there, but does this invocation do anything interesting?

server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
```diff
@@ -71,7 +83,13 @@ export class TinyliciousGitManager extends GitManager {
 		content: string,
 		encoding: "utf-8" | "base64",
 	): Promise<ICreateBlobResponse> {
-		return blobUploadLimiter.run(async () => super.createBlob(content, encoding));
+		const upload = blobUploadQueue.pushAsync<ICreateBlobResponse>(async () =>
+			super.createBlob(content, encoding),
+		);
+		if (blobUploadQueue.running() === maxConcurrentBlobUploads) {
```
ChumpChief: Nit - might be equivalent, but seems like it might be more direct to base the decision to record queue length on the queue length.

```suggestion
		if (blobUploadQueue.length() > 0) {
```

server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
alexvy86: Good question. The `saturated` callback covered synchronous bursts because `async.queue` defers worker startup, so every `createBlob` call could initially observe `running() === 0`; the callback later observed the backlog after 50 workers started. Refactored in 1165ce2bed to calculate the capacity-adjusted waiting backlog on every submission, so the separate `saturated` callback is no longer needed.

server/routerlicious/packages/tinylicious/src/services/tenantManager.ts
alexvy86: Addressed in 1165ce2bed. A raw `length() > 0` check can overcount during a synchronous burst because queued tasks may still have worker capacity available. The code now subtracts available worker slots from `length()` and records that effective backlog on every submission. The test also verifies that 149 submissions remain below the 100-item threshold and the 150th crosses it.

markfields (approved):

ChumpChief (approved):

CraigMacomber (approved):
Changeset looks goof for docs

PR #27996, opened 2026-08-18
feat(tree): add schema change telemetry

Abe27342:
## Description

Adds a `SchemaChangeApplied` telemetry event whenever SharedTree applies an effective schema transition. The event classifies initialization, upgrade, restriction, and rollback transitions and records privacy-safe structural metrics without logging schema payloads or identifiers.

All computed event properties are tagged as code artifacts. Tests cover required-root initialization, which applies an intermediate schema, and a compatible schema expansion.

## Reviewer Guidance

Please review the event shape and the decision to emit once per applied schema inner-change, including intermediate initialization transitions.

### DIFF
--- a/packages/dds/tree/src/shared-tree/treeCheckout.ts
+++ b/packages/dds/tree/src/shared-tree/treeCheckout.ts
@@ -136,8 +136,6 @@ import type { ISharedTreeEditor, SharedTreeEditBuilder } from "./sharedTreeEditB
 import { extractTransactionChangeProcessor } from "./transactionPostProcessor.js";
 import { SerializedChange } from "./serializedChange.js";
 
-const schemaChangeTelemetryEventName = "SchemaChangeApplied";
-
 /**
  * Yields all defined (non-`undefined`) labels from a {@link LabelTree}, depth-first.
  */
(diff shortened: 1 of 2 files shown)


packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1783,6 +1805,54 @@ export class TreeCheckout implements ITreeCheckout {
...
+
+	let removedNodeSchemaCount = 0;
+	for (const identifier of oldSchema.nodeSchema.keys()) {
+		if (!newSchema.nodeSchema.has(identifier)) {
+			removedNodeSchemaCount++;
+		}
+	}
+
+	return {
+		changeKind: isInverse
```
agarwal-navin, line 1840: Consider adding these properties in a "details" object. These will likely end up in "Payload" anyway because we reached the max number of columns in our tables (at least we had some time back).
Adding to "details" will also make it easier to look for in telemetry in one place.

agarwal-navin (approved):

agarwal-navin (approved):

agarwal-navin (approved):

PR #28002, opened 2026-08-18
bump fluid client to 3.0.0

dannimad:
pnpm -r --include-workspace-root exec npm pkg set version=3.0.0
pnpm -r run build:genver
pnpm run --filter=@fluid-private/test-version-utils update-compat-versions


packages/test/test-version-utils/compat-workspaces/generated-versions.cjs
```diff
@@ -9,23 +9,22 @@
...
-		"2.53.1",
-		"2.43.0",
 		"2.40.0",
-		"2.33.2",
-		"2.23.0",
-		"2.13.0",
-		"2.5.0",
 		"2.0.9",
 		"2.0.0-rc.5.0.8",
+		"2.0.0-rc.4.0.10",
```
CraigMacomber, line 16: I don't think we need to add testing with 10 old internal releases as part of 3.0. Something is clearly screwy with the version selection for these tests.

Looks like we removed testing against lots of non-internal versions to test against internal and RC versions instead? Seems wrong.

PR #28004, opened 2026-08-18
Expose a couple string utils

CraigMacomber:
## Description

These are somewhat questionable helper methods that we have some interest in using experimentally in some applications for now until we can think of a better approach (or decide they make sense and stabilize them more).

They are being exposed as alpha instead of internal to help with this.

## Reviewer Guidance


packages/dds/tree/src/text/codePointUtils.ts
```diff
@@ -51,31 +61,32 @@ export function codePointCount(value: string): number {
  * @param start - The UTF-16 index in `value` to start measuring from. Must be in `[0, value.length]`.
  * @param count - The number of Unicode code points to measure. Must be non-negative, and there must
  * be at least `count` code points available in `value` starting at `start`.
+ * @returns The number of UTF-16 code units occupied by the requested code points.
  * @throws A {@link @fluidframework/telemetry-utils#UsageError} if `start` is out of range,
  * `count` is negative, or fewer than `count` code points are available from `start`.
```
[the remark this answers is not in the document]
Josmithr: This is incorrect in terms of how API-Extractor resolves links. The existing syntax is correct.

.vscode/settings.json
```diff
@@ -6,7 +6,6 @@
 	],
 	"js/ts.tsdk.path": "node_modules/typescript/lib",
 	"typescript.tsdk": "node_modules/typescript/lib",
-
```
Josmithr, line 9: Nit: These line formatting changes are probably okay, but don't seem related to your changes. Probably worth reverting.

packages/dds/tree/src/test/text/codePointUtils.spec.ts
```diff
@@ -5,11 +5,48 @@
...
 
-// Allow importing file being tested
-// eslint-disable-next-line import-x/no-internal-modules
+// eslint-disable-next-line import-x/no-internal-modules -- Import the implementations being tested.
 import { codePointCount, utf16LengthForCodePoints } from "../../text/codePointUtils.js";
+import { validateIndex } from "../../util/index.js";
+
+/**
+ * Simpler but slightly slower implementation of utf16LengthForCodePoints using the string iterator.
+ * Used as a reference implementation.
```
Josmithr: Nit
```suggestion
 * Simpler but slightly slower implementation of {@link utf16LengthForCodePoints} using the string iterator.
 * @remarks Used as a reference implementation.
```

packages/dds/tree/src/test/text/codePointUtils.spec.ts
```diff
@@ -5,11 +5,48 @@
...
+	}
+	if (counted !== count) {
+		throw new UsageError(
+			`count (${count}) exceeds the ${counted} code point(s) available from start (${start}) in a value of length ${value.length}.`,
+		);
+	}
+	return utf16;
+}
+
+const implementations = [
```
Josmithr, line 50: Nit: a quick comment explaining why we're testing 2 separate implementations here would be useful. It's hinted at in the docs for `utf16LengthForCodePointsUsingIterator`, but an explicit note here or above the describe block below would be useful.

Josmithr (commented):
Left a few suggestions. Overall looks good.

.vscode/settings.json
CraigMacomber: I really dislike that the automated tools for updating these files violate our formatting rules, and after formatting we get these diffs.

Josmithr (approved):

PR #28005, opened 2026-08-18
fix(external-data): fix flaky webhook relay tests using fixed delays

alexvy86:
## Description

Fixes intermittent CI failures in `examples/external-data` tests, e.g. `register-session-url: Complete data flow`, `register-for-webhook: Complete data flow`/`register-for-webhook` (seen in both public PR builds and internal ADO CI — e.g. build 418109 / build 418015).

Root cause: `MockWebhook.notifySubscribers` and `echoExternalDataWebhookToFluid` relay webhook notifications as fire-and-forget (`fetch(...).catch(...)`, unawaited). Several tests waited a fixed `delay(1000)` before asserting the notification arrived — under CI load/slowness this isn't always enough time, causing races (`TypeError: Cannot read properties of undefined (reading 'signalContent')`, or `wasFluidNotifiedForChange`/`wasHookNotifiedForChange` assertion failures).

Fix: added a `waitForCondition(condition, timeoutMs, pollIntervalMs)` polling helper in `src/test/utilities.ts` and replaced the fixed `delay(1000)` + assert pattern with it in the three affected tests (`customerService.test.ts`: `register-for-webhook`, `register-session-url`; `externalDataService.test.ts`: `register-for-webhook`). Also bumped `this.timeout(...)` on these tests since `waitForCondition`'s internal timeout budget can exceed mocha's default per-test timeout.

**Repro/verification:** Reproduced the failures reliably by temporarily injecting an artificial delay (via a local-only env var) into both fire-and-forget relay hops. At 2x the original fixed wait (2000ms), the original code failed consistently with the same errors seen in CI; with this fix applied, 5 consecutive runs at the same simulated delay all passed with zero failures. A normal (no injected delay) run remains fast (~300ms total).

## Reviewer Guidance

### DIFF
--- a/examples/external-data/src/test/customerService.test.ts
+++ b/examples/external-data/src/test/customerService.test.ts
@@ -18,7 +18,7 @@ import {
 import { externalDataServicePort } from "../mock-external-data-service-interface/index.js";
 import type { ITaskData } from "../model-interface/index.js";
 
-import { closeServer, delay } from "./utilities.js";
+import { closeServer, delay, waitForCondition } from "./utilities.js";
 
 const localServicePort = 5002;
 const externalTaskListId = "task-list-1";
@@ -175,7 +175,10 @@ describe("mock-customer-service", () => {
 
 	// We have omitted `@types/supertest` due to cross-package build issue.
 	// So for these tests we have to live with `any`.
-	it("register-for-webhook: Complete data flow", async () => {
+	it("register-for-webhook: Complete data flow", async function () {
+		// waitForCondition's internal timeout can exceed mocha's default per-test timeout;
+		// give this test enough headroom to let waitForCondition fail on its own terms.
+		this.timeout(10000);
 		// Set up mock local service, which will be registered as webhook listener
 		const localServiceApp = initializeMockFluidService(express());
 		const tenantId = "tinylicious";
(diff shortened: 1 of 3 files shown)


examples/external-data/src/test/utilities.ts
```diff
@@ -27,3 +27,32 @@ export async function closeServer(server: Server): Promise<void> {
...
+	timeoutMs: number = 5000,
+	pollIntervalMs: number = 50,
+): Promise<void> => {
+	const startTime = Date.now();
+	while (!condition()) {
+		if (Date.now() - startTime >= timeoutMs) {
+			throw new Error(`Condition was not satisfied within ${timeoutMs}ms timeout.`);
+		}
+		await delay(pollIntervalMs);
+	}
```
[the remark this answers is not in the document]
alexvy86: Addressed in latest.

jenn-le (approved):

examples/external-data/src/test/customerService.test.ts
```diff
@@ -229,7 +234,10 @@ describe("mock-customer-service", () => {
 	});
 	/* eslint-enable @typescript-eslint/no-non-null-assertion */
 
-	it("register-session-url: Complete data flow", async () => {
+	it("register-session-url: Complete data flow", async function () {
+		// waitForCondition's internal timeout can exceed mocha's default per-test timeout;
+		// give this test enough headroom to let waitForCondition fail on its own terms.
+		this.timeout(10000);
```
Josmithr, line 240: Should we move this to the suite level so we don't have to do this in each test?

Josmithr (approved):

examples/external-data/src/test/customerService.test.ts
alexvy86: Considered this — in `customerService.test.ts` all 3 tests in the suite do need it, so a suite-level `this.timeout(...)` would work there. But `externalDataService.test.ts`'s webhook `describe` only has this one test, so it'd be equivalent to setting it inline. Since the need for the longer timeout isn't uniform across every suite (and isn't needed at all for the other 12 tests in `externalDataService.test.ts`'s main describe block), I opted to keep it local to each affected test — makes it clearer *why* that specific test needs more time (the fire-and-forget webhook relay race) rather than being a suite-wide default that new fast tests would inherit without explanation. Happy to revisit if you feel strongly.

alexvy86: I'm partial to keeping it with each test that actually needs it, I'll merge like this.

PR #28006, opened 2026-08-18
feat(container-loader): add reference-only state capture

alexvy86:
## Description

This PR adds a reference-only mode for `captureFullContainerState` that results in blobs being omitted in the returned pending container state. The new `blobCaptureMode` option defaults to `"inline"` to preserve today's self-contained/offline behavior; `"reference"` retains snapshot blob IDs while omitting structural and attachment payloads so later online loads fetch them from live storage.

The intended scenario is capturing a baseline state that is later used together with a separate (non-Fluid) representation of changes to be applied to the container, so the edits are applied on top of the container state at the moment of the initial capture/read, not on top of the latest container state. The state is returned by an HTTP API call and is submitted back on another HTTP request, where the caller doesn't interact with the state blob, so keeping it size down is preferred to save bandwidth and is possible because the load during the write request happens online and blobs can be rehydrated at that point.

Of note, `getPendingLocalState` is intentionally not suitable for this even when it doesn't include blobs by default because it serializes an ongoing ID-compressor session and pending runtime state; forking that artifact can duplicate session ownership, and we don't want to limit the use of the baseline state blob to a single write request after it is read. `captureFullContainerState` emits `pendingRuntimeState: undefined`, so each load creates a fresh runtime and ID-compressor session.

Reference-only state is online-only. `loadExistingContainer` and the online form of `loadFrozenContainerFromPendingState` fall through to live `readBlob` when the payload cache is empty. The fully offline frozen-load form rejects reference-only state with a clear `UsageError`.

Validation includes:
- Existing default inline capture and fully offline frozen-load coverage.
- Structural and attachment payload omission plus state-size reduction.
- Live attachment reads from omitted payloads.
- Two independent containers loaded from one baseline, each submitting a distinct edit, with both edits present in a subsequent load.
- 300 passing container-loader tests.
- 108 passing local-server tests (4 pending).

This is a request-for-feedback prototype, not a proposal to make `getPendingLocalState` forkable.

## Reviewer Guidance

Please focus on:
- Is `blobCaptureMode: "inline" | "reference"` the right API shape and terminology?
- Should reference-only artifacts carry an explicit serialized marker, rather than being identified by an empty `snapshotBlobs` map for fully offline rejection?
- Is making the existing capture API mode-dependent preferable to introducing a separate reference-only capture API?
- Are the online-only storage dependency and offline failure semantics sufficiently explicit?

This adds a non-breaking property to an existing `@legacy @alpha` interface, so API Council review is required.


packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -418,6 +419,11 @@ export async function loadFrozenContainerFromPendingState(
 		// returns a resolved URL whose `url` equals `pendingLocalState.url`, so the
 		// identity guard in `Loader.resolveCore` is trivially satisfied.
 		const pending = getAttachedContainerStateFromSerializedContainer(pendingLocalState);
+		if (Object.keys(pending.snapshotBlobs).length === 0) {
+			throw new UsageError(
+				`${fnName}: pending state does not contain inlined snapshot blob contents. Reference-only state captured by ${captureFullContainerState.name} requires online driver wiring.`,
+			);
```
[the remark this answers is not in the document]
alexvy86: Addressed in dd932effe96. Reference-only captures now carry an explicit internal `blobContentsMode: "reference"` marker, and the fully offline path checks that marker instead of inferring the mode from an empty `snapshotBlobs` map. Existing/default inline state remains unmarked. I also added coverage showing that an unmarked empty map is not misclassified and falls through to the normal missing-blob error.

packages/test/local-server-tests/src/test/captureFullContainerState.spec.ts
```diff
@@ -510,6 +512,117 @@ describe("captureFullContainerState", () => {
...
+			blobCaptureMode: "reference",
+		});
+		const parsedReferenceState = JSON.parse(referenceState) as {
+			snapshotBlobs: Record<string, string>;
+			attachmentBlobContents?: Record<string, string>;
+		};
+		assert.deepStrictEqual(
+			parsedReferenceState.snapshotBlobs,
+			{},
+			"Reference-only state should preserve the snapshot tree without structural payloads",
```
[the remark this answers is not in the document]
alexvy86: Fixed in dd932effe96. The assertion now says that reference-only state omits structural blob payloads, and a separate assertion verifies that the snapshot carries the explicit live-storage marker.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -505,6 +511,19 @@ export interface ICaptureFullContainerStateProps {
 	 * Optional logger for driver-side telemetry.
 	 */
 	readonly logger?: ITelemetryBaseLogger | undefined;
+	/**
+	 * Controls whether blob payloads are included in the captured state.
+	 *
+	 * `"inline"` (the default) includes referenced structural and attachment
+	 * blob contents, producing self-contained state that supports fully offline
+	 * loading with {@link loadFrozenContainerFromPendingState}.
```
Abe27342: I don't think we get any value here specifying different types of blobs.



```suggestion
	 * `"inline"` (the default) includes blob contents, producing self-contained
	 * state that supports fully offline loading with {@link loadFrozenContainerFromPendingState}.
```

Abe27342 (approved):

Abe27342 (approved):

Abe27342 (approved):

Abe27342 (approved):

.changeset/strict-buckets-sing.md
```diff
@@ -0,0 +1,23 @@
+---
+"@fluidframework/container-loader": minor
+"__section": feature
+---
+
```
Josmithr:
```suggestion
```

.changeset/strict-buckets-sing.md
```diff
@@ -0,0 +1,23 @@
...
+Add reference-only capture for reusable container baselines
+
+`captureFullContainerState` now accepts a `blobCaptureMode` option. The default `"inline"` mode remains self-contained, while `"reference"` omits structural and attachment blob payloads so repeated online loads can fetch them from live storage.
```
Josmithr: Nit: link to API docs

.changeset/strict-buckets-sing.md
alexvy86: Applied the suggestion and removed the extra blank line after the front matter.

.changeset/strict-buckets-sing.md
alexvy86: Linked `captureFullContainerState` to its API docs. I used the package-page function anchor (`/docs/api/container-loader/#capturefullcontainerstate-function`), which is the canonical live link for this legacy export; the otherwise conventional standalone `capturefullcontainerstate-function` page currently returns 404.

jzaffiro (approved):
Approved for docs

PR #28008, opened 2026-08-18
fix(tree): forking during changed event

yann-achard-MS:
## Description

[Forking](https://fluidframework.com/docs/api/fluid-framework/treeviewbeta-interface#fork-methodsignature) (beta) a view during the callback for the ["changed" event](https://fluidframework.com/docs/api/fluid-framework/treebranchevents-interface#changed-methodsignature) (alpha) emitted when a transaction is committed would create a fork with malformed change data.
This could result in asserts being triggered when utilizing the fork (including, but not limited to, error code `0x7ce`).

This PR also changes usages of `activeBranch` to `branch` wherever appropriate in `treeCheckout.ts`. I have added a PR comment explaining my reasoning for each change.

## Breaking Changes

None

### DIFF
--- a/packages/dds/tree/src/shared-tree/treeCheckout.ts
+++ b/packages/dds/tree/src/shared-tree/treeCheckout.ts
@@ -850,7 +850,7 @@ export class TreeCheckout implements ITreeCheckout {
 							);
 							this.revertibleCommitBranches.set(
 								revision,
-								this.#transaction.activeBranch.fork(commit),
+								this.#transaction.branch.fork(commit),
 							);
 							this.revertibles.set(revision, revertible);
 							return revertible;
@@ -1207,9 +1207,9 @@ export class TreeCheckout implements ITreeCheckout {
 				}
 
 				const commitToRevert = revertibleBranch.getHead();
-				const activeBranchHead = targetCheckout.#transaction.activeBranch.getHead();
+				const branchHead = targetCheckout.#transaction.branch.getHead();
 
-				if (isAncestor(commitToRevert, activeBranchHead, true) === false) {
+				if (isAncestor(commitToRevert, branchHead, true) === false) {
 					throw new UsageError(
 						"Cannot clone revertible for a commit that is not present on the given branch.",
 					);
@@ -1308,7 +1308,7 @@ export class TreeCheckout implements ITreeCheckout {
 			throw new UsageError("A view cannot be forked while it has a pending transaction.");
 		}
 
-		const branch = this.#transaction.activeBranch.fork();
+		const branch = this.#transaction.branch.fork();
 		const storedSchema = this.storedSchema.clone();
 		const forkBreaker = new Breakable("TreeCheckout", this.logger);
 		const forest = this.forest.clone(storedSchema, forkBreaker);
@@ -1381,7 +1381,7 @@ export class TreeCheckout implements ITreeCheckout {
 			0xa5d /* Shared branches cannot be rebased onto another branch. */,
 		);
 
-		checkout.#transaction.activeBranch.rebaseOnto(this.#transaction.activeBranch);
+		checkout.#transaction.branch.rebaseOnto(this.#transaction.branch);
 	}
 
 	public rebaseOnto(branch: TreeBranch): void {
@@ -1446,7 +1446,7 @@ export class TreeCheckout implements ITreeCheckout {
 				"Views with an open transaction cannot be merged into another view.",
 			);
 		}
-		this.#transaction.activeBranch.merge(checkout.#transaction.activeBranch);
+		this.#transaction.branch.merge(checkout.#transaction.branch);
 		if (disposeMerged && !checkout.isSharedBranch) {
 			// Dispose the merged checkout unless it is a shared branch.
 			checkout[disposeSymbol]();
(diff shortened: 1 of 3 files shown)


packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -850,7 +850,7 @@ export class TreeCheckout implements ITreeCheckout {
 							);
 							this.revertibleCommitBranches.set(
 								revision,
-								this.#transaction.activeBranch.fork(commit),
+								this.#transaction.branch.fork(commit),
```
yann-achard-MS, line 853: When invoked during the callback of a "changed" event that is triggered by a committing transaction, this would call fork on the transaction branch. This was wrong (we should be forking the main branch) but had no negative impact.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1207,9 +1207,9 @@ export class TreeCheckout implements ITreeCheckout {
 				}
 
 				const commitToRevert = revertibleBranch.getHead();
-				const activeBranchHead = targetCheckout.#transaction.activeBranch.getHead();
+				const branchHead = targetCheckout.#transaction.branch.getHead();
```
yann-achard-MS, line 1211: It shouldn't make a difference which branch we get the head from (because the commit to revert should be upstream of whatever might be on a transaction branch), but it makes more sense to use the main (i.e., non-transaction) branch since that's the one that the inverse of the commit to be reverted will be rebased onto and so that's the branch that needs to be related to that commit.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1308,7 +1308,7 @@ export class TreeCheckout implements ITreeCheckout {
 			throw new UsageError("A view cannot be forked while it has a pending transaction.");
 		}
 
-		const branch = this.#transaction.activeBranch.fork();
+		const branch = this.#transaction.branch.fork();
```
yann-achard-MS, line 1311: >branch [](http://example.com/codeflow?start=35&length=6)

This is where the bug fix is. The above checks for `this.transaction.size > 0` pass when `fork` is invoked from after the transaction is committed, but `this.#transaction.activeBranch` still returns the transaction branch during the "changed" event callback.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1381,7 +1381,7 @@ export class TreeCheckout implements ITreeCheckout {
 			0xa5d /* Shared branches cannot be rebased onto another branch. */,
 		);
 
-		checkout.#transaction.activeBranch.rebaseOnto(this.#transaction.activeBranch);
+		checkout.#transaction.branch.rebaseOnto(this.#transaction.branch);
 	}
```
yann-achard-MS, line 1385: Since we don't intend to support either of these `activeBranch`_es being the transaction branch, we may as well not tempt fate.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1446,7 +1446,7 @@ export class TreeCheckout implements ITreeCheckout {
 				"Views with an open transaction cannot be merged into another view.",
 			);
 		}
-		this.#transaction.activeBranch.merge(checkout.#transaction.activeBranch);
+		this.#transaction.branch.merge(checkout.#transaction.branch);
 		if (disposeMerged && !checkout.isSharedBranch) {
```
yann-achard-MS, line 1450: Same as for rebase

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1545,7 +1545,7 @@ export class TreeCheckout implements ITreeCheckout {
 			revisionForInvert,
 		);
 
-		const headCommit = this.#transaction.activeBranch.getHead();
+		const headCommit = this.#transaction.branch.getHead();
 		// Rebase the inverted change onto any commits that occurred after the undoable commits.
```
yann-achard-MS, line 1549: There might have been a bug here as well except we don't let you revert edits during an event callback. Still, it seems bad to use `activeBranch` when only `branch` makes sense.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1578,7 +1578,7 @@ export class TreeCheckout implements ITreeCheckout {
 		const previousLabelTreeNode = this.labelTreeNode;
 		this.labelTreeNode = labelTree;
 		try {
-			this.#transaction.activeBranch.apply(
+			this.#transaction.branch.apply(
 				change,
```
yann-achard-MS, line 1582: Same as for the `headCommit` above.

packages/dds/tree/src/shared-tree/treeCheckout.ts
noencke: This implies that users can't Undo during a transaction, correct? I can't remember if that's the case off the top of my head but surely it must be disallowed.

noencke (approved):

packages/dds/tree/src/shared-tree/treeCheckout.ts
yann-achard-MS: I don't think that follows from the above. What does follow is that you can't undo transaction steps. We want to fork the main branch because that is where the `commit` is. All that said, you are correct that we don't currently allow undo during transactions (there is a check against it at the top of `revertRevertible`). If we start allowing that, we'll still want to fork form the main branch here.

Josmithr (approved):

PR #20396, opened 2024-03-29
Repro offline initial sequence number mismatch

tyler-cai-microsoft:
[AB#7667](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/7667)

This can be seen working in this PR: https://github.com/microsoft/FluidFramework/pull/20342

In the work to do [AB#7588](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/7588), I found that the core assumptions I had about `DeltaManager.initialSequenceNumber` in the runtime layer was wrong. My understanding was that `DeltaManager.initialSequenceNumber` pointed to the sequence number of the base snapshot. This was broken in an offline scenario that wanted to prevent the `DeltaManager` from processing extra ops. This PR fixes that so that the `DeltaManager` still skips processing extra ops and has its initial sequence number point to the sequence number of the snapshot the container loaded from.

PR #28009, opened 2026-08-18
fix(container-loader): Include websocket ops in captureFullContainerState

Abe27342:
## Description

`captureFullContainerState` previously retrieved operations only from delta storage. On services such as ODSP, recently sequenced operations may still be available through the delta stream but not yet persisted to storage, causing the captured state to omit recent changes.

This change reuses the container load path's `DeltaManager` and `ConnectionManager` construction to merge operations from storage and the live delta stream. Capture waits through the latest sequence known by the connection, then disposes the delta manager to freeze the operation set before serializing the artifact.

The added coverage includes a deterministic socket-only operation test and a real-service frozen offline round-trip test. Against ODSP, the new test passes with this change and fails against the previous implementation because the captured artifact contains no post-snapshot operations.

## Reviewer Guidance

Please focus on the synchronization and cutoff semantics in `captureFullContainerState`, particularly whether disposing the delta manager immediately after catch-up provides the intended stable capture boundary.

### DIFF
--- a/packages/loader/container-loader/src/createAndLoadContainerUtils.ts
+++ b/packages/loader/container-loader/src/createAndLoadContainerUtils.ts
@@ -545,6 +545,7 @@ export interface ICaptureFullContainerStateProps {
  * is a known consumer and end-to-end coverage, the capture refuses rather
  * than silently producing pending state that omits group data.
  *
+ * @privateRemarks
  * Note: if a new snapshot lands between the snapshot fetch and delta catch-up,
  * the returned state may not reflect the very latest snapshot, but remains
  * internally consistent: ops are anchored to the snapshot that was captured.
--- a/packages/loader/container-loader/src/createAndLoadContainerUtils.ts
+++ b/packages/loader/container-loader/src/createAndLoadContainerUtils.ts
@@ -617,6 +617,10 @@ export async function captureFullContainerState({
 	deltaManager.connect({
 		reason: { text: "captureFullContainerState" },
 		mode: "read",
+		// Ops will be fetched from storage using the later "attachOpHandler" call. Attempting to fetch here is a no-op anyway as
+		// no op handler is yet attached.
+		// Fetching the snapshot first such that an op handler could be attached using its sequence number information may simplify
+		// the code, but doing it this way allows parallelization of snapshot fetch and establishing a connection.
 		fetchOpsFromStorage: false,
 	});
 	try {
--- a/packages/loader/container-loader/src/createAndLoadContainerUtils.ts
+++ b/packages/loader/container-loader/src/createAndLoadContainerUtils.ts
@@ -18,7 +18,7 @@ import type {
 } from "@fluidframework/core-interfaces";
 import type { AllOrNone } from "@fluidframework/core-interfaces/internal";
 import { validateAllOrNone } from "@fluidframework/core-utils/internal";
-import type { IClientDetails } from "@fluidframework/driver-definitions";
+import type { IClient, IClientDetails } from "@fluidframework/driver-definitions";
 import type {
 	IDocumentServiceFactory,
 	IResolvedUrl,
@@ -32,6 +32,7 @@ import { getSnapshotTree } from "@fluidframework/driver-utils/internal";
 import {
 	GenericError,
 	UsageError,
+	createChildLogger,
 	normalizeError,
 	createChildMonitoringContext,
 	mixinMonitoringContext,
@@ -51,7 +52,9 @@ import {
 	unreferencedAttachmentBlobLocalIds,
 	type IBlobAttachReference,
 } from "./captureReferencedContents.js";
+import { CatchUpMonitor } from "./catchUpMonitor.js";
 import { DebugLogger } from "./debugLogger.js";
+import { createDeltaManager } from "./deltaManagerFactory.js";
 import { createFrozenDocumentServiceFactory } from "./frozenServices.js";
 import { Loader } from "./loader.js";
 import { pkgVersion } from "./packageVersion.js";
@@ -518,7 +521,10 @@ export interface ICaptureFullContainerStateProps {
  * the latest snapshot, inlined contents of every blob reachable through
  * referenced subtrees, inlined contents of every referenced attachment blob
  * keyed by storage id, and all ops with sequence numbers after the base
- * snapshot's sequence number (as read from its attributes blob).
+ * snapshot's sequence number through the point observed when the delta stream
+ * connects. The same DeltaManager catch-up path used by container loading
+ * combines delta-storage and delta-stream ops, so recently sequenced ops do
+ * not need to have reached durable delta storage before capture.
  *
  * Reachability respects GC. Snapshot subtrees flagged `unreferenced: true`
  * are skipped (their contents are not inlined). Attachment blobs that GC has
@@ -539,7 +545,7 @@ export interface ICaptureFullContainerStateProps {
  * is a known consumer and end-to-end coverage, the capture refuses rather
  * than silently producing pending state that omits group data.
  *
- * Note: if a new snapshot lands between the snapshot fetch and the ops fetch,
+ * Note: if a new snapshot lands between the snapshot fetch and delta catch-up,
  * the returned state may not reflect the very latest snapshot, but remains
  * internally consistent: ops are anchored to the snapshot that was captured.
  *
@@ -579,6 +585,39 @@ export async function captureFullContainerState({
 		resolvedUrl,
 		logger,
 	);
+	const captureLogger = createChildLogger({ logger, namespace: "captureFullContainerState" });
+	const client: IClient = {
+		details: { capabilities: { interactive: true } },
+		mode: "read",
+		permission: [],
+		scopes: [],
+		user: { id: "" },
+	};
+	const deltaManager = createDeltaManager({
+		serviceProvider: () => documentService,
+		logger: captureLogger,
+		active: () => false,
+		containerDirty: () => false,
+		client,
+		reconnectAllowed: true,
+	});
+	const connectedP = new Promise<void>((resolve) => {
+		deltaManager.once("connect", () => resolve());
+	});
+	const closedP = new Promise<never>((_resolve, reject) => {
+		deltaManager.once("closed", (error) => {
+			reject(
+				error === undefined
+					? new GenericError("DeltaManager closed while capturing container state")
+					: normalizeError(error),
+			);
+		});
+	});
+	deltaManager.connect({
+		reason: { text: "captureFullContainerState" },
+		mode: "read",
+		fetchOpsFromStorage: false,
+	});
 	try {
 		const storage = await documentService.connectToStorage();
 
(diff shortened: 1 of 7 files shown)


anthony-murphy:

- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -539,7 +545,7 @@ export interface ICaptureFullContainerStateProps {
  * is a known consumer and end-to-end coverage, the capture refuses rather
  * than silently producing pending state that omits group data.
  *
- * Note: if a new snapshot lands between the snapshot fetch and the ops fetch,
+ * Note: if a new snapshot lands between the snapshot fetch and delta catch-up,
```
Abe27342, line 549: this paragraph and what follows is better suited for a \@privateRemarks block.

packages/test/local-server-tests/src/test/captureFullContainerState.spec.ts
```diff
@@ -84,6 +85,76 @@ function makeFactoryWithFailingReadBlob(
 	};
 }
 
+function makeFactoryWithSocketOnlyOps(
```
alexvy86, line 95: I don't fully follow how this helper factory works. `connectToDeltaStorage` is stubbed, but `connectToDeltaStream` seems to reach into delta storage for something? Some TSDoc could help.

alexvy86 (commented):
I don't have enough familiarity to judge the changes to `captureFullContainerState` itself, but the rest of the PR generally makes sense to me based on offline chats, with one small comment below.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -579,6 +586,43 @@ export async function captureFullContainerState({
...
+	const closedP = new Promise<never>((_resolve, reject) => {
+		deltaManager.once("closed", (error) => {
+			reject(
+				error === undefined
+					? new GenericError("DeltaManager closed while capturing container state")
+					: normalizeError(error),
+			);
+		});
+	});
+	deltaManager.connect({
```
anthony-murphy, line 617: is there any case where we wouldn't want the ops from push? or is it always ok the try for them?

anthony-murphy:

- 🪣 **Mechanical Review** — trimming ballast…
- ❄️ **Historical / Contextual Review** — as the cold settles in…
- 🗺️ **Solution / Problem Space Review** — venturing into uncharted water…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:

- 🛰️ **Mechanical Review** — calibrating the sonar…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:
## Deep Review

*Reviewed commit `03e790b` on 2026-08-19.*

Not ready for sign-off. The rework — merging live delta-stream ops with delta-storage ops by reusing the container load path via the new shared `createDeltaManager` factory — is the sanctioned resolution of vladsud's #27220 socket-ops ask. One contained correctness gap remains: the capture path opens an unbounded, unabortable delta-stream connection that can hang forever on a persistent connection failure (flagged inline).

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add the #27220 origin, vladsud's socket-ops thread, and the ADO work item to the PR description so the originating alpha-surface threads can be resolved

### Context for Reviewers

- This migrates capture from a storage-only op drain (`connectToDeltaStorage().fetchMessages`) to a load-path catch-up (`connect` → `attachOpHandler(..., "all")` → `waitForCatchUp` → dispose), resolving vladsud's #27220 ask that the storage-only prototype "does not have websocket connection … we would need ops from socket."
- The merge reuses the container load path via the new `createDeltaManager` factory rather than a manual sequence-keyed merge — `Container` and capture build an identical delta stack, avoiding the silent drift a reimplemented merge would incur the day `DeltaManager`'s reconciliation changes. Also realizes vladsud's #8423 swappable-connection-manager modularization.
- The `connect → waitForCatchUp → dispose` boundary is the core design change. Sequence-number anchoring here is historically error-prone (#20396's offline `initialSequenceNumber` fix; #7429's no-handler fetch guard), so the `fetchOpsFromStorage: false` → later-`attachOpHandler` ordering is deliberate.

- **Needs human judgment** — anthony-murphy's inline question "is there any case where we wouldn't want the ops from push?" ([discussion_r3817449550](https://github.com/microsoft/FluidFramework/pull/28009#discussion_r3817449550)) is a design/taste call on whether capture should always establish the stream connection. The compatibility half is answered: `ConnectionManager.connectCore` already short-circuits `policies?.storageOnly === true` via a synthesized `FrozenDeltaStream`, so storage-only/frozen services need no capture-specific branch. The "always try push ops" preference is still an owner call.
- **Needs human judgment** — cutoff-semantics sign-off on whether the immediate-dispose fence is correct across services. vladsud (connectivity/#8423/#7429), markfields (#27220 capture author), and tyler-cai-microsoft (#20396 offline seq-number fix) are the right reviewers.
- **Cannot be assessed by the pipeline** — the e2e ODSP frozen round-trip test depends on live-service timing; confirm it is not flaky and that the socket-only fence behaves against real ODSP checkpoint semantics.

</details>

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
```diff
@@ -579,6 +586,43 @@ export async function captureFullContainerState({
...
 	);
+	const captureLogger = createChildLogger({ logger, namespace: "captureFullContainerState" });
+	const client: IClient = {
+		details: { capabilities: { interactive: true } },
+		mode: "read",
+		permission: [],
+		scopes: [],
+		user: { id: "" },
+	};
+	const deltaManager = createDeltaManager({
```
anthony-murphy, line 597: Deep Review: The capture path can hang forever on a persistent connection failure. `createDeltaManager` is called with `reconnectAllowed: true` and no `maxInitialConnectionAttempts`. `ConnectionManager` maps `reconnectAllowed: true` → `ReconnectMode.Enabled`, and with `maxInitialConnectionAttempts` undefined the retry loop only stops on success — so connection establishment retries indefinitely.

Capture then blocks on `await Promise.race([Promise.all([connectedP, attachP]), closedP])` and `await Promise.race([waitForCatchUp(deltaManager), closedP])`. `connectedP` resolves only on `"connect"` and `closedP` rejects only on `"closed"`. If no connection can be established (offline, permissions, a read-only endpoint that never emits `connect`, transient outage), neither promise settles and the manager retries forever — `captureFullContainerState` never returns and never throws. The prior implementation completed purely from storage reads and required no live socket, so this is a new failure mode.

This is historically corroborated: #20793 added an `AbortSignal` to the sibling container-load path *because that process can hang*, and #27220 explicitly requested `abortSignal?: AbortSignal` on `ICaptureFullContainerStateProps` before `@alpha` graduation.

Recommendation: add `abortSignal?: AbortSignal` to `ICaptureFullContainerStateProps`, wire it into the connect/`waitForCatchUp` await sequence, and dispose the `DeltaManager` on abort. If unbounded reconnection isn't intended for capture, also pass a finite `maxInitialConnectionAttempts` (e.g. `1`) and/or an overall timeout so capture fails fast with a clear error instead of pending forever. This is the abort mechanism #27220 asked to land before graduation.

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
Abe27342: I'm not aware of one

packages/test/local-server-tests/src/test/captureFullContainerState.spec.ts
Abe27342: added a doc comment here.

The implementation might be a bit clearer if done without proxies, but will definitely be more verbose. I'm indifferent. Guessing the agent went with this approach as there's already another proxy implementation directly above.

anthony-murphy:

- 🌑 **Historical / Contextual Review** — leaving the photic zone behind…
- 🤖 **Solution / Problem Space Review** — wandering the abyssal plain…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
Abe27342: I don't think it makes sense to bundle that change into this PR. I'd rather do it in a dedicated PR with a more targeted changeset. Will do that as a follow-up.

anthony-murphy:
## Deep Review

*Reviewed commit `cd5b0ac` on 2026-08-20.*

Not ready for sign-off. The design — merging live delta-stream ops with delta-storage ops by reusing the container load path through the new shared `createDeltaManager` factory — is validated as the correct resolution of the #27220 socket-ops ask. One contained correctness gap remains: the capture path opens an unbounded, unabortable delta-stream connection that can hang forever on a persistent retriable connection failure (flagged inline). The author is deferring the fix to a dedicated follow-up PR, so an owner should sign off on whether shipping the interim hang mode is acceptable.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Add the #27220 origin, vladsud's socket-ops thread, and the ADO work item to the PR description so the originating alpha-surface threads can be resolved

### Context for Reviewers

- This migrates capture from a storage-only op drain (`connectToDeltaStorage().fetchMessages`) to a load-path catch-up (`connect` → `attachOpHandler(..., "all")` → `waitForCatchUp` → dispose), resolving vladsud's #27220 ask that the storage-only prototype "does not have websocket connection … we would need ops from socket."
- The merge reuses the container load path via the new `createDeltaManager` factory rather than a manual sequence-keyed merge — `Container` and capture build an identical delta stack, avoiding the silent drift a reimplemented merge would incur the day `DeltaManager`'s reconciliation changes. Also realizes vladsud's #8423 swappable-connection-manager modularization.
- The `connect → waitForCatchUp → dispose` boundary is the core design change. Sequence-number anchoring here is historically error-prone (#20396's offline `initialSequenceNumber` fix; #7429's no-handler fetch guard), so the `fetchOpsFromStorage: false` → later-`attachOpHandler` ordering is deliberate.

- **Needs human judgment** — is deferring the capture hang fix to a follow-up PR acceptable, or must this PR at least bound the connection before merge? The author has committed to a dedicated follow-up ([discussion_r3817919077](https://github.com/microsoft/FluidFramework/pull/28009#discussion_r3817919077)); an owner (markfields / vladsud) should decide whether shipping the new hang mode in the interim is tolerable.
- **Needs human judgment** — anthony-murphy's inline question "is there any case where we wouldn't want the ops from push?" ([discussion_r3817449550](https://github.com/microsoft/FluidFramework/pull/28009#discussion_r3817449550), author answered "not aware of one") is a design/taste call on whether capture should always establish the stream connection. The compatibility half is settled: `ConnectionManager.connectCore` short-circuits `policies?.storageOnly === true` via a synthesized `FrozenDeltaStream`, so storage-only/frozen services need no capture-specific branch.
- **Needs human judgment** — cutoff-semantics sign-off on whether the immediate-dispose fence is correct across services. vladsud (connectivity/#8423/#7429), markfields (#27220 capture author), and tyler-cai-microsoft (#20396 offline seq-number fix) are the right reviewers.
- **Cannot be assessed by the pipeline** — the e2e ODSP frozen round-trip test depends on live-service timing; confirm it is not flaky and that the socket-only fence behaves against real ODSP checkpoint semantics.

</details>

- `03e790b` 2026-08-19 · **5/10** — capture hang flagged inline; #27220 socket-ops resolution

</details>

packages/loader/container-loader/src/createAndLoadContainerUtils.ts
anthony-murphy: Deep Review: Thanks — a dedicated follow-up for the `abortSignal` API makes sense, and #27220 explicitly asked for that surface before graduation. One interim ask: this diff introduces a *new* hang mode. The prior storage-only path returned a valid capture on a healthy-storage/unhealthy-stream service; the load-path `connect` retries a retriable failure forever, so a persistently-unreachable-but-retriable stream emits neither `"connect"` nor `"closed"` and both `Promise.race` calls pend indefinitely. Consider passing a finite `maxInitialConnectionAttempts` (e.g. `1`) to the capture `createDeltaManager` call so it fails fast even while the full `abortSignal` wiring lands in the follow-up. Flagging for owner (markfields / vladsud) sign-off on whether shipping the interim hang mode is acceptable.

Abe27342: I think keeping maxInitialConnectionAttempts is unbounded is probably more in line with other Fluid APIs so would lean towards not doing it. Adding an abort signal as follow-up is probably sufficient.

If human reviewer feels strongly otherwise though, I'll make this change.

alexvy86 (approved):

PR #28010, opened 2026-08-18
test(test-service-load): log driverTenantName to telemetry from stress tests

shlevari:
Stress tests did not log the driverTenantName telemetry dimension that E2E tests and perf benchmarks already emit, so there was no way to tell which ODSP tenant a stress run exercised. That makes it hard to investigate whether a particular tenant is being hit too hard, which is a likely source of throttling errors.

Add the dimension to the orchestrator and to each child runner. The runner now creates its test driver before the logger so the tenant name is available as a dimension; driver creation is wrapped so a failure there still reports runnerFailed telemetry instead of exiting silently.

ADO: 66983


alexvy86 (approved):
The Copilot suggestion seems good, otherwise LGTM. Ideally let's verify this before merging by pushing it to a test/ branch in the upstream repo and doing a manual run of `Build - client packages` -> `Real Service Stress test` and checking that the new field shows up in Kusto telemetry for that run.

alexvy86 (approved):

PR #28012, opened 2026-08-19
feat(tree): history and rewind

yann-achard-MS:
## Description

Adds features to allow applications to rewind a view to a specific revision by switching to a new SharedTreeBranch that is forked from that revision.

## Breaking Changes

None


packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1922,18 +1922,32 @@ export interface TreeBranchAlpha extends TreeBranch, TreeContextAlpha {
...
+    rewindTo(revision: string): void;
     runTransaction<TSuccessValue, TFailureValue>(transaction: () => TransactionCallbackStatusAlpha<TSuccessValue, TFailureValue>, params?: RunTransactionParamsAlpha): TransactionValueResult<TSuccessValue, TFailureValue>;
     runTransaction(transaction: () => VoidTransactionCallbackStatusAlpha | void, params?: RunTransactionParamsAlpha): TransactionVoidResult;
     runTransactionAsync<TSuccessValue, TFailureValue>(transaction: () => Promise<TransactionCallbackStatusAlpha<TSuccessValue, TFailureValue>>, params?: RunTransactionParamsAlpha): Promise<TransactionValueResult<TSuccessValue, TFailureValue>>;
     runTransactionAsync(transaction: () => Promise<VoidTransactionCallbackStatusAlpha | void>, params?: RunTransactionParamsAlpha): Promise<TransactionVoidResult>;
 }
 
+// @alpha @sealed
+export interface TreeBranchCommitMetadata {
+    readonly parent: TreeBranchCommitMetadata | undefined;
```
noencke: This creates a linked list, which may be the best design. I'll propose another, and we can compare:

```ts
interface TreeBranchCommitMetadata {
    readonly parent: string | undefined;
    readonly revision: string;
}

interface TreeBranchHistory {
    readonly head: TreeBranchCommitMetadata | undefined;
    readonly size: number;
    readonly get(revision: string): TreeBranchCommitMetadata | undefined;
}
```

Pros:
* Users need not index/search for commits from a revision themselves, we'll do it for them. This is especially nice if we already have such a map internally, so that they don't have to create the same map themselves.
* `TreeBranchCommitMetadata` is self-contained per commit. JSON.stringifying a commit wouldn't, for example, actually serialize all the ancestor commits as well.
* `TreeBranchHistory` API is more like a std lib collection API.

Cons:
* Not immediately obvious how to deref the parent commit from a child commit - you need the `TreeBranchHistory`.
* Can't do `commit.parent.parent`, etc.

I don't _expect_ apps to typically be cognizant of "how many commits they want to climb upwards" - rather, I expect them to have handles to commits, hence the deref API. But, I could be wrong. #Resolved

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1922,18 +1922,32 @@ export interface TreeBranchAlpha extends TreeBranch, TreeContextAlpha {
     // (undocumented)
     fork(): TreeBranchAlpha;
     hasRootSchema<TSchema extends ImplicitFieldSchema>(schema: TSchema): this is TreeViewAlpha<TSchema>;
+    readonly history: TreeBranchHistory;
     isMissingEditsFrom(branch: TreeBranch): boolean;
+    rewindTo(revision: string): void;
```
noencke: Should this be within the `history` API? Keep things scoped? `history.rewindTo` reads well too. #Resolved

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1922,18 +1922,32 @@ export interface TreeBranchAlpha extends TreeBranch, TreeContextAlpha {
...
+
 // @alpha @sealed
 export interface TreeBranchEvents {
     changed(data: ChangeMetadata, getRevertible?: RevertibleAlphaFactory): void;
 }
 
+// @alpha @sealed
+export interface TreeBranchHistory {
+    readonly commitCount: number;
+    getHeadCommit(): TreeBranchCommitMetadata | undefined;
```
noencke: When can the head commit be undefined, again? In practice, not on document init, because we have to schematize a view (which does at least 1 commit) before being able to create the view object. The other possibility is if we GC every commit? Is that possible, or do we always leave at least one? #Resolved

packages/dds/tree/src/shared-tree/history.ts
```diff
@@ -0,0 +1,140 @@
...
+import type { GraphCommit, RevisionTag } from "../core/index.js";
+import type { SharedTreeBranch, SharedTreeBranchChange } from "../shared-tree-core/index.js";
+import type { TreeBranchCommitMetadata, TreeBranchHistory } from "../simple-tree/index.js";
+
+import type { SharedTreeChange } from "./sharedTreeChangeTypes.js";
+import type { SharedTreeEditBuilder } from "./sharedTreeEditBuilder.js";
+
+class LazyTreeBranchCommitMetadata implements TreeBranchCommitMetadata {
+	public readonly revision: string;
+	private priorParent?: GraphCommit<SharedTreeChange>;
```
noencke: nit: `priorParent` and `cachedParent` should always be set at the same time - never one and not the other - or else caching behavior can break. This can be enforced in the type system by making them required within a single optional object, at the cost of being a little verbose.

```ts
private parent?: {
   prior: GraphCommit<SharedTreeChange>;
   cached: TreeBranchCommitMetadata;
}
``` #Resolved

File: packages/dds/tree/api-report/tree.alpha.api.md (hunk 2 above)
Comment (yann-achard-MS) [reply packages/dds/tree/api-report/tree.alpha.api.md]: I agree that's tempting. Here's my thinking:

Methods that have side effects on some object (here, the `TreeBranch`) are best exposed directly on that object so that it's clear which object is being affected.

This seems particularly prudent here because of the following:
Multiple branches might have a commit with a given revision Foo. This could be because Foo happens to be part of the common ancestry of these branches, but it could also be because the branches have different versions of Foo. To the application, there is no way to tell (at least with the current API) because you'll just get two `TreeBranchCommitMetadata` metadata objects that are structurally identical. My fear is that an application author might get these objects mixed up (or think them interchangeable) and call `rewindTo` on the wrong one.

File: packages/dds/tree/src/shared-tree/history.ts [hunk 5]
```diff
@@ -0,0 +1,140 @@
...
+	public get commitCount(): number {
+		if (!this.commitCountInitialized) {
+			this.cachedCommitCount = this.countCommits();
+			this.subscribeToBranch();
+			this.commitCountInitialized = true;
+		}
+		return this.cachedCommitCount;
+	}
+
+	private subscribeToBranch(): void {
```
noencke: This should either early return if we're already subscribed, or unsubscribe and then resubscribe.

Or, just inline it into `commitCount()` where we know it won't be called more than once, and then don't worry about it. #Resolved

packages/dds/tree/src/shared-tree/history.ts
```diff
@@ -0,0 +1,140 @@
...
+		this.unsubscribeAfterChange = this.branch.events.on(
+			"afterChange",
+			this.onAfterBranchChange,
+		);
+		this.unsubscribeAncestryTrimmed = this.branch.events.on(
+			"ancestryTrimmed",
+			this.onAncestryTrimmed,
+		);
+	}
```
noencke, line 73: Do you want to wire into branch.events.dispose too, and dispose this object when the branch is disposed? #ByDesign

packages/dds/tree/src/shared-tree/history.ts
```diff
@@ -0,0 +1,140 @@
...
+
+	private countCommits(): number {
```
noencke: This seems like a candidate for a property on the branch object itself rather than here, because - since we need it - why not?

`branch.size` or something. It can be documented as O(n). #Resolved

packages/dds/tree/src/shared-tree/history.ts
```diff
@@ -0,0 +1,140 @@
...
+			}
+		}
+	};
+
+	private readonly onAncestryTrimmed = (trimmedRevisions: RevisionTag[]): void => {
+		if (!this.commitCountInitialized) {
+			return;
+		}
+
+		this.cachedCommitCount = Math.max(0, this.cachedCommitCount - trimmedRevisions.length);
```
noencke: Under what circumstances would this calculation be negative and the Math.max necessary? Surely we shouldn't ever trim more revisions than exist (even in the cache) right? #Resolved

packages/dds/tree/src/shared-tree/history.ts
```diff
@@ -0,0 +1,140 @@
...
+				parent === undefined || parent.revision === "root"
+					? undefined
+					: new LazyTreeBranchCommitMetadata(parent, this.idCompressor);
+			this.priorParent = parent;
+		}
+		return this.cachedParent;
+	}
+}
+
+export class TreeBranchHistoryImpl implements TreeBranchHistory {
```
noencke: Pretty much the entirety of the code in this class exists to cache/memoize the computation of the branch length, as opposed to doing necessary work for implementing `TreeBranchHistory` (the exception is `getHeadCommit()`. To that end, do you think we should actually name and scope this class as such (`SharedTreeBranchSizeMemoizer` or whatever), and then wrap it in a small shim to conform it to the `TreeBranchHistory` interface where necessary (i.e. in schematizing tree view)?

Going further - should we just do this caching in the branch object itself? Especially if we [add a size property](https://github.com/microsoft/FluidFramework/pull/28012/changes#r3815521956)? #Resolved

packages/dds/tree/api-report/tree.alpha.api.md
yann-achard-MS: We can indeed GC every commit, though in practice that means keeping the more recent commit object and gutting it out to make it the new root commit. We *could* expose that as the head commit by returning "root" (nobody cares what the revision string is) but I think that would be a mistake for two reasons:
1. It would mean that the app could get "root" as a revision at different times when reading and each such "root" values would actually refer to different points in the history. Even worse, supplying this "root" revision to `rewindTo` will mean different things at different times. So, an app could read a revision from the history (which happens to be "root"), then later pass that to `rewindTo` and it's totally possible that the state they would be brought back to is a different state than the one when they read the "root" value.
2. I expect we'll want to add some valuable information to the metadata, like the sequencing status of a commit. I worry that some of this metadata won't always make sense for the "root" commit.

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -593,6 +595,11 @@ export class TreeCheckout implements ITreeCheckout {
 		this.registerForBranchEvents();
 	}
 
+	public get history(): TreeBranchHistoryImpl {
+		this.branchHistory ??= new TreeBranchHistoryImpl(this.branch, this.idCompressor);
+		return this.branchHistory;
```
noencke: Put this on one line, save the trees! #ByDesign

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1349,6 +1355,12 @@ export class TreeCheckout implements ITreeCheckout {
 		this.unregisterFromBranchEvents();
 
 		this.#transaction = this.createTransactionStack(branch);
+		if (!this.isSharedBranch) {
+			this.branch.dispose();
```
noencke: Is this related to history? Looks like just a bug fix (which is fine, just wondering). #Resolved

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1349,6 +1355,12 @@ export class TreeCheckout implements ITreeCheckout {
...
+		}
+		this.branch = branch;
+		this.branchHistory?.dispose();
```
noencke: This wouldn't be necessary if the history object [auto-disposed in response to the branch dispose](https://github.com/microsoft/FluidFramework/pull/28012/changes#r3815504061). #ByDesign

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1357,6 +1369,24 @@ export class TreeCheckout implements ITreeCheckout {
...
+	public rewindTo(revisionString: string): void {
+		this.checkNotDisposed("The branch has already been disposed and cannot be rewound.");
+		assert(this.#transaction.size === 0, "Cannot rewind during a transaction");
+		const revision = this.idCompressor.tryRecompress(revisionString as StableId);
+		if (revision === undefined) {
+			throw new UsageError(`Unrecognized revision id: ${revisionString}`);
+		}
+		const head = this.#transaction.branch.getHead();
+		let targetCommit = head;
+		while (targetCommit.parent !== undefined && targetCommit.revision !== revision) {
```
noencke: nit: Don't we have a "find ancestor" helper or something somewhere that does this? #Resolved

packages/dds/tree/src/shared-tree/treeCheckout.ts
```diff
@@ -1357,6 +1369,24 @@ export class TreeCheckout implements ITreeCheckout {
...
+			targetCommit = targetCommit.parent;
+		}
+		if (targetCommit.revision !== revision) {
+			throw new UsageError(`No commit found with revision: ${revisionString}`);
+		}
+		this.switchBranch(this.#transaction.branch.fork(targetCommit));
```
noencke, line 1419: This came out so clean since we already have this switcher! Awesome! #Resolved

packages/dds/tree/api-report/tree.alpha.api.md
yann-achard-MS: I was hoping for some discussion on this. :)

On key thing to keep in mind, and perhaps the API should be better at making this obvious, is that the linked list is inherently unstable (even when ignoring trunk GC): calling `get(revFoo)` may return a different ancestry at different times for the same `revFoo`. So, what the current API gives you when you first call `getHeadCommit()` is a fixed snapshot of the history because the implementation captures the graph commit object.

The other aspect that pushed me towards the current API is the desire to keep the middleware (between the ST internal history and the history that we expose to the app) as thin and dumb as possible. Part of this is because time was scarce, but the more important/durable reason is that any smarts we do inject in that middleware increase the risk of doing more work than strictly necessary in some scenario. I think the alternative API you are floating here is in keeping with that, but I still wanted to mention it.

> This is especially nice if we already have such a map internally, so that they don't have to create the same map themselves.

We don't have such a map internally, but would could lazily build one.

> `TreeBranchCommitMetadata` is self-contained per commit. JSON.stringifying a commit wouldn't, for example, actually serialize all the ancestor commits as well.

I hadn't thought of that. It seems like it would be easily circumvented (we could offer a `toJson` method). Either way, do we expect apps to have a need to serialize this? ST will be persisting this data already. Although as mentioned above, the history can change, so what ST give you will be different, but that seems like a good thing: we don't want the app to keep stale history data.

> `TreeBranchHistory` API is more like a std lib collection API.

I wonder if that's a good or a bad thing when it is a shifting thing. No strong opinion.

packages/dds/tree/src/shared-tree/history.ts
```diff
@@ -0,0 +1,140 @@
...
+		private readonly commit: GraphCommit<SharedTreeChange>,
+		private readonly idCompressor: IIdCompressor,
+	) {
+		assert(commit.revision !== "root", "Cannot construct metadata for the root commit");
+		this.revision = idCompressor.decompress(commit.revision);
+	}
+
+	public get parent(): TreeBranchCommitMetadata | undefined {
+		// We cannot merely cache the computation once because the parent of the commit may change over time due to rebases and trunk trimming.
+		if (this.commit.parent !== this.priorParent) {
```
yann-achard-MS: I think the point about trunk trimming is correct but not the point about rebasing. Needs confirmation. #Resolved

packages/dds/tree/src/shared-tree/history.ts
yann-achard-MS: Yeah, this should be an assert of `this.cachedCommitCount >= trimmedRevisions.length` instead.

packages/dds/tree/src/shared-tree/history.ts
yann-achard-MS: I love it!

packages/dds/tree/src/shared-tree/history.ts
yann-achard-MS: I agree that it would make sense to isolate all that `commitCount` logic given the current code. I wonder if that will become less true as more features are added, but we can cross that bridge when we get to it. I'll give that refactor a go.

About doing this caching on the branch object itself, what do you see as the incentive? Is it so that this caching can be shared by multiple views that are riding the same `SharedTreeBranch`? That seems like something not very likely to happen. I think the incentive against it is that we want to keep `SharedTreeBranch` simple and avoid giving it more responsibilities.

If we were to bake in the counting at the `SharedTreeBranch` level, then such counting likely should include the "root" commit. Then the counting at the `TreeBranchHistory` layer would wrap that and decrement it by 1.

packages/dds/tree/src/shared-tree/history.ts
yann-achard-MS: Yeah, I'll make an `unsubscribeFromBranch` method and call it in `subscribeToBranch` and `dispose`.

packages/dds/tree/src/shared-tree/history.ts
yann-achard-MS: mmm... 🤔 I think not.
This object is owned by the `TreeBranch`, so I think it would be strange for it to sneakily dispose itself when the `SharedTreeBranch` is disposed.

Another thing against it: it seems fine for the history to survive (meaning, have lifetime that extends beyond that of) the `SharedTreeBranch`. The subscription to the branch is so that this history can stay current. If the branch goes away, that just means the history will stop changing, but it doesn't mean the history is no longer true.

packages/dds/tree/src/shared-tree/history.ts
yann-achard-MS: I'll copy some of my comments above to the other comment thread so this one can be scoped to your first point.

packages/dds/tree/src/shared-tree/history.ts
yann-achard-MS: About doing this counting/caching on the branch object itself, what do you see as the incentive? Is it so that this counting/caching can be shared by multiple views that are riding the same `SharedTreeBranch`? That seems like something not very likely to happen. I think the incentive against it is that we want to keep `SharedTreeBranch` simple and avoid giving it more responsibilities.

If we were to bake in the counting at the `SharedTreeBranch` level, then such counting likely should include the "root" commit. Then the counting at the `TreeBranchHistory` layer would wrap that and decrement it by 1.

yann-achard-MS: I think that last point above about counting policy (whether to include the root commit or not) just helped me make up my mind. That commit-counting code currently mixes two things together:
1. The knowledge of how to deal with a `SharedTreeBranch` in order to get the true commit count (which includes root commits) and maintain it.
2. A policy about the fact that root commits should not be counted.

It does make sense to keep (2) at the `TreeBranchHistory` layer. The same cannot be said for (1).
Here's my proposal:

* We isolate the counting code as you suggested but house it _alongside_ (not inside) the `SharedTreeBranch` code. It's a helper class that works with a `SharedTreeBranch`, it shouldn't know about "root" commits and so will count them in.
* `TreeBranchHistory` leverages that but removes one from the count.

How does that sound?

packages/dds/tree/src/shared-tree/treeCheckout.ts
yann-achard-MS: No! I don't like statements inside my expressions! 😝

packages/dds/tree/src/shared-tree/treeCheckout.ts
yann-achard-MS: It does look like a bug fix but it's not: this was only called from a context where the `isSharedBranch` would have been true. It is now being called from a context where it might be either.

packages/dds/tree/src/shared-tree/treeCheckout.ts
yann-achard-MS: See other thread about same topic.

packages/dds/tree/api-report/tree.alpha.api.md
noencke: Makes sense, I think it's good how it is then.

packages/dds/tree/src/shared-tree/history.ts
yann-achard-MS: I have done that. LMK what you think.

packages/dds/tree/api-report/tree.alpha.api.md
noencke: Yea, simple is good! Good points... I say keep it how it is for now.

packages/dds/tree/src/shared-tree/history.ts
noencke: Makes sense.

packages/dds/tree/src/shared-tree/history.ts
noencke: I think that's an improvement. A count/size still feels like a natural and intuitive operation to have as a method on a branch IMO, but it gets muddy with the caching, since caching has a cost if you don't use it, so I think on the side is a fine solution too.

packages/dds/tree/src/shared-tree/treeCheckout.ts
noencke: 💀 🌴 🌳 🌲 💀

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1881,11 +1881,23 @@ export type TreeBranch = UntypedTreeView;
...
+}
+
 // @alpha @sealed
 export interface TreeBranchEvents {
     changed(data: ChangeMetadata, getRevertible?: RevertibleAlphaFactory): void;
 }
 
+// @alpha @sealed
+export interface TreeBranchHistory {
+    readonly commitCount: number;
```
noencke: nit: `length`? In the spirit of being conventional. I don't think it risks being ambiguous - it's pretty clear what "the length of the history" would mean, IMO.

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1881,11 +1881,23 @@ export type TreeBranch = UntypedTreeView;
 // @alpha @deprecated
 export type TreeBranchAlpha = UntypedTreeViewAlpha;
 
+// @alpha @sealed
+export interface TreeBranchCommitMetadata {
+    readonly parent: TreeBranchCommitMetadata | undefined;
```
noencke: As much as I like short names and property getters, the regretfully responsible part of me thinks this should be a method, because it can change out from under the user (just like `getHeadCommit()`.

packages/dds/tree/api-report/tree.alpha.api.md
```diff
@@ -1881,11 +1881,23 @@ export type TreeBranch = UntypedTreeView;
...
+
 // @alpha @sealed
 export interface TreeBranchEvents {
     changed(data: ChangeMetadata, getRevertible?: RevertibleAlphaFactory): void;
 }
 
+// @alpha @sealed
+export interface TreeBranchHistory {
+    readonly commitCount: number;
+    getHeadCommit(): TreeBranchCommitMetadata | undefined;
```
noencke: `getHead()`?

.changeset/tree-branch-history.md
```diff
@@ -0,0 +1,23 @@
+---
+"fluid-framework": minor
+"@fluidframework/tree": minor
+"__section": tree
+---
+New alpha APIs for inspecting history and restoring past states
+
+`UntypedTreeViewAlpha` now exposes a `branchHistory` property which returns a `TreeBranchHistory` object with:
```
Josmithr: Nit: can we link to API docs for at least existing APIs here?

packages/dds/tree/src/shared-tree/history.ts
```diff
@@ -0,0 +1,80 @@
...
+import type { IIdCompressor } from "@fluidframework/id-compressor";
+
+import type { GraphCommit } from "../core/index.js";
+import { BranchCommitCounter, type SharedTreeBranch } from "../shared-tree-core/index.js";
+import type { TreeBranchCommitMetadata, TreeBranchHistory } from "../simple-tree/index.js";
+
+import type { SharedTreeChange } from "./sharedTreeChangeTypes.js";
+import type { SharedTreeEditBuilder } from "./sharedTreeEditBuilder.js";
+
+class LazyTreeBranchCommitMetadata implements TreeBranchCommitMetadata {
```
Josmithr, line 19: Nit: docs?

packages/dds/tree/src/shared-tree/history.ts
```diff
@@ -0,0 +1,80 @@
...
+					: {
+							prior: parent,
+							cached: new LazyTreeBranchCommitMetadata(parent, this.idCompressor),
+						};
+		}
+		return this.parentCache?.cached;
+	}
+}
+
+export class DefaultTreeBranchHistory implements TreeBranchHistory {
```
Josmithr, line 53: Nit: docs?

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -335,6 +384,34 @@ export interface UntypedTreeViewAlpha extends UntypedTreeView, TreeContextAlpha
...
+	 *
+	 * @param revision - The {@link TreeBranchCommitMetadata.revision | revision} to rewind to.
+	 * Can be obtained by navigating the commits on the {@link UntypedTreeViewAlpha.branchHistory | branch history}.
+	 *
+	 * @remarks
+	 * Unlike {@link UntypedTreeViewAlpha.revertTo | revertTo}, this does not apply a change to the underlying branch.
+	 * The original underlying branch will be disposed unless it is the main branch or a {@link (ITreeAlpha:interface).createSharedBranch | shared branch}.
+	 * In order to retain the local branch, consider {@link UntypedTreeViewAlpha.fork | forking} before rewinding.
+	 */
+	rewindTo(revision: string): void;
```
Josmithr: Nit: Is it safe to assume that the format of the revision is intended to be opaque to users? I'm wondering if there might be any value in creating a type alias for them that documents any expectations in a single place.

Josmithr (approved):
Left a couple minor suggestions, but otherwise docs and API changes look good. I didn't look at the implementation details in much detail though, would be good to get another tree approval there.

.changeset/tree-branch-history.md
yann-achard-MS: `UntypedTreeViewAlpha` is not in the docs yet, so I can't link to it.

packages/dds/tree/src/simple-tree/api/tree.ts
noencke: I think `revertTo` is a good name and aligns with `Revertible`.

Is "rewind" too similar? Will customers get confused? I can't think of a better name at the moment though... maybe we just refine the API later

packages/dds/tree/src/simple-tree/api/tree.ts
```diff
@@ -335,6 +384,34 @@ export interface UntypedTreeViewAlpha extends UntypedTreeView, TreeContextAlpha
...
 
+	/**
+	 * Switches this view to a new underlying branch with the given commit as the head, updating the view state accordingly.
+	 *
+	 * @param revision - The {@link TreeBranchCommitMetadata.revision | revision} to rewind to.
+	 * Can be obtained by navigating the commits on the {@link UntypedTreeViewAlpha.branchHistory | branch history}.
+	 *
+	 * @remarks
+	 * Unlike {@link UntypedTreeViewAlpha.revertTo | revertTo}, this does not apply a change to the underlying branch.
+	 * The original underlying branch will be disposed unless it is the main branch or a {@link (ITreeAlpha:interface).createSharedBranch | shared branch}.
```
noencke: What is the proper use case in which someone calls `rewindTo` on the main branch? Isn't that usually (always?) a mistake, because they'll lose their reference to the main branch after the switch? They'd have to then make another view from the sharedTree or something to get it back, no? Should we just ban this API from happening on the main branch for now?

noencke (approved):

noencke (approved):

Josmithr (approved):

Josmithr (approved):

PR #27891, opened 2026-08-07
Move marker and load to sequence number API to legacy alpha

sonalideshpandemsft:
- Promotes these APIs from internal imports to legacy-alpha entrypoints:
  - `IVersionMarkResolver`, `ResolveResult`, and `VersionMarkCapture`
  - `loadContainerToSequenceNumber` and `ILoadContainerToSequenceNumberProps`
  - `OdspPointInTimeDocumentServiceFactory`
- Renames `captureVersionMark(): Promise<VersionMarkCapture>` to synchronous `sealAndCaptureVersionMark(): VersionMarkCapture`, making its batch-flush side effect explicit.
- Adds the required legacy-alpha package exports, API Extractor configuration, API reports, and API documentation.
- Updates tests for the renamed synchronous capture API.
- Expands `versionMarks/DEV.md` with implementation details, MSN cache behavior, limitations, and future work

PR #28017, opened 2026-08-19
feat(container-runtime): make version-mark sequenceNumberLowerBound i…

lindsnguyen:
## Description

Changes the `sequenceNumberLowerBound` field of a version-mark `pending` capture from an **exclusive** to an **inclusive** lower bound ([AB#80024](https://dev.azure.com/fluidframework/235294da-091d-4c29-84fc-cdfc3d90890b/_workitems/edit/80024)).

`sealAndCaptureVersionMark()` now stores `referenceSequenceNumber + 1` — the pending batch's first possible sequence number — and `resolveFromHistory` scans directly from it, moving the `+ 1` from read-time to capture-time. The actual scanned op range is unchanged, so resolution results (`resolved` / `pending` / `unresolvable`) are identical; the stored locator is now self-describing.

No API surface change — signatures are unchanged and the API report is unaffected. TSDoc, `DEV.md`, and `classifyMiss` comments are updated to inclusive semantics.

## Reviewer Note

`sequenceNumberLowerBound` is now an **inclusive** lower bound — the mark's ops are at **or after** this sequence number (previously strictly after).

- **If you only round-trip the value** (capture → persist → pass back to `resolve()`): **no change needed** — producer and consumer stay consistent.
- **If you construct or interpret the value yourself:** treat it as the batch's first possible sequence number. Any code that added `+ 1` before calling `resolve()` must **drop that `+ 1`**, and any comparison/display should read it as "at or after," not "strictly after."

The prior exclusive format was never shipped (still `@beta`, no durably persisted marks), so there are no old-format locators to migrate. `DEV.md` documents this and the deferred plan for handling any future persisted-format change compat-safely.

## Testing

Updated the existing `VersionMarkResolver` unit tests to the inclusive semantics (capture now stores `S + 1`; the history scan reads from the stored value directly) and added boundary coverage:

- **Inclusive boundary** — a batch whose op sits exactly at `sequenceNumberLowerBound` resolves and does not trip the reader-range assert.
- **`from == tip` boundary** — an empty read where the lower bound equals the tip classifies as `unresolvable` (pre-existing gap between the `from < tip` and `from > tip` cases).

The scanned op range is unchanged, so all existing `pending` / `resolved` / `unresolvable` classifications are preserved.


packages/runtime/container-runtime/src/test/versionMarks/versionMarkResolver.spec.ts
```diff
@@ -351,6 +353,47 @@ describe("VersionMarkResolver", () => {
...
+			);
+			const resolver = makeResolver({ reader });
+			assert.deepEqual(await resolver.resolve(generateBatchId("clientA", 7), 15), {
+				kind: "resolved",
+				sequenceNumber: 15,
+			});
+			assert.deepEqual(calls, [{ from: 15, to: undefined }]);
+		});
+
+		it("still resolves a locator persisted under the old exclusive scheme", async () => {
```
markfields: Back/forwards compat may not be worth testing since we never shipped the old thing. This isn't a test we need long-term at all. 

That said, it's a good opportunity to think through how we'll handle compat like this in the future. We definitely would need back compat since the pending marks are stored by the app. Is it possible we would need fowards compat, where old code interacts with new persisted format? Something to think about.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -283,7 +286,7 @@ export class VersionMarkResolver implements IVersionMarkResolver {
 		);
 		const tip = this.hooks.getCurrentSequenceNumber();
 		if (from > tip) {
```
markfields, line 288: Should this change by 1 somehow?

markfields: And any of the if checks below?

markfields (commented):
Can you double-check the logic in the last file? Wondering if any of those if-checks need to change by 1.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
lindsnguyen: No, `from` ends up as the identical value it always was (the  +1  just relocated from read-time to capture-time), so `classifyMiss` receives the same input and its boundaries are unchanged.

anthony-murphy:

- 🤖 **Solution / Problem Space Review** — wandering the abyssal plain…
- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:

- ⬜ Adversarial Filter
- ⬜ Synthesis
- ⬜ Comments

anthony-murphy:
## Deep Review

*Reviewed commit `3d04bf6` on 2026-08-19.*

No blocking correctness issues after weighing the conversation. The one substantive risk — old exclusive persisted locators misbehaving on the miss and orphan-batch paths — is gated entirely on whether that format ever shipped, and markfields notes it didn't. Remaining work is a small set of test/doc items flagged inline, resolving that back-compat decision, and getting CI green.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Resolve the back-compat thread with markfields: confirm whether the old exclusive format was ever persisted by any app. If not, drop the back-compat test and add a one-line `DEV.md` note that old locators cannot exist; if yes, add the miss-path and orphan-tail coverage flagged inline.
- [ ] Fix CI: the build is failing (bundle-size comparison unavailable) and the website link-check reports broken links — this PR edits `DEV.md` and carries the `area: website` label.

### Context for Reviewers

- Completes markfields' request in #27885 and finalizes a three-step inclusive-lower-bound evolution: #27834 introduced `resolve(batchId, referenceSequenceNumber)` with a read-time `+ 1`, #27885 renamed the field to `sequenceNumberLowerBound`, and this PR stores `referenceSequenceNumber + 1` at capture and scans directly from it. The scanned op range is numerically identical for new marks.
- Old-locator safety for chunk streams rides on #27919's resolver-side skip of clipped leading chunk streams, not on this capture-side edit — `referenceSequenceNumber` is still `deltaManager.lastSequenceNumber`, so `referenceSequenceNumber + 1` is not batch-aligned mid-chunk-stream.
- Suggested reviewers: **markfields** — design gatekeeper on this file, requested this exact change, and authored the ordinary-batch invariant (in #27919) that the inline threads probe; he is already commenting but is not a requested reviewer. Also **lindsnguyen** (authored #27834/#27919, best on classification and chunk-stream behavior) and **sonalideshpandemsft** (renamed the field; already requested).

- **Needs human judgment** — The back-compat strategy is markfields' call: whether the exclusive format ever shipped, whether long-term back/forward-compat handling is needed, and whether the back-compat test stays or goes. He has already opened this in [the back-compat thread](https://github.com/microsoft/FluidFramework/pull/28017#discussion_r3816016379). His answer decides whether the two inline back-compat concerns are live bugs or dead code paths.
- **Needs human judgment** — #27919 removed the ordinary-batch orphan-end guard on the premise that scans start at `sequenceNumberLowerBound + 1`. Whether that premise still holds under the inclusive anchor for a pre-PR locator scanned from `S` is a design judgment only markfields/lindsnguyen can settle.
- **Cannot be assessed by the pipeline** — a real ODSP `fetchMessages` trim at `S` cannot be exercised here, so the `pending`-vs-`unresolvable` behavior on a genuinely trimmed old locator is unverified at runtime.

</details>

packages/runtime/container-runtime/src/test/versionMarks/versionMarkResolver.spec.ts
anthony-murphy: Deep Review: This is the crux of the back-compat coverage question. The added test only exercises the `resolved` happy-path where op `S` is still present. The miss paths that `classifyMiss` governs are untested for a pre-PR persisted locator: if op `S` is trimmed but ops from `S + 1` remain and the target batch is absent, the scan now starts at `from = S` (was `from = S + 1`), so `firstScannedSequenceNumber (S+1) > from (S)` → `{ kind: "unresolvable" }`, where the old read-time `+ 1` gave `{ kind: "pending" }`. The `tip === S` boundary flips the same way.

If the exclusive format was never shipped, as you note, the cleanest resolution is to drop this test and add a one-line `DEV.md` note that old locators cannot exist. If old locators must be tolerated, keep the test and add a miss-path case (op `S` trimmed, plus the `tip === S` boundary) asserting the result stays `pending`, not `unresolvable`.

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
anthony-murphy: Deep Review: Agreed with lindsnguyen that `from` is identical for new marks — the `+ 1` just relocated from read-time to capture-time, so `classifyMiss` inputs and its boundary checks are unchanged. The one edge this leaves is a locator persisted before this PR under the old exclusive scheme (value `S`): it now scans from `S` instead of `S + 1`, so on a miss where op `S` is trimmed the `firstScannedSequenceNumber > from` check flips `pending` → `unresolvable`. Whether that path is reachable depends on the shipped-format question in [the back-compat thread](https://github.com/microsoft/FluidFramework/pull/28017#discussion_r3816016379).

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
```diff
@@ -232,7 +235,7 @@ export class VersionMarkResolver implements IVersionMarkResolver {
 		batchId: string,
 		sequenceNumberLowerBound: number,
 	): Promise<ResolveResult> {
-		const from = sequenceNumberLowerBound + 1;
+		const from = sequenceNumberLowerBound;
```
anthony-murphy, line 238: Deep Review: `resolveFromHistory` now scans from the stored value directly (`const from = sequenceNumberLowerBound;`) and feeds the first returned op straight into `unpack?.(op)` before deriving batch identity. For a locator persisted under the old exclusive scheme, `S` can be the final op of a preceding ordinary multi-op batch, so the scan begins on that batch's `batch: false` orphan tail. #27919 removed the ordinary-batch orphan-end guard on the premise that scans start at `sequenceNumberLowerBound + 1` — a premise that no longer holds for old locators scanned from `S`. `RemoteMessageProcessor.getResultBasedOnBatchMetadata` asserts `0x9d5` ("Unexpected batch end marker") on a `batch: false` marker with no batch in progress — the exact failure `DEV.md:162` flags. The added back-compat test covers only a single non-matching op at `S`, not a `batch: false` tail.

Gate this on the same [shipped-format question](https://github.com/microsoft/FluidFramework/pull/28017#discussion_r3816016379): if old locators are in scope, add a `batch: false` ordinary-batch-tail-at-`S` test or restore the orphan-end guard for the compat window; if the old format was never shipped, document that this case cannot arise and no guard is needed.

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -306,8 +306,8 @@ The existing real-service ODSP suites under `packages/test/test-end-to-end-tests
 ### Resolver correctness and robustness gaps
```
anthony-murphy, line 308: Deep Review: This edit removes "and a lower bound whose `+ 1` overflows safe integer precision" from the stored-locator-validation gap — a concern deliberately added in #27919's robustness list. Dropping it at the read side is reasonable now that the app value is consumed directly, but the `+ 1` moved to `sealAndCaptureVersionMark` (`referenceSequenceNumber + 1`), so the arithmetic still exists, just at capture time. Either state the overflow concern no longer applies (persisted value consumed directly), or relocate the note to the capture site as the arithmetic now subject to the safe-integer consideration.

packages/runtime/container-runtime/src/test/versionMarks/versionMarkResolver.spec.ts
lindsnguyen: Since the exclusive format was never shipped, I took the drop-the-test route: removed it and added a DEV.md note that no pre-PR persisted locators exist, so the miss-path reclassification (and the  tip === S  boundary) can't arise. Kept the two new boundary tests (op exactly at the inclusive lower bound, and  from == tip ).

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
lindsnguyen: good catch that this is the more severe version: scanning an old locator from `S` could hit a preceding batch's `batch: false` orphan tail and trip `0x9d5`, not just misclassify. Since the exclusive format was never shipped, I documented in DEV.md that this can't arise (scans always start at the batch's own first op) so no guard is needed, and flagged that any future change to the stored anchor's meaning must revisit it.

packages/runtime/container-runtime/src/versionMarks/DEV.md
lindsnguyen: The + 1 moved rather than disappeared. Relocated the note to the stored-locator-validation item to say the safe-integer consideration now applies at capture (`referenceSequenceNumber + 1` in `sealAndCaptureVersionMark`) instead of at read.

packages/runtime/container-runtime/src/test/versionMarks/versionMarkResolver.spec.ts
lindsnguyen: Forgot to address this point: `Is it possible we would need fowards compat, where old code interacts with new persisted format?` — yeah we do need to think about it, but luckily there's precedent for how we'd handle it.

It's actually required in our case since marks get resolved cross-client across a mixed-version fleet, so an older client will eventually read a mark written by a newer one (same version-skew thing as the layer-compat window).

Good news is we don't have to do anything now. Today's inclusive format is basically the implicit v1 baseline, and since the old exclusive format was never shipped there's nothing to migrate. we'd only act if/when the locator's meaning or shape actually changes — and if it never does, nothing to do.

When it does change there's established FF precedent we can follow: either write-gate the new format behind `minVersionForCollab` so old readers never see it, or add a `formatVersion` field we detect on read (like `summaryFormat.getAttributesFormatVersion`, where a missing version maps to a known baseline) so the reader fails closed on something it doesn't recognize instead of silently misinterpreting it. given how small the locator is i think the lightweight `summaryFormat`-style approach fits best. i captured this as a future-work note in DEV.md so the decision + trigger are tracked.

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -155,19 +155,21 @@ For each raw op, the scan records the first returned sequence number before filt
...
 
 An **ordinary (non-chunked) batch tail is never clipped**, on the live queue or the replayed `fetchMessages` stream, so the scan needs no ordinary-batch orphan-end guard. `InboundBatchAggregator` aggregates a runtime-observed batch atomically and `DeltaQueue` drains it synchronously, so a runtime-observed `lastSequenceNumber` never lands mid-ordinary-batch; the capture anchor (`sequenceNumberLowerBound + 1`) therefore cannot start between an ordinary batch's first and last op. This holds for the replay path because it feeds the same `RemoteMessageProcessor`/`OpSplitter` pipeline, and delta storage returns whole batches (only chunk streams expose intermediate sequence numbers). If a clipped ordinary-batch tail ever did reach the unpacker, it would not resolve silently: `RemoteMessageProcessor.getResultBasedOnBatchMetadata` still asserts `0x9d5` on a `batch: false` marker with no batch in progress, so the invariant fails loudly rather than masking a regression from a future replay/capture-path change.
```
anthony-murphy: Deep Review: This orphan-tail paragraph still reads `the capture anchor (\`sequenceNumberLowerBound + 1\`)`, but the PR rewrote every other anchor reference to inclusive form — the chunk-stream sentence, both `from = sequenceNumberLowerBound` edits, and the new paragraph noting the stored value is already `referenceSequenceNumber + 1`. `DEV.md:31` and `DEV.md:57` both define the field as the inclusive first possible sequence number. Applying `+ 1` a second time to a value the same document says already includes it weakens the load-bearing safety rationale for the `0x9d5` orphan-tail invariant. Drop the `+ 1` so it reads `sequenceNumberLowerBound`, matching `resolveFromHistory` and the field definitions.

```suggestion
`InboundBatchAggregator` aggregates a runtime-observed batch atomically and `DeltaQueue` drains it synchronously, so a runtime-observed `lastSequenceNumber` never lands mid-ordinary-batch; the capture anchor (`sequenceNumberLowerBound`) therefore cannot start between an ordinary batch's first and last op. This holds for the replay path because it feeds the same `RemoteMessageProcessor`/`OpSplitter` pipeline, and delta storage returns whole batches (only chunk streams expose intermediate sequence numbers). If a clipped ordinary-batch tail ever did reach the unpacker, it would not resolve silently: `RemoteMessageProcessor.getResultBasedOnBatchMetadata` still asserts `0x9d5` on a `batch: false` marker with no batch in progress, so the invariant fails loudly rather than masking a regression from a future replay/capture-path change.
```

packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts
anthony-murphy: Deep Review: The DEV.md note that scans always start at the batch's own first op — so the `batch: false` orphan tail can't be reached — resolves the severe path, and flagging that any future change to the stored anchor's meaning must revisit it is exactly right. One residual: the orphan-tail paragraph itself (`DEV.md:162`) still literally reads `sequenceNumberLowerBound + 1`, applying the offset a second time to a now-inclusive value. Flagged inline with a one-click fix.

packages/runtime/container-runtime/src/test/versionMarks/versionMarkResolver.spec.ts
anthony-murphy: Deep Review: Dropping the test and documenting that no pre-PR persisted locators exist is the right call, and the forwards-compat plan you [outlined](https://github.com/microsoft/FluidFramework/pull/28017#discussion_r3833207795) — write-gate behind `minVersionForCollab` or a `summaryFormat`-style `formatVersion` — is a solid tripwire. One wording nit: the API did ship at `@legacy @beta` in #27891, so the safety actually rests on *no partner having durably persisted a `pending` locator*, not on the format being "never shipped." Re-grounding DEV.md on "no durably persisted `pending` marks exist yet" keeps the rationale accurate. Confirming that no `@legacy @beta` consumer has persisted one since #27891 is the one item left for a human owner.

packages/runtime/container-runtime/src/versionMarks/DEV.md
anthony-murphy: Deep Review: Relocating the safe-integer note to the capture site (`referenceSequenceNumber + 1` in `sealAndCaptureVersionMark`) is the right fix — the arithmetic moved rather than disappeared, and the note now sits where it applies. Sequence numbers stay far below `Number.MAX_SAFE_INTEGER`, so the documented deferral is fine; no code change needed.

sonalideshpandemsft (approved):

sonalideshpandemsft (approved):

anthony-murphy:
## Deep Review

*Reviewed commit `20ccc33` on 2026-08-21.*

Not signed off. The refactor is behavior-preserving — moving the `+ 1` on `sequenceNumberLowerBound` from read-time to capture-time leaves the scanned op range identical, so `resolved`/`pending`/`unresolvable` outcomes are unchanged for round-tripped locators, and no correctness or API concern survives. Three documentation-precision edits are all that stand between here and sign-off; two have one-click suggestions inline.

### Path to Ready

- [ ] Resolve inline threads
- [ ] Restate the `DEV.md` no-migration rationale — replace "never shipped" with "shipped at `@legacy @beta/@alpha` in #27891 but not known to be durably persisted by any consumer," and confirm the deferred `formatVersion`/`minVersionForCollab` guard stays tracked.

### Context for Reviewers

- Completes markfields' deferred inclusive-lower-bound request from #27885 (tracked as AB#80024) — a reviewer-requested follow-up, not a reversal. It finalizes a three-step evolution: #27834 introduced the read-time `+ 1`, #27885 renamed the field, this PR relocates the `+ 1` to capture.
- Store-side and read-side offset changes cancel: `fetchMessages(from, …)` receives the identical value, so all classifications are unchanged. The separately-tracked chunk-stream scan-anchor gap (AB#80025) is neither fixed nor worsened here.
- Brand-new, high-velocity area — `versionMarkResolver.ts` landed in #27834 and has seen ~5 PRs in two weeks; the persisted-locator meaning is still being settled at `@legacy @beta/@alpha` before any durable consumer, which is the right time to settle inclusive-vs-exclusive.

- **Needs human judgment** — Whether any `@legacy @beta/@alpha` partner has durably persisted a `pending` locator since #27891 (2026-08-07) is the area owners' call. Their answer decides whether the "never shipped" premise is merely imprecise wording or actually load-bearing for the no-migration decision. The author laid out the forward-compat strategy with FF precedent in [this reply](https://github.com/microsoft/FluidFramework/pull/28017#discussion_r3833207795).
- **Needs human judgment** — Whether to rename `sequenceNumberLowerBound` to an explicitly-inclusive name before public graduation (accepting api-report/type-test/changeset churn, and that renaming a serialized field is itself a break) is a taste/timing call; the beta surface makes it optional now.
- **Cannot be assessed by the pipeline** — A real ODSP `fetchMessages` trim at the anchor cannot be exercised here, so `pending`-vs-`unresolvable` behavior on a genuinely trimmed locator is unverified at runtime.

</details>

- `2888c69` 2026-08-21 · **9/10** — orphan-tail `+ 1` doc fix landed; DEV.md "never shipped" premise still imprecise
- `5fdc900` 2026-08-21 · **9/10** — every inline thread addressed; two doc-precision edits remained
- `3d04bf6` 2026-08-19 · **8/10** — no blocking issues; back-compat gated on never-shipped format, CI red

</details>

anthony-murphy:
`packages/runtime/container-runtime/src/versionMarks/versionMarkResolver.ts:194`

Deep Review: Line 194 still reads `// Otherwise scan history from the mark's reference point.`, but this PR redefined `sequenceNumberLowerBound` from an exclusive reference point (`from = value + 1`) to an inclusive lower bound (`const from = sequenceNumberLowerBound;`). The TSDoc three lines below was already reworded to inclusive semantics ("Scans sequenced ops from `sequenceNumberLowerBound` (inclusive)"), leaving this comment as the only place still using the pre-PR "reference point" wording. Align it with the reworded TSDoc.

```suggestion
			// Otherwise scan history from the mark's lower bound.
```

packages/runtime/container-runtime/src/versionMarks/DEV.md
```diff
@@ -155,19 +155,21 @@ For each raw op, the scan records the first returned sequence number before filt
...
 
-- `from > tip` — nothing is sequenced at/after the reference point yet, so the batch cannot have landed → **`pending`**.
+- `from > tip` — nothing is sequenced at/after the lower bound yet, so the batch cannot have landed → **`pending`**.
 - Empty read while ops should exist (`from <= tip`) — the requested range came back empty, which for a strict driver (ODSP's `validateMessages` empties a from-misaligned trimmed range) means the range was trimmed → **`unresolvable`**.
 - First available op is past `from` — a trim gap at the anchor; the mark's batch (sequenced just after the reference point) was trimmed → **`unresolvable`**. A found batch resolves before this check, so a gap **on a miss** genuinely means the batch's ops are gone, not merely preceded by other clients' ops (which are still present at `from`).
```
anthony-murphy, line 174: Deep Review: This `pending` vs `unresolvable` bullet still reads "the mark's batch (sequenced just after the reference point) was trimmed → **`unresolvable`**." The parallel code comment was updated to "(at/after its lower bound)" and this same DEV.md section already describes `from = sequenceNumberLowerBound` as the inclusive anchor, so this line is the last "reference point" holdover — the author applied the change to the sibling code comment and missed this passage. Align it with the inclusive semantics.

```suggestion
- First available op is past `from` — a trim gap at the anchor; the mark's batch (sequenced at/after its lower bound) was trimmed → **`unresolvable`**. A found batch resolves before this check, so a gap **on a miss** genuinely means the batch's ops are gone, not merely preceded by other clients' ops (which are still present at `from`).
```

Issue #27461, opened 2026-06-01
Option createBlobPayloadPending defaulted on in v3.0 when min version is 2.40+


jason-ha:
## Broken API or pattern

Starting in 3.0, when `MinimumVersionForCollab` value is specified at 2.40 or higher, the `createBlobPayloadPending` feature will be enabled by default.

**Preparation References**

There are no required consumer changes.

**Packages**

`@fluidframework/container-runtime` (and those dependent on it)

**Expected Timeline**

v3.0

dannimad:
https://github.com/microsoft/FluidFramework/pull/28018 PR tracks/solves this issue

dannimad:
Won't be resolved this release.

PR #28018, opened 2026-08-19
createBlobPayloadPending default on 2.40.0 and explicit disable option

dannimad:
- turning `createBlobPayloadPending` on by default on >=2.40.0 versions by adding the option on `runtimeOptionsAffectingDocSchemaConfigMap`
- `createBlobPayloadPending` can now be set to false to be able to explicity disable the feature. 
- Added unit test coverage for the new default/explicit-disable behavior:
- `documentSchema.spec.ts`: verifies explicit `false` normalizes to `undefined` in the document schema summary, and `true` persists as `true`.
- `containerCompatibility.spec.ts` (new file): covers `getMinVersionForCollabDefaults` and `validateRuntimeOptions` for `createBlobPayloadPending` across version thresholds.
- Fixed a stale `containerRuntime.spec.ts` expectation that didn't account for the new default turning on at `minVersionForCollab >= 2.40.0`.
- Added a changeset (`.changeset/nice-hotels-act.md`) and regenerated the `container-runtime` legacy alpha/beta API reports to reflect the `createBlobPayloadPending` type widening (`true | undefined` → `boolean | undefined`).
- Fixed a downstream type-test break in `@fluidframework/aqueduct` caused by the widened type: acknowledged the intentional back-compat break via `typeValidation.broken` in `package.json` and regenerated the aqueduct type tests.
- Fixed `test-service-load`'s `optionsMatrix.ts` to include `false` in the `createBlobPayloadPending` option matrix (previously only `[true, undefined]`), and updated the `explicitSchemaControl` override check to be explicit (`=== true`) instead of truthy, since `false` is now a valid, meaningful value.
- Fixed 3 real-service e2e tests that broke due to the new default (these tests predated this change and assumed `undefined` meant "disabled", which is no longer true once `minVersionForCollab >= 2.40.0`):
  - `blobsisAttached.spec.ts`: changed the test matrix from `[undefined, true]` to `[false, true]` so the disabled case is still exercised.
  - `blobs.spec.ts`: same fix — matrix pinned `minVersionForCollab: "2.40.0"`, so `undefined` no longer disabled the feature; switched to `[false, true]` and decoupled `explicitSchemaControl` from the loop variable.
  - `gcSweepAttachmentBlobs.spec.ts` / `gcTombstoneAttachmentBlobs.spec.ts`: fixed a latent bug where a de-duped blob handle that's never referenced in a DDS never had its graph attached, so its upload never started under the payload-pending flow — added an explicit `attachGraph()` call to match the eager-upload behavior these GC/dedup tests rely on.

### DIFF
--- a/packages/runtime/container-runtime/src/containerCompatibility.ts
+++ b/packages/runtime/container-runtime/src/containerCompatibility.ts
@@ -157,7 +157,11 @@ const runtimeOptionsAffectingDocSchemaConfigMap: ConfigMap<RuntimeOptionsAffecti
 			// "3.0.0": { enableGCSweep: true },
 		},
 		createBlobPayloadPending: {
+			// Disabled by default for versions predating 2.40.0, since older clients don't understand the
+			// document format changes associated with pending blob payloads.
 			"1.0.0": undefined,
+			// Enabled by default starting at 2.40.0, since blob functionality is not exposed on the `@public`
+			// API surface, making this a reasonably safe version to flip the default at.
 			"2.40.0": true,
 		},
 	};
--- a/packages/runtime/container-runtime/src/containerCompatibility.ts
+++ b/packages/runtime/container-runtime/src/containerCompatibility.ts
@@ -161,6 +161,7 @@ const runtimeOptionsAffectingDocSchemaConfigMap: ConfigMap<RuntimeOptionsAffecti
 			// closed on the version where that will happen yet.  Probably a .10 release since blob functionality is not
 			// exposed on the `@public` API surface.
 			"1.0.0": undefined,
+			"2.40.0": true,
 		},
 	};
 
--- a/packages/runtime/container-runtime/src/containerCompatibility.ts
+++ b/packages/runtime/container-runtime/src/containerCompatibility.ts
@@ -157,9 +157,6 @@ const runtimeOptionsAffectingDocSchemaConfigMap: ConfigMap<RuntimeOptionsAffecti
 			// "3.0.0": { enableGCSweep: true },
 		},
 		createBlobPayloadPending: {
-			// This feature is new and disabled by default. In the future we will enable it by default, but we have not
-			// closed on the version where that will happen yet.  Probably a .10 release since blob functionality is not
-			// exposed on the `@public` API surface.
 			"1.0.0": undefined,
 			"2.40.0": true,
 		},
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -483,9 +483,9 @@ export interface ContainerRuntimeOptions {
 
 	/**
 	 * Create blob handles with pending payloads when calling createBlob.
-	 * Default is `undefined` (disabled) when `oldestClientSupported` is `<2.40.0`, but enabled when `oldestClientSupported` is `>=2.40.0`).
+	 * Default is `undefined` (disabled) when `minVersionForCollab` is `<2.40.0`, but enabled when `minVersionForCollab` is `>=2.40.0`.
 	 * When enabled (`true`), createBlob will return a handle before the blob upload completes.
-	 * Use explicit `false` to disable this feature when `oldestClientSupported` is `>=2.40.0`.
+	 * Use explicit `false` to disable this feature when `minVersionForCollab` is `>=2.40.0`.
 	 */
 	readonly createBlobPayloadPending: boolean | undefined;
 
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -482,10 +482,12 @@ export interface ContainerRuntimeOptions {
 	readonly explicitSchemaControl: boolean;
 
 	/**
-	 * Create blob handles with pending payloads when calling createBlob (default is `undefined` (disabled)).
+	 * Create blob handles with pending payloads when calling createBlob.
+	 * Default is `undefined` (disabled) when `oldestClientSupported` is `<2.40.0`, but enabled when `oldestClientSupported` is `>=2.40.0`).
 	 * When enabled (`true`), createBlob will return a handle before the blob upload completes.
+	 * Use explicit `false` to disable this feature when `oldestClientSupported` is `>=2.40.0`.
 	 */
-	readonly createBlobPayloadPending: true | undefined;
+	readonly createBlobPayloadPending: boolean | undefined;
 
 	/**
 	 * Controls automatic batch flushing during staging mode.
--- a/packages/runtime/container-runtime/src/containerRuntime.ts
+++ b/packages/runtime/container-runtime/src/containerRuntime.ts
@@ -1101,6 +1101,15 @@ export class ContainerRuntime
 			if (disallowedKeys.length > 0) {
 				throw new UsageError(`explicitSchemaControl must be enabled to use ${disallowedKeys}`);
 			}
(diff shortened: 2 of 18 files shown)


packages/runtime/container-runtime/src/test/containerCompatibility.spec.ts
jason-ha, line 1: Are there no other tests like these that these should be housed with?
@scottn12 @agarwal-navin ?

packages/runtime/container-runtime/src/containerCompatibility.ts
```diff
@@ -157,10 +157,8 @@ const runtimeOptionsAffectingDocSchemaConfigMap: ConfigMap<RuntimeOptionsAffecti
 			// "3.0.0": { enableGCSweep: true },
 		},
 		createBlobPayloadPending: {
-			// This feature is new and disabled by default. In the future we will enable it by default, but we have not
-			// closed on the version where that will happen yet.  Probably a .10 release since blob functionality is not
-			// exposed on the `@public` API surface.
 			"1.0.0": undefined,
+			"2.40.0": true,
```
jason-ha, line 165: No worthy notes - even historical here?

packages/runtime/container-runtime/src/containerRuntime.ts
```diff
@@ -482,10 +482,12 @@ export interface ContainerRuntimeOptions {
 	readonly explicitSchemaControl: boolean;
 
 	/**
-	 * Create blob handles with pending payloads when calling createBlob (default is `undefined` (disabled)).
+	 * Create blob handles with pending payloads when calling createBlob.
+	 * Default is `undefined` (disabled) when `oldestClientSupported` is `<2.40.0`, but enabled when `oldestClientSupported` is `>=2.40.0`).
 	 * When enabled (`true`), createBlob will return a handle before the blob upload completes.
+	 * Use explicit `false` to disable this feature when `oldestClientSupported` is `>=2.40.0`.
```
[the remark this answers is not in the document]
jason-ha: `minVersionForCollab` is the old name. `oldestClientSupported` is the correct new name.

packages/test/test-end-to-end-tests/src/test/gc/gcSweepAttachmentBlobs.spec.ts
```diff
@@ -307,6 +307,11 @@ describeCompat("GC attachment blob sweep tests", "NoCompat", (getTestObjectProvi
 				const blobHandle2 = toFluidHandleInternal(
 					await mainDataStore._runtime.uploadBlob(stringToBuffer(blobContents, "utf-8")),
 				);
+				// blobHandle2 is never referenced in a DDS in this test, so it wouldn't otherwise have its
+				// graph attached. When createBlobPayloadPending is enabled, the blob's upload doesn't begin
+				// until the handle's graph is attached, so we do so explicitly here to ensure it becomes a
+				// tracked (and immediately unreferenced) GC node like it would under the legacy blob flow.
+				blobHandle2.attachGraph();
```
jason-ha, line 314: This makes me nervous.
How confident are we that no consumer would need to respond like this?
How fast and obviously would their stuff break if they picked up 3.0 without setting setting to false or making similar change?
cc: @ChumpChief

packages/test/test-end-to-end-tests/src/test/blobs.spec.ts
```diff
@@ -97,7 +107,9 @@ const ContainerStateEventsOrErrors: ExpectedEvents = {
 	],
 };
 
-for (const createBlobPayloadPending of [undefined, true] as const) {
+// Use false/true (not undefined) since minVersionForCollab is pinned to 2.40.0 below, where
+// createBlobPayloadPending defaults to true; undefined would no longer exercise the disabled path.
+for (const createBlobPayloadPending of [false, true] as const) {
```
jason-ha: There is no compat testing in here?
Customer can use `undefined` so why shouldn't it be tested? Or is the optionsMatrix coverage sufficient?

packages/test/test-service-load/src/optionsMatrix.ts
jason-ha, line 155: In other code there is a comment that if `false` is used, then explicitSchemaControl must be true.
I was a little skeptical and now am more so.

packages/test/test-end-to-end-tests/src/test/gc/gcSweepAttachmentBlobs.spec.ts
dannimad: We need to work on a migration story. I haven't written it but I'm planning to write any action needed to migrate smoothly. 
As for this specific test change, I see some time differences on when the blob will be gc swept vs the legacy path but I don't see any consumer action about this.

packages/test/test-end-to-end-tests/src/test/blobs.spec.ts
dannimad: should be exercised when flag is false but we can be paranoic, I see value on it with little to no cost.

dannimad:
https://github.com/microsoft/FluidFramework/issues/27461

packages/test/test-end-to-end-tests/src/test/gc/gcSweepAttachmentBlobs.spec.ts
ChumpChief: attachGraph() isn't appropriate for a customer to call.  The more typical scenario that would produce garbage like this is to set the handle in a DDS and then delete it (like is done immediately below for the blobHandle1.

This change to the test doesn't seem quite right though, the point of this one is to test the de-duping logic so the attach of blobHandle1 below already does the job of tracking/uploading it?

packages/test/test-end-to-end-tests/src/test/gc/gcTombstoneAttachmentBlobs.spec.ts
```diff
@@ -201,6 +201,11 @@ describeCompat("GC attachment blob tombstone tests", "NoCompat", (getTestObjectP
 				const blobHandle2 = await mainDataStore._runtime.uploadBlob(
 					stringToBuffer(blobContents, "utf-8"),
 				);
+				// blobHandle2 is never referenced in a DDS in this test, so it wouldn't otherwise have its
+				// graph attached. When createBlobPayloadPending is enabled, the blob's upload doesn't begin
+				// until the handle's graph is attached, so we do so explicitly here to ensure it becomes a
+				// tracked (and immediately unreferenced) GC node like it would under the legacy blob flow.
+				toFluidHandleInternal(blobHandle2).attachGraph();
```
ChumpChief, line 208: Same here, if the intent is to attach the handle then do it by setting it in the DDS like blobHandle1 below, but I'm doubtful that it's intentional to attach all the handles in most of these tests.

ChumpChief (commented):
I'm somewhat inclined to defer this change, given that 3.0 is on a tight timeline and we might want more time to evaluate (and/or break up this change into test fixes, schema options, etc.).  I think it's generally in the right direction but just want to be extra careful since changes to document schema are very hard to recover from if there's any issue.

dannimad:
Matt and I took the decision to defer turning this feature on by default. Seems like high risk low reward given the timing and multiple points of failure my PR highlighted: tests failures when turning the flag on, misalignment on multiple flags (explicitSchemaControl, createBlobPayloadPending, type change allowing false value) and versions (what exactly happens on 2.40, 3.0.0 and in between) and given that changes to document schema are very hard to recover from if there's any issue.
I think we need a better story internally about how to turn things on by default. We didn't have anything like that for this and got us into this weird spot. I'll follow up by adding some items to the backlog separating the test bugs we found, changes into the schema options and adding the false value to the flag.

